PR #119, opened 2015-01-27
Add action buttons to the update view page.

ralphbean:
This includes allowing the editing of updates in 2f93ff9, which is essential.

This also adds "revoke" as a possible action on a request which got dropped in
the port from bodhi1 on accident (added back in in e5c248a).

### DIFF
--- a/bodhi/models/models.py
+++ b/bodhi/models/models.py
@@ -51,7 +51,7 @@ from bodhi.util import (
 )
 import bodhi.util
 from bodhi.models.enum import DeclEnum, EnumSymbol
-from bodhi.exceptions import LockedUpdateException
+from bodhi.exceptions import BodhiException, LockedUpdateException
 from bodhi.config import config
 from bodhi.bugs import bugtracker
 
@@ -213,6 +213,7 @@ class UpdateRequest(DeclEnum):
     stable = 'stable', 'stable'
     obsolete = 'obsolete', 'obsolete'
     unpush = 'unpush', 'unpush'
+    revoke = 'revoke', 'revoke'
 
 
 class UpdateSeverity(DeclEnum):
@@ -1457,7 +1458,7 @@ class Update(Base):
         self.untag()
 
         for build in self.builds:
-            koji.tagBuild(self.candidate_tag, build.nvr, force=True)
+            koji.tagBuild(self.release.candidate_tag, build.nvr, force=True)
 
         self.pushed = False
         self.status = UpdateStatus.unpushed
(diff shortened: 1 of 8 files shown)


bodhi/models/models.py
```diff
@@ -977,7 +982,12 @@ def set_request(self, action, request, pathcheck=True):
             notifications.publish(topic=topic, msg=dict(
                 update=self, agent=request.user.name))
             return
-
+        elif action is UpdateRequest.revoke:
+            self.revoke()
+            flash_log("%s has been obsoleted" % self.title)
```
lmacken: s/obsoleted/revoked/

bodhi/models/models.py
```diff
@@ -1448,12 +1458,42 @@ def unpush(self):
...
+                                 "one that is %s" % self.status.description)
+
+        # Remove the 'pending' koji tags from this update so taskotron stops
+        # evalulating them.
+        if self.request is UpdateRequest.testing:
+            self.remove_tag(self.release.pending_testing_tag)
+        elif self.request is UpdateRequest.stable:
+            self.remove_tag(self.release.pending_stable_tag)
+
+        # If we're revoking a request we need to set the status back to what it
```
lmacken: Revoking a 'stable' request on an update with the 'testing' status won't revert the status back to pending. I don't think we need to change any enums other than set the `request` to `None`.

bodhi/models/models.py
```diff
@@ -1448,12 +1458,42 @@ def unpush(self):
...
+        if self.release.testing_tag in current_tags:
+            self.status = UpdateRequest.testing
+        elif self.release.stable_tag in current_tags:
+            self.status = UpdateRequest.stable
+        else:
+            self.status = None
+
+        self.request = None
+
+        mail.send_admin('revoked', self)
```
lmacken: With FMN I'm not sure we need to send mails for most of the triggers in `bodhi/mail.py`. What do you think?

ralphbean: :+1:  I was just keeping the pattern in place that was already there (obsolete sent a mail, so revoke should too).

bodhi/models/models.py
ralphbean: Ah, the trick I think is that when the request got previously set to ``stable``, that code *also* sets the ``status`` to ``pending``, I think.  I guess I should double-check that.

ralphbean: Nope.  I was mistaken.  I'll remove this code.  Thanks!

lmacken: The `status` gets updated & the` request` gets unset towards the end of the push process by the `masher https://github.com/fedora-infra/bodhi/blob/pyramid/bodhi/masher.py#L242

bodhi/models/models.py
```diff
@@ -1448,12 +1458,31 @@ def unpush(self):
...
 
         self.pushed = False
         self.status = UpdateStatus.unpushed
         mail.send_admin('unpushed', self)
 
+    def revoke(self):
+        """ Remove pending request for this update """
+        log.debug("Revoking %s" % self.title)
+
+        if self.status is not UpdateStatus.pending:
```
lmacken: A `testing` update should probably be able to revoke a `stable` request.

ralphbean: Agreed.

lmacken:
:+1:

PR #122, opened 2015-01-28
Stop using TG identity.current.user_name in bodhi.mail.

ralphbean:


lmacken:
Good enough for now, thanks! I'm working on porting over a couple more features from bodhi1, and I'll look into pulling the agent from the request.

PR #123, opened 2015-01-28
Random work with respect to fedmsg messages

ralphbean:
I went through and tried to trigger each kind of message to make sure that
fedmsg.meta could handle the ones from bodhi2 correctly.  That receiving side
of this is in fedora-infra/fedmsg_meta_fedora_infrastructure#195.

Notably, in 6bc9017 I changed the ``Comment`` json to be backwards compatible
since some things (badges iirc) are looking for certain things there.

Some other things are unrelated but I was too lazy to break them out into their
own pull requests, like:  de9c092 and 8a098e0 and 7aa7c3e and 14cd78f -- sorry
for that.  Long day.

### DIFF
--- a/bodhi/masher.py
+++ b/bodhi/masher.py
@@ -210,8 +210,10 @@ class MasherThread(threading.Thread):
         self.init_state()
         self.init_path()
 
-        notifications.publish(topic="mashtask.mashing", msg=dict(repo=self.id,
-                              updates=self.state['updates']))
+        notifications.publish(topic="mashtask.mashing", msg=dict(
+            repo=self.id,
+            updates=self.state['updates'],
+        ))
 
         success = False
         try:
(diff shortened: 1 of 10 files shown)


bodhi/masher.py
```diff
@@ -589,10 +593,13 @@ def wait_for_sync(self):
...
                 log.info("master repomd.xml matches!")
-                notifications.publish(topic="mashtask.sync.done", msg=dict())
+                notifications.publish(topic="mashtask.sync.done", msg=dict(
+                    repo=self.id))
                 return
-            log.debug("master repomd.xml doesn't match! %s != %s",
-                      checksum, newsum)
+
+            # This seems like a pretty big deal..
+            log.error("master repomd.xml doesn't match! %s != %s for %r",
```
lmacken: Nope, it's expected that the repomd.xml won't match for up to 15min or so until the cron job syncs the repos to the master mirror.

ralphbean: Oh!  :athletic_shoe:

ralphbean: OK.  eb04cbf should take care of it.

ralphbean:
OK - rebased against ``pyramid`` to make this merge-able again.

Also, got lots the test suite running again (I totally forgot about it).

lmacken:
:money_with_wings:

PR #124, opened 2015-01-28
Add release to build if one is not already associated.

ralphbean:
The bulk of the change is in 7e6e25c.  I ran into a weird case when trying to
test the buildroot override fedmsgs where some updates would have their
``Build`` already in the db, but it didn't have a ``Release`` associated with
it.  The validator here would just vomit in that situation.

This adds logic that checks if the release isn't associated and tacks it on if
its not already there.


lmacken:
Looks good to me, thanks! :100:

PR #128, opened 2015-02-07
Port the updateinfo generation code to createrepo_c

lmacken:
:dancers:


bodhi/metadata.py
```diff
@@ -12,23 +12,23 @@
 # with this program; if not, write to the Free Software Foundation, Inc.,
 # 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 
-__version__ = '1.4'
+__version__ = '2.0'
```
ralphbean: :moneybag:

bodhi/metadata.py
```diff
@@ -12,23 +12,23 @@
...
 
 import os
-import glob
 import logging
 import shutil
 import tempfile
+import subprocess
```
ralphbean: :sob:

bodhi/metadata.py
```diff
@@ -170,181 +180,104 @@ def _fetch_updates(self):
...
             for rpm in rpms:
-                filename = "%s.%s.rpm" % (rpm['nvr'], rpm['arch'])
+                pkg = cr.UpdateCollectionPackage()
+                pkg.name = rpm['name']
+                pkg.version = rpm['version']
+                pkg.release = rpm['release']
+                pkg.epoch = rpm['epoch'] or '0'
+                pkg.arch = rpm['arch']
+
+                # TODO: how do we handle UpdateSuggestion.logout, etc?
```
ralphbean: Maybe submit an RFE to createrepo_c?

ralphbean: ... and link to the ticket from an inline comment here so future bodhi hackers will know?

bodhi/metadata.py
lmacken: Yeah, we ideally should be using the modifyrepo_c python interface

development.ini
```diff
@@ -327,6 +327,11 @@ admin_packager_groups = provenpackager releng security_respons
 mandatory_packager_groups = packager
 
 ##
+## updateinfo.xml configuraiton
+##
+updateinfo_rights =
```
ralphbean: Perhaps provide an example about what might go here?

bodhi/metadata.py
ralphbean: The actual usage of ``subprocess.check_call`` in the code here is not that bad.  It just references files that get written out to (not lots of arguments getting interpolated or anything like that).

bodhi/metadata.py
lmacken: Right now the metadata for this looks like `<reboot_suggested>True</reboot_suggested>`.

Sounds like we need a more flexible schema, maybe something like `<suggest type="logout"/>`

ralphbean:
:+1:

lmacken:
Thanks!

PR #143, opened 2015-04-21
Add fedmenu.

ralphbean:
For https://fedoraproject.org/wiki/Infrastructure/FY16_frontend

If you leave the fedmenu bits out of the config, it just never shows up.


lmacken:
Thanks! :money_with_wings:

PR #144, opened 2015-04-30
Be more explicit with Content-Security-Policy

ralphbean:
Just like in fedora-infra/datagrepper#166, firefox was *okay* with the
CSP we defined before, but chromium didn't like it.  This makes them
both happy by being more explicit.


lmacken:
Yup, LGTM. :+1:  Thanks!

PR #146, opened 2015-05-14
Generate the update ID/alias upon creation

lmacken:
As opposed to when it's initially pushed to testing, which is how it is currently done.

### DIFF
--- a/bodhi/tests/functional/test_updates.py
+++ b/bodhi/tests/functional/test_updates.py
@@ -375,7 +375,7 @@ class TestUpdatesService(bodhi.tests.functional.base.BaseWSGICase):
         self.assertEquals(up['date_modified'], None)
         self.assertEquals(up['date_pushed'], None)
         self.assertEquals(up['locked'], False)
-        self.assertEquals(up['alias'], None)
+        self.assertEquals(up['alias'], u'FEDORA-2015-0001')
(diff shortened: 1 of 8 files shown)


bodhi/tests/functional/test_updates.py
```diff
@@ -1147,7 +1147,7 @@ def test_edit_update(self, publish, *args):
         self.assertEquals(up['date_approved'], None)
         self.assertEquals(up['date_pushed'], None)
         self.assertEquals(up['locked'], False)
-        self.assertEquals(up['alias'], None)
+        self.assertEquals(up['alias'], u'FEDORA-2015-0002')
```
ralphbean: Do you want to use ``time.localtime`` here for the year?  I.e., will these tests break when we get to 2016?

ralphbean:
Looks good all-in-all.  :+1:

bodhi/tests/functional/test_updates.py
ralphbean: Or, use ``freezegun.freeze_time('2015-06-03')`` on the test.

lmacken: Yep, good call!

lmacken:
Thanks for the review!

PR #149, opened 2015-06-24
Use the kitchen unicode/byte converters to mitigate encoding nightmares

lmacken:
This fixes a number of issues that have popped up recently.

https://fedorahosted.org/bodhi/ticket/782
https://fedorahosted.org/bodhi/ticket/783
https://fedorahosted.org/bodhi/ticket/784

### DIFF
--- a/bodhi/model.py
+++ b/bodhi/model.py
@@ -45,6 +45,8 @@ except ImportError:
 
 import fedmsg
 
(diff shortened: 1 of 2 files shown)


ralphbean:
:+1: looks good here.

lmacken:
Thanks bean!

PR #150, opened 2015-06-30
Fix bodhi trac ticket #670 (bodhi2)

bitlord:
This should fix the https://fedorahosted.org/bodhi/ticket/670 for bodhi2 (develop)


lmacken:
Looks good to me, thanks!

PR #153, opened 2015-08-17
Support bodhi1 RSS feeds as best we can.

ralphbean:
The bodhi2 API is close to compatible with bodhi1 in many ways.  One
place that changed is that where previously ``?release=F19`` was used to
specify that you wanted updates only for F19, now we use
``?releases=F19``.  This patch adds some backwards compatibility so that
the old RSS feeds should work.


lmacken:
:+1:

PR #154, opened 2015-08-17
compat: Return the 'updateid' as well as the 'alias'

lmacken:
Things like blockerbugs rely on this, so we might as well remain compatible for now.


ralphbean:
Looks good to me!  Thanks!

PR #157, opened 2015-08-18
Declare CORS allowed-origins for cornice.

ralphbean:
This was the root cause of why the typeahead.js search box stopped
working.  Apparently, in the last year browsers got much more strict
about enforcing CORS and ajax requests from typeahead to the server were
failing that policy.

I implemented a proxy class to relay pyramid configuration to cornice.
The docstring of that class details why it is as complicated as it is.


lmacken:
Wow, nice catch & fix. :+1:

PR #158, opened 2015-08-18
Redirect to a package's updates if it exists.

ralphbean:
This doesn't fix any bugs, per se.  It's an improvement from a UX perspective.

I realized when working on the bodhi1->bodhi2 redirects that bodhi1 has
a behavior that I use all the time where you can visit
https://admin.fedoraproject.org/updates/nethack and be shown updates of
that package.  In bodhi2, this just gives you a 404, because there's no
update named just "nethack" -- it's a package.

This patch enhances the update_id validator to raise a 302 redirect
*if* the update_id isn't an update id but it *is* a package name.

We could take this or leave it.  It would be nice from a UX point of view, but
maybe confusing to someone if they're trying to use the API and then get a 302
redirect.  On the other hand, I find it hard to imagine somewhere where someone
is programmatically providing a package name where they meant to provide an
update id.  The window of confusion is pretty small.

### DIFF
--- a/bodhi/tests/functional/test_updates.py
+++ b/bodhi/tests/functional/test_updates.py
@@ -1554,3 +1554,12 @@ class TestUpdatesService(bodhi.tests.functional.base.BaseWSGICase):
 
         up = body['updates'][0]
         self.assertEquals(up['title'], u'bodhi-2.0-1.fc17')
+
+    def test_redirect_to_package(self):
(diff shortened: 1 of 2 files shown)


lmacken:
So I think that if someone tries to set_request or edit an update with just a package name instead of a build, it'll trigger this redirect... which seems like an edge case that will probably never get hit? 

:+1:

Issue #167, opened 2015-08-18
don't show taskotron results with status 'ABORTED'


tflink:
Taskotron uses the 'ABORTED' as a way to indicate that something went wrong with the job - the reported results are invalid and possibly incomplete.

The update display in bodhi2.0 seems to be displaying the latest result for an update, regardless of whether or not that update has a 'ABORTED' status.

Please change the logic which is querying resultsdb to ignore results with 'ABORTED' status

tflink:
An example that is currently showing this: https://bodhi.stg.fedoraproject.org/updates/libwmf-0.2.8.4-45.fc22

ralphbean:
Aw man!  I went through all the work to show that nice little trashcan icon too! :)

ralphbean:
This should be an easyfix.. I guess I want to make sure first that we really want to get rid of those results.

For instance, a packager might wonder what happened to that check.  If it is **required** that it passes in order for them to request stable status, but it doesn't show in the UI.. they'll come asking.  As things stand now (with the orange background and the trashcan icon) they can click on it to see that it started and have more information about what to do.

tflink:
We stopped notifying folks on ABORTED results a while back. The biggest thing for us right now is that it's papering over an issue with depcheck that causes ABORTED runs far more often than they should happen. depcheck runs often enough that it effectively doesn't matter in the long run as long as the noise of ABORTED runs doesn't make it to maintainers.

I suspect that it's more likely that folks will be alarmed about the ABORTED result than notice something was missing for 30 minutes.

Keep in mind that ABORTED results can show up for an update at any time before an update is pushed stable - it can go from PASSED to ABORTED and back to PASSED over the course of 3 depcheck runs. In reality, that means that depcheck stumbled and there's nothing to worry about but if there's a state change in the update, I think that indicates that something has gone wrong.

ralphbean:
OK.  I understand.

This applies to three places in bodhi's codebase:

- The javascript-built table in the update web UI (informational only).  This is all we've been discussing up to now.
- The server-side request validator that gets run when a packager requests that their update to from testing to stable.  It checks that all required checks have PASSED in taskotron.
- The masher-side gating that gets run at the start of the mash.  It also checks that all requried checks have PASSED in taskotron.

[...]

PR #168, opened 2015-08-18
Don't deal with ABORTED taskotron checks.

ralphbean:
As discussed in #167.

- Don't show ABORTED taskotron results in the update UI.
- Hide ABORTED taskotron results from server-side checks.


lmacken:
Solid. :+1:

PR #176, opened 2015-08-18
Disable stacks for the initial release.

ralphbean:
We're disabling stacks for the initial release because, while you can create
stacks, you can't automatically create updates *from* a stack (which was the
whole point).  We'll work on that for a later release.


lmacken:
:+1:

PR #178, opened 2015-08-18
Change default cert_prefix for push tool.

ralphbean:
The cert we're going to be using is just called "shell", so we might as well
make that the default to save on typing for releng.


lmacken:
:+1:

PR #180, opened 2015-08-19
Add the 'submitter' field back to Update JSON.

ralphbean:
This will make fedmsg.meta compatibility between bodhi1 and bodhi2 updates
*much* more simple.


lmacken:
:+1:

PR #189, opened 2015-08-19
Add the stub of a checkpoint feature for the mash.

ralphbean:


lmacken:
Yeah, we need this to ensure we're not commenting on updates, bugs, or sending update announcements multiple times when resuming a busted push.

ralphbean:
> we need this to ensure we're not commenting on updates, bugs, or sending update announcements multiple times when resuming a busted push.

OK, I applied it to those four methods in 929adc8.

ralphbean:
This fixes #214.

lmacken:
Excellent :)

PR #195, opened 2015-08-19
Make version and branch for Releases not nullable...

ralphbean:


lmacken:
We'll need to run this alembic upgrade in stg and prod

PR #206, opened 2015-08-20
masher: Don't mash or do any repo-related things to pending releases

lmacken:
The build-branched nightly cron job takes care of the mashing for us, so only
do the other stuff.


ralphbean:
:+1:

PR #210, opened 2015-08-20
Stop validating rpm versions.

ralphbean:
This check is currently broken, in that it won't let you submit a build of
foo-1.2.3 for F22 if you already submitted foo-1.2.3 for F23(!)  That's
crucial functionality we need in place.

This kind of validator is something that got ported over from bodhi1 which
had years of policy baked on to it.  Now that taskotron is doing depcheck
runs and bodhi2 is using the results of those runs to gate in the mash
process... we don't even need this validator anymore, do we?

### DIFF
--- a/bodhi/services/builds.py
+++ b/bodhi/services/builds.py
@@ -25,7 +25,6 @@ import bodhi.schemas
 import bodhi.security
 from bodhi.validators import (
     validate_nvrs,
-    validate_version,
     validate_uniqueness,
     validate_acls,
     validate_builds,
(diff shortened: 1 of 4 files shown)


lmacken:
:dog: :moneybag:

lmacken:
Yep, taskotron will take care of the upgrade path check, and bodhi will handle the gating

kparal:
Please note that upgradepath currently runs just for updates-pending, but not for updates-testing-pending, or even updates-testing. We can work on improving that, but that's the current state. Just making sure it's known.

Issue #233, opened 2015-08-20
Can't push updates to stable


swt2c:
I have a couple of updates that have been in testing for 7+ days, for example: https://bodhi.fedoraproject.org/updates/python-pyqtgraph-0.9.10-4.fc22

However, there is no button for allowing me to push this update to stable.

keszybz:
Also https://bodhi.fedoraproject.org/updates/execstack-0.5.0-8.fc23.

lmacken:
This issue has been resolved. Thanks for the report!

Issue #220, opened 2015-08-20
Bodhi2 sucks


leigh123linux:
Issues so far.

1. No button to push to stable (after seven day limit).
2. unable to create epel7 updates

nullr0ute:
The title is not useful. Also what you have above is two problems so please report them as such with proper details and feedback including NVRs.

leigh123linux:
The epel7 issue is reported here

https://github.com/fedora-infra/bodhi/issues/218

I can't submit tint2-0.12.2-1.el7

nullr0ute:
have you tried on using "fedpkg update" with at least python-fedora-0.5.5-1 and bodhi-client-0.9.12.2-3

leigh123linux:
Yes

```
$ fedpkg switch-branch epel7 
Switched to branch 'epel7'
[leigh@pc-main tint2]$ fedpkg update
/usr/lib/python2.7/site-packages/fedpkg/cli.py:169: DeprecationWarning: Commands._hash_file is deprecated and will be removed eventually.
    Please use Commands.lookasidecache.hash_file instead.
  hash = self.cmd._hash_file('bodhi.template', 'sha1')
Creating a new update for  tint2-0.12.2-1.el7 
Password for leigh123linux: 
Creating a new update for  tint2-0.12.2-1.el7 
ServerError(https://bodhi.fedoraproject.org/updates/, 500, Error returned from json module while processing https://bodhi.fedoraproject.org/updates/: Expecting value: line 1 column 1 (char 0))
Traceback (most recent call last):
  File "/usr/bin/bodhi", line 225, in main
    data = bodhi.save(**update_args)
  File "/usr/lib/python2.7/site-packages/fedora/client/bodhi.py", line 96, in wrapper
    result = method(*args, **kwargs)
  File "/usr/lib/python2.7/site-packages/fedora/client/bodhi.py", line 170, in save
    data=kwargs)
  File "/usr/lib/python2.7/site-packages/fedora/client/openidbaseclient.py", line 374, in send_request
    {'url': to_bytes(method), 'err': to_bytes(e)})
ServerError: ServerError(https://bodhi.fedoraproject.org/updates/, 500, Error returned from json module while processing https://bodhi.fedoraproject.org/updates/: Expecting value: line 1 column 1 (char 0))
Could not generate update request: Command 'bodhi --new --release epel7 --file bodhi.template tint2-0.12.2-1.el7 --username leigh123linux' returned non-zero exit status 255
```

[...]

PR #235, opened 2015-08-20
templates: Display the Push to Stable button if we meet the testing requirements

lmacken:
This should hopefully resolve issue #233 and part of #220


ralphbean:
:+1:

Issue #236, opened 2015-08-20
Properly handle BodhiExceptions from set_request 


lmacken:
From production:
```
Traceback (most recent call last):
  File "/usr/lib/python2.7/site-packages/bodhi/services/comments.py", line 186, in new_comment
    com = update.comment(author=author, anonymous=anonymous, **data)
  File "/usr/lib/python2.7/site-packages/bodhi/models/models.py", line 1355, in comment
    self.check_karma_thresholds(author)
  File "/usr/lib/python2.7/site-packages/bodhi/models/models.py", line 1531, in check_karma_thresholds
    self.set_request(UpdateRequest.stable, username)
  File "/usr/lib/python2.7/site-packages/bodhi/models/models.py", line 1002, in set_request
    raise BodhiException('. '.join(notes))
BodhiException: This critical path update has not yet been approved for pushing to the stable repository.  It must first reach a karma of 2, consisting of 0 positive karma from proventesters, along with 2 additional karma from the communit
y. Or, it must spend 14 days in testing without any negative feedback
```

lmacken:
```
Traceback (most recent call last):
  File "/usr/lib/python2.7/site-packages/bodhi/services/comments.py", line 186, in new_comment
    com = update.comment(author=author, anonymous=anonymous, **data)
  File "/usr/lib/python2.7/site-packages/bodhi/models/models.py", line 1355, in comment
    self.check_karma_thresholds(author)
  File "/usr/lib/python2.7/site-packages/bodhi/models/models.py", line 1531, in check_karma_thresholds
    self.set_request(UpdateRequest.stable, username)
  File "/usr/lib/python2.7/site-packages/bodhi/models/models.py", line 1026, in set_request
    raise BodhiException(flash_notes)
BodhiException: This update has not yet met the minimum testing requirements defined in the <a href="https://fedoraproject.org/wiki/Package_update_acceptance_criteria">Package Update Acceptance Criteria</a>

```

ralphbean:
So, this works now and shows up in a messenger.js bubble.

In the api, it is returned in a ``caveats`` field.

PR #243, opened 2015-08-20
Catch other bodhi exceptions when commenting.

ralphbean:
I think some people are hitting this when they comment
on an update that kicks it over the karma threshold, but
that threshold checking raises an exception.  It blocks the
comment from going through all together because the exception
bubbles all the way up to cornice.

This should make it so that the comment can still go through
if karma threshhold checking fails, but it will still log the
exception so we can dig more into them later.


lmacken:
:+1:

lmacken:
https://github.com/fedora-infra/bodhi/issues/236

PR #247, opened 2015-08-20
Add caveat API and guard against self-karma

ralphbean:
In the process I added a bunch of tests for the comments service.

The main feature here is a new "caveat" API extension that can be used to
return useful messages to users about what happened.  I added some js code
to display those to users in the web UI and we can eventually show them in
the bodhi CLI too.

Tests are added which check that you can't give yourself +1 karma, which
was kind of the point here.


lmacken:
Very nice! :+1:

Issue #202, opened 2015-08-20
Editing an update clears the package list, can't view update


mooninite:
I needed to edit an existing update that was pending to switch to a newer build.

After adding the newer package NVR to the update and clicking "edit" the page took me to a list of all updates. When I returned to my profile page the update line exists, but the package name has been wiped out and I cannot click on anything.

Screenshots
After editing the mediawiki-1.25.2-1.fc23 update to mediawiki-1.25.2-2.fc23: http://i.imgur.com/uYfqD0R.png
After editing the mediawiki-1.25.2-1.fc22 update to mediawiki-1.25.2-2.fc22: http://i.imgur.com/7fQJAc7.png

lmacken:
```
[Thu Aug 20 04:03:34.876486 2015] [:error] [pid 21940] 2015-08-20 04:03:34,876 DEBUG [bodhi][Dummy-1] validated = {'close_bugs': True, 'stable_karma': 2, 'csrf_token': u'<removed>', 'requirements': None, 'builds': [], 'autokarma': True, 'edited': u'mediawiki-1.25.2-1.fc23', 'suggest': <unspecified>, 'notes': u'* (T94116) SECURITY: Compare API watchlist token in constant time\\n* (T97391) SECURITY: Escape error message strings in thumb.php\\n* (T106893) SECURITY: Don\\'t leak autoblocked IP addresses on Special:DeletedContributions\\n* (T102562) Fix InstantCommons parameters to handle the new HTTPS-only policy of Wikimedia Commons.\\n* (T100767) Setting a configuration setting for skin or extension to false in LocalSettings.php was not working.\\n* (T100635) API action=opensearch json output no longer breaks when $wgDebugToolbar is enabled.\\n* (T102522) Using an extension.json or skin.json file which has a "manifest_version" property for 1.26 compatability will no longer trigger warnings.\\n* (T86156) Running updateSearchIndex.php will not throw an error as page_restrictions has been added to the locked table list.\\n* Special:Version would throw notices if using SVN due to an incorrectly named variable. Add an additional check that an index is defined.\\n\\n', 'request': <testing>, 'bugs': [1253281, 1253280, 1252194], 'unstable_karma': -2, 'require_bugs': True, 'require_testcases': True, 'type': <security>, 'severity': <unspecified>}
[...]

PR #248, opened 2015-08-20
Check and test for empty builds for edited updates.

ralphbean:
Confirmed via both the test suite *and* local hand testing of the web UI.


lmacken:
Yep, :+1: Fixes #202

PR #252, opened 2015-08-20
Remove these hrefs which make location.reload() unstable.

### LINKED ISSUE #244: "Push to Stable" silently redirects me back to the homepage
https://bodhi.fedoraproject.org/updates/python-shove-0.6.3-1.el6

This update has yet to meet the testing requirements. When I hit Push to Stable (the button is there because I'm in releng), it just redirects me back to the homepage without any information.

ralphbean:
This should fix #244.


lmacken:
:+1:

PR #281, opened 2015-08-21
Include test_case JSON in update output.

### LINKED ISSUE #270: Return the test cases in the update json
The test cases are associated with each package, so we should iterate over each package for each build and throw the test cases into the json results as well.

This was originally filed accidentally in the wrong place as fedora-infra/fedmsg_meta_fedora_infrastructure#313.

ralphbean:
Fixes #270.


lmacken:
:+1: :helicopter:

PR #291, opened 2015-08-21
Display the update in the comment fragment.

### LINKED ISSUE #215: Mention the package along with the 'comment'/'feedback'
The list of feedback does not contain the information on which package the feedback was given to, see this picture:
screenshot from 2015-08-20 08-57-23

Instead this information is hidden in the link "23 days ago [...]".
Would it be possible to e.g. write the package name below the "23 days ago [...]"?

This would help a lot and one does not hove with the mouse to the link to see which package the feedback was given too.

ralphbean:
But only on the user profile page and the comment list page.

It doesn't make sense to show the name of the update if you're looking
**at** the page for **that** update already... so suppress it there.

Fixes #215.

### DIFF
--- a/bodhi/models/models.py
+++ b/bodhi/models/models.py
@@ -25,6 +25,13 @@ from textwrap import wrap
 from datetime import datetime
 from collections import defaultdict
 
+try:
+    # python3
+    from urllib.parse import quote
+except ImportError:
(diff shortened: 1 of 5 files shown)


lmacken:
Solid. :100:

PR #292, opened 2015-08-21
Fix the ACLs issue for set_request and edit.

### LINKED ISSUE #282: [SECURITY ISSUE] Why can other people push my packages?
The permissions seem buggered on the new bodhi

https://bodhi.fedoraproject.org/updates/tint2-0.12.2-1.fc22

thofmann pushed tint2 to testing (he has no perms for this action)

https://bodhi.fedoraproject.org/updates/tint2-0.12.2-1.fc22#comment-313306

In return I unpushed (I have no right or perms to do so).

https://bodhi.fedoraproject.org/updates/parcimonie.sh-0-0.4.20150804gitc009937.el7

ralphbean:
The root cause was that we have this ``validate_acls`` validator which is
used in two different ways but it was only programmed to handle one of
those two. It became critical when that first manner would fail silently if
it was called in the second manner.

I've fleshed out that validator to now handle its two different code paths
and left lots of comments about my findings.  We should get this out asap..

Longer term, we should break the validator into two different ones that
handle each role accordingly (for updates and lists of builds).

Fixes #282.

### DIFF
--- a/bodhi/tests/functional/test_updates.py
+++ b/bodhi/tests/functional/test_updates.py
@@ -94,10 +94,15 @@ class TestUpdatesService(bodhi.tests.functional.base.BaseWSGICase):
         update['edited'] = update['builds']  # the update title..
         update['builds'] = []
         res = self.app.post_json('/updates/', update, status=400)
-        self.assertEquals(len(res.json_body['errors']), 1)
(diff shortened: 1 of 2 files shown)


lmacken:
Looks good to me, :bow:

PR #303, opened 2015-08-21
Reworded "karma" and "karma_critpath" questions to make more sense.

### LINKED ISSUE #222: Feedback form slightly confusing
Using ticks and crosses in the Feedback section is a bit confusing when the questions are of the form "Does this update break the basic functioning of your system?" because a tick usually means "yes" but I don't want to say "yes, it breaks the basic functioning"

I realise the columns are labelled "FAIL" and "PASS" but that's only in popup text that isn't immediately visible.

How about:

"Is the basic functioning of the system still OK after this update?" so that you can give a positive answer

Conan-Kudo:
The questions for general karma and critical path karma really don't make a lot of sense and are somewhat confusing. I've rewritten them to be more logical in the context of how they are presented on screen.

This should fix #222.


lmacken:
Fine with me, thanks for the patch!

PR #312, opened 2015-08-22
Remove builds and bugs when unchecked.

### LINKED ISSUE #260: Replace checkboxes with a "Remove" button
See the attached screenshot. I am copy-pasting in 83 builds and made a typo. Not sure what to do now; if I clear the checkbox is it enough to remove the wrongly entered build ID or is it going to keep it associated with the update forever?

I think this is bad UI and makes the user not feel in control. I am a perfectionist and want to make the update look perfect before pressing submit; the UI doesn't easily let me do that.

If it had a "Remove" button at the end of each update ID instead of the checkbox, it would make it much clearer what is happening and make the user feel in control again.

bodhi-mistyped-build-name

ralphbean:
This fixes #260 by actually removing typo'd items from the list when creating
new updates.

### DIFF
--- a/bodhi/static/js/update_form.js
+++ b/bodhi/static/js/update_form.js
@@ -86,6 +86,14 @@ $(document).ready(function() {
         });
     }
 
+    // Callback to remove a checkbox when it is unchecked.
+    // https://github.com/fedora-infra/bodhi/issues/260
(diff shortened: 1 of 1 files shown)


lmacken:
:white_check_mark:

PR #313, opened 2015-08-22
Surface set_request errors to the user.

### LINKED ISSUE #258: unpush seems to be broken
Clicking on the "unpush" button leads to the home page, with no visible effect.

ralphbean:
This is the last pull-request of many that will fix #258.  The others involved
standardizing the error reporting popups and suppressing some errant redirects.


lmacken:
:+1:

PR #331, opened 2015-08-23
issue #330 add exception

b10n1k:


ralphbean:
Unless I'm mistaken, this will raise an exception every time the system comments on a bug, no matter how long it is, no?

IMHO, the logic should go:

- Before we try to actually add the comment to bugzilla, first check the length of our comment.
- If it is longer than that max length, then break it into multiple comments.  Also log a warning with the whole comment so we can see where these ultra-long comments are coming from (that seems excessively long).

PR #346, opened 2015-08-24
Disable the ability to request an update be pushed it's current state…

### LINKED ISSUE #345: Disable the ability to push to a state that it's currently in.
For example, if an update is already stable, no one should be able to request that it be pushed to stable...

https://bodhi.fedoraproject.org/updates/FEDORA-2015-13060

lmacken:
… (fixes #345)


lmacken:
It's a simple fix and has 2 unit tests. Merging :)

PR #354, opened 2015-08-25
Bugfix:  keep the "type" set when editing an update.

### LINKED ISSUE #349: On update edit the "type" is lost
When "edit" a submitted update, the Type is lost, need to be check again.

Notice: the error message is not really clear: "required"

ralphbean:
Looks like a typo.  This should fix #349.

### DIFF
--- a/bodhi/templates/new_update.html
+++ b/bodhi/templates/new_update.html
@@ -183,7 +183,7 @@ ${update.notes}
                     % for value in types:
                     <label class="radio-inline">
                       <input type="radio" name="type" value="${value}"
-                      % if update and update.type.description == value:
+                      % if update and update.type.value == value:
                       checked="checked"
                       % endif
                       > ${value}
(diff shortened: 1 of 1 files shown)


pypingou:
Looks safe to me (but I don't know bodhi's sources that well)

ralphbean:
That type is an instance of the [UpdateType](https://github.com/fedora-infra/bodhi/blob/develop/bodhi/models/models.py#L202-L206) object which gets its behavior from [this class](https://github.com/fedora-infra/bodhi/blob/develop/bodhi/models/enum.py#L22-L46).

ralphbean:
On the other hand, now that I look more closely, this may not actually fix the bug:

```
>>> bodhi.models.UpdateType().security.description
'security'
>>> bodhi.models.UpdateType().security.value
'security'
```

ralphbean:
OK, it works locally with this.

pypingou:
Still ok for me (with the same limitation as before)

lmacken:
Hmm, I don't see how this changes anything. `description` and `value` are the same, and we use `description` all over the place still...

Issue #355, opened 2015-08-25
Update the mouse cursor when doing an action


pypingou:
When clicking on the button to request that an update goes to testing/stable, the button changes color but nothing happens.

I would be nice to change the mouse cursor as well as an additional way of showing that something is going on.

It would be setting the mouse cursor to ``wait`` while in ``cabbage.spin()`` and setting it back to ``default`` in ``cabbage.finish()``.

PR #356, opened 2015-08-25
Change the mouse cursor when doing an action

pypingou:
Currently after pressing the button to make an update go to testing/stable
the bodhi logo (the cabbage) starts spinning.
This is really nice but easy to miss.
With this change the mouse cursor will also change, making it clearer that
something is going on.

Fixes https://github.com/fedora-infra/bodhi/issues/355

### DIFF
--- a/bodhi/static/js/cabbage.js
+++ b/bodhi/static/js/cabbage.js
@@ -8,6 +8,7 @@ Cabbage.prototype.frequency = 50;
 Cabbage.prototype.finish_time = 400;  // Finish up in 1 second
 
 Cabbage.prototype.spin = function() {
+    document.body.style.cursor = 'wait';
     var self = this;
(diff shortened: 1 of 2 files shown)


ralphbean:
It doesn't seem to work here.  Other interactions on the page are overriding the cursor setting.

ralphbean:
OK, that gets it working locally.

pypingou:
:+1: for me

lmacken:
Sweet, :+1:

lmacken:
This feature doesn't work for me :(

ralphbean:
What browser, @lmacken?  Firefox?

ralphbean:
Just tested and it works in both Firefox and Chrome here.  Try to hard-reload the page to get the latest CSS assets?

PR #359, opened 2015-08-25
Change yum references to dnf

jwboyer:
Yum is deprecated in Fedora and is replaced by dnf.  We should probably just
reference dnf directly in the bugzilla comments.

### DIFF
--- a/development.ini
+++ b/development.ini
@@ -268,7 +268,7 @@ pkgdb_url = https://admin.fedoraproject.org/pkgdb
 
 initial_bug_msg = %s has been submitted as an update to %s. %s
 stable_bug_msg = %s has been pushed to the %s repository. If problems still persist, please make note of it in this bug report.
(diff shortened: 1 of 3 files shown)


ralphbean:
:+1: here.  This would also need to be changed [in ansible](https://infrastructure.fedoraproject.org/cgit/ansible.git/tree/roles/bodhi2/base/templates/) for staging and production.

lmacken:
There are a few other references to yum, like in `bodhi/mail.py`. We should probably also make it conditional on Fedora only, not EPEL.

jwboyer:
OK, that makes sense.  I'm not sure how to really make it conditional on Fedora at first glance though.  I'd be happy to poke at it with a tip or two.

ralphbean:
I'll work on updating this.

ralphbean:
Closed in favor of #535.  Thanks Josh!

Issue #377, opened 2015-08-26
sent emails not encoded correctly


nirik:
From: https://fedorahosted.org/bodhi/ticket/789

one example is: 

https://bodhi.fedoraproject.org/updates/FEDORA-2015-13400

the notes seems to use "smart quotes" which aren't encoded right... ie: 
“Select commit”
not
"Select commit"

pypingou:
I wonder if the way to fix it is not to use:
https://github.com/pypingou/pagure/blob/master/pagure/lib/notify.py#L200

This allows to declare the encoding of the email sent.

PR #385, opened 2015-08-26
Try converting the final email body to utf8 before sending.

ralphbean:
This is intended to address #377, but I'm not sure if it will fix it or not.

Also made a cosmetic commit to address readability.

### DIFF
--- a/bodhi/mail.py
+++ b/bodhi/mail.py
@@ -431,7 +431,7 @@ def send_mail(from_addr, to_addr, subject, body_text, headers=None):
         for key, value in headers.items():
             msg.append('%s: %s' % (key, to_bytes(value)))
     msg += ['Subject: %s' % subject, '', body_text]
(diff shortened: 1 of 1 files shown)


bodhi/mail.py
```diff
@@ -431,7 +431,7 @@ def send_mail(from_addr, to_addr, subject, body_text, headers=None):
         for key, value in headers.items():
             msg.append('%s: %s' % (key, to_bytes(value)))
     msg += ['Subject: %s' % subject, '', body_text]
-    body = '\r\n'.join(msg)
+    body = to_bytes('\r\n').join(msg)
```
lmacken: You're only converting the \r\n to bytes

lmacken:
:+1:

pypingou:
I think there is another way to do this

pypingou:
cf https://github.com/pypingou/pagure/blob/master/pagure/lib/notify.py#L200

PR #386, opened 2015-08-26
Remove calendar icon.

### LINKED ISSUE #374: Calendar icon in +Override form should open date chooser widget
When submitting a build override the Expiration date field has a calendar icon, but surprisingly clicking it does nothing. Clicking in the textbox opens a date chooser widget.

I'm sure I can't be the only person who expects the calendar icon to open the date chooser widget. Currently the calendar icon serves no purpose, I know that 2015-08-27 is a date, and the title for the textbox also says "Expiration date" just in case I can't figure out what that textbox is meant to contain.

ralphbean:
Super simple.  #374 complains correctly that the calendar icon doesn't do
anything.  The datepicker lib we're using doesn't easily extend to allow being
triggered by a button, so let's just remove the icon so we can move on to
hacking on other bugs/features.

Fixes #374.

### DIFF
--- a/bodhi/templates/override.html
+++ b/bodhi/templates/override.html
@@ -66,7 +66,6 @@
             <span class="pull-left">Expiration date</span>
           </div>
           <div class="form-group noui input-group">
-            <span class="input-group-addon glyphicon glyphicon-calendar"></span>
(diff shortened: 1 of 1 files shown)


pypingou:
:+1: for me

PR #387, opened 2015-08-26
Be more specific with time in our taskotron queries.

### LINKED ISSUE #362: Use `since` param when querying ResultsDB
You query ResultsDB for all the results available for the particular update. That is not "wrong" per-se, but when an update is changed, after the tests are finished, you might be showing (e.g.) PASSED result for Depcheck, although it is more like "not tested yet".

ResultsDB's query engine supports `since` param, that takes a timestamp in ISO8601(YYYY-MM-DDTHH:MM:SS) format, and when used ResultsDB only returns results that are newer than the datetime specified.

I'm not sure if the update has something like "last modified" value, but if the last-modified is anything else than "created at", I'd suggest using the `since` param, to ensure the right information is shown.

ralphbean:
* Fixes #362.

This should cut down on any depcheck false positives we report/use and could
also maybe make our query-load lighter on resultsdb.

### DIFF
--- a/bodhi/models/models.py
+++ b/bodhi/models/models.py
@@ -1538,7 +1538,14 @@ class Update(Base):
         requirements = list(requirements)
 
         try:
-            results = bodhi.util.taskotron_results(settings, title=self.title)
+            # https://github.com/fedora-infra/bodhi/issues/362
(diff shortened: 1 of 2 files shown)


lmacken:
:money_with_wings:

PR #388, opened 2015-08-26
Add a "Releases" dropdown to the toplevel navbar.

### LINKED ISSUE #350: No browsing of updates based on releases
Please add one more tab on right upper top "Releases" that will allow to browse updated submitted for that particular release. 
e.g. Like in old interface, In new interface when clicked on "Releases" tab, it should show all active releases like for "Fedora 23" should show https://bodhi.fedoraproject.org/updates/?releases=F23

ralphbean:
Provide the data in the template for this by hanging another lazily
evaluated
'reified' property to the request that returns the same data that the
``bodhi.services.releases`` cornice service does.

* Fixes #350.

### DIFF
--- a/bodhi/__init__.py
+++ b/bodhi/__init__.py
@@ -12,14 +12,11 @@
 # along with this program; if not, write to the Free Software
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
 
-import pkg_resources
-
 from collections import defaultdict
 from dogpile.cache import make_region
 from sqlalchemy import engine_from_config
 
 from pyramid.settings import asbool
-from pyramid.decorator import reify
 from pyramid.security import unauthenticated_userid
 from pyramid.config import Configurator
 from pyramid.authentication import AuthTktAuthenticationPolicy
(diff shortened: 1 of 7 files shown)


bodhi/__init__.py
```diff
@@ -78,6 +75,10 @@ def get_buildinfo(request):
     return defaultdict(dict)
 
 
+def get_releases(request):
+    import bodhi.services.releases
+    return bodhi.services.releases.query_releases_html(request)['releases']
```
lmacken: Doesn't this mean every time any page is rendered, it has to go through the `releases` cornice `Service`, the `ListReleaseSchema`, 3 different validators, and then a db query? Seems a bit heavyweight. Can't we do a simple `@classmethod`+ class-level-cache hack like we do with `Release.get_tags`? https://github.com/fedora-infra/bodhi/blob/develop/bodhi/models/models.py#L316-L330

ralphbean: Yup, let's do that.

lmacken:
Pushed a commit to optimize this with a new `Release.all_releases` class method.

Issue #348, opened 2015-08-25
Unable to disable various karma/bug/test requirements


lmacken:
I am unable to edit an update and uncheck any of the Bug/Test/Karma requirement checkboxes, or even the "Automatically close bugs".

lmacken:
Here is the POST payload when trying to disable require bugs, testcases, and auto-request stable.
```
{"":[],"term":[],"builds":"python-shove-0.6.4-1.fc23","bugs":["781425","1256191"],"notes":"ChangeLog: https://bitbucket.org/lcrees/shove/raw/f612167f6c2b6eb9d9c39852a09db8477dc0aa8f/CHANGELOG","type":"enhancement","severity":"unspecified","suggest":"unspecified","close_bugs":"on","autokarma":[],"stable_karma":"1","unstable_karma":"-2","require_bugs":[],"require_testcases":[],"requirements":"rpmlint","csrf_token":"fb74195dcb6437428f2dea8025f1282bee48d73b","edited":"python-shove-0.6.4-1.fc23"}
```

jwboyer:
Issue 341 might be the same as this?

ralphbean:
It might be that the colander schema has ``missing=True`` for these fields.  A patch like this might do it:

```diff
diff --git a/bodhi/schemas.py b/bodhi/schemas.py
index 79a60a0..2ec8dc4 100644
--- a/bodhi/schemas.py
+++ b/bodhi/schemas.py
@@ -174,7 +174,6 @@ class SaveUpdateSchema(CSRFProtectedSchema, colander.MappingSchema):
     )
     autokarma = colander.SchemaNode(
         colander.Boolean(),
-        missing=True,
     )
     stable_karma = colander.SchemaNode(
         colander.Integer(),
@@ -201,11 +200,9 @@ class SaveUpdateSchema(CSRFProtectedSchema, colander.MappingSchema):
     )
     require_bugs = colander.SchemaNode(
         colander.Boolean(),
-        missing=True,
     )
     require_testcases = colander.SchemaNode(
         colander.Boolean(),
-        missing=True,
     )
```

lmacken:
Yeah, I think you might be right @ralphbean. It seems that when the checkboxes are enabled, we get `"on"`, else `[]`, hopefully colander will convert the empty list to `False` for us.

PR #400, opened 2015-08-27
Differentiate singleton and sequence checkboxes(lists).

### LINKED ISSUE #397: unable to use bodhi web ui, get 3 errors: ACL validation mechanism wsa unable to determin ACLs...
Per summary, I've been unable to use bodhi web ui at all since yesterday... Tried several browsers (chrome, firefox mostly), (force) reloads, nothing seems to help,

Whenever I try to submit a new update or edit and existing one, it fails with 3 popup errors:
* ACL validation mechanism was unable to determine ACLs
* You may not specify an empty list of builds (I assure you the checked list is non-empty)
* Required (yep, that's it, just one word)

ralphbean:
This fixes #397.

The bug was introduced in 2c3bba9e when we were trying to fix something else.
The bug stems from the way the list of submitted builds gets generated in
javascript before it is POSTed to the server.  It looks like this when you have
one checkbox unchecked:

```javascript
[
        "foo-1.2.3-1.fc23",
        "bar-0.1.0-1.fc23",
        false,
        "baz-900-2.fc23",
]
```

The code in this patch only pushes that ``false`` value into the list **if**
the checkbox is specially marked with a ``data-singleton`` HTML5 attribute
(which I made up).  So, now:

- lists of checkboxes should *not* get the ``false`` value
  pushed (fixing #397).
- singleton checkboxes will still get their ``false`` value
  included (keeping #348 in good shape).

### DIFF
--- a/bodhi/static/js/forms.js
+++ b/bodhi/static/js/forms.js
@@ -75,7 +75,12 @@ Form.prototype.data = function() {
         if (this.type == 'radio' && ! this.checked) {
             // pass - don't add unchecked radio buttons to the submission
         } else if (this.type == 'checkbox' && ! this.checked) {
-            data[this.name].push(false);
+            // Handle series and singletons differently.
+            // The checkbox lists of bugs and builds are series.
+            // The single checkboxes for require_bugs, etc.. are singletons.
+            if (this['data-singleton'] == true) {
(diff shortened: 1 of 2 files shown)


lmacken:
:+1:

PR #407, opened 2015-08-27
Ignore the taskotron user like we did autoqa.

ralphbean:


lmacken:
Works for me.

kparal:
I think taskotron is marked as a bot account in FAS. You could use this information when deciding, instead of relying on a manually maintained user list. Just an idea.

PR #408, opened 2015-08-27
Link to various places from the list of builds on the update page.

### LINKED ISSUE #197: make it easy to go from an update to a list of updates for a given package in that update
We used to make the package names clickable in bodhi1 where people could easily see a list of updates for a specific package.

ralphbean:
* Fixes #197.


lmacken:
Nice, :+1:

Issue #314, opened 2015-08-22
Move bug commenting into async process


ralphbean:
This should help a lot with performance, but let's leverage fedmsg and move bug commenting (the initial bug comment) out of ``Update.new(..)`` and into a new fedmsg consumer that gets run on bodhi-backend01.  It can listen for when new updates are created by the frontend and then do some of the more expensive tasks (like commenting on bugs..).

lmacken:
Great idea!

lmacken:
I think a ton of the slowness is coming from querying the wiki in `Package.fetch_test_cases`, so we should do that in this consumer as well.

cicku:
And I think it's really helpful to reduce the time consumption of waiting.

Issue #301, opened 2015-08-21
Investigate performance issues


lmacken:
We need to profile submitting an update and setting a request, as these things seem to be taking a very long time

ralphbean:
Some ideas:

* We have **zero** postgres indexes in place.  Judicious use of those could help.  If done incorrectly it could make things worse, so let's make sure to measure *before* and *after* speed times.
* The query code in ``bodhi/services/updates.py`` has some double-joins in there which could be slowing things down hard.  Some db-oriented thinking might be able to come up with a better way.

lmacken:
Querying the wiki for test cases is extremely slow. We should move this to
a backend process (issue #314)

[Thu Aug 27 19:03:29.658288 2015] [:error] [pid 11744] 2015-08-27 19:03:29,658 INFO  [bodhi][Dummy-3] Querying the wiki for test cases
[Thu Aug 27 19:05:37.587026 2015] [:error] [pid 11744] 2015-08-27 19:05:37,586 INFO  [bodhi][Dummy-3] Done querying for test cases

lmacken:
Okay, wiki & bug interactions have been moved to a backend fedmsg consumer on bodhi-backend02. 

I think the next step is optimizing our model with indexes and maybe some additional eager-loading, etc.

pypingou:
Note: we also have some tables that have just no primary key, just from a quick look:
- stack_group_table
- stack_user_table
- update_bug_table
- update_cve_table
- user_group_table
- user_package_table
- bug_cve_table

They should be composite primary keys but I think they should be there and since primary keys are indexed we might gain some things there as well.

ralphbean:
Cool.  Will be interesting to see what adding those does to performance.  As before, let's make sure to measure before and after speed times before we merge anything into `develop`.

ralphbean:
Here's an example of a *really* slow query:  https://bodhi.fedoraproject.org/updates/?packages=gnome-2048

ralphbean:
So check these raw sql queries out.  They both come from the ``get_updates`` function in ``bodhi/services/updates.py``.  The first one is the ``count_query`` and the second one is the call to ``query.all()`` that actually gets the updates.

```sql
SELECT count(DISTINCT updates.id) AS count_1
FROM updates
    LEFT OUTER JOIN releases AS releases_1 ON releases_1.id = updates.release_id
    LEFT OUTER JOIN comments AS comments_1 ON updates.id = comments_1.update_id
[...]

Issue #272, opened 2015-08-20
Submitting large updates takes too long


kalev:
I worked on submitting a 83 package GNOME update today and Bodhi took a long long time to process it. threebean asked me to file it to see if it's possible to optimize it.

The experience for me was like this:

1) Submitted the update. 15 minutes later the web UI (or a proxy server?) timed out and returned without any errors.

2) I tried to click Submit again twice after the time out. It timed out similarly each time, after 15 or so minutes.

3) ~1.5 hours after first clicking Submit, it finally posted bugzilla comments and tagged the packages.

cicku:
TBH, I doubt that the model originated from Bodhi1 has flaw.

Bodhi2 doesn't enhance that, but even worser. In fact, Bodhi should immediately receive new update requests and submit them to cache-like queue(with return like `pending` in Bodhi1 but not the same meaning. It means updates received, and they are being handled), and notifies users when updates are submitted exactly to testing channel.

lmacken:
I have a feeling a lot of this time is spent querying the wiki for test cases. More profiling is needed, but we do plan on offloading a lot of the slow external API calls to the wiki, bugzilla, etc to a fedmsg consumer on the backend. See #314 for more info.

lmacken:
Yep, as I suspected, the wiki is slow.

[Thu Aug 27 19:03:29.658288 2015] [:error] [pid 11744] 2015-08-27 19:03:29,658 INFO  [bodhi][Dummy-3] Querying the wiki for test cases
[Thu Aug 27 19:05:37.587026 2015] [:error] [pid 11744] 2015-08-27 19:05:37,586 INFO  [bodhi][Dummy-3] Done querying for test cases


We plan on moving this code to a backend process to speed this up (issue
#314)

cicku:
OK. 

So I think maybe it's time to use separate instance to hold test cases. Traversing the wiki consumes too much time as we can see. I don't see reason test cases should be put on wiki because in many years lots of users just treat Fedora wiki as a place for testing instead of diffusing knowledge.

lmacken:
Wiki querying has been moved to a seperate fedmsg consumer, which will drastically speed things up.

lmacken:
I believe @kalev said he has had pretty good experiences with large updates recently.

PR #419, opened 2015-08-28
Move some rhbz bug handling into the backend.

ralphbean:
This should hopefully help some with performance, and it paves the way for
moving other slow cross-system integration stuff into offline processes.

This handles (as you can see) both new updates which get their request set to
testing as well as updates that get edited.  For updates that get edited, we
only sync up new bugs that have been added.  Pre-existing bugs don't get
bothered.

A couple not related changes worth mentioned:

* I moved the bodhi masher into ``bodhi.consumers.masher``, for namespacing.
* When new updates are created, I switched the order of assigning the alias and
  setting the request.  Setting the request publishes a fedmsg message, and
  it's helpful to have the alias already assigned at that point (so we can use
  the alias in fedmsg consumers).

This changeset is relevant to tickets #314, #301 and #272.


lmacken:
Awesome. :+1:

PR #420, opened 2015-08-28
Eject from mash instead of failing when tag actions fail.

### LINKED ISSUE #418: traceback when determining tag actions
We hit this traceback in a mash today.

```                                                                      
[2015-08-28 17:41:52][    fedmsg    INFO] Updating comps                                                            
[2015-08-28 17:41:52][     bodhi    INFO] Running ['git', 'pull']
[2015-08-28 17:41:52][     bodhi    INFO] Running ['make']
[2015-08-28 17:41:52][    fedmsg    INFO] Mashing f21-updates-testing
[2015-08-28 17:41:52][    fedmsg    INFO] Running post-request actions on updates
[2015-08-28 17:41:52][     bodhi    INFO] Running ['mash', '-o', '/mnt/koji/mash/updates/f21-updates-testing-150828.
[2015-08-28 17:41:52][    fedmsg    INFO] Generating testing digest for F21
[2015-08-28 17:42:20][    fedmsg   ERROR] Cannot find relevant tag for python-affinity-0.1.0-5.fc23: ['f23', 'f23-up
[2015-08-28 17:42:20][    fedmsg   ERROR] Exception in MasherThread(f23-updates-testing)
File "/usr/lib/python2.7/site-packages/bodhi/masher.py", line 298, in work
self.determine_tag_actions()
File "/usr/lib/python2.7/site-packages/bodhi/masher.py", line 513, in determine_tag_actions
raise Exception
Exception
```                                                                      

So, we raise that exception there.  Perhaps instead we could try to gracefully recover by just ejecting whichever updates fail at that point.

The sticking point there is that ``determine_tag_actions`` currently happens *after* we try to update the security bugs.  So, maybe we want to reverse the order there if we can.. and do the tag actions first, and then the security bugs.  Possible?

ralphbean:
Potentially fixes #418.  See that issue for more details.

Before accepting this, we need to think if changing the set of updates being
mashed at this point in the mash would be otherwise problematic.  It just seems
like a waste to kill the whole mash if one update has oddball tags at this
point in the process.

If we *don't* want to accept this, we might want to consider cherry-picking
f168b46 and bba5123, which are probably nice on their own.

### DIFF
--- a/bodhi/consumers/masher.py
+++ b/bodhi/consumers/masher.py
@@ -294,11 +294,7 @@ class MasherThread(threading.Thread):
 
             self.update_security_bugs()
 
-            if not self.state.get('tagged', False):
-                self.determine_tag_actions()
-                self.perform_tag_actions()
-                self.state['tagged'] = True
-                self.save_state()
+            self.determine_and_perform_tag_actions()
 
             self.expire_buildroot_overrides()
             self.remove_pending_tags()
@@ -492,7 +488,12 @@ class MasherThread(threading.Thread):
                 for bug in update.bugs:
                     bug.update_details()
 
-    def determine_tag_actions(self):
+    @checkpoint
+    def determine_and_perform_tag_actions(self):
+        self._determine_tag_actions()
+        self._perform_tag_actions()
+
+    def _determine_tag_actions(self):
         tag_types, tag_rels = Release.get_tags()
         for update in sorted_updates(self.updates):
             if update.status is UpdateStatus.testing:
@@ -518,7 +519,7 @@ class MasherThread(threading.Thread):
                     self.move_tags.append((from_tag, update.requested_tag,
                                            build.nvr))
 
-    def perform_tag_actions(self):
+    def _perform_tag_actions(self):
         self.koji.multicall = True
         for action in self.add_tags:
             tag, build = action
--- a/bodhi/consumers/masher.py
+++ b/bodhi/consumers/masher.py
@@ -292,9 +292,9 @@ class MasherThread(threading.Thread):
             if self.request is UpdateRequest.stable:
                 self.perform_gating()
 
-            self.update_security_bugs()
-
             self.determine_and_perform_tag_actions()
+
+            self.update_security_bugs()
 
             self.expire_buildroot_overrides()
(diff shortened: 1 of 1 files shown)


nirik:
I would think ejecting the update with bad tags would be ok, as long as we knew to look/fix it for the next push. I'm not sure how to highlight that however. Could we note it in the update with something like "Please fix the tags on this package before re-reqesting" ?

And/or emit a fedmsg on it and have something that collects/mails those at the end for attention?

Ideally we would figure out how anything got in this state and just fix that. :)

ralphbean:
> I would think ejecting the update with bad tags would be ok, as long as we knew to look/fix it for the next push. I'm not sure how to highlight that however.   Could we note it in the update with something like "Please fix the tags on this package before re-reqesting" ?

Yeah!  The current code in the masher has it such that when an update is ejected:

- We comment on the update stating that it was ejected and why.  Maintainers should get an email about this.
- We reset the ``request`` to ``None`` so that it won't automatically make its way into the next push.
- We currently log it as a *warning*, but if we made that an *error*, then fedmsg admins would get an email about it (perhaps not as useful.. that's the wrong crew).
- We publish a fedmsg message about it so everyone else can know (admins/releng/whoever) via FMN.



> And/or emit a fedmsg on it and have something that collects/mails those at the end for attention?

Yup.  We could look to [this datagrepper query](https://apps.fedoraproject.org/datagrepper/raw?topic=org.fedoraproject.prod.bodhi.update.eject&delta=1270000) for the list of updates that got ejected after the fact.

> Ideally we would figure out how anything got in this state and just fix that. :)

Yeah, Luke was telling me that we hit this the other day because there were some new builds in advance of the ones in the mash and the *oscar* user garbage-collected the old ones (the ones to be mashed).. and so we wound up with a weird tag state like this.

nirik:
This one was caused by an update that was already stable being requested for testing. ;( Which it shouldn't be allowed to do.

ralphbean:
> Which it shouldn't be allowed to do.

Agreed.  We still need to take that up as a separate issue.

lmacken:
Let's do it. :+1:

Issue #406, opened 2015-08-27
Web UI: related bugs and test cases lists are effectively duplicated between the 'feedback item' list and the 'update information', confusing and wastes space


AdamWill:
In the web UI, the whole list of related test cases and bugs is shown twice: once in read-only form at the top, and once as part of the 'leave your comment' UI below the other comments.

The first time I saw this I found myself clicking on the +/- values in the top section and wondering why I couldn't edit them, then realized it was all duplicated down at the bottom.

This seems unnecessarily confusing and a waste of space; wouldn't it be more efficient to combine the displays?

ralphbean:
@AdamWill can you provide an example URL for this?

kalev:
https://bodhi.fedoraproject.org/updates/FEDORA-2015-13858

AdamWill:
yeah, that's a good one...also the same anaconda one from #405.

ralphbean:
> This seems unnecessarily confusing and a waste of space; wouldn't it be more efficient to combine the displays

Sure, and thanks for the links!  This may have made lots of sense when looking at updates with relatively few test cases and builds.. but it does look super confusing on those mega updates.  Will come up with something!

ralphbean:
OK, #421 didn't change much drastically, but I think the updates look a lot better now that the columns are all straightened out.

If you think it still needs work, feel free to re-open the issue and elaborate... but I'm not sure what else to do (we'll need to call in the design team ninjas!)

PR #421, opened 2015-08-29
Move ``</div>``...

ralphbean:
.. so that the read-only Test Cases are enclosed within the right column.
This should really help with #406.


lmacken:
:+1:

PR #422, opened 2015-08-29
Push testcase feedback into its own row

### LINKED ISSUE #405: Web layout for comments on updates with lots of 'karma items' is messy
Screenshot of messy update

See how the longer karma item names flow off the edge of the screen and there's a lot of wasted space.

Perhaps it would look better if they were shown in sort of horizontal tile form under the comment?

ralphbean:
.. in the "comment" fragment.  This should fix #405.

### DIFF
--- a/bodhi/templates/fragments.html
+++ b/bodhi/templates/fragments.html
@@ -33,10 +33,10 @@
       </div>
 
       <div class="panel-body">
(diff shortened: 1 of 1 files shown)


lmacken:
Nice :+1:

Issue #263, opened 2015-08-20
Bodhi metrics for EPEL are missing


robert-scheck:
https://bodhi.fedoraproject.org/metrics only displays the Bodhi metrics for Fedora releases, not for EPEL. Can we have them back? Additionally, the "old" metrics were more verbose (package most often being updated, packager with the most updates, etc.).

ralphbean:
Ah, check out this page for more info:  https://bodhi.fedoraproject.org/releases/

We haven't yet figured out how/where to promote that stuff in the UI yet.

robert-scheck:
I am sorry, I can't see how the URL would provide the same metrics (also with a timeframe) and some numbers, packages and packagers. https://bodhi.fedoraproject.org/releases/EL-5 doesn't really look usable regarding the diagram here (Firefox 38 ESR).

PR #423, opened 2015-08-29
Add more stats and links(!) to the release pages.

### LINKED ISSUE #403: enhance release page with moar links
Like this page https://bodhi.fedoraproject.org/releases/F23

Let's extend the table on the left to include some more stats:

- number of updates in stable
- number of updates in testing
- number of security updates
- number of bugfix updates

etc.. every type, status, and request.. let's put them there.

*importantly*, let's make each one of those entries *also* a link to a page that shows you the *list* of each of those kinds of updates, so people don't have to manually frob the querystring to find what they want.

ralphbean:
* Fixes #403.
* Helps with #263.

### DIFF
--- a/bodhi/services/releases.py
+++ b/bodhi/services/releases.py
@@ -20,7 +20,14 @@ from sqlalchemy import func, distinct
 from sqlalchemy.sql import or_
 
 from bodhi import log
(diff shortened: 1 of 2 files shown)


lmacken:
Excellent! :+1:

Issue #426, opened 2015-08-30
Expat error in making testing digest


nirik:
Hitting this trying to push f23-updates-testing:

```
Aug 29 19:32:05 bodhi-backend01.phx2.fedoraproject.org fedmsg-hub[2521]: [2015-08-29 19:32:05][    fedmsg   ERROR] Exception in MasherThread(f23-updates-testing)
Aug 29 19:32:05 bodhi-backend01.phx2.fedoraproject.org fedmsg-hub[2521]: Traceback (most recent call last):
Aug 29 19:32:05 bodhi-backend01.phx2.fedoraproject.org fedmsg-hub[2521]: File "/usr/lib/python2.7/site-packages/bodhi/masher.py", line 318, in work
Aug 29 19:32:05 bodhi-backend01.phx2.fedoraproject.org fedmsg-hub[2521]: self.generate_testing_digest()
Aug 29 19:32:05 bodhi-backend01.phx2.fedoraproject.org fedmsg-hub[2521]: File "/usr/lib/python2.7/site-packages/bodhi/masher.py", line 627, in generate_testing_digest
Aug 29 19:32:05 bodhi-backend01.phx2.fedoraproject.org fedmsg-hub[2521]: self.add_to_digest(update)
Aug 29 19:32:05 bodhi-backend01.phx2.fedoraproject.org fedmsg-hub[2521]: File "/usr/lib/python2.7/site-packages/bodhi/masher.py", line 620, in add_to_digest
Aug 29 19:32:05 bodhi-backend01.phx2.fedoraproject.org fedmsg-hub[2521]: update, use_template='maillist_template')):
Aug 29 19:32:05 bodhi-backend01.phx2.fedoraproject.org fedmsg-hub[2521]: File "/usr/lib/python2.7/site-packages/bodhi/mail.py", line 315, in get_template
Aug 29 19:32:05 bodhi-backend01.phx2.fedoraproject.org fedmsg-hub[2521]: h = get_rpm_header(build.nvr)
Aug 29 19:32:05 bodhi-backend01.phx2.fedoraproject.org fedmsg-hub[2521]: File "/usr/lib/python2.7/site-packages/bodhi/util.py", line 72, in get_rpm_header
Aug 29 19:32:05 bodhi-backend01.phx2.fedoraproject.org fedmsg-hub[2521]: result = koji_session.getRPMHeaders(rpmID=rpmID, headers=headers)
[...]

PR #429, opened 2015-08-31
Retry and log errors about koji rpm headers.

ralphbean:
This should in part address #426.


lmacken:
Yup. :+1:

ralphbean:
Noting here that there was an underlying koji bug:  https://lists.fedorahosted.org/pipermail/koji-devel/2015-August/000059.html

PR #437, opened 2015-09-01
Allow submitting multiple NVRs for an override.

### LINKED ISSUE #415: Can not create multiple buildroot overrides
In bodhi 2 the "Candidate build" field in the form for creating buildroot overrides only accepts one value. In bodhi 1 this could be a long list of builds. It is not uncommon that you need the same buildroot override in all releases (except rawhide). This mean that I now have to fill the bodhi 2 form 6 times for F23, F22, F21, EPEL7, EPEL6 and EPEL5, instead of just once in bodhi 1. If I need overrides for two packagas I need to fill the form 12 times instead of once. This takes a lot of time that could be used for more interesting things.

Also, if you list more than one build in the "Candidate build" field and press submit, the submission just silently fails without giving any feedback. There is no information given telling the sumitter why nothing happens.

ralphbean:
Under the hood, it creates separate overrides for each NVR
and the JS web client redirects you appropriately.

Added two new tests to cover these cases.  All previous tests cases still pass.

Fixes #415.

### DIFF
--- a/bodhi/models/models.py
+++ b/bodhi/models/models.py
@@ -1984,7 +1984,8 @@ class BuildrootOverride(Base):
         build = data['build']
 
         if build.override is not None:
-            request.errors.add('body', 'nvr', 'This build already is in a buildroot override')
+            request.errors.add('body', 'nvr',
+                               '%s is already in a override' % build.nvr)
             return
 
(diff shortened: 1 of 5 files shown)


lmacken:
LGTM! :100: :+1: :100:

PR #438, opened 2015-09-01
Make buildroot overrides searchable.

### LINKED ISSUE #208: Buildroot override search/list
In bodhi1, there was no way to search a package that is already added as buildroot override but I can browse through the list to check it.

Is there any plans for this in bodhi2? At least a way to find if someone has already added any version of a package in buildroot override?

ralphbean:
* Fixes #208.


lmacken:
Solid, :+1: :dancer: :dash:

PR #439, opened 2015-09-01
Don't show invalid dates in datepicker.

### LINKED ISSUE #302: bodhi expiration dialogue proposes dates which are invalid
This is really minor, but when picking the duration for a buildroot override, a dialogue pops up which can be used to pick any date in the future. But only dates less than 31 days into the future are allowed. So the dialogue could simply grey out the dates too far in the future, like it already does for dates in the past.

Also, the error is badly worded: "Expiration date may not be longer than 31" → "Exp. date may not be more than 31 days into the future" or something.

ralphbean:
* Fixes #302.


pypingou:
Looks sane :+1:

ralphbean:
Thanks Pierre!

PR #440, opened 2015-09-01
Demote this log statement.

ralphbean:
We get it *all the time* in the mash for trying to send email to users that
don't have a FASUSERNAME@fedoraproject.org email address.

That's a problem -- we want them to get mail, so we should get their preferred
email from FAS or FMN and use that.

In the meantime, let's just complain about it with a **warning** in the logs so
that we (the bodhi admins) don't get a ton of spam system email about it.  It
makes it hard to sift through the emails for more important problems.


pypingou:
Big :+1: :)

PR #441, opened 2015-09-01
Expand latest_candidates to accept testing=true/false.

### LINKED ISSUE #327: add build completion to buildroot override add
From: https://fedorahosted.org/bodhi/ticket/788

"Small regression in Bodhi 2.0: builds completion fails to work in the override page.

I tried to submit an override for qpid-cpp-0.32-8.fc23 (due to boost rebuild induced ABI break), and it failed in both discovering the name of the source package and the nvr."

ralphbean:
This then gets used by the new buildroot override form to search for builds
that either either updates candidates **or** in the testing tags or the pending
testing tags.

* Fixes #327.

### DIFF
--- a/bodhi/static/js/override_form.js
+++ b/bodhi/static/js/override_form.js
@@ -41,11 +41,11 @@ $(document).ready(function() {
     };
 
     // These next couple blocks of code wire up the auto-complete search for
-    // candidates in the override form.  Two technologies are at play here.  The
+    // builds in the override form.  Two technologies are at play here.  The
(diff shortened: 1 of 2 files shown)


pypingou:
Code looks sane, but I'm still a noob for bodhi2 :)

ralphbean:
> I'm still a noob for bodhi2

Gotta hack it! :D

pypingou:
>>  I'm still a noob for bodhi2

>Gotta hack it! :D

Def ! :)

PR #443, opened 2015-09-01
Show a spinner while we have no bug titles.

ralphbean:
We don't use anything fancy to clear the spinner.  People just have to reload
the page.  BUT, it's a quick and easy replacement for the ugly ``None`` we had
on the page before..


pypingou:
Looks alright, I just caught myself wondering when/where the spinner would be replaced, is there a loop in the JS?

ralphbean:
> I just caught myself wondering when/where the spinner would be replaced, is there a loop in the JS?

Nope, it never gets removed.  It's just there to let people know that something is working on it.

pypingou:
Ok, :+1: for me

PR #444, opened 2015-09-01
Don't try to tag stuff with the u'' tag.

### LINKED ISSUE #411: Can't change build in EL6 update
I just wanted to edit all my fedfind updates from fedfind-1.4.2-1.(dist) to fedfind-1.5-1.(dist). I was able to edit the updates for EL7, F21, F22 and F23, but I just can't edit the [update for EL6](https://bodhi.fedoraproject.org/updates/fedfind-1.4.2-1.el6). However I try, it fails with "Unable to create update. Invalid taginfo: "

ralphbean:
This fixes #411.  Turns out that el6 has no pending tags and that's okay, but
it was breaking us anytime someone tried to edit an el6 update.

I think this works in the other cases where tagging is done through
``Update.add_tag`` and ``Update.remove_tag``, because those use
``koji.tagBuild(...., force=True)``.  The force, I imagine, is hiding whatever
underlying issue this generates.  Anyways, we can only get faster by not
bothering koji with requests to tag stuff into empty tags.

### DIFF
--- a/bodhi/models/models.py
+++ b/bodhi/models/models.py
@@ -725,7 +725,15 @@ class Update(Base):
                     b = Build(nvr=build, package=package)
                     b.release = up.release
                     db.add(b)
-                koji.tagBuild(up.release.pending_testing_tag, build)
+
+                if up.release.pending_testing_tag:
+                    koji.tagBuild(up.release.pending_testing_tag, build)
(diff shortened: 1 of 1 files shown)


pypingou:
Looks sane :+1:

Issue #435, opened 2015-09-01
bodhi gives error 'null' when pushing to stable


morxa:
I'm trying to push [this update](https://bodhi.fedoraproject.org/updates/parcimonie.sh-0-0.4.20150804gitc009937.el7) to EPEL7 stable. When I press "Push to stable", it gives an error "null" and the push aborts.

#### How to reproduce
1. Press "Push to stable" on the linked update.

#### How reproducible
Not sure. I can push the same update for all other releases, but I cannot push to EPEL7. I don't know if that's always the case though.

#### Expected behavior
Successful push or a meaningful error message.

#### Actual behavior
Error _null_ and the push fails.

ralphbean:
I wasn't able to find anything helpful in the logs.  Can you ping me (``threebean``) on freenode sometime when you have a moment so we can try some interactive debugging?

ralphbean:
After adding some debug statements, I was able to get this traceback out:

```python
2015-09-01 21:10:22,542 DEBUG [bodhi][Dummy-2] Attempting to set request <stable>            
2015-09-01 21:10:22,545 ERROR [bodhi][Dummy-2] Failed to set request.                        
Traceback (most recent call last):                                                           
  File "/usr/lib/python2.7/site-packages/bodhi/services/updates.py", line 121, in set_request
    update.set_request(action, request.user.name)                                            
  File "/usr/lib/python2.7/site-packages/bodhi/models/models.py", line 1039, in set_request  
    raise BodhiException(flash_notes)                                                        
BodhiException: None                                                                         
```

ralphbean:
@morxa, if you try again you should get a reasonable error message about the EPEL update policy.  We were [missing a config value in production](https://infrastructure.fedoraproject.org/cgit/ansible.git/commit/?id=56856cf89cb2f6e242543ad382ffdcf8add2ed82) and since it was absent, you got ``null`` back instead.

morxa:
I confirm it's working now, I'm getting a reasonable error message. Thanks!

PR #445, opened 2015-09-01
Set epel test requirement message.

ralphbean:
This is mostly debugging I used to figure out what was wrong with #435.

The real fix for that is deployed in [this ansible commit](https://infrastructure.fedoraproject.org/cgit/ansible.git/commit/?id=56856cf89cb2f6e242543ad382ffdcf8add2ed82).


lmacken:
Cool, :+1:

PR #448, opened 2015-09-02
Indicate that an update is a critpath update.

### LINKED ISSUE #339: Make it obvious that an update is in the Critical Path
It doesn't look like there is anything obvious in the per-update UI that says that it is a Critical Path update.

ralphbean:
This adds a little alert panel on the update view for critical path updates.

It would show up just below the Locked panel and the Push/Edit buttons, but
above the Update Notes.

Fixes #339.


lmacken:
Awesome, :+1:

PR #452, opened 2015-09-02
Prune duplicates from /latest_candidates

### LINKED ISSUE #450: overrides auto-search shows f21 candidates twice
overrides
You can see that f21 candidate appearing twice. I also checked few other f21 packages and they are listed twice.

Please fix to show them once.

ralphbean:
Recently adding the extra calls to look for testing and pending testing builds
too created the possibility of duplicates here.

Fixes #450.


pypingou:
:+1: 

One of these days I'll need to test if using ``set()`` is quicker than iterating through a list :)

lmacken:
Yup, :+1:

PR #455, opened 2015-09-02
Support "tab" as a select trigger

### LINKED ISSUE #347: Disallow empty checkboxes in the new update form
screenshot from 2015-08-24 17-45-07

### LINKED ISSUE #391: Autocomplete click versus tab key
I built new versions of gap-pkg-fga for F22+ tonight.  I went to bodhi to create F22 and F23 updates.  I typed in the name "gap-pkg-fga", saw that the autocomplete box just below matched and hit tab to go to the versions box.  Apparently, what I was supposed to do was click on the text in the autocomplete box.  In any case, I typed in the package versions for the update, added some update text, and pushed the Submit button, which then grayed out.  There was no error message, but several minutes later, there was no notification of success either.  I checked my update list.  Not there.  Figuring something had gone wrong, I went back and tried again, this time figuring out that I need to click on the autocomplete text, which then populated the version box for me.  Nice!  Except when I clicked the Submit button, I got this:

"Unable to create update. (IntegrityError) duplicate key value violates unique constraint "builds_nvr_key" DETAIL: Key (nvr)=(gap-pkg-fga-1.3.0-1.fc22) already exists. 'INSERT INTO builds (nvr, inherited, package_id, release_id, update_id) VALUES (%(nvr)s, %(inherited)s, %(package_id)s, %(release_id)s, %(update_id)s) RETURNING builds.id' {'package_id': 10296, 'release_id': None, 'update_id': None, 'nvr': u'gap-pkg-fga-1.3.0-1.fc22', 'inherited': False}

That error message is clearly intended for bodhi developers, not bodhi users.  How about logging messages like that somewhere where the bodhi developers can find them, and showing the users something more comprehensible, like "Hey, dummy, you already created an update with this package!"

Anyway, the updates finally showed up on my update list, so they were created successfully, but I thought I'd better file this anyway so that I can vote for (1) make tab out of the package name box == click on an exact match on the autocomplete list, and (2) change the duplicate update message to something comprehensible to mere mortals.  Thanks!  I like the looks of bodhi 2.

ralphbean:
- in the new update form which fixes #391.
- also disallow empty checkboxes which fixes #347.

### DIFF
--- a/bodhi/static/js/update_form.js
+++ b/bodhi/static/js/update_form.js
@@ -219,18 +219,20 @@ $(document).ready(function() {
         $("#builds-adder input").val('');  // Clear the field
         return false;
     }
+
+    // If you press "enter", make it count
     $("#bugs-adder input").keypress(function (e) {
         if (e.which == 13) { return add_bugs(); }
     });
     $("#builds-adder input").keypress(function (e) {
         if (e.which == 13) { return add_builds(); }
     });
-    $("#bugs-adder button").click(function(e) {
-        return add_bugs();
-    });
-    $("#builds-adder button").click(function(e) {
-        return add_builds();
-    });
+    // If you "tab" away from the input, make it count.
+    $("#bugs-adder input").focusout(function(e) { return add_bugs(); });
+    $("#builds-adder input").focusout(function(e) { return add_builds(); });
+    // Or, if you click the "+" button, make it count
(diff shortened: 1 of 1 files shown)


pypingou:
:+1: for me

lmacken:
Nice! :+1:

PR #460, opened 2015-09-02
Add Update.date_{testing,stable} db timestamps and refactor Update.days_in_testing

lmacken:
... instead of scraping the comments.

### DIFF
--- a/alembic/versions/13cfca635b99_add_update_date_testing_stable_.py
+++ b/alembic/versions/13cfca635b99_add_update_date_testing_stable_.py
@@ -1,0 +1,24 @@
+"""Add update.date_testing/stable timestamps
+
+Revision ID: 13cfca635b99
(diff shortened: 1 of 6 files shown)


lmacken:
Just pushed an untested update to the alembic script to set the timestamps on *all* updates in the db.

ralphbean:
Testing the alembic upgrade destroyed my box.  After some fighting, I ended up needing to use this [.yield_per(...)](http://docs.sqlalchemy.org/en/rel_1_0/orm/query.html?highlight=yield_per#sqlalchemy.orm.query.Query.yield_per) method to get things running.

Like we talked about in g+, it takes forever to run so we'll leave it commented out for now and run that part by hand later to avoid hours of downtime.

lmacken:
:+1:

PR #468, opened 2015-09-03
Check the size of the comment before sending it to bugzilla

pypingou:
With this commit we won't make the comment but instead we raise an
exception and log it if a comment has more than 65535 characters.

Fixes https://github.com/fedora-infra/bodhi/issues/330

### DIFF
--- a/bodhi/bugs.py
+++ b/bodhi/bugs.py
@@ -41,6 +41,13 @@ class FakeBugTracker(BugTracker):
     comment = update_details = modified = close = on_qa = __noop__
 
 
+class InvalidComment(Exception):
+    """ Exception thrown when the comment posted is invalid (for example
+    too long.
+    """
+    pass
+
+
 class Bugzilla(BugTracker):
 
     def __init__(self):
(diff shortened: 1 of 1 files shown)


bodhi/bugs.py
```diff
@@ -64,8 +71,13 @@ def getbug(self, bug_id):
 
     def comment(self, bug_id, comment):
         try:
+            if len(comment) > 65535:
+                raise InvalidComment("Comment is too long: %s" % comment)
             bug = self.bz.getbug(bug_id)
             bug.addcomment(comment)
+        except InvalidComment as err:
+            log.exception("Invalid comment for bug #%d" % bug_id)
+            log.exception(err)
```
ralphbean: So, ``log.exception`` already automatically logs the exception *in addition* to whatever string you pass it.  SO, this second ``log.exception(err)`` statement would log the exception twice (three times in total).

Maybe get rid of line 80 and replace line 79 with:

```python
log.exception("Comment too long for bug #%d:  %s" % (bug_id, comment))
```

That way we can see what the comment is so we can reason about it.

ralphbean:
LGTM!

lmacken:
Looks good :) Thanks!

Issue #221, opened 2015-08-20
URLs should use update ID not title


jwakely:
Related to #186, the URLs in the bodhi web UI should use the update ID not the title formed by joining all the builds.  Not only is the ID stable, but it doesn't get too long for the web server (or some part of the bodhi scripts running on the server) to handle, giving 404 errors when trying to edit a large update with very many builds.

i.e. instead of 

[...]

PR #469, opened 2015-09-03
Make the link to updates display the title but rely on the alias

pypingou:
This should make sure we do no have unbarrable links anywhere but that
we are using the title instead.

Fixes https://github.com/fedora-infra/bodhi/issues/221

### DIFF
--- a/bodhi/util.py
+++ b/bodhi/util.py
@@ -438,7 +438,7 @@ def update2html(context, update):
     title = update.title
     alias = update.alias
 
-    url = request.route_url('update', id=alias)
+    url = request.route_url('update', id=alias or title)
     settings = request.registry.settings
     max_length = int(settings.get('max_update_length_for_ui', 30))
     if len(title) > max_length:
--- a/bodhi/util.py
+++ b/bodhi/util.py
@@ -435,18 +435,10 @@ def request2html(context, request):
 def update2html(context, update):
     request = context.get('request')
 
-    if hasattr(update, 'title'):
-        title = update.title
-    else:
-        title = update['title']
+    title = update.title
+    alias = update.alias
 
-    alias = None
-    if hasattr(update, 'alias'):
-        alias = update.alias
-    else:
-        alias = update.get('alias')
-
-    url = request.route_url('update', id=alias or title)
+    url = request.route_url('update', id=alias)
     settings = request.registry.settings
     max_length = int(settings.get('max_update_length_for_ui', 30))
     if len(title) > max_length:
--- a/bodhi/util.py
+++ b/bodhi/util.py
@@ -435,8 +435,15 @@ def request2html(context, request):
 def update2html(context, update):
     request = context.get('request')
 
-    title = update.title
-    alias = update.alias
+    if hasattr(update, 'title'):
+        title = update.title
+    else:
+        title = update['title']
+
+    if hasattr(update, 'alias'):
+        alias = update.alias
+    else:
+        alias = update['alias']
 
     url = request.route_url('update', id=alias or title)
     settings = request.registry.settings
--- a/bodhi/util.py
+++ b/bodhi/util.py
@@ -440,7 +440,13 @@ def update2html(context, update):
     else:
         title = update['title']
 
-    url = request.route_url('update', id=title)
+    alias = None
+    if hasattr(update, 'alias'):
+        alias = update.alias
+    else:
+        alias = update.get('alias')
+
+    url = request.route_url('update', id=alias)
     settings = request.registry.settings
     max_length = int(settings.get('max_update_length_for_ui', 30))
     if len(title) > max_length:
--- a/bodhi/util.py
+++ b/bodhi/util.py
@@ -446,7 +446,7 @@ def update2html(context, update):
     else:
         alias = update.get('alias')
 
-    url = request.route_url('update', id=alias)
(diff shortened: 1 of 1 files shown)


pypingou:
Tests are failing for me locally, but seems un-related to the change

bodhi/util.py
```diff
@@ -440,7 +440,13 @@ def update2html(context, update):
     else:
         title = update['title']
 
-    url = request.route_url('update', id=title)
+    alias = None
+    if hasattr(update, 'alias'):
```
lmacken: Updates will always have the alias attribute

bodhi/util.py
```diff
@@ -440,7 +440,13 @@ def update2html(context, update):
...
+        alias = update.alias
+    else:
+        alias = update.get('alias')
```
lmacken: I don't fully understand this logic... if it has an alias attr, use it, else, try and get it but default to None?

pypingou: Yup, I was missing a part, see 8933e39

lmacken: I don't think the `if/else` conditional is necessary, since `alias` will always exist. It defaults to `None`, but all updates get assigned an alias immediately upon submission.

pypingou: So all updates are getting an alias but not necessarily a title? Sounds kinda odd to me

lmacken: All updates have both an alias and a title

pypingou: So shall I drop the if/else structure above for the title then?

lmacken: Yeah, I think that's fine. Updates will always have those attributes.

bodhi/util.py
```diff
@@ -435,12 +435,10 @@ def request2html(context, request):
...
 
-    if hasattr(update, 'title'):
-        title = update.title
-    else:
-        title = update['title']
+    title = update.title
+    alias = update.alias
 
-    url = request.route_url('update', id=title)
+    url = request.route_url('update', id=alias)
```
lmacken: There have been bugs in the past where alias was `None`. Hopefully that issue is fixed, but we probably still want the `alias or title` logic in there.

bodhi/util.py
```diff
@@ -435,12 +435,10 @@ def request2html(context, request):
 def update2html(context, update):
     request = context.get('request')
 
-    if hasattr(update, 'title'):
-        title = update.title
-    else:
-        title = update['title']
```
ralphbean: I think this was supposed to handle the case where ``update`` could be either a sqlalchemy ``Update`` object, or a ``dict``.

ralphbean: Yeah, confirmed.  This will work on some templates but not others.. The front page crashes with:

```python
MakoRenderingException: 

  File "/home/threebean/.virtualenvs/bodhi-python2.7/lib/python2.7/site-packages/pyramid_mako-1.0.2-py2.7.egg/pyramid_mako/__init__.py", line 148, in __call__
    result = template.render_unicode(**system)
  File "/home/threebean/.virtualenvs/bodhi-python2.7/lib/python2.7/site-packages/Mako-1.0.0-py2.7.egg/mako/template.py", line 452, in render_unicode
    as_unicode=True)
  File "/home/threebean/.virtualenvs/bodhi-python2.7/lib/python2.7/site-packages/Mako-1.0.0-py2.7.egg/mako/runtime.py", line 803, in _render
    **_kwargs_for_callable(callable_, data))
  File "/home/threebean/.virtualenvs/bodhi-python2.7/lib/python2.7/site-packages/Mako-1.0.0-py2.7.egg/mako/runtime.py", line 835, in _render_context
    _exec_template(inherit, lclcontext, args=args, kwargs=kwargs)
  File "/home/threebean/.virtualenvs/bodhi-python2.7/lib/python2.7/site-packages/Mako-1.0.0-py2.7.egg/mako/runtime.py", line 860, in _exec_template
    callable_(context, *args, **kwargs)
  File "/home/threebean/devel/bodhi/bodhi/templates/master.html", line 162, in render_body
    ${self.body()}
  File "/home/threebean/devel/bodhi/bodhi/templates/home.html", line 62, in render_body
    ${self.tables.updates(critpath_updates)}
  File "/home/threebean/devel/bodhi/bodhi/templates/tables.html", line 26, in render_updates
    ${fragments.update(update, display_user, display_request, display_release, display_karma)}
  File "/home/threebean/devel/bodhi/bodhi/templates/fragments.html", line 77, in render_update
    <td>${util.update2html(update) | n}</td>
  File "/home/threebean/devel/bodhi/bodhi/util.py", line 438, in update2html
    title = update.title
AttributeError: 'dict' object has no attribute 'title'
```

ralphbean:
> Tests are failing for me locally, but seems un-related to the change

What failures are you seeing?  The test suite is usually working.

bodhi/util.py
pypingou: So then the first approach was the right one, also for the alias?

ralphbean: Huh, yeah.  :)  a01f5d3e236 was closer to right, yes.

We could consolidate it into one block like this:

```python
if isinstance(update, Update):
    title = update.title
    alias = update.alias
else:
    title = update['title']
    alias =
hasattr(update, 'title'):

lmacken:
Works for me :) Thanks!

PR #470, opened 2015-09-03
consumers: Query the wiki for test cases on the backend (fixes #314)

### LINKED ISSUE #314: Move bug commenting into async process

lmacken:
:construction: :warning: UNTESTED :skull: :fire:


ralphbean:
:+1: and it fixes a bug we didn't even know about.

PR #471, opened 2015-09-03
A fedmsg data manager

### LINKED ISSUE #458: Create a fedmsg transaction manager
It would be awesome to hook in our fedmsg notification shim to the repoze.tm2 transaction middleware. Right now the consumers on the bodhi backend are getting fedmsgs before the frontend commits them to the database. If we create this layer, then fedmsgs will get sent only when `transaction.commit()` is called.

https://zodb.readthedocs.org/en/latest/transactions.html#the-pickledatamanager

ralphbean:
We had been calling this a "transaction manager" but after reading [the docs](https://zodb.readthedocs.org/en/latest/transactions.html#the-pickledatamanager) I see that the transaction module provides the transaction manager and we provide specialized "data managers" on our side.

This adds a data manager for Fedmsg.

I ended up having to make this ``ManagerMapping`` object that maintains state about which data manager **instance** is associated with which transaction manager instance.  I started trying to implement it in ``notifications.publish(...)`` but found that the logic grew very quickly and felt like I had to abstract it out.

Fixes #458.

We should probably split this out into fedmsg itself some day and just have bodhi.notifications use the data manager that it provides.  Other webapps could benefit from this.  Let's try it out in bodhi first?

### DIFF
--- a/bodhi/__init__.py
+++ b/bodhi/__init__.py
@@ -192,6 +192,7 @@ def main(global_config, testing=None, **settings):
     config.scan('bodhi.views')
     config.scan('bodhi.services')
     config.scan('bodhi.captcha')
+    config.scan('bodhi.events')
 
     # Setup fedmsg for this thread
     import bodhi.notifications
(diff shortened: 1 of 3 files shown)


lmacken:
:boom: :+1: :fire:

ralphbean:
Let's definitely give this a hard test in staging to see how it behaves under mod_wsgi.

Issue #204, opened 2015-08-20
Missing periods in many sentences


cicku:
Hi,

I'm not going to submit a PR about it, but a tiny issue is:

```
This update has been pushed to testing
This update has reached 3 days in testing and can be pushed to stable now if the maintainer wishes
```

Sounds like half-completed sentences to me, and some have period at the end:

```
This update is currently being pushed to the Fedora 23 testing updates repository.
```

Please unify comment patterns from bodhi to make them better. Thanks.

PR #477, opened 2015-09-04
Add period at the end of the comments

pypingou:


pypingou:
This relates to https://github.com/fedora-infra/bodhi/issues/204 but I'm not sure it covers it entirely

bodhi/models/models.py
```diff
@@ -1696,7 +1696,7 @@ def met_testing_requirements(self):
             if comment.user.name == 'bodhi' and \
                comment.text.startswith('This update has reached') and \
                comment.text.endswith('days in testing and can be pushed to'
-                                     ' stable now if the maintainer wishes'):
+                                     ' stable now if the maintainer wishes.'):
```
ralphbean: This will be problematic.

We need to be able to match both the old form without the ``.`` and the new form with it.

Maybe change it to ``' stable now if the maintainer wishes' in comment.text``?

pypingou: Sure can do

lmacken:
This breaks an alembic migration script that @ralphbean ran yesterday in staging, and soon to run in prod. I pushed out a fix for it, and :+1:  the rest of this PR.

Issue #473, opened 2015-09-03
replace request.user with a munch object


ralphbean:
Right now its a sqlalchemy object **but** we could be trying to render an error page *after* the sqlalchemy session is closed at which point references to ``request.user`` in the master template will fail.  Learned this the hard way in staging today.

Making it a munch object will get us around that.

PR #479, opened 2015-09-04
Munch it up!

ralphbean:
This changeset turns ``request.user`` into a munch object where it used to be a sqlalchemy ``User`` object, mediating the *complex failure scenario* detailed in #473.

It was a little more involved than I thought it would be, but not too bad.  All the tests pass (and I didn't have to modify any of them).

### DIFF
--- a/bodhi/__init__.py
+++ b/bodhi/__init__.py
@@ -14,6 +14,7 @@
 
 from collections import defaultdict
 from dogpile.cache import make_region
+from munch import munchify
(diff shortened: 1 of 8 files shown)


ralphbean:
Oh, I also fiddled with ``._to_json(...)`` and turned it into a ``@classmethod`` because I just so happened to notice that it didn't need a reference to ``self``.

lmacken:
Yup, looks good :bullettrain_side: :100: :mag:

PR #482, opened 2015-09-05
Fix an attribute error from the munchification.

ralphbean:


puiterwijk:
Looks sane to me.

bodhi/__init__.py
```diff
@@ -61,8 +61,8 @@ def get_user(request):
 
 
 def groupfinder(userid, request):
-    user = request.user
-    if user:
+    if request.user:
+        user = User.get(request.user.name, request.db)
```
puiterwijk: User with capital U seems to be only imported in get_user(...), maybe missing an import here or need that on a global level?

lmacken: Yup, I think we need to import it at a function level to avoid a circular dependency. :repeat:

PR #483, opened 2015-09-05
models: Index the Update.title and date_submitted

lmacken:
I haven't done any benchmarking to see if this helps, but I'm guessing it will since I think the Search widget searches over this field.


ralphbean:
Nice!  You could try ``tools/perf-test.py`` before and after the alembic upgrade (and you should be able to downgrade and upgrade back and forth to test over and over).

I bet it will *really* help since things are referred to by their title all over the place.

PR #491, opened 2015-09-06
Add an option to bypass the fedmsg transaction/data manager.

ralphbean:
And use that option in the masher so messages don't get queued up until we're all done.


lmacken:
:+1:

ralphbean:
:bullettrain_side: :bank: :tractor:

PR #492, opened 2015-09-06
Add force=True to our mock checks in the bodhi tests.

ralphbean:
Also, remove some dead code from the masher tests.


lmacken:
:+1:

Issue #488, opened 2015-09-05
Support an old update URL format


ralphbean:
like ``/updates/{alias}/{title}``.  it's redundant, but some people are still trying to use it.

pypingou:
This isn't such an easyfix as ``/updates/`` relies on cornice and not on the regular routing mechanism and we need to find a way to say: consider ``/updates/{alias}/{title}`` to be ``/updates/{alias}`` when ``{title}`` is not ``edit`` or ``request``, otherwise we conflict with other URLs

pypingou:
Actually, it can only conflict with ``edit``, ``request`` is in another path

pypingou:
so I tried two things:

a) Adding a new URL routing in the pyramid side that would just redirect the old URLs to the new ones:
````
@view_config(route_name='old_update', renderer='json',
             request_method='GET')
def old_update_view(request):
    """ Method providing a backward compatible url for old updates """
    return HTTPFound("/updates/{0}".format(request.matchdict['alias']))
````

Fails as it catches ``/update/{alias}/edit`` as well.

b) Add a new cornice service:
````
update_old = Service(name='update_edit', path='/updates/{id}/{title}',
                     validators=(validate_update_id,),
                     description='Update submission service',
                     #acl=bodhi.security.package_maintainers_only_acl,
                     acl=bodhi.security.packagers_allowed_acl,
                     cors_origins=bodhi.security.cors_origins_rw)
...
@update_old.get(accept=('application/json', 'text/json'), renderer='json',
                error_handler=bodhi.services.errors.json_handler)
def get_update_old(request):
    return HTTPFound("/updates/{0}".format(request.matchdict['alias']))
````

Fails as then: ``/update/{alias}/foobar`` is considered as ``/update/{alias}/edit``

pypingou:
And this was the mistake:
> update_old = Service(name='update_edit',
The ``name`` there was incorrect, using the right one, it works.

PR #499, opened 2015-09-07
Add a redirect from /updates/alias/title to /updates/alias

pypingou:
This provides some backward compatibility layer with bodhi1

Fixes https://github.com/fedora-infra/bodhi/issues/488

### DIFF
--- a/bodhi/services/comments.py
+++ b/bodhi/services/comments.py
@@ -15,7 +15,7 @@
 import math
 
 from cornice import Service
-from pyramid.httpexceptions import HTTPBadRequest, HTTPFound
+from pyramid.httpexceptions import HTTPBadRequest
 from sqlalchemy import func, distinct
 from sqlalchemy.sql import or_
 
@@ -188,11 +188,3 @@ def new_comment(request):
         return
 
     return dict(comment=comment, caveats=caveats)
-
-
-@comments.get(accept=('application/atom+xml',))
-def rss_redirect(request):
-    url = request.route_url('comments_rss')
-    if request.query_string:
-        url = url + '?' + request.query_string
-    raise HTTPFound(url)
--- a/bodhi/services/overrides.py
+++ b/bodhi/services/overrides.py
@@ -16,7 +16,6 @@ import math
 
 from cornice import Service
 from pyramid.exceptions import HTTPNotFound
-from pyramid.httpexceptions import HTTPFound
 
 from sqlalchemy import func, distinct
 from sqlalchemy.sql import or_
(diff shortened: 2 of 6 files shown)


lmacken:
This breaks the test suite:
```
ERROR: Test submitting an invalid request
  File "/home/lmacken/.virtualenvs/bodhi-python2.7/lib/python2.7/site-packages/mock/mock.py", line 1305, in patched
  File "/home/lmacken/code/github.com/bodhi/bodhi/tests/functional/test_updates.py", line 1517, in test_invalid_request
    {'request': 'foo','csrf_token': self.get_csrf_token()}, status=400)
  File "/home/lmacken/.virtualenvs/bodhi-python2.7/lib/python2.7/site-packages/webtest/utils.py", line 37, in wrapper
    return self._gen_request(method, url, **kw)
  File "/home/lmacken/.virtualenvs/bodhi-python2.7/lib/python2.7/site-packages/webtest/app.py", line 735, in _gen_request
    expect_errors=expect_errors)
  File "/home/lmacken/.virtualenvs/bodhi-python2.7/lib/python2.7/site-packages/webtest/app.py", line 631, in do_request
    self._check_status(status, res)
  File "/home/lmacken/.virtualenvs/bodhi-python2.7/lib/python2.7/site-packages/webtest/app.py", line 666, in _check_status
    "Bad response: %s (not %s)", res_status, status)
AppError: Bad response: 405 Method Not Allowed (not 400)
```

pypingou:
Ok, I see the error, the new service update_old gets the hand when a ``GET`` request is performed.

I've tried to add:
````
@update_request.get(accept=('application/json', 'text/json'), renderer='json',
                error_handler=bodhi.services.errors.json_handler)
@update_request.get(accept=('application/javascript'), renderer='jsonp',
                error_handler=bodhi.services.errors.jsonp_handler)
@update_request.get(accept="text/html", renderer="update.html",
                error_handler=bodhi.services.errors.html_handler)
def update_request_get(request):
    raise exception_response(405)
````
To keep the behavior consistent, but for some reason it does not seem to work, it looks like update_old always gets the hand.

ralphbean:
This is pretty similar to how we did RSS redirects in 2732147.  We could probably simplify the way this works now:

- Remove the error handlers and validators.
- Boil it down to just redirection.
- Avoid using ``zz_`` as a prefix, and instead use something with "redirect" in the name.

We could also move all of our redirect endpoints to a ``bodhi/services/redirects.py`` file to reduce clutter, maybe?

ralphbean:
OK, I added some reorganization here and a test for the test suite.  How's this look?

lmacken:
Looks good to me! :

Issue #489, opened 2015-09-06
optimization.  fast-track provenpackagers through ACL checks


ralphbean:
currently, we check acls on packagers first and *then* check if the update submitter is a provenpackager.  we should do it in the opposite order such that if they are a provenpackager, we can skip *all those pkgdb queries* that we have to make in order to figure out if they have acls on the individual packages.

PR #500, opened 2015-09-07
Check if an user is a provenpackager before checking their ACLs

pypingou:
This should speed things up by avoiding making un-necessary calls to
pkgdb.

Fixes https://github.com/fedora-infra/bodhi/issues/489

### DIFF
--- a/bodhi/validators.py
+++ b/bodhi/validators.py
@@ -53,6 +53,7 @@ def cache_nvrs(request, build):
     if '' in (name, version, release):
         raise ValueError
 
+
 def validate_nvrs(request):
     for build in request.validated.get('builds', []):
         try:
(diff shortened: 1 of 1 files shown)


bodhi/validators.py
```diff
@@ -276,6 +278,20 @@ def validate_acls(request):
...
+        # Allow certain groups to push updates for any package
+        admin_groups = settings['admin_packager_groups'].split()
+        for group in admin_groups:
+            if group in user_groups:
+                log.debug(
+                    '{} is in {} admin group'.format(user.name, group))
+                has_access = True
+                break
+        if has_access:
+            continue
```
ralphbean: Is this ``continue`` a ``SyntaxError``?

ralphbean: Oh, no.  I see.  it's inside the ``for build in builds:`` loop.

ralphbean:
:+1: This is awesome!

(and it should **really** help the performance of big multi-build updates).

Issue #485, opened 2015-09-05
Buildroot override view doesn't tell you who submitted it


AdamWill:
When I view an existing buildroot override submitted by someone else - e.g. https://bodhi.fedoraproject.org/overrides/rpm-4.13.0-0.rc1.1.fc23 - what I get is a view that kinda looks like it's intended for the person who submitted the override; it has 'Submit' and 'Expire' buttons and some instructions on checking if the override is active. But it doesn't tell me who submitted the override, which is what I actually wanted to know.

ralphbean:
Should be easy to add there.

PR #501, opened 2015-09-07
Display the submitter of the override on the override page

pypingou:
fixes https://github.com/fedora-infra/bodhi/issues/485

### DIFF
--- a/bodhi/templates/override.html
+++ b/bodhi/templates/override.html
@@ -1,3 +1,4 @@
+<%namespace name="util" module="bodhi.util"/>
 <%inherit file="master.html"/>
 
 <div class="row">
@@ -60,7 +61,7 @@
         </div>
       </div>
 
-      <div class="col-md-6">
+      <div class="col-md-3">
         <div class="panel formpanel panel-default">
           <div class="panel-heading clearfix">
             <span class="pull-left">Expiration date</span>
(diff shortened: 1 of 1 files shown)


bodhi/templates/override.html
```diff
@@ -76,6 +77,27 @@ <h2 class="pull-left">New Override</span>
...
+          <div class="form-group noui text-left">
+            %if override is not UNDEFINED:
+            <a href="${request.route_url('user', name=override.submitter.name)}">
+              <img class="img-circle" src="${util.avatar(override.submitter.name, size=24)}"/>
+              ${override.submitter.name}
+            </a>
+            %else:
+            <a href="${request.route_url('user', name=request.user.name)}">
+              <img class="img-circle" src="${util.avatar(request.user.name, size=24)}"/>
+              ${request.user.name}
```
lmacken: I'm confused here. If the override is UNDEFINED (is that even possible with the validators?) then it shows the currently logged in persons name?

ralphbean: Confused here too.  Oh, I think I see now!

- If the override is **not** UNDEFINED, then show the override submitter.
- If it *is* undefined, then show the logged in user.

I think this is because the override template here is used for both displaying existing overrides *and* for the form for creating new overrides.  So, if it's new ``<form>`` mode, then show *you* since you're going to be the one who ends up owning the override.

bodhi/templates/override.html
```diff
@@ -60,7 +61,7 @@ <h2 class="pull-left">New Override</span>
         </div>
       </div>
 
-      <div class="col-md-6">
+      <div class="col-md-3">
```
ralphbean: This is changing the columns quite a bit.  How does it look?

bodhi/templates/override.html
lmacken: Oh cool, I didn't realize it was used in both cases. :+1:  here then.

bodhi/templates/override.html
ralphbean: Looks okay here:

override

I guess it could use a little padding.. but so could the other panels.

ralphbean:
:+1: here.

lmacken:
Yup, :+1:

Issue #484, opened 2015-09-05
No way to view buildroot overrides from web UI


AdamWill:
Buildroot overrides show up in the Bodhi 2 web UI search, but there is no place (I can see) where you can just go and view all buildroot overrides. This is sometimes something you want to do.

ralphbean:
Oh, it's https://bodhi.fedoraproject.org/overrides/

We could maybe link to it from the ``/releases/SOME_RELEASE`` page by listing all the overrides there?

It's also linked to from user profiles (you can get there to see all the overrides submitted by *someone*).

PR #502, opened 2015-09-07
Link to lists of buildroot overrides

pypingou:
Relates to https://github.com/fedora-infra/bodhi/issues/484

### DIFF
--- a/bodhi/services/releases.py
+++ b/bodhi/services/releases.py
@@ -25,6 +25,7 @@ from bodhi.models import (
     UpdateStatus,
     UpdateType,
     Build,
+    BuildrootOverride,
     Package,
     Release,
 )
@@ -102,6 +103,30 @@ def get_release_html(request):
     num_updates_newpackage = base_count_query\
         .filter(Update.type==UpdateType.newpackage).count()
 
+    num_active_overrides = request.db.query(
+        BuildrootOverride
+    ).filter(
+        BuildrootOverride.expired_date==None
+    ).join(
+        BuildrootOverride.build
+    ).join(
+        Build.release
+    ).filter(
+        Build.release==release
+    ).count()
+
+    num_expired_overrides = request.db.query(
+        BuildrootOverride
+    ).filter(
+        BuildrootOverride.expired_date!=None
+    ).join(
+        BuildrootOverride.build
+    ).join(
+        Build.release
+    ).filter(
+        Build.release==release
+    ).count()
+
     return dict(release=release,
                 latest_updates=updates.limit(25).all(),
                 count=updates.count(),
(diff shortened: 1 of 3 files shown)


pypingou:
Relates to https://github.com/fedora-infra/bodhi/issues/484 but not sure if that fixes it entirely (doesn't seem to follow completely @ralphbean 's idea)

ralphbean:
:-1: here to adding it to the "Releases" dropdown like that.  It would be confusing, I think, to click on "Releases" and then be shown a drop-down that included buildroot overrides.

My thought was to add them as a 'stat' to left-hand table on the ``release.html`` page.  There it shows that F22 has ``X`` stable updates and ``Y`` testing updates and it provides links to those lists.  We could do the same for buildroot overrides and say that there are ``X`` total overrides and ``Y`` *active* (non-expired) overrides for F22 -- and then provide links to the lists like we do for updates on that page.

Showing them in the context of a single release will make sense, I think, because people will want to know "what overrides are active for a particular buildroot".

pypingou:
On Mon, Sep 07, 2015 at 06:00:23AM -0700, Ralph Bean wrote:
>    :-1: here to adding it to the "Releases" dropdown like that. It would be
>    confusing, I think, to click on "Releases" and then be shown a drop-down
>    that included buildroot overrides.

Fair

>    My thought was to add them as a 'stat' to left-hand table on the
>    release.html page. There it shows that F22 has X stable updates and Y
>    testing updates and it provides links to those lists. We could do the same
>    for buildroot overrides and say that there are X total overrides and Y
>    active (non-expired) overrides for F22 -- and then provide links to the
>    lists like we do for updates on that page.

While I like the idea, this doesn't quite qualifies as an easyfix anymore then
:)

I looked quickly through the code to see if there was a way to restrict the list
of buildroot overrides displayed for a certain release and iiuc there isn't such
mechanism yet. Is that correct?

So that would be:
- Get the number of active/expired overrides to put on the stat section of the
  release page
- Add the possibility to filter overrides for a certain release
- Link from the stat on the release page to this list.

Would a URL arg be ok to filter ``/overrides`` ?

ralphbean:
> - Get the number of active/expired overrides to put on the stat section of the release page

Yeah!  That would need to be added to the overrides cornice service in ``bodhi/services/releases.py`` I think.

> - Add the possibility to filter overrides for a certain release

That would need to be added to:

- The ``ListOverridesSchema`` in ``bodhi/schemas.py``
- And the overrides service in ``bodhi/services/overrides.py``.  That service will probably need to use one of the existing releases validators from ``bodhi/validators.py``.

> - Link from the stat on the release page to this list.

Yup!  That should be a relatively simple last piece added to ``bodhi/templates/release.html``.

> Would a URL arg be ok to filter ``/overrides`` ?

Yes :)  Take a look at the schema and service used to list updates (by release) for an example.

pypingou:
>        * Add the possibility to filter overrides for a certain release
> 
>    That would need to be added to:
> 
>      * The ListOverridesSchema in bodhi/schemas.py
>      * And the overrides service in bodhi/services/overrides.py. That service
>        will probably need to use one of the existing releases validators from
>        bodhi/validators.py.

This was already there actually, cool :)

pypingou:
These should address @ralphbean 's concerns :)

ralphbean:
OK, looks like this now after 4b8ba29:

overrides-list

lmacken:
:+1:

PR #520, opened 2015-09-08
Ensure that fedmsg can publish before voting.

ralphbean:
If fedmsg doesn't have an endpoint, then we will fail in a Very Ugly Way in
``tpc_finish()``.  We should do this check here in ``tpc_vote()`` so that we
can fail early before things get into a weird, inconsistent state.


lmacken:
Cool. :+1:

PR #523, opened 2015-09-08
Include meets_testing_requirements in Update JSON.

ralphbean:
This is for my personal use.

It would be cool if Update JSON included a True/False of whether or not updates
are ready to be pushed to stable or not.  I would use this from local scripts
to figure out what updates I need to go push to stable.  Historically, I would
scrape the comment history looking for bodhi to tell me its good to go.


lmacken:
Nice, :+1:

PR #524, opened 2015-09-08
Correctly map builds to bugs when closing.

### LINKED ISSUE #513: Bodhi changing a version number in Bugzilla
My [bug](https://bugzilla.redhat.com/show_bug.cgi?id=1252902) in bugzilla have bad version number when bodhi closed it.

Seems to me like bodhi have problems when the patch goes to multiple projects with *Resolves* and *Related* tags.

ralphbean:
It used to be that we would take the version of the first build in a
multi-build update and use it everywhere for the "fixed in version" field.
This should more correctly map builds to bugs.

Fixes #513.

### DIFF
--- a/bodhi/bugs.py
+++ b/bodhi/bugs.py
@@ -93,12 +93,12 @@ class Bugzilla(BugTracker):
         except:
             log.exception("Unable to alter bug #%d" % bug_id)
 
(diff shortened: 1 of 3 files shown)


lmacken:
LGTM, :+1:

PR #526, opened 2015-09-08
Flip the order of the feedback items.

### LINKED ISSUE #498: Incorrect order of Feedback columns for positive/negative karma
Hi,

my bad I did not check which radio button adds positive/negative karma properly. However, if you look at any update with some karmas, there are two places:
1) Related Bugs
2) Feedback

1) shows karmas for bugs in positive | negative order of columns. 2) shows negative | neutral | positive order. Which is a little user-unfriendly. Would it be possible to switch ordering of 2) to positive | neutral | negative so it complies with ordering of 1)? If not, the issue can be closed.

[1] https://bodhi.fedoraproject.org/updates/golang-1.5-8.fc23

Jan

ralphbean:
So that they are in the same order as in the comment form at the bottom of the page.

Fixes #498.

### DIFF
--- a/bodhi/models/models.py
+++ b/bodhi/models/models.py
@@ -957,7 +957,7 @@ class Update(Base):
                         good += 1
                     elif feedback.karma < 0:
                         bad += 1
-        return good, bad * -1
+        return bad * -1, good
 
     def get_testcase_karma(self, testcase):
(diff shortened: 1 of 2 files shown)


lmacken:
:+1:

PR #528, opened 2015-09-08
Make the fixedin field additive.

### LINKED ISSUE #428: Don't fiddle with "fixed in" all the time
The "fixed in field" should be probably additive, that way all the releases where the errata was issued should stay listed. E.g. [1] should list rubygem-rack-1.6.1-2.fc22 as well as rubygem-rack-1.5.2-5.fc21 and don't change the field all the time.

[1] https://bugzilla.redhat.com/show_bug.cgi?id=1248923

ralphbean:
This should fix #428.

Furthermore, I try here to stick with how the anaconda team uses the fixedin
field by hand so that we don't disrupt their workflow.  See this bug for an
example:  https://bugzilla.redhat.com/show_bug.cgi?id=1256072

### DIFF
--- a/bodhi/bugs.py
+++ b/bodhi/bugs.py
@@ -97,8 +97,18 @@ class Bugzilla(BugTracker):
         args = {}
         try:
             bug = self.bz.getbug(bug_id)
+            # If this bug is for one of these builds...
             if bug.component in versions:
-                args['fixedin'] = versions[bug.component]
+                version = versions[bug.component]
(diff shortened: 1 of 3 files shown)


lmacken:
Nice! :+1:

voxik:
I think that the way fixed in is used by anaconda is not generic way, i.e. I believe that in RH, it is suggested to used the complete NVR, not just version.

PR #532, opened 2015-09-09
Make the edit comment look pretty.

ralphbean:
This enumerates the list of added and removed builds in markdown format, so it should look nice and comprehensible.


lmacken:
Slick! :+1:

PR #533, opened 2015-09-09
Tweak rhbz fixedin handling (again)

### LINKED ISSUE #531: The "Fixed In Version" Bugzilla field should contain NVRs
Traditionally, the "Fixed In Version" field in Bugzilla contained NVRs (separated by spaces where applicable). The current instance uses only VRs.

This adds an unnecessary inconsistency and confusion in cases where a bug is fixed by multiple packages, each with a different version.

ralphbean:
This is the third and hopefully final time.

We got pinged by some Red Hatters who say that there's a precedent for handling
that field in a particular way.  This should hopefully meet that criteria.

Fixes #531.

### DIFF
--- a/bodhi/bugs.py
+++ b/bodhi/bugs.py
@@ -101,13 +101,18 @@ class Bugzilla(BugTracker):
             if bug.component in versions:
                 version = versions[bug.component]
                 # Get the existing list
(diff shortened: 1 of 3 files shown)


lmacken:
Looks good to me. :+1:

PR #535, opened 2015-09-09
Change yum references to dnf

### LINKED ISSUE #335: Push notification in BZ should not contain '\n'
Please see [1] for example.

[1] https://bugzilla.redhat.com/show_bug.cgi?id=1208316#c106

### LINKED ISSUE #357: Update text in bugs is out of date and strangely formatted
The instructions bodhi puts in each bug for testing an update is strangely formatted and out of date.  Example text:

"kernel-4.2.0-0.rc8.git0.1.fc23 has been pushed to the Fedora 23 testing repository. If problems still persist, please make note of it in this bug report.\nIf you want to test the update, you can install it with \n su -c 'yum --enablerepo=updates-testing update kernel'. You can provide feedback for this update here: https://bodhi.fedoraproject.org/updates/FEDORA-2015-14151"

The \n characters look strange there.  Further, yum is no longer used as the package manager in Fedora 22 and newer.  It will still work for now, but it is likely better to change the text to reference dnf.

ralphbean:
Just like in #359, except here we also conditionalize yum/dnf advice on EPEL/Fedora.

Lastly, also fix newlines in the config, which

- fixes #335
- fixes #357

That will also need to be applied to our ansible production/staging config.

### DIFF
--- a/bodhi/models/models.py
+++ b/bodhi/models/models.py
@@ -1941,7 +1941,15 @@ class Bug(Base):
             update.get_title(delim=', '), "%s %s" % (
                 update.release.long_name, update.status.description))
         if update.status is UpdateStatus.testing:
-            message += config['testing_bug_msg'] % (
+            template = config['testing_bug_msg']
+
+            if update.release.id_prefix == "FEDORA-EPEL":
+                if 'testing_bug_epel_msg' in config:
+                    template = config['testing_bug_epel_msg']
+                else:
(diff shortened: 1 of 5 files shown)


ralphbean:
Config change applied to ansible here:  https://infrastructure.fedoraproject.org/cgit/ansible.git/commit/?id=f3d61d9f2d46eef6d99edf3701e9286799f73314

lmacken:
:+1:

Issue #259, opened 2015-08-20
Document & link to the RSS/JSON/HTML features of each service


lmacken:
We have a super powerful architecture now that allows for RSS/Atom, JSON, and HTML views of updates, releases, users, etc. We need to document this and we should link to it somewhere.

lmacken:
Bodhi1 also linked to a bunch of documentation, like the Updates Policy. We might want to setup a separate menu item for docs?

ryanlerch:
WE now have API docs, and the Docs are linked to in the footer on Bodhi.

PR #538, opened 2015-09-09
Fix RSS.

### LINKED ISSUE #493: create different urls for rss feeds
So, I tried adding those html links that allow for rss auto-discovery, and they worked in Firefox!

```diff
diff --git a/bodhi/templates/master.html b/bodhi/templates/master.html
index 4c88c1f..90e6078 100644
--- a/bodhi/templates/master.html
+++ b/bodhi/templates/master.html
@@ -31,6 +31,11 @@
     <link href="${request.static_url('bodhi:static/css/panel.css') }" rel="stylesheet" />
     <script src="${request.static_url('bodhi:static/js/jquery-1.10.2.min.js')}"></script>
     <script src="${request.static_url('bodhi:static/js/Chart-0.2.0.min.js')}"></script>
+
+    % if request.matched_route:
+    <link rel="alternate" type="application/atom+xml" title="RSS Feed" href="${request.current_route_url()}" />
+    % endif
+
   </head>
 
   <body>
```

However, when you try to navigate to the feed via that little advertisement button, it looks like firefox sends the accept header of ``*/*`` which then gets a response of just HTML from our app.  No rss feed.

We could get around this if we made different urls for rss feeds (which people have been requesting, but we didn't want it.. we thought our implementation was the coolest!)  We should probably do that.

Something like ``/updates/rss?submitter=ralph``?

ralphbean:
This does two things:

- Move our RSS feeds to their own url path.
- Put rss links everywhere.

Bugs:

- Fixes #493.
- Relates to #259.


ralphbean:
Just a note again not to merge this until the font-awesome cdn link is removed.

ralphbean:
OK - ready to go here.

lmacken:
Looks good! :+1: 

Although, people will the existing URLs hooked up will not be happy.

Issue #539, opened 2015-09-10
Can't comment on locked update(?)


AdamWill:
I twice tried to post a comment on https://bodhi.fedoraproject.org/updates/FEDORA-2015-15401 while it was showing "Locked: This update is currently locked and cannot be modified." Both times the comment and karma values did not appear. I wasn't focused on the tab so I didn't see if there was an error message (seriously, those vanishing error messages are an awful idea for something as slow as Bodhi, no-one sits there on the tab for a minute while they wait!)

The update being locked should not prevent comment/karma submissions, I don't think, if that's what's going on.

ralphbean:
- Yes.  The update being locked should not prevent comment/karma submission.  Agreed!  However, we'd have to be careful that a comment with karma that pushes something above or below the karma threshold(s) doesn't send us into a weird/inconsistent state.
- On the error messages vanishing... the error messages are actually supposed to stay-put.  The "info" and "success" notifications vanish, but error ones should stick around until you manually dismiss them.  Try hard-reloading a page to make sure your browser doesn't have any old assets cached (this was introduced a week ago or so...).

ralphbean:
Also, it takes an *insanely* long time to comment on that update.  What's going on!?

ralphbean:
Ha, so I was able to comment on it.. but I just so happened to try it as it was completing the push, so I can't say what was what.

ralphbean:
#542 leads in the direction of closing this as not-a-bug (although there are other problems, as we've both touched on here).

AdamWill:
well, I definitely tried to submit my comment twice, and it definitely didn't wind up appearing in the update.

ralphbean:
Sorry adam :(  I don't have any other information to go on at the moment.  If it happens to you again, can you make a note of the time in UTC and re-open this bug so we can go hunting in the logs?

dhgutteridge:
I just encountered this same problem commenting on [https://bodhi.fedoraproject.org/updates/FEDORA-2017-8a540cb555](https://bodhi.fedoraproject.org/updates/FEDORA-2017-8a540cb555). It was in the process of being pushed to testing (though in the current incarnation of Bodhi, there was no visible notification I could see that this was the case), and Bodhi simply didn't record the input. This was between 19:00 and 20:00 UTC. (I resubmitted after it was pushed, and Bodhi accepted that input.)

bowlofeggs:
[...]

PR #542, opened 2015-09-10
Add tests for commenting on locked updates.

ralphbean:
- Ensure you can comment on a locked update.
- Ensure that the comment won't trigger threshold action if your comment
takes
 things over or under the stable/unstable karma thresholds.

Relates to #539.


lmacken:
:+1:

PR #544, opened 2015-09-10
Show the locked panel to everyone...

### LINKED ISSUE #543: show "locked" panel even when user is not logged in
If you look at a locked update when logged out, you can't tell that it's locked.

ralphbean:
... even to people who cannot edit the update, like not-logged-in users.

Fixes #543, if that's what we want to do.  Could also close as WONTFIX.  Any opinions?


lmacken:
Sounds reasonable to me. :+1:

Issue #267, opened 2015-08-20
fedora-easy-karma fails with limit error


nirik:
With python-fedora-0.5.5-1.fc22.noarch installed, running fedora-easy-karma gives: 

$ fedora-easy-karma 
Getting list of installed packages...
Waiting for Bodhi for a list of packages in updates-testing (F22)...
Cannot query Bodhi: 1000 is greater than maximum value 100

perhaps allow queries of 1000 or we could ask fedora-easy-karma to adjust to 100 I guess.

nirik:
If I manually edit fedora-easy-karma and set it to 100, it gets past that to a new error: 

Waiting for Bodhi for a list of packages in updates-testing (F22)...
Traceback (most recent call last):
  File "/usr/bin/fedora-easy-karma", line 651, in <module>
    FedoraEasyKarma()
  File "/usr/bin/fedora-easy-karma", line 466, in __init__
    self.already_commented(update, self.options.fas_username):
  File "/usr/bin/fedora-easy-karma", line 576, in already_commented
    comment["author"].split(" ")[0] == user:
KeyError: 'author'

JamesWrigley:
This also happens for me with fedora-easy-karma-0-0.24.20150508gitc8e437c0.fc22.noarch and python-fedora-0.5.5-1.fc22.noarch.

tyll:
IMHO the 100 updates limit is too low, at least for f-e-k, this means that 6 slow request are required to get a list of all testing updates which took 4 minutes when I tested this. Since the f-e-k queries are very common, can Bodhi maybe cache them and for example just provide it at a static location to make this faster?

lmacken:
It looks like this issue has been fixed.

PR #546, opened 2015-09-10
Let people query the API up to 1000 items.

### LINKED ISSUE #279: /updates/? returns too much data
I want to continuously consume information about the latest stable updates for all Fedora releases in an automated way.

It's my impression that some updates have status "testing", potentially for a long time, before they transition to status "stable".

When I
```
GET https://bodhi.fedoraproject.org/updates/?release=f21
Accept: application/json
```
the result is one page of information about updates. When I max out the page size (rows_per_page=100), the result is 96 pages of updates. Since these are sorted by date_submitted (not updated when an update moves from "testing" to "stable"), as I continuously check for new updates, I would need to check every single page to see if any update has changed state.

If I instead consume 
```
https://bodhi.fedoraproject.org/updates/?release=f21&status=stable
```
since items are sorted by date_submitted, I would again have to check every page to see if new updates are available and which ones they are (now only 84 pages).

Other users might want to do something similar, filtering on different attributes.

Since I want to consume this information continuously (say check for new updates every couple hours) and not put undue load on your web servers / databases, I would like to make the following request:

I suggest creating a new route /updates/ids that lists only the updateid of updates matching request conditions. For this endpoint, the rows_per_page could safely be much greater, say 1000 or 10000.

```
https://bodhi.fedoraproject.org/updates/ids?release=f21&status=stable
```
This would return something like:
```javascript
{
    "rows_per_page": 1000,
    "total": 9568,
    "pages": 10,
    "updates": [
        "FEDORA-2015-13080",
        "FEDORA-2015-13129",
        ...
    ]
}
```

This is a fairly simple addition, and would make it much easier for users to consume information about updates on an ongoing basis.

Let me know if you think this is a good idea. I'm happy to submit a pull request.

ralphbean:
- Relates to #267.
- Fixes #279.

### DIFF
--- a/bodhi/schemas.py
+++ b/bodhi/schemas.py
@@ -236,7 +236,7 @@ class PaginatedSchema(colander.MappingSchema):
 
     rows_per_page = colander.SchemaNode(
         colander.Integer(),
-        validator=colander.Range(min=1, max=100),
+        validator=colander.Range(min=1, max=1000),
         location="querystring",
         missing=20,
     )


lmacken:
:+1:  :100: * 10

PR #549, opened 2015-09-10
Use email over username here (if we have it).

ralphbean:
We did this just a few lines above for all of the package maintainers, but we
forgot to do it here for all the people who commented on the bug.  This should
cut down on superfluous warnings in our logs *and* get more email to people who
should be getting it.


lmacken:
Yup, nice. :+1:

PR #551, opened 2015-09-10
Use validate_acls to produce the can_edit value.

### LINKED ISSUE #190: sort out the bodhi.security ACLs versus bodhi.validators.validate_acls
They're both used to check if someone has commit rights to a package in the bodhi.services.updates services, but the bodhi.security one doesn't work... the validate_acls one does... and we don't even need both.

We should go back through and fine-tooth comb it and add test cases for it.

### LINKED ISSUE #321: Don't show push buttons if user doesn't have sufficient rights on the update
We fixed #282, but it requires some followup.  Currently, if you try to set the request on an update you don't have rights on, you get correctly denied.  However, if you view an update page you may still be presented with buttons telling you that you *can* set the request (even though you can't).

We can likely just invoke the same server-side logic used to guard the set_request service when deciding whether or not to show buttons.

* The logic for displaying (or not displaying) buttons currently resides in ``bodhi/templates/update.html``.
* It relies on a ``can_edit`` field that gets computed in ``bodhi/services/updates.py``.
* That field is derived from the ``bodhi.security.packagers_allowed_acl`` object which is attached to the cornice service.
* That is wrong in both name and substance.  Really, we only want the package **maintainers** to have ``can_edit=True``... (which is supposed to be embodied in ``bodhi.security.package_maintainers_only_acl``).  The ``packagers_allowed_acl`` evaluates to ``True`` for *any* packager :disappointed:. 
* The logic that currently handles denying set_request requests is in ``bodhi.validators.validate_acls``.  It has code paths in it that are superior to the logic in ``bodhi.security.package_maintainers_only_acl``.  The later, for instance, only checks if you have direct commit on a package in pkgdb -- but it *does not* check if you're in a group which has group commit access (a group like the *infra-sig* group, for instance).

**So**, what we need to do is:

* [x] Write a test-case in ``bodhi/tests/functional/test_updates.py`` that *fails* given the current code (it should hit the JSON api and ensure that ``can_edit`` is True/False given different setup conditions).
* [x] Re-organize ``bodhi.security.package_maintainers_only_acl`` and ``bodhi.validators.validate_acls`` so that they *share code*.  Take care that we do this correctly and without introducing any unnecessary overhead.
* [ ] ~~Apply the ``bodhi.security.package_maintainers_only_acl`` to that get_request service instead of ``bodhi.security.packagers_allowed_acl``~~.
* [x] Ensure that our original test-case we wrote in step 1 *passes* after we re-organize things.

ralphbean:
* Fixes #190
* Fixes #321

This didn't go the way I expected.  I expected to make the ``bodhi.security.package_maintainers_only_acl`` acl smarter by moving some of the logic over from ``validate_acls``, and then rewriting  ``validate_acls`` to use that pyramid acl.  This turned out to not be the best idea.

First, pyramid acls must return a list of principals indicating who is and who isn't allowed to access a resource.  For us, that would mean returning a list of *all* proven packagers and *all* people with commit access (by querying pkgdb) every time we load a page.  `validate_acls` is smarter than this in that it first checks for provenpackager status and then skips querying pkgdb if the person is in that group.  I thought about trying to be clever by returning a *generator* from our acl function there, so that pyramid would only evaluate the first few items before saying "okay", but it is not written to be able to handle that.  I then thought about replacing pyramid ACLAuthorizationPolicy object with our own, but then things got out of hand.

Before this patch, we were querying pkgdb *twice* (in some cases, three times) for all the people with commit on *all the builds in an update*.  Now we should be doing that *at most once*, and sometimes not at all (in the case of provenpackager).

Furthermore, in fixing #321, the buttons that show up at the top of the update page should make a lot more sense.

### DIFF
--- a/bodhi/security.py
+++ b/bodhi/security.py
@@ -14,14 +14,13 @@
 
 from cornice.errors import Errors
 
-from pyramid.security import (Allow, Deny, Everyone, Authenticated,
-                              ALL_PERMISSIONS, DENY_ALL)
+from pyramid.security import (Allow, ALL_PERMISSIONS, DENY_ALL)
 from pyramid.security import remember, forget
 from pyramid.httpexceptions import HTTPFound
 from pyramid.threadlocal import get_current_registry
 
 from . import log
-from .models import User, Group, Update
+from .models import User, Group
 
 
 #
@@ -37,19 +36,10 @@ def admin_only_acl(request):
 
 def packagers_allowed_acl(request):
     """Generate an ACL for update submission"""
-    return [(Allow, 'group:' + group, ALL_PERMISSIONS) for group in
-            request.registry.settings['mandatory_packager_groups'].split()] + \
-           [DENY_ALL]
-
-
-def package_maintainers_only_acl(request):
-    """An ACL that only allows package maintainers for a given package"""
-    acl = admin_only_acl(request)
-    update = Update.get(request.matchdict['id'], request.db)
-    if update:
-        for committer in update.get_maintainers():
-            acl.insert(0, (Allow, committer.name, ALL_PERMISSIONS))
-    return acl
+    groups = request.registry.settings['mandatory_packager_groups'].split()
+    return [
+        (Allow, 'group:' + group, ALL_PERMISSIONS) for group in groups
+    ] + [DENY_ALL]
 
 
 #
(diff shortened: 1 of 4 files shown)


lmacken:
> For us, that would mean returning a list of all proven packagers and all people with commit access (by querying pkgdb) every time we load a page

Are you sure that's the case? I was under the impression that we could simply return `group:provenpackager`. If I go to https://bodhi.fedoraproject.org/admin/ it shows that group in my principals, which should ideally correspond with the Pyramid ACL.

ralphbean:
That's cool.

We would still have to return the list of all the committers after that even if they weren't going to be referenced in the check (we can't short-circuit on proven-packager).

lmacken:
Yep. This looks fine to me, :+1:

PR #557, opened 2015-09-10
Short-circuit recursion specifically for test cases.

### LINKED ISSUE #343: Remove unscalable items from the testcase JSON
See #270.  There may be some unsustainable information included in the testcase json that will need to be removed.

ralphbean:
These get artificially added in, so we need to artificially prune them as well.

Fixes #343.

This cuts the bytes-returned by queries down in half *again*::

❯ http get http://0.0.0.0:6543/updates/ | wc
      0   85607 1087158
❯ http get http://0.0.0.0:6543/updates/ | wc
      0    4919   58298

### DIFF
--- a/bodhi/models/models.py
+++ b/bodhi/models/models.py
@@ -1806,8 +1806,9 @@ class Update(Base):
                 'Unable to determine requested tag for %s.' % self.title)
         return tag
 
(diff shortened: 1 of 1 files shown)


lmacken:
Nice! :+1:

PR #558, opened 2015-09-10
masher: Ensure updates cannot be partially tagged before begin ejected

### LINKED ISSUE #536: Large update can be partially tagged before ejected from push.
In the case of this large KDE update https://bodhi.fedoraproject.org/updates/FEDORA-2015-14271 (issue #504) one of the builds was not tagged properly, and was not ejected until a number of builds were queued up to have tags added.

The part of the code responsible for this bug is in the `_determine_tag_actions` in `masher.py`:

```python
for build in update.builds:                                   
    from_tag = None                                           
    tags = build.get_tags()                                   
    for tag in tags:                                          
        if tag in tag_types[status]:                          
            from_tag = tag                                    
            break                                             
    else:                                                     
        reason = 'Cannot find relevant tag for %s: %s' % (    
            build.nvr, tags)                                  
        self.eject_from_mash(update, reason)                  
        break                                                 
                                                              
    if self.skip_mash:                                        
        self.add_tags.append((update.requested_tag, build.nvr)
    else:                                                     
        self.move_tags.append((from_tag, update.requested_tag,
                               build.nvr))                    
```

So a bunch of builds in the update get to `self.add_tags`, but once ejected those are never removed. Ideally, I think we should look for relevant tags in the previous `verify_updates` method.

lmacken:
fixes #536

### DIFF
--- a/bodhi/consumers/masher.py
+++ b/bodhi/consumers/masher.py
@@ -505,6 +505,9 @@ class MasherThread(threading.Thread):
     def _determine_tag_actions(self):
         tag_types, tag_rels = Release.get_tags()
         for update in sorted_updates(self.updates):
+            add_tags = []
+            move_tags = []
+
             if update.status is UpdateStatus.testing:
                 status = 'testing'
             else:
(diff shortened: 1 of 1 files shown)


lmacken:
I think my latest commit addresses your concerns?

ralphbean:
> I think my latest commit addresses your concerns?

Yes!  Much better than my scrubbery.  :+1:!

PR #560, opened 2015-09-10
Set thread name for masher logging.

### LINKED ISSUE #555: Setup custom loggers per masher thread with the koji tag as a prefix
Right now it's difficult to differentiate which thread in the masher logs corresponds with which koji tag. We should have them setup a `[tagname]` logging prefix so it's clear which thread each message is from.

ralphbean:
With an extra change to our ansible logging configuration in ``/etc/fedmsg.d/``, this will fix #555.

### DIFF
--- a/bodhi/consumers/masher.py
+++ b/bodhi/consumers/masher.py
@@ -260,6 +260,10 @@ class MasherThread(threading.Thread):
                               .filter_by(name=self.release).one()
         self.id = getattr(self.release, '%s_tag' % self.request.value)
 
+        # Set our thread's "name" so it shows up nicely in the logs.
+        # https://docs.python.org/2/library/threading.html#thread-objects
+        self.name = self.id
+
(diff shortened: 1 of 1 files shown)


lmacken:
Elegant! :+1: :dancer:

ralphbean:
See the ``threadName`` value in the [stdlib docs](https://docs.python.org/2/library/logging.html#logrecord-attributes)?

I think for production config all we have to do is modify our [bodhi-specific logging config here](https://infrastructure.fedoraproject.org/cgit/ansible.git/tree/roles/bodhi2/base/templates/bodhi-logging.py) to add a formatter that uses ``threadName`` kind of like the standard [bare formatter that we use for all our other fedmsg instances](https://infrastructure.fedoraproject.org/cgit/ansible.git/tree/roles/fedmsg/base/templates/logging.py.j2#n137).

PR #561, opened 2015-09-11
bugzilla ticket number have 6+ digits

remicollet:
Ticker 99999 is >8 years old.

This is mostly useful to avoid the pattern to be raised when update description is a copy/paste from upstream changelog, with upstream bug ref.

And I think most projects have bug < 100,000 (e.g. PHP is ~70,000)

See https://bodhi.fedoraproject.org/updates/FEDORA-2015-14978

### DIFF
--- a/bodhi/ffmarkdown.py
+++ b/bodhi/ffmarkdown.py
@@ -62,7 +62,7 @@ def inject():
             return el
 
     MENTION_RE = r'(@\w+)'
-    BUGZILLA_RE = r'(#[0-9]{5,})'
+    BUGZILLA_RE = r'(#[0-9]{6,})'
 
     class SurroundProcessor(markdown.postprocessors.Postprocessor):
         def run(self, text):


bochecha:
This is going to break a unit test, specifically `bodhi/tests.functional.test_generic:TestGenericViews.test_markdown_with_bugzilla`.

But then, this will only provide you temporary relief, as long as the PHP bug ids are under 6 digits.

Instead of this pull request, I suggest we adopt something like I implemented in pull request #562.

bochecha:
Also, some bug reports in RHBZ have less than 6 digits, and yet they are still open, and might eventually get closed by a future update. (not holding my breath :stuck_out_tongue_winking_eye: )

remicollet:
sorry for the test

I think both PR are valuable, and can work together

remicollet:
> Also, some bug reports in RHBZ have less than 6 digits,

so rhbz#123 will make sense for this very uncommon case ;)

PR #562, opened 2015-09-11
Markdown bugurls

bochecha:
This allows writing things like `python#12345` and get it markdownified into a link to bug 12345 in the upstream Python bug tracker.

### DIFF
--- a/bodhi/ffmarkdown.py
+++ b/bodhi/ffmarkdown.py
@@ -29,8 +29,16 @@ def user_url(name):
     return request.route_url('user', name=name)
 
 
-def bugzilla_url(idx):
-    return "https://bugzilla.redhat.com/show_bug.cgi?id=%s" % idx
+def bug_url(tracker, idx):
+    try:
+        return {
+            'fedora': "https://bugzilla.redhat.com/show_bug.cgi?id=%s",
+            'rh': "https://bugzilla.redhat.com/show_bug.cgi?id=%s",
+            'rhbz': "https://bugzilla.redhat.com/show_bug.cgi?id=%s",
+            }[tracker.lower()] % idx
+
+    except KeyError:
+        return None
 
 
 def inject():
@@ -55,14 +63,20 @@ def inject():
 
     class BugzillaPattern(markdown.inlinepatterns.Pattern):
         def handleMatch(self, m):
+            tracker = markdown.util.AtomicString(m.group(2))
+            idx = markdown.util.AtomicString(m.group(3))
+            url = bug_url(tracker, idx[1:])
+
+            if url is None:
+                return tracker + idx
+
             el = markdown.util.etree.Element("a")
-            idx = markdown.util.AtomicString(m.group(2))
-            el.set('href', bugzilla_url(idx[1:]))
+            el.set('href', url)
             el.text = idx
             return el
 
     MENTION_RE = r'(@\w+)'
-    BUGZILLA_RE = r'(#[0-9]{5,})'
+    BUGZILLA_RE = r'([\S]+)(#[0-9]{5,})'
 
     class SurroundProcessor(markdown.postprocessors.Postprocessor):
         def run(self, text):
--- a/bodhi/ffmarkdown.py
+++ b/bodhi/ffmarkdown.py
@@ -33,6 +33,12 @@ def bug_url(tracker, idx):
     try:
         return {
             'fedora': "https://bugzilla.redhat.com/show_bug.cgi?id=%s",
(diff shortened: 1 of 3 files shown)


ralphbean:
Looks pretty cool.

Can you update the [docs in the pop-up modal](https://github.com/fedora-infra/bodhi/blob/develop/bodhi/templates/fragments.html#L214-L216) too?

bochecha:
Sure thing.

remicollet:
I think "5" is not correct for all upstream

lmacken:
Nice! I also use `rh#1234` quite a bit too.

bochecha:
So, talking about this with @remicollet on IRC, I'm realizing that what I proposed here doesn't suit his need.

Le me explain...

@remicollet often creates updates (usually PHP packages, but not exclusively), for which he copy-pastes the (usually well-formatted) upstream changelog into the Bodhi note field.

His biggest problem is that the current markdown parser will match strings like `#12345` and change them into links to RHBZ. It is a problem for him because those bug ids are for the upstream bug tracker.

That's why @remicollet originally submitted #561, because RHBZ tickets are on the 6+ digits, but his upstream bug ids all have 5 or less digits.

So that would solve part of his problem, and his update notes would not point to RHBZ tickets incorrectly... until his upstreams get to 6+ digits. (GNOME upstream is well above 700000 for example)

Therefore, I'm starting to think that what we should do is not match any more the unprefixed `#1234` at all, but only `foo#1234`.

That also solves @remicollet's problem (even when his upstreams get to 6+ digits), but then it breaks the current assumption that "no prefix means RHBZ".

In addition, it avoids matching too eagerly things like `#1` which could be used for other things than bug reports. (`libfoobar is the #1 library to foo some bar!`)

So what should we do here?

remicollet:
+1 to drop unprefixed pattern

lmacken:
I'm :+1: for dropping unprefixed patterns as well

ralphbean:
Same here, sounds good.

bochecha:
So while we're at it, what about the fallback case where we don't know the prefix?

Right now, my patch falls back to RHBZ (the `except KeyError` case).

Should we instead just not transform the text into an URL at all?

ralphbean:
> Should we instead just not transform the text into an URL at all?

Yeah, that makes sense to me.

lmacken:
Solid. :+1:

bodhi/ffmarkdown.py
```diff
@@ -29,8 +29,21 @@ def user_url(name):
...
 
 
-def bugzilla_url(idx):
-    return "https://bugzilla.redhat.com/show_bug.cgi?id=%s" % idx
+def bug_url(tracker, idx):
+    try:
+        return {
+            'fedora': "https://bugzilla.redhat.com/show_bug.cgi?id=%s",
+            'gnome': "https://bugzilla.gnome.org/show_bug.cgi?id=%s",
+            'kde': "https://bugs.kde.org/show_bug.cgi?id=335161",
```
ralphbean: the 335161 here needs to be replaced with ``%s``.

bodhi/templates/fragments.html
```diff
@@ -214,6 +214,9 @@ <h4 class="modal-title" id="markdown-help-label">Fedora-Flavored Markdown</h4>
               <p>You can reference <strong>bugzilla tickets</strong> by simply prefixing a ticket id with the <code>&#35;</code> character.</p>
               <pre><code>We really need #1185409</code></pre>
               <p>... we will automatically generate a link to the ticket in place.</p>
+              <p>You can also specify the <strong>ticket tracker</strong> if your update is fixing a ticket reported elsewhere than in the Fedora bugzilla.</p>
```
ralphbean: We should re-work the docs just above this that say you can use the old un-prefixed format.

bodhi/ffmarkdown.py
```diff
@@ -29,8 +29,21 @@ def user_url(name):
...
+            'pear': "http://pear.php.net/bugs/bug.php?id=%s",
+            'php': "https://bugs.php.net/bug.php?id=%s",
+            'python': "https://bugs.python.org/issue%s",
+            'rh': "https://bugzilla.redhat.com/show_bug.cgi?id=%s",
+            'rhbz': "https://bugzilla.redhat.com/show_bug.cgi?id=%s",
```
ralphbean: Could get out of hand.. but do we want to add just ``'bz'`` as a prefix?

bochecha:
Ok, so I'm happy with what's pushed now. Ack? Nack?

ralphbean:
Ack!  :+1:!

PR #568, opened 2015-09-11
Fix the 'pushed' column.

### LINKED ISSUE #564: Should not offer to push to testing on stable update
When I visit https://bodhi.fedoraproject.org/updates/FEDORA-2015-15193 I see a "Push to testing" button in the upper right next to edit.  This shouldn't be offered as the update is already pushed to stable.

ralphbean:
* Fixes #564.


lmacken:
:+1:

PR #569, opened 2015-09-11
Restructure the way we assign aliases.

### LINKED ISSUE #567: Change the structure of update aliases
Right now they are ``<PREFIX>-<YEAR>-<INTEGER>``, where the last part grows monotonically with each new update.

Let's considering changing that to ``<PREFIX>-<YEAR>-<HASH>``, where hash is the first 8 characters of a hash of a random uuid.

The problem with the way things are now is that:

- we have a race condition.  If you try to submit two updates at the same time (via a script, this is easy to do), then they both go into the code where their alias is assigned at the same time, they both get the same integer, and then when they try to commit their database transactions, one wins and the other fails.
- the alias-assigning code is one of the slower parts of the new update process.  It takes 1-2s iirc (because it has to query for the latest update before it and figure out its id).

The downside here is that the update aliases will no longer be meaningful like they were.  You won't be able to look at the latest one and instantly know how many updates there have been so far in 2015, for instance.  IMHO, this isn't worth the trouble they're causing us.  Up for discussion!

ralphbean:
Fixes #567.

Before this change, they were ``<PREFIX>-<YEAR>-<INTEGER>``, where the last
part grows monotonically with each new update.

This commit changes that to ``<PREFIX>-<YEAR>-<HASH>``, where hash is the first
10 characters of a hash of a random uuid.

The problem with the way things are now is that:

- we have a race condition.  If you try to submit two updates at the same time
  (via a script, this is easy to do), then they both go into the code where
  their alias is assigned at the same time, they both get the same integer, and
  then when they try to commit their database transactions, one wins and the
  other fails.
- the alias-assigning code is one of the slower parts of the new update
  process.  It takes 1-2s iirc (because it has to query for the latest update
  before it and figure out its id).

The downside here is that the update aliases will no longer be meaningful like
they were.  You won't be able to look at the latest one and instantly know how
many updates there have been so far in 2015, for instance.  IMHO, this isn't
worth the trouble they're causing us.  Up for discussion!

### DIFF
--- a/bodhi/models/models.py
+++ b/bodhi/models/models.py
@@ -17,8 +17,10 @@
 import os
 import re
 import copy
+import hashlib
 import json
 import time
+import uuid
 
 from textwrap import wrap
 from datetime import datetime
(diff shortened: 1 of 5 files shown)


pypingou:
:+1: I like the change

bodhi/models/models.py
```diff
@@ -989,44 +991,15 @@ def get_testcase_karma(self, testcase):
...
-        for this release, increments it and prefixes it with the id_prefix of
-        the release and the year (ie FEDORA-2007-0001).
+        This function used to construct update IDs in a monotonic sequence, but
+        we ran into race conditions so we do it randomly now.
         """
-        release = self.release
-        releases = DBSession.query(Release)\
-                            .filter_by(id_prefix=release.id_prefix)\
-                            .all()
-        subquery = DBSession.query(Update.date_submitted).filter(
```
lmacken: So happy to kill this query with :fire:

lmacken:
Yes, I'm so sick of struggling to increment an integer. :+1:

PR #584, opened 2015-09-18
masher: Remove the pending tag when ejecting an update from the push …

### LINKED ISSUE #583: Removing pending tag from ejected updates
When an update is ejected from a push, the `request` is removed, which means we should probably remove the `-pending` tag as well so Taskotron doesn't get confused.

lmacken:
…(fixes #583)


ralphbean:
Sounds good to me.  :+1:

lmacken:
Thanks :)

PR #598, opened 2015-09-23
Add links to the expired buildroot overrides of a the user on said user page

pypingou:


ralphbean:
:+1: to the idea.

- How about keeping 'expired', 'active', and 'all'?  (three options, like for updates).
- I think as written, the "active" link will just take you to the "all" page.

pypingou:
From my tests it looked like the ``expired=0`` was equivalent to not specifying ``expired`` in the URL that's why I didn't include but maybe I tested it wrongly

ralphbean:
Yeah, compare:

- [expired for ralph](https://bodhi.fedoraproject.org/overrides/?user=ralph&expired=1)
- [active for ralph](https://bodhi.fedoraproject.org/overrides/?user=ralph&expired=0)
- [all for ralph](https://bodhi.fedoraproject.org/overrides/?user=ralph)

pypingou:
ok didn't test with the right user then, cool :)

PR #609, opened 2015-09-25
Port everything to a request-local db session.

ralphbean:
.. in a quest to kill the memory leak.

One thing we get out of this is the ability to explicitly close the sessions that we open.

For some context:

- We used to use global dbsessions for everything.  This is a habit we learned from TurboGears back in the day and it turns out to be a bad one.  With the rewrite to Bodhi2, we ported *most all* instances of that to instead rely on a request-local db session which is nicely scoped, more testable, etc.  This 27-file change ports over the *last* of those old global session references and it removes the global session all together.  Starting this I thought it would be easy since there were only a few instances, but it turned out to be more involved.  These last spots were some of the more tangled places in the code.

The test suite is all passing and smoke-testing with a local instance works (submitting comments, poking around, etc...).

### DIFF
--- a/bodhi/__init__.py
+++ b/bodhi/__init__.py
@@ -16,6 +16,8 @@ from collections import defaultdict
 from dogpile.cache import make_region
 from munch import munchify
 from sqlalchemy import engine_from_config
+from sqlalchemy.orm import scoped_session, sessionmaker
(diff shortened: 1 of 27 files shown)


lmacken:
Wow, way more work than I expected. Nice job! :+1:

PR #625, opened 2015-10-01
models: Use dynamic lazy loading for User.{comments,updates}

lmacken:
This should hopefully address some of the memory/CPU/db issues discussed in
issue #301


ralphbean:
:+1:  Definitely try it in staging, but :+1:

lmacken:
Yeah, I wrote it in staging first ;)

Issue #89, opened 2014-09-09
Push status page


lmacken:
It would be cool to have a page that displays the status of the updates push process.

PR #629, opened 2015-10-02
Initial stab at a masher status page.

ralphbean:
It's way more simple than the releng dash.  But it may just do the job.

Notice the things that are commented out at the bottom of masher-status.js.  Those can be uncommented if we like, or not.  I thought they were a little spammy, but try it out and see what you think.


ralphbean:
masher-status

ralphbean:
Fixes #89.

lmacken:
:money_with_wings:

PR #634, opened 2015-10-06
Fixing the search issue which was redirecting users to a 404 page when packages are being searched

farhaanbukhsh:


bodhi.spec
```diff
@@ -11,6 +11,7 @@ URL:            https://fedorahosted.org/bodhi
 Source0:        https://fedorahosted.org/releases/b/o/bodhi/bodhi-%{version}.tar.gz
 BuildRoot:      %{_tmppath}/%{name}-%{version}-%{release}-root-%(%{__id_u} -n)
 BuildArch:      noarch
+ExcludeArch:    ppc64 ppc
```
puiterwijk: Why do you want to exclude arches in a no arch package? That doesn't even work, and doesn't make any sense.

puiterwijk:
@farhaanbukhsh is creating a new PR from a newly created branch.

PR #635, opened 2015-10-06
Fixing the search issue which was redirecting users to a 404 page when packages are being searched

farhaanbukhsh:
…n packages are being searched


ralphbean:
Ah, but I think this will break the search if you are actually searching for a user.

The trick is that the code that *used* to handle the redirect for packages is in the code block just above this.  See where it checks if ``datum.builds`` is there?  That ``builds`` field disappeared, which is what caused it to break.  **That** conditional is the one that needs to be modified (and this one needs to be reset back to the way it was).  We need to check for the existence of something else other than ``builds``.

ralphbean:
:+1: let's get this merged and deployed!

farhaanbukhsh:
thanks a ton :dancer:

PR #639, opened 2015-10-06
scripts: Add a bodhi-approve-testing script (fixes #298)

### LINKED ISSUE #298: Implement nagmail script
We need a job that comments on updates once they meet the minimum amount of time in testing, etc.

https://github.com/fedora-infra/bodhi/blob/master/bodhi/jobs.py#L74

lmacken:


ralphbean:
:+1: as is

Some thoughts though:

- It is confusing that we have both ``meets_testing_requirements`` and ``met_testing_requirements`` and that they don't share code.  I think ``met`` does the same things as ``meets`` except it tests for a comment also.  Perhaps ``met`` could be refactored to call ``meets`` first, and then check for the comment?
- It would be nice to add a test that has an update the ``meets`` but hasn't yet ``met``.   Assert those things are true and false respectively.  Add the comment.  Then finally assert that both are true afterwards.

lmacken:
Okay, I think I addressed both of your concerns?

ralphbean:
:+1:

lmacken:
Thanks :)

PR #651, opened 2015-10-11
Include url to an update in JSON body.

### LINKED ISSUE #600: API: provide URL for updates
When I run an API query for updates, for each update I get a dict (well, a `Munch`, fine) with lots of useful keys, but none that gives the update's URL. This would be rather handy. It's relatively easy to construct it from the client instance's `base_url` attribute, but it requires your app to have some random magic knowledge - that the URL is `{0}updates/{1}.format(client.base_url, update['updateid'])` - and it's a bit awkward if the place where you want to use the URL has the update dict but doesn't have access to the client instance (you have to add the URL to the update dicts when you initially get them). It'd be nice if this was just there.

ralphbean:
Fixes #600.

### DIFF
--- a/bodhi/models/models.py
+++ b/bodhi/models/models.py
@@ -584,7 +584,7 @@ class Build(Base):
 class Update(Base):
     __tablename__ = 'updates'
     __exclude_columns__ = ('id', 'user_id', 'release_id', 'cves')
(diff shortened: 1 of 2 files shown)


lmacken:
Look good! :+1:

PR #663, opened 2015-10-19
Let people get quick search results when they press enter.

### LINKED ISSUE #229: Can't search?
I'm trying to use the search functionality to search for a package. All I get is "Unable to find any users matching the current query".

ralphbean:
This should finally fix #229.


pypingou:
:+1: for me

PR #668, opened 2015-10-21
Use https for comps Repo

tyll:
git:// is insecure, therefore do not use it.

Plase also remove the old comps repo on the systems bodhi runs since bodhi will not update the comps URL afaics.


ralphbean:
Nice!  :+1: here.

lmacken:
Yep, nice catch @tyll. Thanks! :+1:

Issue #666, opened 2015-10-20
Cannot push Eclipse update to F22 stable!


mbooth101:
Please see: https://bodhi.fedoraproject.org/updates/FEDORA-2015-2ba0ca6444

I have tried twice to push this to stable, but bodhi later comments on the update with:

> ... ejected from the push because u"Cannot find relevant tag for eclipse-4.5.1-1.fc22: ['f22-updates-candidate', 'f22-updates-pending', 'f22-override']"

It's not obvious to me what I need to do to fix this, please help!

ralphbean:
So, it is getting ejected from this method call: https://github.com/fedora-infra/bodhi/blob/develop/bodhi/consumers/masher.py#L514-L545

ralphbean:
After a little digging, in order for an update to be requested for stable, all of its builds must be in one of these tags:

```python
[u'f23-updates-testing', u'epel7-testing', u'dist-6E-epel-testing', u'dist-5E-epel-testing', u'f22-updates-testing', u'f21-updates-testing']
```

But that ejection notice indicates that eclipse-4.5.1-1.fc22 is somehow mis-tagged.  It is only in ``['f22-updates-candidate', 'f22-updates-pending', 'f22-override']``, and not in ``['f22-updates-testing']``.

I don't have an explanation for why that is, but tagging it correctly would help.  Does it appear in the updates-testing repos?

mbooth101:
Hmm, maybe this is related to #655 then? I tried to edit the update but I was unable to. One of the things I wanted to do was add an updated build for this package (eclipse-4.5.1-2.fc22) but bodhi did not let me

mbooth101:
So to answer your question, there should be an even build than that in updates-testing

ralphbean:
OK, I removed eclipse-4.5.1-1.fc22 from https://bodhi.fedoraproject.org/updates/FEDORA-2015-2ba0ca6444 to try and get things moving again.  @mbooth101, please let us know if you can get these to stable.

mbooth101:
@ralphbean My update failed again, same error, different package :-(

https://bodhi.fedoraproject.org/updates/FEDORA-2015-2ba0ca6444

ralphbean:
OK, as discussed in ``#fedora-admin`` I removed eclipse-jgit-4.1.0-2.fc22 from the update as well.

ralphbean:
Looks like this made it to stable.

PR #669, opened 2015-10-21
Clarify ejection reason.

ralphbean:
This should help to make the reason for cases like #666 more self-evident in the future.


lmacken:
Yep, looks good! :+1:

PR #670, opened 2015-10-21
Allow the same people to give karma to an update that has had it's karma reset

### LINKED ISSUE #659: Incorrect karma after reset
https://bodhi.fedoraproject.org/updates/FEDORA-2015-40a4d314f3

It looks like this update was edited and had it's karma reset, then the same 2 people gave feedback again, but it is not reflected in the overall karma.

lmacken:
Fixes #659

### DIFF
--- a/bodhi/models/models.py
+++ b/bodhi/models/models.py
@@ -1386,41 +1386,48 @@ class Update(Base):
                 notice = 'You may not give karma to your own updates.'
                 caveats.append({'name': 'karma', 'description': notice})
 
-        if not anonymous and karma != 0 and \
-           not filter(lambda c: c.user.name == author and c.karma == karma,
-                      self.comments):
-            mycomments = [
-                c.karma for c in self.comments if c.user.name == author]
-            if karma == 1 and -1 in mycomments:
-                self.karma += 2
-                caveats.append({
-                    'name': 'karma',
-                    'description': 'Your karma standing was reversed.',
-                })
-            elif karma == -1 and 1 in mycomments:
-                self.karma -= 2
-                caveats.append({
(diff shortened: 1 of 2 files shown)


bodhi/models/models.py
```diff
@@ -1386,41 +1386,48 @@ def comment(self, session, text, karma=0, author=None, anonymous=False,
...
-                    pass
-                except BodhiException as e:
-                    # This gets thrown if the karma is pushed over the
-                    # threshold, but it is a critpath update that is not
-                    # critpath_approved. ... among other cases.
-                    log.exception('Problem checking the karma threshold.')
+        if not anonymous and karma != 0:
+            # Take all comments since the previous karma reset
+            reset_index = 0
+            for i, comment in enumerate(self.comments):
```
ralphbean: Do we need to iterate over comments in ``reverse`` here?

I'm thinking that this might work correctly if there was one edit, but it might get screwy if there were two edits (with some karma comments in between).

lmacken: I was initially writing it to iterate in reverse, but this one should start with the most recent edit message. I /think/ it works correctly? Even if there are 2 edits with karma in between, it'll ignore everything before the most recent edit, I believe.

lmacken: I just updated the test to reset the karma twice and have the same person +1 3 times.

ralphbean: Yup.  I think I read this hastily earlier and just assumed that it would ``break`` in the loop.  I see now that it doesn't and you're right -- this should work as written.

ralphbean:
:+1:

lmacken:
Thanks!

PR #674, opened 2015-10-26
Provide a more useful message when the csrf check fails.

### LINKED ISSUE #667: CSRF tokens do not match - but no advice
I get this quite often after trying to submit a Bodhi comment:
bodhi

I suppose the reason is that I tend to keep several Bodhi updates open at the same time and leave them open for quite long before I add the comment (I actually need to test stuff, before commenting:)).

If it is not technically possible to eliminate or reduce the likelihood of this issue, OK. But let's make the error message at least helpful. Could you replace it with something like this?

> Error adding comment - CSRF tokens do not match. This happens if you have the page open for a long time. Please reload the page and try to add the comment again. Make sure to save your feedback somewhere before reloading.

Also, it would be helpful to make this more visible. The floating bars in the right corner are not very visible. For errors like these, it would be better to show it in the center of the screen with an OK button and dim the rest of the screen, or maybe show it in a highly visible bar above the Comment box or the Add Comment button.

Thanks!

ralphbean:
Fixes #667.

### DIFF
--- a/bodhi/validators.py
+++ b/bodhi/validators.py
@@ -34,13 +34,18 @@ from .util import get_nvr, tokenize, taskotron_results
 import bodhi.schemas
 
 
+csrf_error_message = """CSRF tokens do not match.  This happens if you have
+the page open for a long time. Please reload the page and try to submit your
+data again. Make sure to save your input somewhere before reloading.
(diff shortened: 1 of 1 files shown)


lmacken:
:+1:

Issue #671, opened 2015-10-22
Browser severly affected by adding a comment using markdown


vpodzime:
Tried to add this as a comment to an update:

```
An easy way to test the functionality of the package is to install the library and the python bindings, then run ipython and type something like this in:

    >>> from gi.repository.ByteSize import Size
    >>> s = Size("10 GiB")
    >>> s2 = Size("5 GiB")
    >>> s + s2
```
and I got no preview, just *Loading...* and my browser wasn't able to load any other tab -- *firefox-41.0.1-2.fc21.x86_64*

ralphbean:
I can't reproduce this.  Do you still see it?

If so, can you open up your browser's JS console and tell us if you see any errors or weird things there?

vpodzime:
Just tried it and my browser is okay and I got the preview. So I guess it's good! However, I discovered another issue -- https://github.com/fedora-infra/bodhi/issues/677

ralphbean:
Thanks @vpodzime!

PR #464, opened 2015-09-03
Fix overflow in markdown comment areas.

ralphbean:
Fixes #459.

overflow

PR #675, opened 2015-10-26
Explicitly namespace the white-space css rule.

ralphbean:
This fixes an issue with the way code-blocks are rendered while maintaining the
behavior we want introduced in #464.

I noticed it while trying to reproduce #671.


lmacken:
:open_mouth: :+1:

PR #676, opened 2015-10-26
Add convenience links from the update page to override management.

### LINKED ISSUE #661: [RFE] Add override button when an update for a build is created
Hi,

when I update a golang package in Fedora it is always followed by override of its build. Would it be possible to add a button to open a page (bodhi.fedoraproject.org/overrides/new) with override form filled with package name, build and bug id?

Jan

ralphbean:
- If an override exists for each build, add a link to go and edit each one.
- If an override does not exist, add a link to go and create each one.
- If you cannot edit the update, do not show those links.

Fixes #661.

### DIFF
--- a/bodhi/templates/override.html
+++ b/bodhi/templates/override.html
@@ -49,8 +49,14 @@
           </div>
           <div id="candidates-search" class="form-group noui">
             %if override is UNDEFINED:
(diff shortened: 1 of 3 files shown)


lmacken:
Looks good to me! :+1:

Issue #624, opened 2015-10-01
Unhandled exceptions don't tell us anything about themselves


ralphbean:
We often see in the logs something like:

```
Unhandled exception raised
None
```

which is worthless.  We need to figure out what's hiding the exceptions there and uncover them.

ralphbean:
So, after #680, we will be getting the exception message, but not the traceback.

I'm leaving this open until either:

- we figure out how to get the traceback or..
- we decide that the message is enough and we can close this as is.

bowlofeggs:
I haven't observed this problem before, so I am closing this for now.

PR #680, opened 2015-10-27
Log a little more information about unhandled errors.

ralphbean:
What's weird is that we have *two* other error catchers.  In total:

- We have a *per-service* cornice error handler, defined in ``bodhi/services/errors.py``.
- We have a *pyramid-wide* handler in `bodhi/views/generic.py`.
- Lastly, this *cornice-wide* error handler.

In local testing, I can introduce errors that get handled by the first and
second handlers, but not this one.  Oddly, it seems we're hitting it in
production all the time with that ``None`` error mentioned in #624.

So, I'm not sure what's going on, but hopefully this patch will help us figure
out what's happening, fixing #624.

### DIFF
--- a/bodhi/__init__.py
+++ b/bodhi/__init__.py
@@ -109,7 +109,7 @@ def get_releases(request):
 def exception_filter(response, request):
     """Log exceptions that get thrown up to cornice"""
     if isinstance(response, Exception):
-        log.exception('Unhandled exception raised')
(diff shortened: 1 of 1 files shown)


pypingou:
Cool :+1:

lmacken:
Yeah, those `None` errors are seriously weird... hopefully this helps ;) :+1:

PR #681, opened 2015-10-27
Hide the js warning as early as possible.

### LINKED ISSUE #627: Don't show "page only works with JavaScript enabled" warning when JS is enabled
When loading various pages, including the New Update page, Bodhi flashes a warning that "this page only works with JavaScript enabled." Then the warning disappears after a fraction of a second, because I do in fact have JavaScript enabled.

ralphbean:
Fixes #627, by not waiting for ``$(document).ready(...);``.

### DIFF
--- a/bodhi/static/js/override_form.js
+++ b/bodhi/static/js/override_form.js
@@ -84,7 +84,6 @@ $(document).ready(function() {
         theform.expire();
     });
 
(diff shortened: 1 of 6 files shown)


pypingou:
Looks alright code-wise

lmacken:
Cool :+1:

PR #684, opened 2015-10-27
Allow users to disable global popups.

### LINKED ISSUE #660: Allow disabling event popups in web UI
While I'm trying to perform a task via the web UI, Bodhi keeps showing pop-up notifications saying that someone has done something with some random update.  This is distracting and also completely pointless, since those events are not actionable or even relevant to me.  Can we have a way to turn this off?

popups

ralphbean:
Fixes #660.

People will still get popups about actions they're undertaking (like,
commenting on an update or error messages when they submit an incomplete form).

### DIFF
--- a/bodhi/models/models.py
+++ b/bodhi/models/models.py
@@ -37,6 +37,7 @@ from sqlalchemy import Unicode, UnicodeText, Integer, Boolean
 from sqlalchemy import DateTime
 from sqlalchemy import Table, Column, ForeignKey
 from sqlalchemy import and_, or_
+from sqlalchemy.sql import text
 from sqlalchemy.orm import relationship, backref
 from sqlalchemy.orm import class_mapper
 from sqlalchemy.orm.properties import RelationshipProperty
@@ -2006,6 +2007,9 @@ class User(Base):
 
     name = Column(Unicode(64), unique=True, nullable=False)
     email = Column(UnicodeText, unique=True)
+
+    # A preference
+    show_popups = Column(Boolean, default=True, server_default=text('TRUE'))
 
     # One-to-many relationships
     comments = relationship(Comment, backref=backref('user'), lazy='dynamic')
--- a/bodhi/views/generic.py
+++ b/bodhi/views/generic.py
@@ -20,7 +20,8 @@ import cornice.errors
 from pyramid.security import authenticated_userid
 from pyramid.settings import asbool
 from pyramid.view import view_config, notfound_view_config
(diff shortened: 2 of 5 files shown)


bodhi/models/models.py
```diff
@@ -2007,6 +2008,9 @@ class User(Base):
     name = Column(Unicode(64), unique=True, nullable=False)
     email = Column(UnicodeText, unique=True)
 
+    # A preference
+    show_popups = Column(Boolean, default=True, server_default=text('TRUE'))
```
pypingou: ``server_default=True`` doesn't work?

bodhi/views/generic.py
```diff
@@ -193,6 +194,25 @@ def new_override(request):
...
+    from bodhi.models import User
+    userid = authenticated_userid(request)
+    if userid is None:
+        raise HTTPForbidden("You must be logged in.")
+    user = request.db.query(User).filter_by(name=unicode(userid)).first()
+    if user is None:
+        raise HTTPBadRequest("For some reason, user does not exist.")
+
+    # Toggle the value.
+    user.show_popups = not user.show_popups
```
pypingou: I guess we have autocommit on then?

ralphbean: We do, and it does.  :)

bodhi/models/models.py
ralphbean: Yeah, I tried ``server_default=True`` first and it balked.  I tried this a few times with upgrades and downgrades.

bodhi/views/generic.py
```diff
@@ -193,6 +194,25 @@ def new_override(request):
...
 
 
+@view_config(route_name='popup_toggle', request_method='POST')
+def popup_toggle(request):
+    # Get the user
+    from bodhi.models import User
+    userid = authenticated_userid(request)
+    if userid is None:
+        raise HTTPForbidden("You must be logged in.")
+    user = request.db.query(User).filter_by(name=unicode(userid)).first()
```
lmacken: Can't we just do `request.user`?

ralphbean: Nah, ``request.user`` is a ``munch`` object now.  So, we can *get* the ``show_popups`` value from it, but we can't *set* it.

lmacken: Ah, gotcha. LGTM then! :+1:

lmacken:
:fire: :dancers: :boom:

PR #685, opened 2015-10-27
Add more info to the mash summary logging.

### LINKED ISSUE #620: Log a summary instead at the end of the mash
Right now it just says "push complete" https://github.com/fedora-infra/bodhi/blob/develop/bodhi/consumers/masher.py#L207

but... it would be cool if it could print out some sort of summary of all the branches it tried, whether they succeeded or failed, and if any updates were ejected for any of those.  Something so that the operator can know if they need to do anything special or not.

ralphbean:
This should fix #620.  I wasn't able to test this, so give it two looks over before merging.


lmacken:
Looks good to me.

PR #686, opened 2015-10-28
Ensure there is an user before we try to access show_popups

pypingou:
This fix the front page when the user is not logged in


puiterwijk:
Looks good to me. :+1:

ralphbean:
:+1: here too.

Issue #623, opened 2015-10-01
required notifications should tell you what is required


ralphbean:
simo tells me that if you update submit the form under certain conditions, you get a notification back that just says "required".  it means that one of the form fields you submitted is required but empty.. but it doesn't tell you *which* field that is.

It should already be in the JSON data.. we just need to show it via messenger.js

simo5:
As requested, screenshot attached

![bodhi-required-error](https://cloud.githubusercontent.com/assets/8332609/10231671/0d23a8b2-6852-11e5-8300-f3f16920a2c4.png)

ralphbean:
As discussed in IRC, this is because the "updates notes" textarea is blank.

We still need to update the little error popup to indicate that though.

PR #687, opened 2015-10-28
Specify for which field we are returning a notification

pypingou:
Fixes https://github.com/fedora-infra/bodhi/issues/623


ralphbean:
Ha!  I bet this will work.

Does it work?

pypingou:
Oh, sorry I had prepare a screenshot and forgot to upload it, so here it is:
bodhi_notifications

ralphbean:
Looks great!  :+1:

Issue #652, opened 2015-10-12
Prune out feedback from bodhi on the user profile


ralphbean:
Here's an idea.

The bottom half of a user's profile looks like this:

![feedback-received](https://cloud.githubusercontent.com/assets/331338/10431216/f4ba6878-70d1-11e5-9443-f9dd619f0b65.png)

The info on the left includes comments left on my updates by both *real users* and by the system bodhi user.  Those comments from bodhi are useful when looking at the page for a particular update, but they can be kind of spammy on the user profile page.  Can we remove them, and list only comments from people *other* than the bodhi user?

pypingou:
This ticket is still formulated as a question. The pull-request linked above is a proposed implementation of a solution to address this ticket, but if we decide to keep these comments, then we can close the PR w/o merging it.

PR #689, opened 2015-10-28
Add the possibility to ignore the comments made by a specific user

pypingou:
This allows to ignore comments made by the bodhi user to the user's
updates on the user's page.

Fixes https://github.com/fedora-infra/bodhi/issues/652

### DIFF
--- a/bodhi/schemas.py
+++ b/bodhi/schemas.py
@@ -602,6 +602,12 @@ class ListCommentSchema(PaginatedSchema, SearchableSchema):
         missing=None,
     )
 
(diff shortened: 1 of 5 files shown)


pypingou:
This pull-request address a ticket that is still formulated as a question, so we may want to be decide if it is a good idea or not before merging it :)

ralphbean:
Code looks good to me.

Can you add a test case for the ``ignore_user`` option to ``bodhi/tests/functional/test_comments.py``?

pypingou:
Done :)

ralphbean:
Looks good to me!

PR #690, opened 2015-10-29
Add perl# schema for bug links

remicollet:
Perl change often use RT#123456 (e.g. http://cpansearch.perl.org/src/SULLR/Net-SSLGlue-1.055/Changes) but perhaps "RT" is too generic, reason why I use "perl" here.


remicollet:
Perhaps cpan@ could be better

bochecha:
"perl" seems better to me as a prefix than "cpan", because, who knows, maybe one day the perl community will move off of CPAN, and then the "cpan" prefix wouldn't make sense any more?

(granted, this is not a very realistic scenario :stuck_out_tongue_winking_eye: )

Worst case, you could add both "perl" and "cpan" as prefixes, pointing to the same URL (see how I had added "rhbz", "rh" and "fedora").

In any case, :+1: to merge this.

lmacken:
Looks good to me. Thanks @remicollet and @bochecha

Issue #626, opened 2015-10-02
Bodhi emails questions/improvements


nirik:
First, is it still expected/needed that bodhi sends emails? Shouldn't FMN cover it?

If it does, the emails sent could/should be improved: 

- They have a From: of the person who made the comment  at feodraproject.org, sometimes this user doesn't have an alias or doesn't exist... like anonymous, or taskotron, etc. Could this just be 'updates@fedoraproject.org' or bodhi@fedoraproject.org or something?

- They don't have any headers that make it easy to filter on, they just have some X-Bodhi ones, but they all vary. Could we add a X-Bodhi: True or other header people could use to match all of them?

- Is there any way to opt out of them entirely? (If they aren't in FMN, bodhi needs a pref)?

ralphbean:
There was some reason why we decided to keep email sending in bodhi to beginwith.. but I forget what it was.

Is there any reason we can't remove them all and just rely on FMN?  That was the plan at one point.

morxa:
I get some bodhi emails also from FMN, namely emails about my own comments (e.g. '<myownusername> has commented on bodhi update <update>'). These are basically duplicates, as I get a similar email directly from bodhi. It would be really nice if this was more consistent.

I guess adding `X-Fedmsg-Category: bodhi` to the header is not a good idea because this is reserved for FMN? This would at least allow us to filter both FMN and non-FMN messages at the same time.

kparal:
I just received this email:
```
Subject: Undelivered Mail Returned to Sender

This is the mail system at host bastion01.phx2.fedoraproject.org.

I'm sorry to have to inform you that your message could not
be delivered to one or more recipients. It's attached below.

For further assistance, please send mail to postmaster.

If you do so, please include this problem report. You can
delete your own text from the attached returned message.

                   The mail system

<chrissandler@outlook.com>: host mx1.hotmail.com[207.46.8.199] said: 550
    Requested action not taken: mailbox unavailable (in reply to RCPT TO
    command)
```

Appended is the usual Bodhi comment email, which **I** am supposedly sending to chrissandler:
```
--2E7BC605DFCA.1444205035/bastion01.phx2.fedoraproject.org
Content-Description: Undelivered Message
Content-Type: message/rfc822

Return-Path: <kparal@fedoraproject.org>
Received: from bodhi04.phx2.fedoraproject.org (bodhi04.phx2.fedoraproject.org [10.5.126.116])
[...]

PR #691, opened 2015-10-29
Couple of fixes regarding sending emails.

pypingou:
This pull-request addresses the first two points mentioned by @nirik in 
https://github.com/fedora-infra/bodhi/issues/626

- Emails are all sent using the sender email set in the configuration file
- All emails sent by Bodhi have a X-Bodhi header


lmacken:
Yep, looks good. Thanks! :+1:

Issue #630, opened 2015-10-03
Screwy validator messages


ralphbean:
I just tried to submit an update for a nvr that never got built.  It failed (correctly) but I got these error messages back:

- Invalid tag: fedmsg-0.16.1-4.fc21 tagged with ``[u'f23-updates-candidate', u'epel7-testing-candidate', u'dist-6E-epel-testing-candidate', u'dist-5E-epel-testing-candidate', u'f22-updates-candidate', u'f21-updates-candidate']``
- Cannot find release associated with build: fedmsg-0.16.1-4.fc21, tags: ``[]``

The first one seems backwards.  It seems to be saying that that nvr **is** tagged with those tags, but the case is actually the opposite.  The nvr **is not** tagged with any of those.  We just need to change the wording.

The second one is also screwy.  Bodhi cannot find a release associated with that build that has "no" tags?  We should just say that we couldn't find any tags associated with that nvr.  That's more clear.

These should be pretty simple to add into ``bodhi/validators.py``.

farhaanbukhsh:
hey @ralphbean  I would like to fix it , how can i reproduce the bug.

pypingou:
@farhaanbukhsh check the first line:
> I just tried to submit an update for a nvr that never got built.

TridevGuha:
There is lots of nvr in  https://github.com/fedora-infra/bodhi/blob/develop/bodhi/validators.py  getting confused on which line to fix. Need some help

ralphbean:
Try grepping for some parts of the error messages.  "Invalid tag" and "Cannot find release".

TridevGuha:
Hey @ralphbean I updated the file bodhi/validators.py for Cannot find release associated with build: fedmsg-0.16.1-4.fc21, tags: [] 
but I am not sure about the Invalid tag. 
What does 'We just need to change the wording' means in the first comment?  Thanks :)

rahulbajaj0509:
hey, @ralphbean i would like to fix this bug.

mandaltapesh:
@rahulbajaj0509 @farhaanbukhsh are you guys going to fix this ? because i am interested in fixing this. just confirming in order to prevent duplicate effort :)

PR #703, opened 2015-11-14
Update validators.py

TridevGuha:
https://github.com/fedora-infra/bodhi/issues/630

### DIFF
--- a/bodhi/validators.py
+++ b/bodhi/validators.py
@@ -148,7 +148,7 @@ def validate_build_tags(request):
                 if not build_rel:
                     raise KeyError("Couldn't find release from build tags")
             except KeyError:
-                msg = 'Cannot find any tags associated with that nvr: {}, tags: {}'.format(build, tags)
+                msg = 'nvr for {} is not tagged with any of the following: {}'.format(build, tags)
                 log.warn(msg)
                 request.errors.add('body', 'builds', msg)
                 return
--- a/bodhi/validators.py
+++ b/bodhi/validators.py
@@ -148,7 +148,7 @@ def validate_build_tags(request):
                 if not build_rel:
                     raise KeyError("Couldn't find release from build tags")
             except KeyError:
-                msg = 'Cannot find release associated with build: {}, tags: {}'.format(build, tags)
+                msg = 'Cannot find any tags associated with that nvr: {}, tags: {}'.format(build, tags)
                 log.warn(msg)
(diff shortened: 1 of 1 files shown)


bodhi/validators.py
```diff
@@ -148,7 +148,7 @@ def validate_build_tags(request):
                 if not build_rel:
                     raise KeyError("Couldn't find release from build tags")
             except KeyError:
-                msg = 'Cannot find release associated with build: {}, tags: {}'.format(build, tags)
+                msg = 'Cannot find any tags associated with that nvr: {}, tags: {}'.format(build, tags)
```
ashcrow: It would be nice to have the error messages be different. Maybe this could read something like 
```python
'nvr for {} is not tagged with any of the following: {}'.format(build, tags)
```

lmacken: `tags` in this context are what the `build` is tagged with, so that message wouldn't be correct.

Also, I'm not sure that `nvr` is the most user-friendly way to describe a `build`, even though most people probably know what it means.

ashcrow: Fair enough. The same sentiment still remains though. I think having different error messages would be nicer than a single error message for both lines.

TridevGuha: different error massages are updated to issue #630

ralphbean:
OK.  There's really two failure scenarios here.

- The build is not tagged with *any* tags.
- The build is tagged, but not with a tag that can tell us the release.

Seems like we should add a check before the call to ``build_rel = Release.from_tags(tags, request.db)`` to handle the first scenario, like:

```python
if not tags:
    # return some error here indicating that the build has *no* tags
```

and then the error messages that have already been edited here need to be updated to indicate that we have hit the second scenario.

FWIW, I agree with the comment on an earlier commit that we should replace the word 'nvr' in the error messages with 'build'.

ralphbean:
Also, @TridevGuha, did you run the test suite against this change?  I have a feeling that there are test cases that check these error messages (although I'm not sure about that).

TridevGuha:
Nope, I did not run the test suite, need to run the test?

ralphbean:
Yeah, definitely.  They're pretty comprehensive.

ralphbean:
@TridevGuha any luck running the test suite?

ralphbean:
OK, I'm going to close this one for now.  Feel free to re-open it after we can get some feedback from the test suite.

PR #719, opened 2015-12-01
client - handle case where "total" not in the response.

ralphbean:
This happens when you use the bodhi2 client to create a multi-release update (which I just tried to do for the first time).


lmacken:
lgtm, :+1:

Issue #718, opened 2015-11-30
Missing attributes in Fedora 22 Updates repodata


jiridostal:
## Description
*Copied from Fedora-Infrastructure ticket. (https://fedorahosted.org/fedora-infrastructure/ticket/5003)*

I'm creating this ticket in order to resolve our customer's problem. We are working on Spacewalk\Red Hat Satellite. One of our customers found a bug that blocks use of feature in Spacewalk\Satellite for synchronizing repositories. The problem he bumped into was occurring only by Fedora 22 Updates repo while syncing erratas.

I did some investigation and found out that there are some inconsistencies in repodata/updateinfo.xml as described below.
### Bug analysis

Fedora 22 Updates repodata: updateinfo.xml Examples:

#### Actual result
```
<update from="updates@fedoraproject.org" status="stable" type="bugfix" version="2.0">
    <id>FEDORA-2015-12791</id>
    <title>tint2-0.12.2-1.fc22</title>
    <updated date="2015-08-11 13:04:12"/>
    <rights>Copyright (C) 2015 Red Hat, Inc. and others.</rights>
    <release>Fedora 22</release>
    <summary>tint2-0.12.2-1.fc22 bugfix update</summary>
    ...
```
#### Expected result
```
<update from="updates@fedoraproject.org" status="stable" type="bugfix" version="2.0">
    <id>FEDORA-2015-12791</id>
    <title>tint2-0.12.2-1.fc22</title>
    <issued date="some_date"/>
    <updated date="2015-08-11 13:04:12"/>
    <rights>Copyright (C) 2015 Red Hat, Inc. and others.</rights>
    <release>Fedora 22</release>
    <summary>tint2-0.12.2-1.fc22 bugfix update</summary>
    ...
```
In this example there is no "issued date" attribute. We need this attribute for post-processing repository. "Issued date" is critical for us to store informations about erratas.
### Fix recommendation

Add "issued date" attribute to shown example. There might be more records that are missing "issued date".

lmacken:
Interesting... so it looks like there are 5 stable updates across F21-23 & EL5-7 that are missing a `datee_pushed` field, which is set as the `issued` timestamp in the updateinfo. From what I can tell all of these updates were submitted during the initial bodhi2 launch, and are most likely related to some early database migration issue.

I just fixed all occurrences of the missing timestamp in the database, and will research the issue further, although it might already be resolved with the latest code.

lmacken:
[...]

PR #721, opened 2015-12-02
models: Don't reset the `pushed` and `date_pushed` flags upon request.

lmacken:
If a testing update gets a request for stable, currently it will nullify the
`pushed` and `date_pushed` fields. I don't see any reason why we shouldn't keep
these intact, especially considering the fact that we now have `date_stable` and
`date_testing` timestamps. Also, with `date_pushed` set to None, the
`<issued/>` tag in the updateinfo.xml will be omitted, which can lead to other
problems (see
issue #718).

### DIFF
--- a/bodhi/metadata.py
+++ b/bodhi/metadata.py
@@ -206,6 +206,7 @@ class ExtendedMetadata(object):
         rec.rights = config.get('updateinfo_rights')
 
         if update.date_pushed:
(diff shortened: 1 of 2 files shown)


ralphbean:
Sure, :+1:

lmacken:
Thanks!

PR #722, opened 2015-12-02
Tag multiple builds for the same package synchronously.

lmacken:
Performing all of the tagging in a single koji multicall does not guarantee
order, which means the `koji latest-pkg` could potentially be incorrect if two
versions of the same package went out in the same push.

This patch tags all builds for the same package without using koji multicalls.


ralphbean:
:+1: here!

lmacken:
Thanks :100:

PR #731, opened 2015-12-09
overrides: Gracefully handle duplicate submissions (fixes #713)

### LINKED ISSUE #713: No indication when override already exists
When trying to create a buildroot override, there is no warning when one already exists. It just silently fails and the Submit button turns green again. It would be nice if an indication of an already existing override was given, or even better would be to extend the expiration of the existing override to the given date.

lmacken:


ralphbean:
Nice!  :+1:

lmacken:
Thanks!

PR #737, opened 2015-12-10
email: Send comment notifications to the update submitter

### LINKED ISSUE #726: Bodhi doesn't send emails to the update submitter
When I submit an update, I never get any emails about the update afterwards. I need to comment on the update to start getting email from bodhi. Please make it so that update submitters get those emails as well.

Recent example: https://bodhi.fedoraproject.org/updates/FEDORA-2015-48362b03d1 , haven't gotten a single email about it.

lmacken:
fixes #726

### DIFF
--- a/bodhi/models/models.py
+++ b/bodhi/models/models.py
@@ -1565,7 +1565,7 @@ class Update(Base):
         Return a list of User objects that have commit access to all of the
         packages that are contained within this update.
         """
-        people = set()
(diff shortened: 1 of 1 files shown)


ralphbean:
Cool.  +1.  At some point we should double-back on removing email from bodhi all-together... but this works for now.

lmacken:
Yep, agreed.

Issue #599, opened 2015-09-24
auto-obsoleting doesn't work when both are pending


ralphbean:
I just submitted one update, it had a bug in it, and so I submitted another one right afterwards.  Normally, the second update would obsolete the first, but since the first had never made it to testing (I'm guessing), it didn't get auto-obsoleted.

Now they're both sitting there.  One will move forwards and the other will be stuck in testing forever:

https://bodhi.fedoraproject.org/updates/?packages=fedmsg

lmacken:
Fixed by https://github.com/fedora-infra/bodhi/pull/738

PR #738, opened 2015-12-11
models: Obsolete unlocked updates, even if they have an open request …

### LINKED ISSUE #724: When a package is update but change the bug type from bugfix to security, bodhi does not obsolete the previous one
Hello
Something wrong it is happening look at https://bodhi.fedoraproject.org/updates/?packages=p7zip

p7zip-15.09-3.fc23 now says that is security when it isn't .
p7zip-15.09-4.fc23 doesn't obsolete p7zip-15.09-3.fc23 

Anyway let's wait for pushes , to see if situation gets better .

Thanks.

lmacken:
…(fixes #724)

Prior to bodhi2, we would avoid automatically obsoleting updates if they had an
open request. Now that we have proper update locking in place, we can simply
avoid locked updates and potentially obsolete ones with an active request that
have yet to be pushed.

### DIFF
--- a/bodhi/models/models.py
+++ b/bodhi/models/models.py
@@ -880,7 +880,7 @@ class Update(Base):
             for oldBuild in db.query(Build).join(Update).filter(
                 and_(Build.nvr != build.nvr,
                      Build.package == build.package,
-                     Update.request == None,
+                     Update.locked == False,
                      Update.release == self.release,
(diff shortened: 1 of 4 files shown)


ralphbean:
Awesome!  This also fixes #599.

ralphbean:
:+1:

lmacken:
thanks :)

PR #739, opened 2015-12-11
ini: Link to wiki page for how to get updates instead of inline cmd

cgwalters:
Fedora Atomic host uses the `atomic` command and not `dnf` - the wiki
page has more information in either case, at the small cost of
indirection.


lmacken:
Solid. Thanks Colin!

Issue #366, opened 2015-08-25
Improper unicode handling of bugzilla titles


lmacken:
```
Traceback (most recent call last):                                                                           
  File "/usr/lib/python2.7/site-packages/pyramid_mako/__init__.py", line 148, in __call__                    
    result = template.render_unicode(**system)                                                               
  File "/usr/lib/python2.7/site-packages/mako/template.py", line 452, in render_unicode                      
    as_unicode=True)                                                                                         
  File "/usr/lib/python2.7/site-packages/mako/runtime.py", line 783, in _render                              
    **_kwargs_for_callable(callable_, data))                                                                 
  File "/usr/lib/python2.7/site-packages/mako/runtime.py", line 815, in _render_context                      
    _exec_template(inherit, lclcontext, args=args, kwargs=kwargs)                                            
  File "/usr/lib/python2.7/site-packages/mako/runtime.py", line 841, in _exec_template                       
    callable_(context, *args, **kwargs)                                                                      
  File "/usr/lib/python2.7/site-packages/bodhi/templates/master.html", line 141, in render_body              
    ${self.body()}                                                                                           
  File "/usr/lib/python2.7/site-packages/bodhi/templates/update.html", line 522, in render_body              
    <td>${self.util.bug_link(bug) | n}</td>                                                                  
  File "/usr/lib/python2.7/site-packages/bodhi/util.py", line 475, in bug_link                               
    link = link + " " + str(bug.title)                                                                       
UnicodeEncodeError: 'ascii' codec can't encode character u'\\xe9' in position 81: ordinal not in range(128)  

```

PR #742, opened 2015-12-16
Remove unnecessary encoding of bug titles.

### LINKED ISSUE #705: UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 95: ordinal not in range(128)
If I click the following 3 links for the ibus-typing-booster-1.2.15-1
updates on

https://bodhi.fedoraproject.org/users/mfabian

ibus-typing-booster-1.2.15-1.fc23
https://bodhi.fedoraproject.org/updates/FEDORA-2015-c97e8a4bd5

ibus-typing-booster-1.2.15-1.fc22
https://bodhi.fedoraproject.org/updates/FEDORA-2015-ccb91a63a8

ibus-typing-booster-1.2.15-1.fc21
https://bodhi.fedoraproject.org/updates/FEDORA-2015-2c65f7392a

I get a Python backtrace:

500 Server Error

Traceback (most recent call last): File "/usr/lib/python2.7/site-packages/pyramid_mako/__init__.py", line 148, in __call__ result = template.render_unicode(**system) File "/usr/lib/python2.7/site-packages/mako/template.py", line 452, in render_unicode as_unicode=True) File "/usr/lib/python2.7/site-packages/mako/runtime.py", line 784, in _render return context._pop_buffer().getvalue() File "/usr/lib/python2.7/site-packages/mako/util.py", line 150, in getvalue return self.delim.join(self.data) UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 95: ordinal not in range(128)

ralphbean:
Ok, so this fixes a bug that showed up first as #366, was partway fixed, but
then it showed up again as #705.

We were originally casting bug titles to a ``str`` (to ``bytes`` in python3
terminology) which caused (of course) some encoding errors.  We then addressed
this in 06476bf1 by using's kitchen's ``to_bytes`` which still forces things to
bytes, but does so in a graceful way.

The problem is that we're using mako's ``render_unicode(...)`` function at the
end of the day, which looks for things like bytes and has to force them *back*
to unicode before rendering everything out and giving it a final encoding once
done.  It's that re-decoding that caused the issue in #705.  By just keeping
everything as unicode and letting mako deal with it, this fixes #705.  (I
checked, and ``bug.title`` is indeed a ``unicode`` object).

I hacked this in place in staging to figure it out.  Compare:

- Working in stg - https://bodhi.stg.fedoraproject.org/updates/FEDORA-2015-67b99c3223
- Broken in prod - https://bodhi.fedoraproject.org/updates/FEDORA-2015-67b99c3223

### DIFF
--- a/bodhi/tests/models/test_models.py
+++ b/bodhi/tests/models/test_models.py
@@ -386,7 +386,7 @@ class TestUpdate(ModelTest):
         bug.title = u'foo\xe9bar'
         from bodhi.util import bug_link
         link = bug_link(None, bug)
-        eq_(link, "<a target='_blank' href='https://bugzilla.redhat.com/show_bug.cgi?id=1'>#1</a> foo\xc3\xa9bar")
+        eq_(link, u"<a target='_blank' href='https://bugzilla.redhat.com/show_bug.cgi?id=1'>#1</a> foo\xe9bar")
 
     def test_set_request_untested_stable(self):
         """
--- a/bodhi/util.py
+++ b/bodhi/util.py
@@ -41,7 +41,6 @@ from sqlalchemy.orm import scoped_session, sessionmaker
 from zope.sqlalchemy import ZopeTransactionExtension
 from pyramid.i18n import TranslationStringFactory
 from pyramid.settings import asbool
-from kitchen.text.converters import to_bytes
 
 from . import log, buildsys
 from .exceptions import RepodataException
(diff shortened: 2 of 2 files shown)


pypingou:
:+1: for me

lmacken:
Solid, thanks!

PR #753, opened 2015-12-30
readme modified to avoid error:database does not exist for local instance

trishnaguha:
I was getting "bodhi2 database does not exist" while setting up local instance. These changes fix the error.


pypingou:
:+1: for me

Issue #529, opened 2015-09-08
Gracefully handle Bugzilla deadlocks


lmacken:
```
[2015-09-08 19:29:18][     bodhi   ERROR] Unable to add comment to bug #1260329
Traceback (most recent call last):
File "/usr/lib/python2.7/site-packages/bodhi/bugs.py", line 77, in comment
bug.addcomment(comment)
File "/usr/lib/python2.7/site-packages/bugzilla/bug.py", line 295, in addcomment
return self.bugzilla.update_bugs(self.bug_id, vals)
File "/usr/lib/python2.7/site-packages/bugzilla/base.py", line 1270, in update_bugs
return self._proxy.Bug.update(tmp)
File "/usr/lib64/python2.7/xmlrpclib.py", line 1224, in __call__
return self.__send(self.__name, args)
File "/usr/lib/python2.7/site-packages/bugzilla/base.py", line 168, in _ServerProxy__request
ret = ServerProxy._ServerProxy__request(self, methodname, params)
File "/usr/lib64/python2.7/xmlrpclib.py", line 1578, in __request
verbose=self.__verbose
File "/usr/lib/python2.7/site-packages/bugzilla/base.py", line 263, in request
return self._request_helper(url, request_body)
File "/usr/lib/python2.7/site-packages/bugzilla/base.py", line 245, in _request_helper
raise sys.exc_info()[1]
Fault: <Fault -32000: 'DBD::mysql::db do failed: Deadlock found when trying to get lock; try restarting transaction [for Statement "UPDATE bugs SET delta_ts = ? WHERE bug_id = ?"] at /var/www/html/bugzilla/Bugzilla/Bug.pm line 1305\
n\tBugzilla::Bug::update(\'Bugzilla::Bug=HASH(0x7f90972f8710)\') called at /var/www/html/bugzilla/Bugzilla/WebService/Bug.pm line 768\n\tBugzilla::WebService::Bug::update(\'Bugzilla::WebService::Bug\', \'HASH(0x7f9096fe39f0)\') call
ed at /usr/share/perl5/vendor_perl/SOAP/Lite.pm line 2875\n\teval {...} called at /usr/share/perl5/vendor_perl/SOAP/Lite.pm line 2860\n\teval {...} called at /usr/share/perl5/vendor_perl/SOAP/Lite.pm line 2826\n\tSOAP::Server::handl
[...]

PR #756, opened 2016-01-12
Handling Bugzilla deadlock

trishnaguha:
Trying to fix https://github.com/fedora-infra/bodhi/issues/529


lmacken:
Looks good to me, that should hopefully do the trick. Thanks @trishnaguha! Nice job :)

PR #757, opened 2016-01-12
templates: Enable fedmenu on single-package update lists (fixes #755)

### LINKED ISSUE #755: add a link to https://apps.fedoraproject.org/packages/*
Hi there,

almost every time a see a package that I don't know about, I have to go to https://apps.fedoraproject.org/packages/ to understand if i'm able to test or not.

Can you please add one link in the main pages/list ? 

Example : 
global package page : https://bodhi.fedoraproject.org/updates/?packages=kanatest
detailed build page : https://bodhi.fedoraproject.org/updates/kanatest-0.4.8-16.fc23

lmacken:


pypingou:
:+1: for me (but not tested)

bodhi/services/updates.py
```diff
@@ -213,6 +213,10 @@ def query_updates(request):
         query = query.join(Update.builds).join(Build.package)
         query = query.filter(or_(*[Package.name==pkg for pkg in packages]))
 
+    package = None
+    if packages and len(packages):
```
pypingou: Isn't this a little redundant since if ``packages`` is a ``[]`` ``if packages`` will return ``False``?

lmacken: That's what I had initially, but  `packages` can be `None`, which explodes.

lmacken:
Thanks!

bodhi/services/updates.py
pypingou: Ah, yes of course! I should have thought of that :)

PR #771, opened 2016-01-29
push: Only attempt to push unlocked updates

lmacken:
We just hit an issue where a second push was queued up containing locked updates from the first. This lead to a bunch of ejection messages.


ralphbean:
:+1:

lmacken:
Thanks :dango:

PR #779, opened 2016-02-18
Time to stable

### LINKED ISSUE #745: [RFE] Display time to stable
It would be great if, in addition to displaying time at which an update was submitted to testing, the time remaining until the update can be pushed to stable were displayed.  Thanks!

trishnaguha:
#745

### DIFF
--- a/bodhi/models/models.py
+++ b/bodhi/models/models.py
@@ -1757,6 +1757,16 @@ class Update(Base):
         return False
 
     @property
+    def days_to_stable(self):
+        """ Return the number of days until an update can be pushed to stable """
+        for comment in self.comments:
+            if comment.user.name == 'bodhi' and \
+                'can be pushed to stable now if the ' \
+                'maintainer wishes' in comment.text:
+                return (comment.timestamp - self.date_submitted).days
+        return 0
+
+    @property
     def days_in_testing(self):
         """ Return the number of days that this update has been in testing """
         if self.date_testing:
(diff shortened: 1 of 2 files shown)


ralphbean:
I think this is backwards.

- The ticket requested the number of days *until* and update *can* be pushed to stable.
- This implements the number of days *between* when an update was submitted and when it *was* pushed to stable.

So, that will have to be sorted out first.

Second, once a function like this is implemented that returns a value like is requested -- then we'll need to update ``bodhi/templates/update.html`` to *display that value* to users.

trishnaguha:
Here is a screenshot.
screenshot from 2016-02-26 20-53-23

lmacken:
Looks good to me! :dancers:

ralphbean:
I dunno if this is correct still.  If you look at the implementation of ``days_to_stable``, I think its still backwards.  It returns the "number of days between when bodhi said it *could* be pushed to stable, and when it was actually pushed to stable"  which is not the same thing as the "number of days between now and when it *can* be pushed to stable".

lmacken:
Yeah, I think you're right @ralphbean 

Like I mentioned on IRC, 
```
09:37  lmacken> trishnag: I think you can do something like: return self.release.mandatory_days_in_testing - (datetime.utcnow() - self.date_testing).days

```

trishnaguha:
Here is the Modified PR #793

PR #782, opened 2016-02-22
Hide "Push to Stable" if request is already "stable".

ralphbean:
Whenever I request an update go to stable, it then reloads the page for the
update.. but the button is still there.  It should be hidden after you click
it.  This should handle that.


lmacken:
LGTM, thanks! :100:

PR #786, opened 2016-02-26
Add an error message when we fail to query datagrepper on the user page.

ralphbean:


lmacken:
Looks good to me. :+1:

PR #791, opened 2016-03-01
models: Add a unique constraint to the Update.title

### LINKED ISSUE #775: Cannot request stable - 'failing to determine ACLs'
For this update:  https://bodhi.fedoraproject.org/updates/FEDORA-EPEL-2016-50448945cd

Something is screwy when I try to request stable.  Not sure what.  Perhaps we should sync prod data to stg and then try adding some debugging around it over there.

lmacken:
This fixes #775, which is a race-condition that allows multiple updates for the
same build to get created seconds apart. By making the title unique, this would
fail when trying to commit the second update.


ralphbean:
Nice.  :+1:

lmacken:
Thanks! :fried_shrimp:

Issue #787, opened 2016-02-29
Can not push to stable


ellert:
Hi!

I am trying to push this EPEL 5 update to stable:

https://bodhi.fedoraproject.org/updates/FEDORA-EPEL-2016-910bd4684d

This should be possible since it is more that 14 days in testing:

In Testing 	14 days ago, 2016-02-15 08:50:38.668140

However, when I try I get the following error:

Cannot submit xrootd-4.2.3-5.el5 to stable since it is older than xrootd-20100315-2.el5

This does not make sense for at least two reasons.

1. The current version in EPEL 5 is xrootd-4.2.3-3.el5, not xrootd-20100315-2.el5.
2. xrootd-4.2.3-5.el5 has epoch 1, while xrootd-20100315-2.el5 had no epoch, so it is not older.

The xrootd-20100315-2.el5 was the very first build of the package for EPEL 5 and is now in state deleted in koji: https://koji.fedoraproject.org/koji/buildinfo?buildID=178885

ellert:
What makes this really strange is that the submissions to stable of the corresponding update for EPEL 6 two hours earlier worked just fine. For EPEL 6 there also exists in koji an xrootd-20100315-2.el6 build in deleted state, and it did not prevent the submission. Did something change in these two hours?

lmacken:
I manually queued this update up for stable. We'll need some additional code changes to support epochs in the request validator.

ellert:
This is still broken for EPEL 5.
Can you please push https://bodhi.fedoraproject.org/updates/FEDORA-EPEL-2016-c13a4dbfd9

PR #792, opened 2016-03-02
Use the RPM epoch when comparing builds

lmacken:
This resolves bugs like https://fedorahosted.org/bodhi/ticket/793 and https://github.com/fedora-infra/bodhi/issues/787


lmacken:
This currently has commits from https://github.com/fedora-infra/bodhi/pull/791 in it, in order to get the appropriate alembic migration path.

ralphbean:
LGTM.  :+1:

lmacken:
Thanks! :aerial_tramway:

PR #793, opened 2016-03-02
Modifying days to stable

trishnaguha:
Modifying #779

### DIFF
--- a/bodhi/models/models.py
+++ b/bodhi/models/models.py
@@ -1772,12 +1772,10 @@ class Update(Base):
     @property
     def days_to_stable(self):
         """ Return the number of days until an update can be pushed to stable """
-        for comment in self.comments:
-            if comment.user.name == 'bodhi' and \
-                'can be pushed to stable now if the ' \
-                'maintainer wishes' in comment.text:
-                return (comment.timestamp - self.date_submitted).days
-        return 0
+        if self.meets_testing_requirements:
+            return self.release.mandatory_days_in_testing - (datetime.utcnow() - self.date_testing).days
+        else:
+            return 0
 
     @property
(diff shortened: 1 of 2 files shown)


bodhi/models/models.py
```diff
@@ -1763,7 +1763,7 @@ def days_to_stable(self):
             if comment.user.name == 'bodhi' and \
                 'can be pushed to stable now if the ' \
                 'maintainer wishes' in comment.text:
-                return (comment.timestamp - self.date_submitted).days
+                return (datetime.utcnow() - self.date_testing).days - self.release.mandatory_days_in_testing
```
trishnaguha: @lmacken @ralphbean Since the comment gets added just after `self.release.mandatory_days_in_testing` and we are considering `datetime.utcnow()` also, I have made  `(datetime.utcnow() - self.date_testing).days - self.release.mandatory_days_in_testing`. Otherwise it would return negative value.

lmacken:
So from what I understand we want to the number of days until the 'this update can now be pushed to stable' comment gets added automatically, right?

trishnaguha:
@lmacken So it seems that we don't have to check for the comment `"can be pushed to stable now if the maintainer wishes"`, since it gets added automatically after an update meets testing requirements right?

lmacken:
Yeah, that's right @trishnaguha

trishnaguha:
@lmacken Yes it is about subtracting days in testing from `self.release.mandatory_days_in_testing` to get the number of days left. I did that because I was getting negative value. Now I understand why I was getting that. I have modified the PR.  Thanks :)

lmacken:
Cool, looks good to me :) Thanks @trishnaguha!

PR #800, opened 2016-03-09
Display timezone in timestamps

### LINKED ISSUE #789: show timezone in timestamps
When I look at some update ([example](https://bodhi.fedoraproject.org/updates/FEDORA-2016-5ba698daad)), I sometimes really need to know which timezone all the timestamps are in. When Bodhi says this update was pushed stable today at 22:24, is that UTC, or my local time, or the Bodhi server local time?

Please add a timezone notifier behind the timestamp to make it clear, thank you.

trishnaguha:
Fixes #789

### DIFF
--- a/bodhi/templates/fragments.html
+++ b/bodhi/templates/fragments.html
@@ -24,7 +24,7 @@
         <span class="pull-right">
           <a href="${comment.url()}">
             %if not display_update:
-            ${util.age(comment.timestamp)}, <small>${comment.timestamp}</small>
+            ${util.age(comment.timestamp)}, <small>${comment.timestamp} (UTC)</small>
             % else:
             ${util.age(comment.timestamp)}
             % endif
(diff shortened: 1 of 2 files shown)


ralphbean:
@trishnaguha, since this is a UI change, can you attach a screenshot of the affected page in a comment here?  It just makes it easier to review.  :)

trishnaguha:
@ralphbean Here are the screenshots :)

screenshot from 2016-03-09 20-33-46

screenshot from 2016-03-09 20-34-03

ralphbean:
Good stuff!  :+1: here!

kparal:
When I saw those screenshots, I wonder - is the microsecond precision intentional? It seems a bit too precise for human consumption :)

trishnaguha:
@kparal 
So @ralphbean or @lmacken can answer that :)

ralphbean:
> is the microsecond precision intentional?

Nope!  We should get rid of it.  ;)

PR #808, opened 2016-03-16
Only touch tags that we know about.

lmacken:
This avoids situations where bodhi steps on the toes of releng when they are
working with side-tags for outside composes.


ralphbean:
LGTM!  :+1:

lmacken:
Thanks!

PR #814, opened 2016-03-17
Make sure request is None after unpushing

### LINKED ISSUE #813: Unpushing an update with a stable request doesn't remove the Stable request
I unpushed https://bodhi.fedoraproject.org/updates/FEDORA-2016-76e0d3a68e because it was the only button I could press to make sure the update didn't go to stable (why wasn't it automatically removed from stable request once it got -1 karma?), but now the status is:
Request	stable
Status	unpushed

I would have expected the request to be removed as well on unpush.

trishnaguha:
Fixes #813

### DIFF
--- a/bodhi/models/models.py
+++ b/bodhi/models/models.py
@@ -1547,6 +1547,7 @@ class Update(Base):
 
         self.pushed = False
         self.status = UpdateStatus.unpushed
+        self.request = None
 
     def revoke(self):
         """ Remove pending request for this update """
--- a/bodhi/tests/functional/test_updates.py
+++ b/bodhi/tests/functional/test_updates.py
@@ -1915,6 +1915,26 @@ class TestUpdatesService(bodhi.tests.functional.base.BaseWSGICase):
 
     @mock.patch(**mock_taskotron_results)
     @mock.patch(**mock_valid_requirements)
+    @mock.patch('bodhi.notifications.publish')
+    def test_request_after_unpush(self, publish, *args):
+        """Test request of this update after unpushing"""
+        args = self.get_update('bodhi-2.0.0-3.fc17')
+        resp = self.app.post_json('/updates/', args)
+        up = self.db.query(Update).filter_by(title=resp.json['title']).one()
+        up.status = UpdateStatus.testing
+        up.request = UpdateRequest.stable
+        self.db.flush()
+
+        resp = self.app.post_json(
+            '/updates/%s/request' % args['builds'],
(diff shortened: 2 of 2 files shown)


bodhi/models/models.py
```diff
@@ -1044,6 +1044,7 @@ def set_request(self, db, action, username):
         topic = u'update.request.%s' % action
         if action is UpdateRequest.unpush:
             self.unpush()
+            self.request = None
```
ralphbean: Would it make more sense to set the request to None *inside* the definition of the ``unpush`` method?  That way, we could be sure the request is correct whenever that method is called.

trishnaguha: @ralphbean Yeah right. Although this already ensures when `action` is `UpdateRequest.unpush` the `request` is set to `None`, but it would make more sense to set request to `None` inside the definition of `unpush`. I'll try to do that :).

ralphbean: I guess I am a little confused due to the two different ``unpush`` methods in play:

```bash
$ git grep unpush\(
bodhi/models/models.py:    def unpush(self, koji):
bodhi/models/models.py:                b.unpush(koji=request.koji)
bodhi/models/models.py:                up.unpush()
bodhi/models/models.py:            self.unpush()
bodhi/models/models.py:    def unpush(self):
bodhi/tests/models/test_models.py:        self.obj.builds[0].unpush(koji)
```

There's an ``unpush`` method on both the ``Build`` and on the ``Update`` class.. so I just want to make sure we have things straight and that all the cases are covered that should be.

I'll keep reading and will come back with another comment in a moment.

ralphbean: Yeah, there's a *second* place that ``Update.unpush(db)`` is called.

- First there is the place where you added a the ``self.request = None`` line on 1046.
- But there is *also* a place where it is called on line 836 (when an update is edited and the builds are changed).

I *think* (and @lmacken may correct me if I'm wrong here) that ``self.request`` should be set to ``None`` in *both* circumstances.  The best way to code that I think is to remove the ``self.request = None`` line where you added it and to *instead* add it to the end of the ``Update.unpush(db)`` method on line 1550.  That way, the request will be set to ``None`` **everywhere** in the code that something tries to unpush an Update (not just when the user explicitly requests that -- which is the code path that you edited).

Does that make sense?

ralphbean:
Looks good here.  :)

lmacken:
Nice job, thanks @trishnaguha! :palm_tree:

Issue #796, opened 2016-03-04
[RFE] manual stable push based on karma


carlwgeorge:
Bodhi has two scenarios for pushing updates to stable.

* manual stable push based on time
    1. auto-karma disabled
    2. 14 days in testing
    3. maintainer submits to stable via "push to stable" button
* automatic stable push based on karma
    1. auto-karma enabled
    2. karma threshold reached
    3. automatically submitted to stable

There is another scenario that seems to be allowed by policy, but not currently possible in bodhi.

* manual stable push based on karma
    1. auto-karma disabled
    2. karma threshold reached
    3. maintainer submits to stable via "push to stable" button

What would it take to have this feature added?

trishnaguha:
@lmacken @ralphbean I want to work on it. I need some hints. How do I start? :)

lmacken:
I haven't looked too deeply into what this change will require, but here is some info off the top of my head.

* Right now the "autokarma" checkbox enables the stable/unstable karma thresholds. The thresholds are currently ignored unless "autokarma" is enabled.

This proposes being able to keep "autokarma" disabled, but set thresholds that, when met, will allow the package to be manually pushed to stable.

I know the main update template logic will need to be modified. It's a bit hairy right now, so it might require some refactoring...
https://github.com/fedora-infra/bodhi/blob/develop/bodhi/templates/update.html#L515-L543

You'll also want to grep around for 'autokarma' to see if any of that logic needs modification, like:
https://github.com/fedora-infra/bodhi/blob/develop/bodhi/models/models.py#L709
https://github.com/fedora-infra/bodhi/blob/develop/bodhi/models/models.py#L865

trishnaguha:
@lmacken Thanks for the hints :)
So is there any way to see or test the changes I'll be making? It would be really helpful if I can know that.

lmacken:
Yeah, it shouldn't be too difficult to write a unit test for. It should basically:

 * Create an update with `autokarma=False` and `stable_karma` set to 1
 * Give the update 1 karma (as a differrent user). Example: https://github.com/fedora-infra/bodhi/blob/develop/bodhi/tests/functional/test_updates.py#L2363-L2381
 * Ensure the `request` didn't get automatically set due to autokarma, since it should be disabled.
 * Get the html of the page, and ensure the text 'Push to Stable' is in there. Example of getting the html: https://github.com/fedora-infra/bodhi/blob/develop/bodhi/tests/functional/test_updates.py#L422-L428

[...]

Issue #772, opened 2016-01-31
It should be possible to manually push an update with karma 1


kkofler:
Since the minimum value for auto-karma is 1, it would be only logical to be also allowed to manually push updates which have a karma of +1.

In Bodhi 1, this was at least possible with the following workaround:
* edit the update
* enable auto-karma
* set the stable threshold to 1
* confirm the edit
* now hit the "push to stable" button

and then it worked (whereas just hitting "push to stable" without the workaround would complain that the update does not satisfy the criteria).

Now in Bodhi 2, even this workaround does not work anymore. I see no way to trigger the push. (There is no "Push to stable" button shown, neither without nor with the workaround.) It is entirely illogical that we allow updates to go stable with a karma of +1 if it was set to do so automatically with no verification, but not if the packager verified that the feedback is OK. That punishes the packager for wanting to ensure the quality of his/her package, and ultimately hurts the user (because either they will have to wait longer for a verified update, or the packagers will just enable autokarma and then bad updates can end up pushed because no human looked at the feedback, especially at the threshold +1).

Please allow manual pushes at karma +1 for consistency with the minimum auto-karma threshold.

kkofler:
Case in point: https://bodhi.fedoraproject.org/updates/FEDORA-2016-afa165bd29

kkofler:
To elaborate on why I think this should be allowed:

Case A:
* The update is submitted.
* Auto-karma is enabled at the threshold +1.
* A tester gives 1 karma.
* The update is automatically pushed to stable.

Case B:
* The update is submitted.
* Auto-karma is not enabled.
* A tester gives 1 karma.
* The update has exactly the same amount of testing as in case A. Yet, Bodhi will not allow me to push this update to stable. Why?

Case C (the Bodhi 1 workaround, that also no longer works):
* The update is submitted.
* Auto-karma is not enabled.
* A tester gives 1 karma.
* The update is edited, auto-karma is now enabled at the threshold +1.
* The update is now in exactly the same state as in case A. Yet, Bodhi neither automatically pushes the update to stable, nor allows me to push it manually. Why?

bowlofeggs:
I believe that @trishnaguha fixed this in #821.

kkofler:
Bodhi 2.2.2 is now deployed and this still doesn't work. Please reopen this issue!

[...]

PR #821, opened 2016-03-29
Manually push to stable based on stable karma threshold

trishnaguha:
In progress PR for #796 and #772


bodhi/models/models.py
```diff
@@ -714,6 +723,12 @@ def new(cls, request, data):
                 'description': 'Auto-karma stable requests disabled.',
             })
 
+        elif data['autokarma']:
+        # We have to set autokarma column to True here.
```
trishnaguha: @lmacken These tests are currently failing. 

_https://github.com/fedora-infra/bodhi/blob/develop/bodhi/tests/test_masher.py#L977
https://github.com/fedora-infra/bodhi/blob/develop/bodhi/tests/models/test_models.py#L340
https://github.com/fedora-infra/bodhi/blob/develop/bodhi/tests/functional/test_updates.py#L308_

Since the `"autokarma"` column is `"False"` by default, these tests are failing because they want `"autokarma"` to be enabled or `"True"` so that `"request"` gets set to `"stable"` automatically.
So we have to set `"autokarma"` column `"True"` here like when `"autokarma"` is enabled.

bodhi/models/models.py
```diff
@@ -706,7 +707,15 @@ def new(cls, request, data):
                     build.package for build in data['builds']
                 ] if pkg.requirements], []))))
 
-        if not data['autokarma']:
+        if not data['autokarma'] and data['stable_karma'] == 1:
+            # We have to set autokarma column to False here.
```
trishnaguha: `autokarma` column needs to be set to `False` here since `autokarma` is disabled.

bodhi/models/models.py
```diff
@@ -706,7 +707,15 @@ def new(cls, request, data):
...
+            caveats.append({
+                'name': 'autokarma',
+                'description': 'Auto-karma stable requests disabled. Make stable karma 1.',
+            })
+
+        elif not data['autokarma'] and data['stable_karma'] != 1:
+            # We have to set autokarma column to False here.
```
trishnaguha: `autokarma` column needs to be set to `False` here since `autokarma` is disabled.

bodhi/models/models.py
```diff
@@ -1662,13 +1677,18 @@ def check_karma_thresholds(self, db, agent):
...
+                        self.set_request(db, UpdateRequest.stable, agent)
+                        self.request = UpdateRequest.stable
+                        self.date_pushed = None
+                        notifications.publish(
+                            topic='update.karma.threshold.reach',
+                            msg=dict(update=self, status='stable'))
+                    else:
+                        log.info("%s now reaches stable karma threshold and can be manually pushed to stable now" % self.title)
+                        self.request = None
+                        # We have to add the 'testing_approval_msg_based_on_karma' message as comment after this
```
trishnaguha: I will remove this line ["# We have to add the 'testing_approval_msg_based_on_karma' message as comment after this"](https://github.com/fedora-infra/bodhi/pull/821/files#diff-55c0d4c740a773e117d1f620f264f793R1667) from here as the approval_msg is done by scripts/approve_testing.
https://github.com/fedora-infra/bodhi/pull/821/files#diff-08739a77d556ed2e9809494ddb3f7286R2515 confirms that bodhi successfully posted the approval message when karma meets stable karma.

trishnaguha:
@lmacken This is the in progress PR. It is pretty buggy right now. I would like to know what do you think about the changes done. :)
Thanks!

bodhi/models/models.py
lmacken: The web interface has "Auto-request stable?" enabled by default, so I think we can have `autokarma` default to `True`.

bodhi/models/models.py
```diff
@@ -706,14 +707,28 @@ def new(cls, request, data):
...
                 'description': 'Auto-karma stable requests disabled.',
             })
 
+        elif data['autokarma']:
+        # We have to set autokarma column to True here.
+
+        else:
+            pass # For further enhancing
+
         del(data['autokarma'])
```
lmacken: Since `autokarma` is now going to be a database column and is already defined as a boolean in `schemas.py`, I think we can simply remove this `del`, along with all of the stable/unstable logic above it completely. That way, we're initializing the update with `**data` which includes the new `autokarma` field. I don't think the `caveats` above this are necessary anymore. I'm working on this locally and will have more feedback and commits shortly.

bodhi/tests/functional/test_updates.py
```diff
@@ -1950,6 +1950,48 @@ def test_invalid_stable_request(self, *args):
```
trishnaguha: @lmacken I was thinking do we really need two separate test cases to test the following? :
* Test request to stable before an update reaches stable karma and after it reaches stable karma when autokarma is disabled
* Test manually push to stable when autokarma is disabled and karma threshold is reached

I'm not sure merging these two testcases into one is a good idea. Or should we keep them as two separate test cases them like they are now.

bodhi/tests/functional/test_updates.py
```diff
@@ -1950,6 +1950,48 @@ def test_invalid_stable_request(self, *args):
 
     @mock.patch(**mock_taskotron_results)
     @mock.patch(**mock_valid_requirements)
+    def test_request_to_stable_based_on_stable_karma(self, *args):
```
trishnaguha, line 1953: @lmacken I was thinking do we really need two separate test cases to test the following? :
- Test request to stable before an update reaches stable karma and after it reaches stable karma when autokarma is disabled
- Test manually push to stable when autokarma is disabled and karma threshold is reached

I'm not sure merging these two testcases into one is a good idea. Or should we keep them as two separate test cases them like they are now.

bodhi/models/models.py
```diff
@@ -706,16 +707,6 @@ def new(cls, request, data):
...
 
-        if not data['autokarma']:
-            del(data['stable_karma'])
-            del(data['unstable_karma'])
-            caveats.append({
-                'name': 'autokarma',
-                'description': 'Auto-karma stable requests disabled.',
-            })
-
-        del(data['autokarma'])
```
trishnaguha, line 717: @lmacken So I have removed all these from here like https://paste.fedoraproject.org/368261/46363292/ as you told earlier and the current patch is passing the current bodhi test suite along with the testcases which are included in this patch \o/.
So what should be our next step? :)

bodhi/tests/functional/test_updates.py
lmacken, line 1953: Either is fine with me in this case, whichever is easiest.

bodhi/models/models.py
lmacken: I don't think you necessarily need to delete this part... It's there for when an update reaches the karma threshold mid-push, then immediately after the push is done it gets the comment. The `scripts/approve-testing` cron job doesn't run as often, but it will still catch these updates... not a big deal either way.

bodhi/tests/functional/test_updates.py
lmacken, line 1953: After reading through the tests, I think they'd be safe to merge if you want.

bodhi/models/models.py
lmacken, line 717: Everything looks good to me, nice work!! 

I think the next step will be for me to merge this PR, spin up a new release, and get it into staging for further testing.

trishnaguha:
@lmacken Cool so this is good to go \o/.
Thanks a lot for the help and continually reviewing :-).

bodhi/models/models.py
kkofler: Yet, in the deployed Bodhi 2.2.2, it still does not remember the karma values if autokarma is disabled. There must be some more code doing that.

This and the other issue I pointed out in this review make the feature from pull request #821 entirely non-functional.

bodhi/templates/update.html
```diff
@@ -535,6 +535,8 @@
             % endif
           % elif update.meets_testing_requirements and getattr(update.request, 'description', None) != 'stable':
             <a id='stable' class="btn btn-sm btn-success"><span class="glyphicon glyphicon-circle-arrow-right"></span> Push to Stable</a>
+          % elif update.stable_karma not in (0, None) and update.karma >= update.stable_karma and not update.autokarma and getattr(update.request, 'description', None) != 'stable':
```
kkofler, line 538: Why `and not update.autokarma`? This means that the manual push is still not possible if autokarma is enabled, but did not trigger for some reason (e.g., because it was enabled only after the karma value was reached). I don't see any valid reason for that restriction.

This and the other issue I pointed out in this review make the feature from pull request #821 entirely non-functional.

kkofler (commented):
Pull request #821 is entirely non-functional on the deployed Bodhi 2.2.2.

bodhi/templates/update.html
trishnaguha: @kkofler Because this is for `manually push to stable is based of stable karma threshold` here. We don't want `autokarma` to be enabled here, if that happens I am not sure whether there would be any difference between `Autopush` and this feature. 
There have been two ways before i.e, Manually push to stable based on time(14 days in testing) and Autopush based on stable karma threshold.

kkofler: The autopush only triggers when the karma goes from 0 to 1, not when it is already at 1, that's the difference.

kkofler: Please also see my comments on issue #772.

PR #823, opened 2016-04-01
Add an "agent" to all of the masher messages.

ralphbean:
The idea here is that when a mash fails, we'll be able to easily tell _whose_
mash failed.

Then, we can change the IRC message from:
- bodhi masher failed to mash f24-updates-testing

to:
- bodhi masher failed on kevin's mash of f24-updates-testing

The benefit being that the releng team member on `.push_duty` will get a ping
in their IRC client when things fail so they can have a look.

---

Test suite is all passing here.

### DIFF
--- a/bodhi/tests/test_masher.py
+++ b/bodhi/tests/test_masher.py
@@ -77,7 +77,7 @@ class FakeHub(object):
 
 def makemsg(body=None):
     if not body:
-        body = {'updates': ['bodhi-2.0-1.fc17']}
+        body = {'updates': ['bodhi-2.0-1.fc17'], 'agent': 'lmacken'}
     return {
         'topic': u'org.fedoraproject.dev.bodhi.masher.start',
         'body': {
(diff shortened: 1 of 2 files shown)


bodhi/tests/test_masher.py
```diff
@@ -1009,7 +1028,7 @@ def test_push_timestamps(self, publish, *args):
         publish.assert_called_with(
             topic="mashtask.complete",
             force=True,
-            msg=dict(success=True, repo='f17-updates-testing'))
+            msg=dict(success=True, repo='f17-updates-testing', agent='lmacken'))
```
pypingou, line 1031: This kinda surprises me a little since the MasherThread above seems to be created by `ralph`. Is this expected?

lmacken:
Hmm, so right now people run the `bodhi-push` tool as the `apache` user in order to access the fedmsg certs. I think this would make the agent always be `apache`?

ralphbean:
Looks like `bodhi-push` is able to figure out the name of the sudo-ing user:  https://apps.fedoraproject.org/datagrepper/raw?topic=org.fedoraproject.prod.bodhi.masher.start  The agent looks correct in all those recent messages.

bodhi/tests/test_masher.py
ralphbean, line 1031: I think that one (where the `MasherThread` is created with `'ralph'`) is in a different test.  This one uses `self.masher` which is set to be `'lmacken'`.

lmacken:
Sweet, sounds good then :)

ralphbean:
woot!  Thanks @lmacken :)

PR #825, opened 2016-04-06
So adamw doesn't kill me.

ralphbean:
His paragraphs tend to be rather long.

### DIFF
--- a/bodhi/static/css/site.css
+++ b/bodhi/static/css/site.css
@@ -209,6 +209,7 @@ ul.updateslist li {
 .markdown p {
     word-wrap:      break-word; /* old-school */
     overflow-wrap:  break-word; /* new-school */
+    padding:        5px;
 }
 
 #markdown-help strong {


AdamWill:
:gun: :+1:

lmacken:
Way too long... :trollface:

AdamWill:
@lmacken Look, that's not just unfair, it's demonstrably wrong. See, if you go back into the history of all adamw mails, starting in 1994, an interesting trend emerges. Basically, both overall email and paragraph length vary quite widely, but there's no significant difference from the average of other posters in the same context. (Of course, I'm omitting automated messages that just happen to be sent from adamw's email addresses here). There's an interesting tangent here about the differences in message and paragraph lengths between technical and non-technical lists, and even between different technical lists, but I won't go into that in detail here. Roughly, technical lists have longer messages but shorter paragraphs. This might be because of increased use of bottom quoting and quote interleaving (that's when people reply in between blocks of the quoted text), but I'm not totally sure, we'd have to look into it more. But yeah, bottom line, MY PARAGRAPHS AREN'T LONG YOU MEANIE

...goddamnit.

(I made all of that up, but it sure sounds convincing, right? right?!)

lmacken:
@AdamWill Your essay gets a gold star :star2: @ChangeTip

lmacken:
@AdamWill Your essay is doughut :doughnut: worthy. @ChangeTip

lmacken:
@AdamWill here's a dime for the dedication. @ChangeTip :money_with_wings:

changetip:
Hi @adamwill, @lmacken sent you a tip worth a dime (236 bits/$0.10), and I'm here to deliver it ➔ **[collect your tip](https://www.changetip.com/collect/944629)**.

**[Learn more about ChangeTip](https://www.changetip.com/tip-online/github)**

PR #832, opened 2016-04-23
Add timestamp in locked message for updates

### LINKED ISSUE #831: Inform since when the update is locked
Bodhi2 now shows lock messages:

> This update is currently locked and cannot be modified.

Please say since when the update is locked, so that it's easier to discover locking bugs (when we see it has been locked for 24 hours, we immediately know something is wrong).

E.g.

> This update has been locked since YYYY-MM-DD hh:mm:ss (UTC) and cannot be modified.

Thanks.

trishnaguha:
Fixes #831


sayanchowdhury:
Looks good to me :+1:

bodhi/models/models.py
```diff
@@ -1792,6 +1792,12 @@ def days_in_testing(self):
             return 0
 
     @property
+    def date_locked(self):
+        """ Return the date when the update has been locked """
+        if self.locked:
+            return datetime.utcnow()
```
pypingou: Won't this always return `utcnow()`?
So today, it won't it show in the template as:
`
This update is currently locked since <small>2016-05-03 07:59:51 (UTC)</small> and cannot be modified.
`
In one hour:
`
This update is currently locked since <small>2016-05-03 08:59:51 (UTC)</small> and cannot be modified.
`
And tomorrow:
`
This update is currently locked since <small>2016-05-04 07:59:51 (UTC)</small> and cannot be modified.
`
?

trishnaguha:
@pypingou Yes that's right!
What if we can do something like this https://paste.fedoraproject.org/363289/ ? 
I need some hint about its scope if this is good to go :).

<pre>
# Assigns the current date when update is locked
locked_date = False
if self.locked:
    date_now = datetime.utcnow()
    locked_date = True
def date_locked(self):
    if self.locked_date == True:
        return self.date_now
</pre>

pypingou:
Why not just saving the date in the DB itself as NULL by default and filling it at the time the update is updated to be marked as 'locked'? Then the date could be reset to NULL when the update is updated to be marked as 'unlocked' again.

trishnaguha:
@lmacken Can you please take a look at this one? :)
As @pypingou suggested on irc that you may have a better idea to implement this. 
Can you suggest your idea so that I could modify the patch?  :)

lmacken:
Yeah, I think we should probably add a new timestamp for the date locked to the Update model:

https://github.com/fedora-infra/bodhi/blob/develop/bodhi/models/models.py#L633-L639

Then in the masher when we lock/unlock updates, we should modify these timestamps.

https://github.com/fedora-infra/bodhi/blob/develop/bodhi/consumers/masher.py

trishnaguha:
@lmacken I have modified the PR.  
Do we need to write test for the same?
/cc @sayanchowdhury @pypingou

bodhi/templates/update.html
```diff
@@ -506,7 +506,7 @@
       <div class="alert alert-info" role="alert">
         <span class="glyphicon glyphicon-lock" aria-hidden="true"></span>
         <span class="sr-only">Locked:</span>
-        This update is currently locked and cannot be modified.
+        This update is currently locked since <small>${str(update.date_locked).split('.')[0]} (UTC)</small> and cannot be modified.
```
pypingou: Rather than using `str()` here, I would suggest using `.strftime()` https://docs.python.org/2/library/time.html#time.strftime

bodhi/consumers/masher.py
```diff
@@ -378,6 +379,9 @@ def work(self):
             self.check_all_karma_thresholds()
             self.obsolete_older_updates()
 
+            # Update datetime for locking on update
+            self.locked_date_for_update()
```
lmacken: So this will set the `locked_date` for each update _after_ the push is done. Don't we want to set this when it is initially locked, at the beginning of the `work` method?

lmacken:
Sweet, LGTM! :+1:

Issue #838, opened 2016-05-07
Updates stuck in "This update is currently locked and cannot be modified" state


ellert:
Updates submitted to testing for F23, F22, EPEL7, EPEL6 and EPEL5 between approx. 2016-05-02 14:30 UTC and 2016-05-03 09:00 UTC have been put in locked stated, but never got pushed.

Updates submitted after this time have been pushed successfully.

lmacken:
Can you please link to an update in this batch? It looks like there are multiple failed pushes that need to be resumed.

ellert:
Here is one in F23, but there are a lot more.

https://bodhi.fedoraproject.org/updates/FEDORA-2016-28e824b24b

ellert:
You can easily find all of them if you go to https://bodhi.fedoraproject.org/updates/?releases=F23&status=pending (and the following pages) and look for those that have "4 daya" or "5 days" in the age column. And similar of F22, EPEL-7, EL-6 and EL-5.

lupinix:
I recognized the same issue:
https://bodhi.fedoraproject.org/updates/FEDORA-EPEL-2016-64b17587f7
https://bodhi.fedoraproject.org/updates/FEDORA-2016-93684d7e11
https://bodhi.fedoraproject.org/updates/FEDORA-2016-002673fb41

ellert:
These updates have now been stuck for more than a week, and the issue was reported more than 4 days ago. As a packager responsible for same of these updates I feel very frustrated that nothing happens.

The last time this happened (see #828) it also took a long time to fix a similar issue. At that occasion - after having had no response to the ticket for several days - I filed a rel-eng ticket for the same issue and then it was fixed. However, some people got really irate about the second ticket and in what I consider to be an over-reaction wrote an angry reply about "why did you report this a second time - someone is already working on it". I had no idea anyone was working on it at that point because there had been no replies to the first ticket and I simply was trying to open a second line of communication with those that were in a position to fix the problem.

This time there has been a reply on the ticket - thank you for that it is reassuring - but still 3 days later the issue is still there. That updates end up in this undefined state should never happen in the first place. But if they do - and this is at least the second time it happens - I would like to suggest that there should be some automated warning sent to the bodhi admins so that it can be fixed promptly. I don't think it should take more than a week to fix such incidents.

What can I do to increase the priority to fix this? Should I open a rel-eng ticket again?

lmacken:
[...]

PR #839, opened 2016-05-12
Push (and warn about) locked updates that aren't a part of a current push

lmacken:
This will workaround and let us know when issues like #838 occur so we can
investigate things further.


pypingou:
Looks sane to me

lmacken:
Thanks @pypingou!

PR #841, opened 2016-05-18
Set PYTHONPATH to avoid import error

### LINKED ISSUE #840: Import error for modules on system commands in virtualenv
Whenever I try to run the system commands in the virtualenv I get import error for modules.
Starting from alembic command to run server.  

All the time  I have to write `PYTHONPATH=.` before the system commands in the virtualenv in order to execute them.

trishnaguha:
Fixes #840


lmacken:
LGTM, thanks!

PR #847, opened 2016-06-13
Document that users can use Vagrant for development.

bowlofeggs:


lmacken:
Awesome!

trishnaguha:
@bowlofeggs @lmacken Should we use fedora24 in vagrantfile?
fedora 24 cloud base image is already up :) https://atlas.hashicorp.com/fedora/boxes/24-cloud-base

PR #851, opened 2016-06-15
Bundle Bugzilla comments with the close command.

### LINKED ISSUE #404: Comment and close bugs in same operation
It looks like Bodhi 2.0 is performing two separate Bugzilla transactions when closing bugs: first it writes the "has been pushed to the Fedora 23 stable repository. If problems still persist, please make note of it in this bug report." message, then it sends another request to close the bug. This results in two Bugzilla mails to everyone subscribed to the bug. It'd be better if it used one transaction both to post the comment and close the bug (only in the case where 'close bugs when update pushed stable' is set, of course).

bowlofeggs:
Bodhi used to add a comment to a Bugzilla that the ticket was
resolved, and then close it in a separate transation. This commit
adjusts Bodhi to close and comment in the same Bugzilla transaction
to reduce the number of e-mails that users receive.

closes #404

### DIFF
--- a/bodhi/bugs.py
+++ b/bodhi/bugs.py
@@ -102,8 +102,12 @@ class Bugzilla(BugTracker):
         except:
             log.exception("Unable to alter bug #%d" % bug_id)
 
-    def close(self, bug_id, versions):
-        args = {}
+    def close(self, bug_id, versions, comment):
+        """
+        Close the bug given by bug_id, mark it as fixed in the given versions,
+        and add a comment.
+        """
+        args = {'comment': comment}
         try:
             bug = self.bz.getbug(bug_id)
             # If this bug is for one of these builds...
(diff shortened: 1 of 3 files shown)


bowlofeggs:
I haven't been able to access a testing Bugzilla deployment to test this pull request yet, so I've not hand tested this patch.

Additionally, it seems that none of the existing unit tests use the Bugzilla tracker code, but use the FakeBugTracker object instead. Due to that, I did not add unit tests for the changes I made to the Bugzilla tracker. I'm happy to do so if requested.

lmacken:
This looks good to me!

We should be able to test this locally against prod bugzilla in the Bodhi shell.

`pshell development.ini`

then something like:

```
u = m.Update.get(u'some-update', db)
u.modify_bugs()
u.bugs[0].close()
```

lmacken:
I'm just going to go ahead and merge this and test in staging. The patch looks sane, and there are multiple unit tests, the risk is fairly minimal :)

Issue #799, opened 2016-03-08
use different color for '0' in test cases section


matthiasclasen:
A page like https://bodhi.fedoraproject.org/updates/FEDORA-2016-606ca05253 more or less screams 'all green'. But if you get close enough, most of the test results are '0'. It would be less misleading to use blue or white as background for values that don't indicate a succeeded test.

userimack:
I am working on it.

akshayharidas:
Hi ,I am new to Fedroa. Can someone assign me to this bug?

lmacken:
@akshayharidas Absolutely, feel free to dive in! Please let me know if you need any help getting started.

akshayharidas:
Hi
So happy see your message .I am stated working on the bug, I cloned in my
pc and am stucked in the running time . It shows 'waiterss' distribution
was not found an pkg_resources. I  searched for the solution . Presently am
not done with that error . Hope  you can help me . Thank you :)
On 10-May-2016 1:07 AM, "Luke Macken" notifications@github.com wrote:

@akshayharidas https://github.com/akshayharidas Absolutely, feel free to
dive in! Please let me know if you need any help getting started.

akshayharidas:
I am getting "raise DistributionNotFound(req, requirers)
pkg_resources.DistributionNotFound: The 'waitress' distribution was not found and is required by the application"  these kinds of  errors 
I don't know what to do! 
![screenshot from 2016-06-05 13-26-51](https://cloud.githubusercontent.com/assets/16167718/15885303/f4b0e050-2d73-11e6-90f4-474208c31bbc.png)

trishnaguha:
@akshayharidas `setup.py` already has `waitress`. Check here https://github.com/fedora-infra/bodhi/blob/develop/setup.py#L24
Have you followed the instructions given on `readme` properly? 
You need to run `python setup.py develop` in virtualenv as given in readme.

https://github.com/fedora-infra/bodhi#setting-up after bootstrapping the virtualenv https://github.com/fedora-infra/bodhi#bootstrap-the-virtualenv

akshayharidas:
@trishnaguha 
>Have you followed the instructions given on readme properly? 

yes, I followed the instructions in https://github.com/fedora-infra/bodhi .

>You need to run python setup.py develop in virtualenv as given in readme.

[...]

PR #856, opened 2016-07-06
Add info in README:Make sure database is always upgraded

trishnaguha:
Makes sure the database is always upgraded to avoid issue like https://github.com/fedora-infra/bodhi/issues/799#issuecomment-230843661 while setting up development environment.


pnemade:
:+1:  This is really needed to be documented for people trying development environment.

lmacken:
Great, thanks! :+1:

PR #857, opened 2016-07-06
Disallow re-use of solved captchas.

ralphbean:
If someone solved a captcha, they could re-use it over and over again for 5 minutes.  Let's not let them do that.  ;)

This adds the cipherkey of the captcha to the session as a kind of sentinel.  We remove that flag once the captcha is used once.  If they try to use their answer again, it will fail since the flag for that captcha is no longer in the session.

Fixes CVE-2016-1000008.


pypingou:
LGTM :)

puiterwijk:
:+1:

lmacken:
Nice, thanks!

PR #858, opened 2016-07-06
Warp the captcha.

ralphbean:
Currently, you can crack the captcha with simple tesseract OCR.

This adds some warping to the image which should make that harder (but not
impossible...).  This is based off the vasasco_dowty warping buried in
tg2captcha.

Fixes CVE-2016-1000008.


lmacken:
I don't understand any of this, but let's ship it :shipit:

PR #861, opened 2016-07-08
virtualenv dependency update

clime:
fedmsg requires pyzmq, which requires zeromq-devel to be installed on the system as a building requirement for zmq C library.


lmacken:
I think it falls back to a statically compiled version if it can't build it, but this is fine with me. Thanks!

PR #863, opened 2016-07-09
README section ordering update

clime:
Section about creating development.ini file was moved before the section about running the testsuite because the testsuite needs the config file.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -66,6 +66,14 @@ Setting up
 
 ``pip install psycopg2``
 
+Create the `development.ini <https://github.com/fedora-infra/bodhi/blob/develop/development.ini.example>`_ file
+--------------------------------------------------------------------------------------------------------------------------
+
+Copy ``development.ini.example`` to ``development.ini``:
+::
+
+    cp development.ini.example development.ini
+    
 Run the test suite
 ------------------
 ``python setup.py test``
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -66,6 +66,14 @@ Setting up
 
 ``pip install psycopg2``
 
+Create the `development.ini <https://github.com/fedora-infra/bodhi/blob/develop/development.ini.example>`_ file
+--------------------------------------------------------------------------------------------------------------------------
```
pypingou: The underline seems larger than the title, I'm not sure that's expected/desired

README.rst
```diff
@@ -82,15 +90,10 @@ Import the bodhi2 database
...
-Copy ``development.ini.example`` to ``development.ini``:
-::
-
-    cp development.ini.example development.ini
+Adjust database configuration in `development.ini <https://github.com/fedora-infra/bodhi/blob/develop/development.ini.example>`_ file
+-------------------------------------------------------------------------------------------------------------------------------------
 
-Now adjust the configuration key
+Set the configuration key
 `sqlalchemy.url <https://github.com/fedora-infra/bodhi/blob/02d0a883c156d9a27a4dbac994409ecf726d00a9/development.ini#L413>`_
```
pypingou, line 97: Do we really want to point to a specific commit here?

pypingou:
One question but outside the scope of this PR

So :+1: for me, I'll let @lmacken do the merging if he agrees :)

README.rst
lmacken, line 97: Do we want to use https://github.com/fedora-infra/bodhi/blob/develop/development.ini.example#L416 instead?

pypingou, line 97: That would work for me

clime, line 97: Hello, I made the change. I hope it's ok.

clime, line 97: Actually, I have just realized why it was pointing to a specific commit...if line is added or removed before the line with url key, the key stops being on the referenced line 416 as it is now. I was just too quick and dumb. I guess it would be the best to remove my last commit from this PR...?

pypingou, line 97: The question becomes then: does it make sense to have a link at all?

clime, line 97: I think it is friendlier to users :).

lmacken:
This is fine with me. Thanks @clime !

PR #864, opened 2016-07-09
conditional for RHEL7 distro fixed

clime:
I think the operator should be 'greater than or equal' if we check for RHEL7 distro (the python-webob1.4 package inside the conditional exists only for EPEL7).


lmacken:
Cool, thanks!

PR #881, opened 2016-08-10
Remove cornice from the Sphinx project.

bowlofeggs:
The docs project currently does not build with a mysterious error
message. Removing cornice from the project seems to remove this
error message, though currently the entirety of the Sphinx project
is the cornice generated documentation. However, I am working on
using Sphinx to build a man page for the new bodhi 2 CLI, and so
the docs failing to build is a blocker for the work I need to do.

I will file an issue to re-examine cornice in the future when we
have time to determine what the issue was.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -23,13 +23,9 @@ import sys, os
 # If your documentation needs a minimal Sphinx version, state it here.
 #needs_sphinx = '1.0'
 
-import cornice
-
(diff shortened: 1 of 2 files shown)


bowlofeggs:
This commit is included in #883 because it blocks the work being done there. So as to avoid duplicate PR review, I'm going to close this one in favor of that one.

PR #883, opened 2016-08-11
Bodhi man page

bowlofeggs:


docs/man_bodhi.rst
```diff
@@ -0,0 +1,250 @@
...
+    Many commands accept this flag to specify which user's updates should be operated upon.
+
+
+Commands
+========
+
+There are two commands, ``overrides`` and ``updates``. They are described in more detail in their
+own sections below.
+
+``bodhi overrides [options] command <args>``
```
puiterwijk: "There are two commands, overrides and updates", and then there is a "command" blank in here. Wouldn't it be more consistent to call this a sub-command or an operation?
It's even called subcommands the line below.

docs/man_bodhi.rst
```diff
@@ -0,0 +1,250 @@
...
+    Provides an interface to manage updates. Supports subcommands ``comment``, ``download``,
+    ``new``, ``query``, and ``request``, described below.
+
+
+Overrides
+=========
+
+The ``overrides`` command allows users to manage build overrides.
+
+``bodhi overrides query [options]``
```
puiterwijk, line 62: And suddenly [options] is moved from just after "updates" to after the subcommand, and args are gone?

docs/man_bodhi.rst
bowlofeggs: I'll use sub-command to be more consistent with the rest of the document. Thanks!

bowlofeggs: I will also wrap subcommand in <>'s so it doesn't appear to be a literal.

docs/man_bodhi.rst
bowlofeggs, line 62: Haha, yeah I just copied most of these out of the --help text while driving the CLI. I'm guessing it's not particularly important where the flags go. To be honest, I'm not particularly familiar with the bodhi 2 CLI yet.

bowlofeggs, line 62: My guess is that options means things like `--staging` and what not, referenced at the top, and args means things specific to the subcommand. I'll do a pass and try to make that consistent.

bowlofeggs:
I changed the notation to indicate that options and args both come after subcommand (though I bet that they can go anywhere but I haven't tested) and both are optional (square brackets on args now) since not all subcommands take args.

puiterwijk:
:shipit:

PR #885, opened 2016-08-15
Change validator messages (fixes #630)

### LINKED ISSUE #630: Screwy validator messages

k3rn3l3rr0r:
This issue was sleeping for so long. I gave it a try and here is the PR, which should close it.
Tests were OK. Feedback appreciated.

### DIFF
--- a/bodhi/validators.py
+++ b/bodhi/validators.py
@@ -164,7 +164,12 @@ def validate_build_tags(request):
                 if not build_rel:
                     raise KeyError("Couldn't find release from build tags")
             except KeyError:
-                msg = 'Cannot find release associated with build: {}, tags: {}'.format(build, tags)
+                if tags:
+                    msg = 'Cannot find release associated with build: '
+                    '{}, tags: {}'.format(build, tags)
+                else:
+                    msg = 'Cannot find any tags associated with build: '
+                    '{}'.format(build)
                 log.warn(msg)
                 request.errors.add('body', 'builds', msg)
                 return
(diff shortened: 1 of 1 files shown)


bodhi/validators.py
```diff
@@ -164,7 +164,12 @@ def validate_build_tags(request):
                 if not build_rel:
                     raise KeyError("Couldn't find release from build tags")
             except KeyError:
-                msg = 'Cannot find release associated with build: {}, tags: {}'.format(build, tags)
+                if tags:
+                    msg = 'Cannot find release associated with build: '
```
lmacken: I don't think you can add a `\n` here without it being in parens or with a `\`?

k3rn3l3rr0r: Indeed. I was too fast, sorry for this.
I pushed a fix now.

lmacken:
LGTM, thanks!

Issue #361, opened 2015-08-25
ResultsDB Querying - switch from update.title to build names


rajcze:
At the moment, Bodhi queries ResultsDB on per-update.title basis. That effects multi-build updates in a way, which causes no results to be found (https://bodhi.fedoraproject.org/updates/control-center-3.16.3-1.fc22%20gnome-settings-daemon-3.16.3-1.fc22 for example).

This works with single-build results, as the koji-build-name is usually the same as bodhi-update-title.
The multi-build-update result querying did work to some extent in the past, as depcheck stored per-update(-title) results, as well as per-koji-build-name results, but then no rpmlint (for example) results were shown, although the builds contained in the update were tested.

I suggest querying ResultsDB on build-name(s) basis (all our current check store per-build results), which will give you all the results for all the builds in the update.

It might happen, that some multi-build updates will have quite a lot of automated results (gnome mega updates come to mind), but that can be sorted out by (e.g.) showing an overall result (computed from the singular per-build results, e.g. "if at least one build FAILed, overall result is FAIL" etc.), and implementing a drop-down to show the separate per-build results.

ralphbean:
Adjusting the queries will not be difficult, but designing/implementing the drop-down will be a pain.  Can do, though!

ralphbean:
A reminder to whoever implements this: taskotron results are used in three places:

* In javascript in the web client to show results in a table on the update page.
* In python on the server when someone requests an update go to stable.. we check to make sure they pass all *required* taskotron checks at that point.
* In python again, but this time in the masher at the beginning of the push.   We eject any updates from the push which fail any *required* taskotron checks.

In practice, this only needs to be implemented in two places since the python code is shared in ``bodhi/models/models.py`` .. specifically in the ``check_requirements(self, ...)`` method.

ralphbean:
Note that #890 fixed the UI portion of this only.  The server-side mash-time checks still need to be adjusted.

(In retrospect, that's a little dangerous because people might see the UI and think all checks are being consulted, when in reality it's still just the per-update checks).

kparal:
Do I understand correctly that some parts of Bodhi still send `type=bodhi_update` requests and we should continue producing these results until that is resolved?

AdamWill:
[...]

PR #890, opened 2016-08-22
(UI) Check taskotron checks per build, not update.

ralphbean:
Taskotron records results on a per-build basis. We previously only queried
for results on a per-update basis, which meant that we missed a lot of the
new checks.

This change makes it so that we show them all.


kparal:
Btw this is ticket #361.

lmacken:
Yep, looks great. Thanks @ralphbean ! :cake: :balloon:

PR #893, opened 2016-08-26
Convert the Vagrantfile to use Ansible instead of shell commands.

bowlofeggs:


sayanchowdhury:
Looks error-free to me :+1:

PR #894, opened 2016-08-29
Disable AutoPush for critical update if it receives negative karma

### LINKED ISSUE #879: Disable autokarma if anyone gives a -1 to 'basic system functionality'
If someone gives negative feedback with the "Does the system's basic functionality continue to work after this update?", we should disable the automatic pushing based on stablekarma...

We probably want to leave the unstable-karma auto-unpushing in this caes, but disable the automatic pushing to stable when it this happens.

trishnaguha:
Fixes #879

### DIFF
--- a/bodhi/server/models/models.py
+++ b/bodhi/server/models/models.py
@@ -1653,7 +1653,11 @@ class Update(Base):
         """Check if we have reached either karma threshold, and call set_request if necessary"""
         if not self.locked:
             if self.status is UpdateStatus.testing:
-                if self.stable_karma not in (0, None) and self.karma >= self.stable_karma:
+                # If critical update receives negative karma disable autopush
+                if self.critpath and self.autokarma and self.has_negative_karma:
+                    log.info("Disabling Auto Push since the critical update has negative karma")
+                    self.autokarma = False
+                elif self.stable_karma not in (0, None) and self.karma >= self.stable_karma:
                     if self.autokarma:
                         log.info("Automatically marking %s as stable" % self.title)
                         self.set_request(db, UpdateRequest.stable, agent)
@@ -1706,6 +1710,18 @@ class Update(Base):
 
         possibilities.sort()  # Sort smallest to largest (oldest to newest)
         return possibilities[-1]  # Return the last one
+
+    @property
+    def has_negative_karma(self):
+        """Check for negative karma, Returns True if the update has negative karma"""
+        feedback = defaultdict(int)
+        for comment in self.comments:
+            if not comment.anonymous:
+                feedback[comment.user.name] = comment.karma
+        for karma in feedback.values():
+            if karma < 0:
+                return True
+        return False
 
     @property
     def critpath_approved(self):
--- a/bodhi/tests/server/functional/test_updates.py
+++ b/bodhi/tests/server/functional/test_updates.py
@@ -2520,6 +2520,379 @@ class TestUpdatesService(bodhi.tests.server.functional.base.BaseWSGICase):
 
     @mock.patch(**mock_valid_requirements)
     @mock.patch('bodhi.server.notifications.publish')
+    def test_has_negative_karma(self, publish, *args):
+        """The test asserts that has_negative_karma returns True when an update receives negative karma"""
+        user = User(name=u'bob')
+        self.db.add(user)
+        self.db.flush()
+
+        nvr = 'bodhi-2.1-1.fc17'
+        args = self.get_update(nvr)
+        resp = self.app.post_json('/updates/', args).json_body
+        publish.assert_called_with(topic='update.request.testing', msg=ANY)
+
+        up = self.db.query(Update).filter_by(title=nvr).one()
+        up.request = None
+        up.status = UpdateStatus.testing
+        self.db.flush()
+
+        # The user gives negative karma first
+        up.comment(self.db, u'Failed to work', author=u'luke', karma=-1)
+        up = self.db.query(Update).filter_by(title=nvr).one()
+        self.assertTrue(up.has_negative_karma)
+
+    @mock.patch(**mock_valid_requirements)
+    @mock.patch('bodhi.server.notifications.publish')
+    def test_has_no_negative_karma_with_changed_karma(self, publish, *args):
+        """
+        This test asserts that has_negative_karma returns False when a user posts negative karma 
+        and then later posts positive karma.
+        """
+        user = User(name=u'bob')
+        self.db.add(user)
+        self.db.flush()
+
+        nvr = 'bodhi-2.1-1.fc17'
+        args = self.get_update(nvr)
+        resp = self.app.post_json('/updates/', args).json_body
+        publish.assert_called_with(topic='update.request.testing', msg=ANY)
+
+        up = self.db.query(Update).filter_by(title=nvr).one()
+        up.request = None
+        up.status = UpdateStatus.testing
+        self.db.flush()
+
+        # The user gives negative karma first
+        up.comment(self.db, u'Failed to work', author=u'ralph', karma=-1)
+        up = self.db.query(Update).filter_by(title=nvr).one()
+        self.assertTrue(up.has_negative_karma)
+
+        # The same user gives positive karma later
+        up.comment(self.db, u'wfm', author=u'ralph', karma=1)
+        up = self.db.query(Update).filter_by(title=nvr).one()
+        self.assertFalse(up.has_negative_karma)
+        self.assertEquals(up.karma, 1)
+
+    @mock.patch(**mock_valid_requirements)
+    @mock.patch('bodhi.server.notifications.publish')
+    def test_has_negative_karma_with_positive_karma_first(self, publish, *args):
+        """
+        This test asserts that has_negative_karma returns True when a user posts positive karma 
+        and then another user posts negative karma.
+        """
+        user = User(name=u'bob')
+        self.db.add(user)
+        self.db.flush()
+
+        nvr = u'bodhi-2.1-1.fc17'
+        args = self.get_update(nvr)
+        resp = self.app.post_json('/updates/', args).json_body
+        publish.assert_called_with(topic='update.request.testing', msg=ANY)
+
+        up = self.db.query(Update).filter_by(title=nvr).one()
+        up.request = None
+        up.status = UpdateStatus.testing
+        self.db.flush()
+
+        #  user gives positive karma first
+        up.comment(self.db, u'Works for me', author=u'ralph', karma=1)
+        up = self.db.query(Update).filter_by(title=nvr).one()
+        self.assertFalse(up.has_negative_karma)
+
+        # Another user gives negative karma later
+        up.comment(self.db, u'Failed to work', author=u'bowlofeggs', karma=-1)
+        up = self.db.query(Update).filter_by(title=nvr).one()
+        self.assertTrue(up.has_negative_karma)
+        self.assertEquals(up.karma, 0)
+
+    @mock.patch(**mock_valid_requirements)
+    @mock.patch('bodhi.server.notifications.publish')
+    def test_has_no_negative_karma(self, publish, *args):
+        """The test asserts that has_negative_karma returns False when there is no negative karma"""
+        user = User(name=u'bob')
+        self.db.add(user)
+        self.db.flush()
+
+        nvr = 'bodhi-2.1-1.fc17'
+        args = self.get_update(nvr)
+        resp = self.app.post_json('/updates/', args).json_body
+        publish.assert_called_with(topic='update.request.testing', msg=ANY)
+
+        up = self.db.query(Update).filter_by(title=nvr).one()
+        up.request = None
+        up.status = UpdateStatus.testing
+        self.db.flush()
+
+        up.comment(self.db, u'LGTM', author=u'mac', karma=1)
+        up = self.db.query(Update).filter_by(title=nvr).one()
+        self.assertFalse(up.has_negative_karma)
+
+        # Karma with no comment
+        up.comment(self.db, u' ', author=u'bowlofeggs', karma=1)
+        up = self.db.query(Update).filter_by(title=nvr).one()
+        self.assertFalse(up.has_negative_karma)
+        self.assertEquals(up.karma, 2)
+
+    @mock.patch(**mock_valid_requirements)
+    @mock.patch('bodhi.server.notifications.publish')
+    def test_has_negative_karma_with_Anonymous_karma(self, publish, *args):
+        """
+        The test asserts that has_negative_karma returns False when an anonymous user
+        gives negative karma to an update
+        """
+        user = User(name=u'bob')
+        self.db.add(user)
+        self.db.flush()
+
+        nvr = 'bodhi-2.1-1.fc17'
+        args = self.get_update(nvr)
+        resp = self.app.post_json('/updates/', args).json_body
+        publish.assert_called_with(topic='update.request.testing', msg=ANY)
+
+        up = self.db.query(Update).filter_by(title=nvr).one()
+        up.request = None
+        up.status = UpdateStatus.testing
+        self.db.flush()
+
+        up.comment(self.db, u'Not working', author='me', anonymous=True, karma=-1)
+        up = self.db.query(Update).filter_by(title=nvr).one()
+        self.assertFalse(up.has_negative_karma)
+
+    @mock.patch(**mock_valid_requirements)
+    @mock.patch('bodhi.server.notifications.publish')
+    def test_has_negative_karma_with_no_feedback(self, publish, *args):
+        """This test asserts that has_negative_karma return False when an update has no feedback"""
+        user = User(name=u'bob')
+        self.db.add(user)
+        self.db.flush()
+
+        nvr = 'bodhi-2.1-1.fc17'
+        args = self.get_update(nvr)
+        resp = self.app.post_json('/updates/', args).json_body
+        publish.assert_called_with(topic='update.request.testing', msg=ANY)
+
+        up = self.db.query(Update).filter_by(title=nvr).one()
+        up.request = None
+        up.status = UpdateStatus.testing
+        self.db.flush()
+
+        up = self.db.query(Update).filter_by(title=nvr).one()
+        self.assertFalse(up.has_negative_karma)
+
+    @mock.patch(**mock_valid_requirements)
+    @mock.patch('bodhi.server.notifications.publish')
+    def test_disable_autopush_for_critical_updates(self, publish, *args):
+        """Make sure that autopush is disabled if a critical update receives any negative karma"""
+        user = User(name=u'bob')
+        self.db.add(user)
+        self.db.flush()
+
+        nvr = u'kernel-3.11.5-300.fc17'
+        args = self.get_update(nvr)
+        args['autokarma'] = True
+        resp = self.app.post_json('/updates/', args)
+        self.assertTrue(resp.json['critpath'])
+        self.assertEquals(resp.json['request'], 'testing')
+        publish.assert_called_with(topic='update.request.testing', msg=ANY)
+
+        up = self.db.query(Update).filter_by(title=resp.json['title']).one()
+        up.status = UpdateStatus.testing
+        up.request = None
+        self.db.flush()
+
+        # A user gives negative karma first
+        up.comment(self.db, u'Failed to work', author=u'ralph', karma=-1)
+        up = self.db.query(Update).filter_by(title=resp.json['title']).one()
+
+        # Another user gives positive karma
+        up.comment(self.db, u'wfm', author=u'bowlofeggs', karma=1)
+        up = self.db.query(Update).filter_by(title=resp.json['title']).one()
+
+        self.assertEquals(up.karma, 0)
+        self.assertEquals(up.status, UpdateStatus.testing)
+        self.assertEquals(up.request, None)
+
+        # Autopush gets disabled since there is a negative karma from ralph
+        self.assertEquals(up.autokarma, False)
+
+    @mock.patch(**mock_valid_requirements)
+    @mock.patch('bodhi.server.notifications.publish')
+    def test_autopush_critical_update_with_no_negative_karma(self, publish, *args):
+        """Autopush critical update when it has no negative karma"""
+        user = User(name=u'bob')
+        self.db.add(user)
+        self.db.flush()
+
+        nvr = u'kernel-3.11.5-300.fc17'
+        args = self.get_update(nvr)
+        args['autokarma'] = True
+        args['stable_karma'] = 2
+        args['unstable_karma'] = -2
+
+        resp = self.app.post_json('/updates/', args)
+        self.assertTrue(resp.json['critpath'])
+        self.assertEquals(resp.json['request'], 'testing')
+        publish.assert_called_with(topic='update.request.testing', msg=ANY)
+
+        up = self.db.query(Update).filter_by(title=resp.json['title']).one()
+        up.status = UpdateStatus.testing
+        self.db.flush()
+
+        up.comment(self.db, u'LGTM', author=u'ralph', karma=1)
+        up = self.db.query(Update).filter_by(title=resp.json['title']).one()
+
+        up.comment(self.db, u'LGTM', author=u'bowlofeggs', karma=1)
+        up = self.db.query(Update).filter_by(title=resp.json['title']).one()
+        self.assertEquals(up.karma, 2)
+
+        # No negative karma: Update gets automatically marked as stable
+        self.assertEquals(up.autokarma, True)
+
+        up = self.db.query(Update).filter_by(title=resp.json['title']).one()
+        self.assertEquals(up.request, UpdateRequest.stable)
+
+    @mock.patch(**mock_valid_requirements)
+    @mock.patch('bodhi.server.notifications.publish')
+    def test_manually_push_critical_update_with_negative_karma(self, publish, *args):
+        """
+        Manually push critical update when it has negative karma
+        Autopush gets disabled after it receives negative karma
+        A user gives negative karma, but another 3 users give positive karma
+        The critical update should be manually pushed because of the negative karma
+        """
+        user = User(name=u'bob')
+        self.db.add(user)
+        self.db.flush()
+
+        nvr = u'kernel-3.11.5-300.fc17'
+        args = self.get_update(nvr)
+        args['autokarma'] = True
+        args['stable_karma'] = 3
+        args['unstable_karma'] = -3
+
+        resp = self.app.post_json('/updates/', args)
+        self.assertTrue(resp.json['critpath'])
+        self.assertEquals(resp.json['request'], 'testing')
+        publish.assert_called_with(topic='update.request.testing', msg=ANY)
+
+        up = self.db.query(Update).filter_by(title=resp.json['title']).one()
+        up.status = UpdateStatus.testing
+        self.db.flush()
+
+        up.comment(self.db, u'Failed to work', author=u'ralph', karma=-1)
+        up = self.db.query(Update).filter_by(title=resp.json['title']).one()
+
+        up.comment(self.db, u'LGTM', author=u'bowlofeggs', karma=1)
+        up = self.db.query(Update).filter_by(title=resp.json['title']).one()
+
+        up.comment(self.db, u'wfm', author=u'luke', karma=1)
+        up = self.db.query(Update).filter_by(title=resp.json['title']).one()
+
+        up.comment(self.db, u'LGTM', author=u'puiterwijk', karma=1)
+        up = self.db.query(Update).filter_by(title=resp.json['title']).one()
+
+        up.comment(self.db, u'LGTM', author=u'trishnag', karma=1)
+        up = self.db.query(Update).filter_by(title=resp.json['title']).one()
+
+        self.assertEquals(up.karma, 3)
+        self.assertEquals(up.autokarma, False)
+
+        # Checks Push to Stable text in the html page for this update
+        id = 'kernel-3.11.5-300.fc17'
+        resp = self.app.get('/updates/%s' % id,
+                            headers={'Accept': 'text/html'})
+        self.assertIn('text/html', resp.headers['Content-Type'])
+        self.assertIn(id, resp)
+        self.assertIn('Push to Stable', resp)
+
+    @mock.patch(**mock_valid_requirements)
+    @mock.patch('bodhi.server.notifications.publish')
+    def test_manually_push_critical_update_with_autopush_turned_off(self, publish, *args):
+        """
+        Manually push critical update when it has Autopush turned off
+        and make sure the update doesn't get Autopushed
+        """
+        user = User(name=u'bob')
+        self.db.add(user)
+        self.db.flush()
+
+        nvr = u'kernel-3.11.5-300.fc17'
+        args = self.get_update(nvr)
+        args['autokarma'] = False
+        args['stable_karma'] = 3
+        args['unstable_karma'] = -3
+
+        resp = self.app.post_json('/updates/', args)
+        self.assertTrue(resp.json['critpath'])
+        self.assertEquals(resp.json['request'], 'testing')
+        publish.assert_called_with(topic='update.request.testing', msg=ANY)
+
+        up = self.db.query(Update).filter_by(title=resp.json['title']).one()
+        up.status = UpdateStatus.testing
+        self.db.flush()
+
+        up.comment(self.db, u'LGTM Now', author=u'ralph', karma=1)
+        up = self.db.query(Update).filter_by(title=resp.json['title']).one()
+
+        up.comment(self.db, u'wfm', author=u'luke', karma=1)
+        up = self.db.query(Update).filter_by(title=resp.json['title']).one()
+
+        up.comment(self.db, u'LGTM', author=u'puiterwijk', karma=1)
+        up = self.db.query(Update).filter_by(title=resp.json['title']).one()
+
+        self.assertEquals(up.karma, 3)
+        self.assertEquals(up.autokarma, False)
+
+        # Checks Push to Stable text in the html page for this update
+        id = 'kernel-3.11.5-300.fc17'
+        resp = self.app.get('/updates/%s' % id,
+                            headers={'Accept': 'text/html'})
+        self.assertIn('text/html', resp.headers['Content-Type'])
+        self.assertIn(id, resp)
+        self.assertIn('Push to Stable', resp)
+
+    @mock.patch(**mock_valid_requirements)
+    @mock.patch('bodhi.server.notifications.publish')
+    def test_autopush_non_critical_update_with_negative_karma(self, publish, *args):
+        """Autopush Non Critical update even though it receives negative karma"""
+        user = User(name=u'bob')
+        self.db.add(user)
+        self.db.flush()
+
+        nvr = u'bodhi-2.0.0-2.fc17'
+        args = self.get_update(nvr)
+        args['autokarma'] = True
+        args['stable_karma'] = 3
+        args['unstable_karma'] = -3
+
+        resp = self.app.post_json('/updates/', args)
+        self.assertEquals(resp.json['request'], 'testing')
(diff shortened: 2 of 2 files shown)


bodhi/models/models.py
```diff
@@ -1649,19 +1649,35 @@ def check_requirements(self, session, settings):
...
                 if self.stable_karma not in (0, None) and self.karma >= self.stable_karma:
                     if self.autokarma:
-                        log.info("Automatically marking %s as stable" % self.title)
-                        self.set_request(db, UpdateRequest.stable, agent)
-                        self.request = UpdateRequest.stable
-                        self.date_pushed = None
-                        notifications.publish(
-                            topic='update.karma.threshold.reach',
-                            msg=dict(update=self, status='stable'))
+                        if self.critpath and self.negative_karma:
```
bowlofeggs: `negative_karma()` is a method above, but here it is used like a property (i.e., it's not called with ()'s). You probably want to either add ()'s here, or add the `@property` decorator to the function above.

bodhi/models/models.py
```diff
@@ -1649,19 +1649,35 @@ def check_requirements(self, session, settings):
 
         return True, "All checks pass."
 
+    def negative_karma(self):
+        """Checks for negative karma, Returns True if the update has negative karma"""
```
bowlofeggs: Consider renaming the method to sound a little bit more actionable. Something like `has_negative_karma()` is a little clearer, as it hints that it might be returning a boolean value. This is just a nitpick, so you can leave it as is if you prefer.

bodhi/models/models.py
```diff
@@ -1649,19 +1649,35 @@ def check_requirements(self, session, settings):
...
+        feedback = defaultdict(int)
```
bowlofeggs: Also a nitpick that you can ignore, but consider using `defaultdict(lambda: 0)` here instead simply for being more explicit. You can leave it as is if you prefer.

bodhi/models/models.py
```diff
@@ -1649,19 +1649,35 @@ def check_requirements(self, session, settings):
...
+        for comment in self.comments:
+            if not comment.anonymous:
+                feedback[comment.user.name] +=  comment.karma
+        for karma in feedback.values():
+            if karma < 0:
+                return True
+
```
bowlofeggs: This function will return True if there is any negative feedback, but it will return None if there is not any negative feedback. If you tack on a final return False at the end here, this function will always return a boolean, rather than alternating return types. Strange bugs can happen with mixed return types sometimes ☺

bodhi/models/models.py
```diff
@@ -1649,19 +1649,35 @@ def check_requirements(self, session, settings):
...
                     if self.autokarma:
-                        log.info("Automatically marking %s as stable" % self.title)
-                        self.set_request(db, UpdateRequest.stable, agent)
-                        self.request = UpdateRequest.stable
-                        self.date_pushed = None
-                        notifications.publish(
-                            topic='update.karma.threshold.reach',
-                            msg=dict(update=self, status='stable'))
+                        if self.critpath and self.negative_karma:
+                            log.info("Disabling Auto Push since the critic update has negative karma")
```
bowlofeggs: s/critic/critical/

bodhi/models/models.py
```diff
@@ -1649,19 +1649,35 @@ def check_requirements(self, session, settings):
...
+                            self.autokarma = False
+                            self.status = UpdateStatus.testing
+                            self.request = None
```
bowlofeggs: My familiarity with Bodhi is not strong yet, but it is not obvious to me that self gets saved to the database after these values are adjusted. In the else block, I do see the call to self.set_request(), and then some attribute changes similar to what is seen here. Do you know where these values are used and/or set into the database? I really need to sit down and just read all the Bodhi code soon!

bowlofeggs: All that said, it _is_ obvious to me that the other code won't get called if this block is executed, so we do know that the update won't go to stable (which is good!)

bodhi/tests/functional/test_updates.py
```diff
@@ -2520,6 +2520,39 @@ def test_edit_testing_update_reset_karma_with_same_tester(self, publish, *args):
...
+        resp = self.app.post_json('/updates/', args)
+        self.assertTrue(resp.json['critpath'])
+        self.assertEquals(resp.json['request'], 'testing')
+        publish.assert_called_with(topic='update.request.testing', msg=ANY)
+
+        upd = self.db.query(Update).filter_by(title=resp.json['title']).one()
+        upd.status = UpdateStatus.testing
+        upd.request = None
+
+        upd.comment(self.db, u'Failed to work', author=u'ralph', karma=-1)
```
bowlofeggs: @ralphbean is always finding bugs!

bodhi/tests/functional/test_updates.py
```diff
@@ -2520,6 +2520,39 @@ def test_edit_testing_update_reset_karma_with_same_tester(self, publish, *args):
...
+        upd.autokarma = False
+        self.db.flush()
+        self.assertEquals(upd.karma, -1)
+        self.assertEquals(upd.status, UpdateStatus.testing)
+        self.assertEquals(upd.request, None)
+        self.assertEquals(upd.autokarma, False)
```
bowlofeggs: It would be good to ensure that no calls were made to set_request(), since that's the primary action we are trying to prevent.

bodhi/tests/functional/test_updates.py
```diff
@@ -2520,6 +2520,39 @@ def test_edit_testing_update_reset_karma_with_same_tester(self, publish, *args):
...
+        self.assertTrue(resp.json['critpath'])
+        self.assertEquals(resp.json['request'], 'testing')
+        publish.assert_called_with(topic='update.request.testing', msg=ANY)
+
+        upd = self.db.query(Update).filter_by(title=resp.json['title']).one()
+        upd.status = UpdateStatus.testing
+        upd.request = None
+
+        upd.comment(self.db, u'Failed to work', author=u'ralph', karma=-1)
+        upd.autokarma = False
```
bowlofeggs: Why is this getting set here?

bodhi/tests/functional/test_updates.py
bowlofeggs: I believe this test might not be hitting the new code path. The code has an if statement to check if the karma has reached the autopush threshold before the new code will execute. I could be misreading the conditions here, but it seems to me that the karma on `upd` is -1 and thus wouldn't push anyway. I think you probably need three upvoters after this -1 to reach an overall karma of +2 before this test will hit the new code path that is introduced in this commit. Does that make sense?

I didn't actually run this test through pdb to be sure, so I could be incorrect and I apologize if I am ☺

bodhi/tests/functional/test_updates.py
bowlofeggs: You could also add an assertion that your new `log.info()` call is happening, so that you can be sure that this test is hitting the new code path.

bodhi/tests/functional/test_updates.py
```diff
@@ -2520,6 +2520,39 @@ def test_edit_testing_update_reset_karma_with_same_tester(self, publish, *args):
...
+        self.db.flush()
+        self.assertEquals(upd.karma, -1)
```
bowlofeggs: I would think we want to ensure that this is +2, since that is the condition that we are trying to prevent the push from happening in.

bodhi/tests/functional/test_updates.py
bowlofeggs: I think upd.autokarma might only be false because it is getting set to False on line 2547 above.

bodhi/tests/functional/test_updates.py
bowlofeggs: We should remove this, because I believe it is causing an assertion to pass that seems like it should be failing below, since the comment as is didn't go through the new code path.

bodhi/tests/functional/test_updates.py
bowlofeggs: Why not add a few dedicated unit tests for the new `negative_karma()` method as well?

bowlofeggs:
Hey @trishnaguha! I'm done with my first pass on this review. It's looking pretty good so far! If you just fix a few things I think we'll be ready to merge. Here are the high-level things to address, and then I'll take another look:

0) Watch out for that method call, to either make sure to call it with ()'s or make it a property.
1) Make sure that method returns a boolean no matter what, instead of True/None.
2) Fix that spelling error in the log message.
3) Investigate whether those attribute changes are getting saved into the database. If you want help with this, I'm happy to do so. It's nearing 6 PM my local time so I don't have time to test it just now but I'm happy to if you want. It would be good for both of us to know this bit well anyway ☺
4) Make sure that your new unit test is hitting the new code path you've created. I believe that it is not exercising that code as is, as you will need to generate a condition where a new comment would traditionally have caused an autopush.
5) Add a few more assertions that make sure the new code path is getting hit, and that the old one isn't.
6) Add some positive and negative test cases for the new `negative_karma()` method. Here are a few conditions that I think would be easy to test: 0) There are no comments, 1) There are a couple or three positive comments, 2) A user has given a negative and then a positive karma, which shouldn't be counted as the update having negative karma, 3) There are some positive comments and one negative comment.

If you'd like me to explain any of my comments better, feel free to ask here or IRC. Nice work!

trishnaguha:
@bowlofeggs Thanks for the review. You have made it crystal clear :). I am going to work on this today.

trishnaguha:
@bowlofeggs Here is the test and script for the method `has_negative_karma`. https://paste.fedoraproject.org/417833/

trishnaguha:
@bowlofeggs This is the updated patch.
So the test is passing for negative karma. But it is failing to set `autokarma column` to `False`. I am not sure why. Can you please take a look at it?

bowlofeggs:
@trishnaguha my guess is that it isn't setting autokarma to False in the database because there isn't any call that saves the object to the database after that attribute is set. This is what I was talking about in my third point on my last post, that we might want to make sure the object is getting saved to the database.

bodhi/models/models.py
```diff
@@ -1649,6 +1649,27 @@ def check_requirements(self, session, settings):
...
+        return False
+
+    def diasble_autopush_for_critical_update(self, db):
+        """Disable Autopush if a critical update receives any negative karma"""
+        if not self.locked:
+            if self.status is UpdateStatus.testing:
+                if self.autokarma and self.critpath and self.has_negative_karma:
+                    log.info("Disabling Auto Push since the critical update has negative karma")
+                    self.autokarma = False
+                    db.commit()
```
bowlofeggs: I don't think we want to commit the transaction here. In general, we want to commit transactions at a higher level so that any later errors don't cause a partial change to be committed. i.e., if this were committed, and then there were an Exception later in the response handling, we could end up with a "half done" operation. Does that make sense?

We probably just want to make sure there is a db.commit() happening higher in the stack than this, and I suspect there is.

bodhi/models/models.py
```diff
@@ -1649,6 +1649,27 @@ def check_requirements(self, session, settings):
...
+        else:
+            log.debug('%s locked. Ignoring karma thresholds.' % self.title)
+            raise LockedUpdateException
```
bowlofeggs: Generally, it's good practice to instantiate Exception objects when raising them (though it's not strictly necessary).

bodhi/models/models.py
```diff
@@ -1649,6 +1649,27 @@ def check_requirements(self, session, settings):
...
+    @property
+    def has_negative_karma(self):
+        """Check for negative karma, Returns True if the update has negative karma"""
+        for comment in self.comments:
+            if not comment.anonymous:
+                if comment.karma < 0:
+                    return True
+        return False
+
+    def diasble_autopush_for_critical_update(self, db):
```
bowlofeggs: Does anything call this method? I don't see any calls to it in this change, and it's a new method.

bodhi/models/models.py
```diff
@@ -1649,6 +1649,27 @@ def check_requirements(self, session, settings):
 
         return True, "All checks pass."
 
+    @property
+    def has_negative_karma(self):
+        """Check for negative karma, Returns True if the update has negative karma"""
+        for comment in self.comments:
+            if not comment.anonymous:
+                if comment.karma < 0:
+                    return True
```
bowlofeggs: Your previous implementation of this method seemed to take into account the possibility of a user giving an accidental -1 and then a later +1. It seems that this code no longer handles that possibility, and will count accidental -1's and mark it as negative karma.

bodhi/tests/functional/test_updates.py
```diff
@@ -2520,6 +2520,69 @@ def test_edit_testing_update_reset_karma_with_same_tester(self, publish, *args):
 
     @mock.patch(**mock_valid_requirements)
     @mock.patch('bodhi.notifications.publish')
+    def test_for_negative_karma(self, publish, *args):
+        """
+        Checks if an update has any negative karma
+        The test asserts True for has_negative_karma i.e, if there exists a negative karma
+        The test asserts False for has_negative_karma i.e, if there exists no negative karma
```
bowlofeggs: It's good practice to make these be separate test methods. i.e., make one test per condition.

bodhi/tests/functional/test_updates.py
```diff
@@ -2520,6 +2520,69 @@ def test_edit_testing_update_reset_karma_with_same_tester(self, publish, *args):
...
+        up = self.db.query(Update).filter_by(title=nvr).one()
+        self.assertFalse(up.has_negative_karma)
+
+        up.comment(self.db, u'Failed to work', author=u'ralph', karma=-1)
+        up = self.db.query(Update).filter_by(title=nvr).one()
+        self.assertTrue(up.has_negative_karma)
+
+        up.comment(self.db, u'Doesn\'t work', author=u'bowlofeggs', karma=-1)
+        up = self.db.query(Update).filter_by(title=nvr).one()
+        self.assertTrue(up.has_negative_karma)
```
bowlofeggs: As I mentioned, please split this method up into several tests. There are some more tests you should add too, for example, a test where a user sets negative karma in one comment, and then positive karma in a later comment. You should also run up.has_negative_karma in a test where there are no comments to make sure it returns False.

bodhi/tests/functional/test_updates.py
```diff
@@ -2520,6 +2520,69 @@ def test_edit_testing_update_reset_karma_with_same_tester(self, publish, *args):
...
+        upd.request = None
+        self.db.flush()
+
+        upd.comment(self.db, u'Failed to work', author=u'ralph', karma=-1)
+        up = self.db.query(Update).filter_by(title=resp.json['title']).one()
+        self.assertTrue(up.has_negative_karma)
+        self.assertEquals(up.karma, -1)
+        self.assertEquals(up.status, UpdateStatus.testing)
+        self.assertEquals(up.request, None)
+        self.assertEquals(up.autokarma, False)
```
bowlofeggs: It doesn't look like anything in the tests or production code is calling `disable_autopush_for_critical_updates()`. This test seems to mostly be asserting that the has_negative_karma works and that the sqlalchemy writes above worked.

bowlofeggs:
@trishnaguha I think your last commit was closer to what we need. This one seems like it drops the usage of the new code, and it also lost the ability to handle "accidental" karma events from users.

trishnaguha:
@bowlofeggs ah thanks. I felt this would be more simple than the previous `negative_karma_check`, but ignored the accidental karma event concept.

trishnaguha:
@bowlofeggs Is my previous commit I set `False` to `autokarma` after https://github.com/fedora-infra/bodhi/pull/894#discussion-diff-76882047R1666  But what if `self.karma` is not greater than equal to `self.stable_karma`? For crtipath updates we are checking negative karma for each feedback. So it won't exercise the code path if `self.karma is less than self.stable_karma` and it would fail to set `autokarma` to `False` even if it receives negative karma on feedback. So I thought to add a separate method. 
Let me know what is your thought on this.

bowlofeggs:
@trishnaguha I like that you made a separate method, because it keeps the code cleaner. However, this method isn't getting called by anything so it doesn't perform any action. You probably want to call this method in the area that you had added its functionality in your initial commit.

Does that help?

bodhi/tests/functional/test_updates.py
trishnaguha: @bowlofeggs Yeah our purpose is to check whether `autokarma` is False after critical update receives negative karma during testing. I think we don't need to call the method `disable_autopush_for_critical_updates()` in the testcase. If `disable_autopush_for_critical_updates()` is called in the model itself, the test is supposed to pass.
So I am just trying to work on calling the method in the model.

trishnaguha:
This is the updated patch. The test suite seems to pass all the tests :).

bodhi/server/models/models.py
```diff
@@ -1649,11 +1649,31 @@ def check_requirements(self, session, settings):
 
         return True, "All checks pass."
 
+    @property
```
sayanchowdhury: Can you group this property with all the other properties below?

bodhi/server/models/models.py
```diff
@@ -1649,11 +1649,31 @@ def check_requirements(self, session, settings):
...
+                if karma < 0:
+                    return True
+        return False
+
+    def disable_autopush_for_critical_update(self):
+        """Disable Autopush when critical update receives any negative karma"""
+        log.debug("Disabling Auto Push since the critical update has negative karma")
+        self.autokarma = False
+
     def check_karma_thresholds(self, db, agent):
```
sayanchowdhury, line 1652: This method needs a refactor. Too much depth of nested-if

sayanchowdhury:
Don't have much knowledge about the code-base but looked good to me other than the comments.

bodhi/server/models/models.py
trishnaguha: Sure :).

trishnaguha: Grouped the property with other properties :).

bodhi/tests/server/functional/test_updates.py
```diff
@@ -2520,6 +2520,99 @@ def test_edit_testing_update_reset_karma_with_same_tester(self, publish, *args):
 
     @mock.patch(**mock_valid_requirements)
     @mock.patch('bodhi.server.notifications.publish')
+    def test_has_negative_karma(self, publish, *args):
+        """
+        The test asserts True for has_negative_karma i.e,
+        if update has negative karma
+        """
```
bowlofeggs, line 2550: The name of the test should reflect the condition it is testing, rather than just the name of the method it is testing. For example, a better name might be `test_has_negative_karma_with_changed_karma`, since this test is testing a user who gave negative feedback and then changed their mind. Also, the test documentation here is incorrect - this test is asserting False, not true. Make sure that the test documentation covers the condition that is being created below.

bodhi/tests/server/functional/test_updates.py
```diff
@@ -2520,6 +2520,99 @@ def test_edit_testing_update_reset_karma_with_same_tester(self, publish, *args):
...
+
+        up = self.db.query(Update).filter_by(title=nvr).one()
+        up.request = None
+        up.status = UpdateStatus.testing
+        self.db.flush()
+
+        # The user gives negative karma first
+        up.comment(self.db, u'Failed to work', author=u'ralph', karma=-1)
+        up = self.db.query(Update).filter_by(title=nvr).one()
+        self.assertTrue(up.has_negative_karma)
```
bowlofeggs, line 2568: It would be cleaner to split this test condition into another test, rather than to use one test case to try to test two conditions. i.e., let's remove this assertion here and add one more test that has all of the above code but tests the case where a user did not change their mind.

bowlofeggs: It looks like your ```test_has_no_negative_karma``` test below does test this condition. You can drop this assertion to make this test a little bit more readable and cleaner, since the other test already makes sure this works.

bodhi/tests/server/functional/test_updates.py
```diff
@@ -2520,6 +2520,99 @@ def test_edit_testing_update_reset_karma_with_same_tester(self, publish, *args):
...
+        up.comment(self.db, u'LGTM', author=u'mac', karma=1)
+        up = self.db.query(Update).filter_by(title=nvr).one()
+        self.assertFalse(up.has_negative_karma)
+
+        # Karma with no comment
+        up.comment(self.db, u' ', author=u'bowlofeggs', karma=1)
+        up = self.db.query(Update).filter_by(title=nvr).one()
+        self.assertFalse(up.has_negative_karma)
+        self.assertEquals(up.karma, 2)
+
```
bowlofeggs, line 2635: You should add another test case that ensures that anonymous karma is ignored.

bowlofeggs, line 2635: Another good test case would be to test an update that has no feedback.

bodhi/tests/server/functional/test_updates.py
```diff
@@ -2520,6 +2520,99 @@ def test_edit_testing_update_reset_karma_with_same_tester(self, publish, *args):
...
+        upd.comment(self.db, u'Failed to work', author=u'ralph', karma=-1)
+        upd.disable_autopush_for_critical_update()
+        up = self.db.query(Update).filter_by(title=resp.json['title']).one()
+        self.assertTrue(up.has_negative_karma)
+        self.assertEquals(up.karma, -1)
+        self.assertEquals(up.status, UpdateStatus.testing)
+        self.assertEquals(up.request, None)
+        self.assertEquals(up.autokarma, False)
+
```
bowlofeggs, line 2845: You also need tests on `check_karma_thresholds` since the critical logic for this is in that method. Make sure to add both negative and positive tests so it is well covered. Here are some ideas:
1. Test a critical path update that has no negative karma and is ready to autopush.
2. Test a critical path update that has one negative karma, and then four positive karma (reaching a total of 3) with 3 being the stable push threshold. Ensure that the update does not get autopushed.
3. Test a non-critical path update with the same conditions as the previous test I described - this one _should_ get autopushed.
4. Test a critical path update that has autopush turned off, ensuring it doesn't get autopushed.

bowlofeggs:
Hello @trishnaguha! The code looks good! We just need several more tests written as I mentioned in my comments, and we'll be good to go.

bodhi/tests/server/functional/test_updates.py
trishnaguha: Sounds good :-)

trishnaguha:
@bowlofeggs Updated the PR with the testcases :)

bodhi/server/models/models.py
```diff
@@ -1708,6 +1716,18 @@ def last_modified(self):
...
 
     @property
+    def has_negative_karma(self):
+        """Check for negative karma, Returns True if the update has negative karma"""
+        feedback = defaultdict(int)
+        for comment in self.comments:
+            if not comment.anonymous:
+                feedback[comment.user.name] += comment.karma
+        for karma in feedback.values():
+                if karma < 0:
```
bowlofeggs: The indentation here is strange - can you make sure it's 4 spaces?

bodhi/server/models/models.py
```diff
@@ -1649,11 +1649,19 @@ def check_requirements(self, session, settings):
 
         return True, "All checks pass."
 
+    def disable_autopush_for_critical_update(self):
+        """Disable Autopush when critical update receives any negative karma"""
+        log.debug("Disabling Auto Push since the critical update has negative karma")
+        self.autokarma = False
```
bowlofeggs: It looks like this function is only used in one place, which means that we've changed what could have been two lines of code to 1 line of code plus a 4 line function for a total of 5 lines of code. I think this is a case where it makes sense to just put these two lines under the if statement below.

bodhi/tests/server/functional/test_updates.py
```diff
@@ -2520,6 +2520,353 @@ def test_edit_testing_update_reset_karma_with_same_tester(self, publish, *args):
...
+        up.comment(self.db, u'Failed to work', author=u'luke', karma=-1)
+        up = self.db.query(Update).filter_by(title=nvr).one()
+        self.assertTrue(up.has_negative_karma)
+
+    @mock.patch(**mock_valid_requirements)
+    @mock.patch('bodhi.server.notifications.publish')
+    def test_has_negative_karma_with_changed_karma(self, publish, *args):
+        """
+        The test asserts False to has_negative_karma i.e, if update has no negative karma finally
+        Here an user first gives negative karma and gives positive karma later to the update
```
bowlofeggs: s/an user/a user/

bodhi/tests/server/functional/test_updates.py
```diff
@@ -2520,6 +2520,353 @@ def test_edit_testing_update_reset_karma_with_same_tester(self, publish, *args):
...
+
+        up = self.db.query(Update).filter_by(title=nvr).one()
+        up.request = None
+        up.status = UpdateStatus.testing
+        self.db.flush()
+
+        # The user gives negative karma first
+        up.comment(self.db, u'Failed to work', author=u'luke', karma=-1)
+        up = self.db.query(Update).filter_by(title=nvr).one()
+        self.assertTrue(up.has_negative_karma)
```
bowlofeggs, line 2542: Let's add one more test like this one, but the new test should have one user giving positive karma, and one giving negative karma.

bodhi/tests/server/functional/test_updates.py
```diff
@@ -2520,6 +2520,353 @@ def test_edit_testing_update_reset_karma_with_same_tester(self, publish, *args):
...
+        self.db.flush()
+
+        up.comment(self.db, u'Not working', author='me', anonymous=True, karma=-1)
+        up = self.db.query(Update).filter_by(title=nvr).one()
+        self.assertFalse(up.has_negative_karma)
+
+    @mock.patch(**mock_valid_requirements)
+    @mock.patch('bodhi.server.notifications.publish')
+    def test_has_negative_karma_with_no_feedback(self, publish, *args):
+        """This test asserts False to has_negative_karma since there is no feedback for the update"""
```
bowlofeggs: I'd phrase this "This test asserts that has_negative_karma is False when there is no feedback."

bodhi/tests/server/functional/test_updates.py
```diff
@@ -2520,6 +2520,353 @@ def test_edit_testing_update_reset_karma_with_same_tester(self, publish, *args):
...
+        self.assertFalse(up.has_negative_karma)
+        self.assertEquals(up.karma, 2)
+
+    @mock.patch(**mock_valid_requirements)
+    @mock.patch('bodhi.server.notifications.publish')
+    def test_has_negative_karma_with_Anonymous_karma(self, publish, *args):
+        """
+        If update receives negative karma from Anonymous user
+        the test asserts False to it, since anonymous karma is not supposed to affect
+        negative karma count. The test asserts False for the method
```
bowlofeggs: I'd rephrase this "This test asserts that has_negative_karma is False when anonymous users have provided negative feedback."

bodhi/tests/server/functional/test_updates.py
```diff
@@ -2520,6 +2520,353 @@ def test_edit_testing_update_reset_karma_with_same_tester(self, publish, *args):
...
+        up = self.db.query(Update).filter_by(title=nvr).one()
+        self.assertFalse(up.has_negative_karma)
+        self.assertEquals(up.karma, 1)
+
+    @mock.patch(**mock_valid_requirements)
+    @mock.patch('bodhi.server.notifications.publish')
+    def test_has_no_negative_karma(self, publish, *args):
+        """
+        The test asserts False to has_negative_karma i.e,
+        if update has no negative karma
```
bowlofeggs: I'd rephrase this "This test asserts that has_negative_karma is False when an update has no negative karma."

bodhi/tests/server/functional/test_updates.py
bowlofeggs: I'd rephrase this "This test asserts that has_negative_karma returns False when a user posts negative karma and then later posts positive karma."

bodhi/server/models/models.py
```diff
@@ -1708,6 +1716,18 @@ def last_modified(self):
         return possibilities[-1]  # Return the last one
 
     @property
+    def has_negative_karma(self):
+        """Check for negative karma, Returns True if the update has negative karma"""
+        feedback = defaultdict(int)
+        for comment in self.comments:
+            if not comment.anonymous:
+                feedback[comment.user.name] += comment.karma
```
bowlofeggs: It occurred to me that this code will not work correctly if the user gives positive karma and then changes their mind and posts negative karma. This function will count their karma as 0.

bodhi/tests/server/functional/test_updates.py
```diff
@@ -2520,6 +2520,353 @@ def test_edit_testing_update_reset_karma_with_same_tester(self, publish, *args):
...
+        # The user gives negative karma first
+        up.comment(self.db, u'Failed to work', author=u'ralph', karma=-1)
+        up = self.db.query(Update).filter_by(title=nvr).one()
+        self.assertTrue(up.has_negative_karma)
+
+        # The same user gives positive karma later
+        up.comment(self.db, u'wfm', author=u'ralph', karma=1)
+        up = self.db.query(Update).filter_by(title=nvr).one()
+        self.assertFalse(up.has_negative_karma)
+        self.assertEquals(up.karma, 1)
```
bowlofeggs, line 2574: We need another test like this, but with the user giving positive karma and then negative karma (in that order) - `has_negative_karma` should return True in this case. The order matters!

bodhi/tests/server/functional/test_updates.py
```diff
@@ -2520,6 +2520,353 @@ def test_edit_testing_update_reset_karma_with_same_tester(self, publish, *args):
...
+        up.status = UpdateStatus.testing
+        self.db.flush()
+
+        up = self.db.query(Update).filter_by(title=nvr).one()
+        self.assertFalse(up.has_negative_karma)
+
+    @mock.patch(**mock_valid_requirements)
+    @mock.patch('bodhi.server.notifications.publish')
+    def test_disable_autopush_for_critical_updates(self, publish, *args):
+        """Make sure that autopush is disabled if critical update receives any negative karma"""
```
bowlofeggs: s/if critical/if a critical/

bodhi/tests/server/functional/test_updates.py
```diff
@@ -2520,6 +2520,353 @@ def test_edit_testing_update_reset_karma_with_same_tester(self, publish, *args):
...
+        self.assertFalse(up.has_negative_karma)
+
+        up.comment(self.db, u'LGTM', author=u'bowlofeggs', karma=1)
+        up = self.db.query(Update).filter_by(title=resp.json['title']).one()
+        self.assertFalse(up.has_negative_karma)
+        self.assertEquals(up.karma, 2)
+        self.assertEquals(up.autokarma, True)
+
+        up = self.db.query(Update).filter_by(title=resp.json['title']).one()
+        self.assertEquals(up.request, UpdateRequest.stable)
```
bowlofeggs, line 2752: This test should run `check_karma_thresholds` and assert that the correct behaviour happens.

bodhi/tests/server/functional/test_updates.py
```diff
@@ -2520,6 +2520,353 @@ def test_edit_testing_update_reset_karma_with_same_tester(self, publish, *args):
...
+
+        up.comment(self.db, u'LGTM', author=u'puiterwijk', karma=1)
+        up = self.db.query(Update).filter_by(title=resp.json['title']).one()
+        self.assertFalse(up.has_negative_karma)
+
+        self.assertEquals(up.karma, 3)
+        self.assertEquals(up.autokarma, True)
+
+        up = self.db.query(Update).filter_by(title=resp.json['title']).one()
+        self.assertEquals(up.request, UpdateRequest.stable)
```
bowlofeggs, line 2892: This test should run `check_karma_thresholds` and assert that the correct behaviour happens.

bowlofeggs, line 2892: We still need a test for `check_karma_thresholds()` when a critical update does have negative karma - it should not autopush. This was condition 2 from my last review.

bodhi/tests/server/functional/test_updates.py
```diff
@@ -2520,6 +2520,353 @@ def test_edit_testing_update_reset_karma_with_same_tester(self, publish, *args):
...
+
```
bowlofeggs, line 2893: We also still need a test on `check_karma_thresholds()` for condition 4 that I wrote about above, "Test a critical path update that has autopush turned off, ensuring it doesn't get autopushed."

bowlofeggs (changes requested):
There are still a few more tests that we need. Also, I think we have an error in the code for the case where a user gives positive and then negative feedback, so make sure that works correctly.

bodhi/tests/server/functional/test_updates.py
trishnaguha: @bowlofeggs I think this is the test you are looking for https://github.com/fedora-infra/bodhi/pull/894/files#diff-940fb74c59239799d9ed1020a5af9d7cR2777 Here `autokarma_turned_off` means `autopush_turned_off`. This might sound confusing!
I will make sure to change the name of the test to `test_manually_push_critical_update_with_autopush_turned_off` :-).

bodhi/tests/server/functional/test_updates.py
trishnaguha: @bowlofeggs Well this test `test_autopush_non_critical_update_with_negative_karma` is for non critical update where autopush should remain unaltered even it receives negative karma. That's what I am doing :-).

bodhi/tests/server/functional/test_updates.py
trishnaguha: Well we do not need to run the method `check_karma_thresholds`. 
I am attaching the screenshot of debugging for better clarification. 

screenshot from 2016-09-15 18-30-09

bodhi/server/models/models.py
trishnaguha: Oh yeah that was a typo!

trishnaguha:
@bowlofeggs I think we might want to remove `test_disable_autopush_for_critical_update`  if we want to remove the method `disable_autopush_for_critical_update` because this is also going to be tested in `test_manual_push_for_critical_update_with_negative_karma`. So keeping that test sounds kind of duplicate test?

bowlofeggs:
After talking with @trishnaguha on IRC, I came to the realization that I had not been correct about the tests not running `check_karma_thresholds()`. I didn't realize it was part of the karma posting stack and for some reason thought it was a periodic call. Being new to Bodhi I don't yet have the full knowledge of how it does everthing. I apologize for so much confusion!

bodhi/tests/server/functional/test_updates.py
bowlofeggs: I was wrong - it seems that up.comment() calls the function. Apologies!

bodhi/tests/server/functional/test_updates.py
bowlofeggs: I see now, you are correct. Thanks!

bodhi/tests/server/functional/test_updates.py
bowlofeggs: This test has the same user (ralph) giving -1 and then +1, which means that the -1 won't be counted as negative karma, right? Thus, this test actually doesn't have negative karma on the update it is testing.

trishnaguha:
Updated the patch @bowlofeggs  :)

bodhi/server/models/models.py
```diff
@@ -1708,6 +1712,18 @@ def last_modified(self):
         return possibilities[-1]  # Return the last one
 
     @property
+    def has_negative_karma(self):
+        """Check for negative karma, Returns True if the update has negative karma"""
+        feedback = defaultdict(int)
+        for comment in self.comments:
```
bowlofeggs, line 1718: Is self.comments guaranteed to be in chronological order?

bowlofeggs: It does appear that the comments are ordered by timestamp:

https://github.com/fedora-infra/bodhi/blob/3fe3e21946d43d895e38df34b6626b531946676c/bodhi/server/models/models.py#L658-L659

bowlofeggs (approved):
Nice work @trishnaguha, and apologies again for my confusion. These are some nice test cases!

trishnaguha:
@bowlofeggs It's perfectly alright! Thank you :-).

PR #896, opened 2016-08-30
Add a --version option to the CLI.

### LINKED ISSUE #895: Add a --version flag to the bodhi CLI
The bodhi 1 CLI had a --version flag, but the CLI lost that flag during the bodhi 2 dev cycle. As we plan to roll bodhi 2 out to Fedora Rawhide and EL 7, but not to EL 6 or Fedora 23-25, we will need to add a --version flag to the Bodhi 2 CLI so that fedpkg can tell which Bodhi CLI it is interacting with. The fedpkg developers keep fedpkg at the same version across releases, while in Bodhi we'd prefer to offer API stability to our users (with the exception that we will be releasing Bodhi 2 to EPEL 7 soon).

It should be simple to add a --version flag to the new CLI, and then patch fedpkg to use this flag to determine whether it is interacting with Bodhi 1 or Bodhi 2.

bowlofeggs:
closes #895

### DIFF
--- a/bodhi/cli.py
+++ b/bodhi/cli.py
@@ -25,6 +25,7 @@ from fedora.client.bodhi import Bodhi2Client
 
 
 @click.group()
+@click.version_option(message='%(version)s')
 def cli():
     pass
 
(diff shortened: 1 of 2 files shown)


bowlofeggs:
This fedpkg pull request will use this feature to determine which version of bodhi it is dealing with: https://pagure.io/fedpkg/pull-request/40

sayanchowdhury:
:+1:  for me

lmacken:
:+1:

PR #901, opened 2016-09-02
Make sure that an update is obsoleted if it reaches unstable karma threshold in pending state

### LINKED ISSUE #880: Bodhi pushed obsoleted package to testing
Hello,

I originally opened ticket, https://fedorahosted.org/fedora-infrastructure/ticket/5427, but I was redirected here :)

So my issue is: I set negative karma to package (threshold for negative karma was set to -1) during pending state, so I reach obsoleted threshold, but package had been pushed to updates-testing, and obsoleted later (6 seconds later). But because the broken package is in updates-testing it already caused troubles for us.

I'm not sure if this is caused by bodhi too, but this package is in updates-testing for 3 days (I would expect only one)

For reference please see: https://bodhi.fedoraproject.org/updates/FEDORA-2016-86434d618d

trishnaguha:
Fixes #880

### DIFF
--- a/bodhi/server/models/models.py
+++ b/bodhi/server/models/models.py
@@ -1400,6 +1400,18 @@ class Update(Base):
             color = '#00ff00'  # green
         return color
 
+    def obsolete_if_unstable(self, db):
+        """
+        If an update with pending status(autopush enabled) reaches unstable karma
+        threshold, make sure it gets obsoleted.
+        """
+        if self.autokarma and self.status is UpdateStatus.pending and self.request is UpdateRequest.testing \
+                and self.unstable_karma not in (0, None) and self.karma <= self.unstable_karma:
+            log.info("%s has reached unstable karma thresholds" % self.title)
+            self.obsolete(db)
+            flash_log("%s has been obsoleted." % self.title)
+        return
+
     def comment(self, session, text, karma=0, author=None, anonymous=False,
                 karma_critpath=0, bug_feedback=None, testcase_feedback=None,
                 check_karma=True):
@@ -1465,6 +1477,9 @@ class Update(Base):
                         caveats.append({
                             'name': 'karma', 'description': str(e),
                         })
+
+                # Obsolete pending update if it reaches unstable karma threshold
+                self.obsolete_if_unstable(session)
             else:
                 log.debug('Ignoring duplicate %d karma from %s on %s' % (karma, author, self.title))
 
--- a/bodhi/tests/server/functional/test_updates.py
+++ b/bodhi/tests/server/functional/test_updates.py
@@ -1956,6 +1956,108 @@ class TestUpdatesService(bodhi.tests.server.functional.base.BaseWSGICase):
         publish.assert_called_with(
                 topic='update.request.revoke', msg=mock.ANY)
 
+    @mock.patch(**mock_valid_requirements)
+    @mock.patch('bodhi.server.notifications.publish')
+    def test_obsolete_if_unstable_with_autopush_enabled_when_pending(self, publish, *args):
+        """
+        Send update to obsolete state if it reaches unstable karma on
+        pending state where request is testing when Autopush is enabled. Make sure that it
+        does not go to update-testing state.
+        """
+        nvr = u'bodhi-2.0.0-2.fc17'
+        args = self.get_update(nvr)
+        args['autokarma'] = True
+        args['stable_karma'] = 1
+        args['unstable_karma'] = -1
+
+        resp = self.app.post_json('/updates/', args)
+        up = Update.get(nvr, self.db)
+        up.status = UpdateStatus.pending
+        up.request = UpdateRequest.testing
+        up.comment(self.db, u'Failed to work', author=u'ralph', karma=-1)
+        self.db.flush()
+
+        up = self.db.query(Update).filter_by(title=nvr).one()
+        self.assertEquals(up.karma, -1)
+        self.assertEquals(up.status, UpdateStatus.obsolete)
+        self.assertEquals(up.request, None)
+
+    @mock.patch(**mock_valid_requirements)
+    @mock.patch('bodhi.server.notifications.publish')
+    def test_obsolete_if_unstable_with_autopush_disabled_when_pending(self, publish, *args):
+        """
+        Don't automatically send update to obsolete state if it reaches unstable karma on
+        pending state when Autopush is disabled.
+        """
+        nvr = u'bodhi-2.0.0-2.fc17'
+        args = self.get_update(nvr)
+        args['autokarma'] = False
+        args['stable_karma'] = 1
+        args['unstable_karma'] = -1
+
+        resp = self.app.post_json('/updates/', args)
+        up = Update.get(nvr, self.db)
+        up.status = UpdateStatus.pending
+        up.request = UpdateRequest.testing
+        up.comment(self.db, u'Failed to work', author=u'ralph', karma=-1)
+        self.db.flush()
+
+        up = self.db.query(Update).filter_by(title=nvr).one()
+        self.assertEquals(up.karma, -1)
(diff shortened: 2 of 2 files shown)


bodhi/models/models.py
```diff
@@ -1077,6 +1077,16 @@ def set_request(self, db, action, username):
                 update=self, agent=username))
             return
 
+        # If an update with pending status receives negative karma, make sure it gets obsoleted
+        if self.status is UpdateStatus.pending and self.request is UpdateRequest.testing \
+            and self.karma < 0:
```
bowlofeggs: I'm thinking that we should only do this if autokarma is turned on. Do you agree? If so, rather than testing for negative karma perhaps we could test to see if the negative karma threshold is reached.

bowlofeggs:
Can you rebase your commit?

bodhi/tests/functional/test_updates.py
```diff
@@ -1955,6 +1955,33 @@ def test_revoke_action_for_testing_request(self, publish, *args):
         publish.assert_called_with(
                 topic='update.request.revoke', msg=mock.ANY)
 
+    @mock.patch(**mock_valid_requirements)
+    @mock.patch('bodhi.notifications.publish')
+    def test_obsolete_update_on_pending_state(self, publish, *args):
```
bowlofeggs: I'm not familiar with the rest of the tests in this area, so can you please ensure that these five conditions are covered by tests that are either new or existing? There are 4 conditions that will be tested in the new if statement, so we could conceivably test 5 conditions. The first is a positive case (basically, what you have here but modified to set the autokarma and reach the negative threshold), and the other four just ensure the negative cases:
1. The update is pending, is marked for testing, has autokarma true, and has reached the negative karma threshold. This update should get obsoleted.
2. All of the first test, but the update isn't pending.
3. All of the first test, but the update isn't marked for testing (this one should probably also be obsoleted, but by another function in the code. I also bet there's already a test for this.)
4. All of the above, but autokarma is off.
5. All of the above, but the karma is not beyond the thresholds.

You don't necessarily have to write the above tests if they already exist, but please ensure that they do exist.

bodhi/models/models.py
trishnaguha: @bowlofeggs Yeah I agree with `autokarma turned on`.
Okay one more query regarding negative karma. Do we want to check negative karma from recent comment like we did `has_negative_karma` or just checking `unstable_karma(negative_karma) threshold` would be choice here?

bowlofeggs: @trishnaguha I think just checking that method should be fine. Does that method not account for users who changed their minds? If it doesn't, let's file a separate issue (and separate PR) for that and we can fix it later.

trishnaguha:
Updated the patch. I will make sure that we have enough testcases for the patch.

trishnaguha:
@bowlofeggs I have updated the test set. Can you please explain about the test you mentioned at no. 3?

bowlofeggs:
On Wed, 2016-09-21 at 02:09 -0700, Trishna Guha wrote:

> @bowlofeggs I have updated the test set. Can you please explain about
> the test you mentioned at no. 3?

Basically that we should test behavior when `self.request is not
UpdateRequest.testing`, and make sure the right thing happens.

Overall, I want to ensure we hit all the bits that are anded in the if
statement.

trishnaguha:
@bowlofeggs Sure I got you :).

trishnaguha:
@bowlofeggs I just debugged that When we are not making request to testing on pending status it by default gets set to testing. Does it make sense to make request to `None` for this testcase? if that is the case, karma is not going to affect the update unless the request is testing. the status still remains as `pending` because of request being `None`.

bodhi/server/models/models.py
```diff
@@ -1400,6 +1400,18 @@ def get_pushed_color(self):
             color = '#00ff00'  # green
         return color
 
+    def obsolete_if_unstable(self, db):
+        """
+        If an update with pending status(autopush enabled) reaches unstable karma
+        threshold, make sure it gets obsoleted.
+        """
+        if self.autokarma and self.status is UpdateStatus.pending and self.request is UpdateRequest.testing \
+            and self.unstable_karma not in (0, None) and self.karma <= self.unstable_karma:
```
bowlofeggs: Can you add one more level of indentation to this line so that it's visually separated from the code below?

bodhi/tests/server/functional/test_updates.py
```diff
@@ -1956,6 +1956,109 @@ def test_revoke_action_for_testing_request(self, publish, *args):
...
+        up = Update.get(nvr, self.db)
+        up.status = UpdateStatus.pending
+        up.request = UpdateRequest.testing
+        up.comment(self.db, u'Failed to work', author=u'ralph', karma=-1)
+        self.db.flush()
+
+        up = self.db.query(Update).filter_by(title=nvr).one()
+        self.assertEquals(up.karma, -1)
+        self.assertEquals(up.status, UpdateStatus.pending)
+        self.assertNotEquals(up.status, UpdateStatus.obsolete)
```
bowlofeggs: I think the first assertion implies the second assertion, so if it were me I would leave only the assertEqual and drop the assertNotEquals. That said, there's also obviously no direct harm in leaving this assertion, but less lines of code is preferable in my book ☺ You can leave it as is if you want, just a suggestion.

bodhi/tests/server/functional/test_updates.py
```diff
@@ -1956,6 +1956,109 @@ def test_revoke_action_for_testing_request(self, publish, *args):
...
+        resp = self.app.post_json('/updates/', args)
+        up = Update.get(nvr, self.db)
+        up.status = UpdateStatus.pending
+        up.request = UpdateRequest.testing
+        up.comment(self.db, u'Failed to work', author=u'ralph', karma=-1)
+        self.db.flush()
+
+        up = self.db.query(Update).filter_by(title=nvr).one()
+        self.assertEquals(up.karma, -1)
+        self.assertNotEquals(up.status, UpdateStatus.obsolete)
```
bowlofeggs: Same idea here, I think it's slightly cleaner to have one less line of code and keep only the assertEquals instead of this assertNotEquals. Again, it's preference so you can leave it if you want.

bodhi/tests/server/functional/test_updates.py
```diff
@@ -1956,6 +1956,109 @@ def test_revoke_action_for_testing_request(self, publish, *args):
...
+        self.assertEquals(up.status, UpdateStatus.pending)
+        self.assertEquals(up.request, UpdateRequest.testing)
+
+    @mock.patch(**mock_valid_requirements)
+    @mock.patch('bodhi.server.notifications.publish')
+    def test_obsolete_if_unstable_with_autopush_enabled_when_testing(self, publish, *args):
+        """
+        Send update to obsolete state if it reaches unstable karma threshold on
+        testing state when Autopush is enabled.
```
bowlofeggs: It seems that the main difference between this test and the first one is that this one has up.request set to stable and the other has up.request set to testing. Can you clarify that difference in the docblocks of both tests?

bowlofeggs (changes requested):
I think just two more changes and we'll be good to go:
1. Indent that one line so it is visually separated from the code below it.
2. Clarify the docblocks on the two very similar tests.

Nice work!

bodhi/tests/server/functional/test_updates.py
trishnaguha: - The first test makes sure that the update doesn't go to testing(should be obsoleted) if it receives negative karma on the pending state where request is testing when autopush is enabled.
- This test makes sure that the update doesn't got to stable(should be obsoleted) if it receives negative karma on the testing state where request is stable when autopush is enabled.

Yes I will update the PR with that :).

bowlofeggs (approved):

trishnaguha:
Thanks :-).

PR #903, opened 2016-09-03
Remove tests from setup.py

### LINKED ISSUE #899: setup.py install includes the unit tests
Currently running setup.py install will install the unit tests. We should adjust setup.py to exclude the unit tests, or we could move the unit tests to be outside of the code directory so setup.py doesn't find them during install. Either way is fine.

jarifibrahim:
This fixes https://github.com/fedora-infra/bodhi/issues/899


setup.py
```diff
@@ -89,17 +89,10 @@
...
       include_package_data=True,
       zip_safe=False,
       install_requires = requires,
-      tests_require = [
-          'nose',
-          'nose-cov',
-          'webtest',
-          'mock'
-      ],
-      test_suite="nose.collector",
```
bowlofeggs, line 102: Hello @jarifibrahim! Thanks for the PR. I think we want to keep the above two elements so that the tests can be run with `python setup.py test`. It's my fault for not being clear in the issue I filed, but to be more explicit, what I wanted was to make it so `python setup.py install` doesn't install the tests. I think what you did on line 92 will accomplish that goal.

Can you add the tests_require and test_suite back? I'm very sorry for not making that clear in my issue, and thanks for the patch!

jarifibrahim: @bowlofeggs No problem. I've added `test_suite` and `tests_require` back.

setup.py
```diff
@@ -89,7 +89,7 @@
       author_email='',
       url='',
       keywords='web fedora pyramid',
-      packages=find_packages(),
+      packages=find_packages(exclude=['tests']),
```
bowlofeggs, line 92: Thanks @jarifibrahim!

This doesn't seem to be working. You can test by running `python setup.py install` - you'll see that `/usr/lib/python2.7/site-packages/bodhi-2.1.9-py2.7.egg/bodhi/tests` is getting installed. I think this will work though:

```
packages=find_packages(exclude=('bodhi.tests', 'bodhi.tests.*')),
```

Let me know when you want me to take another look, and thanks for the contribution!

bowlofeggs:
I needed to mess with the setup.py files to split up the client and server, so I went ahead and fixed this in #915

I'll update that commit message to reference this issue. Thanks for taking a shot at it @jarifibrahim, your contribution is still appreciated!

PR #904, opened 2016-09-03
Cleanly separate client and server code.

### LINKED ISSUE #900: Reorganize locations of client and server code
Since the client and server code typically gets split into two different RPMs during packaging, it would be easier to separate them if they were more cleanly separated in our git repository. Something simple like moving them into bodhi/client/ and bodhi/server/ would make it a little easier to keep them separate in the spec file. Alternatively, we could create two separate Python packages for the client at server at the top level in this repository, or even start a new git repository for the bodhi client.

bowlofeggs:
Prior to this commit, it was impossible to install the Bodhi client
separately from the server, as the bodhi.**init** module contained
server code. This commit creates new bodhi.client and bodhi.server
modules so there is a clear separation between them. This will help
future developers keep them apart mentally, and it will make
packaging Bodhi simpler.

fixes #900


sayanchowdhury:
Code looks good to me.

pypingou:
:+1: as well

trishnaguha:
After setting this up on my local system, 
I am getting these errors on starting the localserver and running the testsuite.
It is failing to locate the template files.
Local server start: https://paste.fedoraproject.org/424437/
Testsuite: https://paste.fedoraproject.org/424440/41709614/

sayanchowdhury:
@trishnaguha: You need copy the new development.ini.example into development.ini

Now, the template files are in server/templates/

trishnaguha:
Ah I forgot to copy the new development.example.ini 
Sorry for the noise!
Thanks @sayanchowdhury

PR #907, opened 2016-09-05
Always use utf8 for mashing threads

puiterwijk:
This change makes sure that the encoding during mashing is utf8 by default.
This fixes running on Fedora, since Fedora's systemd on server starts processes
with an empty LC_ALL, defaulting to LC_ALL=C, which breaks almost all unicode.

### DIFF
--- a/bodhi/server/consumers/masher.py
+++ b/bodhi/server/consumers/masher.py
@@ -29,6 +29,7 @@ import hashlib
 import threading
 import fedmsg.consumers
 from datetime import datetime
+import sys
 
 from collections import defaultdict
 from pyramid.paster import get_appsettings
@@ -248,6 +249,9 @@ class MasherThread(threading.Thread):
             'completed_repos': []
         }
         self.success = False
+
+        reload(sys)
+        sys.setdefaultencoding('utf8')
 
     def run(self):
         try:


bodhi/consumers/masher.py
```diff
@@ -249,6 +250,9 @@ def __init__(self, release, request, updates, agent,
         }
         self.success = False
 
+        reload(sys)
```
bowlofeggs: This seems a bit unconventional. It seems that you have this call here because `setdefaultencoding()` gets removed from `sys` after the site module runs. However, the API docs for that function warn that it isn't meant to be used by mere mortals:

https://docs.python.org/2.7/library/sys.html#sys.setdefaultencoding

Being new to Bodhi, I'm not sure how the masher is run. Is it launched by fedmsgs? If so, what process listens to that fedmsg? It might be better to have that process set a locale in the environment for this process.

puiterwijk:
The masher is run in fedmsg-hub, which gets started by systemd.
Explanation for this is in the commit message:

> This change makes sure that the encoding during mashing is utf8 by default.
> This fixes running on Fedora, since Fedora's systemd on server starts processes
> with an empty LC_ALL, defaulting to LC_ALL=C, which breaks almost all unicode.

bowlofeggs:
@puiterwijk and I discussed this over IRC yesterday. I'm going to try to fix this by making sure that Bodhi is being given a UTF-8 locale by the environment, rather than having the code deal with not having a proper locale set by the environment. Thanks @puiterwijk!

PR #920, opened 2016-09-10
Include the bootstrap symlink in the server manifest.

### LINKED ISSUE #919: The bootstrap symlink is not getting packaged by setup.py
The 2.1.9 beta is currently failing to load bootstrap because the git repo symlink is not making it into the downstream RPM. Adding bootstrap to the SERVER_MANIFEST.in should fix this problem.

bowlofeggs:
fixes #919


sayanchowdhury:
Verified :+1:

bowlofeggs:
Thanks @sayanchowdhury!

PR #921, opened 2016-09-11
Autofocus the packages input field in New Update form

### LINKED ISSUE #876: [webgui] "New update" form usability
It seems that currently the form is convenient just for the "clicky" people, but those preferring keyboard should be served as well.
Now I am talking about following use case when filling out the mentioned form:
- fill the package name under Packages label and do not select the autocompletion item
- press `TAB` so as to move to the next item

Observed: no completion of possible builds is triggered

Expected: it is triggered just as when the package is selected from autocompletion with a mouse

k3rn3l3rr0r:
(fixes #876)

### DIFF
--- a/bodhi/server/static/js/update_form.js
+++ b/bodhi/server/static/js/update_form.js
@@ -242,5 +242,9 @@ $(document).ready(function() {
 
     // Lastly show the main form
     $("#new-update-form").removeClass('hidden');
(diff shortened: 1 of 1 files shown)


pypingou:
Cool, not tested but looks good to me :)

bowlofeggs:
I tested it and it works. Nice change, thanks!

PR #924, opened 2016-09-12
Write release notes for the upcoming 2.2.0 release.

bowlofeggs:
The next release does contain features that are new since the 2.1.8 release, so this commit also sets the version to 2.2.0 in the `setup.py` and `docs/conf.py` files.


puiterwijk:
:+1:

docs/release_notes.rst
```diff
@@ -0,0 +1,71 @@
...
+Bodhi 2.2.0 is a security and feature release, with a few bug fixes as well.
+
+
+Security
+^^^^^^^^
+
+This update addresses `CVE-2016-1000008 <https://github.com/fedora-infra/bodhi/pull/857>`_ by
+disallowing the re-use of solved captchas. Additionally, the captcha is
+`warped <https://github.com/fedora-infra/bodhi/commit/f0122855>`_ to make it more difficult to
+solve through automation.
```
bowlofeggs: Add a credit here for Patrick

https://github.com/distributedweaknessfiling/DWF-Database/blob/master/DWF-Database-2016.csv#L17

docs/release_notes.rst
```diff
@@ -0,0 +1,71 @@
...
+
+Features
+^^^^^^^^
+
+* Bodhi's ``approve_testing.py`` script will now comment on updates when they have reached a stable
+  karma threshold
+  (`5b0d1c7c <https://github.com/fedora-infra/bodhi/commit/5b0d1c7c>`_).
+* The web interface now displays a push to stable button when the karma reaches the right level when
+  autokarma is disabled
+  (`b65a6ccf <https://github.com/fedora-infra/bodhi/commit/b65a6ccf>`_).
```
trishnaguha: This commit fixes https://github.com/fedora-infra/bodhi/issues/772 and https://github.com/fedora-infra/bodhi/issues/796

bowlofeggs: @trishnaguha both of those issues still show as being open, which was why I didn't list them by issue. Should they be marked closed?

docs/release_notes.rst
bowlofeggs: I've pushed a new commit that credits Patrick here.

docs/release_notes.rst
trishnaguha: @bowlofeggs Yes I think we can close them since they are fixed by this PR #821

bowlofeggs: Thanks @trishnaguha, I'll change the release notes to reference those.

PR #929, opened 2016-09-14
Update email templates to point users to dnf instead of yum

### LINKED ISSUE #927: Update emails still include boilerplate info for "yum" rather than "dnf"
Scott Dowdle [reported in Red Hat's Bugzilla](https://bugzilla.redhat.com/show_bug.cgi?id=1375701) that Fedora update e-mails still reference `yum` instead of `dnf`:

```
I wasn't sure which Component to file against as I did not notice email mentioned anywhere.

Here's a sample email I got today to illustrate the issue:

- - - - - start - - - - 

--------------------------------------------------------------------------------
Fedora Update Notification
FEDORA-2016-5e40d80b1c
2016-09-13 18:03:12.623848
--------------------------------------------------------------------------------

Name        : livecd-tools
Product     : Fedora 25
Version     : 23.4
Release     : 1.fc25
URL         : https://github.com/rhinstaller/livecd-tools
Summary     : Tools for building live CDs
Description :
Tools for generating live CDs on Fedora based systems including
derived distributions such as RHEL, CentOS and others. See
http://fedoraproject.org/wiki/FedoraLiveCD for more details.

--------------------------------------------------------------------------------
Update Information:

- Version 23.4 (bcl) - yuminst.LiveCDYum.runInstall: import yum.Errors
(sandro.bonazzola) - Wrap all parted calls in LC_ALL=C (#1350253) (bcl) - Fix
calling livecd-iso-to-pxeboot with a full path (cadegenn) - Stop truncating
/etc/resolv.conf in SELinux module -- fixes #31 (torrancew) - litp: Copy
ldlinux.c32 to tftpboot for PXE live setup. (bcl) - Always use rsync to copy
Packages (#1343645) (bcl) - Fix extended regular expression [0-9] (fgrose) -
Enable loop device as installation target (fgrose)
--------------------------------------------------------------------------------
References:

  [ 1 ] Bug #1350253 - livecd-iso-to-disk fails using a non English locale
        https://bugzilla.redhat.com/show_bug.cgi?id=1350253
  [ 2 ] Bug #1343645 - livecd-iso-to-disk states "/Packages found, will copy source packages to target", but does not
        https://bugzilla.redhat.com/show_bug.cgi?id=1343645
--------------------------------------------------------------------------------

This update can be installed with the "yum" update program. Use
su -c 'yum update livecd-tools' at the command line.
For more information, refer to "Managing Software with yum",
available at https://docs.fedoraproject.org/yum/.

All packages are signed with the Fedora Project GPG key.  More details on the
GPG keys used by the Fedora Project can be found at
https://fedoraproject.org/keys
--------------------------------------------------------------------------------

- - - - - end - - - - -
```

We should fix this, but we should also make sure that EPEL <= 7 updates continue to reference `yum` when we fix it.

k3rn3l3rr0r:
fixes #927 
- update template texts to not use outdated informations
- add legacy template pointing to yum and use it with updates for EPEL <= 7 distributions

### DIFF
--- a/bodhi/server/mail.py
+++ b/bodhi/server/mail.py
@@ -252,12 +252,12 @@ Description :
 
 --------------------------------------------------------------------------------
 %(notes)s%(changelog)s%(references)s
-This update can be installed with the "yum" update program. Use
-su -c 'yum%(yum_repository)s update %(name)s' at the command line.
-For more information, refer to "Managing Software with yum",
-available at https://docs.fedoraproject.org/yum/.
+This update can be installed with the "dnf" update program. Use
+su -c 'dnf%(yum_repository)s upgrade %(name)s' at the command line.
+For more information, refer to the dnf documentation available at
+http://dnf.readthedocs.io/en/latest/command_ref.html#upgrade-command-label
 
-All packages are signed with the Fedora Project GPG key.  More details on the
+All packages are signed with the Fedora Project GPG key. More details on the
 GPG keys used by the Fedora Project can be found at
 https://fedoraproject.org/keys
 --------------------------------------------------------------------------------
@@ -281,10 +281,41 @@ Description :
 
 --------------------------------------------------------------------------------
 %(notes)s%(changelog)s%(references)s
+This update can be installed with the "dnf" update programs.  Use
+su -c 'dnf%(yum_repository)s upgrade %(name)s' at the command line.
+For more information, refer to the dnf documentation available at
+http://dnf.readthedocs.io/en/latest/command_ref.html#upgrade-command-label
+
+All packages are signed with the Fedora EPEL GPG key. More details on the
+GPG keys used by the Fedora Project can be found at
+https://fedoraproject.org/keys
+--------------------------------------------------------------------------------
+"""
+
+# message template for legacy systems, still pointing users to use "yum"
+fedora_epel_legacy_errata_template = u"""\
+--------------------------------------------------------------------------------
+Fedora EPEL%(testing)s Update Notification
+%(updateid)s
+%(date)s
+--------------------------------------------------------------------------------
+
+Name        : %(name)s
+Product     : %(product)s
+Version     : %(version)s
+Release     : %(release)s
+URL         : %(url)s
+Summary     : %(summary)s
+Description :
+%(description)s
+
+--------------------------------------------------------------------------------
+%(notes)s%(changelog)s%(references)s
 This update can be installed with the "yum" update programs.  Use
 su -c 'yum%(yum_repository)s update %(name)s' at the command line.
-For more information, refer to "Managing Software with yum",
-available at https://docs.fedoraproject.org/yum/.
+For more information, refer to "YUM", available at
+https://access.redhat.com/documentation/en-US/Red_Hat_Enterprise_Linux/7\
+/html/System_Administrators_Guide/ch-yum.html
 
 All packages are signed with the Fedora EPEL GPG key.  More details on the
 GPG keys used by the Fedora Project can be found at
--- a/bodhi/server/models/models.py
+++ b/bodhi/server/models/models.py
@@ -1247,7 +1247,13 @@ class Update(Base):
             mailinglist = config.get('%s_announce_list' % release_name)
         elif self.status is UpdateStatus.testing:
             mailinglist = config.get('%s_test_announce_list' % release_name)
-        templatetype = '%s_errata_template' % release_name
+        
+        # switch email template to legacy if update aims to EPEL <= 7
+        if release_name == 'fedora_epel' and self.release.version_int <= 7:
+            templatetype = '%s_errata_legacy_template' % release_name
+        else:
+            templatetype = '%s_errata_template' % release_name
+
         if mailinglist:
             for subject, body in mail.get_template(self, templatetype):
                 mail.send_mail(sender, mailinglist, subject, body)
@@ -1256,6 +1262,7 @@ class Update(Base):
         else:
             log.error("Cannot find mailing list address for update notice")
             log.error("release_name = %r", release_name)
+
 
     def get_url(self):
         """ Return the relative URL to this update """
--- a/bodhi/tests/server/models/test_models.py
+++ b/bodhi/tests/server/models/test_models.py
@@ -598,6 +598,44 @@ class TestUpdate(ModelTest):
         for value in mail.MESSAGES.values():
             value['body'] % value['fields']('guest', self.obj)
 
+    @mock.patch('bodhi.server.mail.get_template')
+    def test_send_update_notice_message_template(self, get_template):
+        """Ensure update message template reflects aimed distribution"""
+        update = self.obj
+        update.status = UpdateStatus.stable
+        update.send_update_notice()
+        get_template.assert_called_with(update, u'fedora_errata_template')
+
+        update = self.get_update(name=u'TurboGears-3.1-1.el7')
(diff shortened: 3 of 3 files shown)


k3rn3l3rr0r:
All test were OK. But I can barely see any email testcases there...

trishnaguha:
LGTM :+1: @k3rn3l3rr0r Can you squash the commits?

k3rn3l3rr0r:
@trishnaguha Squashed...

bodhi/server/models/models.py
```diff
@@ -1247,7 +1247,13 @@ def send_update_notice(self):
             mailinglist = config.get('%s_announce_list' % release_name)
         elif self.status is UpdateStatus.testing:
             mailinglist = config.get('%s_test_announce_list' % release_name)
-        templatetype = '%s_errata_template' % release_name
+        
+        # switch email template to legacy if update aims to EPEL <= 7
+        if release_name == 'fedora_epel' and self.release.version <= 7:
+            templatetype = '%s_errata_legacy_template' % release_name
+        else:
+            templatetype = '%s_errata_template' % release_name
```
bowlofeggs, line 1255: It does seem that we lack tests on this method, but we do have a test for the class this method is part of here:

https://github.com/fedora-infra/bodhi/blob/dd6add3cfa1392135eb972416f2b892152147850/bodhi/tests/server/models/test_models.py#L172

Could you add unit tests that hit these two paths on the `send_update_notice()` method? You don't have to worry about the rest of the method unless you just want to get some testing done while you are in there, but getting some coverage at least on this new if statement would be good.

bodhi/server/mail.py
```diff
@@ -252,12 +252,12 @@
 
 --------------------------------------------------------------------------------
 %(notes)s%(changelog)s%(references)s
-This update can be installed with the "yum" update program. Use
-su -c 'yum%(yum_repository)s update %(name)s' at the command line.
-For more information, refer to "Managing Software with yum",
-available at https://docs.fedoraproject.org/yum/.
+This update can be installed with the "dnf" update program. Use
+su -c 'dnf%(yum_repository)s update %(name)s' at the command line.
```
bowlofeggs: s/update/upgrade/

bodhi/server/mail.py
```diff
@@ -252,12 +252,12 @@
...
+For more information, refer to "Package Management",
+available at https://docs.fedoraproject.org
```
bowlofeggs: We could probably change this to be more specific, like:

```
For more information, refer to the dnf documentation available at
http://dnf.readthedocs.io/en/latest/command_ref.html#upgrade-command-label
```

bodhi/server/mail.py
```diff
@@ -281,10 +281,40 @@
...
+%(description)s
+
+--------------------------------------------------------------------------------
+%(notes)s%(changelog)s%(references)s
 This update can be installed with the "yum" update programs.  Use
 su -c 'yum%(yum_repository)s update %(name)s' at the command line.
-For more information, refer to "Managing Software with yum",
-available at https://docs.fedoraproject.org/yum/.
+For more information, refer to "Package Management",
+available at https://docs.fedoraproject.org
```
bowlofeggs: We should probably link to documentation about Red Hat Enterprise Linux here, since that URL isn't about enterprise linux and yum isn't covered in the recent versions of dnf documentation there. I found this link with duckduckgo.com but it doesn't seem to respond right now:

https://access.redhat.com/documentation/en-US/Red_Hat_Enterprise_Linux/6/html/Deployment_Guide/ch-yum.html

Presumably there is one for Red Hat Enterprise Linux 7 as well.

bodhi/server/mail.py
```diff
@@ -281,10 +281,40 @@
...
+
+All packages are signed with the Fedora EPEL GPG key. More details on the
+GPG keys used by the Fedora Project can be found at
+https://fedoraproject.org/keys
+--------------------------------------------------------------------------------
+"""
+
+# message template for legacy systems, still pointing users to use "yum"
+fedora_epel_legacy_errata_template = u"""\
```
bowlofeggs, line 296: I don't think we need to add a "legacy" template just yet, since nothing is currently using the `fedora_epel_errata_template` defined above. I assume this was to think ahead for EL8, but I think we can just cross that bridge when we get to it. For now, let's just keep less text around and use the already existing template name.

bowlofeggs: Actually, on second thought I see that your if statement below works with EL 8 already. I'm convinced that what I just wrote is no good, so please do keep these ☺

bodhi/server/mail.py
```diff
@@ -281,10 +281,40 @@
 
 --------------------------------------------------------------------------------
 %(notes)s%(changelog)s%(references)s
+This update can be installed with the "dnf" update programs.  Use
+su -c 'dnf%(yum_repository)s update %(name)s' at the command line.
+For more information, refer to "Package Management",
+available at https://docs.fedoraproject.org
```
bowlofeggs: Let's link the dnf documentation here as well.

bodhi/server/mail.py
```diff
@@ -281,10 +281,40 @@
 
 --------------------------------------------------------------------------------
 %(notes)s%(changelog)s%(references)s
+This update can be installed with the "dnf" update programs.  Use
+su -c 'dnf%(yum_repository)s update %(name)s' at the command line.
```
bowlofeggs: s/update/upgrade/

bowlofeggs (changes requested):
@k3rn3l3rr0r this look pretty good! If you can change just a few things:
1. Can you squash this down to one commit?
2. Can you add three unit tests - one to hit the Fedora path, one for EL < 8, and one for EL == 8?
3. Can you change the dnf links to point at dnf docs, and the yum links to point at EL docs about yum? I struggled to find links that seem to currently be working for yum, so if you need help finding them just ping me and I'll put more time into finding them.
4. Change documentation to say "dnf upgrade" instead of "dnf update".

Thanks for putting in so many commits lately, I really appreciate it!

k3rn3l3rr0r:
@bowlofeggs I tried to address all your requests in 1c1f2ab. Waiting for further comments. Thanks.

bodhi/server/models/models.py
```diff
@@ -1257,6 +1263,7 @@ def send_update_notice(self):
             log.error("Cannot find mailing list address for update notice")
             log.error("release_name = %r", release_name)
 
+
```
bowlofeggs, line 1266: PEP-8 dictates that there should be 1 space here instead of 2 since this separation is between two methods and not between two top-level objects.

bodhi/tests/server/models/test_models.py
```diff
@@ -598,6 +598,44 @@ def test_expand_messages(self):
...
+                          testing_tag=u'dist-8E-epel-testing',
+                          candidate_tag=u'dist-8E-epel-testing-candidate',
+                          pending_testing_tag=u'dist-8E-epel-testing-pending',
+                          pending_stable_tag=u'dist-8E-epel-pending',
+                          override_tag=u'dist-8E-epel-override',
+                          branch=u'el8', version=u'8')
+        update.release = release
+        update.status = UpdateStatus.stable
+        update.send_update_notice()
+        get_template.assert_called_with(update, u'fedora_epel_errata_template')
```
bowlofeggs, line 637: It would be good to split this test into three separate test methods, since they are testing three different conditions.

bowlofeggs (approved):
Thanks for the contribution! I'm going to go ahead and merge this as-is, and I'll submit a new PR on top of yours with the two changes I suggested (so you don't have to do that). Thanks so much!

PR #930, opened 2016-09-14
Add pending_signing_tag

puiterwijk:
The idea is that instead of into pending_testing_tag, we tag into pending_signing_tag,
and then robosignatory will pick them up, sign them, and move them to pending_testing_tag.


bowlofeggs:
Thanks @puiterwijk - this idea makes sense to me. Could you add some test coverage for the new lines? Also, it seems that we should have a data migration to add this new tag to existing repos.

puiterwijk:
I was not intending to do migration inside bodhi. We can just run the standard sigulsign_unsigned for one last time when we enable this.
That means that the same "everything in pending_testing_tag is signed" statement holds both before and after enabling this, without bodhi having to care abut it either way.

bodhi/server/models/models.py
```diff
@@ -1129,7 +1133,7 @@ def set_request(self, db, action, username):
         # Add the appropriate 'pending' koji tag to this update, so tools like
         # AutoQA can mash repositories of them for testing.
         if action is UpdateRequest.testing:
-            self.add_tag(self.release.pending_testing_tag)
+            self.add_tag(self.release.pending_signing_tag)
```
bowlofeggs, line 1136: What process will add the pending_testing_tag when it is time for that?

bowlofeggs (approved):

bodhi/server/models/models.py
puiterwijk: That will be robosignatory's duty.

PR #931, opened 2016-09-15
Send fedmsg notifications during ostree composes

puiterwijk:
This also uses the ref and commitid returned by fedora-infra/fedmsg-atomic-composer#7

### DIFF
--- a/bodhi/server/consumers/masher.py
+++ b/bodhi/server/consumers/masher.py
@@ -958,12 +958,24 @@ class MasherThread(threading.Thread):
                 release['repos']['updates'] = mash_path
 
             # Compose the tree, and raise an exception upon failure
+            notifications.publish(topic="ostree.compose.start",
+                                  msg=dict(tag=tag),
+                                  force=True)
             result = composer.compose(release)
             if result['result'] != 'success':
                 self.log.error(result)
+                notifications.publish(topic="ostree.compose.fail",
+                                      msg=dict(tag=tag),
+                                      force=True)
                 raise Exception('%s atomic compose failed' % tag)
             else:
                 self.log.info('%s atomic tree compose successful', tag)
+                notifications.publish(
+                    topic="ostree.compose.finish",
+                    msg=dict(tag=tag,
+                             ref=result.get('ref'),
+                             commitid=results.get('commitid')),
+                    force=True)
 
 
 class MashThread(threading.Thread):


bodhi/server/consumers/masher.py
```diff
@@ -958,12 +958,24 @@ def compose_atomic_trees(self):
                 release['repos']['updates'] = mash_path
 
             # Compose the tree, and raise an exception upon failure
+            notifications.publish(topic="ostree.start",
+                                  msg=dict(tag=tag),
+                                  force=True)
             result = composer.compose(release)
             if result['result'] != 'success':
                 self.log.error(result)
+                notifications.publish(topic="ostree.start",
```
bowlofeggs: Should this be ostree.failed or something?

puiterwijk: Haha, yes. That had been my intention.

bodhi/server/consumers/masher.py
```diff
@@ -958,12 +958,24 @@ def compose_atomic_trees(self):
...
+                                      force=True)
                 raise Exception('%s atomic compose failed' % tag)
             else:
                 self.log.info('%s atomic tree compose successful', tag)
+                notifications.publish(topic="ostree.compose.finish",
+                                      msg=dict(tag=tag,
+                                               ref=result.get('ref'),
+                                               commitid=
+                                                    result.get('commitid')),
+                                      force=True)
```
bowlofeggs: Maybe we could rewrap the above so it's a little cleaner, like

```
                notifications.publish(
                    topic="ostree.compose.finish",
                    msg=dict(
                        tag=tag,
                        ref=result.get('ref'),
                        commitid=result.get('commitid')),
                    force=True)
```

You can leave it as-is if you prefer, just a suggestion.

puiterwijk: Done

bodhi/server/consumers/masher.py
```diff
@@ -958,12 +958,24 @@ def compose_atomic_trees(self):
...
                 raise Exception('%s atomic compose failed' % tag)
             else:
                 self.log.info('%s atomic tree compose successful', tag)
+                notifications.publish(
+                    topic="ostree.compose.finish",
+                    msg=dict(tag=tag,
+                             ref=result.get('ref'),
+                             commitid=results.get('commitid')),
+                     force=True)
```
bowlofeggs: It looks like this `force` isn't lined up with the `msg` above it, maybe one too many spaces?

bowlofeggs (changes requested):
Two things to fix before merging:
1. Let's fix that indentation on `force`.
2. Since it is programmatically important that these messages get sent correctly for signing to work with it, we should add two tests to ensure they do get sent during mash. One test for the success case, and the other for the failure case.

bodhi/server/consumers/masher.py
pypingou: though a little risky, we could make it something like `topic="ostree.%s" % result['result']` no?

Risky as in: for every new result, we'll need to adjust fedmsg_meta.

puiterwijk: I would prefer just keeping with the ==success !=success cases, as those are the only two that the atomic composer is aware of.

bowlofeggs:
Since @puiterwijk has been so helpful to me over the past week, I'm going to attempt to write the tests for him in a separate pull request and I'll merge this now. Thanks @puiterwijk!

PR #935, opened 2016-09-16
Clear unchecked Candidate Builds and Related Bugs on package change

### LINKED ISSUE #897: Duplicate Candidate Builds and Related Bugs
While creating a new update, I enter fedpkg in Packages, then Bodhi find out candidate builds and related bugs and put them into respective list. Next, I don't leave the page, and change the package name but still the same name fedpkg, Bodhi tries to find out builds and bugs again, and fill them in the list again, as shown below.

screenshot from 2016-09-01 15-29-13

Expected result is to prevent from filling in duplicate candidate builds and related bugs.

k3rn3l3rr0r:
(closes #897)
- compare ID's, while appending the items to get rid of duplicates in the lists
- preserve checked items
- manually added entries are checked in default, so those are preserved in the lists too (as long as those stay checked)

### DIFF
--- a/bodhi/server/static/js/update_form.js
+++ b/bodhi/server/static/js/update_form.js
@@ -163,15 +163,28 @@ $(document).ready(function() {
     $('#packages-search input.typeahead').on('typeahead:selected', function (e, datum) {
         $("#candidate-checkboxes").prepend("<img class='spinner' src='static/img/spinner.gif'>")
         $("#bugs-checkboxes").prepend("<img class='spinner' src='static/img/spinner.gif'>")
+        // Get a list of currently checked items
+        var checked_bug_ids = [];
+        $("#bugs-checkboxes input:checkbox:checked").each(function(){
+            checked_bug_ids.push(parseInt($(this).val()));
+        });
+        var checked_candidate_ids = [];
+        $("#candidate-checkboxes input:checkbox:checked").each(function(){
+            checked_candidate_ids.push(parseInt($(this).attr('data-build-id')));
+        });
         // Get the candidate builds
         $.ajax({
             url: 'latest_candidates',
             data: $.param({package: datum.name}),
             success: function(builds) {
                 $("#candidate-checkboxes .spinner").remove();
+                $("#candidate-checkboxes input:checkbox:not(:checked)").parents("div.checkbox").remove();
                 if (builds.length == 0) {return candidate_error(datum.name);}
                 $.each(builds, function(i, build) {
-                    add_build_checkbox(build.nvr, build.id, false);
+                    // Insert the checkbox only if this ID is not already listed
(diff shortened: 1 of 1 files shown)


k3rn3l3rr0r:
@bowlofeggs Would be nice to test in stg... Btw. how can I convince my local bodhi dev instance to return some latest_candidates results for my testing purposes? :)

bowlofeggs:
Hey @k3rn3l3rr0r!

The easiest way to test Bodhi like this is to point it at the [staging koji instance](http://koji.stg.fedoraproject.org/koji/) and do a fake build there. You will need to do a fake build on a package you own in staging pkgdb so you have the right ACLs.

I'll check this PR out more soon (today is the weekend so what am I doing on github‽ ☺), but thanks for sending this PR. I'm planning to release Bodhi 2.2.0 this week, so once that is deployed to production we'll start staging the next release and we'll be able to test this if it gets in in time.

Sorry I haven't reviewed your several PRs yet, but thank you for sending them. It was an unusually busy week, but I promise I will get to them soon.

bowlofeggs:
@k3rn3l3rr0r,

If you set these lines in your development.ini to be like these, I think that'll get it connected to the stg koji:

```
##
## Buildsystem settings
##

# What buildsystem do we want to use?  For development, we'll use a fake
# buildsystem that always does what we tell it to do.  For production, we'll
# want to use 'koji'.
buildsystem = koji

# Koji's XML-RPC hub
koji_hub = https://koji.stg.fedoraproject.org/kojihub

# Root url of the Koji instance to point to. No trailing slash
koji_url = http://koji.stg.fedoraproject.org
```

You will need a legitimate `.fedora.cert` in your development environment, which means you'll need to inject one into the VM if you use Vagrant. You will also need the `.fedora-server-ca.cert` file.

bowlofeggs:
@k3rn3l3rr0r After forever of messing around in my browser's debug console, I finally figured out what else you need to get the browser to be OK with the latest_updates API call. I had to change `base_address` from `http://0.0.0.0:6543` to `http://localhost:6543`. This address was causing the AJAX calls to go to 0.0.0.0 instead of localhost, and the browser was not happy with that.

bowlofeggs:
I filed https://github.com/fedora-infra/bodhi/pull/940 to change the default `base_address` in `development.ini.example` so this won't trip up other developers.

bowlofeggs:
I also filed https://github.com/fedora-infra/bodhi/pull/941 to document how to connect to the staging Koji during development.

bowlofeggs:
Nice work @k3rn3l3rr0r! After figuring out how to get my dev box connected to staging koji, I was finally able to test your code and it does seem to do the job well. I tried various combinations of checking boxes, changing the search, re-searching the same search, and all did what I would expect to happen. Thanks for the patch!

PR #940, opened 2016-09-20
Change the development base_address to be localhost.

bowlofeggs:
Most developers are likely to develop by pointing their browsers
at localhost, rather than 0.0.0.0. This commit sets the
base_address to use the name "localhost" by default for new
developers.

### DIFF
--- a/bodhi/tests/server/functional/test_updates.py
+++ b/bodhi/tests/server/functional/test_updates.py
@@ -18,6 +18,7 @@ import copy
 import textwrap
 import time
 import mock
+import urlparse
 
 from mock import ANY
 from nose.tools import eq_
@@ -433,7 +434,6 @@ class TestUpdatesService(bodhi.tests.server.functional.base.BaseWSGICase):
         self.assertEquals(len(body['updates']), 1)
 
         alias = u'FEDORA-%s-a3bbe1a8f2' % YEAR
-        baseurl = 'http://0.0.0.0:6543'
 
         up = body['updates'][0]
         self.assertEquals(up['title'], u'bodhi-2.0-1.fc17')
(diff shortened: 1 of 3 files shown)


sayanchowdhury:
:+1:  for me

bowlofeggs:
I need to hold off on merging this - it seems that four unit tests explicitly depend on this setting having 0.0.0.0 in it. We should make those tests dynamically use the setting from development.ini.

development.ini.example
```diff
@@ -123,7 +123,7 @@ fedora_epel_testing_master_repomd = https://download.fedoraproject.org/pub/epel/
 
 ## The base url of this application
 ## Used as the <base/> tag in the master template.
-base_address = http://0.0.0.0:6543/
+base_address = http://localhost:6543/
```
puiterwijk, line 126: Is there any way to instead of changing the binding, change just the line it's printing?
The problem is that people might deploy it on a remote box and want to test, which would be possible with 0.0.0.0 but not localtime.

puiterwijk (changes requested):

development.ini.example
bowlofeggs: @puiterwijk On a local box, 0.0.0.0 doesn't work for lots of the AJAX calls. Since developing on a local system is the common case and since this is the .example file, I think we really do want to change this to "localhost". Keep in mind that the readme directs developers to use Vagrant, so this example will not work for an out-of-the-box setup unless we accept this change.

Developers who want to work against a remote system are free to change their development.ini to be however they please, as the development.ini is in the gitignore. This file is just here to get you started.

puiterwijk (approved):

puiterwijk (approved):

bodhi/tests/server/functional/test_updates.py
```diff
@@ -453,7 +453,7 @@ def test_list_updates(self):
         self.assertEquals(up['locked'], True)
         self.assertEquals(up['alias'], alias)
         self.assertEquals(up['karma'], 1)
-        self.assertEquals(up['url'], '%s/updates/%s' % (baseurl, alias))
+        self.assertEquals(up['url'], (urlparse.urljoin(config['base_address'], '/updates/%s' % alias)))
```
pypingou, line 456: I've run a few times into issues using `urlparse.urljoin()` so I now try to avoid it in favor of a construct such as:

`
'%s/updates/%s' % (baseurl.rstrip('/'), alias))
`
Close to the original one but ensure there are no `//`.

bowlofeggs: Yeah urljoin can be tricky at times. Since this is just a test and it is working as-is, I'm inclined to leave it this way for now. If I have troubles in the future, I'll switch it to be as you suggested. Thanks!

PR #943, opened 2016-09-21
Only put the comment through markdown if there is a comment

puiterwijk:
Because sometimes, the contents are None (null in the database), which
breaks the markdown processing.

### DIFF
--- a/bodhi/server/templates/fragments.html
+++ b/bodhi/server/templates/fragments.html
@@ -34,7 +34,9 @@
 
       <div class="panel-body">
         <div class="col-md-8">
+          %if comment.text:
           ${util.markup(comment.text) | n}
+          $endif
         </div>
         <div class="col-md-4 text-right">
           % if comment.karma:


bowlofeggs:
I think we will fix this by writing a migration in https://github.com/fedora-infra/bodhi/issues/949 instead of merging this patch. However, I will carry this patch in our downstream RPM in the meantime. Thanks @puiterwijk!

bodhi/server/templates/fragments.html
```diff
@@ -34,7 +34,9 @@
 
       <div class="panel-body">
         <div class="col-md-8">
+          %if comment.text:
           ${util.markup(comment.text) | n}
+          $endif
```
bowlofeggs, line 39: I think it's suppsoed to be `% endif` ☺

trishnaguha: Also there should be a space after `%` :)

Issue #952, opened 2016-09-21
Bodhi 2.2.0 does not display the lock banner on locked updates during pushes anymore


bowlofeggs:
These two updates should both display that they are locked right now because there is an active push for them in progress, but neither of them displays the locked banner:

https://bodhi.fedoraproject.org/updates/FEDORA-2016-a6867138e2
https://bodhi.fedoraproject.org/updates/FEDORA-2016-821d2ec8cd

bowlofeggs:
@puiterwijk suggested that the problem is that the `update.locked = True` isn't followed by `session.add(update)` at https://github.com/fedora-infra/bodhi/blob/2.2.0/bodhi/server/consumers/masher.py#L220

trishnaguha:
This also means we don't have test for this :(.

puiterwijk:
Turns out that we WERE locking them, but the interface didn't get the massive banner because of date_locked not being set.
See #959 and its commit messages for the explanation.

bowlofeggs:
Thanks @puiterwijk! I'll update the text here to reflect that.

PR #959, opened 2016-09-22
Fix displaying the locked banner by fixing date_locked

puiterwijk:


puiterwijk:
Please read the commit messages for explanations.

puiterwijk:
The reason this was not a problem with 2.0: it did not have the date_locked attribute.

bodhi/server/consumers/masher.py
```diff
@@ -405,23 +403,9 @@ def unlock_updates(self):
         self.log.debug('Unlocking updates')
         for update in self.updates:
             update.locked = False
+            update.locked_date = None
```
trishnaguha: It should be `update.date_locked` instead?

puiterwijk: Correct, fixed that. Thanks.

bodhi/server/consumers/masher.py
```diff
@@ -406,23 +403,9 @@ def unlock_updates(self):
...
-    def locked_date_for_update(self):
-        """ Return the date & time when an update has been locked """
-        for update in self.updates:
-            # Return datetime when an update is locked and datetime is
-            # not yet assigned to the column date_locked
-            if update.locked and not update.date_locked:
-                update.date_locked = datetime.utcnow()
-                self.db.commit()
-                self.db.refresh(update)
-                return update.date_locked
```
bowlofeggs, line 419: This seems like a pretty strange pattern - to loop over updates until you hit one that's locked without a date, set its date, and… return? What about the rest of the updates in `self.updates`?

bowlofeggs:
@puiterwijk These commits seem like they should be squashed together. The first one shouldn't be applied alone since the second one has the code to clear the date_locked.

Additionally, can you throw a `fixes #952` in your commit message so it automatically ties to the bug? I updated the bug to be about this instead of about the session stuff.

bowlofeggs (approved):
Thanks so much!

bowlofeggs:
I will write unit tests for this later. For now I am happy to go ahead and merge this since we need it ASAP.

PR #967, opened 2016-09-23
Sometimes we have no mash, so dont wait for mash_thread if we dont

puiterwijk:
If self.state['completed_repos'] is True, we don't kick off a mash thread, and
def mash() returns nothing (None).


sayanchowdhury:
:+1:  for me

bowlofeggs (approved):
Thanks @puiterwijk!

bowlofeggs:
I have cherry picked this into the 2.2 branch as well.

Issue #949, opened 2016-09-21
Update comments should not allow NULL values in the comment text


bowlofeggs:
We had a problem in production where update comments were NULL, causing some template rendering to fail. There's not a good reason for update comments ot be NULL vs. empty string, so let's change that column to disallow NULL and write a migration to apply that change on upgrade. The migration will need to first change all NULL values to empty string.

bowlofeggs:
I believe this update might have a NULL comment from Bodhi:

https://bodhi.fedoraproject.org/updates/FEDORA-2016-0ab60daf6d

PR #971, opened 2016-09-23
Add a test to assert that Comment.text cannot be NULL.

bowlofeggs:
re #949

### DIFF
--- a/bodhi/tests/server/models/test_models.py
+++ b/bodhi/tests/server/models/test_models.py
@@ -17,6 +17,7 @@
 """Test suite for the Bodhi models"""
 
 import time
+import unittest
 import cornice
 import mock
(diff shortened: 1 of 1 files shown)


bowlofeggs:
I spent a huge amount of time trying to make the `TestComment` class inherit from the same parent class as the other classes in this module, but it was causing conflicts with objects being created by the other classes. I suspect that the base test class doesn't do a very thorough job of cleaning up after each test case. Unfortunately, I do not have much time to sort this problem out right now and I am also pretty new to SQLAlchemy so I am slow at it. I filed #972 to track improving this new class later, so that I can move on to other urgent issues for now.

puiterwijk (approved):

PR #974, opened 2016-09-24
973

bowlofeggs:


bowlofeggs:
The first of these commits will be cherry-picked to the 2.2 branch by itself, that's why there are two commits instead of one.

puiterwijk (approved):

bowlofeggs:
I have cherry picked the first commit to the 2.2 branch.

PR #976, opened 2016-09-24
Set autopush to True when stable_karma is not NULL

### LINKED ISSUE #950: Autopush seems to be broken
Usually when an update reaches the stable karma threshold, there will be a comment from the bodhi user stating that the update has reached this status and that it will be pushed to stable. For example, this update from bodhi-2.0-212 exibited that behavior:

https://bodhi.fedoraproject.org/updates/FEDORA-2016-ffc86f2b30

Since deploying 2.2.0 to production, it has been noted that updates no longer seem to get that comment from Bodhi when they reach stable karma. For example, see this update in stg:

https://bodhi.stg.fedoraproject.org/updates/FEDORA-2016-4d0198119a

My comment should have caused bodhi to submit it for stable automatically, but it did not. Here is one more example of a non-critical path update that should be autopushed:

https://bodhi.stg.fedoraproject.org/updates/FEDORA-2016-93c58fa1b3

bowlofeggs:
fixes #950

### DIFF
--- a/alembic/versions/5110dfc1a01a_set_autokarma_to_true_for_updates_.py
+++ b/alembic/versions/5110dfc1a01a_set_autokarma_to_true_for_updates_.py
@@ -1,0 +1,32 @@
+"""Set NULL autokarma to True for Updates that have non-NULL stable karma thresholds.
+
+Revision ID: 5110dfc1a01a
+Revises: 37f38ddc4c8d
+Create Date: 2016-09-24 02:53:42.025785
+
+"""
+
+from alembic import op
+import sqlalchemy as sa
+
+
+# revision identifiers, used by Alembic.
+revision = '5110dfc1a01a'
+down_revision = '37f38ddc4c8d'
+
+
+def upgrade():
+    """Set NULL autokarma to True for Updates that have non-NULL stable karma thresholds."""
+    # Build a fake mini version of the updates table so we can form an UPDATE statement.
+    updates = sa.sql.table('updates', sa.sql.column('autokarma', sa.Boolean),
+                           sa.sql.column('stable_karma', sa.Integer))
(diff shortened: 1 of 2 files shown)


bowlofeggs:
The fist of these two commits will need to be cherry picked to the 2.2 branch.

alembic/versions/5110dfc1a01a_set_autokarma_to_true_for_updates_.py
```diff
@@ -0,0 +1,32 @@
...
+def upgrade():
+    """Set NULL autokarma to True for Updates that have non-NULL stable karma thresholds."""
+    # Build a fake mini version of the updates table so we can form an UPDATE statement.
+    updates = sa.sql.table('updates', sa.sql.column('autokarma', sa.Boolean),
+                           sa.sql.column('stable_karma', sa.Integer))
+    # Set autokarma to True if there is a stable threshold set and autokarma is None.
+    op.execute(
+        updates.update().where(
+            updates.c.stable_karma != None).where(updates.c.autokarma == None).values(
+                {'autokarma': True}))
```
puiterwijk, line 27: Why this awkward indentation?

How about something like this?

```
    op.execute(
         updates.update().where(
             updates.c.stable_karma != None).where(
             updates.c.autokarma == None).values(
             {'autokarma': True}))

```

puiterwijk (approved):
Single idea, but looks good.

alembic/versions/5110dfc1a01a_set_autokarma_to_true_for_updates_.py
bowlofeggs: I actually like the indentation the way I did it more because it clearly indicates which bit is part of which method. The rest of the code isn't part of the first `where` call, but is part of other calls that are tacked on later.

PR #985, opened 2016-09-28
Only initialize the Bugzilla connection when needed

puiterwijk:


puiterwijk:
This should prevent the entire web interface of going down when RHBZ is unavailable.
This will most likely also speed up a lot of the requests.

pypingou:
Looks fine to me

bowlofeggs:
It seems that we are missing test coverage on this entire module, but I won't require you to start coverage there. This change looks good to me.

bowlofeggs (approved):

PR #991, opened 2016-10-01
metrics: 'releases' always must be sorted in all loops.

jcfaracco:
Depending on the tag that you are using, the data may be inconsistent
with the labels. So, all releases lists needs to be sorted in all loops.

Example:
Data stored in my database:
- Fedora 23: 3 bugfixes
- Fedora 24: 2 bugfixes
- Fedora 22: 5 bugfixes

Sorted label array:
- Fedora 22
- Fedora 23
- Fedora 24

Chart result (wrong results):
- Fedora 22: 3 bugfixes
- Fedora 23: 2 bugfixes
- Fedora 24: 5 bugfixes

After this patch:
- Fedora 22: 5 bugfixes
- Fedora 23: 3 bugfixes
- Fedora 24: 2 bugfixes

Signed-off-by: Julio Faracco jfaracco@br.ibm.com

### DIFF
--- a/bodhi/server/views/metrics.py
+++ b/bodhi/server/views/metrics.py
@@ -23,8 +23,10 @@ def compute_ticks_and_data(db, releases, update_types):
     """ Return the data and ticks to make the stats graph. """
     data, ticks = [], []
 
-    for i, release in enumerate(sorted(releases, cmp=lambda x, y:
-            cmp(int(x.version_int), int(y.version_int)))):
+    releases = sorted(releases, cmp=lambda x, y:
+            cmp(int(x.version_int), int(y.version_int)))
+
+    for i, release in enumerate(releases):
         ticks.append([i, release.name])
 
     for update_type, label in update_types.items():


jcfaracco:
Hi guys,

I'm using a bodhi fork (customized) in my server here to track packages from RHEL.
We have some generic tags when the package is available for all systems.
Example: EL6 packages are compatible with EL6.6, EL6.7 and EL6.8.

I'm not using a stable tag. I'm using the 'develop' branch.
I opened the metrics page and it was a completely mess.
So, I checked the source code and I noticed that the second loop was not using the 'releases' sorted array. It was producing inconsistent data and labels.

bodhi/server/views/metrics.py
```diff
@@ -23,8 +23,10 @@ def compute_ticks_and_data(db, releases, update_types):
     """ Return the data and ticks to make the stats graph. """
     data, ticks = [], []
 
-    for i, release in enumerate(sorted(releases, cmp=lambda x, y:
-            cmp(int(x.version_int), int(y.version_int)))):
+    releases = sorted(releases, cmp=lambda x, y:
+            cmp(int(x.version_int), int(y.version_int)))
```
bowlofeggs, line 27: At first I didn't see the difference, but now I see that there's another call to `enumerate(releases)` a few lines down, and that line didn't have the sorting that we see here.

I think this change makes sense, thanks for the contribution!

bowlofeggs (approved):
I've checked out the metrics page with this change in place, and everything looks reasonable to me. Thanks!

bowlofeggs:
Hello @jcfaracco, Would you mind if I blogged about your usage of Bodhi outside of the Fedora project? I think many people inside Fedora would be pleasantly surprised to know that there are other users of Bodhi than just Fedora. Thanks!

PR #992, opened 2016-10-03
Karma comments should not set request to None when autopush is off.

### LINKED ISSUE #989: Further comments on pushed updates with autopush disabled remove stable requests
Andre Robatino [reported](https://bugzilla.redhat.com/show_bug.cgi?id=1380622) that Bodhi update stable requests are being reset by subsequent comments. For example, [this kernel update](https://bodhi.fedoraproject.org/updates/FEDORA-2016-7179af3ac1) was set to go to stable by labbott about 15 hours ago at the time of this writing. About 8 hours after that, another user helpfully gave the update a +1 and that comment caused the request for stable to be removed. The update now shows as status testing with no request.

I have not yet investigated the code to determine where the mistake in logic is.
# Known workaround

If you experience this issue on one of your updates, a possible workaround is to edit the update and turn autokarma on. After doing this, any additional +1 should trigger the update to push. You should also be able to push the update once it reaches 7 days.

bowlofeggs:
fixes #989

### DIFF
--- a/bodhi/server/models/models.py
+++ b/bodhi/server/models/models.py
@@ -1697,7 +1697,6 @@ class Update(Base):
                         # Add the 'testing_approval_msg_based_on_karma' message now
                         log.info("%s update has reached the stable karma threshold and can be pushed to "\
                         "stable now if the maintainer wishes" % self.title)
-                        self.request = None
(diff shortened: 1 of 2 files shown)


pypingou:
:+1: for me

PR #1002, opened 2016-10-05
Hard-depend on fedmsg-atomic-composer.

bowlofeggs:


jeremycline (commented):
Looks good :shipit:

I assume people will figure out they need to update their Vagrantfile fairly easily and there's no need to announce it anywere.

bowlofeggs:
@jeremycline That is unfortunately my plan. There's not really a bodhi mailing list. I will announce this in #fedora-apps, but there's not really a better way to announce it that I know of. Of course, `git bisect` would also help people find out that this commit broken them, and then they'd theoretically see the Vagrantfile.example change as well…

PR #1003, opened 2016-10-05
PEP-8 bodhi.server.models.models.

bowlofeggs:
This commit also adds a new test for enforcing that the models.py
file stays compliant with PEP-8 with two exceptions: 100 character
line length and ignoring one particular flake8 error that is very
common for sqlalchemy projects. The plan is to add new modules to
this test as the code base is converted to be PEP-8 compliant over
time, until the entire code base is enforced.

### DIFF
--- a/bodhi/server/models/models.py
+++ b/bodhi/server/models/models.py
@@ -189,13 +189,12 @@ class BodhiBase(object):
         return new, same, removed
 
 
-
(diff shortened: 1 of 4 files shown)


trishnaguha:
LGTM :+1:

jeremycline (approved):
Looks good, :shipit:

PR #1010, opened 2016-10-07
Display buttons on override template to users who belong to packager/provenpackager group 

### LINKED ISSUE #1008: Submit and Expire buttons for Override should be displayed only for the one who has permission
`Submit` and `Expire` buttons on the `overrides` should be displayed only to the one who has `commit access/ permission` for the same, which also means that the buttons will not be displayed if someone is not logged in.

screenshot from 2016-10-06 15-42-57

trishnaguha:
Fixes #1008 
Here is the screenshot:

screenshot from 2016-10-07 11-28-45

### DIFF
--- a/bodhi/server/templates/override.html
+++ b/bodhi/server/templates/override.html
@@ -138,6 +138,7 @@
 
     <input type="hidden" name="csrf_token" value="${request.session.get_csrf_token()}"/>
 
+    %if request.user and ('provenpackager' or 'packager' in request.user.groups):
     <div class="row">
       <div class="col-md-12">
         <div class="panel panel-default">
@@ -157,6 +158,7 @@
         </div>
       </div>
     </div> <!-- end row -->
+    %endif
 
   </div>
 </div>
--- a/bodhi/tests/server/functional/test_overrides.py
+++ b/bodhi/tests/server/functional/test_overrides.py
@@ -17,7 +17,7 @@ from datetime import datetime, timedelta
 import mock
 
 import bodhi.tests.server.functional.base
-from bodhi.server.models import (Build, Package, Release, User)
+from bodhi.server.models import (Build, Package, Release, User, Group)
 
 
 class TestOverridesService(bodhi.tests.server.functional.base.BaseWSGICase):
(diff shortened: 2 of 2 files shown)


bodhi/server/templates/override.html
```diff
@@ -138,6 +138,7 @@ <h2 class="pull-left">New Override</span>
 
     <input type="hidden" name="csrf_token" value="${request.session.get_csrf_token()}"/>
 
+    %if request.user and request.user.name is override.submitter.name:
```
bowlofeggs: This check is more restrictive than the handler for these buttons, as the handler allows anyone in the packager group:

https://github.com/fedora-infra/bodhi/blob/develop/bodhi/server/services/overrides.py#L156

trishnaguha: Ah I missed it. Will update the PR. Thanks @bowlofeggs for pointing it out.

trishnaguha:
@bowlofeggs Could you please review now? :)

bodhi/server/templates/override.html
```diff
@@ -138,6 +138,7 @@ <h2 class="pull-left">New Override</span>
 
     <input type="hidden" name="csrf_token" value="${request.session.get_csrf_token()}"/>
 
+    %if request.user and 'mandatory_packager_groups' in request.user.groups:
```
bowlofeggs: This does not appear to work in my development checkout. I am logged in and I am in the correct group, but I am not shown these buttons.

bowlofeggs (changes requested):
I don't believe this is working correctly. Additionally, we will need tests to go with this change. Going ahead and writing the the tests now will be useful to prove to yourself that this is working as intended, and is often how I work when I'm working on a well defined problem like this. It's super handy ☺

bodhi/server/templates/override.html
```diff
@@ -138,6 +138,7 @@ <h2 class="pull-left">New Override</span>
 
     <input type="hidden" name="csrf_token" value="${request.session.get_csrf_token()}"/>
 
+    %if request.user and ('provenpackager' or 'packager' in request.user.groups):
```
bowlofeggs, line 141: We should not hardcode the group names, as [they are not hardcoded on the view that checks whether the acl is valid or not](https://github.com/fedora-infra/bodhi/blob/2.3.1/bodhi/server/security.py#L39-L42):

```
    groups = request.registry.settings['mandatory_packager_groups'].split()
    return [
        (Allow, 'group:' + group, ALL_PERMISSIONS) for group in groups
    ] + [DENY_ALL]
```

bodhi/tests/server/functional/test_overrides.py
```diff
@@ -180,6 +180,79 @@ def test_list_overrides_by_unexisting_username(self):
...
+
+        expiration_date = datetime.utcnow() + timedelta(days=1)
+        data = {'nvr': build.nvr, 'notes': u'blah blah blah',
+                'expiration_date': expiration_date, 'csrf_token': self.get_csrf_token()}
+        resp = self.app.post('/overrides/', data)
+        o = resp.json_body
+        self.assertEquals(o['build_id'], build.id)
+        self.assertEquals(o['notes'], 'blah blah blah')
+        self.assertEquals(o['expiration_date'], expiration_date.strftime("%Y-%m-%d %H:%M:%S"))
+        self.assertEquals(o['expired_date'], None)
```
bowlofeggs, line 209: Though these assertions are not harmful, they do complicate this test and aren't necessary for the purpose of the test. I prefer to keep test code as short as possible so that it's clearer what the purpose of the test is. These assertions should be present on another test that asserts that creating new updates work correctly instead of here.

bodhi/tests/server/functional/test_overrides.py
```diff
@@ -180,6 +180,79 @@ def test_list_overrides_by_unexisting_username(self):
...
```
bowlofeggs, line 246: Same here - these assertions should be in a test that checks that overrides can be created. They make this test longer and less obvious.

bowlofeggs (changes requested):
We need the code that displays the buttons to follow the same algorithm as the code that allows or disallows the view. That code is dynamic and is based on the settings.

Additionally, we should have a negative test - one that sets up a user who isn't in the approved groups - that user shouldn't see the buttons.

bowlofeggs:
@trishnaguha This PR has been sitting for some time. Do you still want to work on it, or should we close it?

trishnaguha:
@bowlofeggs Sorry, I couldn't manage to get time to work on this. I'm going to close this PR now since this is really not an urgent patch. If I can come with a solution I will reopen it or may be a separate new PR. I really appreciate you taking the time to review the PR :-).

PR #1019, opened 2016-10-10
approve_testing: use required testing time when update can be pushed to stable

### LINKED ISSUE #1017: Reminder email script adds comments that include the number of elapsed days
When #1009 was fixed, I started getting Bodhi reminder emails for updates that passed the minimum number of days in testing some time ago.

Consider this update (chosen more or less at random):

https://bodhi.fedoraproject.org/updates/FEDORA-2016-c22fffe3da

It's an F24 update, so only needed to spend 7 days in testing. It entered testing on 2016-09-29, so the 7 days ended on 2016-10-06, and as of now (2016-10-08) it's been in testing for 9 days.

Due to #1009, the reminder comment was only added this morning (2016-10-08). Notice that it says "9 days", whereas "7 days" would have been more correct (that's the requirement for F24).

In case the script ever stops working again, and needs to 'catch up' with old updates, I think the code needs to change from:

```
text = config.get('testing_approval_msg') % update.days_in_testing
```

to:

```
text = config.get('testing_approval_msg') % update.release.mandatory_days_in_testing
```

so that the comment reflects the requirements for the Fedora release, rather than the time so far that the update has spent in testing.

richardfearn:
The script currently uses the number of days that the update has been in testing. This can lead to comments like the one seen here:

  https://bodhi.fedoraproject.org/updates/FEDORA-2016-c22fffe3da

...where the comment says the update "has reached 9 days in testing", because the script ran 9 days after the update entered testing. For this (F24) update it's better to say it's reached 7 days in testing, since that's the mandatory testing time for this release.

This fixes #1017.

### DIFF
--- a/bodhi/server/scripts/approve_testing.py
+++ b/bodhi/server/scripts/approve_testing.py
@@ -62,7 +62,7 @@ def main(argv=sys.argv):
                 continue
             if update.meets_testing_requirements:
                 print('%s now meets testing requirements' % update.title)
-                text = config.get('testing_approval_msg') % update.days_in_testing
+                text = config.get('testing_approval_msg') % update.release.mandatory_days_in_testing
                 update.comment(db, text, author='bodhi')
 
             # Approval message when testing based on karma threshold


bowlofeggs (approved):
Thanks for your contribution!

PR #1021, opened 2016-10-11
Remove bodhi.server.models.DBSession from development.ini.example.

bowlofeggs:
There is no bodhi.server.models.DBSession, and its presence in the
example development.ini file causes pshell to fail.


pypingou:
:+1:

PR #1022, opened 2016-10-11
Allow shelldb to detect its settings from argv and document it.

bowlofeggs:
This commit adjusts the small shelldb.py utility to be a little
more flexible and allow it to detect which settings files should be
used by inspecting sys.argv for any arguments that end in .ini.
Though this is a "simpleton" approach and normally I'd want
something more robust, it does work and this is just a development
tool.

### DIFF
--- a/tools/shelldb.py
+++ b/tools/shelldb.py
@@ -8,13 +8,24 @@ from pyramid.paster import get_appsettings, setup_logging
 from sqlalchemy import engine_from_config
 from sqlalchemy.orm import scoped_session, sessionmaker
 from zope.sqlalchemy import ZopeTransactionExtension
+import sys
 import transaction
-config_uri = '/etc/bodhi/production.ini'
+
+
+config_uri = None
+for arg in sys.argv:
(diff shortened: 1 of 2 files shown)


pypingou:
:+1: for me

tools/shelldb.py
```diff
@@ -8,14 +8,25 @@
...
 from sqlalchemy.orm import scoped_session, sessionmaker
 from zope.sqlalchemy import ZopeTransactionExtension
+import sys
 import transaction
-config_uri = '/etc/bodhi/production.ini'
+
+
+config_uri = None
+for arg in sys.argv:
+    if '.ini' in arg:
```
sayanchowdhury: Better to check `arg.endswith('.ini')`

sayanchowdhury (changes requested):

bowlofeggs:
@sayanchowdhury I've made the change you requested.

pypingou:
Still :+1: for me

sayanchowdhury:
:+1:

PR #1024, opened 2016-10-12
Add a docblock to a method in bodhi.server.models.enum.

bowlofeggs:


pypingou:
:+1: although it doesn't describe what the method do :)

bowlofeggs:
@pypingou To be honest, that's because I still don't know what it does. I just wanted to add the information that I _do_ know accurately (I used ipdb to get in there and inspect to find out).

pypingou:
^_^ fair, still :+1: to merge, we can always improve later when we figure things out :)

PR #1025, opened 2016-10-12
Document that libvirt must be installed for Vagrant to work.

bowlofeggs:
I had incorrectly assumed that vagrant-libvirt depended on libvirt
and it does not, so developers need to install libvirt as well.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -10,10 +10,14 @@ Vagrant
 =======
 
 `Vagrant <https://www.vagrantup.com/>`_ allows contributors to get quickly up and running with a
-Bodhi development environment by automatically configuring a virtual machine. To get started, simply
+Bodhi development environment by automatically configuring a virtual machine. Before you get
+started, ensure that your host machine has virtualization extensions enabled in its BIOS so the
+guest doesn't go slower than molasses. To get started, simply
 use these commands::
 
-    $ sudo dnf install ansible vagrant-libvirt vagrant-sshfs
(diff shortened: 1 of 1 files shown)


pypingou:
:+1:

README.rst
```diff
@@ -10,10 +10,13 @@ Vagrant
...
 `Vagrant <https://www.vagrantup.com/>`_ allows contributors to get quickly up and running with a
-Bodhi development environment by automatically configuring a virtual machine. To get started, simply
+Bodhi development environment by automatically configuring a virtual machine. Before you get
+started, ensure that your host machine has virtualization extensions enabled in its BIOS so the
+guest doesn't go slower than molasses. To get started, simply
 use these commands::
 
-    $ sudo dnf install ansible vagrant-libvirt vagrant-sshfs
+    $ sudo dnf install ansible libvirt vagrant-libvirt vagrant-sshfs
+    $ for a in {enable,start}; do sudo systemctl $a libvirtd; done;
```
pypingou: Although I read this, I would have kept it two lines for the sake of clarity

bowlofeggs: I like that it's one line to copy/paste instead of two, but I'll fix it as per your preference.

bowlofeggs:
@pypingou I've made the change you requested. Good to merge?

pypingou:
Looks good to me :)

README.rst
pypingou: You're also upstream so you can prefer this way :)

PR #1030, opened 2016-10-14
Send one comment when non-autokarma updates hit the threshold.

### LINKED ISSUE #1009: Users are receiving repeat e-mails every 6 hours
@kparal reported than [an Epiphany update](https://bodhi.fedoraproject.org/updates/FEDORA-2016-f4b5897686) has been e-mailing him every 6 hours with the push to stable message, like this:

```
This update has reached 11 days in testing and can be pushed to stable now if the maintainer wishes
```

He received one at 2016-10-04T08:00, 2016-10-04T14:00, and 2016-10-04T20:00, for example.

We should ensure that these messages only get sent once.

bowlofeggs:
The approve_testing script did not work properly with the new
autokarma feature. For updates that had autokarma set to False, the
approve_testing script would add a bodhi comment every time it ran
to say that the update could be pushed to stable. This commit
adjusts that script, the Update.meets_testing_requirements method,
and the Update.met_testing_requirements method so that only one
comment is left.

Additionally, it was noted that there was not testing coverage on
the approve_testing script so a new test module was created to
cover the script. In order to save some testing effort, a new base
test class was created to perform some common functions.
Additionally, this commit adds several new tests on
Update.meets_testing_requirements and
Update.met_testing_requirements.

fixes #1009

### DIFF
--- a/bodhi/tests/server/base.py
+++ b/bodhi/tests/server/base.py
@@ -1,0 +1,59 @@
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+"""
+This module contains a useful base test class that helps with common testing needs when testing
+bodhi.server modules.
+"""
+
+import unittest
+
+from sqlalchemy import create_engine
+from sqlalchemy import event
+from sqlalchemy.orm import scoped_session, sessionmaker
+from zope.sqlalchemy import ZopeTransactionExtension
+
+from bodhi.server import log
+from bodhi.tests.server import populate
+from bodhi.server.models import (
+    Base,
+)
+
+DB_PATH = 'sqlite://'
+DB_NAME = None
+
+
+class BaseTestCase(unittest.TestCase):
+    def setUp(self):
+        engine = create_engine(DB_PATH)
+        # We want a special session that lasts longer than a transaction
+        Session = scoped_session(
+            sessionmaker(bind=engine, extension=ZopeTransactionExtension(keep_session=True)))
+        log.debug('Creating all models for %s' % engine)
+        Base.metadata.bind = engine
+        Base.metadata.create_all(engine)
+        self.db = Session()
+        populate(self.db)
+
+        # Track sql statements in every test
+        self.sql_statements = []
+
+        def track(conn, cursor, statement, param, ctx, many):
+            self.sql_statements.append(statement)
+
+        event.listen(engine, "before_cursor_execute", track)
+
+    def tearDown(self):
+        log.debug('Removing session')
+        self.db.close()
+        del self.sql_statements
--- a/bodhi/tests/server/functional/base.py
+++ b/bodhi/tests/server/functional/base.py
@@ -13,23 +13,14 @@
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
 
 import os
-import unittest
 
 from webtest import TestApp
-from sqlalchemy import create_engine
-from sqlalchemy import event
-from sqlalchemy.orm import scoped_session, sessionmaker
-from zope.sqlalchemy import ZopeTransactionExtension
 
-from bodhi.server import main, log
-from bodhi.tests.server import populate
-from bodhi.server.models import (
-    Base,
-)
+from bodhi.server import main
+from bodhi.tests.server.base import BaseTestCase, DB_NAME, DB_PATH
+
 
 FAITOUT = 'http://209.132.184.152/faitout/'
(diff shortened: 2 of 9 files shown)


bodhi/tests/server/base.py
```diff
@@ -0,0 +1,59 @@
...
+from sqlalchemy.orm import scoped_session, sessionmaker
+from zope.sqlalchemy import ZopeTransactionExtension
+
+from bodhi.server import log
+from bodhi.tests.server import populate
+from bodhi.server.models import (
+    Base,
+)
+
+DB_PATH = 'sqlite://'
```
pypingou, line 32: Thought about using faitout when running the tests (say in Jenkins)?
http://faitout.fedorainfracloud.org/

bodhi/tests/server/base.py
```diff
@@ -0,0 +1,59 @@
...
+DB_NAME = None
+
+
+class BaseTestCase(unittest.TestCase):
+    def setUp(self):
+        engine = create_engine(DB_PATH)
+        # We want a special session that lasts longer than a transaction
+        Session = scoped_session(
+            sessionmaker(bind=engine, extension=ZopeTransactionExtension(keep_session=True)))
```
pypingou, line 41: there is an odd tweak to try to make sqlite enforce referential constraints if you're interested: https://pagure.io/pagure/blob/master/f/pagure/lib/model.py#_75-81

bodhi/tests/server/functional/base.py
```diff
@@ -13,23 +13,14 @@
...
-from bodhi.server import main, log
-from bodhi.tests.server import populate
-from bodhi.server.models import (
-    Base,
-)
+from bodhi.server import main
+from bodhi.tests.server.base import BaseTestCase, DB_NAME, DB_PATH
+
 
 FAITOUT = 'http://209.132.184.152/faitout/'
```
pypingou, line 23: The URL can be changed to the one I specified above :)

And I guess this answers my question :)

bodhi/tests/server/models/__init__.py
```diff
@@ -32,9 +32,10 @@ class ModelTest(object):
     klass = None
     attrs = {}
 
-    def setup(self):
+    def setUp(self):
         engine = create_engine('sqlite://')
-        Session = scoped_session(sessionmaker(extension=ZopeTransactionExtension()))
+        Session = scoped_session(
+            sessionmaker(extension=ZopeTransactionExtension(keep_session=True)))
         Session.configure(bind=engine)
```
pypingou, line 39: No inheritance here?

bodhi/tests/server/models/test_models.py
```diff
@@ -514,30 +514,154 @@ def test_set_request_stable_after_week_in_testing(self, publish):
...
     @mock.patch('bodhi.server.notifications.publish')
-    def test_set_meets_then_met_requirements(self, publish):
-        req = DummyRequest()
-        req.errors = cornice.Errors()
-        req.koji = buildsys.get_session()
-        req.user = model.User(name='bob')
-
+    def test_met_testing_requirements_at_7_days_after_bodhi_comment(self, publish):
+        """
+        Ensure a correct True return value from Update.met_testing_requirements() after an update
```
pypingou, line 519: interesting formulation :)

bodhi/tests/server/test_metadata.py
```diff
@@ -49,7 +49,7 @@ def __init__(self, *args, **kw):
 
     def setUp(self):
         engine = create_engine(DB_PATH)
-        Session = scoped_session(sessionmaker(extension=ZopeTransactionExtension()))
+        Session = scoped_session(sessionmaker(extension=ZopeTransactionExtension(keep_session=True)))
```
pypingou, line 52: Here also, no inheritance?

pypingou:
Code-wise looks good, a few questions on the tests though :)

bodhi/tests/server/models/__init__.py
bowlofeggs: I attempted to use inheritance here but these tests do not work properly with that superclass. I plan to return to these later to debug that, but I'd rather keep this PR focused on what is needed to fix and test the e-mail issue.

bodhi/tests/server/models/test_models.py
bowlofeggs: ☺

bodhi/tests/server/test_metadata.py
bowlofeggs: Same thing - these tests didn't pass with the full inheritance. In fact, you will find that quite a few other Bodhi modules have made their own TestCases but I don't want to change those for this PR.

bodhi/tests/server/functional/base.py
bowlofeggs: I'd rather do this change as a separate effort as well.

bowlofeggs:
@pypingou I'd prefer to only make the changes needed to get the e-mail issue fixed and tested. In order to test it, I did need some code that was in the functional tests, which is why you see those changed here so I could form a new superclass for them to share. However, the goal of this PR wasn't to fix all the test problems (there are quite a few) but to give me a test superclass that I can use to test the e-mail change without repeating myself.

The changes I made to the two other test cases (the `keep_session=True` bits) were needed because those tests were causing the new ones to fail due to transaction reuse.

I do want to fix these issues, but it will be a significant amount of work and I'd rather separate that concern from this one. We need to get this merged ASAP so I can release bodhi 2.3.0 and stop those e-mails from going out ☺

In general, Bodhi's tests are going to be an area of focus for me over time. I plan to make incremental fixes over time, with a goal to have this new base test case be used everywhere.

pypingou:
Therefore :+1: for me

bowlofeggs:
Thanks @pypingou!

Issue #939, opened 2016-09-19
Add example documentation for using the CLI


bowlofeggs:
We should add a workflows section to the man page that has some practical examples on how to use the CLI.

cosmologist10:
Hey @bowlofeggs, I want to work on this issue. Can you please elaborate this task.

bowlofeggs:
Hello @cosmologist10! The bodhi man page's source code is located in the sphinx project:

https://github.com/fedora-infra/bodhi/blob/2.2.4/docs/man_bodhi.rst

Quality man pages often have an Examples section that show the user a few common tasks they might want to do. For this ticket, we'll want to add an Examples section (probably just before the Bugs section at the end) with some examples of common bodhi CLI operations (perhaps things like creating updates, editing updates, downloading the files for an update, creating buildroot overrides, or anything else that sounds useful).

cosmologist10:
Thanks @bowlofeggs, I am working on this issue.

bowlofeggs:
We have some example documentation now.

Issue #997, opened 2016-10-05
search should be case-insensitive


plughiubra:
searching for "wxgtk3", for example, produces no results. the search should be case-insensitive, in line with user expectations about most search results.

bowlofeggs:
I agree, thanks for the report!

cosmologist10:
Hey! I want to work on this issue. Can you assist me?  
Thanks!

bowlofeggs:
@cosmologist10 I'm not too familiar with this area of the code myself, but I think [this code](https://github.com/fedora-infra/bodhi/blob/2.2.4/bodhi/server/services/updates.py#L214-L217) is probably the issue:

```
    packages = data.get('packages')
    if packages is not None:
        query = query.join(Update.builds).join(Build.package)
        query = query.filter(or_(*[Package.name==pkg for pkg in packages]))
```

My guess is that `name==pkg` is probably the issue, and it should probably be changed to be a case insensitive match instead of a pure equality match.

cosmologist10:
Thanks @bowlofeggs , I am working on this issue.

cosmologist10:
Hey @bowlofeggs , There are two ways to work on  this issue :
1.  name.lower()==pkg.lower()
2.  import re
     re.match("pattern", "string", flags=re.IGNORECASE)

Can you suggest me which one is better, according to you.

pnemade:
@cosmologist10 you need to give meaningful information for a commit log message. I see you just said "Solved" but that is not enough. The reviewer of your commit should get understanding by just reading the log message that for what purpose that commit has been added to the bodhi source.

For this fix your commit message should be saying "Change package search to case-insensitive"

bowlofeggs:
@cosmologist10 Another option that would work well and that would improve the search otherwise is to use the ilike operator (i for case Insensitive, and otherwise the SQL like operator):

http://docs.sqlalchemy.org/en/latest/orm/internals.html#sqlalchemy.orm.properties.ColumnProperty.Comparator.ilike

cosmologist10:
Thanks @bowlofeggs and @pnemade for the suggestions. I'm working on it.

lalitkr:
Is this issue still open. I would like to work on it.

bowlofeggs:
Hello @lalitkr Hmmm, I'm not sure. It seems that the pull request for this issue was closed. I do not recall merging a fix for this. @cosmologist10, do you still plan to work on this?

cosmologist10:
[...]

PR #1034, opened 2016-10-15
Issue #939 solved

cosmologist10:
Issue #939: Add example documentation for using the CLI is done! Have a look on it.


trishnaguha:
@cosmologist10 the fix for #997 should be in a separate PR.

PR #1036, opened 2016-10-17
PEP-8 bodhi.server.services.updates.

bowlofeggs:


bodhi/server/services/updates.py
```diff
@@ -92,9 +92,9 @@ def get_update(request):
 def get_update_for_editing(request):
     """Return a single update from an id, title, or alias for the edit form"""
     return dict(
-        update=request.validated['update'],
-        types=reversed(bodhi.server.models.UpdateType.values()),
-        severities=sorted(bodhi.server.models.UpdateSeverity.values(), key=bodhi.server.util.sort_severity),
+        update=request.validated['update'], types=reversed(bodhi.server.models.UpdateType.values()),
+        severities=sorted(
+            bodhi.server.models.UpdateSeverity.values(), key=bodhi.server.util.sort_severity),
```
pypingou, line 97: I think I would have kept one value/argument per line here

pypingou:
:+1: for me

bodhi/server/services/updates.py
bowlofeggs: Either way is a valid style by PEP-8 rules, but I covet my screen's vertical space which is why I prefer this particular style. I'm fine with people following their own preferences though, so long as flake8 agrees ☺

PR #1042, opened 2016-10-18
Add route to allow trailing slash on masher url

### LINKED ISSUE #1041: GET on /masher/ errors with a 404 response
GET on `/masher` works, but GET on `/masher/` returns an HTTP 404 error code. We should add a route to allow the trailing slash on this URL.

trishnaguha:
Fixes #1041 
This would allow `302` URL redirection from `/masher` to `/masher/`.


bowlofeggs:
Thanks!

PR #1050, opened 2016-10-19
Disallow NULL values in the Update.autokarma column.

### LINKED ISSUE #1048: Autokarma should not be None
Came to know about the issue when I was working on #1027 
We need to make sure that `Autokarma` cannot be `None`. It should be either `True` or `False`.

screenshot from 2016-10-19 10-44-14

bowlofeggs:
fixes #1048

### DIFF
--- a/alembic/versions/4f2f825bcf4a_disallow_null_values_in_autokarma.py
+++ b/alembic/versions/4f2f825bcf4a_disallow_null_values_in_autokarma.py
@@ -1,0 +1,34 @@
+"""Disallow NULL values in autokarma.
+
+Revision ID: 4f2f825bcf4a
+Revises: 06aa0e8aa5d2
+Create Date: 2016-10-19 15:09:50.884185
+
+"""
+
+from alembic import op
+import sqlalchemy as sa
+
+
+# revision identifiers, used by Alembic.
+revision = '4f2f825bcf4a'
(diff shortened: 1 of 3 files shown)


trishnaguha:
Verified :100:  Now it is displaying `Disabled` instead of `None` :). 
Thanks for the fix @bowlofeggs :icecream:

trishnaguha (approved):

alembic/versions/4f2f825bcf4a_disallow_null_values_in_autokarma.py
```diff
@@ -0,0 +1,34 @@
...
+
+"""
+
+from alembic import op
+import sqlalchemy as sa
+
+
+# revision identifiers, used by Alembic.
+revision = '4f2f825bcf4a'
+down_revision = '3cde3882442a'
```
trishnaguha: `down_revision` should now be modified to `'06aa0e8aa5d2'` since another migration is recently added for dropping `update.karma` column https://github.com/fedora-infra/bodhi/commit/2a6bb6fc31a48914a46fe35915caad8829404905

bowlofeggs: Good catch @trishnaguha! I've fixed this.

jeremycline:
Looks good, :shipit:

PR #1055, opened 2016-10-21
Small fixes for signing handler and status indication

puiterwijk:


bowlofeggs (approved):
Thanks @puiterwijk, these all look great to me. Thanks for making the effort to keep the commits atomic too!

PR #1056, opened 2016-10-21
Configure nosetests to require 77% coverage and document its usage.

bowlofeggs:
Though we should and do have the goal of higher test coverage,
bodhi currently only has 77% coverage. This commit alters the
nosetests configuration so that the coverage will not be allowed to
regress and documents that nose should be used to run the tests.

Over time I plan to slowly increase our test coverage and will
adjust these settings as I go.


pypingou:
Neat, :+1: :)

Issue #1057, opened 2016-10-21
Remove from models import * from bodhi.server.models.__init__


bowlofeggs:
We should stop the `from thing import *` pattern, and one particular place it's being used is in `bodhi.server.models.__init__`. To fix this issue, we should remove that line and choose between these two options:
1. Fix all the things that import from the `__init__.py`.
2. Rename models.py to `__init__.py`

I personally prefer the first option, but welcome discussion about the second.

ankit01ojha:
Is second option will be correct?

bowlofeggs:
Neither option is more correct than the other. It's a matter of preference. Some people dislike when Python package `__init__.py` files contain significant code. I personally feel that way, but I can't make any technical argument as to why. It just feels less clean to me. However, that's just a personal preference and I don't want to force my preferences onto others ☺ The second option is certainly easier ☺

trishnaguha:
IMO **first** one would be better choice. It is good to have `DB model` inside `models.py` only :)

bowlofeggs:
I think I agree with @trishnaguha. It is always a little surprising when projects put lots of code in `__init__.py` files.

There is a third option which might be nice: we could get rid of the `bodhi.server.models` package and replace it with a `bodhi.server.models` module. The only other thing in the models package is the `enum.py`. That could just become `bodhi.server.enum`, or it might make sense to just be in `bodhi.server.models`. Either way I think would be cleaner than `from bodhi.server.models import models` which is pretty strange if you ask me ☺

trishnaguha:
+1 to replace `bodhi.server.models` package with `bodhi.server.models` module and `enum.py` with `bodhi.server.enum` . This sounds even better :-).
Any of the first or third sounds good to me :).

bowlofeggs:
@trishnaguha Actually I had proposed just putting the enum stuff inside `bodhi.server.models`, but we could make enum its own thing too. Since it's part of bodhi.server.models now I figured it made sense to keep it there. It is used by the models extensively, so I think it makes sense to be part of them.

trishnaguha:
Fair :)

ankit01ojha:
Can you please tell me now  what basically I have to do with enum.py?

ankit01ojha:
Can you please tell me now  what basically I have to do with enum.py?

ankit01ojha:
Do I have to put this code from bodhi.server.models import models in place of from models import * .

PR #1058, opened 2016-10-21
PEP-8 bodhi.server.models.__init__.

bowlofeggs:
Frankly I was tempted to just get rid of the `from models import *` because I don't generally like that pattern. I'd prefer to either move all the models into `__init__.py` or just fix all the things that import from the `__init__.py`. However, I didn't want to make either of those drastic moves at this time, so I'm just slapping a `# noqa` on it for now. I filed https://github.com/fedora-infra/bodhi/issues/1057 to track getting rid of this line so we remember to do that later.


pypingou:
Fair, :+1:

PR #1059, opened 2016-10-21
PEP-8 server.services.overrides.

bowlofeggs:


bodhi/server/services/overrides.py
```diff
@@ -102,19 +104,19 @@ def query_overrides(request):
     expired = data.get('expired')
     if expired is not None:
         if expired:
-            query = query.filter(BuildrootOverride.expired_date!=None)
+            query = query.filter(BuildrootOverride.expired_date.isnot(None))
```
pypingou, line 107: I didn't know this one, cool

Is it present in all versions of sqlalchemy?

pypingou:
:+1: for me

bodhi/server/services/overrides.py
bowlofeggs: @pypingou Looks like it was introduced in 0.7.9:

http://docs.sqlalchemy.org/en/latest/core/sqlelement.html#sqlalchemy.sql.operators.ColumnOperators.isnot

PR #1062, opened 2016-10-24
Add python-cornice to the development environment.

bowlofeggs:
During the Cloud test day today, I found that the unit tests for
Bodhi did not pass with the Fedora 25 Vagrant image. It turned out
that python-cornice was not getting installed anymore. I didn't
determine exactly which change between Fedora 24 and 25 caused
this, but I suspect something that we were installing stopped
depending on it.


trishnaguha:
:+1:

jeremycline (approved):
I just ran into this and can confirm it's a problem :shipit:

PR #1066, opened 2016-10-27
Drop the Update.karma column and replace it with a property.

### LINKED ISSUE #1018: Bodhi allows users to give +1 twice, counting it as +2 after a Build is updated or removed
As of the moment of this filing, [this znc-1.6.3-6.el7 update](https://bodhi.fedoraproject.org/updates/znc-1.6.3-6.el7#comment-505077) has +2 karma, both of which seem to have come from elyscape. Bodhi should allow users to add further comments with Karma (so that users can change their mind and swap a +1 for a -1 or vice-versa), but it should only count their most recent karma vote in the total.

I am labeling this as critical since it has potential for users to game the system.

### LINKED ISSUE #1064: Bodhi only allows users to change their mind about karma once
Due to [this block](https://github.com/fedora-infra/bodhi/blob/2.2.4/bodhi/server/models/models.py#L1432-L1447) in the `comment()` method, if a user submits a karma, and then later submits the opposite karma, they will be unable to submit the first karma again:

```
            mycomments = [c.karma for c in self.comments if c.user.name == author][reset_index:]
            if karma not in mycomments:
                if karma == 1 and -1 in mycomments:
                    self.karma += 2
                    caveats.append({
                        'name': 'karma',
                        'description': 'Your karma standing was reversed.',
                    })
                elif karma == -1 and 1 in mycomments:
                    self.karma -= 2
                    caveats.append({
                        'name': 'karma',
                        'description': 'Your karma standing was reversed.',
                    })
                else:
                    self.karma += karma
```

This is an edge case for certain, but it is not outside the realm of realistic possibilities and so we should fix it.

@puiterwijk discovered this problem while testing #1018 in staging. Fortunately, the proposed refactor for #1018 to make karma dynamically calculated should also fix this issue as a side effect.

### LINKED ISSUE #1065: Update.has_negative_karma() counts karma from before new or removed builds
Bodhi's karma counting code resets the karma to 0 when builds are added or removed from an Update, and [a user's karma from before the reset is not supposed to be counted](https://github.com/fedora-infra/bodhi/blob/2.2.4/bodhi/server/models/models.py#L1426-L1430):

```
            # Take all comments since the previous karma reset
            reset_index = 0
            for i, comment in enumerate(self.comments):
                if (comment.user.name == u'bodhi' and ('New build' in
                    comment.text or 'Removed build' in comment.text)):
                        reset_index = i
```

Due to #1018 this code doesn't work quite as intended, but it also occurred to me that [`Update.has_negative_karma()`](https://github.com/fedora-infra/bodhi/blob/f515793beaf5d35d3918c808edd9dd153eeba2b0/bodhi/server/models/models.py#L1739-L1748) does count karma that was given before karma reset events:

```
    def has_negative_karma(self):
        """Check for negative karma, Returns True if the update has negative karma"""
        feedback = defaultdict(int)
        for comment in self.comments:
            if not comment.anonymous:
                feedback[comment.user.name] = comment.karma
        for karma in feedback.values():
            if karma < 0:
                return True
        return False
```

This means that critical path updates that add or remove builds after receiving negative karma will still automatically get disabled when receiving new karma comments after the reset event because the old negative karma is still affecting the test. The old karma will not contribute to the Update's karma, however.

This is an edge case scenario, and fails in a "safe" way because the consequence is simply that autopush will get disabled for the Update. Therefore, I consider this a low priority issue.

bowlofeggs:
Prior to this commit Update karma had been denormalized, but that
led to a few issues with keeping the karma column correctly in sync
with the Comments in an Update. In one case, a user was able to
double +1 an Update, causing the karma column to contain an
incorrect value. The code that managed the karma column was
complicated and it was determined that the denormalized column did
not provide a significant enough performance gain to justify the
complex code it takes to maintain correct values in the database.
Nearly all views on Updates need to load the associated Comments
from the database, so calculating the Update.karma property
dynamically will not introduce any new queries in most cases.

This commit drops the karma column and replaces it with two
properties. The first calculates the sums of positive and negative
karma separately, and the second uses the first to generate and
return the overall karma. The first property was created to replace
the has_negative_karma() method, which had an issue where it failed
to correctly filter out karma before karma reset events. Rather
than fix that method, it was determined that it would be simpler to
drop it and rework the caller to use the new _composite_karma()
property to determine if there is negative karma.

This commit also addresses a third issue with karma counting where
users who changed their mind more than once would no longer be able
to change the karma of an Update.

fixes #1018
fixes #1064
fixes #1065

### DIFF
--- a/alembic/versions/06aa0e8aa5d2_drop_the_update_karma_column.py
+++ b/alembic/versions/06aa0e8aa5d2_drop_the_update_karma_column.py
@@ -1,0 +1,30 @@
+"""Drop the Update.karma column.
+
+Revision ID: 06aa0e8aa5d2
+Revises: 3cde3882442a
+Create Date: 2016-10-26 16:55:54.875994
+
+"""
+from alembic import op
+
+
+# revision identifiers, used by Alembic.
+revision = '06aa0e8aa5d2'
+down_revision = '3cde3882442a'
+
+
+def upgrade():
+    """
+    Drop the karma column from the updates table.
+    """
+    op.drop_column('updates', 'karma')
+
+
+def downgrade():
+    """
+    Downgrade is not supported. If we ever do want to do this for some reason, we can use the
+    code from the bodhi.server.models.models.Update.karma() property that was written in the same
+    commit that introduced this migration as a guide for how to calculate the karma column. As it is
+    highly unlikely that will ever be needed, this function simply raises an Exception for now.
+    """
+    raise NotImplemented('Downgrade not supported')
--- a/bodhi/server/models/models.py
+++ b/bodhi/server/models/models.py
@@ -603,7 +603,6 @@ class Update(Base):
 
     title = Column(UnicodeText, unique=True, default=None, index=True)
 
-    karma = Column(Integer, default=0)
     autokarma = Column(Boolean, default=True)
     stable_karma = Column(Integer, nullable=True)
     unstable_karma = Column(Integer, nullable=True)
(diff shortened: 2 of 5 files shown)


sayanchowdhury:
Looks good to me :+1:

PR #1070, opened 2016-10-27
If a release does not have a pending signing tag, always regard as si…

puiterwijk:
…gned


bowlofeggs (approved):
Good catch! I will merge this and I will cherry-pick it into the 2.3 branch so that our next release also has this.

PR #1072, opened 2016-10-28
Result != results

puiterwijk:

### DIFF
--- a/bodhi/server/consumers/masher.py
+++ b/bodhi/server/consumers/masher.py
@@ -954,7 +954,7 @@ class MasherThread(threading.Thread):
                     topic="ostree.compose.finish",
                     msg=dict(tag=tag,
                              ref=result.get('ref'),
-                             commitid=results.get('commitid')),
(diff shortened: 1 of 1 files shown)


bowlofeggs (approved):
I wish we had test coverage for this ☹

pypingou:
:+1: as well, but indeed no tests :(

bowlofeggs:
@pypingou I'm going to allow this one in without tests because we are checking it live on production (really, ooof!). However, I do have a task warrior recurring task to increase bodhi's test coverage slowly to 100% over time, and I plan to require new code to be tested from now on.

Today is a bit of a special case because I want to make a bodh-2.3.2 quickly since production is broken. We are making sure this works by hotfixing production and then committing those changes here. Definitely not best practice, but I don't have much time to get production fixed before the weekend so I think it's allowable in this circumstance.

Also, technically I haven't merged the PR that requires test coverage yet ☺

bowlofeggs:
I have cherry picked this one to the 2.3 branch.

PR #1074, opened 2016-10-28
Use the correct name when looking up EPEL e-mail templates.

### LINKED ISSUE #1073: Mismatch between e-mail template name and key lookup
Bodhi 2.3.0 introduced a new `bodhi.server.mail.fedora_epel_legacy_errata_template` string, but the models.py looked it up by the name `fedora_epel_errata_legacy_template` which caused the masher to fail:

```
  File "/usr/lib/python2.7/site-packages/bodhi/server/mail.py", line 341, in get_template
    use_template = globals()[use_template]
KeyError: u'fedora_epel_errata_legacy_template'
```

bowlofeggs:
fixes #1073

### DIFF
--- a/bodhi/server/models/models.py
+++ b/bodhi/server/models/models.py
@@ -1252,7 +1252,7 @@ class Update(Base):
 
         # switch email template to legacy if update aims to EPEL <= 7
         if release_name == 'fedora_epel' and self.release.version_int <= 7:
-            templatetype = '%s_errata_legacy_template' % release_name
+            templatetype = '%s_legacy_errata_template' % release_name
         else:
             templatetype = '%s_errata_template' % release_name
 


pypingou:
Looks fine but I don't see a unit-tests :D

(:+1: to merge assuming you tested it :))

bowlofeggs:
Copied from my other PR, but here for reference:

```


@pypingou I'm going to allow this one in without tests because we are checking it live on production (really, ooof!). However, I do have a task warrior recurring task to increase bodhi's test coverage slowly to 100% over time, and I plan to require new code to be tested from now on.

Today is a bit of a special case because I want to make a bodh-2.3.2 quickly since production is broken. We are making sure this works by hotfixing production and then committing those changes here. Definitely not best practice, but I don't have much time to get production fixed before the weekend so I think it's allowable in this circumstance.

Also, technically I haven't merged the PR that requires test coverage yet ☺
```

pypingou:
I see the emergency but as long as prod is (hot)fixed and working there is no real rush imho. We could just leave it fixed as it is and make 2.3.2 with the fix and the tests.

I'm just offering thoughts here, I am ok with the change and the PR (as stated above), to me it just depends on how badly we want to follow the (future) `no code without tests` rule. I know I'm bad with it on pagure so I will definitively not judge anyone for merging without tests :)

pypingou:
> I see the emergency but as long as prod is (hot)fixed and working there is no real rush imho. We could just leave it fixed as it is and make 2.3.2 with the fix and the tests.

That is of course assuming we're talking about a day or two to add the tests, not weeks :)

bowlofeggs:
I will be unavailable to write the tests for a about 1.5 weeks actually, and I'd prefer not to just rely on hotfixing during that time. If someone were to reinstall the RPMs for any reason, the hotfixes would be lost. I'm hoping to make a bodhi 2.3.2 release today or tomorrow with this and the three other fixes so the new RPM has everything. That's why I'm in a bit of a rush ☺

pypingou:
Fair, sorry I had forgot you were taking off :s

bowlofeggs:
I have chery picked this to the 2.3 branch.

PR #1076, opened 2016-10-28
Fix push

puiterwijk:


bodhi/server/push.py
```diff
@@ -47,6 +47,9 @@
 @click.option('--resume', help='Resume one or more previously failed pushes',
               is_flag=True, default=False)
 def push(username, cert_prefix, config, **kwargs):
+    if not username:
+        raise Exception("Please provide --username")
```
bowlofeggs: I think there's a way to let click do this for you by declaring the option as required. I'll see if I can find it in their docs.

bowlofeggs: It seems that maybe it's a bit against click's philosophy to require options, since the name "option" implies "optional" ☺

However, there is a `prompt=True` argument we could pass to the username option, which would cause click to automatically prompt the user to give their username if they don't use the `--username` option. How about we just do it that way?

http://click.pocoo.org/5/api/#click.Option

bowlofeggs: Alternatively, we could make it an argument instead of an option, but I actually kinda like the prompt workflow myself. What do you think?

bodhi/server/push.py
```diff
@@ -80,7 +83,7 @@ def push(username, cert_prefix, config, **kwargs):
                     continue
 
                 for update in lockfiles[lockfile]:
-                    update = Update.get(update).first()
+                    update = session.query(Update).filter(Update.title==update).first()
```
bowlofeggs, line 83: One day we will have good test coverage on Bodhi to find things like this ☺

bowlofeggs:
I have cherry picked both of these commits to the 2.3 branch.

PR #1077, opened 2016-10-28
push.py now defaults to current Releases.

### LINKED ISSUE #1071: push.py does not filter archived releases
During our first run of the new `push.py` today, we noticed that it attempted to mash `f21-updates` and `f22-updates`. The old `push.py` used to filter out archives releases, but the new one doesn't do that anymore. We simply need to filter those as well.

bowlofeggs:
fixes #1071

### DIFF
--- a/bodhi/server/push.py
+++ b/bodhi/server/push.py
@@ -27,7 +27,7 @@ from sqlalchemy.sql import or_
 
 import bodhi.server.notifications
 from bodhi.server.util import transactional_session_maker
-from bodhi.server.models import Update, Base, UpdateRequest, Build, Release
+from bodhi.server.models import Update, Base, UpdateRequest, Build, Release, ReleaseState
 
 
 @click.command()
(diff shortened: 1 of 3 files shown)


bodhi/server/push.py
```diff
@@ -157,5 +147,40 @@ def push(username, cert_prefix, config, **kwargs):
...
+
+    if releases:
+        for r in releases.split(','):
+            release = session.query(Release).filter(
+                or_(Release.name == r,
+                    Release.name == r.upper(),
+                    Release.version == r)).first()
+            if not release:
+                click.echo('Unknown release: %s' % r)
+                sys.exit(1)
```
pypingou: From past experience I would recommend raising an exception here rather than killing the program.
The exception could then be handled by the CLI that would do the `sys.exit()`

Food for thought :)

bodhi/server/push.py
```diff
@@ -157,5 +147,40 @@ def push(username, cert_prefix, config, **kwargs):
...
+            else:
+                _releases.append(release)
+    else:
+        # Since the user didn't ask for specific Releases, let's just filter for releases that are
+        # current.
+        _releases = session.query(Release).filter(Release.state == ReleaseState.current)
```
puiterwijk: This skips releases in the Development stage (like F25).
Please make this only check !=archived.

puiterwijk (changes requested):

bodhi/server/push.py
bowlofeggs: @pypingou I attempted to go even further and use click's callback system to validate and convert the `--releases` option into real releases, but it ended up being pretty tricky since `--config` is also an option and we need the config to be parsed in order to get the db connection, and we need the db connection to validate and convert the releases. So I just raised click.BadParameter here as you suggested, and it gives the user a nice error message.

bodhi/server/push.py
bowlofeggs: @puiterwijk I don't think `!=archived` is what we want either, because there is also a `disabled` state. It seems that the four states are `disabled`, `pending` (which might be the same as what you call Development?), `current`, and `archived`.

I will act under the assumption that `pending` is what we want to include and make an update to this commit with that change.

bowlofeggs:
@puiterwijk, @pypingou: I have made the changes that you had each requested. How does it look now?

puiterwijk (approved):

pypingou:
I see something hitching me in there but that's a personal preference so it doesn't matter

:+1: for me!

bowlofeggs:
This has been cherry picked to the 2.3 branch as well.

PR #1079, opened 2016-10-30
Pass the fedmsg name explicitly

puiterwijk:
This broke because of the bodhi -> bodhi.server package move.

### DIFF
--- a/bodhi/server/notifications.py
+++ b/bodhi/server/notifications.py
@@ -20,6 +20,8 @@ import fedmsg
 import fedmsg.config
 import fedmsg.encoding
 
+import socket
+
 import bodhi.server
 import bodhi.server.config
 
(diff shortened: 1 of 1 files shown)


pypingou:
:+1: for me

bowlofeggs:
This change looks fine to me. Would you be willing to add a test to assert correct behaviour?

bowlofeggs:
@puiterwijk and I agreed that I would write the tests for this pull request. I may choose to write them as a separate pull request, but I'll decide about that once I have them ready.

bodhi/server/notifications.py
```diff
@@ -38,6 +40,9 @@ def init(active=None, cert_prefix=None):
     if cert_prefix is not None:
         fedmsg_config['cert_prefix'] = cert_prefix
 
+    hostname = socket.gethostname().split('.', 1)[0]
+    fedmsg_config['name'] = 'bodhi.%s' % hostname
```
bowlofeggs: This always overrides `'name'` from line 39 above. Is that intentional? If so, we should probably drop line 39.

puiterwijk: Nope, that's an issue :)

bowlofeggs:
OK, I now have tests on my dev box for this area of the code and can commit to sending a PR for it very soon. I will merge this PR as-is without tests and will send a new PR that has the tests today.

bowlofeggs:
I've submitted https://github.com/fedora-infra/bodhi/pull/1112 with tests for this pull request.

bowlofeggs:
This has been cherry picked as c25fcc11d4fa17dba43bfc5822c2d48cf1c0b0d9 to the 2.3 branch.

PR #1080, opened 2016-10-30
Add a set of tests for bodhi.server.captcha.math_generator.

bowlofeggs:


bodhi/server/captcha.py
```diff
@@ -29,7 +29,7 @@
 
 def math_generator(plainkey, settings):
     """ Given a plainkey, return its expected value.
-    If plainkey is None, return a random one.
+    If plainkey is None, return a random one. settings is not used.
```
pypingou, line 32: I've learned the use of the `_` argument which allows to do something like:

```
>>> def foo(bar, _):
...     print bar
>>> foo(1)
  File "<input>", line 1, in <module>
    foo(1)
TypeError: foo() takes exactly 2 arguments (1 given)
>>> foo(1, 2)
1
```

Basically, you require two arguments but you never use the second one (marked as `_`) and code lint such as pep8 or pylint won't complain about them.

pypingou:
One suggestion, otherwise looks fine to me.

bowlofeggs:
@pypingou Ah that's an interesting convention. I think I may adopt that in the future! However, since this PR is really just about adding tests to the function (and a clarifying comment) I think I will leave that as-is for now. Thanks for the review and the suggestion!

PR #1081, opened 2016-11-02
Tests server.consumers.updates

bowlofeggs:
The overall goal of this pull request is to add some test coverage for ```bodhi.server.consumers.updates.UpdatesHandler.__init__```. However, that code was more difficult to test because it failed to separate concerns and reuse code, so this pull request also contains three more commits that refactor that code and other related code to make it more suitable for testing.


pypingou:
Nothing standing out for me, :+1:

PR #1088, opened 2016-11-05
Add check-sig-status argument to bodhi-push

puiterwijk:
This argument makes the push command check out to koji to refresh the
signed state of builds if requested.
This handles the case where for some reason the signed handler didn't pick
up on the signed status update, and admins really want to check it.

### DIFF
--- a/bodhi/server/push.py
+++ b/bodhi/server/push.py
@@ -25,6 +25,7 @@ from pyramid.paster import get_appsettings
 from sqlalchemy import engine_from_config
 from sqlalchemy.sql import or_
 
+from bodhi.server import buildsys
 import bodhi.server.notifications
 from bodhi.server.util import transactional_session_maker
 from bodhi.server.models import Update, Base, UpdateRequest, Build, Release, ReleaseState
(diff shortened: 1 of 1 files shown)


bodhi/server/push.py
```diff
@@ -25,6 +25,7 @@
 from sqlalchemy import engine_from_config
 from sqlalchemy.sql import or_
 
+from bodhi.server import buildsys
 import bodhi.server.notifications
```
bowlofeggs, line 29: Very small nitpick, but I find alphabetic sorting of the imports to be pleasing ☺

bodhi/server/push.py
```diff
@@ -108,6 +112,17 @@ def push(username, cert_prefix, config, **kwargs):
...
+                if not update.signed and kwargs['check_sig_status']:
+                    # Only initialize the buildsys instance once
+                    if koji is None:
+                        koji = buildsys.get_session()
+                    for build in update.builds:
+                        if not build.signed:
+                            build_tags = build.get_tags(koji)
+                            if not update.release.pending_signing_tag in build_tags:
+                                click.echo('Update %s was refreshed as signed' % build.nvr)
+                                build.signed = True
```
bowlofeggs, line 125: This function is pretty large as it is and should be refactored to use sub functions so it's cleaner. Can we make the above code block be a helper function instead? We can refactor the rest of this method as a separate effort.

bowlofeggs (changes requested):
Before we merge this, we'll need two changes:
1. We need tests for the new code.
2. Let's move the new code into its own helper function.

bowlofeggs:
@puiterwijk This pull request has been sitting with requested changes for a while. Do you still intend to work on it?

puiterwijk:
@bowlofeggs no, as I told you a while ago I think that since we now have the root cause of the issue fixed, this can be closed.
Howver, you still wanted this code to go in, so closing is up to you.

PR #1092, opened 2016-11-09
Add a new script to clean up old mashes.

### LINKED ISSUE #765: Port the job to clean up old repos
Bodhi1 had a job in `jobs.py` that would clean up old stale mashes. We should port this logic over and have it keep the last 10 mashes around.

bowlofeggs:
This commit creates a script called bodhi-clean-old-mashes that is
intended to be called by cron. The script recursively deletes any
folders with names that end in a dash followed by a string that can
be interpreted as a float, sparing the newest 10 by lexigraphical
sorting.

fixes #765

### DIFF
--- a/bodhi/server/scripts/clean_old_mashes.py
+++ b/bodhi/server/scripts/clean_old_mashes.py
@@ -1,0 +1,65 @@
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+"""
+This tool cleans up old mashes that are left over in mash_dir.
+"""
+import collections
+
+import click
+import os
+import shutil
+
+from bodhi.server import config
+
+
+# How many of the newest mash dirs to keep during cleanup
+NUM_TO_KEEP = 10
+
+
+@click.command()
+@click.version_option(message='%(version)s')
+def clean_up():
+    """
+    Delete any repo mashes that are older than the newest 10 from each repo series.
+    """
+    mash_dir = config.config['mash_dir']
+
+    # This data structure will map the beginning of a group of dirs for the same repo to a list of
+    # the dirs that start off the same way.
+    pattern_matched_dirs = collections.defaultdict(list)
+
+    for directory in [d for d in os.listdir(mash_dir) if os.path.isdir(os.path.join(mash_dir, d))]:
+        # If this directory ends with a float, it is a candidate for potential deletion
+        try:
+            split_dir = directory.split('-')
+            float(split_dir[-1])
+        except ValueError:
+            # This directory didn't end in a float, so let's just move on to the next one.
+            continue
+
+        pattern = directory.replace(split_dir[-1], '')
+        pattern_matched_dirs[pattern].append(directory)
+
+    dirs_to_delete = []
+
+    for dirs in pattern_matched_dirs.values():
+        if len(dirs) > NUM_TO_KEEP:
+            dirs_to_delete.extend(sorted(dirs, reverse=True)[NUM_TO_KEEP:])
+
+    if dirs_to_delete:
+        print 'Deleting the following directories:'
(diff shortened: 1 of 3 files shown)


bodhi/server/scripts/clean_old_mashes.py
```diff
@@ -0,0 +1,67 @@
...
+
+        pattern = directory.replace(split_dir[-1], '')
+        if pattern in pattern_matched_dirs:
+            pattern_matched_dirs[pattern].append(directory)
+        else:
+            pattern_matched_dirs[pattern] = [directory]
+
+    dirs_to_delete = []
+
+    for pattern, dirs in pattern_matched_dirs.items():
```
pypingou: Looking at the code below, wouldn't a simple `for pattern in pattern_matched_dirs` be sufficient?

bodhi/server/scripts/clean_old_mashes.py
```diff
@@ -0,0 +1,67 @@
...
+@click.version_option(message='%(version)s')
+def clean_up():
+    """
+    Delete any repo mashes that are older than the newest 10 from each repo series.
+    """
+    mash_dir = config.config['mash_dir']
+
+    # This data structure will map the beginning of a group of dirs for the same repo to a list of
+    # the dirs that start off the same way.
+    pattern_matched_dirs = {}
```
pypingou: Did you consider using `collections.defaultdict(list)`? It would save the if/else line 51 to 54

bodhi/server/scripts/clean_old_mashes.py
```diff
@@ -0,0 +1,67 @@
...
+        if pattern in pattern_matched_dirs:
+            pattern_matched_dirs[pattern].append(directory)
+        else:
+            pattern_matched_dirs[pattern] = [directory]
+
+    dirs_to_delete = []
+
+    for pattern, dirs in pattern_matched_dirs.items():
+        if len(pattern_matched_dirs[pattern]) > NUM_TO_KEEP:
+            dirs_to_delete.extend(sorted(pattern_matched_dirs[pattern], reverse=True)[NUM_TO_KEEP:])
```
pypingou: Can we assume the alphanumerical ordering will correspond to the historical ordering?

bodhi/server/scripts/clean_old_mashes.py
```diff
@@ -0,0 +1,67 @@
...
+
+    if dirs_to_delete:
+        print 'Deleting the following directories:'
+        for d in dirs_to_delete:
+            d = os.path.join(mash_dir, d)
+            shutil.rmtree(d)
```
pypingou, line 64: For scripts having such a behavior, I like to add either a `--test` that does nothing but tell what it would do, or make that the default and enforce a `-y` to actually do the action.

pypingou:
Few suggestions, only one real question (the sorting, which looking at the tests seem to be fine).

So assuming the tests do answer the sorting question positively, the rest are just suggestions or ideas and thus can be ignored :)

:+1: for me

bodhi/server/scripts/clean_old_mashes.py
bowlofeggs: Good suggestion, I will make it so!

bowlofeggs: (I always forget about the `collections` module… thanks for the reminder!)

bowlofeggs: I've made this change.

bodhi/server/scripts/clean_old_mashes.py
bowlofeggs: @pypingou That's what I am counting on. The end of any one of these patterns is a unix timestamp (seconds since epoch), and we want to delete any of them that are older than the newest 10. The example folder names that I used in my test case were taken directly from production a few weeks ago, so they are real folder names.

bodhi/server/scripts/clean_old_mashes.py
bowlofeggs: I did consider adding some flags to this script, but I chose not to for three reasons:
1. It would increase the testing burden (it's nice to only need one test case since there are no options!)
2. This is meant to be called by cron and not by humans, so there won't be anybody to type `y`.
3. We can always expand the script later and add options if we want to, as a separate effort. This commit meets the requirements stated in #765 as is, so I'd prefer to iterate if we want to expand the requirements.

All that said, feel free to file an RFE if you think we should add that feature!

bodhi/server/scripts/clean_old_mashes.py
bowlofeggs: That's a good point. I think I had intended to write `dirs` instead of `pattern_matched_dirs[pattern]` in the code below when I wrote this line, but then promptly forgot to access the lists that way. Silly me.

Since I really don't need the keys at all (they are really just used to collate the lists of dirs that match the patterns), I've changed this line to:

```
for dirs in pattern_matched_dirs.values():
```

And then I used dirs in the code below. Nice catch on a silly pattern!

PR #1097, opened 2016-11-11
Refactor check_karma_thresholds method for clean code

trishnaguha:
Refactor `check_karma_thresholds` method. Avoid too much nested if-else to maintain clean and hygienic code :-).

### DIFF
--- a/bodhi/server/models/models.py
+++ b/bodhi/server/models/models.py
@@ -1709,39 +1709,41 @@ class Update(Base):
         return True, "All checks pass."
 
     def check_karma_thresholds(self, db, agent):
-        """Check if we have reached either karma threshold, and call set_request if necessary"""
-        if not self.locked:
-            if self.status is UpdateStatus.testing:
-                # If critical update receives negative karma disable autopush
-                if self.critpath and self.autokarma and self._composite_karma[1] != 0:
-                    log.info("Disabling Auto Push since the critical update has negative karma")
-                    self.autokarma = False
-                elif self.stable_karma not in (0, None) and self.karma >= self.stable_karma:
-                    if self.autokarma:
-                        log.info("Automatically marking %s as stable" % self.title)
-                        self.set_request(db, UpdateRequest.stable, agent)
-                        self.request = UpdateRequest.stable
-                        self.date_pushed = None
-                        notifications.publish(
-                            topic='update.karma.threshold.reach',
-                            msg=dict(update=self, status='stable'))
-                    else:
-                        # Add the 'testing_approval_msg_based_on_karma' message now
-                        log.info((
-                            "%s update has reached the stable karma threshold and can be pushed to "
-                            "stable now if the maintainer wishes") % self.title)
-                elif self.unstable_karma not in (0, None) and self.karma <= self.unstable_karma:
-                    log.info("Automatically unpushing %s" % self.title)
-                    self.obsolete(db)
-                    notifications.publish(
-                        topic='update.karma.threshold.reach',
-                        msg=dict(update=self, status='unstable'))
-            else:
-                # Ignore karma thresholds for pending/stable/obsolete updates
-                pass
-        else:
+        """
+        Check if we have reached either karma threshold, call set_request if necessary,
+        and Ignore karma thresholds if the update is locked and raise exception
+        """
+        # Raise Exception if the update is locked
+        if self.locked:
             log.debug('%s locked. Ignoring karma thresholds.' % self.title)
             raise LockedUpdateException
+        # Return if the status of the update is not in testing
+        if self.status != UpdateStatus.testing:
+            return
+        # If critical update receives negative karma disable autopush
+        if self.critpath and self.autokarma and self._composite_karma[1] != 0:
+            log.info("Disabling Auto Push since the critical update has negative karma")
+            self.autokarma = False
+        elif self.stable_karma and self.karma >= self.stable_karma:
+            if self.autokarma:
+                log.info("Automatically marking %s as stable" % self.title)
+                self.set_request(db, UpdateRequest.stable, agent)
+                self.request = UpdateRequest.stable
+                self.date_pushed = None
+                notifications.publish(
+                    topic='update.karma.threshold.reach',
+                    msg=dict(update=self, status='stable'))
+            else:
+                # Add the 'testing_approval_msg_based_on_karma' message now
+                log.info((
+                    "%s update has reached the stable karma threshold and can be pushed to "
+                    "stable now if the maintainer wishes") % self.title)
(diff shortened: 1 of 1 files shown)


bodhi/server/models/models.py
```diff
@@ -1709,39 +1709,43 @@ def check_requirements(self, session, settings):
...
+        Check if we have reached either karma threshold, and
+        call set_request if necessary
+        """
+        # Return if the update is locked
+        if self.locked:
             log.debug('%s locked. Ignoring karma thresholds.' % self.title)
             raise LockedUpdateException
+            return
+        # Return if the status of the update is not in testing
+        if self.status in [UpdateStatus.pending, UpdateStatus.stable,
```
sayanchowdhury: Marginal difference here but using `tuple` instead of `list` is better.

sayanchowdhury:
:+1: to the refactor

ralphbean:
Do we have any test cases in the test suite for `check_karma_thresholds`?

ralphbean:
It looks like we don't have any explicitly for the method right now.  This would be a perfect opportunity to add some.  My suggestion would be to revert this change first.  Then write some tests that exercise all of the code paths in the old code.  Then re-introduce this change.  Run the tests again, and we can be sure it's working as before!  :)

Reading through the change, it looks quite good.  :+1:

bodhi/server/models/models.py
```diff
@@ -1709,39 +1709,43 @@ def check_requirements(self, session, settings):
...
-        else:
+        """
+        Check if we have reached either karma threshold, and
+        call set_request if necessary
+        """
+        # Return if the update is locked
+        if self.locked:
             log.debug('%s locked. Ignoring karma thresholds.' % self.title)
             raise LockedUpdateException
+            return
```
ralphbean: No need to `return` after raising an exception.

trishnaguha:
@ralphbean Yes we have test coverage each and every condition for this method :) in here https://github.com/fedora-infra/bodhi/blob/develop/bodhi/tests/server/functional/test_updates.py . There are testcases which exercise all of the code paths of this method.

bowlofeggs:
It does appear that the new code does have 100% coverage, except for the unreachable return statement (I agree we should drop that anyway).

bodhi/server/models/models.py
```diff
@@ -1709,39 +1709,43 @@ def check_requirements(self, session, settings):
...
-                        msg=dict(update=self, status='unstable'))
-            else:
-                # Ignore karma thresholds for pending/stable/obsolete updates
-                pass
-        else:
+        """
+        Check if we have reached either karma threshold, and
+        call set_request if necessary
+        """
+        # Return if the update is locked
```
bowlofeggs: We aren't actually returning here, we're raising an Exception.

bodhi/server/models/models.py
```diff
@@ -1709,39 +1709,43 @@ def check_requirements(self, session, settings):
...
+        if self.locked:
             log.debug('%s locked. Ignoring karma thresholds.' % self.title)
             raise LockedUpdateException
```
bowlofeggs, line 1719: It might be nice to mention when this Exception gets raised in the docblock.

bodhi/server/models/models.py
```diff
@@ -1709,39 +1709,43 @@ def check_requirements(self, session, settings):
...
+            return
+        # Return if the status of the update is not in testing
+        if self.status in (UpdateStatus.pending, UpdateStatus.stable,
+                           UpdateStatus.obsolete, UpdateStatus.unpushed):
```
bowlofeggs: Why not just say `if self.status != UpdatesStatus.testing`? It's simpler, and is also more future proof in case we ever add new states.

bodhi/server/models/models.py
```diff
@@ -1709,39 +1709,43 @@ def check_requirements(self, session, settings):
...
+            return
+        # If critical update receives negative karma disable autopush
+        if self.critpath and self.autokarma and self._composite_karma[1] != 0:
+            log.info("Disabling Auto Push since the critical update has negative karma")
+            self.autokarma = False
+        elif self.stable_karma not in (0, None) and self.karma >= self.stable_karma:
```
bowlofeggs: While we're in here, I think we could get rid of the `not in (0, None)` and just say `elif self.stable_karma and self.karma >= self.stable_karma:` which is logically equivalent for our purposes and is slightly simpler and more readable.

bodhi/server/models/models.py
```diff
@@ -1709,39 +1709,43 @@ def check_requirements(self, session, settings):
...
+                self.date_pushed = None
+                notifications.publish(
+                    topic='update.karma.threshold.reach',
+                    msg=dict(update=self, status='stable'))
+            else:
+                # Add the 'testing_approval_msg_based_on_karma' message now
+                log.info((
+                    "%s update has reached the stable karma threshold and can be pushed to "
+                    "stable now if the maintainer wishes") % self.title)
+        elif self.unstable_karma not in (0, None) and self.karma <= self.unstable_karma:
```
bowlofeggs: Same thing here, we can just say `if self.unstable_karma` since it's logically equivalent for our purposes.

bodhi/server/models/models.py
```diff
@@ -1709,39 +1709,43 @@ def check_requirements(self, session, settings):
...
-                    notifications.publish(
-                        topic='update.karma.threshold.reach',
-                        msg=dict(update=self, status='unstable'))
-            else:
-                # Ignore karma thresholds for pending/stable/obsolete updates
-                pass
-        else:
+        """
+        Check if we have reached either karma threshold, and
+        call set_request if necessary
```
bowlofeggs: The 80 column line length is the one PEP-8 provision that I actively oppose because it was based on computer monitors that are so old these days that there's no excuse for coding on them (they are bad for your eyes!). I personally prefer to limit line lengths to 100 columns. Thus, I am actually fine with the wrapping that this docblock was at previously. However, I don't mind if you really want this change.

bowlofeggs (changes requested):
In order to merge this change, we'll need to fix the following:
1. Drop the unreachable return statement, as it will hurt our coverage numbers.
2. Compare self.status with testing directly, instead of comparing it with what it isn't.
3. Fix the comment that says we are returning to say we are raising an Exception.

All of the other comments I left are optional, so feel free to take or leave them.

Nice refactor!

ralphbean:
> Yes we have test coverage each and every condition for this method :)

Oh, good.  Thanks!  :)

trishnaguha:
Updated! :racehorse:

bowlofeggs (approved):
Looks great, thanks @trishnaguha!

trishnaguha:
Thanks :-).

PR #1098, opened 2016-11-11
Removed from models import * from bodhi.server.models.__init__  fixed #1057

### LINKED ISSUE #1057: Remove from models import * from bodhi.server.models.__init__

ankit01ojha:
removed bodhi.server.models package and replaced it with a module.


ralphbean:
Thanks @ankit01ojha!

A couple problems to deal with before this can be merged.
- First, the `.idea/` folder should be added to .gitignore.  Those files shouldn't be committed to the repo.
- Same for `python_notes`.
- Lastly, make sure that the bodhi.server.models package was not removed and replaced without `git` knowing about the move.  If it wasn't already so, this should be done with `git mv bodhi/server/models/models.py bodhi/server/models.py` so that git retains history (I think).

ankit01ojha:
Thanks @ralphbean but I am facing this error fatal: bad source, source=bodhi/bodhi/server/models/models.py, destination=bodhi/bodhi/server/models.py while doing git mv bodhi/server/models/models.py bodhi/server/models.py

PR #1101, opened 2016-11-14
PEP-8 bodhi.server.services.releases.

bowlofeggs:


bodhi/server/services/releases.py
```diff
@@ -29,16 +29,16 @@
...
 from bodhi.server.validators import (
     validate_tags,
     validate_enums,
     validate_updates,
     validate_packages,
     validate_release,
 )
+import bodhi.server.schemas
+import bodhi.server.security
+import bodhi.server.services.errors
```
pypingou, line 41: Did you have a reason for moving these around?
I usually do the opposite, keep the `import` above the `from ... import`, so I'm curious :)

pypingou:
One question, not related to the changes, :+1: for me

bodhi/server/services/releases.py
bowlofeggs: @pypingou I just don't like the `import`s to be in the middle of the `from … import`s, so I moved them to be alone.

PR #1102, opened 2016-11-14
This commit turns some test usernames into unicode strings.

bowlofeggs:
The test suite was printing this message:

SAWarning: Unicode type received non-unicode bind param value
'hunter1'.

Now it does not print that message.


pypingou:
Cool, :+1: for me

PR #1103, opened 2016-11-14
Add a test for bodhi.server.bugs.Bugzilla.__init__.

bowlofeggs:


bodhi/tests/server/test_bugs.py
```diff
@@ -0,0 +1,29 @@
...
+from bodhi.server import bugs
+
+
+class TestBugzilla(unittest.TestCase):
+    """This test class contains tests for the Bugzilla class."""
+    def test___init__(self):
+        """Assert that the __init__ method sets up the Bugzilla object correctly."""
+        bz = bugs.Bugzilla()
+
+        self.assertIs(bz._bz, None)
```
pypingou: There is a `self.assertNone` that could be used here if you want (There is also an `self.assertNotNone` if ever needed)

pypingou:
Small comment, not a blocker :+1: for me

bodhi/tests/server/test_bugs.py
bowlofeggs: Looks like it's actually `self.assertIsNone()`, but I'll go ahead and make that change!

PR #1106, opened 2016-11-14
PEP-8 bodhi.server.scripts and enforce style.

bowlofeggs:


bodhi/server/scripts/expire_overrides.py
```diff
@@ -56,7 +56,7 @@ def main(argv=sys.argv):
         now = datetime.utcnow()
 
         overrides = db.query(BuildrootOverride)
-        overrides = overrides.filter(BuildrootOverride.expired_date == None)
+        overrides = overrides.filter(is_(BuildrootOverride.expired_date, None))
```
bowlofeggs: I think this should actually be `BuildrootOverride.expired_date.is_(None)` instead.

pypingou: that's how you used it elsewhere indeed

pypingou:
Except for the line you pointed out, looks good to me

bodhi/server/scripts/expire_overrides.py
bowlofeggs: This is fixed.

PR #1108, opened 2016-11-14
Add Tests for bodhi.server.consumers.signed.SignedHandler.__init__()

bowlofeggs:
This pull request contains two commits. The first commit refactors the signed handler so that it uses ```get_db_factory()``` instead of repeating code. This is needed for two reasons: it reduced repeated code, and it makes the ```__init__()``` method significantly easier to test. The second commit adds tests for the ```__init__()``` method.

### DIFF
--- a/bodhi/server/consumers/signed.py
+++ b/bodhi/server/consumers/signed.py
@@ -22,11 +22,8 @@ import logging
 import pprint
 
 import fedmsg.consumers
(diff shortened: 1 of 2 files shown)


trishnaguha:
Looks good :+1: .

sayanchowdhury:
:+1:  for me

PR #1114, opened 2016-11-16
1111

bowlofeggs:
This pull request has two commits:

1. Move the functional updates tests to ```server/services/test_updates.py``` so they match the heirarchy of the files they are testing.
2. Split tests on ```new_update()``` into their own test case called TestNewUpdate.

None of the tests were changed.


puiterwijk (approved):
Looks good to me. :1234:

bowlofeggs:
Thanks!

PR #1117, opened 2016-11-16
Add tests, drop unreachable code, and fix #1111

### LINKED ISSUE #1111: Race condition between robosignatory signing builds and Builds being added to the db
The signed handler often sees messages from robosignatory that it has signed a build, but then the Build isn't found in Bodhi's database. This is due to the Build not being committed before Bodhi tells Koji to add the build to the pre-signing tag. This causes updates to never make it to the signed state, which causes the masher to never move them to the updates-testing repository.

The fix itself will be easy, but we will need to expend a lot of effort in testing the fix since it is unlikely that there are tests for the specific function in question.

bowlofeggs:
This pull request contains four commits. Each commit is atomic and is desirable to go in by itself (i.e., we shouldn't squash this pull request), but they are also all needed for the final commit which fixes #1111.

Incidentally, the last commit's addition of transactions (and in particular, the abortion of transactions) has reduced the test suite's memory usage by about 50%!

### DIFF
--- a/bodhi/server/services/updates.py
+++ b/bodhi/server/services/updates.py
@@ -18,6 +18,7 @@ import math
 from cornice import Service
 from sqlalchemy import func, distinct
 from sqlalchemy.sql import or_
+import transaction
 
 from bodhi.server import log
 from bodhi.server.exceptions import BodhiException, LockedUpdateException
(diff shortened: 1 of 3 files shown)


bodhi/server/services/updates.py
```diff
@@ -18,6 +18,7 @@
 from cornice import Service
 from sqlalchemy import func, distinct
 from sqlalchemy.sql import or_
+import transaction
```
pypingou, line 21: This is a top level import?

pypingou:
One question, the rest looks good to me, :+1:

bodhi/server/services/updates.py
bowlofeggs: Surprisingly, it is! It's a weird system. I've had quite a time learning how sqlalchemy, zope.sqlalchemy, and pyramid all come together to do transactions. They seem to use this transaction library to "register" transaction events and transaction handlers.

```
$ dnf provides /usr/lib/python2.7/site-packages/transaction
python-transaction-1.6.1-2.fc25.noarch : Transaction management for Python
```

bowlofeggs:
The fourth commit has been cherry picked to the 2.3 branch.

PR #1124, opened 2016-11-20
Fail horribly if we forget to setup the build system

puiterwijk:
Currently, if someone forgets to set the buildsys, we will silently (because who
ever sets up logging for "bodhi" if they're working in a shell?).
This commit makes sure that we require people to have setup a buildsys before we
will allow them to get a session.

### DIFF
--- a/bodhi/server/buildsys.py
+++ b/bodhi/server/buildsys.py
@@ -315,10 +315,15 @@ def get_krb_conf(config):
 def get_session():
     """ Get a new buildsystem instance """
     global _buildsystem
-    if not _buildsystem:
-        log.warning('No buildsystem configured; assuming testing')
-        return DevBuildsys()
+    if _buildsystem is None:
+        raise RuntimeError('Buildsys needs to be setup')
     return _buildsystem()
+
+
+def teardown_buildsystem():
+    global _buildsystem
+    _buildsystem = None
+    DevBuildsys.clear()
 
 
 def setup_buildsystem(settings):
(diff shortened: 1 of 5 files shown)


bowlofeggs (changes requested):
We need test coverage.

bodhi/server/buildsys.py
```diff
@@ -315,12 +315,16 @@ def get_krb_conf(config):
 def get_session():
     """ Get a new buildsystem instance """
     global _buildsystem
-    if not _buildsystem:
-        log.warning('No buildsystem configured; assuming testing')
-        return DevBuildsys()
+    assert _buildsystem is not None, 'Buildsys needs to be setup'
```
bowlofeggs: ```assert``` statements get stripped out in .pyo files - it's important to raise an Exception since we want the production code to behave this way.

bodhi/server/buildsys.py
```diff
@@ -337,6 +341,9 @@ def setup_buildsystem(settings):
         log.debug('Using DevBuildsys')
         _buildsystem = DevBuildsys
```
bowlofeggs: I suggest dropping this space for readability.

bowlofeggs (changes requested):
Let's use an Exception instead of an ```assert``` statement.

bodhi/server/buildsys.py
puiterwijk: Sorry, which space do you mean?

bowlofeggs: Sorry, I meant "newline", not "space".

bowlofeggs (approved):
Approved as-is, or you can drop that newline if you want to at your option. Thanks!

PR #1122, opened 2016-11-20
Disable caching koji sessions during mashing process

puiterwijk:
The mashing process takes a very long time: longer than the session timeout for the
bodhi user in koji.
Since a koji session is started on buildsys.get_session(), that means the koji session
often times out if the mash takes too long.
This patch makes us get a new session for each batch of koji operations.

This will incur a slight performance impact, but that should not be too big, since most
of the koji operations are batched, and this change gets a new session on each batch.

Signed-off-by: Patrick Uiterwijk <puiterwijk@redhat.com>

PR #1128, opened 2016-11-23
Convert bodhi.server.push to use get_db_factory()

trishnaguha:


bowlofeggs (changes requested):
This is a good change, but we will need test coverage to make sure that it is sane. The good news is that I am planning to write tests for this module as part of https://github.com/fedora-infra/bodhi/pull/1122 so I might just do this too while I'm doing that. So all that to say, we can't merge this now, but let's keep it open for when I write some tests (no action needed on your part).

trishnaguha:
@bowlofeggs ACK! Let me know if I can do anything for the PR if you don't have time :).

bowlofeggs:
I just spent 1.5 days doing a huge amount of test writing on push.py and a bit of refactoring. It was easier to write the tests if I went ahead and fixed this problem, but I needed to drop a few more lines too so I did that in #1158. Thanks for the PR, and I apologize for stepping on your toes - I just needed it fixed a slightly different way so I could get those tests written quickly!

PR #1129, opened 2016-11-23
Use krb5 for koji

puiterwijk:

### DIFF
--- a/bodhi/server/buildsys.py
+++ b/bodhi/server/buildsys.py
@@ -291,19 +291,24 @@ class DevBuildsys(Buildsystem):
 def koji_login(config):
     """ Login to Koji and return the session """
     koji_client = koji.ClientSession(_koji_hub, {})
-    if not koji_client.ssl_login(*get_certs(config)):
-        log.error('Koji ssl_login failed')
+    if not koji_client.krb_login(**get_krb_conf(config)):
+        log.error('Koji krb_login failed')
     return koji_client
 
 
-def get_certs(config):
-    """ Return paths to the local certs needed log into koji """
-    client = config.get('client_cert', join(expanduser('~'), '.fedora.cert'))
-    clientca = config.get('clientca_cert', join(expanduser('~'),
-                          '.fedora-upload-ca.cert'))
-    serverca = config.get('serverca_cert', join(expanduser('~'),
-                          '.fedora-server-ca.cert'))
-    return client, clientca, serverca
+def get_krb_conf(config):
+    """ Return arguments for krb_login. """
+    principal = config.get('krb_principal')
+    keytab = config.get('krb_keytab')
+    ccache = config.get('krb_ccache')
+    args = {}
+    if principal:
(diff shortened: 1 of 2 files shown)


bodhi/server/buildsys.py
```diff
@@ -291,19 +291,24 @@ def getRPMHeaders(self, rpmID, headers):
...
-                          '.fedora-upload-ca.cert'))
-    serverca = config.get('serverca_cert', join(expanduser('~'),
-                          '.fedora-server-ca.cert'))
-    return client, clientca, serverca
+def get_krb_conf(config):
+    """ Return arguments for krb_login. """
+    principal = config.get('krb_principal')
+    keytab = config.get('krb_keytab')
+    ccache = config.get('krb_ccache')
+    args = {}
```
bowlofeggs, line 304: nitpick: It'd be more conventional to call this kwargs since they are keyword args.

pypingou:
Does this need a change to the spec file to require a certain koji version?

bowlofeggs:
@pypingou I'm sure it does (and I'd like to know what version is required), but I don't use or maintain the spec file in this repo. I should just delete it… The only spec file I use and maintain is the one in dist-git.

bowlofeggs (changes requested):
We will need tests for this change.

puiterwijk:
@pypingou no need for spec changes. This has been part of Koji since a LONG time ago (I think this was the first auth method they had in 1.0).

puiterwijk:
"LONG time" being about 6+ years if I recall correctly.

bowlofeggs:
For reference, I did submit a PR to delete the spec file here since it's not used or maintained: https://github.com/fedora-infra/bodhi/pull/1133

bowlofeggs:
I'm going to write tests for this PR so we can get 2.3.3 released before @puiterwijk finds another bug ☺

bodhi/server/buildsys.py
```diff
@@ -291,19 +291,24 @@ def getRPMHeaders(self, rpmID, headers):
...
+    if principal:
+        args['principal'] = principal
+    if keytab:
+        args['keytab'] = keytab
+    if ccache:
+        args['ccache'] = ccache
```
bowlofeggs, line 310: Should it be an error if any of the above are missing from the config?

puiterwijk: No. There's various pairs that are valid, and koji will validate that, but it might very well work without having to configure anything in certain setups.

bowlofeggs (approved):

bowlofeggs:
I've got tests for this ready to send in on my dev box, so I'm going to go ahead and merge this PR as-is.

bowlofeggs:
The tests are here: https://github.com/fedora-infra/bodhi/pull/1135

bowlofeggs:
I've cherry picked this commit to the 2.3 branch as 681389f8ad2cfbeec8c3c51bc0a7a9644e754873.

PR #1135, opened 2016-11-29
Add test coverage for bodhi.server.buildsys.

bowlofeggs:
This commit adds two test suites that cover the koji_login() and
the get_krb_conf() functions.


pypingou:
:+1: for me

PR #1141, opened 2016-12-01
Add coverage.xml and nosetests.xml to gitignore.

bowlofeggs:


pypingou:
Tough one, but ok :+1:

trishnaguha:
We could do `*.xml` instead :-)?

bowlofeggs:
@trishnaguha I'd rather not do that in the case that we ever want to add .xml files to the repo in the future (test data perhaps?)

PR #1142, opened 2016-12-02
Lock the buildsystem while logging in

puiterwijk:

### DIFF
--- a/bodhi/server/buildsys.py
+++ b/bodhi/server/buildsys.py
@@ -14,6 +14,7 @@
 
 import time
 import logging
+from threading import Lock
 
 from os.path import join, expanduser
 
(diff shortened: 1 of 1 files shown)


bowlofeggs:
@puiterwijk Why do we want to do this? Also, this will need tests.

puiterwijk:
Because having two sessions that login at the exact same moment confuses kerberos.

bowlofeggs:
@puiterwijk Bodhi does have multiple processes, though I'm not sure which of them use the buildsys module. Also, I'm not sure how we have httpd configured in production (does it use processes and threads… just threads… just processes?) Since this commit only uses thread locking, should we be concerned about the processes causing this confusion?

puiterwijk:
The problematic part is the mash stuff. It uses multiple threads, and LOTS of calls to the buildsystem, so LOTS of chances to hit the race condition.
HTTPD side is not a problem.

bowlofeggs (changes requested):
This needs tests.

jeremycline:
We obviously need to get this into Bodhi for now, but it seems like upstream Koji would benefit from making the login function thread-safe. Do you know if there's an issue or PR for that? I don't see either, but that doesn't mean it doesn't exist, of course.

bowlofeggs:
On Wed, 2017-01-18 at 10:40 -0800, Jeremy Cline wrote:
> We obviously need to get this into Bodhi for now, but it seems like
> upstream Koji would benefit from making the login function thread-
> safe. Do you know if there's an issue or PR for that? I don't see
> either, but that doesn't mean it doesn't exist, of course.

I'm not sure, but I'm a huge +1 to the suggestion of sending this to
Koji. We can keep this code for now until they have it.

bodhi/server/buildsys.py
```diff
@@ -314,18 +315,20 @@ def get_krb_conf(config):
...
+        return _buildsystem()
 
 
 def setup_buildsystem(settings):
-    global _buildsystem, _koji_hub
+    global _buildsystem, _koji_hub, _buildsystem_login_lock
     if _buildsystem:
         return
 
+    _buildsystem_login_lock = Lock()
```
jeremycline, line 331: Out of curiosity, why not just initialize this at the module level rather than involving ``global`` and a function?

puiterwijk: To be consistent with the rest of the code.

jeremycline: Okay, cool. I'm inclined to change it to initialize at the module level since that's less surprising to me.

jeremycline:
I added unit tests on top of this commit in #1230 and had my question answered here so I'm going to go ahead and close this.

PR #1145, opened 2016-12-07
Two fixes on bodhi stable comments

### LINKED ISSUE #1009: Users are receiving repeat e-mails every 6 hours

### LINKED ISSUE #1094: No message when updates can be pushed stable if they have autokarma off and haven't reached karma
I noticed today that one of my F25 updates has been sitting in testing for 13 days, but did not get the message from Bodhi that it could be pushed to stable:

https://bodhi.fedoraproject.org/updates/FEDORA-2016-bf2a514f37

bowlofeggs:
This pull request contains two commits, one to fix #1094 and one to fix #1009. I recommend reviewing each commit independently. Normally I'd make two pull requests, but the second commit depended on the first.

### DIFF
--- a/bodhi/server/scripts/approve_testing.py
+++ b/bodhi/server/scripts/approve_testing.py
@@ -56,21 +56,22 @@ def main(argv=sys.argv):
             if update.met_testing_requirements:
                 continue
 
-            # If autokarma updates have reached the testing threshold, say something! Keep in mind
-            # that we don't care about karma here, because autokarma updates get their request set
-            # to stable by the Update.comment() workflow when they hit the required threshold. Thus,
-            # this function only needs to consider the time requirements because these updates have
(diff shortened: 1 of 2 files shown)


trishnaguha:
Nice. This would be helpful indeed :-).

pypingou:
:+1: for me as well :)

PR #1147, opened 2016-12-08
Test for repomd.xml instead of repodata directory

### LINKED ISSUE #908: Work around missing repodata.xml
Sometimes the previous run of the Bodhi masher leaves some directories with missing repodata.xml files (See #887). That issue is about stopping that from happening in the future, but it is a bit mysterious and may take some time to identify its cause.

In the meantime, we can make Bodhi more robust. In subsequent runs of the masher, there is a test for the existence of the folder that should contain a repodata.xml file and if that folder is present Bodhi assumes that the file is present as well. This assumption is false, and it causes Bodhi to blow up later.

We should adjust that test to test for the existence of the file instead of the folder. This will make Bodhi more robust for releng, and should be a pretty easy change as well.

bowlofeggs:
This pull request contains two commits. The first is from @k3rn3l3rr0r and fixes #908. The second is from me and tests the fix to ensure it works properly.

### DIFF
--- a/bodhi/server/metadata.py
+++ b/bodhi/server/metadata.py
@@ -74,7 +74,7 @@ class ExtendedMetadata(object):
         # Load from the cache if it exists
         self.cached_repodata = os.path.join(self.repo, '..', self.tag +
                                             '.repocache', 'repodata/')
-        if os.path.isdir(self.cached_repodata):
+        if os.path.isfile(os.path.join(self.cached_repodata, 'repomd.xml')):
             log.info('Loading cached updateinfo.xml')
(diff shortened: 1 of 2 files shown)


pypingou:
LGTM :+1:

k3rn3l3rr0r:
@bowlofeggs I definitely lost this one from my radar. I'm sorry for this. Next time things will come with tests included. Thanks for finishing it.

bowlofeggs:
@k3rn3l3rr0r No problem, I still appreciate the patch you sent. Thanks!

PR #1148, opened 2016-12-09
Add a test on FakeBugTracker.getbug().

bowlofeggs:


bodhi/tests/server/test_bugs.py
```diff
@@ -17,6 +17,8 @@
 
 import unittest
 
+import bunch
```
pypingou, line 20: I assume this is already in the dependency list

pypingou:
Assuming my assumption is true, then :+1: for me :)

bowlofeggs:
On Fri, 2016-12-09 at 00:52 -0800, Pierre-Yves Chibon wrote:
> I assume this is already in the dependency list

I sure hope so, since Bodhi is using it in production now!

(I checked the ``setup.py`` and Fedora's ``bodhi.spec`` and bunch
is required in both ☺)

PR #1149, opened 2016-12-09
Raise Bodhi's test coverage to 78%

bowlofeggs:
This pull request contains two commits. The first commit adds three new tests on ```bodhi.server.captcha.jpeg_generator```. Those tests happen to raise the overall test coverage to 78% so the second commit adjusts the test suite to require 78% coverage so we don't slip back down in the future.


pypingou:
Cool :+1:

PR #1150, opened 2016-12-09
Add a database schema image to the docs

jeremycline:
This generates a schema diagram from the SQLAlchemy models.

### DIFF
--- a/devel/ansible/roles/dev/tasks/main.yml
+++ b/devel/ansible/roles/dev/tasks/main.yml
@@ -8,6 +8,7 @@
       - freetype-devel
       - gcc
       - git
+      - graphviz
       - koji
       - liberation-mono-fonts
       - libffi-devel
@@ -35,6 +36,9 @@
       - python2-fedmsg-consumers
       - python2-flake8
       - python2-nose-cov
+      # We can switch this to python2-sqlalchemy_schemadisplay once
+      # https://bodhi.fedoraproject.org/updates/FEDORA-2017-e54acedb77 is stable
+      - https://kojipkgs.fedoraproject.org//packages/python-sqlalchemy_schemadisplay/1.3/1.fc24/noarch/python2-sqlalchemy_schemadisplay-1.3-1.fc24.noarch.rpm
       - redhat-rpm-config
       - zlib-devel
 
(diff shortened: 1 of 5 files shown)


jeremycline:
It looks like this
database

devel/ansible/roles/dev/tasks/main.yml
```diff
@@ -35,6 +36,11 @@
       - redhat-rpm-config
       - zlib-devel
 
+- name: Install sqlalchemy_schemadisplay
+  command: pip install sqlalchemy_schemadisplay
+  args:
+      creates: /usr/lib/python2.7/site-packages/sqlalchemy_schemadisplay.py
```
bowlofeggs: I assume this isn't packaged for Fedora? If so, we will need to package it, because the bodhi package in Fedora builds the docs. If this isn't present, will the docs build fail?

bowlofeggs:
This looks great! I think we should get that sqlalchemy thing packaged for Fedora so the docs will build. I don't mind taking that on, unless you want to do it. Until that is present in Fedora/EPEL 7, I think we need to hold this PR so we don't end up having it be a release blocker.

bowlofeggs:
I talked to @jeremycline in person and I agreed to package ```sqlalchemy_schemadisplay``` myself, and he agreed to review it. We will have to hold the PR until it is in Fedora.

devel/ansible/roles/dev/tasks/main.yml
bowlofeggs: @jeremycline Can you move this to be installed by RPM in the list above, with the assumption that it will be called ```python2-sqlalchemy_schemadisplay```?

bowlofeggs (changes requested):

bowlofeggs:
The package review is now approved, so we just need to get it built and released for the Vagrant environment to merge this. If you want, you can use the koji ```https://``` URL until it's actually in Fedora.

jeremycline:
I have built it for Rawhide through EPEL 7 and updated the PR to install the RPM package from Koji with a note about changing it when the F24 update is stable.

bowlofeggs (approved):
Thanks @jeremycline!

Issue #1146, opened 2016-12-08
Login and logout are nouns and not verbs


andreasn:
Right now the actions for logging in and logging out are called "Login" and "Logout", so it sounds like they are nouns "A logout" instead of a verb, ie "I want to log out".
This clashes a bit when you come to the FAS page, where it says "Log in".

I would suggest calling it "Log in" and "Log out" instead.

Also see https://www.quora.com/Diction-and-Word-Usage-Which-is-correct-Login-or-Log-In

PR #1151, opened 2016-12-09
fixes issue #1146 login logout

anazli:


bowlofeggs (changes requested):
This change does introduce a test failure:

```
FAIL: test_home (bodhi.tests.server.functional.test_generic.TestGenericViews)
  File "/vagrant/bodhi/tests/server/functional/test_generic.py", line 101, in test_home
    self.assertIn('Logout', res)
AssertionError: 'Logout' not found in <200 OK text/html body='\n\n\n\n\...l>\n'/9475>
```

It should be pretty easy to fix that test to look for the new string instead.

anazli:
I've just changed the file test_generic.py

bowlofeggs:
Thanks @anazli, the tests pass now!

Issue #1160, opened 2016-12-14
[RFE] Close BugZilla Bugs based on specific or overall karma votes


anassahmed:
Currently, Bodhi closes the related bugs of an update if it hit the stable repositories.

This is not always correct, especially if reporters had marked specific bug (in bodhi) as not fixed in its karma vote, but the overall karma vote threshold made it pass to stable repositories.

We need to make bodhi close only bugs which doesn't have negative votes (specifically for it) or its positive votes count is higher than the negative ones (specific for this bug).

PR #1162, opened 2016-12-15
Close bugs based on their bug karma votes

anassahmed:
Based on the issue #1160, This pull request checks if the related bugs to the update have negative bug feedback so it doesn't close them automatically, but still sends the comment to notify the bug followers about the new _stable_ release.

The update could be _generally functional_, but it didn't fix this specific bug based on testers feedback, assuming the update hit the stable repos and it marked this bug as a fix as enough reason for closure isn't correct especially if we have the bug feedback implemented in bodhi.

### DIFF
--- a/bodhi/server/models/models.py
+++ b/bodhi/server/models/models.py
@@ -1022,15 +1022,19 @@ class Update(Base):
     def get_bug_karma(self, bug):
         good, bad, seen = 0, 0, set()
         for comment in reversed(self.comments):
-            if comment.user.name in seen:
-                continue
-            seen.add(comment.user.name)
-            for feedback in comment.bug_feedback:
-                if feedback.bug == bug:
-                    if feedback.karma > 0:
-                        good += 1
-                    elif feedback.karma < 0:
-                        bad += 1
+            if comment.user:
+                if (comment.user.name == u'bodhi' and
+                        ('New build' in comment.text or 'Removed build' in comment.text)):
+                    break
+                if comment.user.name in seen:
+                    continue
+                seen.add(comment.user.name)
+                for feedback in comment.bug_feedback:
+                    if feedback.bug == bug:
+                        if feedback.karma > 0:
+                            good += 1
+                        elif feedback.karma < 0:
+                            bad += 1
         return bad * -1, good
 
     def get_testcase_karma(self, testcase):
@@ -1260,11 +1264,26 @@ class Update(Base):
                     # https://github.com/fedora-infra/bodhi/issues/368#issuecomment-135155215
                     for bug in self.bugs:
                         if not bug.parent:
-                            log.debug("Closing tracker bug %d" % bug.bug_id)
-                            bug.close_bug(self)
+                            bad, good = self.get_bug_karma(bug)
+                            bug_karma = good + bad
+                            if bug_karma >= 0:
+                                log.debug("Closing tracker bug %d" % bug.bug_id)
+                                bug.close_bug(self)
+                            else:
+                                log.debug('Adding stable comment to bugs for %s', self.title)
+                                bug.add_comment(self)
                 else:
                     for bug in self.bugs:
-                        bug.close_bug(self)
+                        # Close only when the bug has good or no karma
+                        # https://github.com/fedora-infra/bodhi/issues/1160
+                        bad, good = self.get_bug_karma(bug)
+                        bug_karma = good + bad
+                        if bug_karma >= 0:
+                            log.debug("Closing tracker bug %d" % bug.bug_id)
+                            bug.close_bug(self)
+                        else:
+                            log.debug('Adding stable comment to bugs for %s', self.title)
+                            bug.add_comment(self)
 
     def status_comment(self, db):
         """
--- a/bodhi/tests/server/models/test_models.py
+++ b/bodhi/tests/server/models/test_models.py
@@ -277,6 +277,48 @@ class TestUpdate(ModelTest):
 
     @mock.patch('bodhi.server.models.models.bugtracker.close')
     @mock.patch('bodhi.server.models.models.bugtracker.comment')
+    def test_modify_bugs_karma_stable_close(self, comment, close):
+        """Test the modify_bugs() method with a stable status and bug karmas effect."""
+        update = self.get_update()
+        bug_1 = model.Bug(bug_id=1)
+        bug_2 = model.Bug(bug_id=2)
+        update.bugs.append(bug_1)
+        update.bugs.append(bug_2)
+        update.close_bugs = True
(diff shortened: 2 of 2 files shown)


bodhi/server/models/models.py
```diff
@@ -1022,15 +1022,19 @@ def get_cvestring(self):
...
-            if comment.user.name in seen:
-                continue
-            seen.add(comment.user.name)
-            for feedback in comment.bug_feedback:
-                if feedback.bug == bug:
-                    if feedback.karma > 0:
-                        good += 1
-                    elif feedback.karma < 0:
-                        bad += 1
+            if comment.user:
```
bowlofeggs, line 1025: I would expect that comments would always have users. Have you seen that not to be the case?

bodhi/server/models/models.py
```diff
@@ -1022,15 +1022,19 @@ def get_cvestring(self):
...
+                if (comment.user.name == u'bodhi' and
+                        ('New build' in comment.text or 'Removed build' in comment.text)):
+                    break
```
bowlofeggs, line 1028: I've actually not looked at this method before, but the code that counts "normal karma" resets the counts when it encounters this comment. Instead of breaking, you might want to reset the good, bad, and seen counts.

Also, this line doesn't have test coverage - can you add a second test case to cover this situation?

bowlofeggs: Also, it seems like a bug to me that this method never had this check in it, so thanks for adding it! I do believe it is a separate concern from the purpose of your commit though - would you mind putting this (and a test for it) into a second commit so the commits are atomic and focused on one thing each? You can keep it in the same pull request if you prefer. You can also make it a separate pull request if you like that approach better.

bodhi/server/models/models.py
```diff
@@ -1260,11 +1264,26 @@ def modify_bugs(self):
...
-                            log.debug("Closing tracker bug %d" % bug.bug_id)
-                            bug.close_bug(self)
+                            bad, good = self.get_bug_karma(bug)
+                            bug_karma = good + bad
+                            if bug_karma >= 0:
+                                log.debug("Closing tracker bug %d" % bug.bug_id)
+                                bug.close_bug(self)
+                            else:
+                                log.debug('Adding stable comment to bugs for %s', self.title)
+                                bug.add_comment(self)
```
bowlofeggs, line 1274: Suggestion: it might be good to add some text in this comment that mentions on the BZ that though the update went stable, there was negative feedback about this bug and so the bug isn't being closed with a suggestion that someone should continue to look into the bug. Maybe something like:

```
<normal update moved to stable text>

However, negative feedback was received about this bug and so the Fedora Update System is leaving the bug
open for further review.
```

Feel free to use whatever text you think makes sense if you want to do this, and it's just a suggestion.

bodhi/server/models/models.py
```diff
@@ -1260,11 +1264,26 @@ def modify_bugs(self):
                     # https://github.com/fedora-infra/bodhi/issues/368#issuecomment-135155215
                     for bug in self.bugs:
                         if not bug.parent:
-                            log.debug("Closing tracker bug %d" % bug.bug_id)
-                            bug.close_bug(self)
+                            bad, good = self.get_bug_karma(bug)
+                            bug_karma = good + bad
```
bowlofeggs, line 1268: Suggestion: You could replace these two lines with a single but still readable line: ```bug_karma = sum(self.get_bug_karma(bug))``` since ```bad``` and ```good``` are just intermediate variables.

bodhi/server/models/models.py
```diff
@@ -1022,15 +1022,19 @@ def get_cvestring(self):
...
-                    if feedback.karma > 0:
-                        good += 1
-                    elif feedback.karma < 0:
-                        bad += 1
+            if comment.user:
+                if (comment.user.name == u'bodhi' and
+                        ('New build' in comment.text or 'Removed build' in comment.text)):
+                    break
+                if comment.user.name in seen:
+                    continue
```
bowlofeggs, line 1030: This line is missing test coverage, but it seems like was a pre-existing line so I don't think we have to add coverage for it unless you want bonus points ☺

bodhi/server/models/models.py
```diff
@@ -1260,11 +1264,26 @@ def modify_bugs(self):
...
+                        # Close only when the bug has good or no karma
+                        # https://github.com/fedora-infra/bodhi/issues/1160
+                        bad, good = self.get_bug_karma(bug)
+                        bug_karma = good + bad
+                        if bug_karma >= 0:
+                            log.debug("Closing tracker bug %d" % bug.bug_id)
+                            bug.close_bug(self)
+                        else:
+                            log.debug('Adding stable comment to bugs for %s', self.title)
+                            bug.add_comment(self)
```
bowlofeggs, line 1286: These lines seem to be missing test coverage.

bowlofeggs: This code block seems to be identical to the one above (in the other half of the if statement) - let's make it a helper method so that the code isn't repeated (and so that we have guaranteed consistency in the behaviors if anyone changes one of them in the future.)

bowlofeggs (changes requested):
Thanks for the contribution!

I've left a few comments. Anything that says it's a suggestion is optional so you can take it or leave it at your option.

Also, I'm very pleased that you wrote a test for this! Nice work!

bodhi/server/models/models.py
bowlofeggs: I do believe that comments will always have users:

https://github.com/fedora-infra/bodhi/blob/2.3.3/bodhi/server/models/models.py#L1966

I think we should drop this if statement.

anassahmed:
Thanks @bowlofeggs for taking your time to review this PR.

I'll get to work on these suggestions as soon as I've some free time, then I'll squash the commits into 2 commits to be as atomic as possible.

Thanks again.

bodhi/server/models/models.py
anassahmed: I felt that adding a text was a daunting job. the comment is added to the bug based on the update state, but also handled at the end by the masher with fedmsg, and needed to add that text to the *.ini files.

I don't know how fedmsg works and how to pass a certain value for its context to be able to decide if this text will be added to this bug or not.

bowlofeggs: @anassahmed yeah that does sound like a lot of work - don't worry about it. Maybe I'll do it later in a follow on commit ☺

bowlofeggs:
Hello @anassahmed, do you still intend to work on this pull request?

bowlofeggs:
I'm going to close this for now. If you want to continue, feel free to reopen the pull request!

PR #1163, opened 2016-12-16
Set krb_rdns to False

puiterwijk:


bowlofeggs:
Thanks!

This code already has test coverage, but doesn't have strong assertions about koji configs. I wrote the tests, and I recall it being difficult to get that data out of the koji library in the test suite. Therefore, I think it's OK to merge this as-is since it's basically a config change and the code is covered.

Issue #1165, opened 2016-12-17
bodhi.server.util.flash_log does not flash


bowlofeggs:
The name and docblock of ```bodhi.server.util.flash_log``` says that it flashes and logs a message, but it only logs the message.

ankit01ojha:
@bowlofeggs Is this issue still open?

bowlofeggs:
A fix for this issue is planned for inclusion in the upcoming 4.0.0 release: #3221

bowlofeggs:
Bodhi 4.0.0 beta is built and deployed to staging:

https://copr.fedorainfracloud.org/coprs/bowlofeggs/bodhi-pre-release
https://bodhi.stg.fedoraproject.org/

PR #1167, opened 2016-12-17
PEP-8 bodhi.server.util.

bowlofeggs:


bodhi/server/util.py
```diff
@@ -16,47 +16,50 @@
...
-header = lambda x: u"%s\n     %s\n%s\n" % ('=' * 80, x, '=' * 80)
 
-pluralize = lambda val, name: val == 1 and name or "%ss" % name
+def header(x):
+    """Display a given message as a heading."""
+    return u"%s\n     %s\n%s\n" % ('=' * 80, x, '=' * 80)
+
+
+def pluralize(val, name):
+    return val == 1 and name or "%ss" % name
```
pypingou, line 65: that got me confused a little, still kinda do, but somehow seems to work

(misses the docstring btw)

bodhi/server/util.py
```diff
@@ -126,8 +129,6 @@ def get_age_in_days(date):
 
 def flash_log(msg):
     """ Flash and log a given message """
-    # FIXME: request.session.flash()
-    #flash(msg)
     log.debug(msg)
```
pypingou, line 135: Should this method be entirely dropped then?

pypingou:
Couple of comments, rest looks good

bowlofeggs:
On Mon, 2016-12-19 at 01:50 -0800, Pierre-Yves Chibon wrote:
> that got me confused a little, still kinda do, but somehow seems to
> work
> (misses the docstring btw)

flake8 gets upset if lambdas are assigned to a variable since that
defeats the purpose of lambdas, so this commit is adjusting them to be
normal functions since that's what they really are.

I'll add the docstring.

bowlofeggs:
On Mon, 2016-12-19 at 01:51 -0800, Pierre-Yves Chibon wrote:
> Should this method be entirely dropped then?

It does get used, and I filed an issue to get the flashing working:

https://github.com/fedora-infra/bodhi/issues/1165

pypingou:
> flake8 gets upset if lambdas are assigned to a variable since that defeats the purpose of lambdas, so this commit is adjusting them to be normal functions since that's what they really are.

No, I got the move from lambda to a method (I don't really like lambda either) but I really meant the construct used here, if I had been writing this, I would have used an ``return foo if foo else bar`` or so. The current construct looks a little magical to me (but it seems to work, so cool :))

bowlofeggs:
@pypingou ah yeah, I agree that it is strange (I didn't write it) but yeah, it does seem to work. I guess ```True and "bowlofeggs"``` resolves to ```"bowlofeggs"```… strange!

PR #1176, opened 2016-12-19
Adjust a test to allow packages to be printed in any order.

### LINKED ISSUE #1171: Unittest failure due to sorting order
@puiterwijk reported that one unit test is failing on his system:

```
FAIL: Test correct operation when there are some locked updates that aren't in a push.
  File "/vagrant/bodhi/tests/server/test_push.py", line 442, in test_locked_updates_not_in_a_push
    self.assertEqual(result.output, TEST_LOCKED_UPDATES_NOT_IN_A_PUSH_EXPECTED_OUTPUT)
AssertionError: u'Warning: ejabberd-16.09-4.fc17 is locked but not in a push\nejabberd-16.09-4.fc17\npython-nose-1.3.7-11.fc17\npython-paste-deploy-1.5.2-8.fc17\nPush these 3 updates? [y/N]: y\n\nLocking updates...\n\nSending masher.start fedmsg\n' != 'Warning: ejabberd-16.09-4.fc17 is locked but not in a push\npython-nose-1.3.7-11.fc17\npython-paste-deploy-1.5.2-8.fc17\nejabberd-16.09-4.fc17\nPush these 3 updates? [y/N]: y\n\nLocking updates...\n\nSending masher.start fedmsg\n'
```

This may be due to locale, but in general we don't need to assert that the packages are given in that particular order, just that they are given in some order. We need to refactor the test to allow the package order to be shuffled any way.

bowlofeggs:
fixes #1171

### DIFF
--- a/bodhi/tests/server/test_push.py
+++ b/bodhi/tests/server/test_push.py
@@ -442,7 +442,17 @@ class TestPush(base.BaseTestCase):
             result = cli.invoke(push.push, ['--username', 'bowlofeggs'], input='y')
 
         self.assertEqual(result.exit_code, 0)
-        self.assertEqual(result.output, TEST_LOCKED_UPDATES_NOT_IN_A_PUSH_EXPECTED_OUTPUT)
+        # The packages might be printed in any order and the order isn't important. So let's compare
+        # the output with the package list removed to make sure that it is correct.
+        self.assertEqual(
+            '\n'.join([l for l in result.output.split('\n') if l[-4:] != 'fc17']),
+            '\n'.join([l for l in TEST_LOCKED_UPDATES_NOT_IN_A_PUSH_EXPECTED_OUTPUT.split('\n')
+                       if l[-4:] != 'fc17']))
+        # Now let's assert that the package list is correct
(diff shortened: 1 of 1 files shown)


puiterwijk:
How about just assertEquals(set(a), set(b)) ?

bowlofeggs:
@puiterwijk This way it does still check the order of the remaining output. It's fancier and still will work for arbitrary package ordering ☺

pypingou:
Well, set() does not keep the ordering iirc, so using a sorted(set()) or a plain list() would be more stable I believe.

bowlofeggs:
@puiterwijk, @pypingou: So the test as I've written it now is the best test I can think of - it ensures the correct ordering of all output that isn't the package list, and it uses a set to compare the package list. The order of the package list has no meaning for this CLI output, but the rest of the CLI output *does* have significance, so it really is best to test it this way. I think we should merge this as is.

pypingou:
It's fine with me, I was merely nit-picking on the use of set() for something that required ordering :)
If the ordering in the tests isn't an issue, then it's def all fine for me

PR #1184, opened 2016-12-23
Do not print the Update Munch when rendering updates.

bowlofeggs:
This fixes https://bugzilla.redhat.com/show_bug.cgi?id=1408195


pypingou:
Sure 👍

PR #1194, opened 2017-01-03
Add a dependency to the Vagrant setup and update another.

bowlofeggs:
This commit adds python2-fedmsg-consumers to the Vagrant
environment. It also updates the Vagrant environment to
python2-fedmsg-atomic-composer-2016.3 by getting it directly from
Koji.


jeremycline (approved):
Looks good :shipit:

PR #1195, opened 2017-01-03
Do not depend on it being 2016 for the tests to pass.

bowlofeggs:


jeremycline (approved):
Happy New Year! 🎆

Looks good :shipit:

PR #1198, opened 2017-01-04
Make the Vagrantfile.example fancier

jeremycline:

### DIFF
--- a/Vagrantfile.example
+++ b/Vagrantfile.example
@@ -8,24 +8,66 @@
 # vagrant up
 # vagrant ssh -c "cd /vagrant/; pserve development.ini --reload"
 
-Vagrant.configure(2) do |config|
-  config.vm.box_url = "https://download.fedoraproject.org/pub/fedora/linux/releases/24/CloudImages/x86_64/images/Fedora-Cloud-Base-Vagrant-24-1.2.x86_64.vagrant-libvirt.box"
-  config.vm.box = "f24-cloud-libvirt"
-  config.vm.network "forwarded_port", guest: 6543, host: 6543
-  config.vm.synced_folder ".", "/vagrant", type: "sshfs"
+VAGRANTFILE_API_VERSION = "2"
 
-  config.vm.provider :libvirt do |domain|
-      domain.cpus = 4
-      domain.graphics_type = "spice"
-      # The unit tests use a lot of RAM.
-      domain.memory = 4096
-      domain.video_type = "qxl"
-  end
+Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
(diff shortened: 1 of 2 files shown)


sayanchowdhury:
:+1:

Vagrantfile.example
```diff
@@ -8,24 +8,65 @@
...
+ # ``vagrant-hostmanager``
+ if Vagrant.has_plugin?("vagrant-hostmanager")
+     config.hostmanager.enabled = true
+     config.hostmanager.manage_host = true
+ end
+
+ # Vagrant can share the source directory using rsync, NFS, or SSHFS (with the vagrant-sshfs
+ # plugin). Consult the Vagrant documentation if you do not want to use SSHFS.
+ config.vm.synced_folder ".", "/vagrant", disabled: true
+ config.vm.synced_folder ".", "/vagrant", type: "sshfs", sshfs_opts_append: "-o nonempty"
```
bowlofeggs, line 33: Why don't we go ahead and switch this to ```/home/vagrant/bodhi``` or something? I've been annoyed at working in ```/vagrant```…

Vagrantfile.example
```diff
@@ -8,24 +8,65 @@
...
+
+ # Create the bodhi dev box
+ config.vm.define "bodhi" do |bodhi|
+    bodhi.vm.host_name = "bodhi-dev.example.com"
+
+    bodhi.vm.provider :libvirt do |domain|
+        # Season to taste
+        domain.cpus = 4
+        domain.graphics_type = "spice"
+        domain.memory = 4096
```
bowlofeggs, line 61: Maybe leave the comment that was here about the tests using a lot of RAM.

bowlofeggs (approved):
Two suggestions - take 'em or leave 'em. Thanks!

Vagrantfile.example
jeremycline: I wanted to, but it involves changing all the Ansible stuff that expects that path explicitly. I figured it'd be best to do in a follow-up PR

bowlofeggs:
On Thu, 2017-01-05 at 06:31 -0800, Jeremy Cline wrote:
> I wanted to, but it involves changing all the Ansible stuff that
> expects that path explicitly. I figured it'd be best to do in a
> follow-up PR

Oh right, good call.

Vagrantfile.example
bowlofeggs: I'm going to make that follow-up PR now, to move ```/vagrant``` to ```/home/vagrant/bodhi```.

bowlofeggs: This PR moves the code to ```~/bodhi```: https://github.com/fedora-infra/bodhi/pull/1209

PR #1199, opened 2017-01-04
Add a unit test for bodhi.server.scripts.expire_overrides.

bowlofeggs:


pypingou:
Fair, 👍

PR #1204, opened 2017-01-04
Configure the Koji client to retry

### LINKED ISSUE #1201: Ensure the masher's Koji interactions are fault-tolerant
This is an expansion on https://github.com/fedora-infra/bodhi/issues/1143.

The masher is a long-running task that needs to handle failures gracefully. The masher's interactions with Koji need to work despite intermittent network problems. This includes authentication, which involves the Kerberos server.

- [x] Ensure calls to Koji are retried in the event of network problems

- [ ]  Ensure that if there is an authentication problem, it is retried

jeremycline:
Although the Koji client has a few defaults for the options set here, I
set them all explicitly. The difference here is that ``offline_retry``
is now ``True``. It's not entirely clear from the Koji code what cases
this covers, but it should help some with the reliability problems. This
commit does not handle problems with the Kerberos server.

Fixes #1201 in part

### DIFF
--- a/bodhi/server/buildsys.py
+++ b/bodhi/server/buildsys.py
@@ -291,7 +291,17 @@ class DevBuildsys(Buildsystem):
 
 def koji_login(config):
     """ Login to Koji and return the session """
-    koji_client = koji.ClientSession(_koji_hub, {'krb_rdns': False})
+
+    koji_options = {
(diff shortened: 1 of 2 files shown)


bowlofeggs (approved):
Awesome, thanks!

PR #1207, opened 2017-01-05
PEP-8 bodhi.tests.server.models.

bowlofeggs:

### DIFF
--- a/bodhi/tests/server/models/test_models.py
+++ b/bodhi/tests/server/models/test_models.py
@@ -13,26 +13,24 @@
 # You should have received a copy of the GNU General Public License
 # along with this program; if not, write to the Free Software
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
-
-"""Test suite for the Bodhi models"""
-
+"""Test suite for bodhi.server.models.models."""
+from datetime import datetime, timedelta
 import time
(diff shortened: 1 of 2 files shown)


bodhi/tests/server/models/test_models.py
```diff
@@ -13,26 +13,24 @@
 # You should have received a copy of the GNU General Public License
 # along with this program; if not, write to the Free Software
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
-
-"""Test suite for the Bodhi models"""
-
+"""Test suite for bodhi.server.models.models."""
```
pypingou, line 16: models.models?

pypingou:
One comment, otherwise 👍

bodhi/tests/server/models/test_models.py
bowlofeggs: @pypingou This is actually currently correct:

https://github.com/fedora-infra/bodhi/blob/2.3.3/bodhi/server/models/models.py

#1057 has become about fixing that and replacing it with a simple ```bodhi.server.models```.

pypingou: Ok cool, just wanted to make sure it wasn't a typo, my 👍 below still stands :)

Issue #1126, opened 2016-11-22
Authentication errors print a traceback in the bodhi CLI


bowlofeggs:
If you type the wrong password when performing an action with the bodhi CLI, it will print a traceback:

```
Traceback (most recent call last):
  File "/usr/bin/bodhi", line 9, in <module>
    load_entry_point('bodhi-client', 'console_scripts', 'bodhi')()
  File "/usr/lib/python2.7/site-packages/click/core.py", line 716, in __call__
    return self.main(*args, **kwargs)
  File "/usr/lib/python2.7/site-packages/click/core.py", line 696, in main
    rv = self.invoke(ctx)
  File "/usr/lib/python2.7/site-packages/click/core.py", line 1060, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/lib/python2.7/site-packages/click/core.py", line 1060, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/lib/python2.7/site-packages/click/core.py", line 889, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/lib/python2.7/site-packages/click/core.py", line 534, in invoke
    return callback(*args, **kwargs)
  File "/vagrant/bodhi/client/__init__.py", line 620, in request
    resp = client.request(update, state)
  File "/vagrant/bodhi/client/__init__.py", line 58, in wrapper
    result = method(*args, **kwargs)
  File "/vagrant/bodhi/client/__init__.py", line 153, in request
    'csrf_token': self.csrf()})
  File "/vagrant/bodhi/client/__init__.py", line 58, in wrapper
    result = method(*args, **kwargs)
  File "/vagrant/bodhi/client/__init__.py", line 293, in csrf
    self.login(self.username, self.password)
  File "/usr/lib/python2.7/site-packages/fedora/client/openidbaseclient.py", line 289, in login
    openid_insecure=self.openid_insecure)
  File "/usr/lib/python2.7/site-packages/fedora/client/openidproxyclient.py", line 138, in openid_login
    raise AuthError(output['message'])
fedora.client.AuthError: Authentication failed
```

It should just print a nice error message instead.

ankit01ojha:
I want to work on this can you please tell me how should I start.
Thanks.

trishnaguha:
[...]

PR #1211, opened 2017-01-07
Removed the traceback  #Fixed 1126

ankit01ojha:
Removed the traceback and printed a message instead of it.

### DIFF
--- a/bodhi/client/__init__.py
+++ b/bodhi/client/__init__.py
@@ -80,7 +80,8 @@ def new(user, password, **kwargs):
         except bindings.BodhiClientException as e:
             click.echo(str(e))
         except Exception as e:
-            traceback.print_exc()
+            sys.tracebacklimit = 0
+	    print("Wrong password, Try again")
 
 
 @updates.command()


ankit01ojha:
@bowlofeggs Is this ok :)

jeremycline:
Hi @ankit01ojha, thanks for the contribution!

In this case, I think it's appropriate to print the traceback, since it's a general ``Exception`` block. Is there an issue associated with this pull request?

ankit01ojha:
@jeremycline yes, there it is https://github.com/fedora-infra/bodhi/issues/1126

bowlofeggs:
@ankit01ojha Can you move the "Fixed 1126" out of the subject line and into the commit message body? Also, I think GitHub expects it to be "fixes" instead of "Fixed". So a good commit message would be:

```
Do not traceback when CLI users enter the wrong password.

fixes #1126
```

This way GitHub will automatically close #1126 when the commit is merged.

ankit01ojha:
@bowlofeggs Thanks :)

bodhi/client/__init__.py
```diff
@@ -80,7 +80,8 @@ def new(user, password, **kwargs):
         except bindings.BodhiClientException as e:
             click.echo(str(e))
         except Exception as e:
-            traceback.print_exc()
```
jeremycline, line 83: I think we should keep this except block as it is since it's a general (unexpected) exception and that means it's a bug.

bodhi/client/__init__.py
```diff
@@ -80,7 +80,8 @@ def new(user, password, **kwargs):
...
+            sys.tracebacklimit = 0
+	    print("Wrong password, Try again")
```
jeremycline, line 84: The indentation is not a multiple of 4 here

jeremycline (changes requested):
I think a better approach is to add an ``except`` block for the specific exception raised on authentication failure. According to the issue, that's ``fedora.client.AuthError``

bodhi/client/__init__.py
bowlofeggs, line 84: Let's use the click library for I/O instead of print statements, as it makes testing easier.

bowlofeggs (changes requested):
I agree with @jeremycline, we want to keep the except block the way it is and add a new handler for the authentication exception.

Additionally, we need tests for this change.

bodhi/client/__init__.py
ankit01ojha: @bowlofeggs I'am trying to make it as you both suggested. :)

ankit01ojha:
@bowlofeggs I tried to work on it but I am not able to figure out how I do this, could you just help me bit .

Issue #1174, opened 2016-12-19
OSTree failures should not block RPM pushes from going out


bowlofeggs:
Currently, any failure in OSTree will block release engineering from pushing out RPMs. This could be bad if there were important security updates.

We need to refactor the masher such that:

* Failures in OSTree do not block RPM pushes from going to users (and vice-versa),
* Errors in OSTree and RPM mashing should be independently reported,
* The overall mash success should only be True if both succeeded.

bowlofeggs:
It was suggested by @ausil in #1214 that the updates-testing pushes should be blocked by ostree failures but the updates pushes should not be blocked by them.

ausil:
opposite way around, updates-testing should not be blocked on failures but updates should be

bowlofeggs:
@ausil Thanks for the clarification, and that makes sense to me.

bowlofeggs:
Now that we are using Pungi to do all the mashing, we are no longer in control of what is considered success or failure. I'm going to close this as a result.

PR #1214, opened 2017-01-10
Log, but don't fail if ostree compose fails

puiterwijk:

### DIFF
--- a/bodhi/server/consumers/masher.py
+++ b/bodhi/server/consumers/masher.py
@@ -938,7 +938,7 @@ class MasherThread(threading.Thread):
                 notifications.publish(topic="ostree.compose.fail",
                                       msg=dict(tag=tag),
                                       force=True)
-                raise Exception('%s atomic compose failed' % tag)
+                self.log.error('%s atomic tree compose failed', tag)
             else:
                 self.log.info('%s atomic tree compose successful', tag)
                 notifications.publish(


bodhi/server/consumers/masher.py
```diff
@@ -938,7 +938,7 @@ def compose_atomic_trees(self):
                 notifications.publish(topic="ostree.compose.fail",
                                       msg=dict(tag=tag),
                                       force=True)
-                raise Exception('%s atomic compose failed' % tag)
+                self.log.error('%s atomic tree compose failed', tag)
```
bowlofeggs, line 941: As you might expect, we don't have test coverage for this line. I volunteer to write tests, but it may take me some time.

ausil:
we have intentionally linked the ostree and updates, updating, this is not a change that we can accpet

bowlofeggs:
@ausil, I believe we did this so that we could still release RPMs when there are problems with the OSTree generation, so that in cases where OSTree's tooling were broken we would not be blocked for pushing out potentially critical RPM updates. This pull request would have us log the atomic compose rather than have it stop RPMs from going out to the mirrors.

Do you think it's better to have failed OSTree compositions block the ability to push out RPMs, given the possibility I mentioned above?

ausil:
if we have something like shell shock or heart bleed we absolutely must ship them all together.  I strongly believe we need to update things in lockstep.

dustymabe:
Note this is related to https://github.com/fedora-infra/bodhi/issues/1174

@ausil - would it be ok if we made the "lockstep" configurable. i.e. if we have an issue with the rpm-ostree compose like we have [had today](https://github.com/fedora-infra/bodhi/issues/1174) and we can't get it debugged/fixed in a reasonable time then could we set a variable and re-run it and try to fix it the next day?

bowlofeggs:
On Wed, 2017-01-25 at 15:20 -0800, Dusty Mabe wrote:
> would it be ok if we made the "lockstep" configurable

I like this suggestion - we could make it a flag on the bodhi-push CLI.
What do you think @ausil?

ausil:
A better bet is probably to allow updates-testing to fail but updates having to complete. If updates-testing fails we should then automatically file a bug in bugzilla and eject the update.

Dennis

On 2 February 2017 12:22:55 am GMT+01:00, Randy Barlow <notifications@github.com> wrote:
>On Wed, 2017-01-25 at 15:20 -0800, Dusty Mabe wrote:
>> would it be ok if we made the "lockstep" configurable
>
>I like this suggestion - we could make it a flag on the bodhi-push CLI.
>What do you think @ausil?
>
>-- 
>You are receiving this because you were mentioned.
>Reply to this email directly or view it on GitHub:
>https://github.com/fedora-infra/bodhi/pull/1214#issuecomment-276816060

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

bowlofeggs:
@ausil I like that idea, +1

dustymabe:
The only thing I don't like about that is us not realizing when it failed. If we do this I'd like to have some sort of notifications for it failing. There may be existing ones that I don't know about that we can use.

bowlofeggs:
@puiterwijk Do you want to keep this pull request open, or should we close it and file an issue to solve this in a way we can get consensus on?

bowlofeggs:
I'm going to close this pull request. It's already documented in #1174.

PR #1217, opened 2017-01-11
Disable autopush if an update receives negative karma from logged-in user

### LINKED ISSUE #1191: Disable autokarma on *any* signed-in -1 karma
Issue #879 disables autokarma if anyone gives a -1 to "basic system functionality". Let's extend that to all packages — if any logged-in user gives -1, disable automatic push. The maintainer will still be able to override and push manually, but will apply their judgement in this case, rather than letting the positive/negative sum decide.

trishnaguha:
Fixes #1191 
A point to be noted here: When autopush is disabled the update will be able to be pushed manually based of `karma threshold`.

Note that it will disable an update to go to obsolete state automatically when a testing update receives negative karma. It is going to wait for maintainer to decide. Hence we have this test failing for the same [test_obsolete_if_unstable_with_autopush_enabled_when_testing](https://github.com/fedora-infra/bodhi/blob/04cdcd5f4e7c3972761d8e6969e8cc9e79b50881/bodhi/tests/server/services/test_updates.py#L2165)

### DIFF
--- a/bodhi/server/models/models.py
+++ b/bodhi/server/models/models.py
@@ -1720,9 +1720,9 @@ class Update(Base):
         # Return if the status of the update is not in testing
         if self.status != UpdateStatus.testing:
             return
-        # If critical update receives negative karma disable autopush
-        if self.critpath and self.autokarma and self._composite_karma[1] != 0:
-            log.info("Disabling Auto Push since the critical update has negative karma")
+        # If an update receives negative karma disable autopush
+        if self.autokarma and self._composite_karma[1] != 0:
+            log.info("Disabling Auto Push since the update has received negative karma")
             self.autokarma = False
         elif self.stable_karma and self.karma >= self.stable_karma:
             if self.autokarma:
--- a/bodhi/tests/server/services/test_updates.py
+++ b/bodhi/tests/server/services/test_updates.py
@@ -2165,24 +2165,35 @@ class TestUpdatesService(bodhi.tests.server.functional.base.BaseWSGICase):
     def test_obsolete_if_unstable_with_autopush_enabled_when_testing(self, publish, *args):
         """
         Send update to obsolete state if it reaches unstable karma threshold on
-        testing state where request is stable when Autopush is enabled. Make sure that it
-        does not go to stable state.
+        testing state where request is stable when Autopush is enabled. Make sure that the
+        autopush is disabled and the update does not go to stable state.
         """
         nvr = u'bodhi-2.0.0-2.fc17'
         args = self.get_update(nvr)
         args['autokarma'] = True
-        args['stable_karma'] = 1
-        args['unstable_karma'] = -1
+        args['stable_karma'] = 2
+        args['unstable_karma'] = -2
 
         resp = self.app.post_json('/updates/', args)
         up = Update.get(nvr, self.db)
         up.status = UpdateStatus.testing
         up.request = UpdateRequest.stable
-        up.comment(self.db, u'Failed to work', author=u'ralph', karma=-1)
         self.db.flush()
 
+        up.comment(self.db, u'Failed to work', author=u'ralph', karma=-1)
         up = self.db.query(Update).filter_by(title=nvr).one()
-        self.assertEquals(up.karma, -1)
+
+        up.comment(self.db, u'WFM', author=u'puiterwijk', karma=1)
+        up = self.db.query(Update).filter_by(title=nvr).one()
+
+        up.comment(self.db, u'It has bug', author=u'bowlofeggs', karma=-1)
+        up = self.db.query(Update).filter_by(title=nvr).one()
+
+        up.comment(self.db, u'Still not working', author=u'bob', karma=-1)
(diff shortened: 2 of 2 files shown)


trishnaguha:
The patch has all possible test cases. Let me know if you have other possible cases in mind.

bodhi/server/models/models.py
```diff
@@ -1721,8 +1721,8 @@ def check_karma_thresholds(self, db, agent):
         if self.status != UpdateStatus.testing:
             return
         # If critical update receives negative karma disable autopush
-        if self.critpath and self.autokarma and self._composite_karma[1] != 0:
-            log.info("Disabling Auto Push since the critical update has negative karma")
+        if self.user and self.autokarma and self._composite_karma[1] != 0:
```
bowlofeggs: What is the purpose of testing ```self.user``` in this if statement?

bowlofeggs:
@trishnaguha As for the failing test and what to do there - I think we still want it to automatically obsolete the update if it reaches the negative karma threshold. So there should be a passing test that makes sure that updates get obsoleted when the negative threshold is reached. I assume that's what the failing test you referenced does?

trishnaguha:
@bowlofeggs This is why the test is failing https://github.com/fedora-infra/bodhi/blob/develop/bodhi/server/models/models.py#L1453-L1464 . According to the code an update should be obsoleted when it reaches unstable karma on pending status . I am going to fix the code in there and make it work for testing status as well.

trishnaguha:
@bowlofeggs Do you want me to remove the testcases for *critical* updates? Since this rule is going to be applied for all kind of updates. Those testcases seem duplicate now.

bowlofeggs:
@trishnaguha Do critical updates still have similar behaviour for the "Does the system still work?" question? (i.e., the non-karma question)

trishnaguha:
@bowlofeggs As per this rule, An update will lose the ability of autopush if it receives a negative karma. This is only going to happen for `autokarma` based updates. We have the same rule applied for `critical updates`: https://github.com/fedora-infra/bodhi/blob/develop/bodhi/tests/server/services/test_updates.py#L2877-L3053 .

For `non-karma` based updates (Based on time), critical updates need to spend 14 days in testing with no negative karma. For non-critical updates it is 7 days. But both need to have all positive karma during the period.

bowlofeggs:
@trishnaguha Ah, I thought we had code to disable autokarma if the critical path question was answered in the negative. I guess I was wrong.

I don't see harm in keeping the duplicate tests actually - they run through the code with slightly different data and might could catch a problem one day. The only downside I can think of is that the tests will take a little longer. Besides, we can always change our minds later. I say let's keep them.

bowlofeggs:
@trishnaguha Do all the tests pass now (i.e., is this ready for review?)

trishnaguha:
@bowlofeggs yup! It is now ready for review.

bodhi/tests/server/services/test_updates.py
```diff
@@ -2165,24 +2165,34 @@ def test_obsolete_if_unstable_karma_not_reached_with_autopush_enabled_when_pendi
     def test_obsolete_if_unstable_with_autopush_enabled_when_testing(self, publish, *args):
         """
         Send update to obsolete state if it reaches unstable karma threshold on
-        testing state where request is stable when Autopush is enabled. Make sure that it
-        does not go to stable state.
+        testing state where request is stable when Autopush is enabled. Make sure that the
+        autopush is disabled and the update does not go to stable state.
```
bowlofeggs, line 2169: This comment says that the test should make sure that autopush is disabled, but I don't see an assertion below that autokarma is False.

bodhi/tests/server/services/test_updates.py
```diff
@@ -3080,8 +3090,8 @@ def test_manually_push_critical_update_with_autopush_turned_off(self, publish, *
 
     @mock.patch(**mock_valid_requirements)
     @mock.patch('bodhi.server.notifications.publish')
-    def test_autopush_non_critical_update_with_negative_karma(self, publish, *args):
-        """Autopush Non Critical update even though it receives negative karma"""
+    def test_disable_autopush_non_critical_update_with_negative_karma(self, publish, *args):
+        """Disable Autopush Non Critical update even though it receives negative karma"""
```
bowlofeggs: The wording of the end of this sentence is a little confusing since it's from the way the comment used to be. How about "Disable autokarma on non-critical updates upon negative comment."

bowlofeggs (changes requested):
Just two comments. Let's add that assertion and reword the docblock, and this will be good to go!

bodhi/tests/server/services/test_updates.py
trishnaguha: Ah that docblock I added was confusing :).

trishnaguha:
Modified!

bowlofeggs (approved):
Looks great, @trishnaguha++

trishnaguha:
:taco: :)

PR #1223, opened 2017-01-14
PEP-8 bodhi.server.

bowlofeggs:


bodhi/server/metadata.py
```diff
@@ -11,23 +11,21 @@
...
 import createrepo_c as cr
 
+from bodhi.server.buildsys import get_session
 from bodhi.server.config import config
 from bodhi.server.models import Build, UpdateStatus, UpdateRequest, UpdateSuggestion
-from bodhi.server.buildsys import get_session
 
+
+__version__ = '2.0'
 log = logging.getLogger(__name__)
```
pypingou, line 29: pep8 doesn't complain about the case of global variables?

pypingou:
One question the rest looks fine, but since it conflicts I guess there might be some changes needed before merging.

bodhi/server/metadata.py
bowlofeggs: @pypingou ```flake8``` doesn't complain about this. I think pylint might, but I personally find that I disagree with pylint on many things (it goes beyond PEP-8). I only intend to enforce PEP-8 on Bodhi's code for now, with the exception that I allow a 100 character line length instead of 80 characters because c'mon, anyone on one of those old 80's monitors should upgrade to a 1080p screen, at minimum ☺ 100 chars is far more readable and encourages more meaningful variable names (like ```user``` instead of ```u```).

Additionally, capitalizing ```LOG``` would imply that it is a constant by PEP-8, and it is not a constant - it's a Logger instance ☺

bowlofeggs:
@pypingou I have rebased and resolved the conflict (it was just a conflict on the list of modules that have PEP-8 enforcement, due to another PEP-8 PR I recently merged.) Let me know if you think this is ship-shape (an idiom for production worthy ☺)

pypingou:
👍 for me

bodhi/server/metadata.py
pypingou: I thought capitalized variables were global variables, not necessarily constant, I might be wrong though (and it doesn't really matter for the PR at hand)

bowlofeggs: According to PEP-8, it's constants that should be capitalized:

https://www.python.org/dev/peps/pep-0008/#constants

PR #1224, opened 2017-01-15
Removed from models import* #fixed 1057

ankit01ojha:
Replaced python package bodhi/server/models with python module bodhi/server/models.py


ankit01ojha:
@bowlofeggs If its fine then I will add a commit regarding the .gitignore file

PR #1227, opened 2017-01-16
Changed the position of Add Comment button #fixes 1216

ankit01ojha:
Changed the position of Add Comment button from the bottom right to just right-side below the solution field.

### DIFF
--- a/bodhi/server/templates/update.html
+++ b/bodhi/server/templates/update.html
@@ -726,8 +726,20 @@ $(document).ready(function(){
               <div class="col-sm-10">
                 <input class="form-control" id="captcha_value" name="captcha_value"
                 placeholder="prove that you're a human" required="true"></input>
+                  <div class="container1" style="margin-bottom: 1em;">
+                </div>
               </div>
             </div>
+            <div class="hidden-xs hidden-sm">
+            <div class="row">
+              <div class="col-md-12">
+                <button type="submit" class="pull-right btn btn-default">
+                  <span class="indicator fa fa-comment" data-spinclass="indicator fa fa-spinner fa-spin"></span>
+                  Add Comment
+                </button>
+              </div>
+            </div>
+          </div>
             % endif
 
           </div><! -- end form-group -->
@@ -791,16 +803,17 @@ $(document).ready(function(){
 
         </div>
       </div>
+      <div class="hidden-md hidden-lg">
+         <div class="row">
+            <div class="col-md-12">
+               <button type="submit" class="pull-right btn btn-default">
+                  <span class="indicator fa fa-comment" data-spinclass="indicator fa fa-spinner fa-spin"></span>
+                  Add Comment
+                </button>
+              </div>
+            </div>
+          </div>
 
-
-      <div class="form-group">
-        <div class="col-sm-offset-1 col-sm-5">
-          <button type="submit" class="pull-right btn btn-default">
-            <span class="indicator fa fa-comment" data-spinclass="indicator fa fa-spinner fa-spin"></span>
-            Add Comment
-          </button>
-        </div>
-      </div>
     </form>
 
   </div>


ankit01ojha:
@trishnaguha can you review it :)

trishnaguha:
@ankit01ojha Can you attach screenshot of the page since it is related to UI?

ankit01ojha:
screenshot from 2017-01-18 19-05-19

ankit01ojha:
@trishnaguha Is this fine?

trishnaguha:
It would look better if you can align the button to right :) ? just like where the text-boxes end?

ankit01ojha:
screenshot from 2017-01-18 19-23-55

ankit01ojha:
@trishnaguha I think it looks good here :)

bowlofeggs:
@ankit01ojha It still looks a little too far to the right, and it's also bumped up against the Solution answer box. We probably want it to have a little space. @ryanlerch is the UX x-pert though. Ryan, what do you think?

Additionally, we should also make sure this looks nice on mobile. The button actually looks perfect on mobile as-is with no changes, IMO:

screenshot_20170118-090202

So let's make sure that any changes we make continue to look good in a mobile browser.

trishnaguha:
@bowlofeggs Since the **Add Comment** button is below the `Feedback Guidelines`, when the number `Test Case` grows the button goes more down.. appears right after the last testcase :).
For instance this one: https://bodhi.fedoraproject.org/updates/FEDORA-2016-788a7edfd7

ankit01ojha:
screenshot from 2017-01-19 21-16-14

ankit01ojha:
@trishnaguha and @bowlofeggs I hope this is as required :)

trishnaguha:
LGTM :+1:

ankit01ojha:
@trishnaguha can it be merged?

ankit01ojha:
@bowlofeggs Is this fine?

bowlofeggs:
This does look nice on my desktop now, I tested that same update the @trishnaguha linked with a development checkout of this PR: http://localhost:6543/updates/FEDORA-2016-788a7edfd7

However, when I view this PR on my phone, the button is presented before the karma choices:

screenshot_20170123-105841

I think we'd ideally want to find a way to keep the button below the karma prompts on mobile, while having it look similar to how it looks on desktop in your pull request.

I'm not an HTML guru. Ideas?

ankit01ojha:
@bowlofeggs This might work, I have changed the visibility of add comment button. When it will be in mobile size add comment will appear as you want  and as when its in large size(desktop or more) it will come as I have pasted in my screenshot.

bodhi/server/templates/update.html
```diff
@@ -791,16 +803,17 @@ <h3 class="nomargin">Notes about this update:</h3>
...
+      <div class="hidden-md hidden-lg">
+         <div class="row">
+            <div class="col-md-12">
+               <button type="submit" class="pull-right btn btn-default">
+                  <span class="indicator fa fa-comment" data-spinclass="indicator fa fa-spinner fa-spin"></span>
+                  Add Comment
+                </button>
+              </div>
+            </div>
+          </div>
```
bowlofeggs, line 815: I don't think we want to duplicate code like this. Let's see if we can find a solution that doesn't repeat this block.

bowlofeggs (changes requested):

bodhi/server/templates/update.html
ankit01ojha: I will be searching for it also, but can it be merged untill its new solution is found :)

bowlofeggs:
On Thu, 2017-02-02 at 09:43 -0800, Ankit Raj Ojha wrote:
> I will be searching for it also, but can it be merged untill its new
> solution is found :)

I would prefer not to make a change until we have the right change. As
it is, the code is relatively simple and the button mostly works. This
proposed change will raise the complexity of the code and will violate
the DRY principle:

https://en.wikipedia.org/wiki/Don%27t_repeat_yourself

In other words, the issue isn't severe enough to warrant workaround-
type changes to our code.

bowlofeggs:
I believe this issue is addressed by #1313, so I am closing this pull request.

PR #1228, opened 2017-01-18
Fix dependencies

bowlofeggs:
This pull request has two commits (the second depends on the first):

1. Remove two development dependencies from the setup.py, since they aren't needed in production.
2. Make the Vagrant Ansible dev role install all dependencies by name from the Fedora 24 repository.


jeremycline (approved):
Looks good! :shipit: 

We were living on the edge here, mixing ``sudo pip install`` with system packages.

PR #1230, opened 2017-01-18
Lock the buildsystem while logging in

jeremycline:
This is a rebase of https://github.com/fedora-infra/bodhi/pull/1142 with tests added.

### DIFF
--- a/bodhi/tests/server/test_buildsys.py
+++ b/bodhi/tests/server/test_buildsys.py
@@ -15,6 +15,7 @@
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
 """This test suite contains tests for the bodhi.server.buildsys module."""
 
+from threading import Lock
 import unittest
 
 import koji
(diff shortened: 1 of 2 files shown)


bodhi/tests/server/test_buildsys.py
```diff
@@ -140,3 +141,53 @@ def test_login_success(self, error):
...
+        old_buildsystem = buildsys._buildsystem
+        buildsys.setup_buildsystem({})
+        self.assertTrue(old_buildsystem is buildsys._buildsystem)
+
+    @mock.patch('bodhi.server.buildsys._buildsystem', None)
+    def test_koji_buildsystem(self):
+        """Assert the buildsystem initializes correctly for koji"""
+        self.assertTrue(buildsys._buildsystem is None)
+        buildsys.setup_buildsystem({'buildsystem': 'koji'})
+        self.assertFalse(buildsys._buildsystem is None)
```
bowlofeggs, line 181: We could probably add some stronger assertions here by calling the buildystem (since it's that weird internal function) and making assertions on the return value. Perhaps asserting that the type of the buildsystem is a KojiSomethingOrOther, and that it got the config settings? You could just set up a mock on ```koji_login()``` and assert that it gets called when you call the buildsystem and assert that it gets the right settings.

bowlofeggs (changes requested):
Just one request, looks good otherwise!

bodhi/tests/server/test_buildsys.py
jeremycline: Make it so
I have made it so, Number One

bowlofeggs (approved):
Looks great!

PR #1231, opened 2017-01-19
Specify clone directory for comps file

puiterwijk:
Otherwise if the upstream repo changes names it gets cloned into another
directory.

### DIFF
--- a/bodhi/tests/server/consumers/test_masher.py
+++ b/bodhi/tests/server/consumers/test_masher.py
@@ -1183,3 +1183,48 @@ class TestMasherThread_eject_from_mash(MasherThreadBaseTestCase):
         self.assertEqual(t.updates, set([]))
         # The update's title should also have been removed from t.state['updates']
         self.assertEqual(t.state['updates'], [])
+
+
+class TestMasherThread_update_comps(unittest.TestCase):
+    """This test class contains tests for the MasherThread.update_comps() method."""
+
+    def setUp(self):
+        self.masher_thread = MasherThread(u'F17', u'stable', [u'bodhi-2.0-1.fc17'],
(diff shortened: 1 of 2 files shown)


jeremycline:
@bowlofeggs I added tests for this change. I also adjusted the method a bit to make a test pass, but I can adjust things if you don't want that change as part of this.

bodhi/tests/server/consumers/test_masher.py
```diff
@@ -1183,3 +1183,32 @@ def test_testing_request(self, publish):
...
+                                                         'comps_url': 'https://example.com/'})
+    @mock.patch('bodhi.server.consumers.masher.util.cmd')
+    def test_comps_no_dir(self, mock_cmd):
+        calls = [
+            mock.call(['git', 'clone', 'https://example.com/', '/some/path'], '/some'),
+            mock.call(['git', 'pull'], '/some/path'),
+            mock.call(['make'], '/some/path'),
+        ]
+        self.masher_thread.update_comps()
+        self.assertEqual(calls, mock_cmd.call_args_list)
```
bowlofeggs, line 1207: We should add one more test like this one that tests for when ```os.path.exists``` return True. Also, it'd be good to assert that ```os.path.exists()``` was called with the correct path here and in the new test.

bowlofeggs (changes requested):
Let's add an assertion on the arguments to ```os.path.exists``` and also add a test to assert correct behavior when the path does exist.

bodhi/tests/server/consumers/test_masher.py
jeremycline: Done

bowlofeggs (approved):
Yes, we are inching towards a release!

PR #1137, opened 2016-11-30
Reload builds and releases after we commit the transaction

puiterwijk:
Signed-off-by: Patrick Uiterwijk <puiterwijk@redhat.com>

Issue #219, opened 2015-08-20
Can't submit update to multiple distros at once


AdamWill:
I often build a new NEVR for multiple distros and then submit updates for all of them together. The old web interface happily accepted this: I could submit, say, 'fedfind-1.3-1.fc23' and 'fedfind-1.3-1.fc22' together and it'd create two updates, one for F22, one for F23. This saved me a lot of busywork.

The new web UI doesn't seem to allow this; it pops up a message for a few seconds saying 'multiple fedfind builds' or something like that. The failure cases in the new UI could stand to be more obvious.

NiceandGently:
Happened to me with 2 packages.

lmacken:
This is currently a known limitation of Bodhi2, and a high priority feature.

kkofler:
IMHO, ideally, they should be treated as ONE update group and in particularly share the karma value (so that one needs only +1 or +2 for the identical update on all releases together rather than for every single release).

AdamWill:
That's a significant policy change, not an engineering issue.

ralphbean:
I never knew Bodhi1 could do this.  Sounds super useful.

AdamWill:
:+1: @ralphbean is the best.

PR #1232, opened 2017-01-19
Reload builds and releases after we commit the transaction

jeremycline:
This builds upon  https://github.com/fedora-infra/bodhi/issues/1137 and fixes a regression introduced by it which is currently in production. Specifically, it broke the feature that allowed you to submit multiple updates at a time (https://github.com/fedora-infra/bodhi/issues/219).


bowlofeggs (approved):
LGTM!

PR #1234, opened 2017-01-22
PEP-8 bodhi.tests.server.functional.

bowlofeggs:

### DIFF
--- a/bodhi/tests/server/base.py
+++ b/bodhi/tests/server/base.py
@@ -16,12 +16,14 @@ This module contains a useful base test class that helps with common testing nee
 bodhi.server modules.
 """
 from contextlib import contextmanager
+import os
 import unittest
 
 from sqlalchemy import create_engine
 from sqlalchemy import event
 from sqlalchemy.orm import scoped_session, sessionmaker
 from zope.sqlalchemy import ZopeTransactionExtension
+import requests
 import transaction
 
 from bodhi.server import log, models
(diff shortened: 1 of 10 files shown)


bodhi/tests/server/base.py
```diff
@@ -32,6 +34,21 @@
 DB_NAME = None
 
 
+FAITOUT = 'http://209.132.184.152/faitout/'
+# The BUILD_ID environment variable is set by Jenkins and allows us to detect if
+# we are running the tests in jenkins or not
+# https://wiki.jenkins-ci.org/display/JENKINS/Building+a+software+project#Buildingasoftwareproject-below
+if os.environ.get('BUILD_ID'):
+    try:
+        req = requests.get('%s/new' % FAITOUT)
```
jeremycline, line 43: You might want to add a timeout kwarg here since the default is to never time out and that might cause the tests to hang for an indefinite amount of time. Not a blocker since it was already like that, but just so you know.

jeremycline (approved):
I approve, :shipit:

bodhi/tests/server/base.py
bowlofeggs: @jeremycline Great suggestion! Since this PR/commit is claiming to just PEP-8 this code, I'd prefer to make that change in a follow up commit so everything stays clean and atomic. I'll drop a link to that PR back here once I have it.

bowlofeggs: Here's a PR to add that timeout: https://github.com/fedora-infra/bodhi/pull/1239

PR #1235, opened 2017-01-22
Vagrant improvements

bowlofeggs:
This pull request contains three commits that fix and improve the development environment. The first pip installs pyramid_debugtoolbar, which was recently removed from ```setup.py```. The second adds a systemd unit to run the Bodhi development server. The third adds a set of helpful bash commands to perform common development operations, and documents the commands.


jeremycline (approved):
Excellent, looks good :shipit:

PR #1236, opened 2017-01-22
Do not traceback when CLI users enters the wrong password fixes #1126

### LINKED ISSUE #1126: Authentication errors print a traceback in the bodhi CLI

ankit01ojha:
Instead of traceback returns a good message when user types wrong password in CLI.

### DIFF
--- a/bodhi/client/bindings.py
+++ b/bodhi/client/bindings.py
@@ -35,8 +35,9 @@ import re
 import textwrap
 
 import six
+import click
 
-from fedora.client import OpenIdBaseClient, FedoraClientError
+from fedora.client import OpenIdBaseClient, FedoraClientError, AuthError
 import fedora.client.openidproxyclient
 
 
@@ -297,7 +298,12 @@ class BodhiClient(OpenIdBaseClient):
     def csrf(self):
         if not self.csrf_token:
             if not self.has_cookies():
-                self.login(self.username, self.password)
+                try:
+                    self.login(self.username, self.password)
+                except AuthError:
+                    click.echo("wrong password, please try again",
+                               err=True)
+                    sys.exit(1)
             self.csrf_token = self.send_request(
                 'csrf', verb='GET', auth=True)['csrf_token']
         return self.csrf_token


ankit01ojha:
@bowlofeggs can you review this one :)

bodhi/client/__init__.py
```diff
@@ -79,6 +79,8 @@ def new(user, password, **kwargs):
             print_resp(resp, client)
         except bindings.BodhiClientException as e:
             click.echo(str(e))
+	except Exception as e:
```
trishnaguha: The indentation seems wrong here.

trishnaguha:
Please add `.project` in .gitignore file.

ankit01ojha:
In my local file the indentations are correct, but I don't know why its showing here.

ankit01ojha:
screenshot from 2017-01-23 20-15-57

ankit01ojha:
@trishnaguha  Even in github editor its showing proper indentation, but don't why in my commit its showing wrong indentation.

bowlofeggs:
@ankit01ojha it places a bit of extra burden on us when you open new pull requests rather than working in existing ones, as we lose the comment history that forms the context of the current state of the effort.

Here's a little cheat sheet of how I usually work with pull requests. Say I have a branch called ```foo``` in a ```bowlofeggs``` git remote, and I've submitted a pull request for ```foo```. If @trishnaguha wants me to change something, all I do is make the change in my local branch, ```git add``` those changes, then ```git commit --amend``` (which rewrites history, essentially deleting my last commit and replacing it with a new one). Once I've done this, I will ```git push -f bowlofeggs foo```. That -f tells GitHub to be ok with me force pushing a new history to the remote branch, and now my pull request will show my new commit instead of my old one, while maintaining the history of comments.

.project
```diff
@@ -0,0 +1,11 @@
...
+<projectDescription>
+	<name>bodhi</name>
+	<comment></comment>
+	<projects>
+	</projects>
+	<buildSpec>
+	</buildSpec>
+	<natures>
+	</natures>
+</projectDescription>
```
bowlofeggs: Let's not commit this file.

bodhi/client/__init__.py
bowlofeggs: It seems that this may be a tab character. We only want to use spaces in Bodhi (I should write that in the developer guide). Please set your editor to use a tabstop of 4 spaces (so that when you hit the tab button when working on Bodhi, it'll insert spaces until the next multiple of 4 is reached).

bodhi/client/__init__.py
```diff
@@ -79,6 +79,8 @@ def new(user, password, **kwargs):
...
+	    click.echo("wrong password, please try again")
```
bowlofeggs: As we mentioned in your older PR, we don't want to assume that any Exception here means that the password was wrong. You will need to catch whatever more specific Exception subclass is raised when the password is wrong.

bowlofeggs (changes requested):
In addition to my other comments, this pull request also needs tests.

bowlofeggs:
I documented the use of spaces over tabs here: https://github.com/fedora-infra/bodhi/pull/1240

ankit01ojha:
@bowlofeggs I am not able to get which Exception subclass is raised when the password gets wrong, could you just help me a little.

bodhi/client/__init__.py
bowlofeggs: If you put a ```print type(e)``` in this block and then type a wrong password, you will see what the specific Exception type is that gets raised on authentication error. That's the Exception you probably want to catch here instead of catching all Exceptions.

ankit01ojha:
@bowlofeggs Modified!

bodhi/client/__init__.py
```diff
@@ -79,6 +79,8 @@ def new(user, password, **kwargs):
             print_resp(resp, client)
         except bindings.BodhiClientException as e:
             click.echo(str(e))
+        except Exception as AuthError:
```
cverna: Hi, I think this will work, but only for the `bodhi updates new` command. The exception should be managed when the client tries to login. 

Also I believe AuthError is undefined here it need to be imported.

ankit01ojha: @cverna can you tell me how login is done through bodhi CLI so that I can check my changes.

trishnaguha:
@ankit01ojha the `.project` file is still there in the PR. We do not need to commit this file.

ankit01ojha:
I removed `.project` from `.gitignore` file.

bodhi/client/__init__.py
cverna: @ankit01ojha You should take a look [here](https://github.com/fedora-infra/bodhi/blob/develop/bodhi/client/bindings.py#L307)

Did you test your change ? Did the program print "wrong password, please try again" ?

ankit01ojha: @cverna Yes, the changes were working, when I tried to do this 
`$ bodhi updates new --user bowlofeggs ejabberd-16.09-5.fc25`
and gave wrong password then it printed "wrong password,please try again"

cverna: Ok :) did you try another command ? I am just curious 😄

cverna:
@ankit01ojha I think you need to remove `.project` from the commit, not only the .gitignore 🍩

bodhi/client/__init__.py
ankit01ojha: @cverna , nope its not working for the overrides, comments and request. Almost found the solutions for them :)

ankit01ojha:
@cverna IMO, it would be fine to print the traceback in case of request, overrides and comment along with the wrong password message.
@trishnaguha and @bowlofeggs  what you both think?

bodhi/client/__init__.py
```diff
@@ -79,6 +80,8 @@ def new(user, password, **kwargs):
...
+	        click.echo("wrong password, please try again")
```
trishnaguha: The indentation needs to be fixed here.

bodhi/client/__init__.py
cverna: @ankit01ojha if you want you can look at my branch, I think this would handle all command. Maybe you can test it.
https://github.com/cverna/bodhi/blob/handle_wrong_password/bodhi/client/bindings.py#L309

bowlofeggs: This is catching any type of Exception and calling it AuthError (a context scoped variable name, not a type), which is not what we want to do here. With this code, any Exception raised will tell the user that their password is wrong. For example, an ```IOError``` would print "wrong password, please try again". We need to catch only the Exception that is raised when the password is wrong and then print such a message.

bowlofeggs (changes requested):
Rather than repeating the Exception handler so many times, I would recommend finding a way to avoid repetition and express the password handling in one place. For example, you could create a BodhiCLIClient as a wrapper or subclass around the BodhiClient that does this for you. Then everything in here would benefit from that automatically. Another way to do it could be to look into the click library to see if it has a way to mass-handle certain types of Exceptions.

The most important thing to fix is to make sure we are being specific about which Exception type we are handling.

ankit01ojha:
@bowlofeggs I think solution suggested by @cverna is also nice, it works properly. What you think about it? https://github.com/cverna/bodhi/blob/handle_wrong_password/bodhi/client/bindings.py#L309

bowlofeggs:
@ankit01ojha @cverna The ```bodhi.client.bindings``` module is meant for programmatic access to the REST API for Python programs. Thus, we shouldn't assume that a human is the actor and should avoid using the click library there and should not call ```sys.exit```.

A possible alternative is to subclass the bindings in ```bodhi.client``` and adjust the subclass to have the desired behavior.

ankit01ojha:
@bowlofeggs Can you please tell me how can I subclass the bindings in `bodhi.client` (Subclass a file to a directory). Should I make a new fle in `bodhi.client` ?

bowlofeggs:
@ankit01ojha We wouldn't be subclassing a file. We need to subclass the bindings, which is a class. I recommend reading up on object oriented design - it's a bit much to explain in GitHub comments. Here are a couple links that will get you started:

http://www.jesshamrick.com/2011/05/18/an-introduction-to-classes-and-inheritance-in-python/
https://docs.python.org/2/tutorial/classes.html

In general I would recommend reading the entire Python tutorial. It's good stuff!

ankit01ojha:
WIP, just pushed to remove the merge conflicts :)

bowlofeggs:

PR #1238, opened 2017-01-23
Changed package search to case-insensitive

lalitkr:
@bowlofeggs Please check and review this.

### DIFF
--- a/bodhi/server/services/updates.py
+++ b/bodhi/server/services/updates.py
@@ -217,7 +217,7 @@ def query_updates(request):
     packages = data.get('packages')
     if packages is not None:
         query = query.join(Update.builds).join(Build.package)
-        query = query.filter(or_(*[Package.name.ilike('%pkg%') pkg for pkg in packages]))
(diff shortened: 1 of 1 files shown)


bowlofeggs (changes requested):
Hello @lalitkr! This seems like the correct approach to me, but we will need at least two changes:

1. Squash your commits so there is only one.
2. Add a "fixes #997" into your commit message, so the ticket and the commit get linked (and the issue will also autoclosed when this gets merged).
3. Write a couple of tests that demonstrate that the search works with varying cases in the search. For example, maybe one test that searches for "bodhi" and another test searches for "BoDHi".

Thanks for the contribution!

lalitkr:
Hi @bowlofeggs,
Thank you for your feedback. Will work on the same and update it once done.

bowlofeggs:
Hello @lalitkr! Do you still intend to work on this pull request?

bowlofeggs:
I'm going to close this for now, but feel free to reopen if you still want to work on it!

PR #1239, opened 2017-01-23
Only wait for 60 seconds for the faitout database.

bowlofeggs:
Thanks to @jeremycline for the suggestion to do this.

### DIFF
--- a/bodhi/tests/server/base.py
+++ b/bodhi/tests/server/base.py
@@ -40,7 +40,7 @@ FAITOUT = 'http://209.132.184.152/faitout/'
 # https://wiki.jenkins-ci.org/display/JENKINS/Building+a+software+project#Buildingasoftwareproject-below
 if os.environ.get('BUILD_ID'):
     try:
-        req = requests.get('%s/new' % FAITOUT)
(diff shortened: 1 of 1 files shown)


bodhi/tests/server/base.py
```diff
@@ -40,7 +40,7 @@
 # https://wiki.jenkins-ci.org/display/JENKINS/Building+a+software+project#Buildingasoftwareproject-below
 if os.environ.get('BUILD_ID'):
     try:
-        req = requests.get('%s/new' % FAITOUT)
+        req = requests.get('%s/new' % FAITOUT, timeout=60)
```
bowlofeggs, line 43: I presumed that the unit here is seconds, but the requests docs do not say:

http://docs.python-requests.org/en/master/api/#requests.request

Also, 60 seconds is an arbitrary value I pulled out of nowhere.

bowlofeggs: Ah, it does seem to be seconds:

http://docs.python-requests.org/en/master/user/advanced/#timeouts

jeremycline (approved):
Perfect :shipit:

pypingou:
Definitely going to re-use that in some places :D

PR #1240, opened 2017-01-23
Mention that we use spaces instead of tabs.

bowlofeggs:


jeremycline (approved):
This is fine with me, although PEP8 is very explicit that 4 spaces is the expected indentation except for legacy code that already contains tabs (in which case it should be all tabs or converted to all spaces).

bowlofeggs:
@jeremycline You make a good point. I think I'll favor keeping the docs shorter since this is redundant info, and I'll close this PR. Thanks!

PR #1248, opened 2017-01-25
Expire sessions in the default config.

bowlofeggs:


production.ini
```diff
@@ -411,6 +411,10 @@ session.data_dir = %(here)s/data/sessions/data
 session.lock_dir = %(here)s/data/sessions/lock
 session.key = mykey
 session.cookie_on_exception = true
+# Tell the browser to only send the cookie over TLS
+session.secure = true
```
pypingou, line 415: You may want to specify that this should be False for local development and True in prod (behind TLS)

pypingou (approved):
One small comment otherwise 👍 for me

production.ini
bowlofeggs: @pypingou We actually currently have a separate development.ini.example file that does not have this change in it, so it's sort of implicit that this change would only be for production.

PR #1250, opened 2017-01-26
changed the link to build in koji from http to https fixes #1246

### LINKED ISSUE #1246: link to build in koji should use https
in each update, there's a link to the build in koji, for example, https://bodhi.fedoraproject.org/updates/FEDORA-2017-6d6e2bfd1a links to http://koji.fedoraproject.org/koji/search?terms=libXpm-3.5.12-1.fc24&type=build&match=glob (next to "Build") ... the link should use https so that the connection stays secure throughout. koji supports https and even redirects from http to https, so it would be better to change the link to point to https.

ankit01ojha:


ankit01ojha:
Do I have to any other changes or this is fine ? I think you meant only this.

pypingou:
Looks good on the change side, but on the commit message, shouldn't the ``fixes #1246`` be in body rather than its title?

ankit01ojha:
@pypingou Modified.

pypingou:
Yup, that looks better, 👍 for me, now let's see what @bowlofeggs thinks :)

bowlofeggs:
This looks reasonable to me. Thanks!

PR #1275, opened 2017-02-06
Adjust how ``file`` is mocked in a few tests

jeremycline:
This appears to be a problem with the way mock handles patching builtins
in older version versus newer versions. Although setting ``create=True``
isn't great, I felt like it was the best option without digging very
deeply into the guts of mock and figuring out a way that works in the
old version and the new version.


bowlofeggs (approved):
Works for me, thanks!

PR #1277, opened 2017-02-07
Reset karma note after removal or addition of build(s)

### LINKED ISSUE #1276: mention that karma is reset on update edit
Some people are confused why karma is zero for certain updates, because they don't know that update edit resets karma. It might be a good idea to add a note when this happens, so that it's more obvious.

What about changing e.g. this:
```
jwrdegoede edited this update.

New build(s):
    mesa-13.0.4-1.fc25
    libglvnd-0.2.999-10.gitdc16f8c.fc25
Removed build(s):
    libglvnd-0.2.999-9.gitdc16f8c.fc25
    mesa-13.0.3-7.fc25
```

into this:
```
jwrdegoede edited this update.

New build(s):
    mesa-13.0.4-1.fc25
    libglvnd-0.2.999-10.gitdc16f8c.fc25
Removed build(s):
    libglvnd-0.2.999-9.gitdc16f8c.fc25
    mesa-13.0.3-7.fc25

All karma has been reset.
```
?

Of course don't add the note when just the update is edited, but builds are not changed (e.g. just the description is updated, or linked bug reports, etc) - that does not reset karma.

trishnaguha:
Fixes #1276 .

### DIFF
--- a/bodhi/server/models.py
+++ b/bodhi/server/models.py
@@ -962,6 +962,8 @@ class Update(Base):
             comment += '\n\nRemoved build(s):\n'
             for removed_build in removed_builds:
                 comment += "\n- %s" % removed_build
+        if new_builds or removed_builds:
+            comment += '\n\nKarma has been reset.'
         up.comment(db, comment, karma=0, author=u'bodhi')
         caveats.append({'name': 'builds', 'description': comment})
 
--- a/bodhi/tests/server/services/test_updates.py
+++ b/bodhi/tests/server/services/test_updates.py
@@ -1639,6 +1639,8 @@ class TestUpdatesService(bodhi.tests.server.functional.base.BaseWSGICase):
         Removed build(s):
 
         - bodhi-2.0.0-2.fc17
+
+        Karma has been reset.
         """).strip()
         self.assertMultiLineEqual(up['comments'][-1]['text'], comment)
         self.assertEquals(len(up['builds']), 1)
@@ -1680,6 +1682,8 @@ class TestUpdatesService(bodhi.tests.server.functional.base.BaseWSGICase):
         Removed build(s):
 
         - bodhi-2.0.0-2.fc17
+
+        Karma has been reset.
         """).strip()
         self.assertMultiLineEqual(up['comments'][-2]['text'], comment)
         self.assertEquals(up['comments'][-3]['text'],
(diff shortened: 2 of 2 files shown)


bodhi/server/models/models.py
```diff
@@ -854,6 +854,7 @@ def edit(cls, request, data):
             comment += '\n\nRemoved build(s):\n'
             for removed_build in removed_builds:
                 comment += "\n- %s" % removed_build
+            comment += '\n\n\nAll karma has been reset.'
```
bowlofeggs: Doesn't karma get reset on new builds too? Also, I'd recommend just saying "Karma has been reset." instead of using the word all, just to be less verbose.

trishnaguha: Agreed!

bodhi/server/models/models.py
```diff
@@ -854,6 +854,7 @@ def edit(cls, request, data):
             comment += '\n\nRemoved build(s):\n'
             for removed_build in removed_builds:
                 comment += "\n- %s" % removed_build
+            comment += '\n\n\nKarma has been reset.'
```
bowlofeggs: I think we still want to tell the user that the karma has been reset on new_builds too, for consistency.

bowlofeggs (changes requested):

bodhi/server/models/models.py
trishnaguha: That can be done..But anyhow the text is going to appear at the end(After couple of lines) which means it applies both for `new_builds` and `removed_builds`. But for consistency and clarification we can have it in both the places.

bodhi/server/models/models.py
```diff
@@ -850,10 +850,12 @@ def edit(cls, request, data):
             comment += '\n\nNew build(s):\n'
             for new_build in new_builds:
                 comment += "\n- %s" % new_build
+            comment += '\n\nKarma has been reset.'
         if removed_builds:
-            comment += '\n\nRemoved build(s):\n'
+            comment += '\n\n\nRemoved build(s):\n'
             for removed_build in removed_builds:
                 comment += "\n- %s" % removed_build
+            comment += '\n\nKarma has been reset.'
```
bowlofeggs: One more change I'd suggest - we probably don't want to say that karma was reset twice in any given operation. I suggest moving these ```comment +=``` lines into a new if statement that's outside of both of the above ifs, so that the comment only gets added once. Something similar to this would probably work:

```
if new_builds or removed_build:
    comment += '\n\nKarma has been reset.'
```

This way you don't need to edit the ```Removed build(s)``` comment either, I don't think.

bowlofeggs (changes requested):

bowlofeggs:
@trishnaguha Cool the code change looks good. Two more things:

0) Let's adjust the commit message to say "removal or addition" rather than just "removal".
1) This PR now conflicts due to the movement of ```bodhi.server.models.models``` to ```bodhi.server.models```. Can you rebase to resolve the conflict?

trishnaguha:
@bowlofeggs Updated :-).

PR #1280, opened 2017-02-08
PEP-8 the bodhi/ folder.

bowlofeggs:

### DIFF
--- a/bodhi/tests/server/__init__.py
+++ b/bodhi/tests/server/__init__.py
@@ -4,23 +4,8 @@ import mock
 import sqlalchemy
 
 from bodhi.server.models import (
-    Base,
-    Bug,
-    Build,
-    BuildrootOverride,
-    Comment,
-    CVE,
-    Group,
-    Package,
-    Release,
-    ReleaseState,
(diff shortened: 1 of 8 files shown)


bodhi/tests/server/__init__.py
```diff
@@ -4,23 +4,8 @@
...
-    ReleaseState,
-    Update,
-    UpdateType,
-    User,
-    UpdateStatus,
-    UpdateRequest,
-    TestCase,
-)
+    Bug, Build, BuildrootOverride, Comment,
+    CVE, Group, Package, Release, ReleaseState, Update, UpdateRequest, UpdateType, User, TestCase)
```
pypingou: Any reason for this split? Why not putting CVE, Group and Package in the first line?

pypingou (approved):
One small nitpick that can be ignored so 👍 for me :)

bodhi/tests/server/__init__.py
bowlofeggs: No reason, but I'll fix it!

PR #1281, opened 2017-02-08
Change systemd module to service module in Dev environment setup

trishnaguha:
`systemd` module is introduced in to Ansible 2.2. But `service` module works for every version. 
Changing to `service` module would be more general.


bowlofeggs (approved):
Thanks!

PR #1282, opened 2017-02-10
Remove a useless --request flag from a CLI command.

### LINKED ISSUE #1187: Duplicate parameters for setting the state of request
[Lukas Slebodnik filed this issue in Bugzilla](https://bugzilla.redhat.com/show_bug.cgi?id=1408198):

> Description of problem:
> Acording to the help the state of package can be changed either by parameter --request and also as a last parameter.
>
> I didn't test but whether parameter --request is accepted or has higher priority then the last argument. But such situation is confusing.
> You should decide which way can be used. But there should be just one way.  
> 
> Version-Release number of selected component (if applicable):
> sh$ rpm -q bodhi-client 
> bodhi-client-2.3.3-1.fc26.noarch
> 
> How reproducible:
> Deterministic
> 
> Additional info:
> 1. bodhi updates request --help
> Usage: bodhi updates request [OPTIONS] UPDATE STATE
> 
> Options:
>   --user TEXT
>   --password TEXT
>   --request [testing|stable|unpush|obsolete|revoke]
>                                   Requested repository
>   --staging                       Use the staging bodhi instance
>   --help                          Show this message and exit.
> 
> man bodhi says:
>        bodhi updates request [options] <update> <state>
>           Request  that the given update be changed to the given state. update
>           should be given by update id, and state should be  one  of  testing,
>           stable, unpush, obsolete, or revoke.
>
>
> BTW, if you decide to use state from last parameter then IMHO it would be better to change or der of STATE and UPDATE. It will be simpler to implement bash completion for such order.

bowlofeggs:
fixes #1187

### DIFF
--- a/bodhi/client/__init__.py
+++ b/bodhi/client/__init__.py
@@ -124,8 +124,6 @@ def query(**kwargs):
 @click.argument('state')
 @click.option('--user', envvar='USERNAME')
 @click.option('--password', prompt=True, hide_input=True)
-@click.option('--request', help='Requested repository',
-              type=click.Choice(['testing', 'stable', 'unpush', 'obsolete', 'revoke']))
 @click.option('--staging', help='Use the staging bodhi instance',
               is_flag=True, default=False)
 def request(update, state, user, password, **kwargs):


trishnaguha:
Looks good :cake: .

PR #1283, opened 2017-02-11
PEP-8 half of the alembic migrations.

bowlofeggs:


pypingou:
Looks fine to me

PR #1285, opened 2017-02-14
Docs bindings

bowlofeggs:
This pull request has two commits. The first reorganizes the documentation a bit, and the second adds autodocs for the REST API Python bindings.


pypingou:
Cool 👍

Issue #1291, opened 2017-02-16
Bodhi should comment when disabling autokarma due to -1 votes


bowlofeggs:
@kparal made [this suggestion](https://github.com/fedora-infra/bodhi/issues/1191#issuecomment-280369011):

Thanks for implementing this! As a suggestion, you might add some message to clearly explain what just happened, e.g.:
> Negative karma received, disabling automatic push to stable. Please push manually if you decide the issue is not severe.

trishnaguha:
+1.

PR #1292, opened 2017-02-17
Add comment when disabling autopush on negative karma

trishnaguha:
Fixes https://github.com/fedora-infra/bodhi/issues/1291

### DIFF
--- a/bodhi/server/models.py
+++ b/bodhi/server/models.py
@@ -1832,6 +1832,11 @@ class Update(Base):
         if self.autokarma and self._composite_karma[1] != 0:
             log.info("Disabling Auto Push since the update has received negative karma")
             self.autokarma = False
+            default_comment = (
+                u'Bodhi is disabling automatic push to stable due to negative karma. '
+                u'The maintainer may manually if they determine that the issue is not severe.')
+            text = unicode(config.get('disable_automatic_push_to_stable', default_comment))
+            self.comment(db, text, author=u'bodhi')
         elif self.stable_karma and self.karma >= self.stable_karma:
             if self.autokarma:
                 log.info("Automatically marking %s as stable" % self.title)
--- a/bodhi/tests/server/services/test_updates.py
+++ b/bodhi/tests/server/services/test_updates.py
@@ -3088,6 +3088,12 @@ class TestUpdatesService(bodhi.tests.server.functional.base.BaseWSGICase):
         up.comment(self.db, u'Failed to work', author=u'ralph', karma=-1)
         up = self.db.query(Update).filter_by(title=resp.json['title']).one()
 
+        default_comment = (
+            u'Bodhi is disabling automatic push to stable due to negative karma. '
+            u'The maintainer may manually if they determine that the issue is not severe.')
+        expected_comment = unicode(config.get('disable_automatic_push_to_stable', default_comment))
+        self.assertEquals(up.comments[2].text, expected_comment)
+
         up.comment(self.db, u'LGTM Now', author=u'ralph', karma=1)
         up = self.db.query(Update).filter_by(title=resp.json['title']).one()
 
--- a/development.ini.example
+++ b/development.ini.example
@@ -26,6 +26,8 @@ stablekarma_comment = This update has reached the stable karma threshold and wil
 testing_approval_msg_based_on_karma = This update has reached the stable karma threshold and can be pushed to stable now if the maintainer wishes.
 not_yet_tested_msg_based_on_karma = This update has not reached stable karma threshold.
 
+disable_automatic_push_to_stable = Bodhi is disabling automatic push to stable due to negative karma. The maintainer may manually if they determine that the issue is not severe.
+
 # Libravatar - If this is true libravatar will work as normal. Otherwise, all
 # libravatar links will be replaced with the string "libravatar.org" so that
 # the tests can still pass.
--- a/production.ini
+++ b/production.ini
@@ -19,6 +19,8 @@ stablekarma_comment = This update has reached the stable karma threshold and wil
 testing_approval_msg_based_on_karma = This update has reached the stable karma threshold and can be pushed to stable now if the maintainer wishes.
 not_yet_tested_msg_based_on_karma = This update has not reached the stable karma threshold.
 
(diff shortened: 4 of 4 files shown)


bodhi/server/scripts/approve_testing.py
```diff
@@ -64,6 +64,13 @@ def main(argv=sys.argv):
                 update.comment(db, text, author=u'bodhi')
                 continue
 
+            # Comment when Disabling Automatic push to stable
+            if update.autokarma and update._composite_karma[1] != 0:
+                print('%s has received negative karma' % update.title)
+                text = unicode(config.get('disable_automatic_push_to_stable'))
+                update.comment(db, text, author=u'bodhi')
+                continue
```
bowlofeggs: I think it would be desirable to make this comment at the moment that the negative karma is disabled, rather than later. This script only runs every 6 hours. I think this would be a good place to add such a comment:

https://github.com/fedora-infra/bodhi/blob/2.4.0/bodhi/server/models/models.py#L1724-L1726

bowlofeggs (changes requested):

bodhi/server/scripts/approve_testing.py
trishnaguha: Makes sense! I never knew it runs every 6 hours.

bodhi/tests/server/services/test_updates.py
```diff
@@ -3088,6 +3088,11 @@ def test_disable_autopush_non_critical_update_with_negative_karma(self, publish,
         up.comment(self.db, u'Failed to work', author=u'ralph', karma=-1)
         up = self.db.query(Update).filter_by(title=resp.json['title']).one()
 
+        expected_comment = (
+            u'Negative karma received, disabling automatic push to stable. '
+            u'Please push manually if you decide the issue is not severe.')
+        self.assertEquals(up.comments[2].text, expected_comment)
```
trishnaguha: @bowlofeggs Looks good :-) ?

trishnaguha:
@bowlofeggs Updated :-).

bodhi/server/models.py
```diff
@@ -1832,6 +1832,8 @@ def check_karma_thresholds(self, db, agent):
         if self.autokarma and self._composite_karma[1] != 0:
             log.info("Disabling Auto Push since the update has received negative karma")
             self.autokarma = False
+            text = unicode(config.get('disable_automatic_push_to_stable'))
```
bowlofeggs: It would be a good idea to put a default value here, so that Bodhi won't crash if this new config value isn't added to the config. I like to follow [Semantic Versioning](https://semver.org) where possible, so this would be an easy way to avoid a backwards incompatible change.

I think this might be as simple as changing the above line to something like this:

```
default_comment = u'Bodhi is disabling automatic push to stable due to negative karma. The maintainer may manually if they determine that the issue is not severe.'
text = unicode(config.get('disable_automatic_push_to_stable', default_comment))
```

production.ini
```diff
@@ -19,6 +19,8 @@ stablekarma_comment = This update has reached the stable karma threshold and wil
 testing_approval_msg_based_on_karma = This update has reached the stable karma threshold and can be pushed to stable now if the maintainer wishes.
 not_yet_tested_msg_based_on_karma = This update has not reached the stable karma threshold.
 
+disable_automatic_push_to_stable = Negative karma received, disabling automatic push to stable. Please push manually if you decide the issue is not severe.
```
bowlofeggs: I suggest a slightly different wording. Perhaps something like:

Bodhi is disabling automatic push to stable due to negative karma. The maintainer may manually if they determine that the issue is not severe.

This is at your option.

development.ini.example
```diff
@@ -26,6 +26,8 @@ stablekarma_comment = This update has reached the stable karma threshold and wil
 testing_approval_msg_based_on_karma = This update has reached the stable karma threshold and can be pushed to stable now if the maintainer wishes.
 not_yet_tested_msg_based_on_karma = This update has not reached stable karma threshold.
 
+disable_automatic_push_to_stable = Negative karma received, disabling automatic push to stable. Please push manually if you decide the issue is not severe.
```
bowlofeggs: I suggest a different wording here, also at your option.

bowlofeggs (changes requested):
The only thing I think we need to change is to make the new setting optional, so that bodhi can keep working without it.

production.ini
trishnaguha: Either of them works for me.

bodhi/tests/server/services/test_updates.py
```diff
@@ -3088,6 +3088,9 @@ def test_disable_autopush_non_critical_update_with_negative_karma(self, publish,
         up.comment(self.db, u'Failed to work', author=u'ralph', karma=-1)
         up = self.db.query(Update).filter_by(title=resp.json['title']).one()
 
+        expected_comment = unicode(config.get('disable_automatic_push_to_stable'))
+        self.assertEquals(up.comments[2].text, expected_comment)
```
bowlofeggs: One last thing to fix and I think we'll be good to go:

This test doesn't pass if ```development.ini``` doesn't have the new ```disable_automatic_push_to_stable``` setting. An easy way to fix it would be to put the default comment here in the call to ```config.get()``` too. It would be nice to allow development checkouts with existing development.ini files to keep working ☺

In the long run, I'd like to adjust ```bodhi.server.config``` so that all settings have sane defaults in a dictionary that is then overwritten by whatever is found in the config files. This way, all settings always have a default and we can always use ```config.get``` with one argument and can guarantee that it will be defined, regardless of what is in the ini files.

bowlofeggs (changes requested):

bodhi/tests/server/services/test_updates.py
trishnaguha: Pushed :-).

bowlofeggs (approved):
Looks great, thanks @trishnaguha!

bowlofeggs:
After merging this I noticed a typo in the message, so I made a follow up PR:

https://github.com/fedora-infra/bodhi/pull/1316

Also, rather than fixing the typo in four places, I made it so I only had to fix it in three ☺

PR #1293, opened 2017-02-17
Fix a few comments that reference bodhi.server.models.models.

bowlofeggs:


cverna:
👍

PR #1294, opened 2017-02-17
Adding the edit updates command

### LINKED ISSUE #1125: Cannot edit updates using the Bodhi CLI
I tried working around #1033 by using the CLI instead:
```
bodhi -u kkofler -e --disable-autokarma -S 1 calamares-2.4.4-5.fc25
```
but that does not work either. (Using bodhi-client-0.9.12.2-4.fc24.noarch, the latest in F24 stable.)

I get
```
fedora.client.bodhi.BodhiClientException: type is missing
Required
Update for calamares-2.4.4-5.fc25 already exists
Invalid tag: calamares-2.4.4-5.fc25 not tagged with any of the following tags [u'f22-updates-candidate', u'dist-6E-epel-testing-candidate', u'dist-5E-epel-testing-candidate', u'f21-updates-candidate', u'f24-updates-candidate', u'f23-updates-candidate', u'epel7-testing-candidate', u'f25-updates-candidate']
```
and even if I pass `-t security` to work around the "type is missing", I still get the remaining 3 errors.

The comment for the `save` method in `bodhi.py` says:
```
        To edit an existing update, you must specify the update title in
        the ``edited`` keyword argument.
```
but `/usr/bin/bodhi` passes:
```
                data = bodhi.save(builds=args[0], type_=opts.type_,
                                  bugs=opts.bugs, notes=opts.notes,
                                  request=opts.request)
```
instead.

The combination of #1033 and this bug is preventing me from pushing security updates in a timely manner. Why is it taking so long to fix such critical showstoppers?

cverna:
This commit add the edit command to the updates group it uses the same API as the
new command but set the "edited" option with the update name.

Fixes #1125

### DIFF
--- a/bodhi/client/__init__.py
+++ b/bodhi/client/__init__.py
@@ -22,6 +22,40 @@ import click
 
 from bodhi.client import bindings
 
+new_edit_options = [
+    click.option('--user', envvar='USERNAME'),
+    click.option('--password', prompt=True, hide_input=True),
+    click.option('--type', default='bugfix', help='Update type', required=True,
+                 type=click.Choice(['security', 'bugfix', 'enhancement', 'newpackage'])),
+    click.option('--notes', help='Update description'),
+    click.option('--bugs', help='Comma-seperated list of bug numbers', default=''),
+    click.option('--close-bugs', default=True, is_flag=True, help='Automatically close bugs'),
+    click.option('--request', help='Requested repository',
+                 type=click.Choice(['testing', 'stable', 'unpush'])),
+    click.option('--autokarma', default=True, is_flag=True, help='Enable karma automatism'),
+    click.option('--stable-karma', type=click.INT, help='Stable karma threshold'),
+    click.option('--unstable-karma', type=click.INT, help='Unstable karma threshold'),
+    click.option('--suggest', help='Post-update user suggestion',
+                 type=click.Choice(['logout', 'reboot'])),
+    click.option('--staging', help='Use the staging bodhi instance',
+                 is_flag=True, default=False)]
+
+
+def add_options(options):
+    """ Given a list of click options this creates a decorator that
+    will return a function used to add the options to a click command.
+    :param options: a list of click.options decorator.
+    """
+    def _add_options(func):
+        """ Given a click command and a list of click options this will
+        return the click command decorated with all the options in the list.
+        :param func: a click command function.
+        """
+        for option in reversed(options):
+            func = option(func)
+        return func
+    return _add_options
+
 
 def _warn_if_url_and_staging_set(ctx, param, value):
     """
@@ -66,26 +100,10 @@ def updates():
 
 
 @updates.command()
+@add_options(new_edit_options)
 @click.argument('builds')
-@click.option('--user', envvar='USERNAME')
-@click.option('--password', prompt=True, hide_input=True)
-@click.option('--type', default='bugfix', help='Update type', required=True,
-              type=click.Choice(['security', 'bugfix',
-                                 'enhancement', 'newpackage']))
-@click.option('--notes', help='Update description')
 @click.option('--notes-file', help='Update description from a file')
-@click.option('--bugs', help='Comma-seperated list of bug numbers', default='')
-@click.option('--close-bugs', default=True, is_flag=True, help='Automatically close bugs')
-@click.option('--request', help='Requested repository',
-              type=click.Choice(['testing', 'stable', 'unpush']))
-@click.option('--autokarma', default=True, is_flag=True, help='Enable karma automatism')
-@click.option('--stable-karma', type=click.INT, help='Stable karma threshold')
-@click.option('--unstable-karma', type=click.INT, help='Unstable karma threshold')
-@click.option('--suggest', help='Post-update user suggestion',
-              type=click.Choice(['logout', 'reboot']))
 @click.option('--file', help='A text file containing all the update details')
-@click.option('--staging', help='Use the staging bodhi instance',
-              is_flag=True, default=False)
 @url_option
 def new(user, password, url, **kwargs):
     """
@@ -124,6 +142,38 @@ def new(user, password, url, **kwargs):
             click.echo(str(e))
         except Exception as e:
             traceback.print_exc()
+
+
+@updates.command()
+@add_options(new_edit_options)
+@click.argument('updateid')
+@url_option
+def edit(user, password, url, **kwargs):
+    """
+    Edit an existing update.
+
+    Args:
+        user (unicode): The username to authenticate as.
+        password (unicode): The user's password.
+        url (unicode): The URL of a Bodhi server to create the update on. Ignored if staging is
+                       True.
+        kwargs (dict): Other keyword arguments passed to us by click.
+    """
+    client = bindings.BodhiClient(base_url=url, username=user, password=password,
+                                  staging=kwargs['staging'])
+
+    try:
+        query_param = {'updateid': kwargs['updateid']}
+        del(kwargs['updateid'])
+        resp = client.query(**query_param)
+        title = resp['updates'][0]['title']
+        kwargs['builds'] = title
+        kwargs['edited'] = title
+
+        resp = client.save(**kwargs)
+        print_resp(resp, client)
+    except bindings.BodhiClientException as e:
+        click.echo(str(e))
 
 
 @updates.command()
--- a/bodhi/tests/client/test___init__.py
+++ b/bodhi/tests/client/test___init__.py
@@ -203,7 +203,7 @@ class TestRequest(unittest.TestCase):
                 return_value=client_test_data.EXAMPLE_UPDATE_MUNCH)
     def test_successful_operation(self, send_request, __init__):
         """
-        Assert that a successful operation is handled properly.
+        Assert that a successful updates request is handled properly.
         """
         runner = testing.CliRunner()
 
@@ -358,3 +358,40 @@ class TestWarnIfUrlAndStagingSet(unittest.TestCase):
         self.assertEqual(result, 'http://localhost:6543')
         echo.assert_called_once_with(
             '\nWarning: url and staging flags are both set. url will be ignored.\n', err=True)
+
+
+class TestEdit(unittest.TestCase):
+    """
+    This class tests the edit() function.
+    """
+    @mock.patch('bodhi.client.bindings.BodhiClient.csrf',
+                mock.MagicMock(return_value='a_csrf_token'))
+    @mock.patch('bodhi.client.bindings.BodhiClient.query',
+                return_value=client_test_data.EXAMPLE_QUERY_MUNCH, autospec=True)
+    @mock.patch('bodhi.client.bindings.BodhiClient.send_request',
+                return_value=client_test_data.EXAMPLE_UPDATE_MUNCH, autospec=True)
+    def test_url_flag(self, send_request, query):
+        """
+        Assert that a successful updates edit request is handled properly.
+        """
+        runner = testing.CliRunner()
+
+        result = runner.invoke(
+            client.edit, ['FEDORA-2017-cc8582d738', '--user', 'bowlofeggs',
+                          '--password', 's3kr3t', '--notes', 'this is an edited note',
+                          '--url', 'http://localhost:6543'])
+
+        self.assertEqual(result.exit_code, 0)
+        bindings_client = query.mock_calls[0][1][0]
+        query.assert_called_with(
+            bindings_client, updateid=u'FEDORA-2017-cc8582d738')
+        bindings_client = send_request.mock_calls[0][1][0]
+        send_request.assert_called_with(
+            bindings_client, 'updates/', auth=True, verb='POST',
+            data={
+                'close_bugs': True, 'stable_karma': None, 'csrf_token': 'a_csrf_token',
+                'staging': False, 'builds': u'nodejs-grunt-wrap-0.3.0-2.fc25', 'autokarma': True,
+                'edited': u'nodejs-grunt-wrap-0.3.0-2.fc25', 'suggest': None,
+                'notes': u'this is an edited note', 'request': None, 'bugs': u'',
+                'unstable_karma': None, 'type': 'bugfix'})
+        self.assertEqual(bindings_client.base_url, 'http://localhost:6543/')
--- a/docs/release_notes.rst
+++ b/docs/release_notes.rst
@@ -16,6 +16,8 @@ Features
   (`#1366 <https://github.com/fedora-infra/bodhi/pulls/1366>`_).
 * The web UI footer has been added with the documentation link.
   (`#1321 <https://github.com/fedora-infra/bodhi/issues/1321>`_).
+* The bodhi CLI now supports editing an update using the update id.
+  (`#937 <https://github.com/fedora-infra/bodhi/issues/937>`_)
 
 
 Bugs
@@ -56,6 +58,7 @@ The following contributors submitted patches for REPLACE THIS WITH A VERSION BEF
 * Ryan Lerch
 * Bianca Nenciu
 * Randy Barlow
+* Clement Verna
 
 
 2.5.0
--- a/docs/man_pages/bodhi.rst
+++ b/docs/man_pages/bodhi.rst
@@ -172,6 +172,48 @@ The ``updates`` command allows users to interact with bodhi updates.
 
         A path to a file containing all the update details.
 
+``bodhi updates edit [options] <updateid>``
+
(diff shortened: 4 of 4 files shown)


bodhi/client/__init__.py
```diff
@@ -84,6 +84,39 @@ def new(user, password, **kwargs):
...
+def edit(user, password, **kwargs):
+    client = bindings.BodhiClient(username=user, password=password, staging=kwargs['staging'])
+    kwargs['edited'] = kwargs['builds']
+    try:
+        resp = client.save(**kwargs)
+        print_resp(resp, client)
+    except bindings.BodhiClientException as e:
+        click.echo(str(e))
+    except Exception as e:
+        traceback.print_exc()
```
bowlofeggs: I realize that the new() function does this too, but if we're going to print a traceback and exit, you might as well just not catch this Exception and let Python crash (i.e., I think you can delete the above two lines).

bodhi/client/__init__.py
```diff
@@ -84,6 +84,39 @@ def new(user, password, **kwargs):
...
+@click.option('--request', help='Requested repository',
+              type=click.Choice(['testing', 'stable', 'unpush']))
+@click.option('--autokarma', default=True, is_flag=True, help='Enable karma automatism')
+@click.option('--stable-karma', type=click.INT, help='Stable karma threshold')
+@click.option('--unstable-karma', type=click.INT, help='Unstable karma threshold')
+@click.option('--suggest', help='Post-update user suggestion',
+              type=click.Choice(['logout', 'reboot']))
+@click.option('--file', help='A text file containing all the update details')
+@click.option('--staging', help='Use the staging bodhi instance',
+              is_flag=True, default=False)
```
bowlofeggs: It might be possible to create a new Python function decorator that adds all of these flags. Then you can just use the new decorator on this function and the new function, for at least the flags they have in common. That might be the easiest way to avoid duplicating code.

bowlofeggs (changes requested):
One more change we'll want before we can merge is test coverage for the new code.

cverna:
Reading some of the click issues (https://github.com/pallets/click/issues/108), I found a pretty neat way to use a decorator. 
I ll start looking at unit test for this change.

bodhi/client/__init__.py
```diff
@@ -16,12 +16,37 @@
...
+    click.option('--staging', help='Use the staging bodhi instance',
+                 is_flag=True, default=False)]
+
+
+def add_options(options):
+    def _add_options(func):
+        for option in reversed(options):
+            func = option(func)
+        return func
+    return _add_options
```
bowlofeggs: Nice, this is a clever way to do this!

bodhi/client/__init__.py
```diff
@@ -16,12 +16,37 @@
...
+    click.option('--autokarma', default=True, is_flag=True, help='Enable karma automatism'),
+    click.option('--stable-karma', type=click.INT, help='Stable karma threshold'),
+    click.option('--unstable-karma', type=click.INT, help='Unstable karma threshold'),
+    click.option('--suggest', help='Post-update user suggestion',
+                 type=click.Choice(['logout', 'reboot'])),
+    click.option('--staging', help='Use the staging bodhi instance',
+                 is_flag=True, default=False)]
+
+
+def add_options(options):
```
bowlofeggs, line 44: Let's give this wrapper a docblock that explains its purpose, options argument, and return value.

bodhi/client/__init__.py
```diff
@@ -16,12 +16,37 @@
...
+    def _add_options(func):
```
bowlofeggs: Let's also put a docblock on this function.

bodhi/client/__init__.py
```diff
@@ -16,12 +16,37 @@
 import platform
 import subprocess
 import sys
-import traceback
 
 import click
 
 from bodhi.client import bindings
 
+common_options = [
```
bowlofeggs: Since these options aren't common for all commands, let's rename this to be a little more specific, like ```new_edit_options``` or something.

bodhi/client/__init__.py
```diff
@@ -79,8 +89,20 @@ def new(user, password, **kwargs):
             print_resp(resp, client)
         except bindings.BodhiClientException as e:
             click.echo(str(e))
-        except Exception as e:
-            traceback.print_exc()
```
bowlofeggs: Let's not make this change in this commit, since it's unrelated to adding an edit command.

bodhi/client/__init__.py
```diff
@@ -79,8 +89,20 @@ def new(user, password, **kwargs):
...
+
+
+@updates.command()
+@add_options(common_options)
+@click.argument('builds')
+def edit(**kwargs):
```
bowlofeggs: Even though ```new()``` doesn't have a docblock, let's give this new function one. I can give ```new()``` one sometime in the future, but it's good practice to make sure all new functions in Bodhi have docblocks.

bodhi/tests/client/test___init__.py
```diff
@@ -80,3 +80,66 @@ def test_update_not_found(self, send_request, __init__):
...
+
+        self.assertEqual(result.exit_code, 0)
+        send_request.assert_called_once_with(
+            'updates/', verb='POST', auth=True,
+            data={'close_bugs': True, 'stable_karma': None, 'csrf_token': 'a_csrf_token',
+                  'staging': False, 'builds': 'bodhi-2.2.4-1.el7', 'autokarma': True,
+                  'suggest': None, 'notes': 'this is an edited note', 'request': None, 'bugs': '',
+                  'unstable_karma': None, 'user': 'some_user', 'password': 's3kr3t',
+                  'type': 'bugfix', 'edited': u'bodhi-2.2.4-1.el7'})
+        __init__.assert_called_once_with(username='some_user', password='s3kr3t', staging=False)
```
bowlofeggs: Nice tests!

bowlofeggs (changes requested):
This is a very nice way to do this, nice work! I just have a few more changes to get in here and we'll be good to go!

bowlofeggs:
One more thing - could you add a release note to ```docs/release_notes.rst``` for this new feature? There's not a section for 2.5.0 yet, so feel free to add it.

cverna:
PR updated ;)

cverna:
@bowlofeggs can this be merged ?

docs/release_notes.rst
```diff
@@ -1,6 +1,25 @@
...
 
+2.5.0
+-----
+
+Bodhi 2.5.0 is a feature and bugfix release.
+
+
+Features
+^^^^^^^^
+* The ``bodhi`` CLI now supports the edit command for updates.
```
bowlofeggs: It would be good to include a link to the issue here - you can see examples of that in the 2.4.0 release notes below.

docs/release_notes.rst
```diff
@@ -1,6 +1,25 @@
...
+
+
+Release contributors
+^^^^^^^^^^^^^^^^^^^^
+
+The following contributors submitted patches for Bodhi 2.4.0:
```
bowlofeggs: s/2.4.0/2.5.0/

bowlofeggs (changes requested):
This pull request seems to have a failing test:

```
FAIL: Assert that a successful updates new request is handled properly.
  File "/home/vagrant/bodhi/bodhi/tests/client/test___init__.py", line 112, in test_successful_operation
    'notes_file': None, 'password': 's3kr3t', 'type': 'bugfix'})
  File "/usr/lib/python2.7/site-packages/mock/mock.py", line 948, in assert_called_once_with
    return self.assert_called_with(*args, **kwargs)
  File "/usr/lib/python2.7/site-packages/mock/mock.py", line 937, in assert_called_with
    six.raise_from(AssertionError(_error_message(cause)), cause)
  File "/usr/lib/python2.7/site-packages/six.py", line 718, in raise_from
    raise value
AssertionError: Expected call: send_request('updates/', auth=True, data={'close_bugs': True, 'stable_karma': None, 'csrf_token': 'a_csrf_token', 'staging': False, 'builds': 'bodhi-2.2.4-1.el7', 'autokarma': True, 'suggest': None, 'notes':
 'this is a note', 'request': None, 'bugs': '', 'unstable_karma': None, 'user': 'some_user', 'file': None, 'notes_file': None, 'password': 's3kr3t', 'type': 'bugfix'}, verb='POST')
Actual call: send_request('updates/', auth=True, data={'close_bugs': True, 'stable_karma': None, 'csrf_token': 'a_csrf_token', 'staging': False, 'builds': u'bodhi-2.2.4-1.el7', 'autokarma': True, 'suggest': None, 'notes': u'this is a note
', 'request': None, 'bugs': u'', 'unstable_karma': None, 'file': None, 'notes_file': None, 'type': 'bugfix'}, verb='POST')
```

cverna:
Updated ;), I removed a bit too much from the previous review that's why the test was failing. Good catch ;).

bodhi/client/__init__.py
```diff
@@ -84,6 +103,23 @@ def new(user, password, **kwargs):
 
 
 @updates.command()
+@add_options(new_edit_options)
+@click.argument('builds')
```
bowlofeggs: While using this I realized that we probably don't want to restrict the way that users express which update they want to edit to builds. Some updates have a massive number of builds (like Gnome or KDE) and this will make it difficult to express those updates. I think we want to think of a way to accept update id (like ```FEDORA-2017-2d576b8eb4```). If we can think of a way to use the id or the builds and have the client figure it out, that would also be nice. There's a regex in the bindings that tries to do something like this.

bodhi/client/__init__.py
```diff
@@ -35,28 +69,13 @@ def updates():
...
 @updates.command()
+@add_options(new_edit_options)
 @click.argument('builds')
-@click.option('--user', envvar='USERNAME')
-@click.option('--password', prompt=True, hide_input=True)
-@click.option('--type', default='bugfix', help='Update type', required=True,
-              type=click.Choice(['security', 'bugfix',
-                                 'enhancement', 'newpackage']))
-@click.option('--notes', help='Update description')
 @click.option('--notes-file', help='Update description from a file')
```
bowlofeggs, line 105: Does it make sense to include this in the common list too?

bowlofeggs (changes requested):
A few more changes - apologies I didn't think of the update id one earlier!

I tried to try this out on staging, but apparently staging is unable to edit updates due to postgres replication (I'll file an issue about that in a moment) so I tried it on one of my updates on production and it seemed to work well. Nice!

bodhi/client/__init__.py
cverna: Couldn't you want to edit the type of an update ? Same for notes ? 
I don't know I don't have much user experience of bodhi, maybe it's better to give the user the choice.

bodhi/client/__init__.py
cverna: Ok, I believe this will need some changes on the server side too. I have to look at it.

bodhi/client/__init__.py
bowlofeggs: Yes you would want to be able to edit the notes and the type, but it looks like you have put those in there so I think maybe adding the ```--notes-file``` makes sense too.

bodhi/client/__init__.py
bowlofeggs: I apologize again for not thinking of this sooner, but I think it should be pretty doable.

BTW, I have an open PR that makes it so the CLI can be told to connect to the local development server - it might be helpful for you as you work on this: #1314. With that PR the Vagrant environment automatically uses the local server instead of production, or you can set the BODHI_URL environment variable or the ```--url``` flag to whatever you like.

You can also achieve this by editing the ```__init__()``` method in ```bodhi.client.bindings``` without the PR, but you have to be careful not to commit that change ☺

bowlofeggs:
I just merged https://github.com/fedora-infra/bodhi/pull/1314 which is causing some merge conflicts, but it's also good news because now the cli can be more easily connected to the local development instance. You can set BODHI_URL in your dev environment, or you can just re-run vagrant provision if you are using Vagrant and that'll get set for you in the dev environment.

cverna:
Cool, I ll rebase my branch I did not have much time last week to spend on this, I will see how it goes this week :)

bowlofeggs:

bowlofeggs:
I was thinking about this this morning, and it occurred to me that there may be an easy way to work around the API that might have a slight performance impact but I think will be OK at least for an initial implementation.

There's a regex inside the bindings to detect whether a Fedora or EPEL update is being specified. You could pull that regex to the module level and use it to detect if a user has specified an update by id vs. by title. If you detect that they are using an update id, you can then use the query API to find out what that update's title is, and then set the edited field to the title. It's two API calls instead of one, but it'll work and we can sort the API change out later. I'm thinking about altering the API for containers anyway, so perhaps we can do this nicer than this at a later time. I think this will be a nice approach in the meantime and it won't be so much work for you.

How does that sound?

cverna:
At first I did go that way, but I thought it would be more an hack than a nice solution. I might not be that far with just passing updateid as an argument. What I can do is to update this PR with my current development then we can see which solution to go for.

cverna:
Still need to update the test and the release note.

bowlofeggs (changes requested):
There are currently two test failures:

```
/home/vagrant/bodhi/bodhi/client/__init__.py:154:101: E501 line too long (111 > 100 characters)
F.
FAIL: Assert correct behavior with the --url flag.
  File "/home/vagrant/bodhi/bodhi/tests/client/test___init__.py", line 126, in test_url_flag
    'unstable_karma': None, 'file': None, 'notes_file': None, 'type': 'bugfix'})
  File "/usr/lib/python2.7/site-packages/mock/mock.py", line 318, in assert_called_once_with
    return mock.assert_called_once_with(*args, **kwargs)
  File "/usr/lib/python2.7/site-packages/mock/mock.py", line 948, in assert_called_once_with
    return self.assert_called_with(*args, **kwargs)
  File "/usr/lib/python2.7/site-packages/mock/mock.py", line 937, in assert_called_with
    six.raise_from(AssertionError(_error_message(cause)), cause)
  File "/usr/lib/python2.7/site-packages/six.py", line 718, in raise_from
    raise value
AssertionError: Expected call: send_request(<bodhi.client.bindings.BodhiClient object at 0x7f9a64628b10>, 'updates/', auth=True, data={'close_bugs': True, 'csrf_token': 'a_csrf_token', 'unstable_karma': None, 'file': None, 'notes_file': N
one, 'stable_karma': None, 'staging': False, 'builds': u'bodhi-2.2.4-1.el7', 'autokarma': True, 'suggest': None, 'notes': None, 'request': None, 'bugs': u'', 'type': 'bugfix'}, verb='POST')
Actual call: send_request(<bodhi.client.bindings.BodhiClient object at 0x7f9a64628b10>, 'updates/', auth=True, data={'close_bugs': True, 'stable_karma': None, 'csrf_token': 'a_csrf_token', 'staging': False, 'builds': u'bodhi-2.2.4-1.el7',
 'autokarma': True, 'url': u'http://localhost:6543', 'suggest': None, 'notes': None, 'request': None, 'bugs': u'', 'unstable_karma': None, 'user': u'bowlofeggs', 'file': None, 'notes_file': None, 'password': u's3kr3t', 'type': 'bugfix'}, verb='POST')
fedora.client.openidbaseclient: DEBUG: No pre-existing session for http://localhost:6543/:bowlofeggs

AssertionError: 1 != 0
```

cverna:
Thanks for the catch, it should be ok now 🍍

bowlofeggs (approved):
Cool, this looks nice and seems to work well. I noticed that editing an EL 7 update failed but since editing an F26 update works I'm going to assume it's some server-side problem and I'll just file an issue about it.

I'd also like to make the --notes-file argument be shared, but we can do that as a separate change as well.

I also have an idea for how we can make it accept updates by name or by ID, but again, I'll file an issue to expand support for that later because we can do that in a backwards compatible way.

I'm really happy that you took this on and did a great job of avoiding copied code. This was a major feature regression with the bodhi 0.9 client and with your patch bodhi 2 is much more complete.

Nice work!

bowlofeggs:
I filed https://github.com/fedora-infra/bodhi/issues/1407 about the failure to edit that update. It also fails on the web UI so it's definitely not related to this change (I didn't think it was anyway).

bowlofeggs:
I filed https://github.com/fedora-infra/bodhi/issues/1408 about adding ```--notes-file``` to the CLI edit command too.

bowlofeggs:
I filed https://github.com/fedora-infra/bodhi/issues/1409 to track the idea of allowing editing by name or ID. It's solidly in the nice to have category but I think it'd be cool.

PR #1295, opened 2017-02-18
PEP-8 the tools folder.

bowlofeggs:

### DIFF
--- a/tools/metrics.py
+++ b/tools/metrics.py
@@ -13,29 +13,24 @@
 # You should have received a copy of the GNU General Public License
 # along with this program; if not, write to the Free Software
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
-
 """
 A tool for generating statistics for each release.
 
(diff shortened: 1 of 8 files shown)


tools/metrics.py
```diff
@@ -145,17 +139,21 @@ def main(releases=None):
...
+                                'wishes'):
                             data['num_testingtime'] += 1
                             stablekarma_done = True
 
                 # For figuring out if an update has received feedback or not
                 if not feedback_done:
-                    if (not comment.user.name == 'bodhi'
-                        and comment.karma != 0 and not comment.anonymous):
+                    if (not comment.user.name == 'bodhi' and
+                            comment.karma != 0 and not comment.anonymous):
```
pypingou, line 156: I'm in the side of putting the keyword at the start of the line :)

pypingou:
👍 for me

tools/metrics.py
bowlofeggs: @pypingou My vim linter (which is probably just flake8) complains if I put the and operator on the next line, so I think I'll keep it here. I don't personally have a preference, but I like my vim linter to be happy ☺

PR #1297, opened 2017-02-20
Changed the cursor to pointer when it hovers over Releases button

### LINKED ISSUE #1296: Change the mouse cursor when it hovers over  `Releases` to pointer.
When we hover mouse cursor over the `Releases` button it changes to auto which seems odd as it should change to pointer just like as the `login` button.
https://bodhi.fedoraproject.org/

ankit01ojha:
fixes https://github.com/fedora-infra/bodhi/issues/1296


bowlofeggs (approved):
Seems to work well, thanks!

PR #1299, opened 2017-02-22
Add symlink:README.rst -> README

trishnaguha:
This way we can see the exact syntax of `README.rst` in `README` file over web without affecting development/prod environment. Also it makes sure that updating any of the file will affect the other one. Inspired from https://github.com/elastic/elasticsearch-dsl-py.


bowlofeggs:
I'm not sure I understand the benefit of this change. GitHub seems to render the page the same way before and after this change. Also, if you want to see the syntax of the file, you can always visit its raw link:

https://raw.githubusercontent.com/fedora-infra/bodhi/develop/README.rst

I think I'm in favor of keeping the file count down in the repo in this case.

trishnaguha:
Adding symlink would make Github rendering the README file as rst. This is at your option whether you would like to have it or not. Ofcourse this is not a *must* having feature (Feel free to close the PR in that case). But I liked the idea.

bowlofeggs:
When I compare https://github.com/trishnaguha/bodhi/tree/symlink-readme to https://github.com/fedora-infra/bodhi they seem to get rendered the same. Is there a different set of URLs that demonstrate the difference this PR makes on rendering?

trishnaguha:
@bowlofeggs You will see the difference if you look at https://github.com/trishnaguha/bodhi/blob/symlink-readme/README and https://github.com/fedora-infra/bodhi/blob/develop/README.rst 
https://github.com/trishnaguha/bodhi/blob/symlink-readme/README.rst

bowlofeggs:
It's already possible to get GitHub to show the raw readme, by going to the raw link:

https://raw.githubusercontent.com/fedora-infra/bodhi/develop/README.rst

I think we should avoid adding extra files in this case.

PR #1301, opened 2017-02-22
Some Python bindings fixes, and raise test coverage to 82%

bowlofeggs:
This pull request contains 6 commits. The first 4 commits fix various problems discovered while testing ```bodhi.client.bindings```, the 5th commit adds many tests for the bindings, and the last commit requires test coverage to remain at 82%.

### DIFF
--- a/bodhi/client/bindings.py
+++ b/bodhi/client/bindings.py
@@ -33,7 +33,6 @@ import logging
 import os
 import re
 import textwrap
-import warnings
 
 import six
 
@@ -165,13 +164,6 @@ class BodhiClient(OpenIdBaseClient):
                 raise UpdateNotFound(update)
             else:
                 raise
-
-    @errorhandled
-    def delete(self, update):
-        warnings.warn('Deleting updates has been disabled in Bodhi. '
-                      'This API call will unpush the update instead. '
-                      'Please use `set_request(update, "unpush")` instead')
-        self.request(update, 'unpush')
 
     @errorhandled
     def query(self, **kwargs):
--- a/bodhi/client/bindings.py
+++ b/bodhi/client/bindings.py
@@ -228,7 +228,10 @@ class BodhiClient(OpenIdBaseClient):
                 kwargs['packages'] = kwargs['package']
             del(kwargs['package'])
         if 'release' in kwargs:
-            kwargs['releases'] = kwargs['release']
+            if isinstance(kwargs['release'], list):
+                kwargs['releases'] = kwargs['release']
+            else:
+                kwargs['releases'] = [kwargs['release']]
             del(kwargs['release'])
         if 'type_' in kwargs:
             kwargs['type'] = kwargs['type_']
--- a/bodhi/client/bindings.py
+++ b/bodhi/client/bindings.py
@@ -372,7 +372,8 @@ class BodhiClient(OpenIdBaseClient):
                 for update in update_list:
                     yield update
 
-    def override_str(self, override):
+    @staticmethod
+    def override_str(override):
         """ Return a string representation of a given override dictionary.
 
         :arg override: An override dictionary.
--- a/bodhi/client/bindings.py
+++ b/bodhi/client/bindings.py
@@ -280,10 +280,7 @@ class BodhiClient(OpenIdBaseClient):
 
     @errorhandled
     def list_overrides(self, user=None):
-        """ Save a buildroot overrides.
-
-        This entails either creating a new buildroot override, or editing an
-        existing one.
+        """ List buildroot overrides.
 
         :kwarg user: A username whose buildroot overrides you want returned.
 
--- a/bodhi/client/bindings.py
+++ b/bodhi/client/bindings.py
@@ -222,7 +222,7 @@ class BodhiClient(OpenIdBaseClient):
             # update ID, so try and figure it out
             if re.search(r'(\.el|\.fc)\d\d?', kwargs['package']):
                 kwargs['builds'] = kwargs['package']
-            elif re.search(r'FEDORA-(EPEL)?-\d{4,4}', kwargs['package']):
+            elif re.search(r'FEDORA-(EPEL-)?\d{4,4}', kwargs['package']):
                 kwargs['updateid'] = kwargs['package']
             else:
                 kwargs['packages'] = kwargs['package']
--- a/bodhi/tests/client/test_bindings.py
+++ b/bodhi/tests/client/test_bindings.py
@@ -14,6 +14,7 @@
 # along with this program; if not, write to the Free Software
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
 """This module contains tests for bodhi.client.bindings."""
+from datetime import datetime, timedelta
 import copy
 import unittest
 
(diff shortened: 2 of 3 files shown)


AdamWill:
For the record, the regex bug was not only "originally reported and fixed" by me, it was *created*, reported and fixed by me. :P

bowlofeggs:
@AdamWill hahahah well I'm still happy to have the fix ☺

bodhi/client/bindings.py
```diff
@@ -230,7 +222,7 @@ def query(self, **kwargs):
             # update ID, so try and figure it out
             if re.search(r'(\.el|\.fc)\d\d?', kwargs['package']):
                 kwargs['builds'] = kwargs['package']
-            elif re.search(r'FEDORA-(EPEL)?-\d{4,4}', kwargs['package']):
+            elif re.search(r'FEDORA-(EPEL-)?\d{4,4}', kwargs['package']):
```
jeremycline, line 225: Hmm, what happens in the year 10,000? Think of the children!

bodhi/tests/client/test_bindings.py
```diff
@@ -24,6 +25,435 @@
...
+    """
+    def test_comment(self):
+        """
+        Test the comment() method.
+        """
+        client = bindings.BodhiClient()
+        client.csrf_token = 'a token'
+        client.send_request = mock.MagicMock(return_value='response')
+
+        response = client.comment('bodhi-2.4.0-1.fc25', 'It ate my cat!', karma=-1, email=True)
```
jeremycline, line 74: I hate when that happens 🐱

bodhi/tests/client/test_bindings.py
```diff
@@ -24,6 +25,435 @@
...
+    """
+    def test_with_dict(self):
+        """
+        Test override_str() with a dict argument.
+        """
+        override = bindings.BodhiClient.override_str('this is an override')
+
+        self.assertEqual(override, 'this is an override')
+
+    def test_with_str(self):
```
jeremycline: I think you got these mixed up, this one uses a dict, the other one uses a string

bodhi/tests/client/test_bindings.py
```diff
@@ -24,6 +25,435 @@
...
+        Test with a 'release' kwarg.
+        """
+        client = bindings.BodhiClient()
+        client.send_request = mock.MagicMock(return_value='return_value')
+
+        result = client.query(packages='bodhi', release='f26')
+
+        self.assertEqual(result, 'return_value')
+        client.send_request.assert_called_once_with(
+            'updates/', verb='GET', params={'packages': 'bodhi', 'releases': 'f26'})
```
jeremycline: Should it be wrapping the value of 'releases' in a list if it's not already a list? I just noticed the docblock says it can be a list (and its name implies that) so I wondered.

bodhi/tests/client/test_bindings.py
```diff
@@ -24,6 +25,435 @@
...
+        client.csrf_token = 'a token'
+        kwargs = {
+            'builds': ['bodhi-2.4.0-1.fc26'], 'type_': 'enhancement', 'notes': 'This is a test',
+            'request': 'testing', 'autokarma': True, 'stable_karma': 3, 'unstable_karma': -3,
+            'severity': 'low'}
+
+        response = client.save(**kwargs)
+
+        self.assertEqual(response, 'return_value')
+        kwargs['type'] = kwargs['type_']
```
jeremycline, line 417: Should it be popping the old ``type_`` kwarg before sending the request?

bodhi/tests/client/test_bindings.py
```diff
@@ -24,6 +25,435 @@
...
+        client.send_request.assert_called_once_with(
+            'overrides/', verb='POST', auth=True,
+            data={'nvr': 'python-pyramid-1.5.6-3.el7',
+                  'expiration_date': actual_expiration,
+                  'notes': 'This is needed to build bodhi-2.4.0.', 'csrf_token': 'a token'})
+        # Since we can't mock utcnow() since it's a C extension, let's just make sure the expiration
+        # date sent is within 2 seconds of the now variable. It would be surprising if it took more
+        # than 2 seconds to start the function and execute its first instruction!
+        expected_expiration = now + timedelta(days=2)
+        self.assertTrue((actual_expiration - expected_expiration) < timedelta(seconds=2))
```
jeremycline: This may be a flaky test. Not because it's likely it'll take more than two seconds to run (although I guess that's possibly with very unlucky scheduling), but because if a system clock update occurs between the first call and the second call to ``utcnow`` and the drift was more than 2 seconds (which seems like a very small number for drift) this won't be true.

Maybe that won't happen much (or ever!), just thought I'd bring it up.

jeremycline (changes requested):
Overall it looks good, there's just a pair of tests whose names need switching. There's also a time-based test which you may want to re-consider, but I leave that up to your judgment.

bodhi/client/bindings.py
bowlofeggs: Hahahaha, well I guess we'll have to go back to the drawing board, but only if humans still exist then…

bodhi/tests/client/test_bindings.py
bowlofeggs: ☺

bodhi/tests/client/test_bindings.py
bowlofeggs: The ```release``` argument is actually there for backwards compatibility with the bodhi 1 client. I'm not sure anything would be using this, especially with the newly relocated bindings. However, I decided to keep it for now to minimize changes. Perhaps it can be removed with Bodhi 3.

I think you are right that it should be converting it to a list. I'll add another commit that fixes this and modifies the test.

bodhi/tests/client/test_bindings.py
bowlofeggs: The request got sent two lines above in the call to ```client.save()```. This line is just adjusting the kwargs for the purpose of asserting that ```save()``` properly modified the kwargs.

bodhi/tests/client/test_bindings.py
bowlofeggs: Do you have any specific recommendations for what you think would be better here? I think those scenarios are fairly unlikely, though not impossible. If they were to happen, we could raise the threshold later.

bowlofeggs:
@jeremycline I've rebased this pull request and fixed everything you noted except for the timing issue and the kwargs popping (since that was for test assertion). What do you think?

bodhi/tests/client/test_bindings.py
jeremycline: I think what I meant here was, should it be asserting that both ``type`` and ``type_`` are keywords in the data dictionary that ``send_request`` is called with?

bodhi/tests/client/test_bindings.py
jeremycline: Not really, unfortunately. You could probably raise it to something like 5 minutes if the expected expiration is two days. Two days is 2880 minutes so give or take 5 doesn't seem like a big deal. It's totally up to you.

jeremycline (approved):

jeremycline:
I clarified my comments, but neither is a blocking issue to me, so merge whenever you're ready :shipit:

bowlofeggs:
On Tue, 2017-03-07 at 11:03 -0800, Jeremy Cline wrote:
> I think what I meant here was, should it be asserting that both type
> and type_ are keywords in the data dictionary that send_request is
> called with?

That is what the underlying library is doing, and this is again to keep
compatibility with the v1 library. Since I have no knowledge of Bodhi
v1 it's a little hard to say what should happen here, though it does
seem that dropping the extra kwarg makes sense. I'm inclined to leave
it as-is for now and just drop all this code with Bodhi 3.

Is that OK by you?

bowlofeggs:
Ah I will take your last comment to indicate that it is OK by you. Thanks for the review! I'll fix that window to be 5 minutes.

PR #1306, opened 2017-02-22
Correctly calculate how many days to stable an Update has.

### LINKED ISSUE #1305: The Days to Stable calculation in an update seems backwards
I've noticed that updates don't show "Days to Stable" until they are already past the date they can be pushed to stable, and that it seems to be counting how many days it's been *since* the update was eligible to be pushed to stable. It seems like it's doing the math backwards. [Here is the code](https://github.com/fedora-infra/bodhi/blob/2.4.0/bodhi/server/models/models.py#L1856-L1864) that does this:

```
    @property
    def days_to_stable(self):
        """ Return the number of days until an update can be pushed to stable """
        if self.meets_testing_requirements:
            return (
                self.release.mandatory_days_in_testing -
                (datetime.utcnow() - self.date_testing).days)
        else:
            return 0
```

First, it doesn't make sense to me to guard against whether the update meets testing requirements. We want to show how many days the update has left to get to stable, and if it hasn't met those days it might very well not meet testing requirements. This explains why updates never show this label until after they can be pushed to stable.

Furthermore, the math is actually not entirely wrong, but the template is stripping out the ```-``` symbol on updates after they are able to be pushed to stable. This code is called from the ```update.html``` template like this:

```
        % if update.days_to_stable and update.status.description == 'testing':
        <tr>
          <td>Days to Stable</td>
          <td>${str(update.days_to_stable).strip('-')}</td>
        </tr>
        % endif
```

I think we probably want to fix the logic so it doesn't test for ```meets_testing_requirements```, and so that it returns 0 if the calculation returns a negative result. With that last bit, we can drop the ```.strip('-')``` from the template as well.

bowlofeggs:
The Update.days_to_stable() property only calculated the number of
days to stable for updates that were already eligible to go to
stable, in which case it returned negative values only. This commit
reworks the function so that it correctly returns the number of
days until an update can be pushed to stable, or 0 if the update
can already be pushed to stable.

fixes #1305

### DIFF
--- a/bodhi/server/templates/update.html
+++ b/bodhi/server/templates/update.html
@@ -499,7 +499,7 @@ $(document).ready(function(){
       % if update.days_to_stable and update.status.description == 'testing':
       <tr>
         <td>Days to Stable</td>
-        <td>${str(update.days_to_stable).strip('-')}</td>
+        <td>${str(update.days_to_stable)}</td>
       </tr>
       % endif
 
--- a/bodhi/tests/server/models/test_models.py
+++ b/bodhi/tests/server/models/test_models.py
@@ -219,6 +219,46 @@ class TestUpdate(ModelTest):
         eq_(self.obj.builds[0].release.name, u'F11')
         eq_(self.obj.builds[0].package.name, u'TurboGears')
 
+    def test_days_to_stable_meets_testing_requirements(self):
+        """
+        The Update.days_to_stable() method should return 0 if Update.meets_testing_requirements()
+        returns True.
+        """
+        update = self.obj
+        update.autokarma = False
+        update.stable_karma = 1
+        update.comment(self.db, u'I found $100 after applying this update.', karma=1,
+                       author=u'bowlofeggs')
+        # Assert that our preconditions from the docblock are correct.
+        eq_(update.meets_testing_requirements, True)
+
+        eq_(update.days_to_stable, 0)
+
+    def test_days_to_stable_not_meets_testing_requirements_no_date_testing(self):
+        """
+        The Update.days_to_stable() method should return 0 if Update.meets_testing_requirements()
+        returns False but the Update's date_testing attribute is not truthy.
+        """
+        update = self.get_update()
+        # Assert that our preconditions from the docblock are correct.
+        eq_(update.meets_testing_requirements, False)
+        eq_(update.date_testing, None)
+
+        eq_(update.days_to_stable, 0)
+
+    def test_days_to_stable_not_meets_testing_requirements_with_date_testing(self):
+        """
+        The Update.days_to_stable() method should return a positive integer if
+        Update.meets_testing_requirements() returns False and the Update's date_testing attribute is
+        truthy.
+        """
+        update = self.get_update()
+        update.date_testing = datetime.utcnow() + timedelta(days=-4)
+        # Assert that our preconditions from the docblock are correct.
+        eq_(update.meets_testing_requirements, False)
+
+        eq_(update.days_to_stable, 3)
+
     @mock.patch('bodhi.server.models.bugtracker.close')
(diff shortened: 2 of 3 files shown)


bodhi/server/templates/update.html
```diff
@@ -499,7 +499,7 @@
       % if update.days_to_stable and update.status.description == 'testing':
       <tr>
         <td>Days to Stable</td>
-        <td>${str(update.days_to_stable).strip('-')}</td>
+        <td>${str(update.days_to_stable)}</td>
```
pypingou, line 502: haha :)

bodhi/tests/server/models/test_models.py
```diff
@@ -219,6 +219,46 @@ def test_builds(self):
...
 
+    def test_days_to_stable_meets_testing_requirements(self):
+        """
+        The Update.days_to_stable() method should return 0 if Update.meets_testing_requirements()
+        returns True.
+        """
+        update = self.obj
+        update.autokarma = False
+        update.stable_karma = 1
+        update.comment(self.db, u'I found $100 after applying this update.', karma=1,
```
pypingou, line 230: that's some good update you have there, and good karma :-p

bodhi/tests/server/models/test_models.py
```diff
@@ -219,6 +219,46 @@ def test_builds(self):
...
+        eq_(update.meets_testing_requirements, False)
+        eq_(update.date_testing, None)
+
+        eq_(update.days_to_stable, 0)
+
+    def test_days_to_stable_not_meets_testing_requirements_with_date_testing(self):
+        """
+        The Update.days_to_stable() method should return a positive integer if
+        Update.meets_testing_requirements() returns False and the Update's date_testing attribute is
+        truthy.
```
pypingou, line 253: Are you testing if the date_testing attribute is truthy somewhere?

pypingou:
One question, otherwise 👍

bowlofeggs:
On Thu, 2017-02-23 at 09:47 -0800, Pierre-Yves Chibon wrote:
> -        <td>${str(update.days_to_stable).strip('-')}</td>
> +        <td>${str(update.days_to_stable)}</td>
> 
> haha :)

Yeah I laughed when I saw this too. days_to_stable always returns 0 or
a negative number (without this patch) so rather than fixing that
someone just stripped off the minus sign. With this patch, it always
return 0 or a positive number ☺

bowlofeggs:
On Thu, 2017-02-23 at 09:48 -0800, Pierre-Yves Chibon wrote:
> that's some good update you have there, and good karma :-p

Yeah but then I got a speeding ticket for giving out too much karma in
a day and I had to pay $100. So in the end it all balances out…

bodhi/tests/server/models/test_models.py
```diff
@@ -219,6 +219,46 @@ def test_builds(self):
...
+        """
+        update = self.get_update()
+        update.date_testing = datetime.utcnow() + timedelta(days=-4)
```
bowlofeggs, line 256: @pypingou I'm explicitly setting date_testing "truthy" here, so I didn't think it made sense to then test that it is truth immediately afterwards. If I did that, I'd just be making sure that Python works ☺

pypingou: My only comment then would be to define somewher what "truthy" means in this case :)

pypingou:
All my questions have been answered, 👍 for me

bowlofeggs:
On Fri, 2017-02-24 at 10:09 -0800, Pierre-Yves Chibon wrote:
> My only comment then would be to define somewher what "truthy" means
> in this case :)

Truthy is a common phrase in the Python world. If you search for
"python truthy" you will find a lot of references. I think this is the
formal definition of truth in Python:

https://docs.python.org/3/library/stdtypes.html?highlight=truth#truth-v
alue-testing

PR #1309, opened 2017-02-23
Rename bodhi/tests/server/models/test_models.py.

bowlofeggs:
bodhi.server.models.models recently got renamed to
bodhi.server.models, but the tests that match the file were not
renamed. This commit renames the corresponding tests and merges in
code from the no-longer needed __init__.py.


pypingou:
👍

Issue #1304, opened 2017-02-22
The masher should clean up old mashes at the end of every run


bowlofeggs:
Bodhi 2.4.0 introduced a new CLI tool called ```bodhi-clean-old-mashes``` that deletes all but the 10 newest mashes. I had intended the script to be used in cron, but after talking with releng today about enabling it they suggested it would be nice if the masher called it automatically after each successful push to the mirrors. It's a good idea, so let's do it!

All we'll need to do is have the masher import and use the [```clean_up```](https://github.com/fedora-infra/bodhi/blob/2.4.0/bodhi/server/scripts/clean_old_mashes.py#L32) function when it completes a successful mash. I believe the print statements will be OK as they should just end up going to journald with the other log messages.

ankit01ojha:
only thing needed is to import clean_up function and add its code??

bowlofeggs:
On Fri, 2017-02-24 at 10:38 -0800, Ankit Raj Ojha wrote:
> only thing needed is to import clean_up function and add its code??

We will also need to test the change.

bowlofeggs:
When we do this, let's add a new state to the ```ComposeState``` model for "Cleaning old composes" (please do use the word compose and not mash). This way the /composes/ UI, API, and CLI will be able to display this new state to the users. It'll be rad.

bowlofeggs:
You will need to create a database migration. Here's another migration that extends an enum that you can use as an example:

bodhi/server/migrations/versions/22858ba91115_add_new_side_tag_update_states.py

vismay-golwala:
Do we clean_up if a mash fails?

bowlofeggs:
On 06/14/2018 10:05 AM, Vismay Golwala wrote:
> Do we clean_up if a mash fails?

No, we should only clean up when all the composes succeeded.

bowlofeggs:
This feature is planned to be included in the upcoming 3.9.0 release: #2477.

dustymabe:
do we know what the policy is for cleaning up old composes? i.e. is there some amount of time where they won't get cleaned up. Since we now produce atomic host artifacts in the bodhi runs we might need them to be around temporarily.

bowlofeggs:
On 08/22/2018 09:40 PM, Dusty Mabe wrote:
> do we know what the policy is for cleaning up old composes?

I don't know of any policy around this, but Bodhi will keep 10 old
composes around when it does its clean up run.

dustymabe:
[...]

PR #1312, opened 2017-02-25
masher should clean up old mashes at the end of every run

ankit01ojha:
fixes #https://github.com/fedora-infra/bodhi/issues/1304

### DIFF
--- a/bodhi/server/consumers/masher.py
+++ b/bodhi/server/consumers/masher.py
@@ -43,6 +43,7 @@ from bodhi.server.metadata import ExtendedMetadata
 from bodhi.server.models import (Update, UpdateRequest, UpdateType, Release,
                                  UpdateStatus, ReleaseState, Base)
 from bodhi.server.util import sorted_updates, sanity_check_repodata, transactional_session_maker
+from bodhi.server.scripts.clean_old_mashes import clean_up
 
 
 def checkpoint(method):
@@ -365,6 +366,7 @@ class MasherThread(threading.Thread):
 
             self.check_all_karma_thresholds()
             self.obsolete_older_updates()
+            self.clean_up()
 
         except:
             self.log.exception('Exception in MasherThread(%s)' % self.id)
@@ -973,6 +975,9 @@ class MasherThread(threading.Thread):
             raise ValueError("Could not find %s in the config file" % key)
 
         return master_repomd % (self.release.version, arch)
+
+    def clean_up(self):
+        pass
 
 
 class MashThread(threading.Thread):
--- a/bodhi/tests/server/consumers/test_masher.py
+++ b/bodhi/tests/server/consumers/test_masher.py
@@ -155,6 +155,10 @@ class TestMasher(unittest.TestCase):
 
         set_bugtracker.assert_called_once_with()
 
+    @mock.patch('bodhi.server.scripts.clean_old_mashes.NUM_TO_KEEP', 2)
+    def test_clean_up(self, clean_up):
+        self.assertEquals(len(clean_up.call_args_list), 1)
+
     @mock.patch('bodhi.server.notifications.publish')
     def test_invalid_signature(self, publish):
         """Make sure the masher ignores messages that aren't signed with the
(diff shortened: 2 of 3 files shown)


ankit01ojha:
@bowlofeggs Can you please review this.

bodhi/server/consumers/masher.py
```diff
@@ -979,3 +981,41 @@ def run(self):
...
+    for dirs in pattern_matched_dirs.values():
+        if len(dirs) > NUM_TO_KEEP:
+            dirs_to_delete.extend(sorted(dirs, reverse=True)[NUM_TO_KEEP:])
+
+    if dirs_to_delete:
+        print 'Deleting the following directories:'
+        for d in dirs_to_delete:
+            d = os.path.join(mash_dir, d)
+            shutil.rmtree(d)
+            print d
```
bowlofeggs: This code is duplicated with the script. We don't want to copy code as that violates the DRY Principle:

https://en.wikipedia.org/wiki/Don't_repeat_yourself

Instead, we want to import the existing code that does this and use it from the masher.

bodhi/tests/server/consumers/test_masher.py
```diff
@@ -1231,3 +1235,69 @@ def test_comps_unsafe_http_url(self, mock_cmd):
...
+            actual_files = [f for f in os.listdir(mash_dir)
+                            if os.path.isfile(os.path.join(mash_dir, f))]
+            self.assertEqual(actual_files, ['COOL_FILE.txt'])
+            # Make sure the printed output is correct
+            expected_output = set(dirs) - expected_dirs
+            expected_output = {os.path.join(mash_dir, d) for d in expected_output}
+            expected_output = expected_output | {'Deleting the following directories:', ''}
+            self.assertEqual(set(result.output.split('\n')), expected_output)
+        finally:
+            shutil.rmtree(mash_dir)
```
bowlofeggs: This test is copied from the existing test. We don't need to retest that this code works, but we do need to assert that the masher calls it when done.

bowlofeggs (changes requested):
I recommend adjusting the commit message to say "mashes" instead of "mashers". Also, the commit message has two #'s in front of the issue number instead of one.

ankit01ojha:
@bowlofeggs I don't know why but the clean_up function is not getting called, I am getting the test failed `AssertionError: 0 != 1`.

bowlofeggs:
@ankit01ojha it's not getting called because nothing is explicitly calling it. Python doesn't run functions when they are defined; it runs them when they are called, i.e., you put ```()```'s after its name.

bodhi/server/consumers/masher.py
```diff
@@ -979,3 +980,7 @@ def run(self):
         else:
             self.success = True
         return out, err, returncode
+
+
+def clean_up():
+    pass
```
bowlofeggs: I dont' think we need a new function here. We just need to call ```clean_old_mashes()``` when the mash is done and successful.

bowlofeggs:

ankit01ojha:
done!

bodhi/server/consumers/masher.py
```diff
@@ -974,6 +976,9 @@ def _get_master_repomd_url(self, arch):
 
         return master_repomd % (self.release.version, arch)
 
+    def clean_up(self):
+        pass
```
bowlofeggs, line 980: This method doesn't do anything. Why put it here?

bodhi/tests/server/consumers/test_masher.py
```diff
@@ -155,6 +155,10 @@ def test___init___sets_bugtracker(self, set_bugtracker):
 
         set_bugtracker.assert_called_once_with()
 
+    @mock.patch('bodhi.server.scripts.clean_old_mashes.NUM_TO_KEEP', 2)
+    def test_clean_up(self, clean_up):
+        self.assertEquals(len(clean_up.call_args_list), 1)
```
bowlofeggs, line 160: This test doesn't run the code and doesn't assert that the clean up call was made.

bowlofeggs (changes requested):
This can

PR #1316, opened 2017-03-01
Make the new Bodhi comment be a constant and fix a typo in it.

bowlofeggs:
This commit allows the test to import the default comment that
Bodhi uses when it disables autokarma, rather than duplicate the
default. It also corrects a typo in the comment.


trishnaguha:
Thanks for the fix @bowlofeggs :hamburger: .

cverna:
👍

puiterwijk (approved):
🔢

PR #1333, opened 2017-03-06
Added about bodhi and some related links

ankit01ojha:
Added what is bodhi, its advantages and added some related links.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -7,6 +7,36 @@
 Welcome to bodhi's documentation!
 =================================
 
+Bodhi is designed to democratize the package update testing and release process for RPM based Linux
+distributions. It provides an interface for developers to propose updates to a distribution, and an
+interface for testers to leave feedback about updates through a +1/-1 karma system..
+
+Features of bodhi
+-------------------
+
+* Provides an interface for developers and release engineers to manage pushing out 
+  package updates for multiple version releases.
+* Helps in delivering quality packages and repository sustainment with automated testing.
+* Gives community testers an interface to leave feedback about package updates.
+* Provides a framework that will allow future integrations to various other QA and 
+  developer tools.
+* Announces the arrival of new packages entering the collection.
+* Bodhi publishes end-user release notes known as errata.
+* Generates yum repositories.
+* Queries ResultsDB for automated test results and displays them on updates.
+* Provides pre-release test updates for end user and QA folks/users who wish to test these updates.
+
+Links
+^^^^^
+
+* `Online documentation <http://bodhi.readthedocs.io>`_
+*  IRC: #bodhi on Freenode
+* `Mailing list <https://lists.fedoraproject.org/archives/list/bodhi@lists.fedorahosted.org/>`_
+* `Report an issue <https://github.com/fedora-infra/bodhi/issues/new>`_
+* `Source code <https://github.com/fedora-infra/bodhi>`_
+* `Fedora's production instance <https://bodhi.stg.fedoraproject.org/>`_
+* `Fedora's staging instance <https://bodhi.stg.fedoraproject.org/>`_
+
 Contents:
 
 .. toctree::


ankit01ojha:
@bowlofeggs Is this fine, or anything else need to be added. Should I remove the last part , `contents` and `tables and indices`.

docs/index.rst
```diff
@@ -7,6 +7,39 @@
 Welcome to bodhi's documentation!
 =================================
 
+What is Bodhi?
+--------------
```
bowlofeggs: I think we can drop this heading and just let the paragraph below stand on its own.

docs/index.rst
```diff
@@ -7,6 +7,39 @@
...
+Bodhi is a modular web-based system that facilitates the process of publishing package updates for Fedora. Bodhi is the application that manages updates for Fedora releases. It maintains a single stage of repositories by adding/updating/removing packages. It can be used to create or modify updates and overrides.
```
bowlofeggs: When you say that Bodhi is modular, what do you mean? I haven't thought of it as being modular before.

The first and second sentence seem a little redundant.

What does it mean that Bodhi maintains a single stage of repositories?

I think we can go for something a little simpler here. Perhaps something like "Bodhi is designed to democratize the package update testing and release process for RPM based Linux distributions. It provides an interface for developers to propose updates to a distribution, and an interface for testers to leave feedback about updates through a +1/-1 karma system."

Also, let's try to keep line lengths no longer than 100 columns.

docs/index.rst
```diff
@@ -7,6 +7,39 @@
...
+
+Advantages of bodhi
+-------------------
+
+1.  Provides an intuitive interface for developers and release engineers to manage pushing out package updates for multiple version releases.
+2.  Helps in delivering quality packages and repository sustainment with automated testing.
```
bowlofeggs: What do you mean by sustainment?

docs/index.rst
```diff
@@ -7,6 +7,39 @@
...
+3.  Helps Community testers to get involved in testing of the package updates and provide quality feedback which would reach the packagers/developers.
```
bowlofeggs: I'd probably reword the above a bit. Something like:

"Gives community testers an interface to leave feedback about package updates."

docs/index.rst
```diff
@@ -7,6 +7,39 @@
...
 =================================
 
+What is Bodhi?
+--------------
+Bodhi is a modular web-based system that facilitates the process of publishing package updates for Fedora. Bodhi is the application that manages updates for Fedora releases. It maintains a single stage of repositories by adding/updating/removing packages. It can be used to create or modify updates and overrides.
+
+Advantages of bodhi
+-------------------
+
+1.  Provides an intuitive interface for developers and release engineers to manage pushing out package updates for multiple version releases.
```
bowlofeggs: I don't think we need to say it's intuitive - keeping the prose short and to the point makes it more readable. Keep in mind that many people reading the home page of a project are going to be quickly skimming it to find out what it is, so we want to be as direct as possible. We don't need to try to convince them to use Bodhi, we just need to inform them what Bodhi is.

docs/index.rst
```diff
@@ -7,6 +7,39 @@
...
+2.  Helps in delivering quality packages and repository sustainment with automated testing.
+3.  Helps Community testers to get involved in testing of the package updates and provide quality feedback which would reach the packagers/developers.
+4.  Provides a modular framework that will allow future integrations to various other QA and developer tools.
```
bowlofeggs: I'm not sure that Bodhi is modular.

docs/index.rst
```diff
@@ -7,6 +7,39 @@
...
+5.  Announces the arrival of New packages entering the collection.
```
bowlofeggs: s/New/new/

docs/index.rst
```diff
@@ -7,6 +7,39 @@
...
+6.  The reasons for an update can be communicated to end users.
```
bowlofeggs: "Bodhi publishes end-user release notes known as errata."

docs/index.rst
```diff
@@ -7,6 +7,39 @@
...
+7.  Generates extended update metadata which is parsed client-side by yum.
```
bowlofeggs: I'd probably word this more generally. Something like "Generates yum repositories."

docs/index.rst
```diff
@@ -7,6 +7,39 @@
...
+8.  Allows security updates to be seen by users as a separate channel, with links to the security advisories.
```
bowlofeggs: What do you mean by separate channel? The RSS feed? That security updates can be searched in Bodhi?

docs/index.rst
```diff
@@ -7,6 +7,39 @@
...
+9.  Accomodates automated testing for updates before the latter are released.
```
bowlofeggs: Bodhi doesn't actually do any automated testing - it displays automated test results that it retrieves from ResultsDB. I'd probably reword the above:

"Queries ResultsDB for automated test results and displays them on updates."

docs/index.rst
```diff
@@ -7,6 +7,39 @@
...
+10. Provides pre-release test updates for end users and QA folks who wish to test these updates.
```
bowlofeggs: I'd probably s/end users and QA folks/users/.

docs/index.rst
```diff
@@ -7,6 +7,39 @@
...
+
+Create a Bodhi development environment
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+For getting a Bodhi development environment you can refer `here <https://pypi.python.org/pypi/bodhi-client/2.3.3>`_
```
bowlofeggs: We already have this documented in the README and in this Sphinx project. I don't think the docs home page should repeat this information.

docs/index.rst
```diff
@@ -7,6 +7,39 @@
...
+
+
+Bodhi Deployment
+^^^^^^^^^^^^^^^^
+You can checkout `bodhi's deployment <https://bodhi.stg.fedoraproject.org/>`_
```
bowlofeggs: This is the staging deployment, not the real deployment.

docs/index.rst
```diff
@@ -7,6 +7,39 @@
 Welcome to bodhi's documentation!
 =================================
 
+What is Bodhi?
+--------------
+Bodhi is a modular web-based system that facilitates the process of publishing package updates for Fedora. Bodhi is the application that manages updates for Fedora releases. It maintains a single stage of repositories by adding/updating/removing packages. It can be used to create or modify updates and overrides.
+
+Advantages of bodhi
+-------------------
```
bowlofeggs: I would probably rename this section "Features". Also, I recommend using bullet points instead of numbers below.

docs/index.rst
```diff
@@ -7,6 +7,39 @@
...
+
+
+Bodhi Deployment
+^^^^^^^^^^^^^^^^
+You can checkout `bodhi's deployment <https://bodhi.stg.fedoraproject.org/>`_
+
+
+Reporting Issues
+^^^^^^^^^^^^^^^^
+For `reporting issues <https://github.com/fedora-infra/bodhi/issues>`_ , Click on the New Issue button and create a issue.
```
bowlofeggs: How about we just make a "Links" section with common links? Something like this:

```
Links
^^^^^

* `Online documentation <http://bodhi.readthedocs.io>`_
* IRC: #bodhi on Freenode
* `Mailing list <https://lists.fedoraproject.org/archives/list/bodhi@lists.fedorahosted.org/>`_
* `Report an issue <https://github.com/fedora-infra/bodhi/issues/new>`_
* `Source code <https://github.com/fedora-infra/bodhi>`_
* `Fedora's production instance <https://bodhi.stg.fedoraproject.org/>`_
* `Fedora's staging instance <https://bodhi.stg.fedoraproject.org/>`_
```

bowlofeggs (changes requested):

docs/index.rst
ankit01ojha: Its like it supports the quality packages with automated testing.

ankit01ojha:
@bowlofeggs Is this fine now?

bowlofeggs (approved):
I plan to fix a few things in here, but I'll just do it in a follow up commit so we can get this merged. Thanks!

PR #1334, opened 2017-03-06
Inject the repository root into the Python path for the docs build

jeremycline:
This lets Sphinx find and autodoc the Bodhi code.

While I was in the area, I switched Bodhi's theme to Alabaster since it's pretty (and the default for Sphinx now).

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -92,7 +92,7 @@ pygments_style = 'sphinx'
 
 # The theme to use for HTML and HTML Help pages.  See the documentation for
 # a list of builtin themes.
-html_theme = 'default'
+html_theme = 'alabaster'
 
 # Theme options are theme-specific and customize the look and feel of a theme
 # further.  For a list of options available for each theme, see the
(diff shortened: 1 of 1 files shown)


docs/conf.py
```diff
@@ -11,12 +11,14 @@
...
 import datetime
+import os
+import sys
 
 
 # If extensions (or modules to document with autodoc) are in another directory,
 # add these directories to sys.path here. If the directory is relative to the
 # documentation root, use os.path.abspath to make it absolute, like shown here.
-# sys.path.insert(0, os.path.abspath('.'))
+sys.path.insert(0, os.path.abspath('../'))
```
bowlofeggs: I believe this will depend on the CWD, which is probably not what we want. I think we probably want this instead:

```
sys.path.insert(0, os.path.abspath(os.path.join(os.path.dirname(__file__), '../')))
```

bowlofeggs (changes requested):

docs/conf.py
jeremycline: Good call, fixed

bowlofeggs (approved):
LGTM

PR #1341, opened 2017-03-09
Docs home page fixes

bowlofeggs:
This pull request has some commits to clean up and improve wording on the docs home page.

The docs URL I used in this pull request isn't live yet but it will be live when this commit makes it to production. You can visit https://bodhi.stg.fedoraproject.org/docs/ to see the docs today. I considered making the URL in the docs use staging until production is live, but I decided to go ahead and put the real URL there now.


puiterwijk (approved):
🐳

PR #1354, opened 2017-03-16
Ensure Karma thresholds can be set when Autopush is disabled

### LINKED ISSUE #1033: Disabling autokarma (still) removes the karma thresholds
Whenever I edit an update to disable autokarma (I have to edit it, as opposed to just unchecking the box from the beginning, due to issue #1032), the thresholds vanish. The only way to get the thresholds back is to enable autokarma. As a result, the improvement from issue #772 (manual push to stable based on karma) does not work: if I disable autokarma, it does not work because the thresholds are deleted (which is clearly a bug); if I enable autokarma, it does not work because the button explicitly gets disabled if autokarma is enabled for some reason. (Why? It can be valid to trigger a manual push based on karma even if autokarma is enabled, because, if the update already had a karma of +1 and I set the autokarma threshold to +1, the automatic push will not trigger.)

trishnaguha:
Signed-off-by: Trishna Guha <trishnaguha17@gmail.com>
Fixes #1033 .

### DIFF
--- a/bodhi/server/models.py
+++ b/bodhi/server/models.py
@@ -1008,14 +1008,6 @@ class Update(Base):
         if req is not None:
             up.set_request(db, req, request.user.name)
 
-        if not data['autokarma']:
-            data['stable_karma'] = None
-            data['unstable_karma'] = None
-            caveats.append({
-                'name': 'autokarma',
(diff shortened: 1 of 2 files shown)


bowlofeggs (approved):
Looks good, thanks @trishnaguha!

PR #1355, opened 2017-03-17
Correct the flake8 test docblock to say that all code is tested.

bowlofeggs:


pypingou:
👍

PR #1356, opened 2017-03-17
Introduce a new PEP-257 test and fix one module to comply.

bowlofeggs:
This commit introduces a new style test to check for PEP-257
compliance and converts bodhi.server.scripts.approve_testing to
comply.


pypingou:
Cool 👍

PR #1358, opened 2017-03-20
Fix a PEP-8 error that was caught on Rawhide.

bowlofeggs:


pypingou:
👍

PR #1359, opened 2017-03-20
Build docs without install

bowlofeggs:


bodhi/server/bugs.py
```diff
@@ -175,9 +177,14 @@ def modified(self, bug_id):
...
-    log.info('Using the FakeBugTracker')
-    bugtracker = FakeBugTracker()
+def set_bugtracker():
+    """
+    Set the module-level bugtracker attribute to the correct bugtracker, based on the config.
+    """
+    global bugtracker
+    if config.get('bugtracker') == 'bugzilla':
+        log.info('Using python-bugzilla')
+        bugtracker = Bugzilla()
```
pypingou, line 187: Should we check if it has been instantiated before?

bodhi/server/consumers/updates.py
```diff
@@ -36,7 +36,7 @@
 
 import fedmsg.consumers
 
-from bodhi.server.bugs import bugtracker
+from bodhi.server import bugs as bug_module
```
pypingou, line 39: For the other modules you just import bugs, do we want to be consistent there?

pypingou:
Couple of question but nothing blocking for me, 👍

bodhi/server/bugs.py
bowlofeggs: @pypingou We could do that, but I don't believe there would be harm in replacing it. If that happened, the old one would get garbage collected and things would keep on working.

bodhi/server/consumers/updates.py
bowlofeggs: @pypingou I did this because there are already variables used in this module called ```bugs```, and it was easier to import this under a different name than to rename all those variables. The variables actually have a more natural usage anyway (such as ```for bug in bugs```) and the bugs module isn't so natural (it doesn't actually contain bugs, it just has a class that can help you query bugzilla.)

PR #1363, opened 2017-03-21
Update the contributor guidelines.

bowlofeggs:
This commit updates the contributor guidelines to mention the new
PEP-257 test, to remove wording that said that PEP-8 was only
partially enforced, and to use the word "must" instead of "should".


jeremycline (approved):
Yes! :shipit:

PR #1364, opened 2017-03-21
Require test coverage to be at least 83%.

bowlofeggs:
The PR I wrote to have the masher support alternative architectures bumped the test coverage up by one more percent!


jeremycline (approved):
Looking good :shipit:

Issue #1362, opened 2017-03-21
The shield and flame icons could use hover text


bowlofeggs:
In the new Bodhi design, shield and flame icons can appear next to update titles, such as this kernel update:

https://bodhi.stg.fedoraproject.org/updates/FEDORA-2017-e8e717e2b7

It would be good for us to add mouse hover text explaining what they mean. The shield means it's a security update and the flame means that it's a critical path update. There are other icons for the type of update as well: a bug for bugs, an archive for new packages, and a bolt for enhancement updates.

PR #1365, opened 2017-03-21
Add titles to the header icons in the Update page

ryanlerch:
Previously, the icons at the top of the Update page that provide glancable context for the update type and critial path status did not have titles. Consequently, if the user did not know what they meant, they were a little cryptic.

This change adds a title to each that explains the icon when the user hovers over them.

bodhitopicons2
bodhitopicons1

Fixes: #1362

### DIFF
--- a/bodhi/server/templates/update.html
+++ b/bodhi/server/templates/update.html
@@ -410,7 +410,7 @@ $(document).ready(function(){
         ${self.util.type2icon(update.type) | n}
       % endif
       % if update.critpath:
-      <span class="label label-danger"> <i class="fa fa-fire"></i></span>
+      <span data-toggle="tooltip" title="This update is in the critical path" class="label label-danger"> <i class="fa fa-fire"></i></span>
       % endif
 
       % if can_edit:
--- a/bodhi/server/util.py
+++ b/bodhi/server/util.py
@@ -437,7 +437,9 @@ def type2icon(context, kind):
         'enhancement': 'fa-bolt',
     }.get(kind)
 
-    return "<span class='label label-%s'><i class='fa %s'></i></span>" % (cls, fontawesome)
+    return "<span class='label label-%s' data-toggle='tooltip'\
+            title='This is a %s update'><i class='fa %s'></i></span> \
+            " % (cls, kind, fontawesome)
 
 
 def severity2html(context, severity):
--- a/docs/release_notes.rst
+++ b/docs/release_notes.rst
@@ -1,5 +1,21 @@
 Release notes
 =============
+develop
+-------
+
+Features
+^^^^^^^^
+
+
+Bugs
+^^^^
+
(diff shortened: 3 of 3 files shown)


pypingou:
wfm 👍 :)

voxik:
Actually, it might be interesting to add these icons as part of the badges on the right side, so there could be seen relation ....

bodhi/server/util.py
```diff
@@ -437,7 +437,7 @@ def type2icon(context, kind):
         'enhancement': 'fa-bolt',
     }.get(kind)
 
-    return "<span class='label label-%s'><i class='fa %s'></i></span>" % (cls, fontawesome)
+    return "<span class='label label-%s' data-toggle='tooltip' title='%s update'><i class='fa %s'></i></span>" % (cls, kind, fontawesome)
```
bowlofeggs: Since the critical path one is more of a complete sentence, I suggest either making this one also be a complete sentence, or making the critical path one just say "critical path". For example, this title bit could be:

```
title='This is a %s update'
```

bodhi/server/templates/update.html
```diff
@@ -376,7 +376,7 @@
         ${self.util.type2icon(update.type) | n}
       % endif
       % if update.critpath:
-      <span class="label label-danger"> <i class="fa fa-fire"></i></span>
+      <span data-toggle="tooltip" title="This update is in the Critical Path" class="label label-danger"> <i class="fa fa-fire"></i></span>
```
bowlofeggs: Since "bugfix" will be lowercased on the other tooltip, should we lowercase "critical path"?

bowlofeggs (commented):
I've got a couple of thoughts. What do you think?

bodhi/server/util.py
bowlofeggs: This pull request also fails an automated style checking test that is asserting that no code is more than 100 columns wide. It is upset about this line:

```
/home/vagrant/bodhi/bodhi/server/util.py:440:101: E501 line too long (137 > 100 characters)
F.
AssertionError: 1 != 0
```

bowlofeggs:

ryanlerch:
Updated the strings to match cases and make them both sentences. Also fixed the extra long line.

ryanlerch:
Also updated the release notes.

bowlofeggs (changes requested):
Hey @ryanlerch, this is great! Only one more request from me: This change probably won't go into 2.5.0, so can you rebase your branch on the develop branch and move your release note to the "develop" release in the notes?

https://github.com/fedora-infra/bodhi/blob/64a76a00140da74069290e1a2db358bfd6e706e3/docs/release_notes.rst#develop

ryanlerch:
@bowlofeggs done and done.

bowlofeggs (approved):

PR #1366, opened 2017-03-21
Restyle the footer

ryanlerch:
Previously, the footer use code tags to show the bodhi version name and hostname. This did not look so good with the new dark footer, so this change makes them bold.

Also, removed the old glyphicons that were not showing any more and re-organized the content a little.

### Before:
footer-before

### After:
footer-after


bowlofeggs (approved):
Beautiful, this is much better.

PR #1368, opened 2017-03-22
Ignore comments that don't carry karma when counting karma.

### LINKED ISSUE #829: karma not counted correctly
I believe the Karma for the following bodhi update should be +3 rather than +1. Several others (people who understand how karma counting work) agree. Can you please have a look and determine if there is a bug here:

https://bodhi.fedoraproject.org/updates/FEDORA-EPEL-2016-5d93f1948c

bowlofeggs:
fixes #829


jeremycline (approved):
Makes perfect sense, :shipit:

Issue #957, opened 2016-09-22
Page title for multiple updates can be quite large


ryanlerch:
On updates that have many packages, like this GNOME one for example -- https://bodhi.fedoraproject.org/updates/FEDORA-2016-78efe5595f -- the &lt;title&gt; of the page contains all the package names.

While this might work for updates with just one package, for updates with many, it is unweildy and not useful at all. On a tab in your browser, you only see the name of the first package in the set, which may not be useful to describe the nature of the update.

genodeftest:
On another page, FEDORA-2016-77ffa78abb, it seems to make [firefox run amok](https://bugzilla.redhat.com/show_bug.cgi?id=1378601) and [Epiphany is having a hard time too](https://bugs.webkit.org/show_bug.cgi?id=162465). Both issues are probably related to the title being so long.

Issue #1351, opened 2017-03-16
Updates with lots of Packages look unweildy


ryanlerch:
The recent changes I made to bodhi on the updates page have not translated well for an update with a lot of builds attached to it. The title of the update includes all the packagenames of all the builds, resulting in:

![bodhi-large-update](https://cloud.githubusercontent.com/assets/592259/23982983/f211f908-0a5c-11e7-855c-2f0d7b840898.png)

ryanlerch:
My proposal to fix this (and also #957) would be a two-pronged solution.

First, make the logic for this title a little saner. For an update with a single build, it behaves as normal:

    enhancement update for inkscape in Fedora 25

For an update with 2 builds:

    bugfix update for python-arrow & python-requests in Fedora 24

For an update with 3 or more builds:

    security update for firefox, epiphany & 23 more in Fedora 25

The second prong would be to provide a new field tied to an update allowing a package maintainer to add a short description to a multi-build update. With it falling back to the above logic if one is not given. In the example given in the inital report, the short description could simply be "GNOME", so the label would read:

    enhancement update for GNOME in Fedora 25

bowlofeggs:
@ryanlerch I'm a +1 on your proposal. I suggest we do it in two separate issues/PRs since one is easier than the other.

ryanlerch:
Okay!

I can do the first prong, but the second involves adding some database entries, so it's a little out of my skill range. Will get the PR done for the first one now.

ryanlerch:
Filed #1369 for the request of the new short description field for an update

bowlofeggs:
This was fixed by #1370.

PR #1370, opened 2017-03-22
make the html title and update description shorter for large updates

ryanlerch:
This adds the logic described in #1351 to format and truncate the html title and the subheader line if of updates with more than one build.

Some screenshots:

### 1 Build
1-build

### 2 Builds
2-builds

### 3 Builds
3-builds

### 5 Builds
5-builds

### 53 Builds
53-builds

Fixes: #957 #1351


bowlofeggs (approved):
Thanks!

PR #1373, opened 2017-03-22
Fix #1172. Show about 25 points on the X-axis.

### LINKED ISSUE #1172: The releases graph is undiscernable for releases that have a long history, like EPEL 6
The EPEL 6 releases graph has far too many dates on the X-axis to be useful:

https://bodhi.fedoraproject.org/releases/EL-6

We probably need some code to decide how many dates to include there so it doesn't get too crazy. I'm going to assume this is HTML/JS/CSS work and that it's easy, but I could be wrong.

nbianca:
Fixes #1172.

The ideal way would be using a newer version of Chart.js that has this functionality built-in. I did not want to break anything.

Basically, this fix filters data depending to show about 25 points. I chose the number of 25 empirically.

### DIFF
--- a/bodhi/server/templates/release.html
+++ b/bodhi/server/templates/release.html
@@ -26,7 +26,15 @@
             % endfor
           ]
         }
-        var ctx = document.getElementById("releasechart").getContext("2d");
+        // About 25 labels should be printed at most.
+        let every_nth = Math.max(1, Math.round(data.labels.length / 25));
+        let every_filter = function(element, index) {
+            return index % every_nth == 0;
+        }
+        data.labels = data.labels.filter(every_filter);
+        for (let dataset of data.datasets) {
+            dataset.data = dataset.data.filter(every_filter);
+        }
         var myNewChart = new Chart(ctx).Line(data);
       </script>
       <table class="table">


bodhi/server/templates/release.html
```diff
@@ -26,7 +26,15 @@ <h2 class="pull-left">${release['long_name']}</h2>
             % endfor
           ]
         }
-        var ctx = document.getElementById("releasechart").getContext("2d");
```
bowlofeggs, line 29: I think this line is necessary, since ctx is used below on line 38. Also, without this line the chart is not displayed when I view http://localhost:6543/releases/EL-6 but with it the chart does appear.

bodhi/server/templates/release.html
```diff
@@ -26,7 +26,15 @@ <h2 class="pull-left">${release['long_name']}</h2>
...
+        // About 25 labels should be printed at most.
+        let every_nth = Math.max(1, Math.round(data.labels.length / 25));
+        let every_filter = function(element, index) {
+            return index % every_nth == 0;
+        }
+        data.labels = data.labels.filter(every_filter);
+        for (let dataset of data.datasets) {
+            dataset.data = dataset.data.filter(every_filter);
+        }
```
bowlofeggs, line 37: This looks reasonable, and it seems to work when line 29 is added back.

bowlofeggs (changes requested):

bowlofeggs:

nbianca:
For some reason, I cannot push any new changes to this pull request. I will close and open another one.

PR #1374, opened 2017-03-22
Fix #997. Makes search case-insensitive.

### LINKED ISSUE #997: search should be case-insensitive

nbianca:
Fixes #997.

### DIFF
--- a/bodhi/server/services/updates.py
+++ b/bodhi/server/services/updates.py
@@ -217,7 +217,7 @@ def query_updates(request):
     packages = data.get('packages')
     if packages is not None:
         query = query.join(Update.builds).join(Build.package)
-        query = query.filter(or_(*[Package.name == pkg for pkg in packages]))
+        query = query.filter(or_(*[Package.name.lower() == pkg.lower() for pkg in packages]))
 
     package = None
     if packages and len(packages):


bodhi/server/services/updates.py
```diff
@@ -217,7 +217,7 @@ def query_updates(request):
     packages = data.get('packages')
     if packages is not None:
         query = query.join(Update.builds).join(Build.package)
-        query = query.filter(or_(*[Package.name == pkg for pkg in packages]))
+        query = query.filter(or_(*[Package.name.lower() == pkg.lower() for pkg in packages]))
```
bowlofeggs, line 220: This causes an error:

```

500 Server Error

Neither 'InstrumentedAttribute' object nor 'Comparator' object associated with Package.name has an attribute 'lower'
```

I've also done a little more looking into this since I wrote my comment on the ticket that I thought this was the line to change, and I no longer think this is the only line to change. I'll add another comment to the issue with my findings.

bowlofeggs (changes requested):

bowlofeggs:

PR #1375, opened 2017-03-22
Fix #1161. Use alias instead of title in URLs.

### LINKED ISSUE #1161: [web] cannot edit large bodhi update: should use hash alias
I can't edit [this large fedora update](https://bodhi.fedoraproject.org/updates/FEDORA-EPEL-2016-25cbafa070) in the Fedora Bodhi web interface...
when I try I get an error page saying the URL is invalid:

```
Not Found

The requested URL /updates/nvr1 nvr2 nvr3 .... nvr1XYZ/edit was not found on this server.
```

nbianca:
Fixes #1161.


bowlofeggs (approved):
This seems to work, and seems to be using the aliases instead of the names. Nice!

Issue #951, opened 2016-09-21
 Bodhi web UI unusable for large updates after 2.2.0 update


kalev:
After latest Bodhi update, the sheer number of taskatron test results seems to cause most web browsers to hang. Try any larger update, such as https://bodhi.fedoraproject.org/updates/FEDORA-2016-78efe5595f -- Firefox locks up here after clicking on this link and Chrome as well.

frieben:
Firefox fills all available real memory and swap space after opening https://bodhi.fedoraproject.org/updates/FEDORA-2016-78efe5595f. I killed the process at 8 GB and counting .. .

bowlofeggs:
Thanks for the bug report! A quick fix would probably be to limit the number of taskotron results that are rendered in the template to some arbitrary but small-enough number. I think we could get a fix like that in soon so that the pages at least render.

Something perhaps a little more sohpisticated would be to sort the tasks so that the failing ones sort highest, limit them to some arbitary number, and if there are more than our limit we can just link to the full taskotron output (something like this https://taskotron.fedoraproject.org/artifacts/all/8ac4e78e-7e6a-11e6-bab6-525400120b80/task_output/ ?)

What do you think?

kalev:
Sure, makes sense to me to limit the results for now, thanks!

@kparal, any comments about sorting taskotron tasks output?

kparal:
There are multiple issues, I think. I'll try to describe them in detail, sorry for a long post.

First, tasks IDs look like `namespace.task_name.any.number.of.subresults`. We currently have just a single important namespace, `dist`. So some of results are reported simply as `dist.depcheck`. Other tasks, like rpmgrill, decided to report an overall result as `dist.rpmgrill`, but also individual subresults as `dist.rpmgrill.subresult` (see [here](https://taskotron.fedoraproject.org/resultsdb/testcases)). Publishing the overall result as `dist.rpmgrill` is not mandatory, some tasks might not have that (only the subresults). The tasks can simply do what they like under `namespace.task_name`. In the future the namespaces might be structured themselves, like `group.cloud`, but the list of namespaces will be available through API.

[...]

PR #1379, opened 2017-03-23
Optimize the ajax calls to results db and reorgainze the output

ryanlerch:
The current code for pulling down the data from resultsdb was being done by getting a list of all the taskotron testcases, and then pulling down the data for each package. This resulted in a *lot* of requests to pull down a small amount of data. This is somewhat related to #951 

This patch primarily changes the javascript ajax calls to use the resultsdb "results/" api endpoint which greatly reduces the number of calls to resultsdb, and speeds up the rendering of all updates significantly. 

Some quick metrics:
http://localhost:6543/updates/FEDORA-2017-9f83ba7048
Now issues 241 Requests, where it used to be > 2500 (had to stop it)

http://localhost:6543/updates/FEDORA-2017-9a6a91a4b6
Now issues 54 Requests, previously 374 Requests

http://localhost:6543/updates/FEDORA-2017-6cb36132c0
Now: 87 requests was > 2500 (stopped)

http://localhost:6543/updates/FEDORA-2017-e736c47270
now 46 requests was 162 requests

## other changes
These changes are primarily for fixing #983 I have also grouped the output in the automated results tab to be by build name, and added the "note" output for results to give a little more context:

fedora 2017 9a6a91a4b6 enhancement update for globus ftp client globus gridftp server control 1 more fedora updates system

### DIFF
--- a/bodhi/server/templates/update.html
+++ b/bodhi/server/templates/update.html
@@ -40,7 +40,7 @@ ${update.type} update for ${generateupdatetitlestring()}
 
 <script type="text/javascript">
   $(document).ready(function() {
-    var base_url = '${request.registry.settings["resultsdb_api_url"]}' + '/api/v1.0/';
+    var base_url = '${request.registry.settings["resultsdb_api_url"]}' + '/api/v2.0/';
 
     // Some handy lookups that map taskotron states to bootstrap CSS classes.
     var classes = {
@@ -74,7 +74,7 @@ ${update.type} update for ${generateupdatetitlestring()}
     // These are the required taskotron tests
     var requirements = ${update.requirements_json | n};
 
-    var make_row = function(outcome, testcase, item, arch, time, url) {
+    var make_row = function(outcome, testcase, note, arch, time, url) {
       var icon = '<span data-toggle="tooltip" data-placement="top" ' +
         'title="' + outcome + '" ' +
         'class="fa fa-' + icons[outcome] + ' text-' + classes[outcome] + '">' +
@@ -89,7 +89,8 @@ ${update.type} update for ${generateupdatetitlestring()}
       }
 
       if (arch != undefined) {
-        item = item + "<small>(" + arch + ")</small>";
+        testcase = testcase + "&nbsp;<span class='label label-default'>"
+          + arch + "</label>";
       }
 
       var age = '';
@@ -98,25 +99,31 @@ ${update.type} update for ${generateupdatetitlestring()}
         // https://github.com/fedora-infra/bodhi/issues/217
         var age = moment(time + ' Z').fromNow();
       }
-
-      return '<tr class="table-' + classes[outcome] + '" ' +
+      if (note == null){
+        note = '';
+      }
+      return '<tr class="row-automatedtests table-' + classes[outcome] + '" ' +
         'style="cursor: pointer;"' +
         'data-href="' + url + '">' +
         '<td>' + required + '</td>' +
         '<td>' + icon + '</td>' +
-        '<td>' + testcase + '</td>' +
-        '<td>' + item + '</td>' +
-        '<td>' + age + '</td>' +
+        '<td class="nowrap">' + testcase + '</td>' +
+        '<td class="stretch-table-column text-xs-right">' + note + '</td>' +
+        '<td class="nowrap">' + age + '</td>' +
         '</tr>';
     };
 
     var latest = {};
+    var activeResultsAjaxCalls = 0;
     var receive_resultsdb = function(data) {
+      if (data.next){
+        request_results(data.next);
+      }
+      activeResultsAjaxCalls--;
 
       // First, prune duplicate results.  For instance, some depcheck runs
       // happen multiple times on an update.  We only (imho) want to display
       // the latest ones for each arch.  So, prune like this:
-      var changed = false;
       $.each(data.data, function(i, result) {
 
         // However, we want to skip over any ABORTED results:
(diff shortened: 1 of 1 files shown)


pypingou:
Did you look at latest? it should basically return all the results for a build for example: https://taskotron.fedoraproject.org/resultsdb_api/api/v2.0/results/latest?type=koji_build&item=GeoIP-1.6.9-4.fc26

So ultimately that should allow to do 1 query / build in the update

ryanlerch:
@pypingou i tried to :) but latest was giving me 500 errors last night when i tried it.

does latest paginate the results? that is what was making to do multiple requests / build. 

I assume recent returns the most recent result for each textcase that is run? If so that would be awesome, so we wouldnt have to prune the results.

ryanlerch:
Just the simple change to latest knocked another 100 requests off this one:

http://localhost:6543/updates/FEDORA-2017-9f83ba7048
with /results it was 241 Requests, /results/latest knocks it down to 111.

thanks @pypingou!

pypingou:
> with /results it was 241 Requests, /results/latest knocks it down to 111.

Nice!

jskladan:
Just wanted to note, that the /latest endpoint does not really work with depcheck/upgradepath correctly, or at least it does not, if you don't query properly. For what the /latest cares things like:
`testcase: depecheck, item:foo, arch: x86_64` is the same as `testcase: depecheck, item:foo, arch: i386`, if the query is `.../latest?item=foo` (as the 'latest' filter is set to 'item'). If you wanted to have the 'proper' latest results for depcheck/upgradepath (as the others do not produce multiarch results), you'd have to do `.../latest?item=foo&arch=i386`, which would net you all the latest i386 results for item foo.

Thing is, with just arch, it is easy, if we start to have testcases where the stuff is differentiated with multiple fields, it gets even more complicated, and we (as in fedoraqa) do not believe that bodhi should know about all the possibilities. Stuff to solve this reasonably are under way, so I just wanted to advise you, that using the /latest endpoint has its problems.
If you want to keep the special cases for depecheck/upgradepath you can still do /latest query, and then do `/results/testcase=dist.depcheck&item=foo&arch=i386&limit=1` to  get that latest one for that special cases

jskladan:
One other thing, that I'd love to see is if you started using the `since` param. Since no results can by definition be newer than the time when the update was created, I suggest adding `...&since=$UPDATE_CREATED_TIME` to the queries. Where `$UPDATE_CREATED_TIME` is in the UTC time in ISO 8601 format. No need to use it with the /latest endpoint, generaly, but if you decide to not use it, please consider putting the time constraint into the queries. It will speed things up for you in the end :)

bodhi/server/templates/update.html
```diff
@@ -40,7 +40,7 @@
 
 <script type="text/javascript">
   $(document).ready(function() {
-    var base_url = '${request.registry.settings["resultsdb_api_url"]}' + '/api/v1.0/';
+    var base_url = '${request.registry.settings["resultsdb_api_url"]}' + '/api/v2.0/';
```
bowlofeggs, line 43: +1 I've been meaning to look into doing this. @ryanlerch you are my hero.

bowlofeggs, line 43: Very nice, I've been meaning to do this for a while. @ryanlerch, you are my hero!

bodhi/server/templates/update.html
```diff
@@ -98,122 +99,90 @@
...
+      if (data.next){
+        request_results(data.next);
+      }
+      activeResultsAjaxCalls--;
+      console.log(data);
       // First, prune duplicate results.  For instance, some depcheck runs
       // happen multiple times on an update.  We only (imho) want to display
       // the latest ones for each arch.  So, prune like this:
-      var changed = false;
+      //var changed = false;
```
bowlofeggs: If we don't need this var, let's remove it instead of commenting it.

bodhi/server/templates/update.html
```diff
@@ -98,122 +99,90 @@
...
 
     var finish = function() {
-      // Furthermore, remove the spinner if its still around.
+      // Furthermore, remove the spinners.
+      $("#tab-automatedtests .fa-spinner").remove();
       $('#resultsdb .spinner').remove();
 
+      // add the count of tests to the automated tests tab
+      $("#tab-automatedtests").append(" <span class='label label-default'>"
+        +$('.row-automatedtests').length+"</span>");
```
bowlofeggs: +1 this is really cool

bowlofeggs (changes requested):
It sounds like we will need to query for some results differently than others, based on other comments here.

We will also need release notes for this change in docs/release_notes.rst.

bowlofeggs:
This really does perform noticeably better, way to go!

AdamWill:
I don't have any specific comments, just a note that I'm gonna want to change this all again soon so it displays openQA results (which must be found by update ID, not NEVR) as well. I haven't had time to do that yet, though.

nirik:
It would be nice if the tab in bodhi had some kind of visual indicator that there were failures there so people would click on that tab to look. LIke a little 'failures' tab on top of the "Automated tests" tab or a "Failed: 10" after it?

kparal:
I'm going to repeat what Josef already said, but just to clarify... We've been discussing exactly this on [resultsdb-users list](https://lists.fedoraproject.org/archives/list/resultsdb-users@lists.fedoraproject.org/thread/DJSIOLA3YEH6HNSJEXVPLIZT5TW4CDED/) and will continue to discuss it, feedback welcome. Our goal is to come up with an approach that can be easily used by Bodhi (display the latest unique results, where results might not be identified just by `testcase`+`item`, but also by additional fields like `arch`), it does not require special knowledge about every available test case, and it's not overly taxing to the database (and seeing current bodhi performance, let's also mention not taxing to the user's web browser).

At this moment, there's no magic solution. Using the `/latest` endpoint is a good tradeoff for the moment, I think. Only depcheck is now arch-based, so you can do a special query for that (once for x86_64 and once for i386), but note that it seems we'll stop submitting i386 results in the near future anyway. Adjustments will be needed to hook up openqa results, but those can be tackled separately.

It's a very good idea to use `limit=1` for directed queries (`/results/?item=foobar&testcase=dist.depcheck&arch=x86_64&limit=1`), and `since` if you can (for `bodhi_update` type it would the last modification timestamp of that update, for `koji_build` type it would be the build timestamp of that build, provided you have that at hand). The generic queries should be limited to the `dist` namespace (`&testcases:like=dist.*`), because we don't have useful tests in other namespaces yet (that will change with openqa, but again, it will be confined to another specific namespace). That will limit the stress on both the database and the web browser.

I hope we will have a better and more universal approach available for you soon-ish.

Regarding UI, moving the results to a different tab will help with the page jumping up and down all the time, that's great. It might also make sense to update the HTML with a timer (only once per a few seconds, for example), to process the retrieved results in batch rather than individually. A thing to consider is to add `Show automated results` button and perform the queries only after it's clicked - that would save a lot of resources (most people don't look at it and won't click), but it would also prevent showing "there are some errors!" until intentionally clicked. The number in the tab could be colored based on the "worst" color in the results - orange or red, if some result like that is present.

ryanlerch:
Okay!

I have updated the API calls so we get the depcheck results for both arches, and added labels to the automated test tabs showing failurecounts for @nirik :)

bowlofeggs:
I really like the counts in the tab, nice!

bodhi/server/templates/update.html
```diff
@@ -98,122 +99,121 @@
...
+          // the latest/ api endpoint doesnt take into account multiple arches
+          // for dist.depcheck. So if we hit one of these for the first time
+          // we do a call to get the other one.
+          if (name == 'dist.depcheck'){
+            var theurl = base_url+"results?testcases=dist.depcheck&item="+item+"&type=koji_build&limit=1";
+            if (arch == 'x86_64'){
+              theurl = theurl +"&arch=i386"
+            } else {
+              theurl = theurl +"&arch=x86_64"
+            }
```
bowlofeggs, line 155: @kparal, @rajcze: You mentioned ```x86_64``` and ```i386``` - are there also results for the other arches, like aarch64? I'm wondering if we should avoid hardcoding arches here, or if we should at least look for results for more arches than just these two. What do you think?

bodhi/server/templates/update.html
```diff
@@ -98,122 +99,121 @@
...
+      if (data.next){
+        request_results(data.next);
+      }
+      activeResultsAjaxCalls--;
 
       // First, prune duplicate results.  For instance, some depcheck runs
       // happen multiple times on an update.  We only (imho) want to display
       // the latest ones for each arch.  So, prune like this:
-      var changed = false;
+      //var changed = false;
```
bowlofeggs: Don't forget to drop this commented code.

bodhi/server/templates/update.html
```diff
@@ -98,122 +99,121 @@
...
+          // we do a call to get the other one.
+          if (name == 'dist.depcheck'){
+            var theurl = base_url+"results?testcases=dist.depcheck&item="+item+"&type=koji_build&limit=1";
+            if (arch == 'x86_64'){
+              theurl = theurl +"&arch=i386"
+            } else {
+              theurl = theurl +"&arch=x86_64"
+            }
+            request_results(theurl);
+            console.log(theurl);
```
bowlofeggs: I think we drop this, since it's just a debug statement. FF's debug toolbar can be used to verify which URLs are being accessed if needed later.

bowlofeggs (changes requested):
The main things I care about are dropping the commented code and the console statement. Can you also squash your commits? I think it makes sense for there to be two commits - one to optimize/reorganize, and one to give the tab the counts. Or you can do one commit if you prefer. Or three - I mostly want to avoid the "rework" one in the commit log.

The arch thing is an open question.

bowlofeggs:
@kparal, @rajcze: I'd also like your feedback on the use of API calls in @ryanlerch's most recent commits since you are most familiar with resultsdb.

jskladan:
Looks good to me. Thanks!

ryanlerch:
@bowlofeggs 

* Removed the commented out variable
* removed 2 console.log()s that were used for debugging
* squashed the commits down into 1

bowlofeggs:
The current code looks OK to me, but it looks like it is making lots of requests again. For example, http://localhost:6543/updates/FEDORA-2017-9f83ba7048 makes 2,581 requests, where the first implementation made 241. Is that just how it is, given the caveats of the ```/latest/``` endpoint?

I do think it's still an improvement over the current state in terms of presentation so I'm happy to merge as-is, but it seems that the performance is still an issue with the current code. Given that, do we want to merge and iterate, or do you want to keep poking at this PR? Either way is fine by me, so I'll put the "Approved" on it and you can merge it if you please.

bowlofeggs (approved):

ryanlerch:
Yeah, that's not good!

I figured it out though -- it was the limit=1 that was causing this. I assumed (wrongly) that this would limit the results, but it just provides the number of entries per page. So we were pulling down all the depcheck results one item to a page, even though we only want the first one. The easiest way to fix this was to just remove the limit, and we are back to 245ish requests for this case.

We could probably do this more efficiently, but recursive ajax calls are doing my head in, and trying to identifiy if the call is one of the hack calls for depcheck arches was pretty complex, so happy to just not have the limit on this call for now.

bodhi/server/templates/update.html
jskladan: Not to my knowledge. Also, I'd say that you should not worry about it - we are in the process of solving this, and all the similar kinds of problems for you anyway :)

jskladan:
> I figured it out though -- it was the limit=1 that was causing this. ...

Yeah, I guess it can be counter-intuitive... 

>  The easiest way to fix this was to just remove the limit, and we are back to 245ish requests for this case <snipped> trying to identifiy if the call is one of the hack calls for depcheck arches was pretty complex, so happy to just not have the limit on this call for now.

Not that it is a _huge_ issue but you could easily do just 2 requests with minimal changes. SInce you are consuming the `/latest` endpoint, the only thing you want to do on top of that is getting that "other depcheck result". The thousands/hundreds of requests is caused by the fact, that you continue to read subsequent pages of results. You really don't need to, since `/latest` has no pagination at all, and any other request you (now) make only needs the first page(result) anyway.

So instead of removing the `limit=1` I suggest putting that back (not necessary, but nice), and removing the:
```
      if (data.next){
        request_results(data.next);
      }
```
From the `receive_resultsdb()` method. This will get you (with the code you have) to 2 requests per _any update_, and you will also not load of data you are actually not interested in.

Makes sense?

kparal:
I don't completely understand the code, but it seems to me that @rajcze is right here. Just change this line:
```
var theurl = base_url+"results?testcases=dist.depcheck&item="+item+"&type=koji_build";
```
to this:
```
var theurl = base_url+"results?testcases=dist.depcheck&item="+item+"&type=koji_build&limit=1";
```
and remove the 3 lines that @rajcze mentioned and everything should be ponies and rainbows. Since `/latest` returns everything, and for depcheck you only want a single (latest) result, there's no need to paginate at all. You'll avoid pulling the whole depcheck history for that particular build.

@ryanlerch @bowlofeggs Can you please try if it works correctly, and push if it does?

bowlofeggs:
I filed https://github.com/fedora-infra/bodhi/issues/1426 to track the suggestion in the two comments before this one.

PR #1381, opened 2017-03-23
Add docblocks to the initialize_bodhi_db script

bowlofeggs:


bodhi/tests/test_style.py
```diff
@@ -46,8 +46,9 @@ def test_code_with_pydocstyle(self):
         slowly. This test enforces only modules that have been corrected to comply with pydocstyle.
         The goal is that this test would one day check the entire codebase.
         """
-        enforced_paths = ['bodhi/server/scripts/approve_testing.py',
-                          'bodhi/server/scripts/expire_overrides.py']
+        enforced_paths = [
```
puiterwijk, line 49: Didn't we get rid of the enforced paths since everything's enforced now?

puiterwijk (approved):
You will want to revert the test_style changes, but adding docs is always good. ⚡️

bodhi/tests/test_style.py
bowlofeggs: @puiterwijk that was for the PEP-8 tests. This is a new test that will fail on most of the codebase, so I'm taking a similar approach of converting a bit here and there until it is all enforced. Thus, we do want this change.

puiterwijk: Oooh, I see now. pydocstyle :)

ACK for the PR still stands.

PR #1383, opened 2017-03-23
Fix #1172. Show about 25 points on the X-axis.

### LINKED ISSUE #1172: The releases graph is undiscernable for releases that have a long history, like EPEL 6

nbianca:
The ideal way would be using a newer version of Chart.js that has this functionality built-in. I did not want to break anything.

Basically, this fix filters data depending to show about 25 points. I chose the number of 25 empirically.

### DIFF
--- a/docs/release_notes.rst
+++ b/docs/release_notes.rst
@@ -20,6 +20,8 @@ Features
 
 Bugs
 ^^^^
+* Make the release graph easier to read when there are many datapoints
+  (`#1172 <https://github.com/fedora-infra/bodhi/issues/1172>`_).
 * The position of the Add Comment button is now the bottom right.
   (`#902 <https://github.com/fedora-infra/bodhi/issues/902>`_).
 
--- a/bodhi/server/templates/release.html
+++ b/bodhi/server/templates/release.html
@@ -26,6 +26,15 @@
             % endfor
           ]
         }
+        // About 25 labels should be printed at most.
+        let every_nth = Math.max(1, Math.round(data.labels.length / 25));
+        let every_filter = function(element, index) {
+            return index % every_nth == 0;
+        }
+        data.labels = data.labels.filter(every_filter);
+        for (let dataset of data.datasets) {
+            dataset.data = dataset.data.filter(every_filter);
+        }
         var ctx = document.getElementById("releasechart").getContext("2d");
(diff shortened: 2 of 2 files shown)


bowlofeggs:
Hello @nbianca!

In the future, please keep the original pull request open when implementing requested changes. You can amend your commits with ```git commit --amend``` and then you can force push to your branch with ```git push -f``` and the pull request will be updated with your changes. It's better to keep the pull request open so we can maintain the history of conversation in once place.

For the record, the original pull request was #1373.

docs/release_notes.rst
```diff
@@ -20,6 +20,7 @@ Features
 
 Bugs
 ^^^^
+* Make the release graph easier to read when there are many datapoints.
```
bowlofeggs: Can you add a reference to the bug number here, similar to the line below this? Don't forget to remove the period on the end of this line. Basically, this will do:

```
* Make the release graph easier to read when there are many datapoints
(`#1172 <https://github.com/fedora-infra/bodhi/issues/1172>`_).
```

bowlofeggs (changes requested):
Please alter your commit message to mention the release graph and so that "fixes #1172" appears in the body of the commit message instead of the title. Something like this will work:

```
Show about 25 points on the release graph x-axis.

fixes #1172
```

nbianca:
Done. Should be ready to merge.

As I mentioned in the other PR, GitHub would not recognize my push (if you look closely the source branch is marked as `unknown repository`).

bowlofeggs:
@nbianca The commit message still seems to be:

```
Fix #1172. Show about 25 points on the X-axis.
```

Can you use ```git commit --amend``` to change it to something like this?

```
Show about 25 points on the release graph x-axis.

fixes #1172
```

Once you've amended, you can force push it to your branch with ```git push -f```.

nbianca:
Done.

bowlofeggs (approved):
Nice!

PR #1384, opened 2017-03-24
Document the USERNAME environment variable and clarify --user.

bowlofeggs:
This commit documents the USERNAME environment variable in the
bodhi man page and clarifies that the --user flag is overloaded and
is sometimes used for authentication and other times is used as a
query parameter.

### DIFF
--- a/docs/man_pages/bodhi.rst
+++ b/docs/man_pages/bodhi.rst
@@ -39,7 +39,11 @@ Most of the commands will accept these options:
 
 ``--user <username>``
 
-    Many commands accept this flag to specify which user's updates should be operated upon.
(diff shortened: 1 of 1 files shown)


pypingou:
👍 for me, nice and clear

nbianca:
Hi,

There are many other flags that can be specified using environment variables. Shouldn't we document all of them?

bowlofeggs:
@nbianca The only other environment variable I see in the client is ```BODHI_URL``` and that one is already documented. Are there others you have found that I am missing?

nbianca:
No, sorry. It is just `BODHI_URL`. I saw `USERNAME` being used in more than one place and for some reason I thought it were different variables.

PR #1385, opened 2017-03-27
Implement bodhi overrides --mine.

### LINKED ISSUE #1382: Implement bodhi overrides --mine
Zbigniew Jędrzejewski-Szmek [filed an issue in Red Hat's Bugzilla](https://bugzilla.redhat.com/show_bug.cgi?id=1218042) requesting the bodhi client to support a ```--mine``` flag as a shortcut to express ```--user <username>```.

I.e., I could search for my overrides like this:

```
$ bodhi overrides query --mine
```

instead of:

```
$ bodhi overrides query --user bowlofeggs
```

nbianca:
Fixes #1382.

### DIFF
--- a/bodhi/client/__init__.py
+++ b/bodhi/client/__init__.py
@@ -298,6 +298,8 @@ def overrides():
 @overrides.command('query')
 @click.option('--user', default=None,
               help='Updates submitted by a specific user')
+@click.option('--mine', 'user', flag_value=os.getenv('USERNAME'),
+              help='Updates submitted by current user')
 @click.option('--staging', help='Use the staging bodhi instance',
               is_flag=True, default=False)
 @url_option
--- a/docs/man_pages/bodhi.rst
+++ b/docs/man_pages/bodhi.rst
@@ -41,6 +41,11 @@ Most of the commands will accept these options:
 
     Many commands accept this flag to specify which user's updates should be operated upon.
 
+``--mine``
+
+    Some commands accept this flag instead of the ``--user <username>`` option. This uses the
+    value from the ``USERNAME`` environment variable.
+
 ``--version``
 
     Show version and exit. Not accepted by subcommands.


bodhi/client/__init__.py
```diff
@@ -298,6 +298,8 @@ def overrides():
 @overrides.command('query')
 @click.option('--user', default=None,
               help='Updates submitted by a specific user')
+@click.option('--mine', 'user', flag_value=os.getenv('USERNAME'),
+              help='Updates submitted by current user')
```
bowlofeggs, line 302: If we add ". user can be set with USERNAME environment variable" to the help text, it'll be consistent with how the ```--url``` flag is documented.

bowlofeggs: I think we should use ```is_flag=True``` on this with ```default=False```. It seems like it isn't very useful to pass an argument to ```--mine``` since that can be done with ```--user``` already.

Also, we should add a callback/validator since it would be an error to use ```--mine``` without setting the ```USERNAME``` environment variable for now. The validator could basically just say "You must set the USERNAME environment variable to use --mine" and raise the appropriate click Exception. For an example of a validator, see ```_warn_if_url_and_staging_set()``` in this same file.

We do plan to rework how authentication works in the Bodhi CLI sometime soon, and I'm hoping we will be able to know the user's name without environment varaibles or CLI flags, but for now I think this is an OK way to do this.

docs/man_pages/bodhi.rst
```diff
@@ -41,6 +41,11 @@ Most of the commands will accept these options:
 
     Many commands accept this flag to specify which user's updates should be operated upon.
 
+``--mine``
+
+    Some commands accept this flag instead of the ``--user <username>`` option. This uses the
+    value from the ``USERNAME`` environment variable.
```
bowlofeggs, line 47: As is, only one command uses it. However, it does seem natural to add this to ```bodhi updates query --mine```. Want to add that as a flag in the code too, so this docs statement is accurate about "some commands"? ☺

bowlofeggs (changes requested):
The validator will need a couple of tests as well - those can be added to https://github.com/fedora-infra/bodhi/blob/10cd1c1e45d1845bceb4d7de7ce391777b0447da/bodhi/tests/client/test___init__.py

bowlofeggs:
I found a potentially nice way to find out the user's username. If the user has authenticated to Bodhi before with the CLI, there will be a file at ```fedora.client.openidbaseclient.b_SESSION_FILE```. This file is a JSON formatted database of authentication sessions the user has established with Fedora systems. There is some base64 encoded data in there, but we can also find out if the user ever has authenticated to Bodhi and if so, what username they used:

```
>>> import json
>>> with open("/home/rbarlow/.fedora/openidbaseclient-sessions.cache") as cache_f:
...     auth_data = cache_f.read()
>>> auth_data = json.loads(auth_data)
>>> auth_data.keys()
[u'https://bodhi.fedoraproject.org/:bowlofeggs', u'https://bodhi.stg.fedoraproject.org/:bowlofeggs']
```

Unfortunately, if the user has ever tried to authenticate as someone else an entry will also appear in there for the other username. So if there is more than one entry with the Bodhi URL and different usernames, it won't be clear which one we should use with ```--mine```, but if there's only one entry we can use this as a fallback if the ```USERNAME``` env var is not given.

To me, doing all of this extra complicated work isn't necessary to solve this issue, but I thought I'd mention it in case you just want to. If you decide to opt for just doing the env var for now, I'll just file the above idea as its own feature request.

bowlofeggs:
Also, I just talked to @puiterwijk and he is keen on getting Bodhi to use OIDC instead of OpenID. Perhaps my last comment should be ignored for now, because we might just rework all the auth anyway.

bowlofeggs:
Hello @nbianca, do you still want to work on this?

bowlofeggs:
Closing due to inactivity.

PR #1387, opened 2017-03-27
Days to stable is displayed as a negative number on updates that can be pushed

### LINKED ISSUE #1386: Days to stable is displayed as a negative number on updates that can be pushed
I noticed today that an update that was eligible to be pushed to stable yesterday shows "days to stable" as -1:

https://bodhi.stg.fedoraproject.org/updates/FEDORA-2017-d0d830d7d2

It probably makes sense to either not show this field when the number is negative, or to render it as 0.

nbianca:
Fix #1386.

### DIFF
--- a/bodhi/server/models.py
+++ b/bodhi/server/models.py
@@ -980,6 +980,20 @@ class Update(Base):
         return build
 
     @property
+    def mandatory_days_in_testing(self):
+        """
+        Calculate and return how many days an update should be in testing before becoming stable.
+
+        :return: The number of mandatory days in testing.
+        :rtype:  int
+        """
+        if self.critpath:
+            return int(config.get('critpath.stable_after_days_without_negative_karma'))
+
+        days = self.release.mandatory_days_in_testing
+        return days if days else 0
+
+    @property
     def karma(self):
         """
         Calculate and return the karma for the Update.
@@ -1481,7 +1495,7 @@ class Update(Base):
             else:
                 # If we haven't met the stable karma requirements, check if it
                 # has met the mandatory time-in-testing requirements
-                if self.release.mandatory_days_in_testing:
+                if self.mandatory_days_in_testing:
                     if not self.met_testing_requirements and \
                        not self.meets_testing_requirements:
                         if self.release.id_prefix == "FEDORA-EPEL":
@@ -2127,6 +2141,8 @@ class Update(Base):
         If this release does not have a mandatory testing requirement, then
         simply return True.
         """
+        num_days = self.mandatory_days_in_testing
+
         if self.critpath:
             # Ensure there is no negative karma. We're looking at the sum of
             # each users karma for this update, which takes into account
@@ -2138,10 +2154,8 @@ class Update(Base):
             for karma in feedback.values():
                 if karma < 0:
                     return False
-            num_days = int(config.get('critpath.stable_after_days_without_negative_karma'))
             return self.days_in_testing >= num_days
 
-        num_days = self.release.mandatory_days_in_testing
         if not num_days:
             return True
 
@@ -2166,7 +2180,7 @@ class Update(Base):
         If this release does not have a mandatory testing requirement, then
         simply return True.
         """
-        min_num_days = self.release.mandatory_days_in_testing
+        min_num_days = self.mandatory_days_in_testing
         if min_num_days:
             if not self.meets_testing_requirements:
                 return False
@@ -2190,7 +2204,7 @@ class Update(Base):
         mandatory_days_in_testing or longer."""
         if not self.meets_testing_requirements and self.date_testing:
             return (
-                self.release.mandatory_days_in_testing -
+                self.mandatory_days_in_testing -
                 (datetime.utcnow() - self.date_testing).days)
         return 0
 
--- a/bodhi/tests/server/test_models.py
+++ b/bodhi/tests/server/test_models.py
@@ -360,6 +360,52 @@ class TestUpdate(ModelTest):
         eq_(self.obj.builds[0].release.name, u'F11')
         eq_(self.obj.builds[0].package.name, u'TurboGears')
 
+    def test_mandatory_days_in_testing_critpath(self):
+        """
+        The Update.mandatory_days_in_testing method should be the configured value
+        for critpath if it is a critpath update.
+        """
+        update = self.obj
+        update.critpath = True
+
+        # Configured value.
+        expected = int(config.get('critpath.stable_after_days_without_negative_karma'))
+
+        eq_(update.mandatory_days_in_testing, expected)
+
+    def test_mandatory_days_in_testing(self):
+        """
+        The Update.mandatory_days_in_testing method should be a positive integer if the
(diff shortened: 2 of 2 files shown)


bodhi/server/templates/update.html
```diff
@@ -785,7 +785,7 @@ <h5 class="m-t-2">
                     </tr>
                     % endif
 
-                    % if update.days_to_stable and update.status.description == 'testing':
+                    % if update.days_to_stable > 0 and update.status.description == 'testing':
```
bowlofeggs: I think the bug is actually in the calculation of ```days_to_stable```. That code lives [here](https://github.com/fedora-infra/bodhi/blob/28dd380a1d077de3953a9464d765cea9e6db1235/bodhi/server/models.py#L1965-L1977):

```
        if not self.meets_testing_requirements and self.date_testing:
            return (
                self.release.mandatory_days_in_testing -
                (datetime.utcnow() - self.date_testing).days)
        return 0
```

The issue seems to be that critical path updates have a different setting for how many days they should stay in testing, and this code doesn't take that into account. [```meets_testing_requirements```](https://github.com/fedora-infra/bodhi/blob/28dd380a1d077de3953a9464d765cea9e6db1235/bodhi/server/models.py#L1903-L1937) returns ```False``` for critical path updates (like the one I cited in the bug I filed), because it is using ```config.get('critpath.stable_after_days_without_negative_karma')``` to find the number of days required in testing.

So I think the thing to do is to create a new method on the Update model, perhaps ```Update.mandatory_days_in_testing()``` (analogous to the ```Release.mandatory_days_in_testing()``` method), that could look something like this:

```
if self.critpath:
    return int(config.get('critpath.stable_after_days_without_negative_karma'))

num_days = self.release.mandatory_days_in_testing
return num_days if num_days else 0
```

Then we could modify ```meets_testing_requirements``` and ```days_to_stable``` to both use this source of truth for their respective calculations. I found one other place in ```models.py``` that also uses the ```critpath.stable_after_days_without_negative_karma``` setting that could be updated to use this. There are also a few spots that use ```self.release.mandatory_days_in_testing()``` that could use the new ```self.mandatory_days_in_testing()```. How does that sound?

bowlofeggs (changes requested):

bodhi/server/models.py
```diff
@@ -789,6 +789,14 @@ class Update(Base):
     user_id = Column(Integer, ForeignKey('users.id'))
 
     @property
+    def mandatory_days_in_testing(self):
```
bowlofeggs: Let's add a docblock here to describe the method.

bodhi/server/models.py
```diff
@@ -1936,6 +1944,9 @@ def meets_testing_requirements(self):
         If this release does not have a mandatory testing requirement, then
         simply return True.
         """
+
```
bowlofeggs: It's not a requirement to merge, but I suggest removing this newline.

bodhi/server/models.py
```diff
@@ -789,6 +789,14 @@ class Update(Base):
     user_id = Column(Integer, ForeignKey('users.id'))
 
     @property
+    def mandatory_days_in_testing(self):
+        if self.critpath:
+            return int(config.get('critpath.stable_after_days_without_negative_karma'))
+
+        days = self.release.mandatory_days_in_testing
+        return days if days else 0
```
bowlofeggs, line 994: We should add three tests for this new method:

1. Test a critpath update to make sure this returns an integer of the setting.
2. Test a regular update with ```release.mandatory_days_in_testing``` set to some "truthy" value, like 7.
3. Test a regular update with ```release.mandatory_days_in_testing``` set to some "falsy" value, like None.

bodhi/server/models.py
```diff
@@ -1999,7 +2008,7 @@ def days_to_stable(self):
         mandatory_days_in_testing or longer."""
         if not self.meets_testing_requirements and self.date_testing:
             return (
-                self.release.mandatory_days_in_testing -
+                self.mandatory_days_in_testing -
                 (datetime.utcnow() - self.date_testing).days)
         return 0
```
bowlofeggs, line 2209: We need to add a test for this method to prove that the original bug that was filed is now solved. Basically, set up a critical path update and make sure this returns the correct number of days.

bowlofeggs (changes requested):
Let me know if you need help getting started with the tests!

bodhi/server/models.py
nbianca: How can I achieve test number 3? There should be a release with `mandatory_days_in_testing` equals to 0. Is there a way to mock that config nicely?

My current solution looks something like [this](https://gist.github.com/nbianca/4e6a9c1a244395e46ca4de0251c73d0e). It ain't beautiful and it ain't working either (but it should be like this).

bowlofeggs: Hello @nbianca!

[```Release.mandatory_days_in_testing()```](https://github.com/fedora-infra/bodhi/blob/2.5.0/bodhi/server/models.py#L411-L424) does something that is admittedly a bit convoluted in that it loads a dynamically selected config key to determine what the value should be (instead of getting it from the DB or the object).

Thus, you could use [```mock.patch.dict()```](http://www.voidspace.org.uk/python/mock/patch.html#patch-dict) to patch in a config item for a given release. I've not tested this, but something along these lines should work:

```
@mock.patch.dict('bodhi.server.models.config', {'fc10.status': 'stable'})
@mock.patch.dict('bodhi.server.models.config', {'fc10.stable.mandatory_days_in_testing': '0'})
def test_blah_blah_blah(self):
    # Code in here should see 0 as the mandatory days in testing
```

By the way I didn't realize it worked this way when I wrote my previous comment but since it's using ```int()``` on the value you won't be able to encode ```None```, so let's use ```'0'``` as the "falsy" value.

bowlofeggs: Btw, there's another path in that code that would let you set only one config item if you want to do that instead - it involves setting ```{id_prefix}.mandatory_days_in_testing``` in the config. Slightly simpler and it'll achieve the same result. Either way you prefer is fine of course ☺

bodhi/tests/server/test_models.py
```diff
@@ -269,6 +269,42 @@ def test_builds(self):
...
+        """
+        update = self.get_update()
+        update.critpath = True
+        update.date_testing = datetime.utcnow() + timedelta(days=-4)
+
+        criptath_days_to_stable = int(
+            config.get('critpath.stable_after_days_without_negative_karma'))
+
+        eq_(update.days_to_stable, criptath_days_to_stable - 4)
+
```
bowlofeggs, line 408: These tests cover conditions 1 and 2 that I listed above, but we could still use a third test for the condition where ```release.mandatory_days_in_testing``` is set to 0. You should be able to use ```mock``` to achieve this as I mentioned in my earlier comment.

bowlofeggs (changes requested):

nbianca:
Hello @bowlofeggs,

I did the changes. I hope it is ready to merge now.

bowlofeggs (approved):

bowlofeggs:
Thanks!

PR #1388, opened 2017-03-28
Require contributors to sign the FPCA.

bowlofeggs:


bowlofeggs:
It was [decided](https://lists.fedoraproject.org/archives/list/legal@lists.fedoraproject.org/thread/WGICHTUDCXOIDBPXTLZO7T57Y5ACLLAJ/) on Fedora's legal list that we don't need or want to require contributors to sign the FPCA.

PR #1391, opened 2017-03-30
s/mage page/man page/.

bowlofeggs:


cverna:
👍

pypingou:
I think I like the old one better, kinda make me think about [LBA](https://en.wikipedia.org/wiki/Little_Big_Adventure)

bowlofeggs:
@pypingou hahaha

PR #1396, opened 2017-03-31
Fix #1310: If a user edits an update after Bodhi comments it can be pushed, it won't comment again

### LINKED ISSUE #1310: If a user edits an update after Bodhi comments it can be pushed, it won't comment again
I noticed that my Bodhi update in Bodhi reached 14 days in testing this week, but Bodhi never commented on it to tell me that it had been in testing long enough to be pushed to stable:

https://bodhi.fedoraproject.org/updates/FEDORA-EPEL-2016-cd060eeb3c

After poking around in the staging db for a while to find out what gives, I found that the update's [```met_testing_requirements()```](https://github.com/fedora-infra/bodhi/blob/2.4.0/bodhi/server/models/models.py#L1830-L1854) method was returning True because long ago Bodhi had commented to state that the update could be pushed to stable, before I had edited the update and reset its karma and date_testing. The code doesn't take the karma reset event into account:

```
        min_num_days = self.release.mandatory_days_in_testing
        if min_num_days:
            if not self.meets_testing_requirements:
                return False
        else:
            return True
        for comment in self.comments:
            if comment.user.name == u'bodhi' and \
               comment.text.startswith('This update has reached') and \
               'and can be pushed to stable now if the ' \
               'maintainer wishes' in comment.text:
                return True
        return False
```

I think we probably want to take [this loop](https://github.com/fedora-infra/bodhi/blob/2.4.0/bodhi/server/models/models.py#L691-L698) from the ```karma``` property and make it into a reusable method that just returns comments that should be considered for karma and testing requirements:

```
        for comment in reversed(self.comments):
            if (comment.user.name == u'bodhi' and
                    ('New build' in comment.text or 'Removed build' in comment.text)):
                # We only want to consider comments since the most recent karma reset, which happens
                # whenever a build is added or removed from an Update. Since we are traversing the
                # comments in reverse order, once we find one of these comments we can simply exit
                # this loop.
                break
```

This way we can have ```met_testing_requirements()``` and ```karma()``` both use this new method to get the comments that they use to do what they do.

crungehottman:
should fix issue #1310

### DIFF
--- a/bodhi/server/models.py
+++ b/bodhi/server/models.py
@@ -797,17 +797,7 @@ class Update(Base):
         positive_karma = 0
         negative_karma = 0
         users_counted = set()
-        # We want to traverse the comments in reverse order so we only consider the most recent
-        # comments from any given user, and only the comments since the most recent karma reset
-        # event.
-        for comment in reversed(self.comments):
-            if (comment.user.name == u'bodhi' and
-                    ('New build' in comment.text or 'Removed build' in comment.text)):
-                # We only want to consider comments since the most recent karma reset, which happens
-                # whenever a build is added or removed from an Update. Since we are traversing the
-                # comments in reverse order, once we find one of these comments we can simply exit
-                # this loop.
-                break
+        for comment in self.comments_since_karma_reset:
             if comment.karma and not comment.anonymous and comment.user.name not in users_counted:
                 # Make sure we only count the last comment this user made
                 users_counted.add(comment.user.name)
@@ -817,6 +807,28 @@ class Update(Base):
                     negative_karma += comment.karma
 
         return positive_karma, negative_karma
+
+    @property
+    def comments_since_karma_reset(self):
+        """
+        Return the comments since the most recent karma reset event, which
+        occurs whenever a build is added or removed from an Update.
+        :return: an iterable of recent comments
+        :rtype: generator
+        """
+        # We want to traverse the comments in reverse order so we only consider
+        # the most recent comments from any given user and only the comments
+        # since the most recent karma reset event.
+        for comment in reversed(self.comments):
+            if (comment.user.name == u'bodhi' and
+                    ('New build' in comment.text or 'Removed build' in comment.text)):
+                # We only want to consider comments since the most recent karma
+                # reset, which happens whenever a build is added or removed
+                # from an Update. Since we are traversing the comments in
+                # reverse order, once we find one of these comments we can
+                # simply exit this loop.
+                break
+            yield comment
 
     @classmethod
     def new(cls, request, data):
@@ -1954,7 +1966,7 @@ class Update(Base):
                 return False
         else:
             return True
-        for comment in self.comments:
+        for comment in self.comments_since_karma_reset:
             if comment.user.name == u'bodhi' and \
                comment.text.startswith('This update has reached') and \
                'and can be pushed to stable now if the ' \
--- a/bodhi/tests/server/scripts/test_approve_testing.py
+++ b/bodhi/tests/server/scripts/test_approve_testing.py
@@ -333,3 +333,38 @@ class TestMain(BaseTestCase):
             c.user.name
             for c in self.db.query(models.Comment).order_by(models.Comment.timestamp).all()]
         self.assertEqual(usernames, [u'guest', u'anonymous', u'hunter2'])
+
+    @patch.dict(config, [('fedora.mandatory_days_in_testing', 14)])
+    def test_subsequent_comments_after_initial_push_comment(self):
+        """
+        If a user edits an update after Bodhi comments a testing_approval_msg,
+        Bodhi should send an additional testing_approval_msg when the revised
+        update is eligible to be pushed to stable.
+
+        See https://github.com/fedora-infra/bodhi/issues/1310
+        """
+        update = self.db.query(models.Update).all()[0]
+        update.request = None
+        update.status = models.UpdateStatus.testing
+        update.date_testing = datetime.now() - timedelta(days=14)
+        self.db.flush()
+
+        with patch(
+                'bodhi.server.scripts.approve_testing._get_db_session', return_value=self.db):
+            approve_testing.main(['nosetests', 'some_config.ini'])
+            update.comment(self.db, u"Removed build", 0, u'bodhi')
+            approve_testing.main(['nosetests', 'some_config.ini'])
+
+        bodhi = self.db.query(models.User).filter_by(name=u'bodhi').one()
+        cmnts = self.db.query(models.Comment).filter_by(update_id=update.id, user_id=bodhi.id)
+        # There are 3 comments: testing_approval_msg, build change, testing_approval_msg
+        self.assertEqual(cmnts.count(), 3)
+        self.assertEqual(
+            cmnts[0].text,
+            config.get('testing_approval_msg') %
+            update.release.mandatory_days_in_testing)
+        self.assertEqual(cmnts[1].text, 'Removed build')
+        self.assertEqual(
+            cmnts[2].text,
+            config.get('testing_approval_msg') %
+            update.release.mandatory_days_in_testing)


bodhi/tests/server/scripts/test_approve_testing.py
```diff
@@ -333,3 +333,44 @@ def test_non_autokarma_update_without_mandatory_days_in_testing(self):
...
+            approve_testing.main(['nosetests', 'some_config.ini'])
+            update.comment(self.db, u"Removed build", 0, u'bodhi')
+            update.status = models.UpdateStatus.testing
+            update.date_testing = datetime.now()
+            self.assertEqual(update.meets_testing_requirements, False)
+            update.date_testing = datetime.now() - timedelta(days=14)
+            approve_testing.main(['nosetests', 'some_config.ini'])
+
+        bodhi = self.db.query(models.User).filter_by(name=u'bodhi').one()
+        cmnts = self.db.query(models.Comment).filter_by(update_id=update.id, user_id=bodhi.id)
```
bowlofeggs: It would be good to assert that there are three comments here.

bodhi/tests/server/scripts/test_approve_testing.py
```diff
@@ -333,3 +333,44 @@ def test_non_autokarma_update_without_mandatory_days_in_testing(self):
...
+        update.comment(self.db, text, author=u'bodhi')
+        self.assertEqual(update.meets_testing_requirements, True)
+
+        with patch(
+                'bodhi.server.scripts.approve_testing._get_db_session', return_value=self.db):
+            approve_testing.main(['nosetests', 'some_config.ini'])
+            update.comment(self.db, u"Removed build", 0, u'bodhi')
+            update.status = models.UpdateStatus.testing
+            update.date_testing = datetime.now()
+            self.assertEqual(update.meets_testing_requirements, False)
```
bowlofeggs: I don't think we need the above two lines, as they are really asserting that ```meets_testing_requirements``` works.

bodhi/tests/server/scripts/test_approve_testing.py
```diff
@@ -333,3 +333,44 @@ def test_non_autokarma_update_without_mandatory_days_in_testing(self):
...
+        """
+        If a user edits an update after Bodhi comments saying that it can be
+        pushed, Bodhi should comment again saying that it has been in testing
+        long enough to be pushed to stable.
+        """
+        update = self.db.query(models.Update).all()[0]
+        update.request = None
+        update.status = models.UpdateStatus.testing
+        update.date_testing = datetime.now() - timedelta(days=14)
+        self.assertEqual(update.days_in_testing, 14)
```
bowlofeggs: I recommend dropping this line too, since it's just testing that the line above it works.

bodhi/tests/server/scripts/test_approve_testing.py
```diff
@@ -333,3 +333,44 @@ def test_non_autokarma_update_without_mandatory_days_in_testing(self):
...
+        text = unicode(config.get('testing_approval_msg') % update.days_in_testing)
+        update.comment(self.db, text, author=u'bodhi')
+        self.assertEqual(update.meets_testing_requirements, True)
```
bowlofeggs: I recommend dropping this line since it's testing ```meets_testing_requirements()```. We can rely on tests elsewhere to make sure that method works.

bodhi/server/models.py
```diff
@@ -818,6 +808,31 @@ def _composite_karma(self):
...
+            if (comment.user.name == u'bodhi' and
+                    ('New build' in comment.text or 'Removed build' in comment.text)):
+                # We only want to consider comments since the most recent karma
+                # reset, which happens whenever a build is added or removed
+                # from an Update. Since we are traversing the comments in
+                # reverse order, once we find one of these comments we can
+                # simply exit this loop.
+                break
+            else:
+                tempcomments.append(comment)
```
bowlofeggs: If you use Python's [```yield``` statement](https://docs.python.org/2/reference/simple_stmts.html#yield) in this method, it will become a [generator](https://wiki.python.org/moin/Generators) and will not have to load all of the comments into a temporary list. This will make the code a bit more efficient, nearly as efficient as it had been before (though incorrect before ☺). Also, you do not need the ```else``` clause because not hitting the ```break``` implies that the ```if``` statement wasn't matched. Thus, I recommend refactoring the above to something like this:

```
for comment in reversed(self.comments):
    if (comment.user.name == u'bodhi' and
            ('New build' in comment.text or 'Removed build' in comment.text)):
    yield comment
```

bodhi/server/models.py
```diff
@@ -818,6 +808,31 @@ def _composite_karma(self):
 
         return positive_karma, negative_karma
 
+    def get_comments_since_karma_reset(self):
```
bowlofeggs: I recommend naming this ```comments_since_karma_reset()``` and using the [```@property```](https://docs.python.org/2/library/functions.html#property) decorator on it. Something like this:

```
@property
def comments_since_karma_reset(self):
    …
```

This will allow callers to use it like an attribute instead of having to explicitly call it, which is kind of nice. For example, callers can use ```update.comments_since_karma_reset``` instead of ```update.get_comments_since_karma_reset()```. A lot of the methods on the update class work like this, so it would fit in nicely.

bowlofeggs (changes requested):
Nice work, just a few suggestions!

crungehottman:
@bowlofeggs I implemented the changes. Using the yield statement is much more efficient. Thank you so much for your helpful feedback!

bodhi/server/models.py
```diff
@@ -818,6 +808,28 @@ def _composite_karma(self):
 
         return positive_karma, negative_karma
 
+    @property
+    def comments_since_karma_reset(self):
+        """
+        Return the comments since the most recent karma reset event, which
+        occurs whenever a build is added or removed from an Update.
+        :return: a list of comments since karma reset
+        :rtype: list
```
bowlofeggs: Technically this is returning a generator now. I'd change this from list to generator, and you can change the line above this to say "an iterable of comments…" or generator if you prefer.

bodhi/tests/server/scripts/test_approve_testing.py
```diff
@@ -333,3 +333,41 @@ def test_non_autokarma_update_without_mandatory_days_in_testing(self):
...
+        update.date_testing = datetime.now() - timedelta(days=14)
+        text = unicode(config.get('testing_approval_msg') % update.days_in_testing)
+        update.comment(self.db, text, author=u'bodhi')
+
+        with patch(
+                'bodhi.server.scripts.approve_testing._get_db_session', return_value=self.db):
+            approve_testing.main(['nosetests', 'some_config.ini'])
+            update.comment(self.db, u"Removed build", 0, u'bodhi')
+            update.status = models.UpdateStatus.testing
+            update.date_testing = datetime.now() - timedelta(days=14)
```
bowlofeggs: We also don't need this line or the one above it since lines 346 and 348 above already makes these same changes.

bodhi/tests/server/scripts/test_approve_testing.py
```diff
@@ -333,3 +333,41 @@ def test_non_autokarma_update_without_mandatory_days_in_testing(self):
...
+        update = self.db.query(models.Update).all()[0]
+        update.request = None
+        update.status = models.UpdateStatus.testing
+        update.date_testing = datetime.now() - timedelta(days=14)
+        text = unicode(config.get('testing_approval_msg') % update.days_in_testing)
+        update.comment(self.db, text, author=u'bodhi')
+
+        with patch(
+                'bodhi.server.scripts.approve_testing._get_db_session', return_value=self.db):
+            approve_testing.main(['nosetests', 'some_config.ini'])
```
bowlofeggs: You probably don't need this line, since line 349 already adds the comment. So you could drop this line, or you could drop lines 348 and 349.

bowlofeggs (changes requested):
Just a few more changes. I prefer to keep tests short and as simple as possible so they are easy to read. When tests fail, it's more obvious what is going wrong if they are pretty focused.

crungehottman:
Changes made! Thanks again.

bowlofeggs (changes requested):
@crungehottman This looks great! I'm sorry I didn't notice this before, but I have oooone more change to ask for. Can you change your commit message to be more descriptive, and move the bug reference number in to the commit message body? Something like this would do:

```
Bodhi's testing approval comment now respects the karma reset event

<feel free to expand if you want by putting a longer description here, but not required>

fixes #1310
```

Feel free to season to taste, but the general idea is to give a short but descriptive summary of the change, an optional more descriptive body, and a last line with the ```fixes #1310``` so that GitHub will autoclose the associated issue. That fixes is also handy when people want to know what issue this was associated with.

crungehottman:
I think it should be okay now! Thanks for the tip

bowlofeggs (approved):
Nice, this looks great!

PR #1398, opened 2017-03-31
Turn the Package model into a single-table inheritance model

### LINKED ISSUE #1392: Split the Package model into two
We need to split the Package model into two models, a superclass (probably still called Package) and a new RPM-specific subclass, perhaps simply called RPM. All of Package's relationships should be analyzed to determine which ones should be adjusted to reference RPM vs. continuing to reference Package. Also, a database migration will be needed so that Bodhi's existing data can be preserved in the new split package design. All code will need to be altered to use the new RPM model rather than the Package model when the data involved is RPM-specific.

jeremycline:
This turns the Package model into the parent class using single-table
inheritance that sub-classes can inherit from to create different
Package types. It includes a migration that sets the type of all
existing Packages to RpmPackage.

fixes #1392

### DIFF
--- a/alembic/versions/fc6b0169c596_add_a_column_for_single_table_.py
+++ b/alembic/versions/fc6b0169c596_add_a_column_for_single_table_.py
@@ -17,8 +17,8 @@ down_revision = '4f2f825bcf4a'
 def upgrade():
     """Add the column used for polymorphic identity in Packages."""
     # The default of ``1`` is the RPM package type.
-    op.add_column('packages', sa.Column('type', sa.Integer(), nullable=False, default=1))
-    op.alter_column('packages', 'type', default=None)
+    op.add_column('packages', sa.Column('type', sa.Integer(), nullable=False, server_default=u'1'))
+    op.alter_column('packages', 'type', server_default=None)
 
 
 def downgrade():
--- a/alembic/versions/fc6b0169c596_add_a_column_for_single_table_.py
+++ b/alembic/versions/fc6b0169c596_add_a_column_for_single_table_.py
@@ -1,0 +1,26 @@
+"""
+Add a column for single-table inheritance to the packages table.
+
+Revision ID: fc6b0169c596
+Revises: 4f2f825bcf4a
+Create Date: 2017-03-31 16:40:07.136469
+"""
+
+from alembic import op
+import sqlalchemy as sa
+
+# revision identifiers, used by Alembic.
+revision = 'fc6b0169c596'
+down_revision = '4f2f825bcf4a'
+
+
+def upgrade():
+    """Add the column used for polymorphic identity in Packages."""
+    # The default of ``1`` is the RPM package type.
+    op.add_column('packages', sa.Column('type', sa.Integer(), nullable=False, default=1))
+    op.alter_column('packages', 'type', default=None)
+
+
+def downgrade():
+    """Remove the column used for polymorphic identity in Packages."""
+    op.drop_column('packages', 'type')
--- a/bodhi/server/models.py
+++ b/bodhi/server/models.py
@@ -476,12 +476,28 @@ class TestCase(Base):
 
 
 class Package(Base):
+    """
+    This model represents a package.
+
+    This model uses single-table inheritance to allow for different package types.
+
+    Attributes:
+        name (unicode): A unicode string that uniquely identifies the package.
+        requirements (unicode): A unicode string that describes the requirements of a package.
+        type (int): The polymorphic identity column. This is used to identify what Python
+            class to create when loading rows from the database.
+        builds (list): A list of :class:`Build` objects.
+        test_cases (list): A list of :class:`TestCase` objects.
+        committers (list): A list of :class:`User` objects who are committers.
+        stack_id (int): A foreign key to the :class:`Stack`
+    """
     __tablename__ = 'packages'
     __get_by__ = ('name',)
     __exclude_columns__ = ('id', 'committers', 'test_cases', 'builds',)
 
     name = Column(UnicodeText, unique=True, nullable=False)
     requirements = Column(UnicodeText)
+    type = Column(Integer, nullable=False)
 
     builds = relationship('Build', backref=backref('package', lazy='joined'))
     test_cases = relationship('TestCase', backref='package')
@@ -489,6 +505,11 @@ class Package(Base):
                               backref='packages')
 
     stack_id = Column(Integer, ForeignKey('stacks.id'))
+
+    __mapper_args__ = {
+        'polymorphic_on': type,
+        'polymorphic_identity': 0,
+    }
 
     def get_pkg_pushers(self, branch, settings):
         """ Pull users who can commit and are watching a package.
@@ -586,6 +607,14 @@ class Package(Base):
                     x += "    o %s\n" % update.get_title()
         del states
         return x
+
+
+class RpmPackage(Package):
+    """Represents a RPM package."""
+
+    __mapper_args__ = {
+        'polymorphic_identity': 1,
+    }
 
 
 class Build(Base):
--- a/bodhi/server/models.py
+++ b/bodhi/server/models.py
@@ -483,12 +483,15 @@ class Package(Base):
 
     Attributes:
         name (unicode): A unicode string that uniquely identifies the package.
-        requirements (unicode): A unicode string that describes the requirements of a package.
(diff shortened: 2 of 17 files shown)


alembic/versions/fc6b0169c596_add_a_column_for_single_table_.py
```diff
@@ -0,0 +1,26 @@
...
+
+# revision identifiers, used by Alembic.
+revision = 'fc6b0169c596'
+down_revision = '4f2f825bcf4a'
+
+
+def upgrade():
+    """Add the column used for polymorphic identity in Packages."""
+    # The default of ``1`` is the RPM package type.
+    op.add_column('packages', sa.Column('type', sa.Integer(), nullable=False, default=1))
```
bowlofeggs: This migration fails to apply:

```
[vagrant@bodhi-dev bodhi]$ alembic upgrade head
/usr/lib64/python2.7/pkgutil.py:186: ImportWarning: Not importing directory '/usr/lib64/python2.7/site-packages/zope': missing __init__.py
  file, filename, etc = imp.find_module(subname, path)
/usr/lib64/python2.7/pkgutil.py:186: ImportWarning: Not importing directory '/usr/lib/python2.7/site-packages/zope': missing __init__.py
  file, filename, etc = imp.find_module(subname, path)
/usr/lib64/python2.7/pkgutil.py:186: ImportWarning: Not importing directory '/usr/lib/python2.7/site-packages/paste': missing __init__.py
  file, filename, etc = imp.find_module(subname, path)
/usr/lib64/python2.7/pkgutil.py:186: ImportWarning: Not importing directory '/usr/lib/python2.7/site-packages/moksha': missing __init__.py
  file, filename, etc = imp.find_module(subname, path)
/usr/lib64/python2.7/pkgutil.py:186: ImportWarning: Not importing directory '/usr/lib/python2.7/site-packages/repoze': missing __init__.py
  file, filename, etc = imp.find_module(subname, path)
/home/vagrant/bodhi/bodhi/server/__init__.py:26: DeprecationWarning: unauthenticated_userid: As of Pyramid 1.5 the "pyramid.security.unauthenticated_userid" API is now deprecated.  It will be removed in Pyramd 1.8.  Use the "unauthenticat
ed_userid" attribute of the Pyramid request instead.
  from pyramid.security import unauthenticated_userid
INFO  [alembic.runtime.migration] Context impl PostgresqlImpl.
INFO  [alembic.runtime.migration] Will assume transactional DDL.
INFO  [alembic.runtime.migration] Running upgrade 4f2f825bcf4a -> fc6b0169c596, Add a column for single-table inheritance to the packages table.
  File "/usr/bin/alembic", line 9, in <module>
    load_entry_point('alembic==0.8.6', 'console_scripts', 'alembic')()
  File "/usr/lib/python2.7/site-packages/alembic/config.py", line 479, in main
    CommandLine(prog=prog).main(argv=argv)
  File "/usr/lib/python2.7/site-packages/alembic/config.py", line 473, in main
    self.run_cmd(cfg, options)
  File "/usr/lib/python2.7/site-packages/alembic/config.py", line 456, in run_cmd
    **dict((k, getattr(options, k)) for k in kwarg)
  File "/usr/lib/python2.7/site-packages/alembic/command.py", line 174, in upgrade
    script.run_env()
  File "/usr/lib/python2.7/site-packages/alembic/script/base.py", line 397, in run_env
    util.load_python_file(self.dir, 'env.py')
  File "/usr/lib/python2.7/site-packages/alembic/util/pyfiles.py", line 93, in load_python_file
    module = load_module_py(module_id, path)
  File "/usr/lib/python2.7/site-packages/alembic/util/compat.py", line 79, in load_module_py
    mod = imp.load_source(module_id, path, fp)
  File "alembic/env.py", line 75, in <module>
    run_migrations_online()
  File "alembic/env.py", line 68, in run_migrations_online
    context.run_migrations()
  File "<string>", line 8, in run_migrations
  File "/usr/lib/python2.7/site-packages/alembic/runtime/environment.py", line 797, in run_migrations
    self.get_context().run_migrations(**kw)
  File "/usr/lib/python2.7/site-packages/alembic/runtime/migration.py", line 312, in run_migrations
    step.migration_fn(**kw)
  File "/home/vagrant/bodhi/alembic/versions/fc6b0169c596_add_a_column_for_single_table_.py", line 20, in upgrade
    op.add_column('packages', sa.Column('type', sa.Integer(), nullable=False, default=1))
  File "<string>", line 8, in add_column
  File "<string>", line 3, in add_column
  File "/usr/lib/python2.7/site-packages/alembic/operations/ops.py", line 1535, in add_column
    return operations.invoke(op)
  File "/usr/lib/python2.7/site-packages/alembic/operations/base.py", line 318, in invoke
    return fn(self, operation)
  File "/usr/lib/python2.7/site-packages/alembic/operations/toimpl.py", line 123, in add_column
    schema=schema
  File "/usr/lib/python2.7/site-packages/alembic/ddl/impl.py", line 172, in add_column
    self._exec(base.AddColumn(table_name, column, schema=schema))
  File "/usr/lib/python2.7/site-packages/alembic/ddl/impl.py", line 118, in _exec
    return conn.execute(construct, *multiparams, **params)
  File "/usr/lib64/python2.7/site-packages/sqlalchemy/engine/base.py", line 914, in execute
    return meth(self, multiparams, params)
  File "/usr/lib64/python2.7/site-packages/sqlalchemy/sql/ddl.py", line 68, in _execute_on_connection
    return connection._execute_ddl(self, multiparams, params)
  File "/usr/lib64/python2.7/site-packages/sqlalchemy/engine/base.py", line 968, in _execute_ddl
    compiled
  File "/usr/lib64/python2.7/site-packages/sqlalchemy/engine/base.py", line 1146, in _execute_context
    context)
  File "/usr/lib64/python2.7/site-packages/sqlalchemy/engine/base.py", line 1341, in _handle_dbapi_exception
    exc_info
  File "/usr/lib64/python2.7/site-packages/sqlalchemy/util/compat.py", line 202, in raise_from_cause
    reraise(type(exception), exception, tb=exc_tb, cause=cause)
  File "/usr/lib64/python2.7/site-packages/sqlalchemy/engine/base.py", line 1139, in _execute_context
    context)
  File "/usr/lib64/python2.7/site-packages/sqlalchemy/engine/default.py", line 450, in do_execute
    cursor.execute(statement, parameters)
sqlalchemy.exc.IntegrityError: (psycopg2.IntegrityError) column "type" contains null values
 [SQL: 'ALTER TABLE packages ADD COLUMN type INTEGER NOT NULL']
```

I think we need to make it nullable, then run a query to set existing rows to ```1```, then run an update to make it non-nullable.

bodhi/server/models.py
```diff
@@ -476,12 +476,28 @@ class TestCase(Base):
...
 
 class Package(Base):
+    """
+    This model represents a package.
+
+    This model uses single-table inheritance to allow for different package types.
+
+    Attributes:
+        name (unicode): A unicode string that uniquely identifies the package.
+        requirements (unicode): A unicode string that describes the requirements of a package.
```
bowlofeggs: Perhaps "A unicode string that lists space-separated taskotron test results that must pass for this package."

bodhi/server/models.py
```diff
@@ -476,12 +476,28 @@ class TestCase(Base):
...
+        type (int): The polymorphic identity column. This is used to identify what Python
+            class to create when loading rows from the database.
+        builds (list): A list of :class:`Build` objects.
+        test_cases (list): A list of :class:`TestCase` objects.
+        committers (list): A list of :class:`User` objects who are committers.
```
bowlofeggs: Are these really lists, or are they generators or some sqlalchemy type?

bowlofeggs (changes requested):

alembic/versions/fc6b0169c596_add_a_column_for_single_table_.py
jeremycline: The setting I wanted was ``server_default`` rather than ``default``. This lets you immediately set the column to be non-nullable (because it'll use the default for everything) and then you just alter the column to drop the server default. It comes to the same thing, though.

bodhi/server/models.py
jeremycline: I thought they were just lists because I checked on Anitya, but it turned out that was a property making them a list. It's actually an SQLAlchemy type. I updated the docs.

bowlofeggs (approved):
LGTM.

PR #1410, opened 2017-04-04
Add update type icons to update list tables

ryanlerch:
In the new bodhi webinterface, we now use icons to signify
if the update is one of the 4 types -- security, newpackage,
bugfix or enhancement. This adds these icons to the lists of
updates that we render

icons-in-update-page

### DIFF
--- a/bodhi/server/templates/fragments.html
+++ b/bodhi/server/templates/fragments.html
@@ -77,7 +77,7 @@
 </%def>
 
 <%def name="update(update, display_user=True, display_request=False, display_release=True, display_karma=True)">
-  <td class="stretch-table-column">${util.update2html(update) | n}</td>
+  <td class="stretch-table-column">${util.update2html(update) | n} ${util.type2icon(update['type'])|n} </td>
   <td class="nowrap">
     <span title="${update['date_submitted']} UTC">
       ${util.age(update['date_submitted'], True)}


bodhi/server/templates/fragments.html
```diff
@@ -77,7 +77,7 @@
 </%def>
 
 <%def name="update(update, display_user=True, display_request=False, display_release=True, display_karma=True)">
-  <td class="stretch-table-column">${util.update2html(update) | n}</td>
+  <td class="stretch-table-column">${util.update2html(update) | n} ${util.type2icon(update['type'])|n} </td>
```
pypingou, line 80: Will this add a ``title="..."`` to it?

ryanlerch: yes! util.type2icon is the same call we make to render the the top icons on the update page, so the same title will be there. Reuse FTW!. Here is a screenshot for good measure:

screenshot from 2017-04-04 18-39-20

pypingou: Cool thanks :)

pypingou:
Looks good to me :)

bowlofeggs (approved):
I'm OK with this change as is, but I have a suggestion to consider - what if we put the type icons to the left of the package name instead of the right? Then they will all line up which might look nice.

Either way, merge as you please!

Issue #1408, opened 2017-04-03
The --notes-file parameter should be shared between the new and edit CLI commands


bowlofeggs:
Right now the CLI's new update command has a ```--notes-file``` parameter that can be used to pass a file with notes to the CLI. The edit command has ```--notes``` but would also benefit from ```--notes-file```.

PR #1411, opened 2017-04-04
Fix issue #1408  --notes-file

crungehottman:
The --notes-file parameter is now shared between the new and edit CLI commands.

### DIFF
--- a/bodhi/client/__init__.py
+++ b/bodhi/client/__init__.py
@@ -28,6 +28,7 @@ new_edit_options = [
     click.option('--type', default='bugfix', help='Update type', required=True,
                  type=click.Choice(['security', 'bugfix', 'enhancement', 'newpackage'])),
     click.option('--notes', help='Update description'),
+    click.option('--notes-file', help='Update description from a file'),
     click.option('--bugs', help='Comma-seperated list of bug numbers', default=''),
     click.option('--close-bugs', default=True, is_flag=True, help='Automatically close bugs'),
     click.option('--request', help='Requested repository',
@@ -102,7 +103,6 @@ def updates():
 @updates.command()
 @add_options(new_edit_options)
 @click.argument('builds')
-@click.option('--notes-file', help='Update description from a file')
 @click.option('--file', help='A text file containing all the update details')
 @url_option
 def new(user, password, url, **kwargs):
@@ -125,14 +125,7 @@ def new(user, password, url, **kwargs):
     else:
         updates = client.parse_file(os.path.abspath(kwargs['file']))
 
-    if kwargs['notes_file'] is not None:
-        if kwargs['notes'] is None:
-            with open(kwargs['notes_file'], 'r') as fin:
-                kwargs['notes'] = fin.read()
-
-        else:
-            click.echo("ERROR: Cannot specify --notes and --notes-file")
-            sys.exit(1)
+    kwargs['notes'] = _get_notes(**kwargs)
 
     for update in updates:
         try:
@@ -161,6 +154,8 @@ def edit(user, password, url, **kwargs):
     """
     client = bindings.BodhiClient(base_url=url, username=user, password=password,
                                   staging=kwargs['staging'])
+
+    kwargs['notes'] = _get_notes(**kwargs)
 
     try:
         query_param = {'updateid': kwargs['updateid']}
@@ -335,6 +330,29 @@ def download(url, **kwargs):
                     ret = subprocess.call(args)
                     if ret:
                         click.echo("WARNING: download of {0} failed!".format(build['nvr']))
+
+
+def _get_notes(**kwargs):
+    """
+    If the user provides a --notes-file, _get_notes processes the contents of the notes-file.
+    If the user does not provide a --notes-file, _get_notes() returns the notes from the kwargs.
+    One cannot specify both --notes and --notesfile. Doing so will result in an error.
+
+    Args:
+        kwargs (dict): Keyword arguments passed to us by click.
+
+    :returns: the contents of the notes file or the notes from kwargs
+    :rtype: string
+    """
+    if kwargs['notes_file'] is not None:
+        if kwargs['notes'] is None:
+            with open(kwargs['notes_file'], 'r') as fin:
+                return fin.read()
+        else:
+            click.echo("ERROR: Cannot specify --notes and --notes-file")
+            sys.exit(1)
+    else:
+        return kwargs['notes']
 
 
 @cli.group()
--- a/bodhi/tests/client/test___init__.py
+++ b/bodhi/tests/client/test___init__.py
@@ -392,6 +392,61 @@ class TestEdit(unittest.TestCase):
                 'close_bugs': True, 'stable_karma': None, 'csrf_token': 'a_csrf_token',
                 'staging': False, 'builds': u'nodejs-grunt-wrap-0.3.0-2.fc25', 'autokarma': True,
                 'edited': u'nodejs-grunt-wrap-0.3.0-2.fc25', 'suggest': None,
-                'notes': u'this is an edited note', 'request': None, 'bugs': u'',
-                'unstable_karma': None, 'type': 'bugfix'})
+                'notes': u'this is an edited note', 'notes_file': None,
+                'request': None, 'bugs': u'', 'unstable_karma': None, 'type': 'bugfix'})
         self.assertEqual(bindings_client.base_url, 'http://localhost:6543/')
+
+    @mock.patch('bodhi.client.bindings.BodhiClient.csrf',
+                mock.MagicMock(return_value='a_csrf_token'))
+    @mock.patch('bodhi.client.bindings.BodhiClient.query',
+                return_value=client_test_data.EXAMPLE_QUERY_MUNCH, autospec=True)
+    @mock.patch('bodhi.client.bindings.BodhiClient.send_request',
+                return_value=client_test_data.EXAMPLE_UPDATE_MUNCH, autospec=True)
+    def test_notes_file(self, send_request, query):
+        """
+        Assert that a valid notes-file is properly handled in a successful updates
+        edit request.
+        """
+        runner = testing.CliRunner()
+        with runner.isolated_filesystem():
+            with open('notefile.txt', 'w') as f:
+                f.write('This is a --notes-file note!')
+
+            result = runner.invoke(
+                client.edit, ['FEDORA-2017-cc8582d738', '--user', 'bowlofeggs',
+                              '--password', 's3kr3t', '--notes-file', 'notefile.txt',
+                              '--url', 'http://localhost:6543'])
+
+            self.assertEqual(result.exit_code, 0)
+            bindings_client = query.mock_calls[0][1][0]
+            query.assert_called_with(
+                bindings_client, updateid=u'FEDORA-2017-cc8582d738')
+            bindings_client = send_request.mock_calls[0][1][0]
+            send_request.assert_called_with(
+                bindings_client, 'updates/', auth=True, verb='POST',
+                data={
+                    'close_bugs': True, 'stable_karma': None, 'csrf_token': 'a_csrf_token',
+                    'staging': False, 'builds': u'nodejs-grunt-wrap-0.3.0-2.fc25',
+                    'autokarma': True, 'edited': u'nodejs-grunt-wrap-0.3.0-2.fc25',
+                    'suggest': None, 'notes': 'This is a --notes-file note!',
+                    'notes_file': 'notefile.txt', 'request': None, 'bugs': u'',
+                    'unstable_karma': None, 'type': 'bugfix'})
+            self.assertEqual(bindings_client.base_url, 'http://localhost:6543/')
+
+    def test_notes_and_notes_file(self):
+        """
+        Assert providing both --notes-file and --notes parameters to an otherwise successful
+        updates edit request results in an error.
+        """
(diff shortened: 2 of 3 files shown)


bodhi/client/__init__.py
```diff
@@ -162,14 +162,21 @@ def edit(user, password, url, **kwargs):
     client = bindings.BodhiClient(base_url=url, username=user, password=password,
                                   staging=kwargs['staging'])
 
+    if kwargs['notes_file'] is not None:
+        if kwargs['notes'] is None:
+            with open(kwargs['notes_file'], 'r') as fin:
+                kwargs['notes'] = fin.read()
+        else:
+            click.echo("ERROR: Cannot specify --notes and --notes-file")
+            sys.exit(1)
```
bowlofeggs: This same code block appears in the ```new()``` function above. It would be good to make it a new function and have both ```new()``` and ```edit()``` use it.

bodhi/tests/client/test___init__.py
```diff
@@ -392,6 +392,6 @@ def test_url_flag(self, send_request, query):
                 'close_bugs': True, 'stable_karma': None, 'csrf_token': 'a_csrf_token',
                 'staging': False, 'builds': u'nodejs-grunt-wrap-0.3.0-2.fc25', 'autokarma': True,
                 'edited': u'nodejs-grunt-wrap-0.3.0-2.fc25', 'suggest': None,
-                'notes': u'this is an edited note', 'request': None, 'bugs': u'',
-                'unstable_karma': None, 'type': 'bugfix'})
+                'notes': u'this is an edited note', 'notes_file': None,
+                'request': None, 'bugs': u'', 'unstable_karma': None, 'type': 'bugfix'})
         self.assertEqual(bindings_client.base_url, 'http://localhost:6543/')
```
bowlofeggs, line 397: Let's add a test in here to assert that the --notes-file flag works on the edit command. There's a ```TestEdit``` class that would be a good place to put such a test. I can think of a couple tests that would be useful:

1. Assert that giving --notes-file properly hands the data found in a file to the ```notes``` API parameter.
2. Assert that giving ```--notes-file``` and ```--notes``` prints an error.

bowlofeggs (changes requested):

bowlofeggs:
It would also be good to squash the two commits into one.

bodhi/client/__init__.py
```diff
@@ -162,14 +162,21 @@ def edit(user, password, url, **kwargs):
...
+
```
bowlofeggs: The blank space on this line is causing the flake8 test to fail:

```
/home/vagrant/bodhi/bodhi/client/__init__.py:172:1: W293 blank line contains whitespace
F.
AssertionError: 1 != 0
```

bowlofeggs:
One more thing:

If you change you commit message to this format, GitHub will automatically close #1408 when this PR is merged:

```
<MESSAGE>

fixes #1408
```

bodhi/client/__init__.py
```diff
@@ -337,6 +331,17 @@ def download(url, **kwargs):
...
 
+def read_notes(**kwargs):
+    if kwargs['notes_file'] is not None:
+        if kwargs['notes'] is None:
+            with open(kwargs['notes_file'], 'r') as fin:
+                kwargs['notes'] = fin.read()
+        else:
+            click.echo("ERROR: Cannot specify --notes and --notes-file")
+            sys.exit(1)
+    return kwargs['notes']
```
bowlofeggs: There's sort of an unspoken (or maybe it is spoken somewhere deep in Python's docs…) convention in Python that I think is a good rule to live by - functions/methods that change an object should not return the object, and functions/methods that do not change an object should return the result. If you study Python's built-in object APIs, they all follow this convention. For example, ```list.append``` does not return the list after modifying it:

```
>>> a = [1, 2, 3]
>>> x = a.append(4)
>>> print x
None
```

As another example, ```random.choice()``` does return an object and does not modify the list, where ```random.shuffle()``` does modify the object and thus does not return it:

```
>>> import random
>>> a = [1, 2, 3]
>>> x = random.choice(a)
>>> x
3
>>> a
[1, 2, 3]
>>> x = random.shuffle(a)
>>> print x
None
>>> a
[3, 1, 2]
```

It's actually a pretty nice convention, because when software follows it you can pretty easily guess whether the function is modifying objects or not without trawling around the docs too much (even though some people like @jeremycline read docs instead of eating food).

All this to say, you can design this function to work either way (modify the notes in place, or simply return what the notes should be), but it would be conventional to not have it do both. If I were to decide, I think it'd be simplest to just have it modify the dict in place and just drop the return statement. Of course, you shoudl also document this behavior in the function's docblock.

bodhi/client/__init__.py
```diff
@@ -337,6 +331,17 @@ def download(url, **kwargs):
                         click.echo("WARNING: download of {0} failed!".format(build['nvr']))
 
 
+def read_notes(**kwargs):
```
bowlofeggs: This function is probably not intended to be used outside of this module, so it would be conventional to name it with a leading underscore, like ```_read_notes```.

We should also give it a docblock, explaining what it does, what kwargs is, and what it returns (if anything - I suggest having it just modify kwargs in place and not returning anything).

bodhi/tests/client/test___init__.py
bowlofeggs, line 397: It could also be good to add a couple of true unit tests in here on the new ```_process_notes_file()``` function. You could pass in dictionaries of various forms to exercise the various if statements within the function. Here are some ideas:

1. Assert that not passing ```--notes-file``` doesn't do anything.
2. Assert that passing ```--notes-file``` works in the normal case.
3. Assert that passing ```--notes-file``` and ```--notes``` makes it angry.

Unit tests like this can actually be pretty fun to write. Let me know if you need help getting started on these.

bowlofeggs (changes requested):
Can you also add a "fixes #1408" into the body (not the subject, so make sure it's on its own line with at least one blank line between it and the subject) of your commit message?

bodhi/tests/client/test___init__.py
crungehottman: Hi, @bowlofeggs , I do need help getting started. Actually, I think I need plenty of help. I seem to have been running in circles and thus made 0 headway since your initial comments a couple of days ago. First, would I also place these unit tests in /bodhi/tests/client/test__init__.py in the class TestEdit? 

I think my biggest problem is that I'm confused as to how to mock a file, or whether I even need to be mocking a file. I wrote [a simple unit test](https://gist.github.com/crungehottman/e6cd109242502e7a02c5a91aa289cf03) that entailed the three tests you mentioned above, but I don't know how to incorporate this tests into the bodhi project codebase. I'm not even sure if this is how I should be testing --notes-file... I'd prefer to be testing it the way @cverna implemented TestEdit -- BUT -- I made a copy of cverna's test, replacing --notes with --notes-file, but for some reason way beyond me, that causes cverna's test to fail (ha! how on earth did I manage that?) at assert_called_with. I'm sure it has something to do with my not knowing how to test this using a file. Should I be testing this with essentially the same test cvernas created (TestEdit.test_url_flag), except replacing --notes with --notes-file, and somehow creating a notes_file to test? Also, I don't understand why everyone in the class is naming their tests test_url_flag and testing 

> self.assertEqual(bindings_client.base_url, 'http://localhost:6543')

in addition to testing assert_called_with.

Is [this](http://click.pocoo.org/5/testing/#file-system-isolation) something I should be looking at re: testing a --notes-file, or am I way off?

bowlofeggs: Hello @crungehottman!

No problem at all, I'm happy to help. To answer your first question, yes that is the correct file and class to put the test for the edit command in. Not all of bodhi's test follow this convention, but I try to name the test files after the file they are testing (i.e., ```bodhi/foo/bar.py```'s tests would be in a file called ```bodhi/tests/foo/test_bar.py```) and I like to have a class per "unit" I'm testing, where the ```edit()``` function would be a unit here.

bowlofeggs: @crungehottman For your second question, the simple unit test you linked looks like a reasonable start to me, though I have a few recommendations:

0. You can use [```with open```](https://docs.python.org/2/tutorial/inputoutput.html#methods-of-file-objects) (see the end of that section) to open the file, rather than having to remember to close the file. It's a small change, but it's a nice way to do file stuff in Python.
1. I actually didn't know about that click thing you linked me, but it looks pretty sweet. I'd use that. Another way you could do it is to ask Python for a [tempfile](https://docs.python.org/2.7/library/tempfile.html) and pass it to the click client. Either way is fine - I say try out that click thing since it sounds cool.
2. I'm not sure why a modified version of @cverna's test would cause his to fail - can you update this PR with that code so I can take a look? Maybe there's some shared object between them that just needs to be separate or something like that.

Oh, and the reason so many things have a test_url_flag is that there used to be no tests at all on the client and one day I wanted to add a ```--url``` flag to the client. So I had to write tests for my new code, and thus those were the first client tests. Now there are some more tests, but those are still there. Ideally, we'd have lots of tests in there - we're getting there!

Hopefully I got the questions you've mentioned so far, let me know if there's anything else you need to know!

bowlofeggs: Oh, and to make it clear - I do think copying @cverna's test and modifying it to use the --notes-file instead of --notes is a good way to do this. That's the way I'd go ☺

cverna: copying my test is always a good idea 🤗

crungehottman: @cverna  hahaha I'll remember that!

@bowlofeggs thanks so much for your advice! The click's test file thing worked wonderfully once I finally figured out how to use it. So, I added a couple of tests:

I added the two tests you mentioned 4 days ago:

> Assert that giving --notes-file properly hands the data found in a file to the notes API parameter.
> Assert that giving --notes-file and --notes prints an error.

These should also cover the tests you mentioned 2 days ago:

> Assert that passing --notes-file works in the normal case.
> Assert that passing --notes-file and --notes makes it angry.

@cverna 's test accounts for the following case you mentioned 2 days ago:

> Assert that not passing --notes-file doesn't do anything.

So I didn't add an additional test because I figured it might be redundant.

Thoughts?

bodhi/client/__init__.py
```diff
@@ -337,6 +332,27 @@ def download(url, **kwargs):
...
+def _process_notes_file(**kwargs):
+    """
+    Processes the contents of the notes_file if one is specified.
+
+    Args:
+        kwargs (dict): Keyword arguments passed to us by click.
+
+    :returns: the contents of the notes file to be placed in kwargs['notes']
+    :rtype: string
+    """
```
bowlofeggs: Sorry to nitpick a bit here, but the method now either processes the notes file *or* just returns the notes. It's a bit different than before (and it works great!), so I think my previous name recommendation doesn't make as much sense with this functionality. So I recommend again renaming this and changing the docs. Perhaps just ```_get_notes()``` would be a good name, and the docblock should mention that it will get the notes from the notes-file if present, or will return the notes from the kwargs. It would also be good to mention in the docblock that it is an error for the user to provide both ```--notes``` and ```--notes-file```.

bodhi/tests/client/test___init__.py
bowlofeggs: @crungehottman cool, yeah I think the tests you've got here are good!

bowlofeggs (changes requested):
OK, just one more change and I think this will be good to merge!

bodhi/tests/client/test___init__.py
crungehottman: @bowlofeggs, I changed the name to `_get_notes()` and updated the doc block! I think it should be okay now. Unless... by 
> So I recommend again renaming this and changing the **docs**. 

Did you mean to change something in the manpages?

bowlofeggs: @crungehottman Oh sorry, I mean docblock, and it looks like you got that change done!

bodhi/client/__init__.py
```diff
@@ -337,6 +332,29 @@ def download(url, **kwargs):
                         click.echo("WARNING: download of {0} failed!".format(build['nvr']))
 
 
+def _get_notes(**kwargs):
+    """
+    If the user provides a --notes-file, _get_notes processes the contents of the notes-file. 
+    If the user does not provide a --notes-file, _get_notes() returns the notes from the kwargs.
```
bowlofeggs: There is a style test that is failing on these two lines due to trailing whitespace:

```
/home/vagrant/bodhi/bodhi/client/__init__.py:337:94: W291 trailing whitespace
/home/vagrant/bodhi/bodhi/client/__init__.py:338:97: W291 trailing whitespace
F.
AssertionError: 1 != 0
```

I think if you remove that whitespace we'll be good to merge this!

bowlofeggs (changes requested):

bodhi/tests/client/test___init__.py
crungehottman: @bowlofeggs  oh geez... how embarrassing! sorry about the whitespace. 

all fixed! (finally!)

bowlofeggs:
On Wed, 2017-04-12 at 10:26 -0700, Caleigh Runge-Hottman wrote:
> @bowlofeggs oh geez... how embarrassing! sorry about the whitespace.

Hahah, no worries, we've all been there ☺

I'll take another look soon!

bowlofeggs (approved):

PR #1413, opened 2017-04-05
update the label colour for testing on table lists

ryanlerch:
previously, when displaying a list of updates in a table,
the if the request field was "testing", we used the bootstrap
class for success (blue) as the colour for the label. However,
elsewhere in bodhi when showing the testing label, we used the
yellow. This changes the colour of the label to the yellow on
table list views of updates.


pypingou:
Change looks fine :)

bowlofeggs (approved):
Cool, looks nice!

PR #1414, opened 2017-04-05
Switch to a scoped session and drop zope.sqlalchemy

jeremycline:
This is a semi-large change that's broken into a few commits, but they all need to land at once.

Bodhi was littered with database initialization code and engines were
being created for every single request. This unifies all the
initialization code and uses a plain old scoped session rather than
zope.sqlalchemy. This is better because Bodhi uses the database
extensively outside the request/response cycle and zope.sqlalchemy gets
in the way more than it helps there.
    
All the code should be using the scoped session now rather than the
transaction provided by zope.

In addition, this changes how fedmsgs are sent because they relied on zope's transaction manager. This is a much smaller, cleaner method using SQLAlchemy events. In the future, we should stop calling our publish wrapper and start using SQLAlchemy events to queue messages.


bowlofeggs:
This pull request does seem to make a big positive impact on the test RAM usage. I observed a max of 2.45 G when running the test suite, where they would use nearly 4 GB before. 2.45 is still a lot of RAM, but this seems to really help.

bowlofeggs:
Can you add a release note that zope.sqlalchemy is no longer needed? I guess we can add a "dependencies" subsection, or maybe just use the "development" subsection?

Since the first commit is the one that removes the requirement for ```zope.sqlalchemy```, it might make sense to merge the third commit into it. I'm also fine with leaving them as separate commits if you prefer.

bodhi/server/__init__.py
```diff
@@ -45,23 +44,28 @@
...
-    engine = engine_from_config(request.registry.settings, 'sqlalchemy.')
-    Sess = scoped_session(sessionmaker(extension=ZopeTransactionExtension()))
-    Sess.configure(bind=engine)
-    session = Sess()
+    session = request.registry.sessionmaker()
 
     def cleanup(request):
-        # No need to do rollback/commit ourselves.  the zope transaction manager takes care of that
-        # for us. However, we want to explicitly close the session we opened
+        if request.exception is not None:
```
bowlofeggs: Though this function didn't actually have a docblock before, the comment that has been removed here was kind of like one. So it would be good to give it a docblock describing the new behavior.

bodhi/server/__init__.py
```diff
@@ -45,23 +44,28 @@
...
+            session.rollback()
+        else:
+            session.commit()
```
bowlofeggs: This will help Bodhi to not have commitment issues.

bodhi/server/__init__.py
```diff
@@ -45,23 +44,28 @@
...
         session.close()
 
     request.add_finished_callback(cleanup)
-
```
bowlofeggs: This seems like an unrelated style change.

bodhi/server/__init__.py
```diff
@@ -121,6 +125,38 @@ def exception_filter(response, request):
...
+    #: The SQLAlchemy database engine. This is constructed using the value of
+    #: ``DB_URL`` in :mod:`anitya.config``.
+    engine = engine_from_config(config, 'sqlalchemy.')
+    # When using SQLite we need to make sure foreign keys are enabled:
+    # http://docs.sqlalchemy.org/en/latest/dialects/sqlite.html#foreign-key-support
+    if config['sqlalchemy.url'].startswith('sqlite:'):
+        event.listen(
+            engine,
+            'connect',
+            lambda db_con, con_record: db_con.execute('PRAGMA foreign_keys=ON')
```
bowlofeggs, line 156: oooooh, I wonder if we could use something like this in our alembic migrations to get them to work with BDR? @puiterwijk, this could be a cool way to do it.

bodhi/server/consumers/updates.py
```diff
@@ -55,7 +54,8 @@ class UpdatesHandler(fedmsg.consumers.FedmsgConsumer):
     config_key = 'updates_handler'
 
     def __init__(self, hub, *args, **kwargs):
-        self.db_factory = models.get_db_factory()
+        engine = initialize_db(config)
+        self.db_factory = util.transactional_session_maker(engine)
 
         prefix = hub.config.get('topic_prefix')
         env = hub.config.get('environment')
```
bowlofeggs, line 61: I don't think this pull request *has* to be the one to deal with it, but there are still other consumers that are creating their own DB connections their own way. ```bodhi.server.consumers.masher``` and ```bodhi.server.consumers.signed``` are both creating connections their own way.

bodhi/server/scripts/approve_testing.py
```diff
@@ -96,19 +95,8 @@ def main(argv=sys.argv):
...
-    """
-    # There are many blocks of code like this in the codebase. We should consolidate them into a
-    # single utility function as described in https://github.com/fedora-infra/bodhi/issues/1028
-    settings = get_appsettings(config_uri)
-    engine = engine_from_config(settings, 'sqlalchemy.')
-    Session = scoped_session(sessionmaker(extension=ZopeTransactionExtension()))
-    Session.configure(bind=engine)
-    return Session()
+        db.commit()
+    except SQLAlchemyError as e:
```
bowlofeggs: What if some other exception happens here? Lke what if the string formatting fails above and raises an Exception? I assume that postgres will automatically roll a transaction back if the connection is lost from the process exiting?

bowlofeggs: I notice that in the untag branched script you caught the general exception to rollback, but here and in expire overrides you catch only this one class of exceptions. It might be good to be consistent.

jeremycline:
I added release notes and squashed as you recommended.

bodhi/server/scripts/untag_branched.py
```diff
@@ -53,15 +49,13 @@ def main(argv=sys.argv):
...
-    Session = scoped_session(sessionmaker(extension=ZopeTransactionExtension()))
-    Session.configure(bind=engine)
+    initialize_db(settings)
     db = Session()
     koji = buildsys.get_session()
     one_day = timedelta(days=1)
     now = datetime.utcnow()
 
-    with transaction.manager:
+    try:
```
bowlofeggs, line 58: Could it be nice to use ```bodhi.server.util.TransactionalSessionMaker``` here to use the ```with``` statement? I guess it doesn't have a built-in way to do things like ```log.error()```, but maybe we could add a feature to it so you can register a failure handler that gets called when a rollback happens. That might be overkill. I'm fine with this is you prefer it, or whatever. I'm just braindumping.

bodhi/tests/server/base.py
```diff
@@ -52,15 +48,13 @@
...
-        engine = create_engine(DB_PATH)
-        # We want a special session that lasts longer than a transaction
-        self.Session = scoped_session(
-            sessionmaker(bind=engine, extension=ZopeTransactionExtension(keep_session=True)))
-        log.debug('Creating all models for %s' % engine)
-        models.Base.metadata.bind = engine
-        models.Base.metadata.create_all(engine)
+        self.config = {'sqlalchemy.url': DB_PATH}
+        self.engine = initialize_db(self.config)
+        self.Session = Session
```
bowlofeggs, line 53: If ```self.Session = bodhi.server.Session``` now, maybe we can drop this line and adjust tests that were using ```self.Session```. Actually, it may be that none or few tests do use this - I may have only added this line so I could call ```self.Session.remove()``` in ```tearDown()```.

bodhi/tests/server/scripts/test_approve_testing.py
```diff
@@ -129,13 +127,12 @@ def test_non_autokarma_critpath_update_meeting_karma_requirements_gets_one_comme
...
-                'bodhi.server.scripts.approve_testing._get_db_session', return_value=self.db):
-            approve_testing.main(['nosetests', 'some_config.ini'])
-
-            # Now we will run main() again, but this time we expect Bodhi not to add any further
-            # comments.
-            approve_testing.main(['nosetests', 'some_config.ini'])
+        with patch('bodhi.server.scripts.approve_testing.initialize_db'):
+            with patch('bodhi.server.scripts.approve_testing.get_appsettings', return_value=''):
+                approve_testing.main(['nosetests', 'some_config.ini'])
+                # Now we will run main() again, but this time we expect Bodhi not to add any
```
bowlofeggs: It's a small detail, but I have a testing style where I use spacing to try to create three overall blocks in my tests: setup, line that calls thing I'm testing, and then assertions. This test actually had four since the with statement would make flake8 mad, but I actually like the space between this line and the one above, because this test is actually asserting that the second call to approve_testing works correctly (it assumes the one above does work correctly).

Same in other places in this PR.

bodhi/tests/server/test_metadata.py
```diff
@@ -129,10 +126,9 @@ def __init__(self, *args, **kw):
 
     def setUp(self):
         setup_buildsystem({'buildsystem': 'dev'})
-        engine = create_engine(DB_PATH)
-        Session = scoped_session(
-            sessionmaker(extension=ZopeTransactionExtension(keep_session=True)))
-        Session.configure(bind=engine)
+        config = {'sqlalchemy.url': DB_PATH}
+        engine = initialize_db(config)
+        self.Session = Session
```
bowlofeggs: You probably don't need this line, right?

bowlofeggs (changes requested):

bodhi/server/notifications.py
```diff
@@ -46,6 +51,31 @@ def init(active=None, cert_prefix=None):
...
 
 
+@event.listens_for(Session, 'after_commit')
+def send_fedmsgs_after_commit(session):
+    """
+    An SQLAlchemy event listener to send fedmsgs after a database commit.
+
+    This relies on the session ``info`` dictionary being populated. At the moment,
+    this is done by calling the :func:`publish` function. In the future it should
+    be done automatically using SQLAlchemy listeners.
```
bowlofeggs, line 61: Let's document the session parameter here.

bodhi/tests/server/test_notifications.py
```diff
@@ -81,3 +83,112 @@ def test_with_cert_prefix(self, init, info):
...
+        # Ensure nothing at all is in our session
+        Session.remove()
+        session = Session()
+        notifications.publish('demo.topic', {'new': 'package'})
+        session.commit()
+        mock_fedmsg_publish.assert_called_once_with(
+            topic='demo.topic', msg={'new': 'package'})
+
+    def test_repeated_commit(self, mock_fedmsg_publish):
+        """Assert a single message for a single topic is published."""
```
bowlofeggs: Many of these tests have this same docblock - can you make sure the tests all have accurate docblocks?

bowlofeggs (changes requested):
The tests look great. I like the new fedmsg integration!

bodhi/server/consumers/updates.py
jeremycline: Ahh, I missed that. I'm inclined to file an issue and clean that up in a later commit. There's lots more cleanup that needs to get done that I didn't include in this PR :(

bodhi/server/scripts/untag_branched.py
jeremycline: I'd prefer to keep it as-is since I don't want to just re-raise the exception. It is a little bit repetitive, but I'm not sure it's worth adding more complexity and abstraction for this sort of thing.

bodhi/tests/server/base.py
jeremycline: I just didn't want to get into the business of adjusting tests more than I absolutely had to. The tests really need a clean-up, but I think we want to bite that off separately. 

Edit: I checked, there's ~30 places it's used.

bodhi/server/__init__.py
```diff
@@ -121,6 +125,38 @@ def exception_filter(response, request):
...
+    This *must* be called before you can use the :data:`Session` object.
+
+    Args:
+        config (dict): The Bodhi server configuration dictionary.
+
+    Returns:
+        sqlalchemy.engine: The database engine created from the configuration.
+    """
+    #: The SQLAlchemy database engine. This is constructed using the value of
+    #: ``DB_URL`` in :mod:`anitya.config``.
```
cverna: anitya.config ??

jeremycline: Whoops. I started out with what I just did in Anitya and missed this. Good eye! It's gone ✨

bodhi/server/consumers/updates.py
bowlofeggs: ©ol, fair enough.

bodhi/tests/server/base.py
bowlofeggs: Fair enough.

bowlofeggs (approved):
LGTM - but please squash your fourth commit (the one that says its addressing PR comments) before merging.

This PR is very nice!

PR #1416, opened 2017-04-06
Clean up imports in bodhi.server.models.

bowlofeggs:

### DIFF
--- a/bodhi/server/models.py
+++ b/bodhi/server/models.py
@@ -14,58 +14,43 @@
 # along with this program; if not, write to the Free Software
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
 
-import os
-import re
+from collections import defaultdict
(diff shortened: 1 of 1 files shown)


cverna:
Nice, clean and ordered 🥇 
LGTM

bodhi/server/models.py
```diff
@@ -14,60 +14,49 @@
...
-import time
-import uuid
-
-from textwrap import wrap
-from datetime import datetime
 from collections import defaultdict
-
+from datetime import datetime
+from textwrap import wrap
 try:
```
jeremycline: You _could_ use six to make this import less ugly, but it's up to you.

jeremycline (approved):
Looks tidy. There's one totally optional note :shipit:

bodhi/server/models.py
bowlofeggs: Cool, I'll use six. Thanks!

PR #1417, opened 2017-04-06
Checks exit code of process to determine where stderr is logged.

### LINKED ISSUE #1412: Bodhi is logging stderr at error level in bodhi.server.util.cmd
Bodhi is [logging subprocess stderr output at the error level](https://github.com/fedora-infra/bodhi/blob/2.5.0/bodhi/server/util.py#L604-L605), which is not correct. Many programs use stderr for debug or info type output, and it doesn't necessarily indicate that there was an error.

Instead, Bodhi should check the exit code of the process that was called. If the exit code is 0, the stderr should be logged at info. If it is non-zero, then it is OK to log messages at error.

This is causing e-mails like this to get sent:

```
Message
-------
[2017-04-03 22:56:16][bodhi.server   ERROR]
python
From https://pagure.io/fedora-comps
   8cfc441..3ea2f69  master     -> origin/master



Process Details
---------------
- host:     bodhi-backend01.phx2.fedoraproject.org
- PID:      52590
- name:     fedmsg-hub
- command:  /usr/bin/python /usr/bin/fedmsg-hub
- msg_id:   

Callstack that lead to the logging statement
--------------------------------------------
python
  File "/usr/lib64/python2.7/threading.py", line 777 in __bootstrap
    self.__bootstrap_inner()
  File "/usr/lib64/python2.7/threading.py", line 804 in __bootstrap_inner
    self.run()
  File "/usr/lib/python2.7/site-packages/bodhi/server/consumers/masher.py", line 259 in run
    self.work()
  File "/usr/lib/python2.7/site-packages/bodhi/server/consumers/masher.py", line 317 in work
    self.update_comps()
  File "/usr/lib/python2.7/site-packages/bodhi/server/consumers/masher.py", line 641 in update_comps
    util.cmd(['git', 'pull'], comps_dir)
  File "/usr/lib/python2.7/site-packages/bodhi/server/util.py", line 581 in cmd
    log.error(err)
```

crungehottman:
If exit code is 0, the stderr is logged at info. If it is non-zero, stderr logs messages at error.

fixes #1412

### DIFF
--- a/bodhi/server/util.py
+++ b/bodhi/server/util.py
@@ -601,7 +601,10 @@ def cmd(cmd, cwd=None):
     if out:
         log.debug(out)
     if err:
-        log.error(err)
+        if p.returncode == 0:
+            log.debug(err)
+        else:
+            log.error(err)
     if p.returncode != 0:
         log.error('return code %s', p.returncode)
     return out, err, p.returncode
--- a/bodhi/tests/server/test_utils.py
+++ b/bodhi/tests/server/test_utils.py
@@ -12,6 +12,9 @@
 # along with this program; if not, write to the Free Software
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
 
+import mock
+import unittest
+import subprocess
 from bodhi.server.buildsys import setup_buildsystem, teardown_buildsystem
 from bodhi.server.config import config
 from bodhi.server.util import (get_critpath_pkgs, get_nvr, markup,
@@ -70,3 +73,59 @@ class TestUtils(object):
         b1, b2 = sorted_builds([new, old])
         assert b1 == new, b1
         assert b2 == old, b2
+
+
+class TestCMDFunctions(unittest.TestCase):
+    @mock.patch('bodhi.server.log.debug')
+    @mock.patch('bodhi.server.log.error')
+    @mock.patch('subprocess.Popen')
+    def test_err_nonzero_return_code(self, mock_popen, mock_error, mock_debug):
+        """
+        Ensures proper behavior when there is err output and the exit code isn't 0.
+        See https://github.com/fedora-infra/bodhi/issues/1412
+        """
+        mock_popen.return_value = mock.Mock()
+        mock_popen_obj = mock_popen.return_value
+        mock_popen_obj.communicate.return_value = ('output', 'error')
+        mock_popen_obj.returncode = 1
+        cmd('/bin/echo', '"home/imgs/catpix"')
+        mock_popen.assert_called_once_with(['/bin/echo'], cwd='"home/imgs/catpix"',
+                                           stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+        mock_error.assert_any_call('error')
+        mock_debug.assert_called_once_with('output')
+
+    @mock.patch('bodhi.server.log.debug')
+    @mock.patch('bodhi.server.log.error')
+    @mock.patch('subprocess.Popen')
+    def test_no_err_zero_return_code(self, mock_popen, mock_error, mock_debug):
+        """
+        Ensures proper behavior when there is no err output and the exit code is 0.
+        See https://github.com/fedora-infra/bodhi/issues/1412
+        """
+        mock_popen.return_value = mock.Mock()
+        mock_popen_obj = mock_popen.return_value
+        mock_popen_obj.communicate.return_value = ('output', None)
+        mock_popen_obj.returncode = 0
+        cmd('/bin/echo', '"home/imgs/catpix"')
+        mock_popen.assert_called_once_with(['/bin/echo'], cwd='"home/imgs/catpix"',
+                                           stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+        mock_error.assert_not_called()
+        mock_debug.assert_called_once_with('output')
+
+    @mock.patch('bodhi.server.log.debug')
+    @mock.patch('bodhi.server.log.error')
+    @mock.patch('subprocess.Popen')
+    def test_err_zero_return_code(self, mock_popen, mock_error, mock_debug):
+        """
+        Ensures proper behavior when there is err output, but the exit code is 0.
+        See https://github.com/fedora-infra/bodhi/issues/1412
+        """
+        mock_popen.return_value = mock.Mock()
+        mock_popen_obj = mock_popen.return_value
+        mock_popen_obj.communicate.return_value = ('output', 'error')
+        mock_popen_obj.returncode = 0
+        cmd('/bin/echo', '"home/imgs/catpix"')
+        mock_popen.assert_called_once_with(['/bin/echo'], cwd='"home/imgs/catpix"',
+                                           stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+        mock_error.assert_not_called()
+        mock_debug.assert_called_with('error')


bodhi/server/util.py
```diff
@@ -604,7 +604,10 @@ def cmd(cmd, cwd=None):
     if out:
         log.debug(out)
     if err:
-        log.error(err)
+        if p.returncode == 0:
+            log.info(err)
+        else:
+            log.error(err)
```
bowlofeggs: This change looks good, but we will want to add some tests to make sure it works correctly (and stays working correctly).

I suggest using [```mock```](http://www.voidspace.org.uk/python/mock/index.html) to fake a call to subprocess so you can create conditions for a few tests:

0. There is no err output.
1. There is err output and the exit code is 0.
2. There is err output and the exit code is not 0.

Let me know if you need help getting started with these!

bowlofeggs (changes requested):

bodhi/server/util.py
crungehottman: Hello, @bowlofeggs !

I started asserting the proper behavior like such:
`result = cmd('')`
`self.assertEqual(result[0], 'output')`
`self.assertEqual(result[1], None)`
`self.assertEqual(result[2], 0)`

Then I realized checking the result of the cmd() function was pretty silly, because cmd() is returning stderr from the subprocess regardless. I need to check the messages at the INFO and ERROR levels of a logger. I have been searching online all day, but I still don't understand how to check the messages at specific levels of a logger... I found this tool: http://testfixtures.readthedocs.io/en/latest/logging.html, but I'm not sure whether using such a library is acceptable, or if there's a simpler way to check... Any guidance would be greatly appreciated!

bowlofeggs: Hello @crungehottman!

I would expect that you would be able to use [```mock```](http://www.voidspace.org.uk/python/mock/index.html) to force ```subprocess.Popen()``` to return a ```MagicMock``` that does what you want for each test. Here's a toy example that shows how I can use ```mock``` to force a fake ```subprocess``` call and make it give the return values and exit code I want for testing:

```
import subprocess                                                                                                    
                                                                                                                     
import mock


def x():
    p = subprocess.Popen(['/bin/echo', '"Hello World!"'], stdout=subprocess.PIPE,
                         stderr=subprocess.PIPE)
    out, err = p.communicate()

    if out:
        print "stdout: {}".format(out)
    if err:
        if p.returncode == 0:
            print "info: {}".format(err)
        else:
            print "ERROR: {}".format(err)


with mock.patch('subprocess.Popen') as Popen:
    Popen.return_value.communicate.return_value = ('yooo', 'sup')
    Popen.return_value.returncode = 1

    x()
```

If you run that, you'll see this output:

```
stdout: yooo
ERROR: sup
```

Then you can basically make the three tests I suggested by messing with the return value and exit codes. By the way, mock can be done via the with statement or through a decorator. If you ```git grep mock``` in the bodhi codebase, you'll see lots of examples of how to do it as a decorator on a test method.

Does that help answer your question?

bowlofeggs: I think I missed the second component to your question - as for the log levels, you can use ```mock``` to mock the logger too, and then you can assert that the mock was called. After an object is mocked, you can inspect the mock object to see all calls to it and make your assertions. In the example I gave above, you can say things like ```Popen.mock_calls``` to see all the calls made to Popen and you can assert it was called with ```['/bin/echo', '"Hello World!"']``` and other nifty things. ```mock``` also has built in assertion helpers, like [```assert_called_with```](http://www.voidspace.org.uk/python/mock/mock.html#mock.Mock.assert_called_with). So I think you can use that technique to mock both the call to ```Popen``` and the calls to the log, and then you can assert that the correct log level was used.

Hope that helps!

crungehottman: Hey @bowlofeggs, thank you so much for the advice! You are so very helpful. Can't believe I didn't think of also mocking the logger... doh! :)

How do my tests look? I'm not sure whether it's sufficient to only check whether the output is being logged at the error level, or if I should also test whether they are being logged at info. Also, I'm a little inconsistent by mocking using the `with` statement *and* the decorator, but when I tried to use the following decorator: `@mock.patch('bodhi.server.log.error')`, I got the following error: `ValueError: need more than 0 values to unpack`. Not sure what that's about...but mocking using the `with` statement works flawlessly.

bowlofeggs: Hello @crungehottman!

That ```ValueError``` probably happened because you might have forgotten to add a parameter to the test function to accept the mock object. Unless you provide ```patch``` with an object to replace the object with, you need to make sure the test accepts the mock itself as a parameter. This is also useful because you can use that parameter to inspect the mock. Here's some code to illustrate:

```
@mock.patch('some.thing')
def test_cool_stuff(self, thing):
    """Note that thing is in my parameter list, and can be used to see what calls to thing happened."""
    self.assertTrue(thing is 'cool')
```

Now, you can actually also use patch to set the object to something explicitly, in which case the mock isn't handed to you. This is handy when you don't need to assert that the mock was used in any particular way, but just want to cause some side effect (for example, maybe you just want to change a setting or something.) Here's an example of that style:

```
@mock.patch('some.thing', mock.MagicMock(return_value='cool'))
def test_cool_stuff(self):
    """Note that I didn't get the mock as a parameter this time."""
    self.assertEqual("This test is useless", "This test is useless")
```

bodhi/tests/server/test_utils.py
```diff
@@ -70,3 +72,50 @@ def test_sorted_builds(self):
...
+    def test_stderr(self, mock_popen):
+        """
+        Ensures proper behavior when there is err output and the exit code isn't 0.
+        See https://github.com/fedora-infra/bodhi/issues/1412
+        """
+        mock_popen.return_value = mock.Mock()
+        mock_popen_obj = mock_popen.return_value
+        mock_popen_obj.communicate.return_value = ('output', 'error')
+        mock_popen_obj.returncode = 1
+        with mock.patch('bodhi.server.log.error') as mock_error:
```
bowlofeggs: You can mock both log.error and log.info to make assertions on both of them. It's possible to do that by nesting the ```with``` statements but that won't look as pretty as using the decorator style so I recommend doing it that way.

bodhi/tests/server/test_utils.py
```diff
@@ -70,3 +72,50 @@ def test_sorted_builds(self):
...
+            cmd('/bin/echo', '"Hello World"')
+            self.assertTrue(mock_popen.called)
```
bowlofeggs: Let's make a stronger assertion than just that popen was called - you can use something like this:

```
mock_popen.assert_called_once_with(['bin/echo', '"Hello World"'], cwd=None, stdout=subprocess.PIPE,
                                   stderr=subprocess.PIPE)
```

This would be good to do in the other tests as well.

bodhi/tests/server/test_utils.py
```diff
@@ -70,3 +72,50 @@ def test_sorted_builds(self):
...
+            mock_error.assert_any_call('error')
```
bowlofeggs: Let's also mock the info logger and assert that "output" was logged to it. Let's also do this in the two tests below.

bowlofeggs (changes requested):

bodhi/tests/server/test_utils.py
crungehottman: Hi @bowlofeggs, as far as I understand, "output" should never be logged to info. `log.info('Running %r', cmd)` (first line of the `cmd()` function) should be the only call to log.info. Unless, of course, there is stderr output and the return code is 0:
```
if err:
    if p.returncode == 0:
        log.info(err)
```
I did mock the log.info to ensure that this was the case, though I'm not sure it's necessary. 

I could very well be wrong, but then I'm not sure what I'm missing.

bodhi/tests/server/test_utils.py
crungehottman: If you're curious, I figured out what was causing the ValueError. The order of my decorators was the problem. I assumed the order in which you placed the patch was the order in which you added the parameter.
E.g.)

````
@mock.patch('some.thing.a')
@mock.patch('some.thing.b')
@mock.patch('some.thing.c')
def test_alphabet(a, b, c)
```
But it's really 
````
@mock.patch('some.thing.c')
@mock.patch('some.thing.b')
@mock.patch('some.thing.a')
def test_alphabet(a, b, c)
```

Well, now I know. :)

bodhi/server/util.py
crungehottman: If you're curious, I figured out what was causing the ValueError. The order of my decorators was the problem. I assumed the order in which you placed the patch was the order in which you added the parameter.
E.g.)

```
@mock.patch('some.thing.a')
@mock.patch('some.thing.b')
@mock.patch('some.thing.c')
def test_alphabet(a, b, c)

```
Well, now I know. :)

bodhi/server/util.py
```diff
@@ -601,7 +601,10 @@ def cmd(cmd, cwd=None):
     if out:
         log.debug(out)
```
bowlofeggs, line 602: ooooh - I had mistakenly thought this line was logging out at info rather than debug. My apologies for the confusion on the test. Let's use the test to assert that the stdout is getting logged as debug.

bodhi/server/util.py
```diff
@@ -601,7 +601,10 @@ def cmd(cmd, cwd=None):
...
     if err:
-        log.error(err)
+        if p.returncode == 0:
+            log.info(err)
```
bowlofeggs: Since we log stdout at debug, let's change this info to debug as well. No reason to log stderr at info if we're logging stdout at debug.

bodhi/server/util.py
bowlofeggs: @crungehottman ah yes, when you stack them up they kind of apply on top of each other, which counterintuitively gets them passed in reverse order. It is weird, but it makes sense if you think about how decorators work - each decorator decorates the function under it (and returns a function) - so c is applied first, then b, then a, and each tacks on its new parameter at the end of the list. So the list is ```[c, b, a]``` strangely ☺

bodhi/tests/server/test_utils.py
```diff
@@ -70,3 +73,59 @@ def test_sorted_builds(self):
         b1, b2 = sorted_builds([new, old])
         assert b1 == new, b1
         assert b2 == old, b2
+
+
+class TestCMDFunctions(unittest.TestCase):
+    @mock.patch('bodhi.server.log.info')
```
bowlofeggs: Again apologies about my confusion leading you astray here - let's also mock the debug log and assert that the right things are getting logged at debug in each of these three cases.

bowlofeggs (changes requested):

bodhi/server/util.py
crungehottman: @bowlofeggs no worries! I think I understood and implemented your changes. I pushed my changes, too, so you can find them here. But because I squashed my commits, my changes might not be _super_ obvious, so I'll recap to make sure we are on the same page:

- if an err is present, but the exit code is 0, log to DEBUG (**not** log to info like we did before)
- test to make sure stdout is logged at debug
- change commit message to reflect err output to debug (**not** log to info like we did before)
- no reason to mock the logger's info level when we're mocking the debug level instead

bowlofeggs (approved):
Good work!

PR #1419, opened 2017-04-07
updated to use published package

jmtaylor90:
Update to use the python2-sqlalchemy_schemadisplay package in stable instead of kojipkgs link.


bowlofeggs (approved):
Thanks!

PR #1420, opened 2017-04-07
Split the Build model to be polymorphic.

### LINKED ISSUE #1393: Split the Build model into two
We need to split the Build model into two models, a superclass (probably still called Build) and a new RPM-specific subclass, perhaps simply called RPMBuild. All of Build's relationships should be analyzed to determine which ones should be adjusted to reference RPM vs. continuing to reference Build. Also, a database migration will be needed so that Bodhi's existing data can be preserved in the new split Build design. All code will need to be altered to use the new RPM model rather than the Build model when the data involved is RPM-specific.

bowlofeggs:
This commit reworks the Build model to be polymorphic to allow it
to support more types than just RPMs, and it adds a new RpmBuild
model.

fixes #1393

### DIFF
--- a/bodhi/server/models.py
+++ b/bodhi/server/models.py
@@ -616,22 +616,141 @@ class RpmPackage(Package):
 
 
 class Build(Base):
+    """
+    This model represents a specific build of a package.
+
+    This model uses single-table inheritance to allow for different build types.
+
+    Attributes:
+        inherited (bool): The purpose of this column is unknown, and it appears to be unused. At the
+            time of this writing, there are 112,234 records with inherited set to False and 0 with
+            it set to True in the Fedora Bodhi deployment.
+        nvr (unicode): The nvr field is really a mapping to the Koji build_target.name field, and is
+            used to reference builds in Koji. It is named nvr in reference to the dash-separated
+            name-version-release Koji name for RPMs, but it is used by other types as well. At the
+            time of this writing, it was not practical to rename nvr since it is used in the REST
+            API to reference builds. Thus, it should be thought of as a Koji build identifier rather
+            than strictly as an RPM's name, version, and release.
+        package_id (int): A foreign key to the Package that this Build is part of.
+        release_id (int): A foreign key to the Release that this Build is part of.
+        signed (bool): If True, this package has been signed by robosignatory. If False, it has not
+            been signed yet.
+        update_id (int): A foreign key to the Update that this Build is part of.
+        release (sqlalchemy.orm.relationship): A relationship to the Release that this build is part
+            of.
+        type (int): The polymorphic identify of the row. This is used by sqlalchemy to identify
+            which subclass of Build to use.
+    """
     __tablename__ = 'builds'
     __exclude_columns__ = ('id', 'package', 'package_id', 'release',
                            'release_id', 'update_id', 'update', 'override')
     __get_by__ = ('nvr',)
 
     nvr = Column(Unicode(100), unique=True, nullable=False)
-    epoch = Column(Integer, default=0)
     package_id = Column(Integer, ForeignKey('packages.id'))
     release_id = Column(Integer, ForeignKey('releases.id'))
-    update_id = Column(Integer, ForeignKey('updates.id'))
     signed = Column(Boolean, default=False, nullable=False)
+    update_id = Column(Integer, ForeignKey('updates.id'))
 
     release = relationship('Release', backref='builds', lazy=False)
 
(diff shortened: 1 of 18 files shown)


bodhi/server/models.py
```diff
@@ -621,24 +621,74 @@ class RpmPackage(Package):
...
+
+    Attributes:
+        nvr (unicode): A dash (-) separated string of an RPM's name, version, and release (e.g.
+            u'bodhi-2.5.0-1.fc26')
+        epoch (int): The RPM's epoch.
+    """
+
+    __get_by__ = ('nvr',)
+
+    nvr = Column(Unicode(100), unique=True, nullable=False)
```
jeremycline: From the docs, "the columns which correspond to a single subclass are nullable" for single-table inheritance. This might work alright at the moment, but I suspect if you were to define a third class and it didn't have a ``nvr`` column the database would throw a fit since this column is not marked as nullable.

You can enforce the non-nullability of the column in a validator, but I don't think you can have the database enforce it.

bowlofeggs: @jeremycline Good catch! I've dropped the ```nullable=False``` here and I also updated the migration to make this change.

jeremycline: One thing I was thinking about a bit yesterday: RPMs have NVRs and modules will have the name-stream-version thing. Will containers have some string ID?

Would it be better to put this column as a non-nullable, unique field on the parent and call it... I don't know, ``string_identifier`` (that's not great, but you get the idea)? You could add properties to the RpmBuild so the column can be worked with via ``nvr``.

bowlofeggs: @jeremycline Yeah I like that suggestion. I will make it so.

bowlofeggs: We could call it ```koji_identifier```. So far, all the types we are adding are in koji and these strings are used to identify the artifact in koji… How does that sound?

jeremycline: Sounds good to me

bowlofeggs: @jeremycline So there's something that will be tricky about renaming that field - the ```__get_by__``` class attribute references ```nvr```, which means that the ```nvr``` field is part of the API. In order to call it something else than ```nvr```, we would be making an API change unless we put something in place to translate this field (which might be doable).

Another option may be to keep calling it nvr on the superclass. I just spend an unfortunate amount of time trying to find where koji's ```getBuild()``` is defined. I never found it (```git grep getBuild\(``` doesn't even find it?). I was hoping to see if the parameter to that xml-rpc function is called ```nvr``` or not. If it is, the name might actually make sense to stay as it is.

jeremycline: Ah. In that case, I'd say keep calling it ``nvr`` on the superclass. We can change it later if we really want to, but in the meantime we don't break anything.

bowlofeggs: OK, I've moved nvr back to the superclass. I also moved all the methods that used nvr back to the superclass (I had moved them to the subclass since they used that field before). I also resolved the conflict, so this is ready for re-review.

bodhi/server/models.py
```diff
@@ -616,24 +616,143 @@ class RpmPackage(Package):
...
+    __mapper_args__ = {
+        'polymorphic_on': type,
+        'polymorphic_identity': 0,
+    }
+
+    def get_url(self):
+        """
+        Return a the url to details about this build.
+
+        This method appears to be unused and incorrect.
```
jeremycline, line 667: Heh

bodhi/server/models.py
```diff
@@ -616,24 +616,143 @@ class RpmPackage(Package):
...
+
+
+class RpmBuild(Build):
+    """
+    Represents an RPM build.
+
+    Note that this model uses single-table inheritance with its Build superclass.
+
+    Attributes:
+        nvr (unicode): A dash (-) separated string of an RPM's name, version, and release (e.g.
```
jeremycline, line 735: I assume this is leftover from when nvr was on this model. I think it's okay to remove this since its documented on the parent class.

jeremycline (approved):
I think there's one docblock that's incorrect, but otherwise it looks good to me. :shipit:

bodhi/server/models.py
bowlofeggs: @jeremycline Actually, I had left it here on purpose so that the meaning of nvr could be defined for this model specifically. The parent class documents it as a koji identifier, but this adds that it's name-version-release for RPMs.

PR #1422, opened 2017-04-07
Support editing an override via the CLI

### LINKED ISSUE #1049: [RFE] Expire overrides using CLI.
It seems it is not possible to expire override using CLI. The web UI seems to be the only option ATM. Please consider adding this feature.

```
$ rpm -q bodhi-client
bodhi-client-2.2.4-1.fc26.noarch
```

I noted this originally here:

https://bugzilla.redhat.com/show_bug.cgi?id=1366114#c6

cverna:
This commit add support for editing overrides to the CLI.
By using the overrides edit command you can now expire an override using the
--expired flag.
Fixed #1049

### DIFF
--- a/bodhi/client/__init__.py
+++ b/bodhi/client/__init__.py
@@ -432,14 +432,14 @@ def print_resp(resp, client):
         click.echo(client.update_str(resp))
     elif 'overrides' in resp:
         if len(resp.overrides) == 1:
-            click.echo(client.override_str(resp.overrides[0]))
+            click.echo(client.override_str(resp.overrides[0], minimal=False))
         else:
             for override in resp.overrides:
                 click.echo(client.override_str(override).strip())
         click.echo(
             '%s overrides found (%d shown)' % (resp.total, len(resp.overrides)))
     elif 'build' in resp:
-        click.echo(client.override_str(resp))
+        click.echo(client.override_str(resp, minimal=False))
     elif 'comment' in resp:
         click.echo('The following comment was added to %s' % resp.comment['update'].title)
         click.echo(resp.comment.text)
--- a/bodhi/client/__init__.py
+++ b/bodhi/client/__init__.py
@@ -22,41 +22,6 @@ import click
 
 from bodhi.client import bindings
 
-new_edit_options = [
-    click.option('--user', envvar='USERNAME'),
-    click.option('--password', prompt=True, hide_input=True),
-    click.option('--type', default='bugfix', help='Update type', required=True,
-                 type=click.Choice(['security', 'bugfix', 'enhancement', 'newpackage'])),
-    click.option('--notes', help='Update description'),
-    click.option('--notes-file', help='Update description from a file'),
-    click.option('--bugs', help='Comma-seperated list of bug numbers', default=''),
-    click.option('--close-bugs', default=True, is_flag=True, help='Automatically close bugs'),
-    click.option('--request', help='Requested repository',
-                 type=click.Choice(['testing', 'stable', 'unpush'])),
-    click.option('--autokarma', default=True, is_flag=True, help='Enable karma automatism'),
-    click.option('--stable-karma', type=click.INT, help='Stable karma threshold'),
-    click.option('--unstable-karma', type=click.INT, help='Unstable karma threshold'),
-    click.option('--suggest', help='Post-update user suggestion',
-                 type=click.Choice(['logout', 'reboot'])),
-    click.option('--staging', help='Use the staging bodhi instance',
-                 is_flag=True, default=False)]
-
-
-def add_options(options):
-    """ Given a list of click options this creates a decorator that
-    will return a function used to add the options to a click command.
-    :param options: a list of click.options decorator.
-    """
-    def _add_options(func):
-        """ Given a click command and a list of click options this will
-        return the click command decorated with all the options in the list.
-        :param func: a click command function.
-        """
-        for option in reversed(options):
-            func = option(func)
-        return func
-    return _add_options
-
 
 def _warn_if_url_and_staging_set(ctx, param, value):
     """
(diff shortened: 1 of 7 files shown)


bodhi/client/__init__.py
```diff
@@ -41,6 +41,19 @@
...
+save_edit_options = [
+    click.argument('nvr'),
+    click.option('--duration', default=7, type=click.INT,
+                 help='Number of days the override should exist.'),
+    click.option('--notes', default="No explanation given...",
+                 help='Notes on why this override is in place.'),
+    click.option('--user', envvar='USERNAME'),
+    click.option('--password', prompt=True, hide_input=True),
+    click.option('--staging', help='Use the staging bodhi instance',
+                 is_flag=True, default=False)]
```
bowlofeggs: If you add ```url_option``` to this list, you won't need to include ```@url_option``` on both the save and edit commands.

bodhi/client/__init__.py
```diff
@@ -368,33 +381,44 @@ def query_buildroot_overrides(url, user=None, **kwargs):
...
+        password (unicode): The user's password.
         url (unicode): The URL of a Bodhi server to create the update on. Ignored if staging is
                        True.
+        staging (bool): Whether to use the staging server or not.
         kwargs (dict): Other keyword arguments passed to us by click.
     """
     client = bindings.BodhiClient(base_url=url, username=user, password=password, staging=staging)
-    resp = client.save_override(nvr=nvr, duration=duration, notes=notes)
+    kwargs['edited'] = kwargs['nvr']
+    resp = client.save_override(**kwargs)
```
bowlofeggs: It seems that the only difference between save and edit is the ```kwargs['edited']``` line. What if we make a helper function that accepts a boolean called ```edit``` and then we can make the code just do

```
if edit:
    kwargs['edited'] = kwargs['nvr']
```

Then the save and edit function here can both share the other code.

bodhi/client/bindings.py
```diff
@@ -273,7 +273,7 @@ def comment(self, update, comment, karma=0, email=None):
                   'csrf_token': self.csrf()})
 
     @errorhandled
-    def save_override(self, nvr, duration, notes):
+    def save_override(self, **kwargs):
```
bowlofeggs: This change is not backwards compatible, and the bindings aren't only used by the Bodhi client. I think instead we should add a new ```edited``` parameter that defaults to ```False``` to preserve backwards compatibility, while continuing to allow ```nvr```, ```duration```, and ```notes``` to be positional arguments.

bodhi/tests/client/test___init__.py
```diff
@@ -395,3 +394,40 @@ def test_url_flag(self, send_request, query):
                 'notes': u'this is an edited note', 'request': None, 'bugs': u'',
                 'unstable_karma': None, 'type': 'bugfix'})
         self.assertEqual(bindings_client.base_url, 'http://localhost:6543/')
+
+
+class TestEditBuilrootOverrides(unittest.TestCase):
+    """
+    Test the save_buildroot_overrides() function.
```
bowlofeggs: s/save/edit/

bodhi/client/bindings.py
```diff
@@ -391,7 +387,7 @@ def testable(self):
                     yield update
 
     @staticmethod
-    def override_str(override):
+    def override_str(override, minimal=False):
```
bowlofeggs: To preserve backward compatibility, let's make ```minimal``` default to ```True```.

bodhi/client/bindings.py
```diff
@@ -400,12 +396,22 @@ def override_str(override):
...
+
+        val = "%s\n%s\n%s\n" % ('=' * 60, '\n'.join(
+            textwrap.wrap(override['build']['nvr'].replace(',', ', '), width=60,
+                          initial_indent=' ' * 5, subsequent_indent=' ' * 5)), '=' * 60)
+        val += "  Submitter: {}\n".format(override['submitter']['name'])
+        val += "  Expiration Date: {}\n".format(override['expiration_date'])
+        val += "  Notes: {}\n".format(override['notes'])
+        val += "  Expired: {}".format(override['expired_date'] is not None)
+
+        return val
```
bowlofeggs, line 423: This change seems fine in general, but it's also separate of adding editable buildroot overrides. Please break it into a second commit so each commit is atomic and specific in scope. You can include that commit in this same PR if you want to.

docs/man_pages/bodhi.rst
```diff
@@ -90,6 +90,21 @@ The ``overrides`` command allows users to manage build overrides.
...
+    ``--duration <days>``
+
+        The number of days the override should exist, given as an integer.
+
+    ``--notes <text>``
+
+        Notes on why this override is in place.
+
+    ``--expired``
+        Force an override to the expired state.
```
bowlofeggs: Since these options are shared with the save command, maybe there's a way to document them once and just mention that they are shared by the edit and save subcommands?

bowlofeggs (changes requested):
Cool, this looks great! Just a few things to change. Thanks!

cverna:
@bowlofeggs Thanks for the review, I have followed your recommendations.

bowlofeggs:
Can you rebase this pull request to resolve the conflicts? Thanks!

bodhi/client/__init__.py
```diff
@@ -17,11 +17,36 @@
 import subprocess
 import sys
 import traceback
-
```
bowlofeggs: This space is intentional - by PEP-8 imports should be grouped into three groups - Python built-ins, third party libraries, and then imports from your own project:

https://www.python.org/dev/peps/pep-0008/#imports

bodhi/tests/client/test___init__.py
```diff
@@ -450,3 +449,39 @@ def test_notes_and_notes_file(self):
 
             self.assertEqual(result.exit_code, 1)
             self.assertEqual(result.output, u'ERROR: Cannot specify --notes and --notes-file\n')
+
```
bowlofeggs, line 451: There should be two blank lines here for PEP-8. This is causing a test to fail:

```
/home/vagrant/bodhi/bodhi/tests/client/test___init__.py:453:1: E302 expected 2 blank lines, found 1
F.
AssertionError: 1 != 0
```

bodhi/client/__init__.py
```diff
@@ -386,34 +423,38 @@ def query_buildroot_overrides(url, user=None, **kwargs):
...
+                       True.
         staging (bool): Whether to use the staging server or not.
+        kwargs (dict): Other keyword arguments passed to us by click.
+    """
+    _save_override(url=url, user=user, password=password, staging=staging, **kwargs)
+
+
+@overrides.command('edit')
+@add_options(save_edit_options)
+@click.option('--expired', help='Expire the override', is_flag=True, default=False)
```
bowlofeggs: Even though the API uses the past-tense for ```expired```, this isn't quite the best human interface because it's not an active verb. I think we should change this to ```--expire``` and then have the function convert the ```expire``` kwarg to ```expired``` for the API.

bodhi/client/bindings.py
```diff
@@ -273,7 +273,7 @@ def comment(self, update, comment, karma=0, email=None):
                   'csrf_token': self.csrf()})
 
     @errorhandled
-    def save_override(self, nvr, duration, notes):
+    def save_override(self, nvr, duration, notes, edit=False, expired=False):
         """ Save a buildroot override.
 
         This entails either creating a new buildroot override, or editing an
```
bowlofeggs, line 279: Let's add the new parameters to the docblock here.

bodhi/client/bindings.py
```diff
@@ -287,14 +287,17 @@ def save_override(self, nvr, duration, notes):
...
-
-        return self.send_request(
-            'overrides/', verb='POST', auth=True, data={
-                'nvr': nvr,
+        data = {'nvr': nvr,
                 'expiration_date': expiration_date,
                 'notes': notes,
                 'csrf_token': self.csrf(),
-            })
+                'duration': duration}
```
bowlofeggs: Does the API use ```duration```? I didn't find it after an admittedly brief search.

docs/release_notes.rst
```diff
@@ -35,6 +35,8 @@ Features
   (`#1321 <https://github.com/fedora-infra/bodhi/issues/1321>`_).
 * The bodhi CLI now supports editing an update using the update id.
   (`#937 <https://github.com/fedora-infra/bodhi/issues/937>`_)
+* The bodhi CLI now supports editing an override using the nvr.
```
bowlofeggs: I think we can leave off the "using the nvr" because it makes it sound like it was possible to edit overrides some other way before (and thus minimizes how great this change is!)

bodhi/tests/client/test_bindings.py
```diff
@@ -191,9 +191,9 @@ class TestBodhiClient_override_str(unittest.TestCase):
     """
```
bowlofeggs, line 191: Optional: It would be good to add two tests for the two new bindings API flag. We do have tests that hit that code through the CLI, but the bindings are their own interface that some third parties use too, so it would be good to test them independently. This is not required to merge.

bowlofeggs (changes requested):

bodhi/client/bindings.py
cverna: Hum it does not but we have the `--duration` in the CLI , which is apparently not used.  I guess I will remove this and I can create an issue to implement the server side to set the duration using the CLI.

bodhi/tests/client/test_bindings.py
cverna: I would rather submit another PR for that.

bodhi/client/bindings.py
bowlofeggs: The CLI ```--duration``` seems to be a convenience that is used to [calculate the ```expiration_date```](https://github.com/fedora-infra/bodhi/blob/2.5.0/bodhi/client/bindings.py#L288-L289) so I don't think we need to implement this server-side.

bodhi/client/__init__.py
bowlofeggs: Please add this space back to satisfy PEP-8.

bowlofeggs (approved):
This change looks good and seems to work well. Please add that one space back before merging!

PR #1425, opened 2017-04-10
Add a query property to the base model class

jeremycline:
This lets you perform queries using the model directly without
explicitly creating a session:

```
Package.query.first()
```

### DIFF
--- a/bodhi/server/models.py
+++ b/bodhi/server/models.py
@@ -39,7 +39,7 @@ from sqlalchemy.orm.properties import RelationshipProperty
 from sqlalchemy.sql import text
 from sqlalchemy.types import SchemaType, TypeDecorator, Enum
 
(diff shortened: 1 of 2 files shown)


bodhi/server/models.py
```diff
@@ -162,13 +162,25 @@ def drop(self, bind=None, checkfirst=False):
...
+        query (sqlalchemy.orm.query.Query): a class property which produces a
+            Query object against the class and the current Session when called.
+    """
+    __exclude_columns__ = ('id',)
+    __include_extras__ = tuple()
+    __get_by__ = ()
 
     id = Column(Integer, primary_key=True)
 
+    query = Session.query_property()
```
bowlofeggs, line 182: Whoah that was way easier than I expected. Thanks for the docblock too!

bowlofeggs (approved):
Looks awesome, I can't wait to start using it.

PR #1428, opened 2017-04-10
Optimize AJAX calls.

### LINKED ISSUE #1426: Further possible AJAX optimizations
@rajcze and @kparal have [suggested](https://github.com/fedora-infra/bodhi/pull/1379#issuecomment-292986803) some further optimizations to the AJAX code that loads results db data. The suggestion is to replace

```
var theurl = base_url+"results?testcases=dist.depcheck&item="+item+"&type=koji_build";
```

with

```
var theurl = base_url+"results?testcases=dist.depcheck&item="+item+"&type=koji_build&limit=1";
```

and to drop these three lines of code:

```
      if (data.next){
        request_results(data.next);
      }
```

nbianca:
Fixes #1426.

Basically, recursive calls will no longer be made.

### DIFF
--- a/bodhi/server/templates/update.html
+++ b/bodhi/server/templates/update.html
@@ -116,9 +116,6 @@ ${update.type} update for ${generateupdatetitlestring()}
     var latest = {};
     var activeResultsAjaxCalls = 0;
     var receive_resultsdb = function(data) {
-      if (data.next){
-        request_results(data.next);
-      }
       activeResultsAjaxCalls--;
 
       // First, prune duplicate results.  For instance, some depcheck runs
@@ -147,7 +144,7 @@ ${update.type} update for ${generateupdatetitlestring()}
           // for dist.depcheck. So if we hit one of these for the first time
           // we do a call to get the other one.
           if (name == 'dist.depcheck'){
-            var theurl = base_url+"results?testcases=dist.depcheck&item="+item+"&type=koji_build";
+            var theurl = base_url+"results?testcases=dist.depcheck&item="+item+"&type=koji_build&limit=1";
             if (arch == 'x86_64'){
               theurl = theurl +"&arch=i386"
             } else {


kparal:
Has it been tested? Please note we haven't tested this when suggesting the changes :) Ideally you'd compare the results before the patch and after the patch and it should be the same (one latest result for each testcase, and two results for depcheck - x86_64 and i386). For items with many depcheck results only the first page (with a single item) should be pulled.

bowlofeggs:
@nbianca, thanks for this patch! It looks like it's exactly what was suggested. 

@kparal and @rajcze, I did some *light* testing with this patch, and I'm not sure there is a clear difference actually. I only tested two URLs and I only tested them twice each (one with this patch, one wtihout). Note that this test isn't statistically significant since I only ran each path once - in order to get "real" data, I'd have to test the load times several times which I have not done here. Here is what I found:

The number of tests loaded seems the same with or without this patch (I just used the success/warning/failure count in the tab to make sure the numbers were the same). Note that I did *not* make sure the actual test data were identical in each test, just that the counts were the same.

With http://localhost:6543/updates/FEDORA-2017-9f83ba7048 I found that the page took 256s and 235 requests to load with this patch, and it look 124s and 238 requests to load without this patch.

With http://localhost:6543/updates/FEDORA-2017-6ec305fa93 I found that the page took 50.4s to load with this patch and 26.8s to load without the patch. I didn't count the number of requests that time (sorry).

So my statistically insignificant test shows this patch somehow being slower, with a slight reduction in request counts.

To really know if the performance is impacted, we are going to need more data of course (like loading the page n times each way and collecting mean/avg. or something). Even doing that, it's hard to know for sure that other factors aren't at play, like ResultsDB system load - maybe my test got unlucky and was competing with a bunch of other internet requests against the server. Unfortunately, I don't have time to do thorough testing ATM. It's actually pretty easy to load up a [Bodhi dev environment](https://bodhi.fedoraproject.org/docs/developer_docs.html#create-a-bodhi-development-environment) if you would like to check out this PR and compare it to the develop branch for yourselves.

What do you think?

jskladan:
Honestly, the time it took to load does not mean anything, especially if you tested the calls just once each- we have this huge IO bound problems in resultsdb right now, where the database data are dumped in and out of memory all the time. So the request you did "later" was (probably) with the database already in memory, thus returning the data basically instantly. I dare say that the "slower" times have no meaning at all, and the "faster" mostly say how long does the javascript take to render the data. Or it both means nothing - really the DB disk IO is the real enemy here.

But there is something fishy about the number of requests you are making. Taking https://bodhi.fedoraproject.org/updates/FEDORA-2017-9f83ba7048 as an example. There are 63 builds, so you should be doing 127 requests on ResultsDB for that page - 63*`/latest` + 63* "the other depcheck architecture" + 1 *dist.updates" request. The fact that you do almost double the amount of requests (if the "requests" in you comment are "requests to resutlsdb") is weird, and I'd say that there is something wrong there.

kparal:
The vangrant setup for Bodhi dev is very nice! Good work. 

http://localhost:6543/updates/FEDORA-2017-9f83ba7048
no patch: 190 requests to taskotron (64 /latest, 126 /results -- each item is requested for both arches for depcheck), ~34 sec
patched: 190 requests to taskotron (64 /latest, 126 /results -- each item is requested for both arches for depcheck, but limited to 1), ~19 sec

http://localhost:6543/updates/FEDORA-2017-6ec305fa93
no patch: 2 requests to taskotron (2 /latest), ~1 sec
patched: 2 requests to taskotron (2 /latest), ~1 sec

http://localhost:6543/updates/FEDORA-2017-5337793e7c
no patch: 6 requests to taskotron (2 /latest, 4 /results -- depcheck results are paginated), ~3 sec
patched: 4 requests to taskotron (2 /latest, 2 /results -- depcheck results are not paginated and limited to 1), ~2 sec

So overall I think the patch is working as expected. @bowlofeggs you probably used a different branch when measuring the numbers? 

The one problem I noticed is that depcheck results are always retrieved for both arches, instead of just pulling the one remaining arch (one is already pulled from /latest), so the code from #1379 doesn't seem to work right. That's a code we could further improve (as a separate patch).

bowlofeggs:
@rajcze To clarify, I was just using Firefox's developer console to count the total number of requests that happened when loading the page, so that would include the page itself, the CSS/fonts/JS and what not. So I didn't count the actual calls to ResultsDB, but I thought it was still good to show the relative difference in number of calls that the PR introduces. My apologies for not making that clear up front!

bowlofeggs:
@kparal @rajcze I'm good with merging this patch if you are. What do you think?

kparal:
> I didn't count the actual calls to ResultsDB

I also used FF's dev console, just used the filter box to filter all requests to "taskotron" :-)

> I'm good with merging this patch if you are. What do you think?

Yes, please do.

AdamWill:
Er, is this safe for openQA results? There are 19 openQA tests, and they can be re-run if they fail, and re-run manually for any reason. So you could well have 38 or more results from a single query there. I'm not sure it's safe to assume all the relevant results will always be on the first page, if I'm understanding the change right.

jskladan:
@AdamWill Why would this change be of any concern to OpenQA tests? Those are obtained here: https://github.com/nbianca/bodhi/blob/366f22de7b382f6e159cd3237bdf77cf7c3f8e6c/bodhi/server/templates/update.html#L245 and use `/latest` endpoint. `/latest` is not paginated, and with the recent changes, the only pagination left was "all the other depecheck results". This commit just gets rid of reading the (possibly) thousands of depcheck results, when the only one Bodhi cares about is the "latest result of depcheck for that other arch".

AdamWill:
Ah, OK. I didn't know that `/latest` isn't paginated.

AdamWill:
This seems to have gotten really slow again. See #2138 .

PR #1436, opened 2017-04-12
Rename the API docs to server docs and include a link to fedmsg docs

bowlofeggs:


docs/server_api/index.rst
```diff
@@ -1,5 +1,17 @@
-Bodhi REST API
-==============
+=================
+Bodhi Server APIs
+=================
+
+fedmsg API
+==========
+
+You can read about the fedmsgs that Bodhi sends here:
```
jeremycline: You could write this something like:
```
The fedmsgs bodhi sends `are documented in the fedora fedmsg project<https://fedora-fedmsg.readthedocs.io/en/latest/topics.html#bodhi>`_ 
```

Just a suggestion. You do what you think is right.

jeremycline (approved):
I approve, :shipit:

docs/server_api/index.rst
bowlofeggs: @jeremycline cool, I went with that.

PR #1437, opened 2017-04-12
Drop an unused inherited column from the builds table.

bowlofeggs:
While converting the Build model to support multiple types, I
noticed that there's a boolean column on the builds table called
inherited that is always False. In fact, the API did not present
this field and it seems that no code used it at all. Thus, this
commit removes the field from the models and provides a migration
to remove it in the database as well.


jeremycline (approved):
:shipit: looks good

Issue #1394, opened 2017-03-30
Split the Update model into two


bowlofeggs:
We need to split the Update model into two models, a superclass (probably still called Update) and a new RPM-specific subclass, perhaps simply called RPMUpdate. All of Update's relationships should be analyzed to determine which ones should be adjusted to reference RPM vs. continuing to reference Update. Also, a database migration will be needed so that Bodhi's existing data can be preserved in the new split Update design. All code will need to be altered to use the new RPM model rather than the Update model when the data involved is RPM-specific.

Once this is done, all of Bodhi's functionality should continue to work exactly as it does today.

jeremycline:
So I've taken a survey of the Update model (and written docs in #1445), and I think we can get away without having a special update type for each content type. An Update can know its type(s) based on the builds it's composed of, since builds have types.

There's a single catch I know of which is a problem we have to handle whether or not we have type-specific Update models. At the moment, bodhi has [a utility](https://github.com/fedora-infra/bodhi/blob/develop/bodhi/server/util.py#L199) that gets all the packages marked critical path, but this only applies to RPM (I think? The [pkgdb2-cli](https://pagure.io/pkgdb-cli/blob/master/f/pkgdb2client/__init__.py#_348) seems to hit [the critpath API](https://admin.fedoraproject.org/pkgdb/api/#critical_path_packages) and that makes no mention of namespaces) so that's probably something we need to address first).

bowlofeggs:
Your conclusion sounds reasonable to me. Should we use this ticket to add code to enforce that an ```Update``` only has one type of ```Build``` associated with it through a validator (or any other way that makes sense)? If we did something like that, would we need to also add a validator to the other end of the relationship (i.e., the ```Build```) so that the other end can't link to an ```Update``` that has other types? I presume that sqlalchemy validators don't validate from both ends if only defined on one end, but I could be wrong.

bowlofeggs:
As for the critical path packages, I think it would be reasonable for Bodhi to assume for now that only RpmPackages can be critpath. We can watch pkgdb's progress and when its API gains the ability to make other types be critical path we can revisit our code.

It would be good to file a Bodhi issue about this, so I'll do that now.

bowlofeggs:
I filed #1448 to track the critical path problem you mentioned here.

PR #1445, opened 2017-04-17
Document the Update model

jeremycline:
This documents the Update model in preparation for #1394.

### DIFF
--- a/bodhi/server/models.py
+++ b/bodhi/server/models.py
@@ -828,6 +828,76 @@ class RpmBuild(Build):
 
 
 class Update(Base):
+    """
+    This model represents an update.
+
+    Attributes:
+        title (unicode): The update's title which uniquely identifies the update.
+            This is generally an ordered list of the build NVRs contained in the
+            update.
+        autokarma (bool): A boolean that indicates whether or not the update will
+            be automatically pushed when the stable_karma threshold is reached.
+        stable_karma (int): A positive integer that indicates the amount of "good"
+            karma the update must receive before being automatically marked as stable.
+        unstable_karma (int): A positive integer that indicates the amount of "bad"
+            karma the update must receive before being automatically marked as unstable.
+        requirements (unicode): A list of taskotron tests that must pass for this
+            update to be considered stable.
+        require_bugs (bool): Indicates whether or not positive feedback needs to be
+            provided for the associated bugs before the update can be considered
+            stable.
+        require_testcases (bool): Indicates whether or not the update requires that
+            positive feedback be given on all associated wiki test cases before the
+            update can pass to stable. If the update has no associated wiki test cases,
+            this option has no effect.
+        notes (unicode): Notes about the update. This is a human-readable field that
+            describes what the update is for (e.g. the bugs it fixes).
+        type (EnumSymbol): The type of the update (e.g. enhancement, bugfix, etc). It
+            must be one of the values defined in :class:`UpdateType`.
+        status (EnumSymbol): The current status of the update. Possible values include
+            'pending' to indicate it is not yet in a repository, 'testing' to indicate it
+            is in the testing repository, etc. It must be one of the values defined in
+            :class:`UpdateStatus`.
+        request (EnumSymbol): The requested status of the update. This must be one of the
+            values defined in :class:`UpdateRequest` or ``None``.
+        severity (EnumSymbol): The update's severity. This must be one of the values defined
+            in :class:`UpdateSeverity`.
+        suggest (EnumSymbol): Suggested action a user should take after applying the update.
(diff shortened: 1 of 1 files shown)


bodhi/server/models.py
```diff
@@ -735,6 +735,75 @@ def unpush(self, koji):
 
 
 class Update(Base):
+    """
+    This model represents a package.
```
bowlofeggs: s/package/update/

bodhi/server/models.py
```diff
@@ -735,6 +735,75 @@ def unpush(self, koji):
...
+
+    This model uses single-table inheritance to allow for different package types.
```
bowlofeggs: We should probably drop this line.

bodhi/server/models.py
```diff
@@ -735,6 +735,75 @@ def unpush(self, koji):
...
+
+    Attributes:
+        title (unicode): The update's title which uniquely identifies the update.
```
bowlofeggs: Would it be good to document all the weirdness with this field, like that it is generally an ordered list of the build NVRs contained in the update? This is fine if you want to keep it simple.

bodhi/server/models.py
```diff
@@ -735,6 +735,75 @@ def unpush(self, koji):
...
+        autokarma (bool): A boolean that indicates whether or not the update will
+            be automatically pushed when the stable_karma threshold is reached.
+        stable_karma (int): A positive integer that indicates the amount of "good"
+            karma the update must receive before being marked as stable.
```
bowlofeggs: s/marked/automatically marked/ (The update can still be manually marked as stable after 7 days.)

bodhi/server/models.py
```diff
@@ -735,6 +735,75 @@ def unpush(self, koji):
...
+        unstable_karma (int): A positive integer that indicates the amount of "bad"
+            karma the update must receive before being marked as unstable.
```
bowlofeggs: automatically

bodhi/server/models.py
```diff
@@ -735,6 +735,75 @@ def unpush(self, koji):
...
+        requirements (unicode): A list of taskotron tests that must pass for this
+            update to be considered stable. Note: This does not appear to actually
+            work and is likely to be replaced by PolicyEngine in the long-term.
```
bowlofeggs: I'm not sure I'd put that note here just yet because I'm not sure if I used it correctly when I tried.

bodhi/server/models.py
```diff
@@ -735,6 +735,75 @@ def unpush(self, koji):
...
+        notes (unicode): Notes about the update. This is a human-readable field that
+            describes what the update is for (e.g. the bugs it fixes).
+        type (EnumSymbol): The type of the update (e.g. enhancement, bugfix, etc). It
+            must be one of the values defined in :class:`UpdateType`.
+        status (EnumSymbol): The current status of the update. Possible values include
+            'pending' to indicate it is not yet in a repository, 'testing' to indicate it
+            is in the testing repository, etc. It must be one of the values defined in
+            :class:`UpdateStatus`.
+        request (EnumSymbol): The requested status of the update. This must be one of the
+            values defined in :class:`UpdateRequest` or ``None``.
```
bowlofeggs, line 862: These two fields *almost* sound like a finite state machine…

bodhi/server/models.py
```diff
@@ -735,6 +735,75 @@ def unpush(self, koji):
...
+        severity (EnumSymbol): The update's severity. This must be one of the values defined
+            in :class:`UpdateSeverity`.
+        suggest (EnumSymbol): Suggested action a user should take after applying the update.
+            This must be one of the values defined in :class:`UpdateSuggestion`.
```
bowlofeggs, line 866: We should add an option that suggests they buy us a beer…

bodhi/server/models.py
```diff
@@ -735,6 +735,75 @@ def unpush(self, koji):
...
+        locked (bool): Indicates whether or not the update is locked and un-editable.
```
bowlofeggs, line 867: Maybe it would be good to mention that this usually happens when the update is going through a state transition by the masher?

bodhi/server/models.py
```diff
@@ -735,6 +735,75 @@ def unpush(self, koji):
...
+        pushed (bool): Indicates whether or not the update has been pushed to its requested
+            repository.
```
bowlofeggs: Huh, I didn't know about this one.

bowlofeggs (approved):

bodhi/server/models.py
jeremycline: ✨

PR #1451, opened 2017-04-18
Add a validator to Update that enforces Build types

### LINKED ISSUE #1394: Split the Update model into two

jeremycline:
This adds a validator to the Update model that raises an exception if an
update is composed of builds of different types.

fixes #1394

### DIFF
--- a/bodhi/tests/server/test_models.py
+++ b/bodhi/tests/server/test_models.py
@@ -209,6 +209,106 @@ class TestRpmBuild(ModelTest):
         eq_(self.obj.get_url(), u'/TurboGears-1.0.8-3.fc11')
 
 
+class TestUpdateValidateBuilds(BaseTestCase):
+    """Tests for the :class:`Update` validator for builds."""
+
+    def setUp(self):
+        super(TestUpdateValidateBuilds, self).setUp()
+        self.package = model.Package(name='the-greatest-package')
+        self.update = model.Update(
+            title='The best update of all time.',
+            user=model.User.query.filter_by(name=u'guest').one(),
+            request=model.UpdateRequest.testing,
+            notes=u'Useless details!',
+            release=model.Release.query.filter_by(name=u'F17').one(),
+            date_submitted=datetime(1984, 11, 02),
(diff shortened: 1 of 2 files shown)


bodhi/tests/server/test_models.py
```diff
@@ -216,6 +216,106 @@ def test_url(self):
...
+        )
+        build2 = model.RpmBuild(
+            nvr='the-greatest-package-1.1.0-fc17.1',
+            package_id=self.package.id,
+            release_id=self.update.release.id,
+        )
+        self.update.builds += [build1, build2]
+
+    def test_different_build_types(self):
+        """Assert when all builds are a different type, validation passes."""
```
bowlofeggs: s/passes/fails/

bodhi/tests/server/test_models.py
```diff
@@ -216,6 +216,106 @@ def test_url(self):
...
+        build2 = model.RpmBuild(
+            nvr='the-greatest-package-1.1.0-fc17.1',
+            package_id=self.package.id,
+            release_id=self.update.release.id,
+        )
+        build1.update = self.update
+        with self.assertRaises(ValueError) as cm:
+            build2.update = self.update
+
+        self.assertEqual(str(cm.exception), u'An update must contain builds of the same type.')
```
bowlofeggs, line 309: When looking at the code I was allllmost going to add a comment asking if the backrefs still get validated, but then I decided to go see what tests you wrote and here it is. Bam.

bowlofeggs (changes requested):
Just one comment to fix, but LGTM otherwise. Nice!

bowlofeggs (approved):

PR #1452, opened 2017-04-18
Document the BuildrootOverride module

### LINKED ISSUE #1395: Split the BuildrootOverride model into two, if needed
We may need to split the BuiltrootOverride model into two models, a superclass (probably still called BuildrootOverride) and a new RPM-specific subclass, perhaps simply called RPMBuildrootOverride.

Whether we do this depends on whether modules need BuildrootOverrides, as I do not believe containers or flatpaks need them.

If we do need them, all of BuildrootOverride's relationships should be analyzed to determine which ones should be adjusted to reference RPM vs. continuing to reference BuildrootOverride. Also, a database migration will be needed so that Bodhi's existing data can be preserved in the new split BuildrootOverride design. All code will need to be altered to use the new RPM model rather than the BuildrootOverride model when the data involved is RPM-specific.

jeremycline:
fixes #1395

### DIFF
--- a/bodhi/server/models.py
+++ b/bodhi/server/models.py
@@ -2496,6 +2496,24 @@ class Group(Base):
 
 
 class BuildrootOverride(Base):
+    """
+    This model represents a Koji buildroot override.
+
+    A buildroot override is a way to add a build to the Koji buildroot (the set of packages used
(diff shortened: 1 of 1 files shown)


bowlofeggs (approved):
Very nice!

PR #1458, opened 2017-04-20
Add a uniqueness constraint on (name, type) for Packages

jeremycline:
Packages of different types may have overlapping names. For example,
there is a python-requests RPM and a python-requests module. This adds a
uniqueness constraint on the combination of name and type and removes
the uniqueness constraint on name alone.


bowlofeggs (approved):
Nice!

PR #1460, opened 2017-04-21
add a new command to the vagrant setup to run the style tests

ryanlerch:
This adds a new alias in the .bashrc that we use in the default
vagrant setup that allows just the flake8 style tests to be run


cverna:
LGTM, this will be useful for me 😄

bowlofeggs (approved):
Thanks!

PR #1462, opened 2017-04-21
Add a test to ensure there is only one alembic head.

bowlofeggs:


pypingou:
I think I saw this test somewhere before, but looks fine to me :) 👍

PR #1465, opened 2017-04-23
Makes search case-insensitive.

### LINKED ISSUE #997: search should be case-insensitive

nbianca:
Fixes #997.

### DIFF
--- a/bodhi/server/services/updates.py
+++ b/bodhi/server/services/updates.py
@@ -217,7 +217,7 @@ def query_updates(request):
     packages = data.get('packages')
     if packages is not None:
         query = query.join(Update.builds).join(Build.package)
-        query = query.filter(or_(*[RpmPackage.name == pkg for pkg in packages]))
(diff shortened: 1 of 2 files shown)


pypingou:
Looks fine to me, do we want to add some tests for this?

bowlofeggs (changes requested):
This doesn't seem to address the example search given in #997. This URL doesn't return any results:

http://localhost:6543/updates/?packages=wxgtx3

Whereas this one does:

http://localhost:6543/updates/?packages=wxGTK3

Additionally, the search box at the top of the page also searches updates, usernames, and buildroot overrides in addition to packages. It would be good to make those case-insensitive as well.

bowlofeggs:
Do you still wish to work on this pull request?

bowlofeggs:
Closing due to inactivity.

PR #1466, opened 2017-04-23
Remove noisy docblocks that end up in --help text.

### LINKED ISSUE #1457: The CLI docblocks end up in the --help text and are noisy
I noticed today that the CLI docblocks show up in the ```--help``` output, including argument types. This text isn't particularly useful for users and it is noisy and badly formatted. We should probably just not document the parameters for the CLI functions since they already get documented through the ```click options```. For example,

```
$ bodhi updates comment --help
Usage: bodhi updates comment [OPTIONS] UPDATE TEXT

  Comment on an update.

  Args:     update (unicode): The update you wish to modify.     text
  (unicode): The text of the comment you wish to leave on the update.
  karma (int): The karma you wish to leave on the update. Must be +1, 0, or
  -1.     user (unicode): The username to authenticate as.     password
  (unicode): The user's password.     staging (bool): Whether to use the
  staging server or not.     url (unicode): The URL of a Bodhi server to
  create the update on. Ignored if staging is                    True.

Options:
  --karma INTEGER  The karma for this comment (+1/0/-1)
  --user TEXT
  --password TEXT
  --staging        Use the staging bodhi instance
  --url TEXT       URL of a Bodhi server. Ignored if --staging is set. Can be
                   set with BODHI_URL environment variable
  --help           Show this message and exit.
```

To fix this, let's drop all the documentation on function arguments from the docblocks. We can add expanded descriptions in their place if desirable.

nbianca:
Fixes #1457.

### DIFF
--- a/bodhi/client/__init__.py
+++ b/bodhi/client/__init__.py
@@ -144,16 +144,6 @@ def updates():
 @click.option('--file', help='A text file containing all the update details')
 @url_option
 def new(user, password, url, **kwargs):
-    """
-    Create a new update.
-
-    Args:
-        user (unicode): The username to authenticate as.
-        password (unicode): The user's password.
-        url (unicode): The URL of a Bodhi server to create the update on. Ignored if staging is
-                       True.
-        kwargs (dict): Other keyword arguments passed to us by click.
-    """
     client = bindings.BodhiClient(base_url=url, username=user, password=password,
                                   staging=kwargs['staging'])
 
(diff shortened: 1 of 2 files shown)


pypingou:
Not entirely sure @bowlofeggs will agree with this :)

bodhi/client/__init__.py
```diff
@@ -144,16 +144,6 @@ def updates():
 @click.option('--file', help='A text file containing all the update details')
 @url_option
 def new(user, password, url, **kwargs):
-    """
-    Create a new update.
```
bowlofeggs, line 148: We do want to keep the descriptions - it's just the Args we don't want. The descriptions are useful in the help text.

bowlofeggs (changes requested):
Let's make sure the human readable descriptions remain, and just drop the Args blocks.

bowlofeggs:
@pypingou I actually requested a change like this in #1457 because the Args aren't "user consumable" and actually kind of render badly too. I didn't realize that click would render those when they were written. Normally I insist on documenting parameters so that developers have good docs, but in the rare case where developer docs are at odds with good user docs, I'd rather give the users good docs ☺

bowlofeggs:
Do you still wish to work on this pull request?

bowlofeggs:
Closing due to inactivity.

PR #1476, opened 2017-04-25
Improve logging in the notification module

jeremycline:

### DIFF
--- a/bodhi/server/notifications.py
+++ b/bodhi/server/notifications.py
@@ -65,16 +65,19 @@ def send_fedmsgs_after_commit(session):
         session (sqlalchemy.orm.session.Session): The session that was committed.
     """
     if 'fedmsg' in session.info:
+        _log.debug('Emitting all queued fedmsgs for %r', session)
         # Initialize right before we try to publish, but only if we haven't
         # initialized for this thread already.
         if not fedmsg_is_initialized():
             init()
 
         for topic, messages in session.info['fedmsg'].items():
-            _log.info('emitting {n} fedmsgs to the "{topic}" topic.'.format(
-                n=len(messages), topic=topic))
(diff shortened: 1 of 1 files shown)


pypingou:
Looks fine to me 👍

bodhi/server/notifications.py
```diff
@@ -64,16 +64,19 @@ def send_fedmsgs_after_commit(session):
         session (sqlalchemy.orm.session.Session): The session that was committed.
     """
     if 'fedmsg' in session.info:
+        _log.info('Emitting all queued fedmsgs for %r', session)
```
bowlofeggs: I wonder if we should make this debug? This could make the logs pretty noisy since so many actions in Bodhi generate fedmsgs.

bodhi/server/notifications.py
```diff
@@ -64,16 +64,19 @@ def send_fedmsgs_after_commit(session):
...
         # Initialize right before we try to publish, but only if we haven't
         # initialized for this thread already.
         if not fedmsg_is_initialized():
             init()
 
         for topic, messages in session.info['fedmsg'].items():
-            _log.info('emitting {n} fedmsgs to the "{topic}" topic.'.format(
-                n=len(messages), topic=topic))
+            _log.info('emitting %d fedmsgs to the "%s" topic queued by %r',
+                      len(messages), topic, session)
```
bowlofeggs: This one might make sense to be debug as well.

bowlofeggs (approved):
You are welcome to merge as-is, but I have two thoughts to consider. Take them or leave them.

jeremycline:
I dropped both down to debugging. I don't think I realized when I wrote this that comments emitted fedmsgs.

PR #1477, opened 2017-04-25
Remove the __init__() method from TransactionalSessionMaker.

bowlofeggs:
The TransactionalSessionMaker used to have an __init__() method so
that the database engine could be passed in. However, it was
recently modified to use the threadlocal Session object and no
longer used self.engine. This commit drops the no longer necessary
code and fixes all calling code.


pypingou:
Looks fine to me

PR #1478, opened 2017-04-25
If rolling back a db transaction fails, log and raise original exc.

### LINKED ISSUE #1475: The TransactionalSessionMaker will eat error messages if session.rollback() fails
It is possible for the call to [```session.rollback()```](https://github.com/fedora-infra/bodhi/blob/2.6.0/bodhi/server/util.py#L659) in ```bodhi.server.util.TransactionalSessionMaker``` to raise an ```Exception```. When this happens, the call to ```raise``` is not hit and we lose the information about the original ```Exception``` which makes debugging difficult.

We should probably wrap the call to ```rollback``` in another ```try/except``` block with its own log statement so we can know about both the ```rollback``` issue and the original issue.

bowlofeggs:
fixes #1475.

### DIFF
--- a/bodhi/server/util.py
+++ b/bodhi/server/util.py
@@ -655,9 +655,15 @@ class TransactionalSessionMaker(object):
         try:
             yield session
             session.commit()
-        except:
-            session.rollback()
-            raise
+        except Exception as e:
+            # It is possible for session.rolback() to raise Exceptions, so we will wrap it in an
+            # Exception handler as well so we can log the rollback failure and still raise the
+            # original Exception.
+            try:
+                session.rollback()
+            except:
+                log.exception('An Exception was raised while rolling back a transaction.')
+            raise e
         finally:
             session.close()
             Session.remove()
--- a/bodhi/tests/server/test_utils.py
+++ b/bodhi/tests/server/test_utils.py
@@ -18,7 +18,7 @@ import subprocess
 from bodhi.server.buildsys import setup_buildsystem, teardown_buildsystem
 from bodhi.server.config import config
 from bodhi.server.util import (get_critpath_pkgs, get_nvr, markup,
-                               get_rpm_header, cmd, sorted_builds)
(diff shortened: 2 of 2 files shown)


pypingou:
Looks fine to me

bowlofeggs:
@pypingou In full disclosure, I had to further modify this patch before merging, due to https://github.com/fedora-infra/bodhi/pull/1477:

```
git diff
diff --git a/bodhi/tests/server/test_utils.py b/bodhi/tests/server/test_utils.py
index 242d866..7feb12e 100644
--- a/bodhi/tests/server/test_utils.py
+++ b/bodhi/tests/server/test_utils.py
@@ -143,7 +143,7 @@ class TestTransactionalSessionMaker(unittest.TestCase):
         __call__() should log the failure to roll back, and then close and remove the Session, and
         should raise the original Exception again.
         """
-        tsm = TransactionalSessionMaker(mock.MagicMock())
+        tsm = TransactionalSessionMaker()
         exception = ValueError("u can't do that lol")
         # Now let's make it super bad by having rollback raise an Exception
         Session.return_value.rollback.side_effect = IOError("lol now u can't connect to the db")
@@ -169,7 +169,7 @@ class TestTransactionalSessionMaker(unittest.TestCase):
         If the wrapped code raises an Exception, __call__() should roll back the transaction, and
         close and remove the Session, and should raise the original Exception again.
         """
-        tsm = TransactionalSessionMaker(mock.MagicMock())
+        tsm = TransactionalSessionMaker()
         exception = ValueError("u can't do that lol")
 
         with self.assertRaises(ValueError) as exc_context:
@@ -192,7 +192,7 @@ class TestTransactionalSessionMaker(unittest.TestCase):
         __call__() should commit the transaction, and close and remove the Session upon a successful
         operation.
         """
-        tsm = TransactionalSessionMaker(mock.MagicMock())
+        tsm = TransactionalSessionMaker()
 
         with tsm():
             pass
```

PR #1480, opened 2017-04-25
Setup fedmsg so published messages are viewable

jeremycline:
This sets up a fedmsg relay which can be tailed to see what messages
Bodhi is publishing. A service, fedmsg-tail, is enabled and started in
Vagrant. This causes all fedmsgs sent by Bodhi to be logged to the
systemd journal.

### DIFF
--- a/devel/ansible/roles/dev/files/fedmsg-tail.service
+++ b/devel/ansible/roles/dev/files/fedmsg-tail.service
@@ -1,0 +1,11 @@
+[Unit]
+Description=fedmsg tail
+After=network-online.target
+Wants=network-online.target
+
+[Service]
+User=vagrant
+ExecStart=/usr/bin/fedmsg-tail
+
+[Install]
+WantedBy=multi-user.target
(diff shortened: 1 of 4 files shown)


bowlofeggs (approved):
fedmsgs with this change aren't being displayed as-is, but it's actually just because this isn't merged:

https://github.com/fedora-infra/bodhi/pull/1484

So I'm +1 to going ahead and merging it. (I did an offline test of rebasing your commit on mine and confirmed that the fedmsgs are working so this commit actually is working correctly - bodhi just isn't sending fedmsgs due to that other issue.)

I do have a suggestion: You could make a bfedmsg-tail in the bashrc that does one of the following:

1. Runs sudo journalctl -u fedmsg-tail
2. Runs fedmsg-tail, maybe with --really-pretty or something.

Just an idea, feel free to ignore.

bowlofeggs:
Let's make sure we rebase this commit on whichever commit makes it OK to set ```fedmsg_enabled = True``` in the ini file before we merge this one.

bowlofeggs:
#1484 is merged now, so if you rebase on the develop branch this should work with ```fedmsg_enabled = True```.

Issue #1472, opened 2017-04-25
Transactional failure during masher


bowlofeggs:
Message
-------
[2017-04-25 02:27:57][    fedmsg   ERROR]
```python
MasherThread failed. Transaction rolled back.
```

Process Details
---------------
- host:     bodhi-backend01.phx2.fedoraproject.org
- PID:      31792
- name:     fedmsg-hub
- command:  /usr/bin/python /usr/bin/fedmsg-hub
- msg_id:   

Callstack that lead to the logging statement
--------------------------------------------
```python
  File "/usr/lib64/python2.7/threading.py", line 777 in __bootstrap
    self.__bootstrap_inner()
  File "/usr/lib64/python2.7/threading.py", line 804 in __bootstrap_inner
    self.run()
  File "/usr/lib/python2.7/site-packages/bodhi/server/consumers/masher.py", line 263 in run
    self.log.exception('MasherThread failed. Transaction rolled back.')
```

```
Traceback (most recent call last):
  File "/usr/lib/python2.7/site-packages/bodhi/server/consumers/masher.py", line 261, in run
    self.db = None
  File "/usr/lib64/python2.7/contextlib.py", line 24, in __exit__
    self.gen.next()
  File "/usr/lib/python2.7/site-packages/bodhi/server/util.py", line 659, in __call__
    session.rollback()
  File "/usr/lib64/python2.7/site-packages/sqlalchemy/orm/session.py", line 765, in rollback
    self.transaction.rollback()
  File "/usr/lib64/python2.7/site-packages/sqlalchemy/orm/session.py", line 408, in rollback
    self._assert_active(prepared_ok=True, rollback_ok=True)
  File "/usr/lib64/python2.7/site-packages/sqlalchemy/orm/session.py", line 196, in _assert_active
    "This session is in 'committed' state; no further "
InvalidRequestError: This session is in 'committed' state; no further SQL can be emitted within this transaction.
```

bowlofeggs:
At this point, it seems likely that this is the same underlying issue as #1470, and I believe the fix I've got for it will fix this too.

bowlofeggs:
My previous comment was incorrect, but @jeremycline and I figured out how to fix this and he will make it so. Thanks Jeremy!

bowlofeggs:
This e-mail reveals the true problem:

Message
-------
[2017-04-27 19:54:51][bodhi.server   ERROR]
```python
An Exception was raised during a transactional session.
```


Process Details
---------------
[...]

PR #1483, opened 2017-04-27
Turn all fedmsgs to dictionaries prior to queuing

jeremycline:
This is required because at post-commit time, the session is complete
and it's not possible to emit SQL, but when referencing the model
objects SQL must be emitted to refresh the attributes.


bowlofeggs:
fixes #1472

bowlofeggs (approved):
This is good to go, except please rebase it on the commit that makes it OK to have ```fedmsg_enabled = True``` before merging.

bowlofeggs:
#1484 is merged to develop now so if you rebase this ```fedmsg_enabled = True``` should work.

PR #1487, opened 2017-05-01
Edit updates with the update title

### LINKED ISSUE #1409: The CLI edit command could also accept update titles to identify which update is being edited
Right now the CLI accepts the update id as a parameter to identify which update is being edited.

We could make [a couple regexes used in the bindings](https://github.com/fedora-infra/bodhi/blob/develop/bodhi/client/bindings.py#L238-L241) (or even a function that does what the cited code block does) available at the bindings module level so the client can share the code:

```
if re.search(r'(\.el|\.fc)\d\d?', kwargs['package']):
    kwargs['builds'] = kwargs['package']
elif re.search(r'FEDORA-(EPEL-)?\d{4,4}', kwargs['package']):
    kwargs['updateid'] = kwargs['package']
```

If we did this, then users could type ```bodhi updates edit --user bowlofeggs --notes "check it out" FEDORA-2017-4917bb3d50``` or ```bodhi updates edit --user bowlofeggs --notes "check it out" bodhi-2.5.0-1.fc26``` and the two commands would edit the same update.

cverna:
This commit makes possible to edit an update using the update title
Fixes #1409

### DIFF
--- a/bodhi/client/__init__.py
+++ b/bodhi/client/__init__.py
@@ -17,6 +17,7 @@ import platform
 import subprocess
 import sys
 import traceback
+import re
 
 import click
 
@@ -175,13 +176,26 @@ def new(user, password, url, **kwargs):
             traceback.print_exc()
 
 
+def _validate_edit_update(ctx, param, value):
+    """
+    Callback used by click to validate the update argument given to the updates edit command.
+    the update argument can only be update id or update title
+    """
+    if re.search(bindings.UPDATE_ID_RE, value)\
+       or re.search(bindings.UPDATE_TITLE_RE, value):
+        return value
+    else:
+        raise click.BadParameter("Please provide an Update ID or an Update Title")
+
+
 @updates.command()
 @add_options(new_edit_options)
-@click.argument('updateid')
+@click.argument('update', callback=_validate_edit_update)
 @url_option
 def edit(user, password, url, **kwargs):
     """
     Edit an existing update.
+    The update argument can be an update id or the update title.
 
     Args:
         user (unicode): The username to authenticate as.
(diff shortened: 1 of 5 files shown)


bodhi/tests/client/test___init__.py
```diff
@@ -449,6 +449,68 @@ def test_notes_and_notes_file(self):
...
+        result = runner.invoke(
+            client.edit, ['FEDORA-20-cc8582d738', '--user', 'bowlofeggs',
+                          '--password', 's3kr3t', '--notes', 'this is an edited note',
+                          '--url', 'http://localhost:6543'])
+        self.assertEqual(result.exit_code, 2)
+        expected = u'Usage: edit [OPTIONS] UPDATE\n\n' \
+                   u'Error: Invalid value for "update": ' \
+                   u'Please provide an Update ID or an Update Title\n'
+
+        self.assertEqual(result.output, expected)
```
bowlofeggs, line 512: I think we probably only need one of the two tests above, since they are both asserting that identifiers that don't match the regexes raise an error and the only difference is which wrong identifier we try. That said, I'm not opposed to keeping both tests if you prefer - your choice ☺

bodhi/client/__init__.py
```diff
@@ -175,9 +176,21 @@ def new(user, password, url, **kwargs):
...
+        raise click.BadParameter("Please provide an Update ID or an Update Title")
+
+
 @updates.command()
 @add_options(new_edit_options)
-@click.argument('updateid')
+@click.argument('update', callback=_validate_edit_update)
 @url_option
 def edit(user, password, url, **kwargs):
     """
```
bowlofeggs, line 196: You could mention here that ```UPDATE``` can be the update alias (FEDORA-blah) or the title. I think that would cause ```--help``` to display that info to the user. Optional.

bowlofeggs (approved):
Looks great! Just a couple of suggestions - you can merge as-is, or you can make the suggested changes and merge (no need for re-review). Nice!

cverna:
I have added the doc string to mention than update can be an update id or title. I have also added the change to 2.7.0 release notes.

And thanks for the review .

PR #1490, opened 2017-05-01
Add an alias for fedmsg and pretty-print messages

jeremycline:
This makes inspecting fedmsgs easier in bodhi by adding an alias and
pretty-printing the messages so they're easy to read (although not
terribly compact).

### DIFF
--- a/devel/ansible/roles/dev/files/.bashrc
+++ b/devel/ansible/roles/dev/files/.bashrc
@@ -16,6 +16,7 @@ alias bshell="pshell /home/vagrant/bodhi/development.ini"
 alias bstart="sudo systemctl start bodhi && echo 'The Application is running on http://localhost:6543'"
 alias bstop="sudo systemctl stop bodhi"
 alias bteststyle="pushd /home/vagrant/bodhi && nosetests -sx ~/bodhi/bodhi/tests/test_style.py; popd"
+alias bfedmsg="sudo journalctl -u fedmsg-tail"
 
 function btest {
     find /home/vagrant/bodhi -name "*.pyc" -delete;
--- a/devel/ansible/roles/dev/files/fedmsg-tail.service
+++ b/devel/ansible/roles/dev/files/fedmsg-tail.service
@@ -5,7 +5,7 @@ Wants=network-online.target
 
 [Service]
 User=vagrant
-ExecStart=/usr/bin/fedmsg-tail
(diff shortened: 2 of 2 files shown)


devel/ansible/roles/dev/files/.bashrc
```diff
@@ -15,6 +15,7 @@ alias brestart="sudo systemctl restart bodhi && echo 'The Application is running
 alias bstart="sudo systemctl start bodhi && echo 'The Application is running on http://localhost:6543'"
 alias bstop="sudo systemctl stop bodhi"
 alias bteststyle="pushd /home/vagrant/bodhi && nosetests -sx ~/bodhi/bodhi/tests/test_style.py; popd"
+alias bfedmsg="sudo journalctl -u fedmsg-tail -f"
```
bowlofeggs: If you leave the ```-f``` off, it still allows the user to pass the ```-f```, but also allows them to see the history. This is what I did with ```blog```.

devel/ansible/roles/dev/files/motd
```diff
@@ -8,6 +8,7 @@ bstart:     Start the Bodhi service.
 bstop:      Stop the Bodhi service.
 btest:      Run Bodhi's test suite.
 bteststyle: Run just the flake8 code style tests.
+bfedmsg:    View all fedmsgs Bodhi has emitted.
```
bowlofeggs: Perhaps it is dumb that this is double documented, but there is also a section in the docs that talks about these commands:

https://bodhi.fedoraproject.org/docs/developer_docs.html#quick-tips-about-the-bodhi-vagrant-environment

You think we should delete the double docs, or should we add this to it?

bowlofeggs (approved):
I suggest dropping the ```-f```, but it's just a suggestion.

devel/ansible/roles/dev/files/motd
bowlofeggs: I think it's best to delete the double docs, so I filed a PR for that:

https://github.com/fedora-infra/bodhi/pull/1527

devel/ansible/roles/dev/files/.bashrc
jeremycline: Done!

devel/ansible/roles/dev/files/motd
jeremycline: Also done.

PR #1494, opened 2017-05-02
Acquire the global DDL lock if BDR is enabled

jeremycline:
This should allow users to run Alembic in "online" mode with a Postgres
BDR system. It will also emit the correct SQL for offline migrations.

It introduces a new boolean setting in alembic.ini, "bdr", which
defaults to false.

### DIFF
--- a/alembic.ini
+++ b/alembic.ini
@@ -14,9 +14,13 @@ script_location = alembic
 # Don't bother, this is obtained from the Bodhi config file
 sqlalchemy.url = sqlite://bodhi.db
 
-# Set to true to aquire the global DDL lock for BDR
-# See http://bdr-project.org/docs/stable/ddl-replication-advice.html
-bdr = false
+# Set to true to acquire the global DDL lock for BDR for offline SQL scripts.
+# BDR is automatically detected during online migrations and this setting is
+# ignored.
+#
+# See http://bdr-project.org/docs/stable/ddl-replication-advice.html for more
+# information about Postgres-BDR.
+offline_postgres_bdr = false
 
 
 # Logging configuration
--- a/alembic.ini
+++ b/alembic.ini
@@ -14,6 +14,10 @@ script_location = alembic
 # Don't bother, this is obtained from the Bodhi config file
 sqlalchemy.url = sqlite://bodhi.db
 
+# Set to true to aquire the global DDL lock for BDR
+# See http://bdr-project.org/docs/stable/ddl-replication-advice.html
+bdr = false
+
 
 # Logging configuration
 [loggers]
--- a/alembic/env.py
+++ b/alembic/env.py
@@ -2,7 +2,7 @@ from __future__ import with_statement
 from logging.config import fileConfig
 
 from alembic import context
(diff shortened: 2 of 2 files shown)


bowlofeggs:
I tried this on staging today and it worked, nice!

alembic/env.py
```diff
@@ -43,6 +43,8 @@ def run_migrations_offline():
     context.configure(url=url)
 
     with context.begin_transaction():
+        if config.get_main_option('bdr').strip().lower() == 'true':
```
bowlofeggs: It might be nice to make it possible for this option to not be present and default to False so upgrading users don't have to add this to their configs.

alembic/env.py
```diff
@@ -58,6 +60,12 @@ def run_migrations_online():
         prefix='sqlalchemy.',
         poolclass=pool.NullPool)
 
+    if config.get_main_option('bdr').strip().lower() == 'true':
```
bowlofeggs: Same here.

bowlofeggs (approved):

mizdebsk:
Wouldn't it be better not to require configuration option for this? BDR can be autodetected and `bdr.permit_ddl_locking` set only when running with BDR. For example, see [how this was achieved in Koschei](https://github.com/msimacek/koschei/commit/938605ea14372ac0ecbd266f38b6307e9f79ee11)

bowlofeggs:
@mizdebsk Oh very nice, I like that suggestion. @jeremycline, what do you think?

jeremycline:
Oh, yes, that's very nice. @mizdebsk++, thanks for the tip.

jeremycline:
@bowlofeggs Updated. I left some of what I had before to support the offline migration case, but it's no longer a required setting. I also added a few log statements so it's clear when it permits global DDL locks.

bowlofeggs:
On Thu, 2017-05-18 at 11:24 -0700, Jeremy Cline wrote:
> I left some of what I had before to support the offline migration
> case

Oh, good idea. That's officer thinking!

alembic.ini
```diff
@@ -14,6 +14,14 @@ script_location = alembic
 # Don't bother, this is obtained from the Bodhi config file
 sqlalchemy.url = sqlite://bodhi.db
 
+# Set to true to acquire the global DDL lock for BDR for offline SQL scripts.
+# BDR is automatically detected during online migrations and this setting is
+# ignored.
+#
+# See http://bdr-project.org/docs/stable/ddl-replication-advice.html for more
+# information about Postgres-BDR.
+offline_postgres_bdr = false
```
bowlofeggs, line 23: Nice.

bowlofeggs (approved):
This looks great. I think we can go ahead and merge it, even though we technically haven't tried it. We'll know with 2.8.0 beta if it works or not when we deploy to staging, and if there's a problem we can fix it before the release.

bowlofeggs:
This has been cherry-picked to the 2.7 branch as 05ab85a124c916e0f5d4eacb8b873335ffdd624e.

PR #1495, opened 2017-05-02
Add new ModuleBuild and ModulePackage models.

### LINKED ISSUE #1324: AddModuleModels
In order to support modules, we need to add new database models. Before we can do this, we need to modify Bodhi's existing models to be multi-type friendly. Thus, the following modules must be converted before this issue can be worked on:

- [x] Package - #1392
- [x] Build - #1393
- [x] Update - #1394
- [x] BuildrootOverride - #1395

Each of these models will need to be reworked to handle multiple types by splitting them into superclass/subclass (with the subclass being the RPM-specific bits of the current model), and database migrations will be needed. When we are done with this phase, Bodhi should still work like it does today with RPMs.

Once each of these models are able to support multiple types, we will next be able to add more models to satisfy this ticket:

- [x] Module
- [x] ModuleBuild

### LINKED ISSUE #1459: We need to validate that RpmBuilds are only associated with RpmPackages and vice-versa
This is also true for ModuleBuilds and ModulePackages, though we can solve that one in #1324.

bowlofeggs:
fixes #1324
fixes #1459

### DIFF
--- a/bodhi/server/models.py
+++ b/bodhi/server/models.py
@@ -76,10 +76,10 @@ class EnumSymbol(object):
         return "<%s>" % self.name
 
     def __unicode__(self):
-        return unicode(self.description)
+        return unicode(self.value)
 
     def __json__(self, request=None):
-        return self.description
+        return self.value
 
 
 class EnumMeta(type):
@@ -307,6 +307,11 @@ metadata = Base.metadata
 ##
 #  Enumerated type declarations
 ##
+class ContentType(DeclEnum):
+    base = 'base', 'Base'
+    rpm = 'rpm', 'RPM'
+    module = 'module', 'Module'
+
 
 class UpdateStatus(DeclEnum):
     pending = 'pending', 'pending'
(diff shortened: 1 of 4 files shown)


bodhi/server/models.py
```diff
@@ -729,6 +765,20 @@ def unpush(self, koji):
...
+    """
+    Represents a Module build.
+
+    Note that this model uses single-table inheritance with its Build superclass.
+
+    Attributes:
+        nvr (unicode): A unique Koji identifier for the module build.
+    """
+    __mapper_args__ = {
+        'polymorphic_identity': ContentType.module,
```
ralphbean, line 778: Fancy!

ralphbean (approved):
You, sir, are a gentleman and a scholar.

bodhi/tests/server/test_models.py
```diff
@@ -90,7 +90,7 @@ class TestQueryProperty(BaseTestCase):
 
     def test_session(self):
         """Assert the session the query property uses is from the scoped session."""
-        query = model.Package.query
+        query = model.RpmPackage.query
```
jeremycline: Out of curiosity, why did this need to be changed?

jeremycline (approved):
Very nice, :shipit:

ralphbean:
OK to merge?

bodhi/tests/server/test_models.py
bowlofeggs: @jeremycline I'm not sure I can recall why it needed to be changed. Presumably some test was failing at some point, but when I change it back the tests all pass so I'll set it back. Thanks for pointing that out!

bowlofeggs:
@ralphbean Yeah I'm working on merging this today, then we can move on to looking at the API.

PR #1497, opened 2017-05-03
Improved vagrant example file

pypingou:


Vagrantfile.example
```diff
@@ -30,14 +28,20 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
  # Vagrant can share the source directory using rsync, NFS, or SSHFS (with the vagrant-sshfs
  # plugin). Consult the Vagrant documentation if you do not want to use SSHFS.
  config.vm.synced_folder ".", "/vagrant", disabled: true
- config.vm.synced_folder ".", "/home/vagrant/bodhi", type: "sshfs", sshfs_opts_append: "-o nonempty"
+ config.vm.synced_folder ".", "/home/vagrant/bodhi",
+   ssh_opts_append: "-o IdentitiesOnly=true",
+   sshfs_opts_append: "-o nonempty",
+   type: "sshfs"
```
bowlofeggs, line 34: I remember having some issues when we were looking at this last week. Does this PR have what you came up with, or will it still cause issues?

pypingou: No finally to get vagrant to work I have to run before vagrant up: ``export SSH_AUTH_SOCK=˙``

So part of the changes here can be discarded

bowlofeggs: @pypingou Do you want to close this PR, or do you think it should still be modified?

pypingou:
Nah let's close

PR #1498, opened 2017-05-03
Fix a couple of warnings thrown when running the unit-tests

pypingou:


bowlofeggs (approved):
Thanks!

Issue #1489, opened 2017-05-01
config.py should do more for us


bowlofeggs:
Currently, Bodhi has config processing code scattered all over the place. [For example](https://github.com/fedora-infra/bodhi/blob/2.6.0/bodhi/server/notifications.py#L33),

```
bodhi.server.config.config.get('fedmsg_enabled')
```

This is problematic for a few reasons:

1. The code that wants to use the config is always given a string of what it is set to in the config file, regardless of what type the config makes sense to be. In the above example, it makes sense for the config to be a boolean, and it being a string and not interpreted as a bool caused #1488.
2. There is no central place in code to see what all of Bodhi's defaults are, as the defaults need to be coded at the places that the config is read. Also, many of Bodhi's settings actually do not define defaults and Bodhi will explode if some settings don't exist.
3. Some settings are used in more than once place (including the above cited example), introducing the risk that a setting will be interpreted differently in one place than it is in another (which could be especially bad if they choose different defaults).

There are a couple things we can do to help:

1. Have config.py interpret all the settings' types, so that the rest of the code can say things just like the above and get the right type (bool, int, string, whatever) without having to remember to cast it.
2. Put all defaults into config.py so that the bodhi config file can be mostly commented out. Some settings can't be commented out because they are pyramid settings, but the bodhi specific ones could all have sane defaults. We need to be careful not to change any existing defaults when doing this.

# Outside the scope

This is outside the scope of this ticket, but I wanted a place to write down some related thoughts.

Something else I've pondered is giving Bodhi a nicer config file format. AFAIK, we are actually stuck with the ini file for Pyramid settings, but there's no reason we couldn't split Bodhi-specific settings into a new config file. TOML or YAML would be nice. This way, the config file itself encodes the types of the settings rather than us doing awkward things like parsing a string that has commas and deciding that means a list.

bowlofeggs:
This has come up as an issue three times this week, so I'm going to go ahead and make it happen. It's a good Friday project, and it will help us avoid a whole class of bugs in the future.

PR #1499, opened 2017-05-03
Add a ci_status to updates

pypingou:
This status is used to reflect the status of the update in the CI
pipeline.
It can be:
- ignored: all the builds in the update are and will not be tested by the
  CI pipeline
- queued: at least one build in the update has been queued to be tested
  in the CI pipeline
- running: at least one build in the update is being tested in the CI
  pipeline
- passed: all the builds in the update have passed testing in the CI
  pipeline
- failed: at least one build in the update has failed testedin in the CI
  pipeline

Note: the status is currently implemented using an update wide flag while
it is the builds composing the update that are being tested.
We may want to rework this approach to make the ci_status be a property
that inspects the ci_status of the builds composing the update and
implements the logic described above.

### DIFF
--- a/alembic/versions/4b357c65441e_add_ci_status_to_builds.py
+++ b/alembic/versions/4b357c65441e_add_ci_status_to_builds.py
@@ -1,0 +1,38 @@
+"""Add ci_status to builds
+
+Revision ID: 4b357c65441e
+Revises: b01a62d98aa4
+Create Date: 2017-05-11 20:13:41.879435
+"""
+from alembic import op
+import sqlalchemy as sa
+
+
+# revision identifiers, used by Alembic.
+revision = '4b357c65441e'
+down_revision = 'b01a62d98aa4'
+
+
+def upgrade():
+    """ Add the ci_status to builds with the corresponding enum. """
+
+    op.execute(
+        "CREATE TYPE ck_ci_status AS ENUM "
+        "('ignored', 'queued', 'running', 'passed', 'failed', 'waiting')")
+
+    op.add_column(
+        'builds',
+        sa.Column(
+            'ci_status',
+            sa.Enum(
+                'ignored', 'queued', 'running', 'passed', 'failed', 'waiting',
+                name='ck_ci_status'),
+            nullable=True
+        )
+    )
+
+
+def downgrade():
+    """ Remove the ci_status from builds with the corresponding enum. """
+    op.drop_column('builds', 'ci_status')
+    op.execute("DROP TYPE ck_ci_status")
--- a/alembic/versions/8eaacb38b036_add_the_ci_url_field_to_builds.py
+++ b/alembic/versions/8eaacb38b036_add_the_ci_url_field_to_builds.py
@@ -1,0 +1,27 @@
+"""Add the ci_url field to builds
+
+Revision ID: 8eaacb38b036
+Revises: 4b357c65441e
+Create Date: 2017-05-18 12:01:20.698762
+"""
+from alembic import op
+import sqlalchemy as sa
+
+
+# revision identifiers, used by Alembic.
+revision = '8eaacb38b036'
+down_revision = '4b357c65441e'
+
+
+def upgrade():
+    """ Add the ci_url to builds. """
+
+    op.add_column(
+        'builds',
+        sa.Column('ci_url', sa.UnicodeText, nullable=True)
+    )
+
+
+def downgrade():
+    """ Remove the ci_url from builds. """
+    op.drop_column('builds', 'ci_url')
--- a/bodhi/server/models.py
+++ b/bodhi/server/models.py
@@ -374,6 +374,23 @@ class UpdateStatus(DeclEnum):
     processing = 'processing', 'processing'
 
 
+class CiStatus(DeclEnum):
+    """ This class lists the different status the ci_status flag can have. """
+    # None: Updates created while bodhi has the configuration key `ci.required` not set to True
+    # waiting: Updates created while bodhi has the configuration key `ci.required` set to True
+    waiting = 'waiting', 'Waiting'
+    # ignored: a message from the CI system said that this build is ignored
+    ignored = 'ignored', 'Ignored'
+    # queued: a message from the CI system said that the tests for this build have been queued
+    queued = 'queued', 'Queued'
+    # running: a message from the CI system said that the tests for this build are running
+    running = 'running', 'Running'
+    # passed: a message from the CI system said that the tests for this build have passed
+    passed = 'passed', 'Passed'
+    # failed: a message from the CI system said that the tests for this build have been failed
+    failed = 'failed', 'Failed'
+
+
 class UpdateType(DeclEnum):
     bugfix = 'bugfix', 'bugfix'
     security = 'security', 'security'
@@ -770,6 +787,9 @@ class Build(Base):
             of.
         type (int): The polymorphic identify of the row. This is used by sqlalchemy to identify
             which subclass of Build to use.
+        ci_status (EnumSymbol): The CI status of the update. This must be one of the values
+            defined in :class:`CiStatus` or ``None``. None indicates that CI was not enabled when
+            the update was created.
     """
     __tablename__ = 'builds'
     __exclude_columns__ = ('id', 'package', 'package_id', 'release',
@@ -781,6 +801,8 @@ class Build(Base):
     release_id = Column(Integer, ForeignKey('releases.id'))
     signed = Column(Boolean, default=False, nullable=False)
     update_id = Column(Integer, ForeignKey('updates.id'))
+    ci_status = Column(CiStatus.db_type(), default=None, nullable=True)
+    ci_url = Column(UnicodeText, default=None, nullable=True)
 
     release = relationship('Release', backref='builds', lazy=False)
 
@@ -1269,6 +1291,12 @@ class Update(Base):
         log.debug("Triggering db flush for new update.")
         db.flush()
 
+        if config.get('ci.required'):
+            log.debug(
+                'CI required is enforced, marking the build as waiting on CI')
+            for build in up.builds:
+                build.ci_status = CiStatus.waiting
+
         log.debug("Done with Update.new(...)")
         return up, caveats
 
@@ -1414,6 +1442,52 @@ class Update(Base):
         """
         if self.builds:
             return self.builds[0].type
+
+    @property
+    def ci_passed(self):
+        """
+        Returns a boolean representing if all the builds associated with
+        this update are ignored or have passed their CI testing.
+
+        Returns:
+            bool: Returns True if the Update's ci_status property is None, ignored, or passed.
+                Otherwise it returns False.
+        """
+        if self.ci_status in (
+                None, CiStatus.ignored, CiStatus.passed):
+            return True
+        return False
+
+    @property
+    def ci_status(self):
+        """
+        Returns a label representing if all the builds associated with
+        this update are ignored or have passed their CI testing.
+
+        This will return None when bodhi is configured with the configuration
+        key `ci.required` not set to True (ie: when CI support is not enabled).
+
+        Returns:
+            bodhi.server.models.CiStatus or None: The overall status of this update.
+        """
+        ci_status = set([
+            build.ci_status for build in self.builds
+            if build.ci_status is not None
+        ])
+        if ci_status == set():
+            return None
+        if ci_status == set([CiStatus.ignored]):
+            return CiStatus.ignored
+        if ci_status == set([CiStatus.passed]):
+            return CiStatus.passed
+        if ci_status == set([CiStatus.waiting]):
+            return CiStatus.waiting
+        if CiStatus.failed in ci_status:
+            return CiStatus.failed
+        if CiStatus.running in ci_status:
+            return CiStatus.running
+        if CiStatus.queued in ci_status:
+            return CiStatus.queued
 
     def obsolete_older_updates(self, db):
         """Obsolete any older pending/testing updates.
@@ -2208,6 +2282,9 @@ class Update(Base):
                     return False, "Required task %s returned %s" % (
                         latest['testcase']['name'], latest['outcome'])
 
+        if config.get('ci.required') and not self.ci_passed:
+            return False, "CI did not pass on this update"
+
         # TODO - check require_bugs and require_testcases also?
 
         return True, "All checks pass."
@@ -2307,6 +2384,9 @@ class Update(Base):
         simply return True.
         """
         num_days = self.mandatory_days_in_testing
+
+        if config.get('ci.required') and not self.ci_passed:
+            return False
 
         if self.critpath:
             # Ensure there is no negative karma. We're looking at the sum of
--- a/bodhi/server/templates/update.html
+++ b/bodhi/server/templates/update.html
@@ -686,6 +686,23 @@ $(document).ready(function(){
                 </div>
               </div>
 
+              % if request.registry.settings.get('ci.required'):
+              <div class="p-b-1">
+                <div>
+                  % if request.registry.settings.get('ci.url'):
+                    <a href="${request.registry.settings.get('ci.url')}" title="More information about CI">
+                      <strong>CI Status</strong>
+                    </a>
+                  % else:
+                    <strong>CI Status</strong>
+                  % endif
+                </div>
+                <div>
+                  ${self.util.ci_status2html(update.ci_status.description) | n}
+                </div>
+              </div>
+              % endif
+
               % if update.request:
               <div class="p-b-1">
                 <div>
@@ -862,6 +879,17 @@ $(document).ready(function(){
           <td>
             <a href="https://koji.fedoraproject.org/koji/search?terms=${build.nvr}&type=build&match=glob" target="_blank">
                 ${build.nvr}</a>
+          </td>
+          % if request.registry.settings.get('ci.required'):
+            <td>
+              % if build.ci_url:
+                <a href="${build.ci_url}">${self.util.ci_status2html(build.ci_status.description) | n}</a>
+              % else:
+                ${self.util.ci_status2html(build.ci_status.description) | n}
+              % endif
+            </td>
+          % endif
+          <td class="pull-right">
             % if build.signed:
               <span class="fa fa-key text-muted" aria-hidden="true" data-toggle="tooltip" data-placement="top" title="Build signed"></span>
             % endif
--- a/bodhi/server/util.py
+++ b/bodhi/server/util.py
@@ -320,6 +320,32 @@ def status2html(context, status):
     return "<span class='label label-%s'>%s</span>" % (cls, status)
 
 
+def ci_status2html(context, status):
+    """ Convert the specified status into a html string used to present the
+    CI status in a human friendly way.
+
+    Args:
+        context (mako.runtime.Context): Unused.
+        status (unicode): The description of the CI status
+
+    Returns:
+        unicode: A human friendly HTML label of the CI status
+
+    """
+    cls = {
+        'Ignored': 'success',
+        'Running': 'warning',
+        'Passed': 'success',
+        'Failed': 'danger',
+        'Queued': 'info',
+        'Waiting': 'info',
+        'None': 'primary',
+    }[str(status)]
+    if status is None:
+        status = 'not running'
+    return u"<span class='label label-%s'>Tests %s</span>" % (cls, status)
+
+
 def state2class(context, state):
     state = unicode(state)
     cls = {
--- a/bodhi/tests/server/services/test_updates.py
+++ b/bodhi/tests/server/services/test_updates.py
@@ -29,7 +29,8 @@ from bodhi.server import main
 from bodhi.server.config import config
 from bodhi.server.models import (
     BuildrootOverride, Group, RpmPackage, Release,
-    ReleaseState, RpmBuild, Update, UpdateRequest, UpdateStatus, UpdateType, User)
+    ReleaseState, RpmBuild, Update, UpdateRequest, UpdateStatus, UpdateType,
+    User, CiStatus)
 import bodhi.tests.server.functional.base
 
 
@@ -614,6 +615,7 @@ class TestUpdatesService(bodhi.tests.server.functional.base.BaseWSGICase):
             topic='update.request.testing', msg=mock.ANY)
 
         build = self.db.query(RpmBuild).filter_by(nvr=nvr).one()
+        build.ci_status = CiStatus.passed
         eq_(build.update.request, UpdateRequest.testing)
 
         # Try and submit the update to stable as a non-provenpackager
@@ -703,6 +705,211 @@ class TestUpdatesService(bodhi.tests.server.functional.base.BaseWSGICase):
         app = TestApp(main({}, session=self.db, **anonymous_settings))
         res = app.get('/updates/%s' % str(nvr), status=200)
         eq_(res.json_body['can_edit'], False)
+
+    @mock.patch(**mock_taskotron_results)
+    @mock.patch(**mock_valid_requirements)
+    @mock.patch('bodhi.server.notifications.publish')
+    def test_provenpackager_request_update_queued_in_ci(self, publish, *args):
+        """Ensure provenpackagers cannot request changes for any update which
+        CI status is `queued`"""
+        nvr = u'bodhi-2.1-1.fc17'
+        user = User(name=u'bob')
+        self.db.add(user)
+        group = self.db.query(Group).filter_by(name=u'provenpackager').one()
+        user.groups.append(group)
+        self.app_settings['ci.required'] = True
+
+        app = TestApp(main({}, testing=u'bob', session=self.db, **self.app_settings))
+        up_data = self.get_update(nvr)
+        up_data['csrf_token'] = app.get('/csrf').json_body['csrf_token']
+        res = app.post_json('/updates/', up_data)
+        publish.assert_called_once_with(
+            topic='update.request.testing', msg=mock.ANY)
+
+        build = self.db.query(RpmBuild).filter_by(nvr=nvr).one()
+        build.ci_status = CiStatus.queued
+        eq_(build.update.request, UpdateRequest.testing)
+
+        # Try and submit the update to stable as a provenpackager
+        post_data = dict(update=nvr, request='stable',
+                         csrf_token=app.get('/csrf').json_body['csrf_token'])
+        res = app.post_json('/updates/%s/request' % str(nvr), post_data, status=400)
+
+        # Ensure we can't push it until it passed CI
+        eq_(res.json_body['status'], 'error')
+        eq_(res.json_body['errors'][0]['description'],
+            'Requirement not met CI did not pass on this update')
+
+    @mock.patch(**mock_taskotron_results)
+    @mock.patch(**mock_valid_requirements)
+    @mock.patch('bodhi.server.notifications.publish')
+    def test_provenpackager_request_update_running_in_ci(self, publish, *args):
+        """Ensure provenpackagers cannot request changes for any update which
+        CI status is `running`"""
+        nvr = u'bodhi-2.1-1.fc17'
+        user = User(name=u'bob')
+        self.db.add(user)
+        group = self.db.query(Group).filter_by(name=u'provenpackager').one()
+        user.groups.append(group)
+        self.app_settings['ci.required'] = True
+
+        app = TestApp(main({}, testing=u'bob', session=self.db, **self.app_settings))
+        up_data = self.get_update(nvr)
+        up_data['csrf_token'] = app.get('/csrf').json_body['csrf_token']
+        res = app.post_json('/updates/', up_data)
+        publish.assert_called_once_with(
+            topic='update.request.testing', msg=mock.ANY)
+
+        build = self.db.query(RpmBuild).filter_by(nvr=nvr).one()
+        build.ci_status = CiStatus.running
+        eq_(build.update.request, UpdateRequest.testing)
+
+        # Try and submit the update to stable as a provenpackager
+        post_data = dict(update=nvr, request='stable',
+                         csrf_token=app.get('/csrf').json_body['csrf_token'])
+        res = app.post_json('/updates/%s/request' % str(nvr), post_data, status=400)
+
+        # Ensure we can't push it until it passed CI
+        eq_(res.json_body['status'], 'error')
+        eq_(res.json_body['errors'][0]['description'],
+            'Requirement not met CI did not pass on this update')
+
+    @mock.patch(**mock_taskotron_results)
+    @mock.patch(**mock_valid_requirements)
+    @mock.patch('bodhi.server.notifications.publish')
+    def test_provenpackager_request_update_failed_ci(self, publish, *args):
+        """Ensure provenpackagers cannot request changes for any update which
+        CI status is `failed`"""
+        nvr = u'bodhi-2.1-1.fc17'
+        user = User(name=u'bob')
+        self.db.add(user)
+        group = self.db.query(Group).filter_by(name=u'provenpackager').one()
+        user.groups.append(group)
+        self.app_settings['ci.required'] = True
+
+        app = TestApp(main({}, testing=u'bob', session=self.db, **self.app_settings))
+        up_data = self.get_update(nvr)
+        up_data['csrf_token'] = app.get('/csrf').json_body['csrf_token']
+        res = app.post_json('/updates/', up_data)
+        publish.assert_called_once_with(
+            topic='update.request.testing', msg=mock.ANY)
+
+        build = self.db.query(RpmBuild).filter_by(nvr=nvr).one()
+        build.ci_status = CiStatus.failed
+        eq_(build.update.request, UpdateRequest.testing)
+
+        # Try and submit the update to stable as a provenpackager
+        post_data = dict(update=nvr, request='stable',
+                         csrf_token=app.get('/csrf').json_body['csrf_token'])
+        res = app.post_json('/updates/%s/request' % str(nvr), post_data, status=400)
+
+        # Ensure we can't push it until it passed CI
+        eq_(res.json_body['status'], 'error')
+        eq_(res.json_body['errors'][0]['description'],
+            'Requirement not met CI did not pass on this update')
+
+    @mock.patch(**mock_taskotron_results)
+    @mock.patch(**mock_valid_requirements)
+    @mock.patch('bodhi.server.notifications.publish')
+    def test_provenpackager_request_update_ignored_by_ci(self, publish, *args):
+        """Ensure provenpackagers can request changes for any update which
+        CI status is `ignored`"""
+        nvr = u'bodhi-2.1-1.fc17'
+        user = User(name=u'bob')
+        self.db.add(user)
+        group = self.db.query(Group).filter_by(name=u'provenpackager').one()
+        user.groups.append(group)
+        self.app_settings['ci.required'] = True
+
+        app = TestApp(main({}, testing=u'bob', session=self.db, **self.app_settings))
+        up_data = self.get_update(nvr)
+        up_data['csrf_token'] = app.get('/csrf').json_body['csrf_token']
+        res = app.post_json('/updates/', up_data)
+        assert 'does not have commit access to bodhi' not in res, res
+        publish.assert_called_once_with(
+            topic='update.request.testing', msg=mock.ANY)
+
+        build = self.db.query(RpmBuild).filter_by(nvr=nvr).one()
+        build.ci_status = CiStatus.ignored
+        eq_(build.update.request, UpdateRequest.testing)
+
+        # Try and submit the update to stable as a provenpackager
+        post_data = dict(update=nvr, request='stable',
+                         csrf_token=app.get('/csrf').json_body['csrf_token'])
+        res = app.post_json('/updates/%s/request' % str(nvr), post_data, status=400)
+
+        # Ensure the reason we cannot push isn't CI this time
+        eq_(res.json_body['status'], 'error')
+        eq_(res.json_body['errors'][0]['description'],
+            config.get('not_yet_tested_msg'))
+
+    @mock.patch(**mock_taskotron_results)
+    @mock.patch(**mock_valid_requirements)
+    @mock.patch('bodhi.server.notifications.publish')
+    def test_provenpackager_request_update_waiting_on_ci(self, publish, *args):
+        """Ensure provenpackagers cannot request changes for any update which
+        CI status is `waiting`"""
+        nvr = u'bodhi-2.1-1.fc17'
+        user = User(name=u'bob')
+        self.db.add(user)
+        group = self.db.query(Group).filter_by(name=u'provenpackager').one()
+        user.groups.append(group)
+        self.app_settings['ci.required'] = True
+
+        app = TestApp(main({}, testing=u'bob', session=self.db, **self.app_settings))
+        up_data = self.get_update(nvr)
+        up_data['csrf_token'] = app.get('/csrf').json_body['csrf_token']
+        res = app.post_json('/updates/', up_data)
+        publish.assert_called_once_with(
+            topic='update.request.testing', msg=mock.ANY)
+
+        build = self.db.query(RpmBuild).filter_by(nvr=nvr).one()
+        build.ci_status = CiStatus.waiting
+        eq_(build.update.request, UpdateRequest.testing)
+
+        # Try and submit the update to stable as a provenpackager
+        post_data = dict(update=nvr, request='stable',
+                         csrf_token=app.get('/csrf').json_body['csrf_token'])
+        res = app.post_json('/updates/%s/request' % str(nvr), post_data, status=400)
+
+        # Ensure we can't push it until it passed CI
+        eq_(res.json_body['status'], 'error')
+        eq_(res.json_body['errors'][0]['description'],
+            'Requirement not met CI did not pass on this update')
+
+    @mock.patch(**mock_taskotron_results)
+    @mock.patch(**mock_valid_requirements)
+    @mock.patch('bodhi.server.notifications.publish')
+    def test_provenpackager_request_update_with_none_ci(self, publish, *args):
+        """Ensure provenpackagers can request changes for any update which
+        CI status is `None`"""
+        nvr = u'bodhi-2.1-1.fc17'
+        user = User(name=u'bob')
+        self.db.add(user)
+        group = self.db.query(Group).filter_by(name=u'provenpackager').one()
+        user.groups.append(group)
+        self.app_settings['ci.required'] = True
+
+        app = TestApp(main({}, testing=u'bob', session=self.db, **self.app_settings))
+        up_data = self.get_update(nvr)
+        up_data['csrf_token'] = app.get('/csrf').json_body['csrf_token']
+        res = app.post_json('/updates/', up_data)
+        publish.assert_called_once_with(
+            topic='update.request.testing', msg=mock.ANY)
+
+        build = self.db.query(RpmBuild).filter_by(nvr=nvr).one()
+        build.ci_status = None
+        eq_(build.update.request, UpdateRequest.testing)
+
+        # Try and submit the update to stable as a provenpackager
+        post_data = dict(update=nvr, request='stable',
+                         csrf_token=app.get('/csrf').json_body['csrf_token'])
+        res = app.post_json('/updates/%s/request' % str(nvr), post_data, status=400)
+
+        # Ensure the reason we can't push is not CI
+        eq_(res.json_body['status'], 'error')
+        eq_(res.json_body['errors'][0]['description'],
+            config.get('not_yet_tested_msg'))
 
     @mock.patch(**mock_valid_requirements)
     @mock.patch('bodhi.server.notifications.publish')
@@ -2355,6 +2562,8 @@ class TestUpdatesService(bodhi.tests.server.functional.base.BaseWSGICase):
         up = Update.get(nvr, self.db)
         up.status = UpdateStatus.testing
         up.request = None
+        eq_(len(up.builds), 1)
+        up.builds[0].ci_status = CiStatus.passed
         self.db.flush()
 
         # Checks failure for requesting to stable push before the update reaches stable karma
@@ -2392,6 +2601,8 @@ class TestUpdatesService(bodhi.tests.server.functional.base.BaseWSGICase):
         up.request = None
         up.comment(self.db, u'This update has been pushed to testing', author=u'bodhi')
         up.date_testing = up.comments[-1].timestamp - timedelta(days=7)
+        eq_(len(up.builds), 1)
+        up.builds[0].ci_status = CiStatus.passed
         self.db.flush()
         eq_(up.days_in_testing, 7)
         eq_(up.meets_testing_requirements, True)
@@ -2415,6 +2626,8 @@ class TestUpdatesService(bodhi.tests.server.functional.base.BaseWSGICase):
         up.status = UpdateStatus.pending
         up.request = None
         up.release.state = ReleaseState.archived
+        eq_(len(up.builds), 1)
+        up.builds[0].ci_status = CiStatus.passed
         self.db.flush()
         resp = self.app.post_json(
             '/updates/%s/request' % args['builds'],
@@ -2436,6 +2649,8 @@ class TestUpdatesService(bodhi.tests.server.functional.base.BaseWSGICase):
         up.request = None
         up.comment(self.db, u'This update has been pushed to testing', author=u'bodhi')
         up.date_testing = up.comments[-1].timestamp - timedelta(days=7)
+        eq_(len(up.builds), 1)
+        up.builds[0].ci_status = CiStatus.passed
         self.db.flush()
         eq_(up.days_in_testing, 7)
         eq_(up.meets_testing_requirements, True)
@@ -2458,6 +2673,8 @@ class TestUpdatesService(bodhi.tests.server.functional.base.BaseWSGICase):
         up.request = None
         up.comment(self.db, u'This update has been pushed to testing', author=u'bodhi')
         up.date_testing = up.comments[-1].timestamp - timedelta(days=7)
+        eq_(len(up.builds), 1)
+        up.builds[0].ci_status = CiStatus.passed
         self.db.flush()
         eq_(up.days_in_testing, 7)
         eq_(up.meets_testing_requirements, True)
@@ -2482,6 +2699,8 @@ class TestUpdatesService(bodhi.tests.server.functional.base.BaseWSGICase):
         up.comment(self.db, u'This update has been pushed to testing', author=u'bodhi')
         up.date_testing = up.comments[-1].timestamp - timedelta(days=14)
         up.comment(self.db, u'This update has been pushed to stable', author=u'bodhi')
+        eq_(len(up.builds), 1)
+        up.builds[0].ci_status = CiStatus.passed
         self.db.flush()
         eq_(up.days_in_testing, 14)
         eq_(up.meets_testing_requirements, True)
@@ -2510,6 +2729,8 @@ class TestUpdatesService(bodhi.tests.server.functional.base.BaseWSGICase):
         up.request = None
         up.comment(self.db, u'This update has been pushed to testing', author=u'bodhi')
         up.date_testing = up.comments[-1].timestamp - timedelta(days=14)
+        eq_(len(up.builds), 1)
+        up.builds[0].ci_status = CiStatus.passed
         self.db.flush()
         eq_(up.days_in_testing, 14)
         eq_(up.meets_testing_requirements, True)
@@ -2654,7 +2875,7 @@ class TestUpdatesService(bodhi.tests.server.functional.base.BaseWSGICase):
         self.assertEquals(len(r.json['bugs']), 1)
         publish.assert_called_with(topic='update.request.testing', msg=ANY)
 
-        # Pretend it was pushed to testing
+        # Pretend it was pushed to testing and tested
         update = self.db.query(Update).filter_by(title=build).one()
         update.request = None
         update.status = UpdateStatus.testing
@@ -2713,7 +2934,7 @@ class TestUpdatesService(bodhi.tests.server.functional.base.BaseWSGICase):
         self.assertEquals(up['stable_karma'], 3)
         self.assertEquals(up['unstable_karma'], -3)
 
-        # Pretend it was pushed to testing
+        # Pretend it was pushed to testing and tested
         update = self.db.query(Update).filter_by(title=build).one()
         update.request = None
         update.status = UpdateStatus.testing
@@ -2749,7 +2970,7 @@ class TestUpdatesService(bodhi.tests.server.functional.base.BaseWSGICase):
         up = r.json_body
         self.assertEquals(up['type'], u'newpackage')
 
-        # Pretend it was pushed to testing
+        # Pretend it was pushed to testing and tested
         update = self.db.query(Update).filter_by(title=build).one()
         update.request = None
         update.status = UpdateStatus.testing
@@ -2780,7 +3001,7 @@ class TestUpdatesService(bodhi.tests.server.functional.base.BaseWSGICase):
         r = self.app.post_json('/updates/', args)
         publish.assert_called_with(topic='update.request.testing', msg=ANY)
 
-        # Mark it as testing and give it 2 karma
+        # Mark it as testing, tested and give it 2 karma
         upd = Update.get(nvr, self.db)
         upd.status = UpdateStatus.testing
         upd.request = None
@@ -2814,7 +3035,7 @@ class TestUpdatesService(bodhi.tests.server.functional.base.BaseWSGICase):
         r = self.app.post_json('/updates/', args)
         publish.assert_called_with(topic='update.request.testing', msg=ANY)
 
-        # Mark it as testing
+        # Mark it as testing and as tested
         upd = Update.get(nvr, self.db)
         upd.status = UpdateStatus.testing
         upd.request = None
--- a/bodhi/tests/server/test_models.py
+++ b/bodhi/tests/server/test_models.py
@@ -31,7 +31,7 @@ from bodhi.server.config import config
 from bodhi.server.exceptions import BodhiException
 from bodhi.server.models import (
     BugKarma, ReleaseState, UpdateRequest, UpdateSeverity, UpdateStatus,
-    UpdateSuggestion, UpdateType)
+    UpdateSuggestion, UpdateType, CiStatus)
 from bodhi.tests.server.base import BaseTestCase
 
 
@@ -591,7 +591,7 @@ class TestUpdate(ModelTest):
         return dict(
             builds=[model.RpmBuild(
                 nvr=u'TurboGears-1.0.8-3.fc11', package=model.RpmPackage(**TestRpmPackage.attrs),
-                release=release)],
+                release=release, ci_status=CiStatus.passed)],
             bugs=[model.Bug(bug_id=1), model.Bug(bug_id=2)],
             cves=[model.CVE(cve_id=u'CVE-2009-0001')],
             release=release,
@@ -627,6 +627,62 @@ class TestUpdate(ModelTest):
         eq_(self.obj.builds[0].nvr, u'TurboGears-1.0.8-3.fc11')
         eq_(self.obj.builds[0].release.name, u'F11')
         eq_(self.obj.builds[0].package.name, u'TurboGears')
+
+    def test_ci_status_ignored(self):
+        """Test the ci_status property when all Builds have their ci_status set to ignored."""
+        self.obj.builds[0].ci_status = CiStatus.ignored
+        second_build = model.RpmBuild(nvr='bodhi-2.7.0-1.fc26', ci_status=CiStatus.ignored)
+        self.obj.builds.append(second_build)
+
+        self.assertEqual(self.obj.ci_status, CiStatus.ignored)
+
+    def test_ci_status_none(self):
+        """Test the ci_status property when all Builds have their ci_status set to None."""
+        self.obj.builds[0].ci_status = None
+        second_build = model.RpmBuild(nvr='bodhi-2.7.0-1.fc26', ci_status=None)
+        self.obj.builds.append(second_build)
+
+        self.assertEqual(self.obj.ci_status, None)
+
+    def test_ci_status_one_failed(self):
+        """Test the ci_status property when one Build passed and the other failed."""
+        self.obj.builds[0].ci_status = CiStatus.passed
+        second_build = model.RpmBuild(nvr='bodhi-2.7.0-1.fc26', ci_status=CiStatus.failed)
+        self.obj.builds.append(second_build)
+
+        self.assertEqual(self.obj.ci_status, CiStatus.failed)
+
+    def test_ci_status_one_queued(self):
+        """Test the ci_status property when one Build passed and the other is queued."""
+        self.obj.builds[0].ci_status = CiStatus.passed
+        second_build = model.RpmBuild(nvr='bodhi-2.7.0-1.fc26', ci_status=CiStatus.queued)
+        self.obj.builds.append(second_build)
+
+        self.assertEqual(self.obj.ci_status, CiStatus.queued)
+
+    def test_ci_status_one_running(self):
+        """Test the ci_status property when one Build passed and the other is running."""
+        self.obj.builds[0].ci_status = CiStatus.passed
+        second_build = model.RpmBuild(nvr='bodhi-2.7.0-1.fc26', ci_status=CiStatus.running)
+        self.obj.builds.append(second_build)
+
+        self.assertEqual(self.obj.ci_status, CiStatus.running)
+
+    def test_ci_status_passed(self):
+        """Test the ci_status property when all Builds have their ci_status set to passed."""
+        self.obj.builds[0].ci_status = CiStatus.passed
+        second_build = model.RpmBuild(nvr='bodhi-2.7.0-1.fc26', ci_status=CiStatus.passed)
+        self.obj.builds.append(second_build)
+
+        self.assertEqual(self.obj.ci_status, CiStatus.passed)
+
+    def test_ci_status_waiting(self):
+        """Test the ci_status property when all Builds have their ci_status set to waiting."""
+        self.obj.builds[0].ci_status = CiStatus.waiting
+        second_build = model.RpmBuild(nvr='bodhi-2.7.0-1.fc26', ci_status=CiStatus.waiting)
+        self.obj.builds.append(second_build)
+
+        self.assertEqual(self.obj.ci_status, CiStatus.waiting)
 
     def test_content_type(self):
         eq_(self.obj.content_type, model.ContentType.rpm)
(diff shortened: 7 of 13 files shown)


bodhi/server/models.py
```diff
@@ -923,6 +933,7 @@ class Update(Base):
     status = Column(UpdateStatus.db_type(),
                     default=UpdateStatus.pending,
                     nullable=False)
+    ci_status = Column(UpdateCiStatus.db_type(), default=None, nullable=True)
```
jeremycline: Given that the column is nullable, does it need a default of ``None``? Shouldn't that be a no-op?

bodhi/tests/server/services/test_updates.py
```diff
@@ -517,9 +518,21 @@ def test_provenpackager_request_privs(self, publish, *args):
                          csrf_token=app.get('/csrf').json_body['csrf_token'])
         res = app.post_json('/updates/%s/request' % str(nvr), post_data, status=400)
 
-        # Ensure we can't push it until it meets the requirements
+        # Ensure we can't push it until it passed CI
         eq_(res.json_body['status'], 'error')
-        eq_(res.json_body['errors'][0]['description'], config.get('not_yet_tested_msg'))
+        eq_(res.json_body['errors'][0]['description'],
```
jeremycline: Given that nose is a dead project, it's probably better to use the standard unittest assertions to ease later migration.

bodhi/tests/server/services/test_updates.py
```diff
@@ -517,9 +518,21 @@ def test_provenpackager_request_privs(self, publish, *args):
...
+            'Requirement not met CI did not pass on this update')
+
+        # Make the update pass CI
+        update = self.db.query(Update).filter_by(title=nvr).one()
```
jeremycline: You should be able to write this as ``Update.query.filter_by(title=nvr).one()``

bodhi/server/models.py
```diff
@@ -317,6 +317,14 @@ class UpdateStatus(DeclEnum):
     processing = 'processing', 'processing'
 
 
+class UpdateCiStatus(DeclEnum):
```
jeremycline: I recommend adding documentation here.

bodhi/server/models.py
pypingou: I think you're right, thought I prefer to be explicit, I'm fine with removing this if you want.

bodhi/server/models.py

bodhi/tests/server/services/test_updates.py
pypingou: Since I'm just reformatting an existing assertion, I think I would prefer leaving this for another PR

bodhi/tests/server/services/test_updates.py
pypingou: Will do

bodhi/server/models.py
jeremycline: Since it doesn't do anything I don't have a particularly strong opinion one way or the other, but I think marking a column as nullable is explicit since it indicates that the column itself allows null values which _must_ mean the mapped Python object's attribute is allowed to be null and if you don't provide a value... Well, it's going to be null.

bowlofeggs:
Here's an example of how I'm adding a new enum to Bodhi: https://github.com/fedora-infra/bodhi/pull/1495/files#diff-ba24598f13c40d69ee832a77c76e43abR50

pypingou:
bodhi_ci_status

Small preview :)

pypingou:
One item missing, a way to turn on/off these checks via the configuration file.

Any suggestion on this?

alembic/versions/4b357c65441e_add_ci_status_to_builds.py
```diff
@@ -0,0 +1,35 @@
...
+# revision identifiers, used by Alembic.
+revision = '4b357c65441e'
+down_revision = 'c6a5e2849ca4'
+
+
+def upgrade():
+    """ Add the ci_status to builds with the corresponding enum. """
+
+    op.execute(
+        "CREATE TYPE ck_ci_status AS ENUM ('ignored','queued','running', 'passed', 'failed')")
```
bowlofeggs: For consistency, it might be nice to include a space after each comma in here so that it's "PEP-8-like". PEP-8 doesn't really care about SQL scripts. Your option.

alembic/versions/4b357c65441e_add_ci_status_to_builds.py
```diff
@@ -0,0 +1,35 @@
...
+            sa.Enum('ignored', 'queued', 'running', 'passed', 'failed', name='ck_ci_status'),
+            nullable=True
+        )
+    )
+
+
+def downgrade():
+    """ Remove the ci_status to builds with the corresponding enum. """
+    op.execute("DROP TYPE ck_ci_status")
+    op.drop_column('builds', 'ci_status')
```
bowlofeggs: We should reverse the order of these statements - Postgres is be upset that the type is removed before the column that uses it:

```
sqlalchemy.exc.InternalError: (psycopg2.InternalError) cannot drop type ck_ci_status because other objects depend on it
DETAIL:  table builds column ci_status depends on type ck_ci_status
HINT:  Use DROP ... CASCADE to drop the dependent objects too.
 [SQL: 'DROP TYPE ck_ci_status']
```

bodhi/server/models.py
```diff
@@ -317,6 +317,16 @@ class UpdateStatus(DeclEnum):
...
 
 
+class CiStatus(DeclEnum):
+    """ This class lists the different status the ci_status flag can have.
+    """
+    ignored = 'ignored', 'ignored'
+    queued = 'queued', 'queued'
+    running = 'running', 'running'
+    passed = 'passed', 'passed'
+    failed = 'failed', 'failed'
```
bowlofeggs: For each of these, the second one is meant to be used for a human-readable label. I recommend capitalizing each of them. I know that's not consistent with the others here, but I've been meaning to fix them, and the new one I just added for type uses the human-readable form.

bodhi/server/models.py
```diff
@@ -317,6 +317,16 @@ class UpdateStatus(DeclEnum):
     processing = 'processing', 'processing'
 
 
+class CiStatus(DeclEnum):
+    """ This class lists the different status the ci_status flag can have.
+    """
```
bowlofeggs: Can we fit the second """ onto the same line? Optional.

bodhi/server/models.py
```diff
@@ -1259,6 +1272,41 @@ def signed(self):
...
+        this update are ignored or have passed their CI testing.
+
+        Returns False if at least one of the tests of the builds associated
+        with this update are queued, running or failed.
+        """
+
+        ci_status = set([build.ci_status for build in self.builds])
+        if not ci_status.difference(set([CiStatus.ignored, CiStatus.passed])):
+            return True
+        return False
```
bowlofeggs: Since the label ```ci_status``` elsewhere would refer to the Enum type, it's a little confusing that this method returns a bool. I suggest renaming the method to ```ci_passed``` so its name sounds like something that should return a bool.

bodhi/server/models.py
```diff
@@ -1259,6 +1272,41 @@ def signed(self):
...
+
+    @property
+    def ci_status_label(self):
+        """ Returns a label representing if all the builds associated with
+        this update are ignored or have passed their CI testing.
+
+        Returns one of: ignored, passed, failed, queued, running
```
bowlofeggs: Technically it can return "unknown" as written too.

bodhi/server/models.py
```diff
@@ -1259,6 +1272,41 @@ def signed(self):
...
+        """
+
+        ci_status = set([build.ci_status for build in self.builds])
+        if ci_status == set([CiStatus.ignored]):
+            return 'ignored'
```
bowlofeggs: This line is missing test coverage.

bodhi/server/models.py
```diff
@@ -1259,6 +1272,41 @@ def signed(self):
...
+        if ci_status == set([CiStatus.passed]):
+            return 'passed'
+        if CiStatus.failed in ci_status:
+            return 'failed'
+        if CiStatus.running in ci_status:
+            return 'running'
+        if CiStatus.queued in ci_status:
+            return 'queued'
```
bowlofeggs: For each of the above, you could take advantage of the Enum's built-in human readable string. I *think* you can just do ```str()``` (or maybe it's ```unicode()```? I can't remember…) on it. You can probably also just return the Enum itself and the template should just render it.

bodhi/server/models.py
```diff
@@ -1259,6 +1272,41 @@ def signed(self):
...
+        Returns one of: ignored, passed, failed, queued, running
+        """
+
+        ci_status = set([build.ci_status for build in self.builds])
+        if ci_status == set([CiStatus.ignored]):
+            return 'ignored'
+        if ci_status == set([CiStatus.passed]):
+            return 'passed'
+        if CiStatus.failed in ci_status:
+            return 'failed'
```
bowlofeggs: This line is missing test coverage.

bodhi/server/models.py
```diff
@@ -1259,6 +1272,41 @@ def signed(self):
...
+        if CiStatus.running in ci_status:
+            return 'running'
```
bowlofeggs: This line is missing test coverage.

bodhi/server/models.py
bowlofeggs: This line is missing test coverage.

bodhi/server/models.py
```diff
@@ -1259,6 +1272,41 @@ def signed(self):
...
+        if CiStatus.queued in ci_status:
+            return 'queued'
+        return 'unknown'
```
bowlofeggs: It might be best to drop this line and let ```None``` represent "unknown", just like ```None``` is used on the Build.

bodhi/server/util.py
```diff
@@ -343,6 +343,20 @@ def status2html(context, status):
...
+    cls = {
+        'ignored': 'success',
+        'running': 'warning',
+        'passed': 'success',
+        'failed': 'danger',
+        'queued': 'info',
+        'unknown': 'primary',
+    }[status]
+    if status == 'unknown':
+        status = 'info awaited'
```
bowlofeggs: Since a great many existing updates are going to have ```None``` as their ```Build.ci_status```, I think this will cause them to show "Tests info awaited". I think we instead probably just don't want to display CI Status for historic builds at all.

bodhi/server/templates/update.html
```diff
@@ -650,6 +650,15 @@ <h5 class="m-t-2">
...
               </div>
 
+              <div class="p-b-1">
+                <div>
+                  <strong>CI Status</strong>
+                </div>
+                <div>
+                  ${self.util.ci_status2html(update.ci_status_label) | n}
+                </div>
+              </div>
```
bowlofeggs: I suggest renaming ```ci_status_label``` to ```ci_status``` and checking to see if it returns None or not before displaying the above ```<div>```. That way our existing Updates won't show CI Status since they will never have tests.

bowlofeggs (changes requested):
I need to go for now, and I haven't gotten to reviewing the tests yet. I'll look again tomorrow. Also, please accept my apologies for taking so long to get to this pull request. Last week was busy (as you know) and the week before was when Bodhi was very broken. That's my excuse and I'm sticking to it ☺

bodhi/server/models.py
```diff
@@ -1259,6 +1272,41 @@ def signed(self):
...
+        with this update are queued, running or failed.
+        """
+
+        ci_status = set([build.ci_status for build in self.builds])
+        if not ci_status.difference(set([CiStatus.ignored, CiStatus.passed])):
+            return True
+        return False
+
+    @property
+    def ci_status_label(self):
```
bowlofeggs: I recommend renaming this to ```ci_status``` and having it just return the Enum. Then you can use the Enum in the template and the human readable bit should "just work".

bodhi/server/models.py
bowlofeggs: If you go with my suggestion on the other method to rename it and have it return an enum, you can also simplify this method to use the other one. Something like this maybe:

```
def ci_passed(self):
    if self.ci_status in (CiStatus.ignored, CiStatus.passed):
        return True
    return False
```

What do you think?

bowlofeggs:
One more thought - I'm not sure whether the masher calls either of the methods where ```ci_status``` was checked to see if the tests passed (I could be wrong, need to double check but don't have time ATM).

AIUI, we want to prevent updates from even getting pushed to testing if their CI tests failed. Is that true? If so, we might need to add a check in the masher too, or perhaps in ```bodhi-push``` (or both).

bodhi/server/models.py
bowlofeggs: One more thought - it might be good to allow ```None``` to mean "passed" too.

alembic/versions/4b357c65441e_add_ci_status_to_builds.py
pypingou: Done

alembic/versions/4b357c65441e_add_ci_status_to_builds.py
pypingou: Done

bodhi/server/models.py
pypingou: Done

bodhi/server/models.py
pypingou: Done

bodhi/server/models.py
pypingou: > One more thought - it might be good to allow None to mean "passed" too.

We agreed that then there is no distinction between: ``I do not know whether this update should be tested`` and ``I missed the info about it being ignored``, as well as potentially run into race-condition issue between the time the update is creating and the time it is checked for CI for inclusion into the mash process.

bodhi/server/models.py
pypingou: Adjusted

bodhi/server/models.py
pypingou: Sure can do

bodhi/server/models.py
pypingou: Logic adjusted around ci_status vs ci_passed

bodhi/server/util.py
pypingou: I propose another way: ``op.execute("UPDATE builds set ci_status='ignored'")`` in the alembic migration.

Not displaying the CI status sounds more like something we would want to condition based on a configuration key.

bowlofeggs:
On Wed, 2017-05-17 at 01:48 -0700, Pierre-Yves Chibon wrote:
> We agreed that then there is no distinction between: I do not know
> whether this update should be tested and I missed the info about it
> being ignored, as well as potentially run into race-condition issue
> between the time the update is creating and the time it is checked
> for CI for inclusion into the mash process.

Wouldn't we create all updates in the "queued" state if the CI stuff
were configured? In other words, I don't think any of the new Fedora
updates will ever have the None state once we turn this feature on -
only old updates that pre-date this new system would have None. And
Bodhi deployments that don't turn the feature on will have all updates
in the None state. Do you think that will work?

bowlofeggs:
On Wed, 2017-05-17 at 01:53 -0700, Pierre-Yves Chibon wrote:
> I propose another way: op.execute("UPDATE builds set
> ci_status='ignored'") in the alembic migration.
> Not displaying the CI status sounds more like something we would want
> to condition based on a configuration key.

Yeah I thought about doing that too, and making the column non-
nullable. I don't see a major advantage or disadvantage to either
approach, but I do see one small advantage to keeping it nullable: we
will be able to tell the difference between historic updates that
predate CI vs. updates from the glorious era of CI because the old
updates would have None for the status and the new updates would have
an enum.

That said, I don't mind going either way so if you prefer setting them
all to ignored I think that's fine.

bodhi/server/templates/update.html
```diff
@@ -689,6 +689,15 @@ <h5 class="m-t-2">
                 </div>
               </div>
 
+              <div class="p-b-1">
+                <div>
+                  <strong>CI Status</strong>
```
ralphbean: It would be rad if this was a link to a wiki page describing the CI project.

ralphbean (changes requested):

bodhi/server/templates/update.html
pypingou: Tbh, I would prefer this be a link to the actual results or somewhere where one can investigate the CI results.

However I don't think we have this URL so far, so I'm considering to add this in a second step.

Or we start with the wiki link and change it later on, also wfm.

ralphbean: Ah, ok.  I miscommunicated my request:

- Can you have the ``<strong>CI Status</strong>`` be a link to a static wiki page explaining the CI project.
- Can you have the ``${self.util.ci_status2html(update.ci_status) | n}`` label be a link to the actual results for *this* job.

pypingou: Oh I see, you're totally right and sorry for missing what you were saying, we were talking about two different places.

pypingou: Implemented ;-)

ralphbean: LGTM!

alembic/versions/4b357c65441e_add_ci_status_to_builds.py
```diff
@@ -0,0 +1,38 @@
...
+        'builds',
+        sa.Column(
+            'ci_status',
+            sa.Enum('ignored', 'queued', 'running', 'passed', 'failed', name='ck_ci_status'),
+            nullable=True
+        )
+    )
+
+    # Give a default version to all the old builds
+    op.execute("UPDATE builds set ci_status='ignored'")
```
bowlofeggs: If we're giving a default to all builds, we should make the field non-nullable as well. Otherwise a value of NULL is ambiguous.

alembic/versions/8eaacb38b036_add_the_ci_url_field_to_builds.py
```diff
@@ -0,0 +1,27 @@
...
+revision = '8eaacb38b036'
+down_revision = '4b357c65441e'
+
+
+def upgrade():
+    """ Add the ci_url to builds. """
+
+    op.add_column(
+        'builds',
+        sa.Column('ci_url', sa.UnicodeText, nullable=True)
```
bowlofeggs, line 21: Is there any chance we can know the URL based on the commit hash and avoid having another column? No big deal if not, just a thought.

bodhi/server/models.py
```diff
@@ -691,6 +702,8 @@ class Build(Base):
     release_id = Column(Integer, ForeignKey('releases.id'))
     signed = Column(Boolean, default=False, nullable=False)
     update_id = Column(Integer, ForeignKey('updates.id'))
+    ci_status = Column(CiStatus.db_type(), default=None, nullable=True)
```
bowlofeggs, line 804: It seems inconsistent for the migration to mark all updates as ignored, but for the default to be None.

I think I slightly prefer letting the old updates stay at None, but I think it's more important that we pick a consistent policy. Which do you prefer?

bodhi/server/models.py
```diff
@@ -1168,6 +1181,13 @@ def new(cls, request, data):
         log.debug("Triggering db flush for new update.")
         db.flush()
 
+        if not config.get('ci.required', False):
+            log.debug(
+                'CI required is not enforced, marking the update as being ignored')
+            for build in up.builds:
+                build.ci_status = CiStatus.ignored
```
bowlofeggs: If we do allow None, we wouldn't need this block of code.

bodhi/server/models.py
```diff
@@ -1168,6 +1181,13 @@ def new(cls, request, data):
...
+            db.flush()
```
bowlofeggs: Do we need to flush here?

bodhi/server/models.py
```diff
@@ -1309,6 +1329,39 @@ def signed(self):
...
+        if ci_status == set([CiStatus.ignored]):
+            return CiStatus.ignored.description
+        if ci_status == set([CiStatus.passed]):
+            return CiStatus.passed.description
+        if CiStatus.failed in ci_status:
+            return CiStatus.failed.description
+        if CiStatus.running in ci_status:
+            return CiStatus.running.description
+        if CiStatus.queued in ci_status:
+            return CiStatus.queued.description
```
bowlofeggs: I recommend having this simply return the Enum - the template can use the Enum to render, and this way we aren't doing string comparisons in code that doesn't need human readable text.

Also, technically nothing says the human readable form of an Enum has to be unique.

bodhi/server/models.py
```diff
@@ -1309,6 +1329,39 @@ def signed(self):
...
+    @property
+    def ci_passed(self):
+        """ Returns a boolean representing if all the builds associated with
+        this update are ignored or have passed their CI testing.
+
+        Returns False if at least one of the tests of the builds associated
+        with this update are queued, running or failed.
+        """
+        if self.ci_status in (
+                CiStatus.ignored.description, CiStatus.passed.description):
```
bowlofeggs: We should have ```ci_status``` simply return the Enum so these don't have to do string comparisons.

bodhi/server/models.py
```diff
@@ -1309,6 +1329,39 @@ def signed(self):
...
+            return True
+        return False
+
+    @property
+    def ci_status(self):
+        """ Returns a label representing if all the builds associated with
+        this update are ignored or have passed their CI testing.
+
+        Returns one of: ignored, passed, failed, queued, running, None
```
bowlofeggs: We should update these to mention that it's returning the Enum equivalent of these.

bodhi/server/models.py
```diff
@@ -2197,6 +2253,9 @@ def meets_testing_requirements(self):
         """
         num_days = self.mandatory_days_in_testing
 
+        if config.get('ci.required', False) and not self.ci_passed:
```
bowlofeggs: One thing to note that's pretty important: Bodhi's settings are strings - so if ```ci.required``` is "False", this will actually interpret it to be True since ```bool("False") is True```.

This config is used in lots of places in this PR, so be sure to fix the other uses of it too.

This has been the subject of other Bodhi bugs in the past. I really want to build a central settings interpreter that validates all settings, applies all defaults (so you don't have to remember to type ```, False``` everywhere like here), and converts all settings to their logical types (like bools, or lists, or whatever). It's a real pain how it is, and it's error prone. I did file a ticket about this:

https://github.com/fedora-infra/bodhi/issues/1489

bodhi/server/templates/update.html
```diff
@@ -692,6 +692,23 @@ <h5 class="m-t-2">
                 </div>
               </div>
 
+              % if request.registry.settings.get('ci.required', 'False') == 'True':
```
bowlofeggs: It might be good to throw a .lower() on there so the setting isn't case sensitive.

bodhi/server/templates/update.html
```diff
@@ -692,6 +692,23 @@ <h5 class="m-t-2">
...
+              <div class="p-b-1">
+                <div>
+                  % if request.registry.settings.get('ci.url'):
```
bowlofeggs: Should we put the if statement over the ```<div>``` so we don't create an empty div when the setting isn't present?

Also, if this setting isn't defined, will it raise an Exception?

bodhi/server/models.py
```diff
@@ -1309,6 +1329,39 @@ def signed(self):
...
+    def ci_passed(self):
+        """ Returns a boolean representing if all the builds associated with
+        this update are ignored or have passed their CI testing.
+
+        Returns False if at least one of the tests of the builds associated
+        with this update are queued, running or failed.
+        """
+        if self.ci_status in (
+                CiStatus.ignored.description, CiStatus.passed.description):
+            return True
```
bowlofeggs: This line is missing test coverage.

bodhi/server/models.py
```diff
@@ -1309,6 +1329,39 @@ def signed(self):
...
+    def ci_status(self):
+        """ Returns a label representing if all the builds associated with
+        this update are ignored or have passed their CI testing.
+
+        Returns one of: ignored, passed, failed, queued, running, None
+        """
+
+        ci_status = set([build.ci_status for build in self.builds])
+        if ci_status == set([CiStatus.ignored]):
+            return CiStatus.ignored.description
```
bowlofeggs: This line is missing test coverage.

bodhi/server/models.py
```diff
@@ -1309,6 +1329,39 @@ def signed(self):
...
+        if ci_status == set([CiStatus.passed]):
+            return CiStatus.passed.description
```
bowlofeggs: This line is missing test coverage.

bodhi/server/templates/update.html
```diff
@@ -868,6 +885,17 @@ <h5 class="m-t-2">
           <td>
             <a href="https://koji.fedoraproject.org/koji/search?terms=${build.nvr}&type=build&match=glob" target="_blank">
                 ${build.nvr}</a>
+          </td>
+          % if request.registry.settings.get('ci.required', 'False') == 'True':
```
bowlofeggs: Might want to make this one case insensitive too.

bodhi/server/templates/update.html
```diff
@@ -868,6 +885,17 @@ <h5 class="m-t-2">
...
+            <td>
+              % if build.ci_url:
+                <a href="${build.ci_url}">${self.util.ci_status2html(build.ci_status.description) | n}</a>
+              % else:
+                ${self.util.ci_status2html(build.ci_status.description) | n}
```
bowlofeggs, line 888: I think you might be able to leave description off here? I could be wrong.

bodhi/server/util.py
```diff
@@ -343,6 +343,20 @@ def status2html(context, status):
     return "<span class='label label-%s'>%s</span>" % (cls, status)
 
 
+def ci_status2html(context, status):
```
bowlofeggs, line 323: Let's give this a docblock so we know what it does and what its arguments are.

bodhi/server/util.py
```diff
@@ -343,6 +343,20 @@ def status2html(context, status):
...
+        'Ignored': 'success',
+        'Running': 'warning',
+        'Passed': 'success',
+        'Failed': 'danger',
+        'Queued': 'info',
+        'None': 'primary',
+    }[str(status)]
+    if status is None:
+        status = 'info awaited'
+    return "<span class='label label-%s'>Tests %s</span>" % (cls, status)
```
bowlofeggs: This function is missing test coverage.

bodhi/server/util.py
```diff
@@ -343,6 +343,20 @@ def status2html(context, status):
...
+    cls = {
+        'Ignored': 'success',
+        'Running': 'warning',
+        'Passed': 'success',
+        'Failed': 'danger',
+        'Queued': 'info',
+        'None': 'primary',
+    }[str(status)]
+    if status is None:
+        status = 'info awaited'
```
bowlofeggs: If we're going to set historic updates to "Ignored", we should probably not allow status to be None.

bodhi/server/util.py
```diff
@@ -343,6 +343,20 @@ def status2html(context, status):
...
 
 
+def ci_status2html(context, status):
+    cls = {
+        'Ignored': 'success',
+        'Running': 'warning',
+        'Passed': 'success',
+        'Failed': 'danger',
+        'Queued': 'info',
+        'None': 'primary',
```
bowlofeggs: We can use the Enums here too so we don't have to do string comparison - the return statement below should automatically convert the Enum to human readable form, I think.

bodhi/tests/server/services/test_updates.py
```diff
@@ -509,17 +510,18 @@ def test_provenpackager_request_privs(self, publish, *args):
...
         build = self.db.query(RpmBuild).filter_by(nvr=nvr).one()
+        build.ci_status = CiStatus.passed
         eq_(build.update.request, UpdateRequest.testing)
 
         # Try and submit the update to stable as a non-provenpackager
         app = TestApp(main({}, testing=u'ralph', session=self.db, **self.app_settings))
         post_data = dict(update=nvr, request='stable',
                          csrf_token=app.get('/csrf').json_body['csrf_token'])
         res = app.post_json('/updates/%s/request' % str(nvr), post_data, status=400)
-
```
bowlofeggs: This change seems unrelated to the goal of the pull request - I prefer to keep commits atomic.

bodhi/tests/server/services/test_updates.py
```diff
@@ -509,17 +510,18 @@ def test_provenpackager_request_privs(self, publish, *args):
...
-        # Ensure we can't push it until it meets the requirements
+        # Ensure we can't push it unitl it meets the requirements
         eq_(res.json_body['status'], 'error')
-        eq_(res.json_body['errors'][0]['description'], config.get('not_yet_tested_msg'))
+        eq_(res.json_body['errors'][0]['description'],
+            config.get('not_yet_tested_msg'))
```
bowlofeggs: This change seems unrelated as well.

bodhi/tests/server/services/test_updates.py
```diff
@@ -599,6 +601,195 @@ def test_provenpackager_request_privs(self, publish, *args):
         res = app.get('/updates/%s' % str(nvr), status=200)
         eq_(res.json_body['can_edit'], False)
 
+    @mock.patch(**mock_taskotron_results)
+    @mock.patch(**mock_valid_requirements)
+    @mock.patch('bodhi.server.notifications.publish')
+    @mock.patch.dict('bodhi.server.config.config', {'ci.required': True})
```
bowlofeggs: To be more realistic, you should use ```"True"``` or ```"true"``` instead of ```True```.

bodhi/tests/server/services/test_updates.py
```diff
@@ -599,6 +601,195 @@ def test_provenpackager_request_privs(self, publish, *args):
...
+    def test_provenpackager_request_queued_in_ci(self, publish, *args):
+        "Ensure provenpackagers can change the request for any update"
```
bowlofeggs: Can we convert this to a docblock instead of a normal string (triple quotes)? Also, the name of the test doesn't seem to correlate with this description.

bodhi/tests/server/services/test_updates.py
```diff
@@ -599,6 +601,195 @@ def test_provenpackager_request_privs(self, publish, *args):
...
+        nvr = u'bodhi-2.1-1.fc17'
+        user = User(name=u'bob')
+        self.db.add(user)
+        self.db.add(User(name=u'ralph'))  # Add a non proventester
+        self.db.add(User(name=u'someuser'))  # An unrelated user with no privs
```
bowlofeggs: Why create this user? I don't see it getting used below.

bodhi/tests/server/services/test_updates.py
```diff
@@ -599,6 +601,195 @@ def test_provenpackager_request_privs(self, publish, *args):
...
 
+    @mock.patch(**mock_taskotron_results)
+    @mock.patch(**mock_valid_requirements)
+    @mock.patch('bodhi.server.notifications.publish')
+    @mock.patch.dict('bodhi.server.config.config', {'ci.required': True})
+    def test_provenpackager_request_queued_in_ci(self, publish, *args):
+        "Ensure provenpackagers can change the request for any update"
+        nvr = u'bodhi-2.1-1.fc17'
+        user = User(name=u'bob')
+        self.db.add(user)
```
bowlofeggs: Why create this user?

bodhi/tests/server/services/test_updates.py
```diff
@@ -599,6 +601,195 @@ def test_provenpackager_request_privs(self, publish, *args):
...
+        self.db.add(User(name=u'ralph'))  # Add a non proventester
+        self.db.add(User(name=u'someuser'))  # An unrelated user with no privs
+        self.db.flush()
+        group = self.db.query(Group).filter_by(name=u'provenpackager').one()
+        user.groups.append(group)
```
bowlofeggs: This test's name and description say it has something to do with proven packagers, but the user that gets added to the provenpackager group isn't the user that performs actions in the test.

bodhi/tests/server/services/test_updates.py
```diff
@@ -599,6 +601,195 @@ def test_provenpackager_request_privs(self, publish, *args):
...
+
+        # Ensure we can't push it until it passed CI
+        eq_(res.json_body['status'], 'error')
+        eq_(res.json_body['errors'][0]['description'],
+            'Requirement not met CI did not pass on this update')
+
+    @mock.patch(**mock_taskotron_results)
+    @mock.patch(**mock_valid_requirements)
+    @mock.patch('bodhi.server.notifications.publish')
+    @mock.patch.dict('bodhi.server.config.config', {'ci.required': True})
```
bowlofeggs: Be sure to make this a string.

bodhi/tests/server/services/test_updates.py
```diff
@@ -599,6 +601,195 @@ def test_provenpackager_request_privs(self, publish, *args):
...
+    def test_provenpackager_request_running_in_ci(self, publish, *args):
+        "Ensure provenpackagers can change the request for any update"
```
bowlofeggs: I have similar comments about the docblock and name here - this test doesn't' seem to do anything with proven packagers.

bodhi/tests/server/services/test_updates.py
```diff
@@ -599,6 +601,195 @@ def test_provenpackager_request_privs(self, publish, *args):
...
+        nvr = u'bodhi-2.1-1.fc17'
+        user = User(name=u'bob')
+        self.db.add(user)
+        self.db.add(User(name=u'ralph'))  # Add a non proventester
+        self.db.add(User(name=u'someuser'))  # An unrelated user with no privs
+        self.db.flush()
+        group = self.db.query(Group).filter_by(name=u'provenpackager').one()
+        user.groups.append(group)
```
bowlofeggs: These extra users don't seem to get used.

bodhi/tests/server/services/test_updates.py
```diff
@@ -599,6 +601,195 @@ def test_provenpackager_request_privs(self, publish, *args):
...
+
+        app = TestApp(main({}, testing=u'ralph', session=self.db, **self.app_settings))
+        up_data = self.get_update(nvr)
+        up_data['csrf_token'] = app.get('/csrf').json_body['csrf_token']
+        res = app.post_json('/updates/', up_data)
+        assert 'does not have commit access to bodhi' not in res, res
```
bowlofeggs: You don't need this assert statement, because later things would fail if the update weren't created successfully.

bodhi/tests/server/services/test_updates.py
```diff
@@ -599,6 +601,195 @@ def test_provenpackager_request_privs(self, publish, *args):
...
```
bowlofeggs: You don't need this assert statement, because later things would fail if the update weren't created successfully.

bodhi/tests/server/services/test_updates.py
```diff
@@ -599,6 +601,195 @@ def test_provenpackager_request_privs(self, publish, *args):
...
+        eq_(res.json_body['status'], 'error')
+        eq_(res.json_body['errors'][0]['description'],
+            'Requirement not met CI did not pass on this update')
+
+    @mock.patch(**mock_taskotron_results)
+    @mock.patch(**mock_valid_requirements)
+    @mock.patch('bodhi.server.notifications.publish')
+    @mock.patch.dict('bodhi.server.config.config', {'ci.required': True})
+    def test_provenpackager_request_failed_ci(self, publish, *args):
+        "Ensure provenpackagers can change the request for any update"
```
bowlofeggs: Similar docblock and test name comments here.

bodhi/tests/server/services/test_updates.py
```diff
@@ -599,6 +601,195 @@ def test_provenpackager_request_privs(self, publish, *args):
...
+
+        # Ensure we can't push it until it passed CI
+        eq_(res.json_body['status'], 'error')
+        eq_(res.json_body['errors'][0]['description'],
+            'Requirement not met CI did not pass on this update')
+
+    @mock.patch(**mock_taskotron_results)
+    @mock.patch(**mock_valid_requirements)
+    @mock.patch('bodhi.server.notifications.publish')
+    @mock.patch.dict('bodhi.server.config.config', {'ci.required': True})
```
bowlofeggs: Be sure to use a string instead of a bool here.

bodhi/tests/server/services/test_updates.py
```diff
@@ -599,6 +601,195 @@ def test_provenpackager_request_privs(self, publish, *args):
...
+    def test_provenpackager_request_failed_ci(self, publish, *args):
+        "Ensure provenpackagers can change the request for any update"
+        nvr = u'bodhi-2.1-1.fc17'
+        user = User(name=u'bob')
+        self.db.add(user)
+        self.db.add(User(name=u'ralph'))  # Add a non proventester
+        self.db.add(User(name=u'someuser'))  # An unrelated user with no privs
+        self.db.flush()
+        group = self.db.query(Group).filter_by(name=u'provenpackager').one()
+        user.groups.append(group)
```
bowlofeggs: Similarly, these users aren't used.

bodhi/tests/server/services/test_updates.py
```diff
@@ -599,6 +601,195 @@ def test_provenpackager_request_privs(self, publish, *args):
...
+
+        app = TestApp(main({}, testing=u'ralph', session=self.db, **self.app_settings))
+        up_data = self.get_update(nvr)
+        up_data['csrf_token'] = app.get('/csrf').json_body['csrf_token']
+        res = app.post_json('/updates/', up_data)
+        assert 'does not have commit access to bodhi' not in res, res
```
bowlofeggs: We should drop this to keep the test simpler and easier to follow.

bodhi/tests/server/services/test_updates.py
```diff
@@ -599,6 +601,195 @@ def test_provenpackager_request_privs(self, publish, *args):
...
+        res = app.post_json('/updates/%s/request' % str(nvr), post_data, status=400)
+
+        # Ensure we can't push it until it passed CI
+        eq_(res.json_body['status'], 'error')
+        eq_(res.json_body['errors'][0]['description'],
+            'Requirement not met CI did not pass on this update')
+
+    @mock.patch(**mock_taskotron_results)
+    @mock.patch(**mock_valid_requirements)
+    @mock.patch('bodhi.server.notifications.publish')
```
bowlofeggs, line 813: Shouldn't this test also mock the ci being enabled?

bodhi/tests/server/services/test_updates.py
```diff
@@ -599,6 +601,195 @@ def test_provenpackager_request_privs(self, publish, *args):
...
+    def test_provenpackager_request_ignored_by_ci(self, publish, *args):
+        "Ensure provenpackagers can change the request for any update"
```
bowlofeggs: Name and docblock. I'll stop pointing this, the extra users, and the assert statement out - please fix those things in all the tests.

bodhi/tests/server/services/test_updates.py
```diff
@@ -599,6 +601,195 @@ def test_provenpackager_request_privs(self, publish, *args):
...
+        # Try and submit the update to stable as a non-provenpackager
+        app = TestApp(main({}, testing=u'ralph', session=self.db, **self.app_settings))
+        post_data = dict(update=nvr, request='stable',
+                         csrf_token=app.get('/csrf').json_body['csrf_token'])
+        res = app.post_json('/updates/%s/request' % str(nvr), post_data, status=400)
+
+        # Ensure we can't push it until it passed CI
+        eq_(res.json_body['status'], 'error')
+        eq_(res.json_body['errors'][0]['description'],
+            config.get('not_yet_tested_msg'))
```
bowlofeggs: Isn't this the incorrect behavior? Shouldn't the ignored state allow the update to go to stable?

bodhi/tests/server/services/test_updates.py
```diff
@@ -599,6 +601,195 @@ def test_provenpackager_request_privs(self, publish, *args):
...
+
+    @mock.patch(**mock_taskotron_results)
+    @mock.patch(**mock_valid_requirements)
+    @mock.patch('bodhi.server.notifications.publish')
+    @mock.patch.dict('bodhi.server.config.config', {'ci.required': True})
+    def test_provenpackager_request_no_ci(self, publish, *args):
```
bowlofeggs: This test name says "no_ci", but the mock is enabling CI.

bodhi/tests/server/services/test_updates.py
```diff
@@ -599,6 +601,195 @@ def test_provenpackager_request_privs(self, publish, *args):
...
+        app = TestApp(main({}, testing=u'ralph', session=self.db, **self.app_settings))
+        up_data = self.get_update(nvr)
+        up_data['csrf_token'] = app.get('/csrf').json_body['csrf_token']
+        res = app.post_json('/updates/', up_data)
+        assert 'does not have commit access to bodhi' not in res, res
+        publish.assert_called_once_with(
+            topic='update.request.testing', msg=mock.ANY)
+
+        build = self.db.query(RpmBuild).filter_by(nvr=nvr).one()
+        build.ci_status = None
```
bowlofeggs: In what case would you expect the ci_status to be None? If an update is created when CI is enabled, shouldn't its CI status be "queued"?

bodhi/tests/server/services/test_updates.py
```diff
@@ -2238,6 +2429,8 @@ def test_request_to_stable_based_on_stable_karma(self, *args):
         up = Update.get(nvr, self.db)
         up.status = UpdateStatus.testing
         up.request = None
+        eq_(len(up.builds), 1)
```
bowlofeggs, line 2565: I don't think we need this assertion or the similar ones below. This test module isn't a good example of the following ideology, but I believe that tests are best when they are as short as possible and test a condition with "laser" precision. A lot of the tests in this module are huge and are really hard to understand because they have a lot of extraneous code and extraneous assertions, which make it unclear what their purpose is.

bodhi/tests/server/test_models.py
```diff
@@ -626,6 +626,81 @@ def test_days_to_stable_not_meets_testing_requirements_with_date_testing(self):
 
         eq_(update.days_to_stable, 3)
 
+    @mock.patch.dict('bodhi.server.config.config', {'ci.required': True})
```
bowlofeggs, line 776: Be sure to use a string "tRuE" (or whatever weird capitalization you like) here, and elsewhere in this module.

bodhi/tests/server/test_models.py
```diff
@@ -626,6 +626,81 @@ def test_days_to_stable_not_meets_testing_requirements_with_date_testing(self):
...
+        update = self.obj
+        update.autokarma = False
+        update.stable_karma = 1
+        update.builds[0].ci_status = None
+        update.comment(self.db, u'I found $100 after applying this update.', karma=1,
+                       author=u'bowlofeggs')
+        # Assert that our preconditions from the docblock are correct.
+        eq_(update.meets_testing_requirements, False)
+
+    def test_ci_ci_required_off(self):
```
bowlofeggs: It could be useful to mock the setting being set to ```"False"``` here to test that the bool is properly parsed.

bowlofeggs (changes requested):
In addition to my comments, I think it would be very useful to write unit tests specifically on the new ci_status and ci_passed methods - i.e., tests that just focus on those methods and exercise all their if branches and set logic. It's much easier to test them that way than to use what I would say are more "integration" tests in the test_updates.py module.

alembic/versions/8eaacb38b036_add_the_ci_url_field_to_builds.py
pypingou: Currently, I have no idea what the URL will look like, it may end up being a link to jenkins or resultsdb, no idea atm

bodhi/server/models.py
pypingou: When looking at the output of btest, I didn't see this line to be missing, how do you check it?

bowlofeggs:
On Thu, 2017-05-18 at 08:09 -0700, Pierre-Yves Chibon wrote:
> Currently, I have no idea what the URL will look like, it may end up
> being a link to jenkins or resultsdb, no idea atm

Cool.

bowlofeggs:
On Thu, 2017-05-18 at 08:14 -0700, Pierre-Yves Chibon wrote:
> When looking at the output of btest, I didn't see this line to be
> missing, how do you check it?

I just looked at the output of btest, but it's possible I commented on
the wrong line or that I misread it. If you're sure it's covered, feel
free to ignore my comment and I'll recheck it when you are ready for
re-review!

bowlofeggs:
I just made #1544 to solve the problem I mentioned here where Bodhi configs are always strings. If we make sure to merge #1544 before this one, then you don't have to fix all those places I pointed out that problem. It also gives you a nice way to express the default for each setting, so you can also assume that a default exists and just do ```config['setting_string']```.

bodhi/server/models.py
pypingou: We do it earlier, so I just flushed after adjusting the build, kinda in case :)

bodhi/server/models.py
pypingou: ok

bodhi/server/models.py
pypingou: adjusted

bodhi/server/models.py
pypingou: Dropped the ``, False``

bodhi/server/templates/update.html
pypingou: Following #1489 , this got simplified to ``% if request.registry.settings.get('ci.required'):``

bodhi/server/templates/update.html
pypingou: There is an ``% else`` so it won't be empty.

Why would settings not be defined? (it is used elsewhere in the template, so I guess we would have more troubles if it isn't)

bodhi/server/templates/update.html
pypingou: Dropped, the ``, 'False') == 'True'``

bodhi/server/templates/update.html
pypingou: I believe it is needed

bodhi/server/util.py
pypingou: sure

bodhi/tests/server/test_models.py
pypingou: I guess with #1489 this can be ignored?

bodhi/tests/server/test_models.py
pypingou: same here, wrt #1489

pypingou (commented):
I am rebasing and pushing the changes you asked, I think we need to clear the situation around the nullable condition, might be easier to do this on IRC than here.

bowlofeggs:
@pypingou and I talked about this a bit today and we came to a consensus about how we would handle the ci_status database column.

Pierre-Yves, let me know when you want me to re-review this PR!

pypingou:
@bowlofeggs let me know what you think of this :)

bodhi/server/models.py
```diff
@@ -1376,6 +1397,47 @@ def content_type(self):
...
+        """
+
+        ci_status = set([
+            build.ci_status for build in self.builds
+            if build.ci_status is not None
+        ])
+        if ci_status == set():
+            return CiStatus.ignored
+        if ci_status == set([CiStatus.ignored]):
+            return CiStatus.ignored
```
bowlofeggs: This line is missing coverage.

bodhi/server/models.py
```diff
@@ -1376,6 +1397,47 @@ def content_type(self):
...
+        if ci_status == set([CiStatus.passed]):
+            return CiStatus.passed
```
bowlofeggs: This line is missing test coverage.

bodhi/server/util.py
```diff
@@ -315,6 +315,31 @@ def status2html(context, status):
...
+    }[str(status)]
+    if status is None:
+        status = 'info awaited'
+    return u"<span class='label label-%s'>Tests %s</span>" % (cls, status)
```
bowlofeggs: This function is missing test coverage.

bowlofeggs (changes requested):
You can easily test your PR for test coverage with this command:

```
[vagrant@bodhi-dev bodhi]$ diff-cover coverage.xml --compare-branch=upstream/develop --fail-under=100                 
-------------
Diff Coverage
Diff: upstream/develop...HEAD, staged, and unstaged changes
-------------
bodhi/server/models.py (94.7%): Missing lines 1429,1431
bodhi/server/util.py (20.0%): Missing lines 330,338-340
-------------
Total:   43 lines
Missing: 6 lines
Coverage: 86%
-------------
```

You can ```sudo dnf install python3-diff-cover``` to get that command. There's an open PR from @mprahl to document this.

bodhi/server/models.py
```diff
@@ -738,6 +748,8 @@ class Build(Base):
             of.
         type (int): The polymorphic identify of the row. This is used by sqlalchemy to identify
             which subclass of Build to use.
+        ci_status (EnumSymbol): The CI status of the update. This must be one of the values
+            defined in :class:`CiStatus` or ``None``.
```
bowlofeggs: It might be good to document that ```None``` is used for Updates that were created when the CI integration was not enabled.

bodhi/server/models.py
```diff
@@ -380,6 +380,16 @@ class UpdateStatus(DeclEnum):
...
 
 
+class CiStatus(DeclEnum):
+    """ This class lists the different status the ci_status flag can have. """
+    ignored = 'ignored', 'Ignored'
+    queued = 'queued', 'Queued'
+    running = 'running', 'Running'
+    passed = 'passed', 'Passed'
+    failed = 'failed', 'Failed'
+    waiting = 'waiting', 'Waiting'
```
bowlofeggs: Could we put some comments over these to explain what each one means? I think the comments you had in that fpaste recently would be perfect here.

bodhi/server/util.py
```diff
@@ -315,6 +315,31 @@ def status2html(context, status):
     return "<span class='label label-%s'>%s</span>" % (cls, status)
 
 
+def ci_status2html(context, status):
+    """ Convert the specified status into a html string used to present the
+    CI status in a human friendly way.
+
+    Args:
+        context ():
```
bowlofeggs: I recently learned that the context is a ```mako.runtime.Context```, so we could document that here.

bodhi/tests/server/services/test_updates.py
bowlofeggs: Please add this line back since it's an unrelated change.

bodhi/tests/server/services/test_updates.py
bowlofeggs: Please also restore this change since it is unrelated, or put it into its own commit.

bodhi/tests/server/services/test_updates.py
bowlofeggs: Please correct the test name and the docblock (so that it's triple quoted).

bodhi/tests/server/services/test_updates.py
bowlofeggs: It seems you can still drop this user.

bodhi/tests/server/services/test_updates.py
bowlofeggs: And this user.

bodhi/tests/server/services/test_updates.py
bowlofeggs: This user doesn't need to be added to provenpackager since it's not the user performing actions later.

bodhi/tests/server/services/test_updates.py
bowlofeggs: You don't need this assert statement.

bodhi/tests/server/services/test_updates.py
bowlofeggs: The docblock and name here need fixing.

bodhi/tests/server/services/test_updates.py
bowlofeggs: The extra users aren't needed, let's drop them to simplify the test.

bodhi/tests/server/services/test_updates.py
bowlofeggs: You should drop this assertion.

bodhi/tests/server/services/test_updates.py
bowlofeggs: This docblock and test name need to be fixed.

bodhi/tests/server/services/test_updates.py
bowlofeggs: These users aren't used.

bodhi/tests/server/services/test_updates.py
bowlofeggs: We don't need this assertion.

bodhi/tests/server/services/test_updates.py
bowlofeggs: Should we mock the CI being enabled?

bodhi/tests/server/services/test_updates.py
bowlofeggs: Name and docblock. I'll stop pointing this, the extra users, and the assert statement out - please fix those things in all the tests.

bodhi/tests/server/services/test_updates.py
bowlofeggs: Isn't this the incorrect behavior? Shouldn't the ignored state allow the update to go to stable?

bodhi/tests/server/services/test_updates.py
bowlofeggs: This test name says "no_ci", but the mock is enabling CI.

bodhi/tests/server/services/test_updates.py
bowlofeggs: I don't think we need this assertion or the similar ones below. This test module isn't a good example of the following ideology, but I believe that tests are best when they are as short as possible and test a condition with "laser" precision. A lot of the tests in this module are huge and are really hard to understand because they have a lot of extraneous code and extraneous assertions, which make it unclear what their purpose is.

bodhi/tests/server/test_models.py
bowlofeggs: @pypingou sure!

bowlofeggs (changes requested):
There are still a lot of comments I made earlier that need to be addressed. I repeated a few of them here.

bodhi/server/models.py
pypingou: sure

bodhi/server/models.py
pypingou: sure

bodhi/server/models.py
pypingou: I think this is now addressed

bodhi/server/models.py
pypingou: I believe this got fixed

bodhi/server/util.py
pypingou: Done

bodhi/server/util.py
pypingou: Not sure using the enum here would be easier, so I'd prefer to leave it as it is.

bodhi/server/util.py
pypingou: I believe this is no longer a concern

bodhi/tests/server/services/test_updates.py
pypingou: sure if you want

bodhi/tests/server/services/test_updates.py
pypingou: ok

bodhi/tests/server/services/test_updates.py
pypingou: afaiu this is no longer applicable, no?

bodhi/tests/server/services/test_updates.py
pypingou: I'll triple quote it but note that this is consistent with the rest of the file (and triple quoting it will make it inconsistent)

bodhi/tests/server/services/test_updates.py
pypingou: Then it's no longer testing the provenpackager case

bodhi/tests/server/services/test_updates.py
pypingou: I can remove it but that's a left over from ``test_provenpackager_request_privs``

bodhi/tests/server/services/test_updates.py
pypingou: I removed ralph and switched to the bob user instead

bodhi/tests/server/services/test_updates.py
pypingou: Same as above, switched from ralph to bob

bodhi/tests/server/services/test_updates.py
pypingou: Removed, was coming from ``test_provenpackager_request_privs``

bodhi/tests/server/services/test_updates.py
pypingou: I believe this no longer applies

bodhi/tests/server/services/test_updates.py
pypingou: fixed

bodhi/tests/server/services/test_updates.py
pypingou: fixed

bodhi/tests/server/services/test_updates.py
pypingou: fixed

bodhi/tests/server/services/test_updates.py
pypingou: I believe this no longer applies

bodhi/tests/server/services/test_updates.py
pypingou: fixed

bodhi/tests/server/services/test_updates.py
pypingou: fixed

bodhi/tests/server/services/test_updates.py
pypingou: fixed

bodhi/tests/server/services/test_updates.py
pypingou: indeed, fixed

bodhi/tests/server/services/test_updates.py
pypingou: fixed

bodhi/tests/server/services/test_updates.py
pypingou: it is and does, the reason it's not going isn't CI this time

bodhi/tests/server/services/test_updates.py
pypingou: the update has no CI (ci_status is None)

bodhi/tests/server/services/test_updates.py
pypingou: I believe we sorted out this question (also now documented in the doc of the CiStatus object)

bodhi/tests/server/services/test_updates.py
pypingou: I added this line so that it would fail there rather than at the next line which does ``up.builds[0]``, if there are more than one build, they should all be adjusted or the behaviour may differ.

pypingou (commented):
Thanks for the review, going to look at the coverage

pypingou:
On Thu, Jun 01, 2017 at 08:42:43AM -0700, Randy Barlow wrote:
>  [vagrant@bodhi-dev bodhi]$ diff-cover coverage.xml --compare-branch=upstream/develop --fail-under=100                
>  Failure. Coverage is below 100.0%.
>  -------------
>  Diff Coverage
>  Diff: upstream/develop...HEAD, staged, and unstaged changes
>  -------------
>  bodhi/server/models.py (94.7%): Missing lines 1429,1431
>  bodhi/server/util.py (20.0%): Missing lines 330,338-340
>  -------------
>  Total:   43 lines
>  Missing: 6 lines
>  Coverage: 86%
>  -------------

Ok so I guess this is why I got different results. This is what I see in the
output of ``btest``:

````
bodhi/server/models.py                      1499    160    89%   70, 82, 97, 110-111, 151-154, 158-161, 250, 260-266, 493-496, 506, 614-630, 659-660, 815-819, 838-839, 925-926, 928, 941-960, 1283-1284, 1287-1289, 1298, 1323-1324, 1368, 1395, 1557, 1560-1564, 1574-1578, 1599, 1650-1654, 1675-1676, 1695, 1704, 1718, 1736-1737, 1749-1750, 1810-1811, 1818-1820, 1826-1827, 1842-1843, 1942, 1952-1964, 1988, 2043-2048, 2080, 2087, 2099-2100, 2103, 2120, 2125, 2194-2195, 2200-2201, 2289, 2303-2308, 2337, 2369, 2410-2411, 2442-2443, 2445, 2611-2614, 2622, 2636, 2652, 2764-2766, 2836-2837
...
bodhi/server/util.py                         360     68    81%   47-48, 71-79, 84, 115-123, 152, 166, 175-179, 187-189, 218-220, 257, 269, 286, 330-341, 393, 441-449, 453-459, 491, 504-511, 581, 625-645
````

As you can see in `models.py` it sees the lines 1429,1431 as covered.

For util.py I guess it's just a change of the line numbers, I'll see to add
coverage for them.

File: bodhi/server/models.py (hunk 68 above)
Comment (bowlofeggs) [reply bodhi/server/models.py]: These comments are great, thanks!

File: alembic/versions/8eaacb38b036_add_the_ci_url_field_to_builds.py (hunk 21 above)
Comment (bowlofeggs) [reply alembic/versions/8eaacb38b036_add_the_ci_url_field_to_builds.py:21]: Would it be good to make this a fixed width column? I guess since we don't know the URL scheme yet, we also don't know what length to put here. We could migrate it later I suppose.

File: bodhi/server/models.py [hunk 70]
```diff
@@ -380,6 +380,23 @@ class UpdateStatus(DeclEnum):
...
+    # ignored: a message from the CI system said that this build is ignored
+    ignored = 'ignored', 'Ignored'
+    # queued: a message from the CI system said that the tests for this build have been queued
+    queued = 'queued', 'Queued'
+    # running: a message from the CI system said that the tests for this build are running
+    running = 'running', 'Running'
+    # passed: a message from the CI system said that the tests for this build have passed
+    passed = 'passed', 'Passed'
+    # failed: a message from the CI system said that the tests for this build have been failed
+    failed = 'failed', 'Failed'
```
bowlofeggs, line 391: Very clear comments, thanks!

bodhi/server/models.py
bowlofeggs: I wrote a test to hit this line (and a bunch of tests just to focus specifically on this method).

bodhi/tests/server/services/test_updates.py
bowlofeggs: @pypingou Correct - my config PR will make this OK once it's merged, so no need to worry about it.

bodhi/tests/server/services/test_updates.py
bowlofeggs: Correct.

bowlofeggs (commented):
I've added a commit that gets us up to 100% test coverage on this patch and also makes a few tweaks here and there:

```
-------------
Diff Coverage
-------------
bodhi/server/models.py (100%)
bodhi/server/util.py (100%)
-------------
Total:   40 lines
Missing: 0 lines
Coverage: 100%
-------------
```

Can you review my commit? If you agree with it, let's squash this whole PR down to one commit and merge!

bowlofeggs:
https://github.com/fedora-infra/bodhi/issues/1544 is now merged, so let's make sure the new settings in this PR are added to ```bodhi.server.config.BodhiConfig._defaults```. Additionally, all the code that uses the settings should not define defaults anymore (i.e., don't do ```config.get('setting_name', 'default')```).

pypingou:
👍 to your commit, thanks for this!

Feel free to go ahead and squash/merge otherwise I will try to do it tomorrow morning :)

pypingou:
``bodhi/server/config.py`` adjusted

All commits squashed into one with a rather extensive commit message, though we can likely still improve it :)

bowlofeggs:
@pypingou I am adding this patch to the commit:

```
diff --git a/development.ini.example b/development.ini.example
index 9e14795..eee08c2 100644
--- a/development.ini.example
+++ b/development.ini.example
@@ -56,7 +56,7 @@ badge_ids = binary-star|both-bull-and-self-transcended-tester-viii|catching-the-
 
 
 ##
-## Wiki Test Cases
+## Testing
 ##
 
 ## Query the wiki for test cases
@@ -64,6 +64,16 @@ badge_ids = binary-star|both-bull-and-self-transcended-tester-viii|catching-the-
 # wiki_url = https://fedoraproject.org/w/api.php
 # test_case_base_url = https://fedoraproject.org/wiki/
 
+# URL of the resultsdb for integrating checks and stuff
+# resultsdb_api_url = https://taskotron.fedoraproject.org/resultsdb_api/
+
+# Set this to True to enable gating based on Continuous Integration test results.
+# ci.required = False
+
+# If this is set to a URL, a "More information about CI" link will appear on update pages for users
+# to click and learn more.
+# ci.url =
+
 # Email domain to prepend usernames to
 # default_email_domain = fedoraproject.org
 
@@ -208,9 +218,6 @@ dogpile.cache.arguments.filename = %(here)s/dogpile-cache.dbm
 # URL of where users should go to set up their notifications
 # fmn_url = https://apps.fedoraproject.org/notifications/
 
-# URL of the resultsdb for integrating checks and stuff
-# resultsdb_api_url = https://taskotron.fedoraproject.org/resultsdb_api/
-
 # fedmenu.url = https://apps.fedoraproject.org/fedmenu
 # fedmenu.data_url = https://apps.fedoraproject.org/js/data.js
 
index 019868b..e45da9e 100644
--- a/production.ini
+++ b/production.ini
@@ -58,7 +58,7 @@ use = egg:bodhi-server
 
 
 ##
-## Wiki Test Cases
+## Testing
 ##
 
 ## Query the wiki for test cases
@@ -66,6 +66,16 @@ use = egg:bodhi-server
 # wiki_url = https://fedoraproject.org/w/api.php
 # test_case_base_url = https://fedoraproject.org/wiki/
 
+# URL of the resultsdb for integrating checks and stuff
+# resultsdb_api_url = https://taskotron.fedoraproject.org/resultsdb_api/
+
+# Set this to True to enable gating based on Continuous Integration test results.
+# ci.required = False
+
+# If this is set to a URL, a "More information about CI" link will appear on update pages for users
+# to click and learn more.
+# ci.url =
+
 # Email domain to prepend usernames to
 # default_email_domain = fedoraproject.org
 
@@ -208,9 +218,6 @@ use = egg:bodhi-server
 # URL of where users should go to set up their notifications
 # fmn_url = https://apps.fedoraproject.org/notifications/
 
-# URL of the resultsdb for integrating checks and stuff
-# resultsdb_api_url = https://taskotron.fedoraproject.org/resultsdb_api/
-
 # fedmenu.url = https://apps.fedoraproject.org/fedmenu
 # fedmenu.data_url = https://apps.fedoraproject.org/js/data.js
```

bowlofeggs:
There are 5 failing unit tests. Maybe some things changed during the squash?

bowlofeggs (changes requested):
Hmmm, I wonder if something odd happened when squashing the commits? There are 5 failing unit tests, and now we don't have 100% coverage anymore:

```
$ diff-cover coverage.xml --compare-branch=develop --fail-under=100
-------------
Diff Coverage
-------------
bodhi/server/models.py (89.2%): Missing lines 1295,1297-1298,2259
bodhi/server/util.py (100%)
-------------
Total:   42 lines
Missing: 4 lines
Coverage: 90%
-------------
```

bowlofeggs:
I believe this patch will fix the failing tests:

```
index 7ee18c5..11b12e7 100644
@@ -692,7 +692,6 @@ class TestUpdatesService(bodhi.tests.server.functional.base.BaseWSGICase):
-    @mock.patch.dict('bodhi.server.config.config', {'ci.required': True})
     def test_provenpackager_request_update_queued_in_ci(self, publish, *args):
         """Ensure provenpackagers cannot request changes for any update which
         CI status is `queued`"""
@@ -701,6 +700,7 @@ class TestUpdatesService(bodhi.tests.server.functional.base.BaseWSGICase):
         self.db.add(user)
         group = self.db.query(Group).filter_by(name=u'provenpackager').one()
         user.groups.append(group)
+        self.app_settings['ci.required'] = True
 
         app = TestApp(main({}, testing=u'bob', session=self.db, **self.app_settings))
         up_data = self.get_update(nvr)
@@ -726,7 +726,6 @@ class TestUpdatesService(bodhi.tests.server.functional.base.BaseWSGICase):
-    @mock.patch.dict('bodhi.server.config.config', {'ci.required': True})
     def test_provenpackager_request_update_running_in_ci(self, publish, *args):
         """Ensure provenpackagers cannot request changes for any update which
         CI status is `running`"""
@@ -735,6 +734,7 @@ class TestUpdatesService(bodhi.tests.server.functional.base.BaseWSGICase):
         self.db.add(user)
         group = self.db.query(Group).filter_by(name=u'provenpackager').one()
         user.groups.append(group)
+        self.app_settings['ci.required'] = True
 
         app = TestApp(main({}, testing=u'bob', session=self.db, **self.app_settings))
         up_data = self.get_update(nvr)
@@ -760,7 +760,6 @@ class TestUpdatesService(bodhi.tests.server.functional.base.BaseWSGICase):
-    @mock.patch.dict('bodhi.server.config.config', {'ci.required': True})
     def test_provenpackager_request_update_failed_ci(self, publish, *args):
         """Ensure provenpackagers cannot request changes for any update which
         CI status is `failed`"""
@@ -769,6 +768,7 @@ class TestUpdatesService(bodhi.tests.server.functional.base.BaseWSGICase):
         self.db.add(user)
         group = self.db.query(Group).filter_by(name=u'provenpackager').one()
         user.groups.append(group)
+        self.app_settings['ci.required'] = True
 
         app = TestApp(main({}, testing=u'bob', session=self.db, **self.app_settings))
         up_data = self.get_update(nvr)
@@ -794,7 +794,6 @@ class TestUpdatesService(bodhi.tests.server.functional.base.BaseWSGICase):
-    @mock.patch.dict('bodhi.server.config.config', {'ci.required': True})
     def test_provenpackager_request_update_ignored_by_ci(self, publish, *args):
         """Ensure provenpackagers can request changes for any update which
         CI status is `ignored`"""
@@ -803,6 +802,7 @@ class TestUpdatesService(bodhi.tests.server.functional.base.BaseWSGICase):
         self.db.add(user)
         group = self.db.query(Group).filter_by(name=u'provenpackager').one()
         user.groups.append(group)
+        self.app_settings['ci.required'] = True
 
         app = TestApp(main({}, testing=u'bob', session=self.db, **self.app_settings))
         up_data = self.get_update(nvr)
@@ -829,7 +829,6 @@ class TestUpdatesService(bodhi.tests.server.functional.base.BaseWSGICase):
-    @mock.patch.dict('bodhi.server.config.config', {'ci.required': True})
     def test_provenpackager_request_update_waiting_on_ci(self, publish, *args):
         """Ensure provenpackagers cannot request changes for any update which
         CI status is `waiting`"""
@@ -838,6 +837,7 @@ class TestUpdatesService(bodhi.tests.server.functional.base.BaseWSGICase):
         self.db.add(user)
         group = self.db.query(Group).filter_by(name=u'provenpackager').one()
         user.groups.append(group)
+        self.app_settings['ci.required'] = True
 
         app = TestApp(main({}, testing=u'bob', session=self.db, **self.app_settings))
         up_data = self.get_update(nvr)
@@ -863,7 +863,6 @@ class TestUpdatesService(bodhi.tests.server.functional.base.BaseWSGICase):
-    @mock.patch.dict('bodhi.server.config.config', {'ci.required': True})
     def test_provenpackager_request_update_with_none_ci(self, publish, *args):
         """Ensure provenpackagers can request changes for any update which
         CI status is `None`"""
@@ -872,6 +871,7 @@ class TestUpdatesService(bodhi.tests.server.functional.base.BaseWSGICase):
         self.db.add(user)
         group = self.db.query(Group).filter_by(name=u'provenpackager').one()
         user.groups.append(group)
+        self.app_settings['ci.required'] = True
 
         app = TestApp(main({}, testing=u'bob', session=self.db, **self.app_settings))
         up_data = self.get_update(nvr)
```

I'm currently running through another test and coverage check. If those pass, I'll fix this commit.

pypingou:
🎆 🍾 🎉 👍 💯

PR #1500, opened 2017-05-03
Do not prompt CLI users for the password.

bowlofeggs:
We do not need to prompt CLI users for the password because the
bindings superclass will handle that for us if it is needed. This
commit drops the prompt=True flag on the CLI commands that had it.


cverna (approved):
LGTM

bowlofeggs:
This has been cherry picked to the 2.6 branch as 3dd907bafccd7e7f36eae3951e1a6b85626e26f6.

PR #1502, opened 2017-05-03
Modify BodhiClient to retry on AuthError after deleting session.

### LINKED ISSUE #1474: bodhi-client cannot handle expiring sessions
The bodhi client cannot work with expiring sessions. Once its session expires, it will never attempt to get a new session but will instead assume it can use the existing session. Thus, it does not matter if the user types the correct password or not:

```
$ time fedpkg update
Password:
  File "/usr/lib/python2.7/site-packages/bodhi/client/__init__.py", line 132, in new
    resp = client.save(**update)
  File "/usr/lib/python2.7/site-packages/bodhi/client/bindings.py", line 66, in wrapper
  File "/usr/lib/python2.7/site-packages/bodhi/client/bindings.py", line 161, in save
    data=kwargs)
  File "/usr/lib/python2.7/site-packages/fedora/client/openidbaseclient.py", line 254, in send_request
    'output': to_bytes(output.text),
ServerError: ServerError(https://bodhi.fedoraproject.org/updates/, 403, Error returned from json module while processing https://bodhi.fedoraproject.org/updates/: Expecting value: line 6 column 1 (char 5)
```

# Workaround

Removing your cached sessions will work around the problem:

```
$ rm ~/.fedora/openidbaseclient-sessions.cache
```

Note that this workaround will only work until your new session expires, and then you will hit this problem again (and you can ```rm``` again to workaround once more). Fedora's Bodhi instance is configured to expire sessions after 24 hours.

bowlofeggs:
fixes #1474 

This depends on https://github.com/fedora-infra/python-fedora/pull/197

### DIFF
--- a/bodhi/client/bindings.py
+++ b/bodhi/client/bindings.py
@@ -36,7 +36,7 @@ import textwrap
 
 import six
 
-from fedora.client import OpenIdBaseClient, FedoraClientError
+from fedora.client import AuthError, OpenIdBaseClient, FedoraClientError
 import fedora.client.openidproxyclient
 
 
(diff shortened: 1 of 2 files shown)


jeremycline (approved):
Looks nice and tidy to me, I like the comment explaining the error handling! :shipit:

PR #1504, opened 2017-05-04
Move the type icons to the left of the update titles so they align.

bowlofeggs:

### DIFF
--- a/bodhi/server/templates/fragments.html
+++ b/bodhi/server/templates/fragments.html
@@ -77,7 +77,7 @@
 </%def>
 
 <%def name="update(update, display_user=True, display_request=False, display_release=True, display_karma=True)">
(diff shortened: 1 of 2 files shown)


pypingou:
👍 for me though a screenshot would be nice :)

ryanlerch:
I like this idea, but i noticed that you opened issue #1506 about this.

There is a font-awesome class "fa-fw" to make the icons fixed width, for example:

```
<i class="fa fa-fw fa-shield"></i>
```

ryanlerch:
@pypingou FWIW, issue #1506 has a screenshot of how the icons look on the left...

pypingou:
Still looks fine to me

bowlofeggs:
@ryanlerch Thanks for the tip! I've applied that change in a second commit (since it's a distinct change I decided to keep it atomic).

@pypingou here is a screenshot of what it looks like with both commits applied:

delme1

pypingou:
Cool, looks good to me

PR #1511, opened 2017-05-11
Configure the dev environment to cd to the bodhi folder by default.

bowlofeggs:


pypingou:
👍 though with the btest/bstart/bstop commands that wasn't really bothering me

PR #1516, opened 2017-05-15
fix half-hidden button on smaller (mobile) screen sizes

ryanlerch:
Previously, on the update page, when viewing bodhi
from a mobile device with a narrow screen width,
the "Add Comment & Feedback" button may have been
partally or wholly obscured. This was caused by
an inapproprite use of a float class.

This patch makes the button sit towards the right
without a float, and fixes the issue.


bowlofeggs (approved):
Cool, looks great!

PR #1518, opened 2017-05-15
[WIP] DO NOT MERGE: example of using pungi for ostree

kushaldas:
Putting it here for comments.

### DIFF
--- a/bodhi/server/consumers/masher.py
+++ b/bodhi/server/consumers/masher.py
@@ -27,6 +27,7 @@ import os
 import threading
 import time
 import urllib2
+import subprocess
 from collections import defaultdict
 from datetime import datetime
 
@@ -902,6 +903,36 @@ class MasherThread(threading.Thread):
         mashed_repos = dict([('-'.join(os.path.basename(repo).split('-')[:-1]), repo)
                              for repo in self.state['completed_repos']])
         for tag, mash_path in mashed_repos.items():
+            #  TODO: Find out the actual values for the tag
+            #  also for the code below, should we use the mash_path instead
+            #  of current_workdir?
+            #  * Pungi config repos hsa $arch, does it matter in bodhi?
+            #  * Do we need any other config files in the pungi-fedora config directory before we fireup the pungi
+            if tag.find('f26') != -1:  # This is a F26 build
+                current_workdir = os.path.join(self.path, self.id)
+                repo_dir = os.path.join(current_workdir, "pungi-fedora")
+                branch = "f26"
+                git_checkout(repo_dir, "https://pagure.io/pungi-fedora", branch)
+                # Prepare the Repo URL
+                dirpath = os.path.join(current_workdir, tag)[9:]
+                url = "https://kojipkgs.fedoraproject.org/{0}/$arch/".format(dirpath)
+                pungi_config_file = os.path.join(repo_dir, "fedora-atomic.conf")
+                with open(pungi_config_file) as fobj:
+                    lines = fobj.readlines()
+
+                idx = 0  # The place to insert the new repo path
+                for i, line in enumerate(lines):
+                    if line.find('"source_repo_from": [ "Cloud",') != -1:
+                        idx = i
+                        break
+                if idx != 0:
+                    lines.insert(idx + 1, url)
+                # Now we update to write the new config file
+                with open(pungi_config_file, "w") as fobj:
+                    fobj.write("".join(lines))
+                cmd = "pungi-koji --notification-script=/usr/bin/pungi-fedmsg-notification --config={0}".format(pungi_config_file)
+                system(cmd)
+
             if tag not in atomic_config['releases']:
                 log.warn('Cannot find atomic configuration for %r', tag)
                 continue
@@ -1006,3 +1037,24 @@ class MashThread(threading.Thread):
         else:
             self.success = True
         return out, err, returncode
+
+def system(cmd):
+    """
+    Invoke a shell command.
+    :returns: A tuple of output, err message, and return code
+    """
+    ret = subprocess.Popen(cmd, shell=True, stdin=subprocess.PIPE,
+                stdout=subprocess.PIPE, stderr=subprocess.PIPE, close_fds=True)
+    out, err = ret.communicate()
+    return out, err, ret.returncode
+
+
+def git_checkout(dirpath, url, branch):
+    """Checks out the given url in the directory and given branch    
+    :param dirpath: Path to checkout  
+    :param url: URL to clone
+    :param branch: The branch to checkout
+    :return: None
+    """
+    cmd = "git clone -b %s %s %s" % (branch, url, dirpath)
+    system(cmd)


bodhi/server/consumers/masher.py
```diff
@@ -902,6 +903,36 @@ def compose_atomic_trees(self):
         mashed_repos = dict([('-'.join(os.path.basename(repo).split('-')[:-1]), repo)
                              for repo in self.state['completed_repos']])
         for tag, mash_path in mashed_repos.items():
+            #  TODO: Find out the actual values for the tag
+            #  also for the code below, should we use the mash_path instead
+            #  of current_workdir?
+            #  * Pungi config repos hsa $arch, does it matter in bodhi?
```
lubomir, line 909: Does Bodhi create a repo for multiple architectures? I think it does. Using `$arch` in the URL then allows to have the same config for multiple arches.

bodhi/server/consumers/masher.py
```diff
@@ -902,6 +903,36 @@ def compose_atomic_trees(self):
...
+            #  * Do we need any other config files in the pungi-fedora config directory before we fireup the pungi
```
lubomir, line 910: If you go the `pungi-koji` way, then you should probably clone everything from the repo. The config files can include other files, so it's hard to tell what exactly you need. Admittedly this is not used currently for Fedora, but it would be nice to be future proof.

If you call just the `pungi-make-ostree` command, you would need some config from https://pagure.io/fedora-atomic (particularly the fedora-atomic-docker-host.json file) for ostree and no special files for Pungi itself.

lubomir:
I think this is a good start, but there are a few things that could be improved.

First, I don't think you want to run the whole `pungi-koji` machinery when you are only really interested in updating an ostree repo with packages from an existing YUM repo. Running just `koji runroot pungi-
make-ostree` command should be sufficient.

Second, the approach for putting the repo url into the config file seems really fragile. A better option would be to write a JSON file somewhere and pass it to `pungi-make-ostree` via `--extra-config`. The
contents would be something like

    {"repo": "path to yum repo with packages"}

and Pungi will take care of tweaking the config.

dustymabe:
> I think this is a good start, but there are a few things that could be improved.
> 
> First, I don't think you want to run the whole pungi-koji machinery when you are only really interested in updating an ostree repo with packages from an existing YUM repo. Running just koji runroot pungi- make-ostree command should be sufficient.

I think the desire is that we are basically running the same code regardless of how we start an ostree generation. The best way to do that (that I know of) is to run pungi. If we make a change to the pungi config it would update all invocations of pungi and we wouldn't have to chase down differences.

Can we make pungi re-use an existing repo and *just* do the ostree generation?

Also it is worth noting that we might get to the point where we are not only creating ostrees, but also disk images at this step in the process. If we use pungi directly then all we have to do is modify the pungi config and we get the rest of the functionality. 

> 
> Second, the approach for putting the repo url into the config file seems really fragile. A better option would be to write a JSON file somewhere and pass it to pungi-make-ostree via --extra-config. The
> contents would be something like
> 
> {"repo": "path to yum repo with packages"}
> 
> and Pungi will take care of tweaking the config.

Does that extra-config functionality only work for the case where we are calling pungi-make-ostree directly?

bodhi/server/consumers/masher.py
```diff
@@ -27,6 +27,7 @@
 import threading
 import time
 import urllib2
+import subprocess
```
bowlofeggs, line 30: Small nitpick: The imports here were alphabetic, and you can keep the trend going by moving this between os and threading ☺ Optional.

bodhi/server/consumers/masher.py
```diff
@@ -902,6 +903,36 @@ def compose_atomic_trees(self):
...
+                    if line.find('"source_repo_from": [ "Cloud",') != -1:
+                        idx = i
+                        break
+                if idx != 0:
+                    lines.insert(idx + 1, url)
+                # Now we update to write the new config file
+                with open(pungi_config_file, "w") as fobj:
+                    fobj.write("".join(lines))
+                cmd = "pungi-koji --notification-script=/usr/bin/pungi-fedmsg-notification --config={0}".format(pungi_config_file)
+                system(cmd)
```
bowlofeggs, line 934: This function is quite long. I think it would be good to split it into a helper function or two so they are more targeted. This will also make it easier to write tests.

bodhi/server/consumers/masher.py
```diff
@@ -902,6 +903,36 @@ def compose_atomic_trees(self):
         mashed_repos = dict([('-'.join(os.path.basename(repo).split('-')[:-1]), repo)
                              for repo in self.state['completed_repos']])
         for tag, mash_path in mashed_repos.items():
+            #  TODO: Find out the actual values for the tag
+            #  also for the code below, should we use the mash_path instead
+            #  of current_workdir?
+            #  * Pungi config repos hsa $arch, does it matter in bodhi?
+            #  * Do we need any other config files in the pungi-fedora config directory before we fireup the pungi
+            if tag.find('f26') != -1:  # This is a F26 build
```
bowlofeggs, line 911: I assume this is placeholder?

bodhi/server/consumers/masher.py
```diff
@@ -902,6 +903,36 @@ def compose_atomic_trees(self):
...
+                current_workdir = os.path.join(self.path, self.id)
+                repo_dir = os.path.join(current_workdir, "pungi-fedora")
+                branch = "f26"
+                git_checkout(repo_dir, "https://pagure.io/pungi-fedora", branch)
```
bowlofeggs, line 915: I don't think this is a problem you need to solve with this effort, but I wanted to highlight that git pulls are one of the many things that can fail during a Bodhi mash and cause mashes to fail. It actually happened just yesterday due to a Pagure outage.

We really need to make code that does this handle outages by retrying and so forth. Again, I'm just raising awareness here, you don't need to solve that problem in this PR.

bodhi/server/consumers/masher.py
```diff
@@ -902,6 +903,36 @@ def compose_atomic_trees(self):
...
+                with open(pungi_config_file) as fobj:
+                    lines = fobj.readlines()
+
+                idx = 0  # The place to insert the new repo path
+                for i, line in enumerate(lines):
+                    if line.find('"source_repo_from": [ "Cloud",') != -1:
+                        idx = i
+                        break
+                if idx != 0:
+                    lines.insert(idx + 1, url)
```
bowlofeggs, line 929: fwiw, I think you can probably get rid of the ```idx``` variable - I think ```i``` is still accessible here in this scope (unlike Rust ☺)

bowlofeggs: Why not use a string replace or a regex here with a ```\n```? This way you don't need to read through the file, find an index, and do all this machinery. Optional.

bodhi/server/consumers/masher.py
```diff
@@ -902,6 +903,36 @@ def compose_atomic_trees(self):
...
+                # Now we update to write the new config file
+                with open(pungi_config_file, "w") as fobj:
+                    fobj.write("".join(lines))
+                cmd = "pungi-koji --notification-script=/usr/bin/pungi-fedmsg-notification --config={0}".format(pungi_config_file)
```
bowlofeggs, line 933: This line is a bit wide, and will probably fail the ```flake8``` test. Please trim to 100 characters.

bodhi/server/consumers/masher.py
```diff
@@ -1006,3 +1037,24 @@ def run(self):
         else:
             self.success = True
         return out, err, returncode
+
```
bowlofeggs, line 1040: Two spaces here for PEP-8.

bodhi/server/consumers/masher.py
```diff
@@ -1006,3 +1037,24 @@ def run(self):
...
+def system(cmd):
+    """
+    Invoke a shell command.
+    :returns: A tuple of output, err message, and return code
+    """
+    ret = subprocess.Popen(cmd, shell=True, stdin=subprocess.PIPE,
+                stdout=subprocess.PIPE, stderr=subprocess.PIPE, close_fds=True)
+    out, err = ret.communicate()
+    return out, err, ret.returncode
```
bowlofeggs, line 1049: I think there's already a function to do something like this in ```bodhi.server.util.cmd()```, and I think it is used a few places here in ```masher.py```. I recommend consolidating around the existing one rather than creating a new one.

bodhi/server/consumers/masher.py
```diff
@@ -1006,3 +1037,24 @@ def run(self):
...
+
+def git_checkout(dirpath, url, branch):
+    """Checks out the given url in the directory and given branch    
+    :param dirpath: Path to checkout  
+    :param url: URL to clone
+    :param branch: The branch to checkout
+    :return: None
+    """
+    cmd = "git clone -b %s %s %s" % (branch, url, dirpath)
+    system(cmd)
```
bowlofeggs, line 1060: This is OK, but probably not that necessary. ```git clone``` is already used elsewhere in masher as a one-liner so I'm not sure it's that beneficial to have another function do it. That said, if you really want such a function I'm fine with it, but please refactor the other use of it to also use this function if you do so. Optional.

bowlofeggs (changes requested):
I can't comment much about the way pungi is used, since I am not very familiar with it, but I've got a few thoughts about the code.

bodhi/server/consumers/masher.py
bowlofeggs: I did think of a benefit to having a git checkout function - we could make this function gracefully handle Pagure outages by retrying. I'd recommend retrying for a relatively long time before giving up - maybe even 3-4 hours. It should log to err level on each failure so we get e-mail notifications about the problem.

lubomir:
> I think the desire is that we are basically running the same code regardless of how we start an ostree generation. The best way to do that (that I know of) is to run pungi. If we make a change to the pungi config it would update all invocations of pungi and we wouldn't have to chase down differences.
> 
> Can we make pungi re-use an existing repo and just do the ostree generation?
> 
> Also it is worth noting that we might get to the point where we are not only creating ostrees, but also disk images at this step in the process. If we use pungi directly then all we have to do is modify the pungi config and we get the rest of the functionality.

It would not be very straightforward, but I think it's doable for ostree. You could basically turn everything off and only run the phase that runs `rpm-ostree` with the bodhi repo as . For anything that needs a `boot.iso` image (live media, disk images) this would not work, as the install tree has to be generated from packages in the compose, and in the way you want it to work there would be none.

Alternative solution would be to set pkgset source to use the bodhi repo instead of a koji tag. Regular depsolving would run on that and create repos in the compose, with which you could do anything that Pungi supports. It's a bit of an overhead though.

Neither of these two options would allow you to reuse the exact same config file with a regular compose taking packages from koji tag. (But a lot could be deduplicated by importing shared configuration from other files.)
 
> Does that extra-config functionality only work for the case where we are calling pungi-make-ostree directly?

Yes, it only works for this case. When running via `pungi-koji` all the information is read from main pungi config file.

bodhi/server/consumers/masher.py
dustymabe: a fedora-infra-utils package with a python library for common retryables like this would be nice. is that something that exists or perhaps a feature request that can be made?

bowlofeggs:
On Thu, 2017-05-18 at 06:12 -0700, Dusty Mabe wrote:
> a fedora-infra-utils package with a python library for common
> retryables like this would be nice. is that something that exists or
> perhaps a feature request that can be made?

There is a ```python-fedora``` package (that ```bodhi.client``` already
uses). I don't know whether it has such a utility, but it could be a
reasonable place to put such a thing perhaps.

Then again, the retry thing would be pretty simple, and adding a
dependency on another library makes bug fixing and releasing more
complicated. IMO, the complication is worth it when there is complex
code involved, but for something simple like this I think it might be
less overhead for each project to just do it the way they need it done.

bodhi/server/consumers/masher.py
bowlofeggs: @lubomir I'm not clear on the question - do you mean "Does Bodhi create one repo per architecture?", or "Does Bodhi create one repo that has multiple architectures in it", or something else?

lubomir: @bowlofeggs that's sort of what I meant: If Bodhi creates separate repos for each arch, then the `$arch` placeholder in the URL is needed. If Bodhi creates just one big repo with all architectures, the URL for Pungi should not contain it.

ausil:
@lubomir we want pungi-koji not running pungi-make-ostree directly. There is a few reasons. ! is that if we ran it directly we would need to manage threads so we can parallelise ostree creation. It also lays a foundation for us to move updates and updates-testing repo generation from mash to pungi

bodhi/server/consumers/masher.py
bowlofeggs: @lubomir It does create one repo per arch:

```
$ ls /mnt/koji/mash/updates/f26-updates/
aarch64  armhfp  i386  ppc64  ppc64le  SRPMS  x86_64
```

kushaldas:
@lubomir @ausil @Then should I just clone a given repo of pungi-koji configs, and then create a JSON file to pass the new repo using ```--extra-config``` ? I am ready to throw away all the code to handle the repo/config related parts. @bowlofeggs @dustymabe

lubomir:
@kushaldas If the goal is to run a full compose in terms of `pungi-koji`, then I think a good starting point would be to prepare a config that can be used for that manually. It would probably need to set package set source to repos (created by bodhi), and tweak how gathering works.

dustymabe:
@lubomir @kushaldas - I'll be creating a new pungi config that we can use for this. Inside of the pungi config we can easily embed a constant that can easily be search/replaced for the repo locations. Would that work?

lubomir:
Yup, that sounds like a sound plan.

bowlofeggs:
@kushaldas Are you still planning to work on this?

PR #1522, opened 2017-05-16
Don't show the edit update form if a user can't edit

### LINKED ISSUE #1521: Edit Form accessable by anyone
Currently, anyone can access the edit page for an update. It is viewable even if not logged in, or if you are logged in and not in the edit ACLs for the update.

https://bodhi.fedoraproject.org/updates/FEDORA-2017-8733e65d13/edit

It is not a huge issue, as it does not let you save the changes to the update, but we really shouldnt show the update edit form to people that don't have access.

ryanlerch:
Previously, both users that were not logged in, and
users logged in, but not having permissions to edit
an update could view the edit update form. These
users were correctly unable to submit the form, and
buttons that linked them to the form were not shown,
but if the URL was typed in, the user could still
access the form page.

This commit changes it so users that don't have
edit privs on an update get a 403 rather than the
edit form.

Fixes #1521

### DIFF
--- a/bodhi/server/services/updates.py
+++ b/bodhi/server/services/updates.py
@@ -89,7 +89,11 @@ def get_update(request):
 
 
 @update_edit.get(accept="text/html", renderer="new_update.html",
-                 error_handler=bodhi.server.services.errors.html_handler)
+                 error_handler=bodhi.server.services.errors.html_handler,
+                 permission='edit',
+                 validators=(
+                     validate_acls,
+                 ))
 def get_update_for_editing(request):
     """Return a single update from an id, title, or alias for the edit form"""
     return dict(
--- a/bodhi/tests/server/services/test_updates.py
+++ b/bodhi/tests/server/services/test_updates.py
@@ -446,6 +446,38 @@ class TestNewUpdate(bodhi.tests.server.functional.base.BaseWSGICase):
         self.assertEquals(up.comments[-1].text, expected_comment)
 
 
+class TestEditUpdateForm(bodhi.tests.server.functional.base.BaseWSGICase):
+
+    def test_edit_with_permission(self):
+        """
+        Test a logged in User with permissions on the update can see the form
+        """
+        resp = self.app.get('/updates/FEDORA-2017-a3bbe1a8f2/edit')
+        self.assertIn('Editing an update requires JavaScript', resp)
+
+    def test_edit_without_permission(self):
+        """
+        Test a logged in User without permissions on the update can't see the form
+        """
+        app = TestApp(main({}, testing=u'anonymous', session=self.db, **self.app_settings))
+        resp = app.get('/updates/FEDORA-2017-a3bbe1a8f2/edit', status=400)
+        self.assertIn('anonymous does not have commit access to bodhi', resp)
+
+    def test_edit_not_loggedin(self):
+        """
+        Test a non logged in User can't see the form
+        """
+        anonymous_settings = copy.copy(self.app_settings)
+        anonymous_settings.update({
+            'authtkt.secret': 'whatever',
+            'authtkt.secure': True,
+        })
+        app = TestApp(main({}, session=self.db, **anonymous_settings))
+        resp = app.get('/updates/FEDORA-2017-a3bbe1a8f2/edit', status=403)
+        self.assertIn('<h1>403 <small>Forbidden</small></h1>', resp)
+        self.assertIn('<p class="lead">Access was denied to this resource.</p>', resp)
+
+
 class TestUpdatesService(bodhi.tests.server.functional.base.BaseWSGICase):
 
     def test_home_html(self):


bodhi/server/services/updates.py
```diff
@@ -19,6 +19,8 @@
 from sqlalchemy import func, distinct
 from sqlalchemy.sql import or_
 
+from pyramid.exceptions import HTTPForbidden
```
bowlofeggs: For [PEP-8](https://www.python.org/dev/peps/pep-0008/#imports), this import should be grouped with the other third party imports (like ```sqlalchemy```).

bodhi/server/services/updates.py
```diff
@@ -92,6 +94,12 @@ def get_update(request):
                  error_handler=bodhi.server.services.errors.html_handler)
 def get_update_for_editing(request):
     """Return a single update from an id, title, or alias for the edit form"""
+    proxy_request = bodhi.server.security.ProtectedRequest(request)
+    validate_acls(proxy_request)
+    # If validate_acls produced 0 errors, then we can edit this update.
+    can_edit = len(proxy_request.errors) == 0
+    if not can_edit:
+        raise HTTPForbidden()
```
bowlofeggs: I think we could probably accomplish this by using the Pyramid ACL system. There is some code in the definition of ```update_edit``` near the top of this file that I would think should do this, but I suppose something is wrong there:

```
 53 update_edit = Service(
 54     name='update_edit', path='/updates/{id}/edit', validators=(validate_update_id,),
 55     description='Update submission service', acl=bodhi.server.security.packagers_allowed_acl,
 56     cors_origins=bodhi.server.security.cors_origins_rw)
```

I'd expect that ```acl=``` line to do the trick, but I'm not sure why it's not working. I think we should focus on fixing it that way so we're doing it "the Pyramid way".

bowlofeggs (changes requested):

bodhi/server/services/updates.py
ryanlerch: okies, will look into this. Although i copied this from similar code up a bit higher in the same file:

https://github.com/ryanlerch/bodhi/blob/46271695395fbd1028a523f5244b0928f5bd5e7e/bodhi/server/services/updates.py#L88

So i guess we need to change it there too.

bowlofeggs: @ryanlerch I *think* that one is actually using that to determine if the user should be presented with the edit buttons, for which the Pyramid ACL system wouldn't help I don't think. I'm not 100% sure, but if that is true I think that usage is probably OK.

ryanlerch:
Had a bit more of a dig around, and the ACL system goes a little over my head. For the "new update" page, we also aren't using the ACL system, but it is a little different, as we dont care there if a user can edit or not.

https://github.com/fedora-infra/bodhi/blob/develop/bodhi/server/views/generic.py#L171-L182

ryanlerch:
Updated the PR.

Ok, so i think i figured out why the `acl=` line in the `Service()` call was not working. AFAICT, we need to add a permission='something' in the @update_edit.get() decorator.

AFAICT, we don't actually specify specific permissions in bodhi/security.py. We just allow all permissions. so as long as we set something here (i went with 'edit'), the non-logged in user is not granted access to the form. We do this in other places where acls are set in the `Service()` call at the top of the file, rather than down in the decorator for the method.

Setting the permission parameter to something fixes the case for the non-logged in user, However, for the case of a logged-in user (with packager privs), but not commit access to the update, adding the  validate_acls validator checks for this. The results pass the tests, and i did a quick sanity check on a package that i own, and it seems to work, but would appriecate some other testing on this too.

bowlofeggs:
By my reading of [the Cornice docs](http://cornice.readthedocs.io/en/latest/api.html#cornice.service.Service), the acl attribute should be a function that returns a ```bool```, but it is set to [```bodhi.server.security.packagers_allowed_acl```](https://github.com/fedora-infra/bodhi/blob/2.6.2/bodhi/server/security.py#L37-L42), which returns a ```list``` instead. I suspect that may be the true issue.

The approach you have taken here does seem to be in line with the approach taken elsewhere in Bodhi, and is also [documented](http://docs.pylonsproject.org/projects/pyramid/en/1.8-branch/api/config.html#pyramid.config.Configurator.add_view) as being a reasonable approach, so I think it's fine.

As for testing, I think it would be good to write some tests in ```bodhi.tests.server.services.test_updates``` to demonstrate that the logged in user can see the view and that the non-logged-in user cannot see it.

bowlofeggs:
My hand testing also concludes that this approach works.

ryanlerch:
Updated the PR to include tests for the 3 cases of:

* Unlogged in User
* Logged in User without permissions on the update
* Logged in User with permissions on the update

bowlofeggs (approved):
Thanks!

PR #1523, opened 2017-05-16
enable fenced_code python-markdown extension

### LINKED ISSUE #1509: feature request - multiline blockquote using triple tick marks
it would be good if we can get multiline blockquotes in bodhi comments by using triple tick marks.

    ```
    like this
    ```

Here is an example where it line wraps the triple tick marks: https://bodhi.fedoraproject.org/updates/atomic-1.17.1-4.fc25#comment-605115

ryanlerch:
Enables the markdown.extensions.fenced_code extension on
markdown output so codeblocks with triple-backticks work

fixes #1509

### DIFF
--- a/bodhi/tests/server/functional/test_generic.py
+++ b/bodhi/tests/server/functional/test_generic.py
@@ -202,6 +202,17 @@ class TestGenericViews(bodhi.tests.server.functional.base.BaseWSGICase):
             "</div>"
         )
 
+    def test_markdown_with_fenced_code_block(self):
+        res = self.app.get('/markdown', {
+            'text': '```\nsudo dnf install bodhi\n```',
+        }, status=200)
+        self.assertEquals(
+            res.json_body['html'],
+            "<div class='markdown'>"
+            '<pre><code>sudo dnf install bodhi\n</code></pre>\n'
+            "</div>"
+        )
+
     def test_metrics(self):
         res = self.app.get('/metrics')
         self.assertIn('$.plot', res)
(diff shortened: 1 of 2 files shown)


pypingou:
👍 for me

Do we want tests for this?

ryanlerch:
@pypingou not sure. I did have a glance a the tests to see where the UI itself is being tested, but came up blank. Any pointers on where this is done in the tests?

bowlofeggs:
This does seem to work, but I do think it would be good to add a test for it. We could just have a test that passes in some text with the "fences" and asserts that the output HTML is reasonable.

@ryanlerch I'm happy to write a test for you if you want. Let me know!

ryanlerch:
thanks @bowlofeggs i figured it out. 

updated the commit with a test included.

bodhi/tests/server/functional/test_generic.py
```diff
@@ -202,6 +202,18 @@ def test_markdown_with_unknown_prefixed_bugzilla(self):
             "</div>"
         )
 
+    def test_markdown_with_fenced_code_block(self):
+        res = self.app.get('/markdown', {
+            'text': '```\nsudo dnf install bodhi\n```',
+        }, status=200)
+        print res.json_body['html']
```

bowlofeggs (approved):
Be sure to drop that print statement, but LGTM otherwise. No need to re-review, so feel free to merge once that print statement is gone. Nice!

bodhi/tests/server/functional/test_generic.py
ryanlerch: whoops! thanks for picking that up!

ryanlerch:
Thanks for the review @bowlofeggs 

Merging now!

PR #1524, opened 2017-05-16
Add the signing status of an update to the update page

### LINKED ISSUE #1469: Bodhi should show a lock icon next to the status if all of the builds in an update are signed
I believe that Bodhi's old UI had a set of lock/gear icons to indicate whether all the builds in the update were signed yet or not. The new UI does show whether individual builds are signed in the Builds tab, but it doesn't appear to show the overall signature status of the update anymore. I believe the Update model does have a signed attribute, so it should be pretty straightforward to include the icons for the overall update state, perhaps next to the "status" bit on the main tab.

ryanlerch:
This changes the icon on the builds tab to signify that
a build is signed to be a key rather than a lock.

Also adds the key icon to the status area in the right
bar on the Details tab of an update if the Update is
signed.

Removed the cog icon that stated that the "Build is being Signed"
as this was just shown if the build was not signed, not that
the build was in the process of being signed.

Fixes #1469

### DIFF
--- a/bodhi/server/templates/update.html
+++ b/bodhi/server/templates/update.html
@@ -647,6 +647,9 @@ $(document).ready(function(){
                 </div>
                 <div>
                   ${self.util.status2html(update.status) | n}
+                  % if update.signed:
+                    <span class="fa fa-key text-muted" aria-hidden="true" data-toggle="tooltip" data-placement="top" title="Update signed"></span>
+                  % endif
                 </div>
               </div>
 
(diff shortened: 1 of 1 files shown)


pypingou:
👍 based on the commit description, but screenshot? :)

bowlofeggs:
This looks good to me.

@pypingou, here are some screenshots:

delme3

delme2

bowlofeggs (approved):

PR #1526, opened 2017-05-16
PEP-257 bodhi.server.scripts.untag_branched.

bowlofeggs:


ralphbean (approved):
Nice and simple.

PR #1527, opened 2017-05-16
Don't double document the dev aliases.

bowlofeggs:
The bashrc aliases are already documented in the /etc/motd file, so
this commit removes a second place they were documented in the
developer docs so we don't have to maintain both copies. In its
place it instructs the developer to familiarize themselves with the
aliases printed upon login.


pypingou:
Sure 👍

ralphbean (approved):
Yes.

PR #1533, opened 2017-05-18
Make all the icon be of the class ``fa-fw`` so that they have a consistent width

pypingou:


bowlofeggs (approved):
The only difference I notice is that the title shifts a bit more to the right, but I might be missing something. In either case, it's fine with me ☺

pypingou:
What triggered that change for me was ``FEDORA-2017-dcc53ddb5f``, check the icon next to the title.

From there, when I changed one, I went ahead and change the others as well.

bowlofeggs:
@pypingou Oh I see what you mean. Thanks!

PR #1534, opened 2017-05-18
Add a btail command to watch live bodhi's log

pypingou:


devel/ansible/roles/dev/files/.bashrc
```diff
@@ -11,6 +11,7 @@ fi
 shopt -s expand_aliases
 alias bdocs="pushd /home/vagrant/bodhi/docs && make html && make man; popd"
 alias blog="sudo journalctl -u bodhi"
+alias btail="sudo journalctl -lfu bodhi"
```
bowlofeggs: You can already achieve this by using ```$ blog -f```, since ```blog``` is just an alias. You can add any other journalctl flags you like with it too. I'd advocate that we make that clearer in the motd rather than making more aliases.

devel/ansible/roles/dev/files/motd
```diff
@@ -3,6 +3,7 @@ Welcome to the Bodhi development environment! Here are some helpful commands:
 
 bdocs:      Build Bodhi's documentation.
 blog:       View Bodhi's log.
```
bowlofeggs: Perhaps instead of adding ```btail```, we could add something like this to this line "(Hint: this accepts journalctl's flags.)

bowlofeggs (approved):
I'm fine with the change if you really want it, but I suggest clarifying that the existing blog can be used for this instead.

devel/ansible/roles/dev/files/.bashrc
pypingou: Adjusted

ralphbean (approved):
The doc change works too.

PR #1535, opened 2017-05-18
Give the development environment a long login session by default.

bowlofeggs:


ralphbean (approved):
Should be fine.

PR #1536, opened 2017-05-18
[Factory 2.0] Add the ability to restrict Bodhi updates by using ACLs in Pagure

mprahl:
This PR does the following things:
* Adds the ability to restrict Bodhi updates by using ACLs in Pagure
* Adds the ability to get the critpath packages from PDC

This functionality is to enable the Arbitrary Branching project as described at https://fedoraproject.org/wiki/Infrastructure/Factory2/Focus/ArbitraryBranching

This change will not currently work in production because Pagure over dist-git is not yet deployed, but it worked in my local environment. This should not affect the production Bodhi instance because to enable this change, you'd need to set `acl_system` to `pagure` in the config file.

If you are interested, here is a short video demo showing the functionality of the first bullet point from above:
https://fedorapeople.org/groups/factory2/sprint-031/mprahl-bodhi-acls-using-pagure.mp4

Thank you for time and help reviewing this PR. If you have any questions or concerns, please let me know.

### DIFF
--- a/bodhi/server/models.py
+++ b/bodhi/server/models.py
@@ -43,7 +43,7 @@ from bodhi.server import bugs, buildsys, log, mail, notifications, Session
 from bodhi.server.config import config
 from bodhi.server.exceptions import BodhiException, LockedUpdateException
 from bodhi.server.util import (
-    avatar as get_avatar, build_evr, flash_log, get_critpath_pkgs,
+    avatar as get_avatar, build_evr, flash_log, get_critpath_components,
     get_nvr, get_rpm_header, header, tokenize, pagure_api_get)
 import bodhi.server.util
 
@@ -639,7 +639,7 @@ class Package(Base):
             # the former.
             for group_name in package_json['access_groups'][access_type]:
                 group_pagure_url = '{0}/api/0/group/{1}'.format(
-                    config['pagure_url'].rstrip('/'), group_name)
+                    pagure_url.rstrip('/'), group_name)
                 group_json = pagure_api_get(group_pagure_url)
                 committers = committers | set(group_json['members'])
 
@@ -1201,6 +1201,26 @@ class Update(Base):
                 break
             yield comment
 
+    @staticmethod
+    def contains_critpath_component(builds, release_name):
+        """
+        Determines if there is a critpath component in the builds passed in.
+        :param builds: a list of Build objects
+        :param release_name: a string representing the name of the release
+        such as "f25"
+        :return: a boolean determining if the update contains a critpath
+        component
+        """
+        for build in builds:
+            # This function call is cached, so there is no need to optimize
+            # it here.
+            critpath_components = get_critpath_components(
+                release_name.lower(), build.package.type.value)
+            if build.package.name in critpath_components:
+                return True
+
+        return False
+
     @classmethod
     def new(cls, request, data):
         """ Create a new update """
@@ -1208,18 +1228,9 @@ class Update(Base):
         user = User.get(request.user.name, request.db)
         data['user'] = user
         data['title'] = ' '.join([b.nvr for b in data['builds']])
-
         caveats = []
-
-        critical = False
-        critpath_pkgs = get_critpath_pkgs(data['release'].name.lower())
-        if critpath_pkgs:
-            for build in data['builds']:
-                if build.package.name in critpath_pkgs:
-                    critical = True
-                    break
-
-        data['critpath'] = critical
+        data['critpath'] = cls.contains_critpath_component(
+            data['builds'], data['release'].name)
 
         # Create the Bug entities, but don't talk to rhbz yet.  We do that
         # offline in the UpdatesHandler fedmsg consumer now.
@@ -1326,15 +1337,8 @@ class Update(Base):
                     # an override
                     db.delete(b)
 
-        critical = False
-        critpath_pkgs = get_critpath_pkgs(up.release.name.lower())
-        if critpath_pkgs:
-            for build in up.builds:
-                if build.package.name in critpath_pkgs:
-                    critical = True
-                    break
-
-        data['critpath'] = critical
+        data['critpath'] = cls.contains_critpath_component(
+            up.builds, up.release.name)
 
         del(data['builds'])
 
--- a/bodhi/server/models.py
+++ b/bodhi/server/models.py
@@ -44,7 +44,7 @@ from bodhi.server.config import config
 from bodhi.server.exceptions import BodhiException, LockedUpdateException
 from bodhi.server.util import (
     avatar as get_avatar, build_evr, flash_log, get_critpath_pkgs,
-    get_nvr, get_rpm_header, header, tokenize)
+    get_nvr, get_rpm_header, header, tokenize, pagure_api_get)
 import bodhi.server.util
 
 
@@ -611,6 +611,45 @@ class Package(Base):
                             committers.append(name)
 
         return (committers, watchers), (committergroups, watchergroups)
+
+    def get_pkg_committers_from_pagure(self):
+        """
+        Pull users who can commit and are watching a package in Pagure.
+
+        Return two two-tuples of lists:
+        * The first tuple is for usernames.  The second tuple is for groups.
+        * The first list of the tuple is for committers. The second is for
+          watchers.
+        """
+        pagure_url = config.get(
+            'pagure_url', 'https://src.fedoraproject.org/pagure/')
+        # Pagure uses plural names for its namespaces such as "rpms"
+        namespace = self.type.name + 's'
+        package_pagure_url = '{0}/api/0/{1}/{2}'.format(
+            pagure_url.rstrip('/'), namespace, self.name)
+        package_json = pagure_api_get(package_pagure_url)
+
+        committers = set()
+        for access_type in ['owner', 'admin', 'commit']:
+            committers = committers | set(
+                package_json['access_users'][access_type])
+        for access_type in ['admin', 'commit']:
+            # We resolve the group membership here since these are Pagure
+            # groups, and not FAS groups. Bodhi understands the latter but not
+            # the former.
+            for group_name in package_json['access_groups'][access_type]:
+                group_pagure_url = '{0}/api/0/group/{1}'.format(
+                    config['pagure_url'].rstrip('/'), group_name)
+                group_json = pagure_api_get(group_pagure_url)
+                committers = committers | set(group_json['members'])
+
+        # The first list contains users with commit access. The second list is
+        # supposed to contain groups with commit access. Since Pagure uses
+        # Pagure groups, which Bodhi doesn't understand, the code above
+        # resolves the group membership. That is why the second list is empty.
+        # These return values maintain compatibility with other functions such
+        # as get_pkg_pushers.
+        return list(committers), []
 
     def fetch_test_cases(self, db):
         """ Get a list of test cases from the wiki """
--- a/bodhi/server/util.py
+++ b/bodhi/server/util.py
@@ -167,21 +167,32 @@ class memoized(object):
 
 
 @memoized
-def get_critpath_pkgs(collection='master'):
-    """Return a list of critical path packages for a given collection"""
-    critpath_pkgs = []
+def get_critpath_components(collection='master', component_type='rpm'):
+    """ Return a list of critical path packages for a given collection.
+    :param collection: a string representing the collection/branch to search.
+    for
+    :param component_type: a string representing the component type to search
+    for. This only affects PDC queries.
+    :return: a list of critpath components
+    """
+    critpath_components = []
     critpath_type = config.get('critpath.type')
+    if critpath_type != 'pdc' and component_type != 'rpm':
+        log.warning('The critpath.type of "{0}" does not support searching for'
+                    ' non-RPM components'.format(component_type))
+
     if critpath_type == 'pkgdb':
         from pkgdb2client import PkgDB
         pkgdb = PkgDB(config.get('pkgdb_url'))
         results = pkgdb.get_critpath_packages(branches=collection)
         if collection in results['pkgs']:
-            critpath_pkgs = results['pkgs'][collection]
+            critpath_components = results['pkgs'][collection]
     elif critpath_type == 'pdc':
-        critpath_pkgs = get_critpath_pkgs_from_pdc(collection)
+        critpath_components = get_critpath_components_from_pdc(
+            collection, component_type)
     else:
-        critpath_pkgs = config.get('critpath_pkgs', '').split()
-    return critpath_pkgs
+        critpath_components = config.get('critpath_pkgs', '').split()
+    return critpath_components
 
 
 class Singleton(object):
@@ -659,7 +670,7 @@ def sort_severity(value):
     return value_map.get(value, 99)
 
 
-def get_critpath_pkgs_from_pdc(branch):
+def get_critpath_components_from_pdc(branch, component_type='rpm'):
     """
     Searches PDC for critical path packages based on the specified branch.
     :param branch: string representing the branch
@@ -674,7 +685,7 @@ def get_critpath_pkgs_from_pdc(branch):
         'critical_path': 'true',
         'name': branch,
         'page_size': 100,
-        'type': 'rpm'
+        'type': component_type
     })
     pdc_api_url_with_args = '{0}?{1}'.format(pdc_api_url, query_args)
 
--- a/bodhi/server/util.py
+++ b/bodhi/server/util.py
@@ -177,6 +177,8 @@ def get_critpath_pkgs(collection='master'):
         results = pkgdb.get_critpath_packages(branches=collection)
         if collection in results['pkgs']:
             critpath_pkgs = results['pkgs'][collection]
+    elif critpath_type == 'pdc':
+        critpath_pkgs = get_critpath_pkgs_from_pdc(collection)
     else:
         critpath_pkgs = config.get('critpath_pkgs', '').split()
     return critpath_pkgs
@@ -655,3 +657,96 @@ def sort_severity(value):
     }
 
     return value_map.get(value, 99)
+
+
+def get_critpath_pkgs_from_pdc(branch):
+    """
+    Searches PDC for critical path packages based on the specified branch.
+    :param branch: string representing the branch
+    :param component_type: string representing the component type to get
+    critpath packages for. This only affects queries to PDC.
+    :return: a list of critical path packages
+    """
+    pdc_api_url = '{}/rest_api/v1/component-branches/'.format(
+        config.get('pdc_url', 'https://pdc.fedoraproject.org/').rstrip('/'))
+    query_args = urllib.urlencode({
+        'active': 'true',
+        'critical_path': 'true',
+        'name': branch,
+        'page_size': 100,
+        'type': 'rpm'
+    })
+    pdc_api_url_with_args = '{0}?{1}'.format(pdc_api_url, query_args)
+
+    critpath_pkgs_set = set()
+    while True:
+        pdc_request_json = pdc_api_get(pdc_api_url_with_args)
+
+        for branch_rv in pdc_request_json['results']:
+            critpath_pkgs_set.add(branch_rv['global_component'])
+
+        if pdc_request_json['next']:
+            pdc_api_url_with_args = pdc_request_json['next']
+        else:
+            # There are no more results to iterate through
+            break
+    return list(critpath_pkgs_set)
+
+
+def api_get(api_url, service_name, error_key=None):
+    """
+    A wrapper to get API resources with some error handling.
+    :param api_url: a string of the URL to query
+    :param service_name: a string of the service name to use in error messages
+    :param error_key: a string of the key that holds the error message in the
+    JSON body. If this is set to None, the whole JSON will be used as the
+    error message.
+    :return: dictionary of the returned JSON or a RuntimeError
+    """
+    base_error_msg = (
+        'Bodhi failed to get a resource from {0} at the following URL '
+        '"{1}". The status code was "{2}".')
+    rv = requests.get(api_url, timeout=60)
+
+    if rv.status_code == 200:
+        return rv.json()
+    elif rv.status_code == 500:
+        # There will be no JSON with an error message here
+        error_msg = base_error_msg.format(
+            service_name, api_url, rv.status_code)
+        log.error(error_msg)
+        raise RuntimeError(error_msg)
+    else:
+        # If it's not a 500 error, we can assume that the API returned an error
+        # message in JSON that we can log
+        try:
+            rv_error = rv.json()
+            if error_key is not None:
+                rv_error = rv_error.get(error_key)
+        except ValueError:
+            rv_error = ''
+        error_msg = base_error_msg.format(
+            service_name, api_url, rv.status_code)
+        error_msg = '{0} The error was "{1}".'.format(error_msg, rv_error)
+        log.error(error_msg)
+        raise RuntimeError(error_msg)
+
+
+def pagure_api_get(pagure_api_url):
+    """
+    A wrapper to get API resources from Pagure.
+    :param pagure_api_url: a string of the URL to query
+    :return: JSON of the API resource(s) or a RuntimeError
+    """
+    return api_get(pagure_api_url, service_name='Pagure', error_key='error')
+
+
+def pdc_api_get(pdc_api_url):
+    """
+    A wrapper to get API resources from PDC.
+    :param pdc_api_url: a string of the URL to query
+    :return: JSON of the API resource(s) or a RuntimeError
+    """
+    # There is no error_key specified because the error key is not consistent
+    # based on the error message
+    return api_get(pdc_api_url, service_name='PDC')
--- a/bodhi/server/validators.py
+++ b/bodhi/server/validators.py
@@ -369,6 +369,19 @@ def validate_acls(request):
         if has_access:
             continue
 
+        # Make sure the user is in the mandatory packager groups. This is a
+        # safeguard in the event a user has commit access on the ACL system
+        # but isn't part of the mandatory groups.
+        mandatory_groups = request.registry.settings[
+            'mandatory_packager_groups'].split()
+        for mandatory_group in mandatory_groups:
+            if mandatory_group not in user_groups:
+                error = ('{0} is not a member of "{1}", which is a '
+                         'mandatory packager group').format(
+                    user.name, mandatory_group)
+                request.errors.add('body', 'builds', error)
+                return
+
         if acl_system == 'pkgdb':
             try:
                 people, groups = package.get_pkg_pushers(
@@ -381,6 +394,24 @@ def validate_acls(request):
                                    "Unable to access the Package "
                                    "Database to check ACLs. Please "
                                    "try again later.")
+                return
+        elif acl_system == 'pagure':
+            try:
+                committers, groups = package.get_pkg_committers_from_pagure()
+                people = committers
+            except RuntimeError as error:
+                # If it's a RuntimeError, then the error will be logged
+                # and we can return the error to the user as is
+                log.error(error)
+                request.errors.add('body', 'builds', unicode(error))
+                return
+            except Exception as error:
+                # This is an unexpected error, so let's log it and give back
+                # a generic error to the user
+                log.exception(error)
+                error_msg = ('Unable to access Pagure to check ACLs. '
+                             'Please try again later.')
+                request.errors.add('body', 'builds', error_msg)
                 return
         elif acl_system == 'dummy':
             people = (['ralph', 'bowlofeggs', 'guest'], ['guest'])
--- a/bodhi/tests/server/services/test_updates.py
+++ b/bodhi/tests/server/services/test_updates.py
@@ -163,7 +163,14 @@ class TestNewUpdate(bodhi.tests.server.functional.base.BaseWSGICase):
         app = TestApp(main({}, testing=u'bodhi', session=self.db, **self.app_settings))
         res = app.post_json('/updates/', self.get_update(u'bodhi-2.1-1.fc17'),
                             status=400)
-        assert 'bodhi does not have commit access to bodhi' in res, res
+        expected_error = {
+            "location": "body",
+            "name": "builds",
+            "description": ("bodhi is not a member of \"packager\", which is a"
+                            " mandatory packager group")
+        }
+        assert expected_error in res.json_body['errors'], \
+            res.json_body['errors']
         self.assertEquals(publish.call_args_list, [])
 
     @mock.patch(**mock_taskotron_results)
@@ -503,11 +510,14 @@ class TestUpdatesService(bodhi.tests.server.functional.base.BaseWSGICase):
         nvr = u'bodhi-2.1-1.fc17'
 
         user = User(name=u'lloyd')
+        user2 = User(name=u'ralph')
         self.db.add(user)
-        self.db.add(User(name=u'ralph'))  # Add a non proventester
+        self.db.add(user2)  # Add a packager but not proventester
         self.db.flush()
         group = self.db.query(Group).filter_by(name=u'provenpackager').one()
         user.groups.append(group)
+        group2 = self.db.query(Group).filter_by(name=u'packager').one()
+        user2.groups.append(group2)
 
         app = TestApp(main({}, testing=u'ralph', session=self.db, **self.app_settings))
         up_data = self.get_update(nvr)
@@ -535,12 +545,15 @@ class TestUpdatesService(bodhi.tests.server.functional.base.BaseWSGICase):
         "Ensure provenpackagers can change the request for any update"
         nvr = u'bodhi-2.1-1.fc17'
         user = User(name=u'bob')
+        user2 = User(name=u'ralph')
         self.db.add(user)
-        self.db.add(User(name=u'ralph'))  # Add a non proventester
+        self.db.add(user2)  # Add a packager but not proventester
         self.db.add(User(name=u'someuser'))  # An unrelated user with no privs
         self.db.flush()
         group = self.db.query(Group).filter_by(name=u'provenpackager').one()
         user.groups.append(group)
+        group2 = self.db.query(Group).filter_by(name=u'packager').one()
+        user2.groups.append(group2)
 
         app = TestApp(main({}, testing=u'ralph', session=self.db, **self.app_settings))
         up_data = self.get_update(nvr)
--- a/bodhi/tests/server/test_models.py
+++ b/bodhi/tests/server/test_models.py
@@ -25,7 +25,7 @@ from sqlalchemy.exc import IntegrityError
 import cornice
 import mock
 
-from bodhi.server import models as model, buildsys, mail
+from bodhi.server import models as model, buildsys, mail, util
 from bodhi.server.config import config
 from bodhi.server.exceptions import BodhiException
 from bodhi.server.models import (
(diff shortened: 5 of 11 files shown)


bodhi/server/models.py
```diff
@@ -549,6 +550,45 @@ def get_pkg_pushers(self, branch, settings):
...
+    def get_pkg_committers_from_pagure(self, settings):
+        """
+        Pull users who can commit and are watching a package in Pagure.
+
+        Return two two-tuples of lists:
+        * The first tuple is for usernames.  The second tuple is for groups.
+        * The first list of the tuple is for committers. The second is for
+          watchers.
+        """
+        namespace = 'rpms'
```
bowlofeggs: This method is being put in the ```Package``` superclass, which is more general than RPMs. There is a ```self.type``` column that can be used to determine whether the ```Package``` is an RPM or a Module (or in the future, a container). I recommend avoiding hardcoding this and instead mapping ```self.type``` to the appropriate string.

bodhi/server/models.py
```diff
@@ -549,6 +550,45 @@ def get_pkg_pushers(self, branch, settings):
...
+        package_pagure_url = '{0}/api/0/{1}/{2}'.format(
+            settings['pagure_url'].rstrip('/'), namespace, self.name)
```
bowlofeggs: This will cause a KeyError if ```pagure_url``` isn't defined in the settings. It might be good to explicitly raise a RuntimeError with a message that the setting must be defined in order for this method to work. Alternatively, we could ensure that this method never will get used unless that setting is present, which might be an even better approach.

bodhi/server/models.py
```diff
@@ -549,6 +550,45 @@ def get_pkg_pushers(self, branch, settings):
...
+        package_json = requests.get(package_pagure_url)
```
bowlofeggs: This call doesn't have a timeout, and requests will wait forever if a timeout isn't specified.

bodhi/server/models.py
```diff
@@ -549,6 +550,45 @@ def get_pkg_pushers(self, branch, settings):
...
+        if package_json.status_code != 200:
+            raise BodhiException(
+                'The specified package was not found in Pagure')
```
bowlofeggs: This error message could be misleading if the ```status_code``` isn't 404. I suggest handling 404 this way, and having another block to handle other codes with a generic error message like "Bodhi couldn't communicate with Pagure: received code {code}", or similar.

bodhi/server/models.py
```diff
@@ -549,6 +550,45 @@ def get_pkg_pushers(self, branch, settings):
...
+        package_json = package_json.json()
```
bowlofeggs: It could be nice to turn all of the above code into a separate helper method that just returns a ```dict```, and does all the error handling and requests stuff. We could keep it in ```bodhi.server.util``` since we might have other Pagure integration points in the future. It could even make sense to make a ```bodhi.server.pagure```. Alternatively, there might be a Pagure Python API bindings library that we could import and use (I'm not sure, but such a thing does exist for Bodhi).

bodhi/server/models.py
```diff
@@ -549,6 +550,45 @@ def get_pkg_pushers(self, branch, settings):
...
+        committers = set()
+        for access_type in ['owner', 'admin', 'commit']:
+            committers = committers | set(
+                package_json['access_users'][access_type])
+        for access_type in ['admin', 'commit']:
+            for group_name in package_json['access_groups'][access_type]:
+                group_pagure_url = '{0}/api/0/group/{1}'.format(
+                    settings['pagure_url'].rstrip('/'), group_name)
+                group_json = requests.get(group_pagure_url).json()
```
bowlofeggs: This would also benefit from there being a separate Pagure method that can do the error handling and timeouts for us.

bodhi/server/models.py
```diff
@@ -549,6 +550,45 @@ def get_pkg_pushers(self, branch, settings):
...
+                committers = committers | set(group_json['members'])
+
+        watchers = set()
+        package_watchers_pagure_url = '{0}/watchers'.format(package_pagure_url)
+        watchers_json = requests.get(package_watchers_pagure_url).json()
+        for watcher, watcher_levels in watchers_json['watchers'].items():
+            # TODO: Should this include issue watchers, commit watchers, or
+            # both?
+            if 'issues' in watcher_levels:
+                watchers.add(watcher)
```
bowlofeggs: Did you determine that we do need the watcher, or is that still an open question? If we don't need it, it might be nice to avoid doing this (and testing/maintaining it). If you need more help answering the question, let me know and I'll see if I can figure it out.

bodhi/server/models.py
```diff
@@ -549,6 +550,45 @@ def get_pkg_pushers(self, branch, settings):
...
+        for access_type in ['admin', 'commit']:
+            for group_name in package_json['access_groups'][access_type]:
+                group_pagure_url = '{0}/api/0/group/{1}'.format(
+                    settings['pagure_url'].rstrip('/'), group_name)
+                group_json = requests.get(group_pagure_url).json()
+                committers = committers | set(group_json['members'])
+
+        watchers = set()
+        package_watchers_pagure_url = '{0}/watchers'.format(package_pagure_url)
+        watchers_json = requests.get(package_watchers_pagure_url).json()
```
bowlofeggs: We probably also want error handling and timeouting on this call.

bodhi/server/models.py
```diff
@@ -549,6 +550,45 @@ def get_pkg_pushers(self, branch, settings):
...
+        for watcher, watcher_levels in watchers_json['watchers'].items():
+            # TODO: Should this include issue watchers, commit watchers, or
+            # both?
```
bowlofeggs: If Bodhi even does anything with watchers, I'm not sure where I fall on this question. One the one hand, I'd say that only committers want to watch Bodhi updates automatically because issue watchers are focusing on problems and they will get Bodhi notifications when Bodhi comments on BZs to say there are updates that address them. On the other hand, someone watching issues probably wants to know about updates to the package too.

Either way, I think the decision is easily defensible and also easy to change. I prefer not to put TODOs in the code because that's what bug trackers are for, so I recommend making a decision (either way is fine) and removing this comment. If you want to raise a discussion about whether we should change it, we can file a ticket and label it "discussion" and let that track it. Or we can just make a decision and wait for someone to file an issue because they disagree ☺

bodhi/server/models.py
bowlofeggs: As a user of Bodhi, I've noticed that I don't get e-mails about updates to packages I have ACLs on unless I participate in the update (by either making it, or commenting on it). That would suggest that my watcher status isn't really being used. As a developer, I think we should double check ☺

bodhi/server/models.py
```diff
@@ -549,6 +550,45 @@ def get_pkg_pushers(self, branch, settings):
...
+            if 'issues' in watcher_levels:
+                watchers.add(watcher)
+
+        return (list(committers), list(watchers)), ([], [])
```
bowlofeggs: These empty lists are supposed to represent groups? Bodhi does know about groups - should they be filled in?

bodhi/server/util.py
```diff
@@ -206,6 +206,34 @@ def get_critpath_pkgs(collection='master'):
...
+
+            for branch in pdc_request_json['results']:
+                critpath_pkgs_set.add(branch['global_component'])
+
+            if pdc_request_json['next']:
+                pdc_api_url_with_args = pdc_request_json['next']
+            else:
+                # There are no more results to iterate through
+                break
+        critpath_pkgs = list(critpath_pkgs_set)
```
bowlofeggs: This function was already very long and hard to read, but adding this much code to it makes it more unwieldy than it already is. Can we break the above code into an isolated helper function? It'll make it much easier to test too!

bodhi/server/util.py
```diff
@@ -206,6 +206,34 @@ def get_critpath_pkgs(collection='master'):
         results = pkgdb.get_critpath_packages(branches=collection)
         if collection in results['pkgs']:
             critpath_pkgs = results['pkgs'][collection]
+    elif critpath_type == 'pdc':
+        pdc_api_url = '{}/rest_api/v1/component-branches/'.format(
+            config.get('pdc_url').rstrip('/'))
+        pdc_api_url_with_args = (
+            '{}?critical_path=true&type=rpm&name={}&active=true&page_size=100'
+            .format(pdc_api_url, collection))
```
bowlofeggs: It might be nice to use [```urlencode()```](https://docs.python.org/2/library/urllib.html#urllib.urlencode) to do this.

bodhi/server/util.py
```diff
@@ -206,6 +206,34 @@ def get_critpath_pkgs(collection='master'):
...
+
+        critpath_pkgs_set = set()
+        while True:
+            pdc_request = requests.get(pdc_api_url_with_args)
```
bowlofeggs: Similar problems here - we need to handle error codes and timeouts.

bodhi/server/util.py
```diff
@@ -206,6 +206,34 @@ def get_critpath_pkgs(collection='master'):
...
+            pdc_request_json = pdc_request.json()
+
+            if pdc_request.status_code != 200:
```
bowlofeggs: It might be wise to check the status code before calling ```json()``` on it, because an HTTP 500 response might blow up in the call to ```json()``` above and we'll never get here to get the more informative error message.

bodhi/server/util.py
```diff
@@ -206,6 +206,34 @@ def get_critpath_pkgs(collection='master'):
...
+                error_msg = (
+                    'The request to get the critpath packages failed with the '
+                    'status code "{0}" and the following body "{1}"').format(
+                        str(pdc_request_json))
```
bowlofeggs: I think this string substitution will fail, since there are two items that need to be substituted but only one parameter given to ```format()```.

bodhi/server/validators.py
```diff
@@ -342,6 +355,18 @@ def validate_acls(request):
...
+            try:
+                people, groups = package.get_pkg_committers_from_pagure(
+                    settings)
+                committers, watchers = people
+                groups, notify_groups = groups
+            except Exception, e:
+                log.exception(e)
+                error_msg = ('Unable to access Pagure to check ACLs. '
+                             'Please try again later.')
+                request.errors.add('body', 'builds', error_msg)
```
bowlofeggs: I wonder if we want to give the user a clue about which type of error we got from Pagure? If it was a 404, should we tell the user, or would that definitely be a failure in the system somewhere?

development.ini.example
```diff
@@ -282,6 +282,16 @@ acl_system = dummy
...
 
+##
+## Pagure
+##
+pagure_url = https://src.fedoraproject.org/pagure/
+
+##
+## Product Definition Center (PDC)
+##
+pdc_url = https://pdc.fedoraproject.org/
```
bowlofeggs: It could be nice to comment both of these settings and to make sure that these values are the defaults if the settings aren't set.

development.ini.example
```diff
@@ -327,11 +337,14 @@ reboot_pkgs = kernel kernel-smp kernel-xen-hypervisor kernel-PAE kernel-xen0 ker
 ## https://fedoraproject.org/wiki/Critical_path_package
 ##
 
-# Enable this to query the Fedora Package Database for the list of Critical
-# Path Packages.  If disabled, it'll just use the hardcoded list below.
+# Enable this to query the Fedora Package Database or the Product Definition
+# Center for the list of Critical Path Packages.  If disabled, it'll just use
+# the hardcoded list below.
 #critpath.type = pkgdb
+#critpath.type = pdc
```
bowlofeggs: Let's only define this once and leave it at the default. It would be good to instead mention in the comment that "pdc" means to query product definition center.

production.ini
```diff
@@ -253,8 +253,8 @@ resultsdb_api_url = https://taskotron.fedoraproject.org/resultsdb_api/
 ##
```
bowlofeggs, line 253: Similar comments for this file as I gave on development.ini.example above.

bowlofeggs (changes requested):
There are two code test failures, and two code style test failures. You can replicate this by running ```btest``` if you use the Vagrant environment:

```
/home/vagrant/bodhi/bodhi/tests/server/test_utils.py:43:13: E126 continuation line over-indented for hanging indent
/home/vagrant/bodhi/bodhi/tests/server/test_utils.py:45:9: E121 continuation line under-indented for hanging indent
F.
ERROR: Ensure an error is thrown in Bodhi if there is an error in PDC
  File "/usr/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/usr/lib/python2.7/site-packages/mock/mock.py", line 1724, in _inner
    return f(*args, **kw)
  File "/home/vagrant/bodhi/bodhi/tests/server/test_utils.py", line 57, in test_get_critpath_pkgs_pdc_error
    util.get_critpath_pkgs('f25')
  File "/home/vagrant/bodhi/bodhi/server/util.py", line 185, in __call__
    value = self.func(*args)
  File "/home/vagrant/bodhi/bodhi/server/util.py", line 225, in get_critpath_pkgs
    str(pdc_request_json))
IndexError: tuple index out of range
bodhi: DEBUG: Using DevBuildsys

FAIL: test_no_privs (bodhi.tests.server.services.test_updates.TestNewUpdate)
  File "/home/vagrant/bodhi/bodhi/tests/server/services/test_updates.py", line 166, in test_no_privs
    assert 'bodhi does not have commit access to bodhi' in res, res
AssertionError: Response: 400 Bad Request
Content-Type: application/json; charset=UTF-8
Set-Cookie: session=wfYTeIqjaSNucQeLm5YssTZqGAMt5U3_HXPIXJm1Z2MeTpM1bq4ZQ8VFN8_ddJYB5Ky09B7uflAcIeOC9QGKo4ACSvjaHVlHQdZHdr4Ge6J9cQFVB19jc3JmdF9xAlgoAAAANjJiZjBjNjQ3NGNhMmJlYWJiYmZhZjY0MTcwY2NhZmFiNGI1ZTZhOXEDc4dxBC4; Path=/
{"status": "error", "errors": [{"location": "body", "name": "csrf_token", "description": "CSRF tokens do not match.  This happens if you have the page open for a long time. Please reload the page and try to submit your data again. Make s
ure to save your input somewhere before reloading. "}, {"location": "body", "name": "builds", "description": "bodhi is not a member of \"packager\", which is a mandatory packager group"}]}
bodhi: INFO: Using the FakeBugTracker
bodhi.server: DEBUG: Creating all models for Engine(sqlite://)
bodhi.server: DEBUG: Setting alias for bodhi-2.0-1.fc17 to FEDORA-2017-a3bbe1a8f2
bodhi: INFO: Using the FakeBugTracker
bodhi: INFO: Using the FakeBugTracker

AssertionError: 1 != 0
```

bodhi/server/models.py
mprahl: As far as I can tell, the watchers aren't used for anything, I only included them to mimic the function for PkgDB. Should we remove this ability?

bodhi/server/models.py
mprahl: Pagure only supports local groups and FAS groups which is why this is empty. Let me add a comment here to make this clear.

bowlofeggs:
On Thu, 2017-05-18 at 11:26 -0700, Matt Prahl wrote:
> As far as I can tell, the watchers aren't used for anything, I only
> included them to mimic the function for PkgDB. Should we remove this
> ability?

I'll do a code audit soonish to double check, so hold off on removing
this for now until I can do that.

mprahl:
@bowlofeggs I'm ready for another round of review. Thanks :smile:

ralphbean (approved):
LGTM.

bodhi/server/models.py
```diff
@@ -554,6 +555,54 @@ def get_pkg_pushers(self, branch, settings):
...
+        """
+        Pull users who can commit and are watching a package in Pagure.
+
+        Return two two-tuples of lists:
+        * The first tuple is for usernames.  The second tuple is for groups.
+        * The first list of the tuple is for committers. The second is for
+          watchers.
+        """
+        pagure_url = settings.get(
+            'pagure_url', 'https://src.fedoraproject.org/pagure/')
```
bowlofeggs: I just put up https://github.com/fedora-infra/bodhi/pull/1544 which should make it possible to just do ```config['pagure_url']```. Not sure what is the best way to coordinate the two PRs, so let's just say that whichever one gets merged last should be sure to fix this and add the new setting to the new ```bodhi.server.config.BodhiConfig._defaults``` dictionary.

bodhi/server/models.py
```diff
@@ -554,6 +555,54 @@ def get_pkg_pushers(self, branch, settings):
 
         return (committers, watchers), (committergroups, watchergroups)
 
+    def get_pkg_committers_from_pagure(self, settings):
```
bowlofeggs: You should be able to just use ```bodhi.server.config.config``` instead of passing the settings to this method. The config is already imported at the top of this module.

bodhi/server/models.py
```diff
@@ -554,6 +555,54 @@ def get_pkg_pushers(self, branch, settings):
...
+                committers = committers | set(group_json['members'])
+
+        watchers = set()
+        package_watchers_pagure_url = '{0}/watchers'.format(package_pagure_url)
+        watchers_json = pagure_api_get(package_watchers_pagure_url)
+        for watcher, watcher_levels in watchers_json['watchers'].items():
+            # Anyone who is watching commits in Pagure will most likely
+            # be interested in Bodhi updates on the package.
+            if 'commits' in watcher_levels:
+                watchers.add(watcher)
```
bowlofeggs: I can't find any parts of Bodhi that use watchers. @ralphbean, do you think we can drop watchers? If so, I say let's just drop it. We can always add it later if it turns out we are wrong.

bodhi/server/models.py
```diff
@@ -554,6 +555,54 @@ def get_pkg_pushers(self, branch, settings):
...
+
+        # The first tuple contains two lists of users. The second tuple
+        # contains two lists of groups. Since Pagure uses Pagure groups, which
+        # Bodhi doesn't understand, the code above resolves the group
+        # membership. That is why the second tuple contains two empty lists.
+        # These return values maintain compatibility with other functions such
+        # as get_pkg_pushers.
```
bowlofeggs: Nice docs!

bodhi/server/models.py
```diff
@@ -554,6 +555,54 @@ def get_pkg_pushers(self, branch, settings):
...
+        return (list(committers), list(watchers)), ([], [])
```
bowlofeggs: This entire method is missing test coverage.

bodhi/server/util.py
```diff
@@ -684,3 +686,101 @@ def sort_severity(value):
...
+
+    critpath_pkgs_set = set()
+    while True:
+        pdc_request_json = pdc_api_get(pdc_api_url_with_args)
+
+        for branch in pdc_request_json['results']:
+            critpath_pkgs_set.add(branch['global_component'])
+
+        if pdc_request_json['next']:
+            pdc_api_url_with_args = pdc_request_json['next']
```
bowlofeggs: This line is missing test coverage.

bodhi/server/util.py
```diff
@@ -684,3 +686,101 @@ def sort_severity(value):
     }
 
     return value_map.get(value, 99)
+
+
+def get_critpath_pkgs_from_pdc(branch):
+    """
+    Searches PDC for critical path packages based on the specified branch
```
bowlofeggs: s/branch/branch./

bodhi/server/util.py
```diff
@@ -684,3 +686,101 @@ def sort_severity(value):
...
+    :param branch: string representing the branch
+    :return: a list of critical path packages
+    """
+    pdc_api_url = '{}/rest_api/v1/component-branches/'.format(
+        config.get('pdc_url', 'https://pdc.fedoraproject.org/').rstrip('/'))
```
bowlofeggs: With the new config PR I made we can move this default to ```bodhi.server.config.BodhiConfig._defaults```, once #1544 is merged.

bodhi/server/util.py
```diff
@@ -684,3 +686,101 @@ def sort_severity(value):
...
+    query_args = urllib.urlencode({
+        'active': 'true',
+        'critical_path': 'true',
+        'name': branch,
+        'page_size': 100,
+        'type': 'rpm'
```
bowlofeggs: Wouldn't we want to know the critical path packages for all types that Bodhi supports?

bodhi/server/util.py
```diff
@@ -684,3 +686,101 @@ def sort_severity(value):
...
+            # There are no more results to iterate through
+            break
+    return list(critpath_pkgs_set)
+
+
+def pagure_api_get(pagure_api_url, timeout=60):
+    """
+    A wrapper to get API resources from Pagure
+    :param pagure_api_url: a string of the URL to query
+    :param timeout: the timeout on the GET request
```
bowlofeggs: s/request/request, in seconds/

I also wouldn't be opposed to just hardcoding the timeout. I doubt different callers will want to set different timeouts, and this way you wouldn't need to test the setting.

bodhi/server/util.py
```diff
@@ -684,3 +686,101 @@ def sort_severity(value):
...
+            pdc_api_url_with_args = pdc_request_json['next']
+        else:
+            # There are no more results to iterate through
+            break
+    return list(critpath_pkgs_set)
+
+
+def pagure_api_get(pagure_api_url, timeout=60):
+    """
+    A wrapper to get API resources from Pagure
```
bowlofeggs: s/Pagure/Pagure./

bodhi/server/util.py
```diff
@@ -684,3 +686,101 @@ def sort_severity(value):
...
+        # If it's not a 500 error, we can assume that Pagure returned an error
+        # message in JSON that we can log
+        try:
+            rv_error = rv.json().get('error')
+        except ValueError:
+            rv_error = ''
+        error_msg = base_error_msg.format(pagure_api_url, rv.status_code)
+        error_msg = '{0}. The error was "{1}"'.format(error_msg, rv_error)
+        log.exception(error_msg)
+        raise RuntimeError(error_msg)
```
bowlofeggs: This entire function is missing test coverage.

bodhi/server/util.py
```diff
@@ -684,3 +686,101 @@ def sort_severity(value):
...
+    rv = requests.get(pagure_api_url, timeout=timeout)
+
+    if rv.status_code == 200:
+        return rv.json()
+    elif rv.status_code == 500:
+        # There will be no JSON with an error message here
+        error_msg = base_error_msg.format(pagure_api_url, rv.status_code)
+        log.exception(error_msg)
+        raise RuntimeError(
+            base_error_msg.format(pagure_api_url, rv.status_code))
```
bowlofeggs: Seems like you could reuse ```error_msg``` from above.

bodhi/server/util.py
```diff
@@ -684,3 +686,101 @@ def sort_severity(value):
...
+        'Bodhi failed to get a resource from Pagure at the following URL '
+        '"{0}". The status code was "{1}"')
+    rv = requests.get(pagure_api_url, timeout=timeout)
+
+    if rv.status_code == 200:
+        return rv.json()
+    elif rv.status_code == 500:
+        # There will be no JSON with an error message here
+        error_msg = base_error_msg.format(pagure_api_url, rv.status_code)
+        log.exception(error_msg)
```
bowlofeggs: How about ```log.error```? ```log.exception``` will put a traceback in the log even though it's not actually a Bodhi crash.

bodhi/server/util.py
```diff
@@ -684,3 +686,101 @@ def sort_severity(value):
...
+    else:
+        # If it's not a 500 error, we can assume that Pagure returned an error
+        # message in JSON that we can log
+        try:
+            rv_error = rv.json().get('error')
+        except ValueError:
+            rv_error = ''
+        error_msg = base_error_msg.format(pagure_api_url, rv.status_code)
+        error_msg = '{0}. The error was "{1}"'.format(error_msg, rv_error)
+        log.exception(error_msg)
```
bowlofeggs: I suggest ```log.error``` instead of ```log.exception``` here too.

bodhi/server/util.py
```diff
@@ -684,3 +686,101 @@ def sort_severity(value):
...
+        # If it's not a 500 error, we can assume that PDC returned an error
+        # message in JSON that we can log
+        try:
+            rv_error = rv.json()
+        except ValueError:
+            rv_error = ''
+        error_msg = base_error_msg.format(pdc_api_url, rv.status_code)
+        error_msg = '{0}. The error was "{1}"'.format(error_msg, rv_error)
+        log.exception(error_msg)
+        raise RuntimeError(error_msg)
```
bowlofeggs: This ```else``` block is missing test coverage.

bodhi/server/util.py
```diff
@@ -684,3 +686,101 @@ def sort_severity(value):
...
+            rv_error = ''
+        error_msg = base_error_msg.format(pagure_api_url, rv.status_code)
+        error_msg = '{0}. The error was "{1}"'.format(error_msg, rv_error)
+        log.exception(error_msg)
+        raise RuntimeError(error_msg)
+
+
+def pdc_api_get(pdc_api_url, timeout=60):
+    """
+    A wrapper to get API resources from PDC
```
bowlofeggs: s/PDC/PDC./

bodhi/server/util.py
```diff
@@ -684,3 +686,101 @@ def sort_severity(value):
...
+    :param pdc_api_url: a string of the URL to query
+    :param timeout: the timeout on the GET request
```
bowlofeggs: Either specify that this is in seconds, or also feel free to drop it - I don't think we need to make the timeout configurable to the caller. If you do want it to be configurable, we could add it to the new ```bodhi.server.config.BodhiConfig._defaults``` from #1544, but I don't think hardcoding this is a problem.

bodhi/server/util.py
bowlofeggs: This function is extremely similar to the one for Pagure handling - the only difference I notice is whether we know the error from the json or not. Because of this, I recommend making a single unified helper function they can both use, and have each of these functions just use that. Then the special error handling can happen at the higher layer and we don't have so much copied code.

bodhi/server/validators.py
```diff
@@ -342,6 +355,25 @@ def validate_acls(request):
...
+                request.errors.add('body', 'builds', e.message)
+                return
+            except Exception as e:
+                # This is an unexpected error, so let's log it and give back
+                # a generic error to the user
+                log.exception(e)
+                error_msg = ('Unable to access Pagure to check ACLs. '
+                             'Please try again later.')
+                request.errors.add('body', 'builds', error_msg)
+                return
```
bowlofeggs: This entire ```elif``` block is missing test coverage.

bodhi/server/validators.py
```diff
@@ -342,6 +355,25 @@ def validate_acls(request):
...
+        elif acl_system == 'pagure':
+            try:
+                people, groups = package.get_pkg_committers_from_pagure(
+                    settings)
+                committers, watchers = people
+                groups, notify_groups = groups
+            except RuntimeError as e:
+                # If it's a RuntimeError, then the error will be logged
+                # and we can return the error to the user as is
+                request.errors.add('body', 'builds', e.message)
```
bowlofeggs: I believe that ```e.message``` is deprecated. I suggest ```unicode(message)``` instead.

bodhi/tests/server/services/test_updates.py
```diff
@@ -163,7 +163,14 @@ def test_no_privs(self, publish, *args):
...
                             status=400)
-        assert 'bodhi does not have commit access to bodhi' in res, res
+        expected_error = {
+            "location": "body",
+            "name": "builds",
+            "description": ("bodhi is not a member of \"packager\", which is a"
+                            " mandatory packager group")
+        }
+        assert expected_error in res.json_body['errors'], \
+            res.json_body['errors']
```
bowlofeggs, line 173: Optional: It might be nice to use ```self.assertTrue()``` here, but I can't think of a particularly strong advantage to one over the other.

bowlofeggs (changes requested):
This pull request drops the coverage by a percentage.

```
TOTAL                                       6275   1049    83%
nose.plugins.cover: ERROR: TOTAL Coverage did not reach minimum required: 84%
```

bodhi/server/models.py
ralphbean: Sure!  Yeah, I don't remember why it was ever there.  Maybe for "completeness"?

:+1: to drop it.

mprahl: I'll remove the watches from my new code but I'll keep it in the "pkgdb" and "dummy" code blocks. We can file a separate PR to remove those.

bodhi/server/util.py
mprahl: I think yes, but I'm trying to emulate PkgDB's current behavior. Does it support things other than RPMs?

bodhi/server/util.py
mprahl: Fixed

production.ini
mprahl: Fixed

bodhi/server/util.py
mprahl: I just checked with Ralph, and he pointed that it does support RPMs, containers, and modules. I removed the type limitation here. So now it will filter just by the collection/branch like PkgDB does.

mprahl:
@bowlofeggs thanks for the review! I believe I've addressed all your comments. Please let me know if you require further changes.

bodhi/server/models.py
```diff
@@ -1143,25 +1143,34 @@ def comments_since_karma_reset(self):
...
+        Determines if there is a critpath component in the builds passed in.
+        :param builds: a list of Build objects
+        :return: a boolean determining if the update contains a critpath
+        component
+        """
+        for build in builds:
+            # This function call is cached, so there is no need to optimize
+            # it here.
+            critpath_components = get_critpath_components(
+                release_name.lower(), build.package.type)
```
ralphbean: Is ``.type`` really a string here?  I would guess that it is a ``ContentType`` ``EnumSymbol``.  Do you need ``build.package.type.value`` instead?

mprahl: You're right. It seems to work both ways but I fixed it to your suggestion. Thank you!

mprahl:
@bowlofeggs I ended up adding a separate commit to allow critpath checking based on content types. This isn't supported in PkgDB's API though, so I added a warning in Bodhi but kept the existing functionality the same for PkgDB querying. The alternative is to return an empty list if the query is for PkgDB and the type isn't `rpm`. What are your thoughts?

bowlofeggs:
There is a failing test right now due to the lack of #1544:

```
ERROR: Ensure that the package commiters can be found using the Pagure
  File "/home/vagrant/bodhi/bodhi/tests/server/test_models.py", line 417, in test_get_pkg_committers_from_pagure
    rv = self.package.get_pkg_committers_from_pagure()
  File "/home/vagrant/bodhi/bodhi/server/models.py", line 584, in get_pkg_committers_from_pagure
    config['pagure_url'].rstrip('/'), group_name)
  File "/home/vagrant/bodhi/bodhi/server/config.py", line 47, in __getitem__
    return super(BodhiConfig, self).__getitem__(*args, **kw)
KeyError: 'pagure_url'
```

It's up to you whether you want to make this code work the old way so we can merge it before #1544, or whether you want to wait for me to merge #1544 so we can rebase this PR on it and update it to use the new central config system. I'd recommend the latter, but I'm OK with either you prefer.

mprahl:
@bowlofeggs that was a bug. I fixed that. Thank you.

bowlofeggs:
On Thu, 2017-05-25 at 13:15 -0700, Matt Prahl wrote:
> I think yes, but I'm trying to emulate PkgDB's current behavior. Does
> it support things other than RPMs?

I think it does know about containers.

mprahl:
@bowlofeggs please let me know if you have additional feedback.

bodhi/server/models.py
```diff
@@ -1104,25 +1143,34 @@ def comments_since_karma_reset(self):
                 break
             yield comment
 
+    @staticmethod
+    def contains_critpath_component(builds, release_name):
+        """
+        Determines if there is a critpath component in the builds passed in.
+        :param builds: a list of Build objects
```
bowlofeggs, line 1208: We need to document the ```release_name``` parameter here, including its type.

bodhi/server/util.py
```diff
@@ -167,19 +167,32 @@ def __get__(self, obj, objtype):
...
+    for
+    :param component_type: a string representing the component type to search
+    for. This only affects PDC queries.
+    :return: a list of critpath components
+    """
+    critpath_components = []
     critpath_type = config.get('critpath.type')
+    if critpath_type != 'pdc' and component_type != 'rpm':
+        log.warning('The critpath.type of "{0}" does not support searching for'
+                    ' non-RPM components'.format(critpath_type))
```
bowlofeggs: This line is missing test coverage.

bodhi/server/util.py
```diff
@@ -167,19 +167,32 @@ def __get__(self, obj, objtype):
...
+
     if critpath_type == 'pkgdb':
         from pkgdb2client import PkgDB
         pkgdb = PkgDB(config.get('pkgdb_url'))
         results = pkgdb.get_critpath_packages(branches=collection)
         if collection in results['pkgs']:
-            critpath_pkgs = results['pkgs'][collection]
+            critpath_components = results['pkgs'][collection]
```
bowlofeggs, line 189: This line is missing test coverage.

bodhi/server/util.py
```diff
@@ -655,3 +668,94 @@ def sort_severity(value):
     }
 
     return value_map.get(value, 99)
+
+
+def get_critpath_components_from_pdc(branch, component_type='rpm'):
+    """
+    Searches PDC for critical path packages based on the specified branch.
+    :param branch: string representing the branch
```
bowlofeggs, line 676: Let's document the ```component_type``` parameter and its type here.

bodhi/server/validators.py
```diff
@@ -342,6 +355,24 @@ def validate_acls(request):
...
+                request.errors.add('body', 'builds', unicode(error))
+                return
+            except Exception as error:
+                # This is an unexpected error, so let's log it and give back
+                # a generic error to the user
+                log.exception(error)
+                error_msg = ('Unable to access Pagure to check ACLs. '
+                             'Please try again later.')
+                request.errors.add('body', 'builds', error_msg)
+                return
```
bowlofeggs, line 415: Both of these exception handlers do not have test coverage.

bodhi/tests/server/test_models.py
```diff
@@ -361,6 +361,62 @@ def test_same_build_types(self):
...
+            }
+        }
+        json_two = {
+            "creator": {
+                "fullname": "Tom Brady",
+                "name": "tbrady"
+            },
+            "date_created": "1495035052",
+            "description": "Players of the New England Patriots",
+            "display_name": "Patriots",
```
bowlofeggs: Since this is a trademarked name, we probably should avoid using it. Same for "patriots" on line 373 above.

bowlofeggs (changes requested):
Just a few more changes. By the way, if you run ```btest``` in the Vagrant environment, it will print out a list of lines in Bodhi that are missing test coverage. That way you can see if you are missing coverage on your patch or not.

bodhi/server/util.py
mprahl: I added test coverage for it but I think this one was unfair. I just renamed the variable 😛

bodhi/server/validators.py
mprahl: Done

bodhi/server/util.py
mprahl: Done

bodhi/server/models.py
mprahl: Done

mprahl:
@bowlofeggs okay I should have 100% coverage now on the changes thanks to the help of [diff-cover](https://github.com/Bachmann1234/diff-cover).

Please review again.

bowlofeggs:
On Tue, 2017-05-30 at 16:29 -0700, Matt Prahl wrote:
> I added test coverage for it but I think this one was unfair. I just
> renamed the variable 😛

Actually, there was a Bodhi release just last fall where I had allowed
a commit that renamed a variable in without test coverage because it
looked fine, but it turned out to be a typo and the release blew up in
production. That's when I wrote the rule that all new code must have
100% coverage ☺

bowlofeggs (approved):
Very nice @mprahl!

```diff-cover``` is pleased with you too:

```
[vagrant@bodhi-dev bodhi]$ diff-cover coverage.xml --compare-branch=upstream/develop --fail-under=100                 
-------------
Diff Coverage
Diff: upstream/develop...HEAD, staged, and unstaged changes
-------------
bodhi/server/models.py (100%)
bodhi/server/util.py (100%)
bodhi/server/validators.py (100%)
-------------
Total:   86 lines
Missing: 0 lines
Coverage: 100%
-------------
```

ralphbean:
\ó/\ó/\ó/

PR #1537, opened 2017-05-18
PEP-8 docs/conf.py by reducing a line length.

bowlofeggs:
I have no idea how I missed that this was causing the flake8 test to fail, but it is ☹


jeremycline (approved):
:shipit:

PR #1538, opened 2017-05-19
Fix a failure in the pep8 style test.

ralphbean:
I found this was failing when I went to run the test suite today.


bowlofeggs:
I think I fixed this in https://github.com/fedora-infra/bodhi/pull/1537 though I did it in a slightly different way.

ralphbean:
LGTM!  :)

PR #1539, opened 2017-05-19
Add more tests for server/functional/generic.py

ryanlerch:


bodhi/server/buildsys.py
```diff
@@ -179,15 +193,15 @@ def listBuildRPMs(self, id, *args, **kw):
...
-                 'release': '2.fc7',
+                 'release': '2.fc17',
                  'size': 1993385,
                  'version': '1.0.2.2'},
                 ]
         if id == 16059:  # for updateinfo.xml tests
-            rpms[0]['nvr'] = rpms[1]['nvr'] = 'TurboGears-1.0.2.2-3.fc7'
-            rpms[0]['release'] = rpms[1]['release'] = '3.fc7'
+            rpms[0]['nvr'] = rpms[1]['nvr'] = 'TurboGears-1.0.2.2-3.fc17'
+            rpms[0]['release'] = rpms[1]['release'] = '3.fc17'
```
bowlofeggs, line 206: It seems like these lines are missing test coverage, which would imply that they weren't necessarily changed for the tests. In either case, they lacked test coverage before anyway and this is just test code so no big deal.

bodhi/server/buildsys.py
```diff
@@ -78,6 +78,8 @@ class DevBuildsys(Buildsystem):
     __added__ = []
     __tagged__ = {}
     __rpms__ = []
+    multicall = False
+    multicallresult = []
```
bowlofeggs: Since these are class variables, all the places that alter them will alter them across all instances. I suspect that is probably not what you want. I recommend instead setting these in an ```__init__()``` method like this instead:

```
self.multicall = False
self.multicallresult = []
```

That way each instance of DevBuildSys will have independent data on these variables.

Or was it intentional that these are class variables?

bodhi/tests/server/functional/test_generic.py
```diff
@@ -12,8 +12,13 @@
 # along with this program; if not, write to the Free Software
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
 
+import copy
+
 from pyramid.testing import DummyRequest
 
+from webtest import TestApp
```
bowlofeggs: Since webtest is a third party library it does make sense to be grouped with ```pyramid```, so let's remove the space between them.

bowlofeggs (approved):
Just a few thoughts - you can ignore them all if you want. No need for re-review if you decide to change anything based on the comments.

Huge +1 for bumping up test coverage by a percentage!

bodhi/server/buildsys.py
ryanlerch: Yeah, i had to change another set of these from fc7 to fc17 to get the fake buildsystem to give me what i needed. All the other fake turbogears packages had were fc17, so to make our fake buildsys a little more "real" i changed these ones to fc17 too.

bodhi/server/buildsys.py
ryanlerch: Not intentional at all. Just me not really knowing the difference there. Thanks for the lesson!

I have changed this now. I won't merge this one though, i'd feel comfortable if you just double check my changes on this one.

ryanlerch:
Updated the commit.

@bowlofeggs are you able to just check my changes in buildsys.py? want ot make sure i got it right.

cheers,
ryanlerch

bowlofeggs (approved):
@ryanlerch looks great! There's a merge conflict, but feel free to merge once that's resolved. Nice work!

Issue #1326, opened 2017-03-03
Modify the fedmsgs to handle multiple content types


bowlofeggs:
We need to redesign our fedmsgs so they can handle multiple content types.

ralphbean:
I think this is fixed by #1543.

PR #1543, opened 2017-05-24
Support non-rpm content (modules) in the API.

### LINKED ISSUE #1325: Modify the REST API to handle multiple content types
We need to modify the REST API so that it can handle multiple content types. It may be possible to do this in a backwards compatible way, but we can also name the next release Bodhi 3.0 if we need to make backwards-incompatible changes.

ralphbean:
This should fix #1325 and #1326.

As mentioned in #1325, this:

- Expands the GET api to return the `content_type` of a given update.
- Expands the search api to allow filtering on the `content_type` of updates.
- Enhances the submission api to *infer* the content_type from the given NVR,
  which should have minimal impact on users.

To this last point, we have to add an additional call to koji during module
submission to get the `extra` info back from the `koji.getBuild()` API.
Certain markers live in that block that can help us figure out what kind of
build this is.  Containers built in OSBS and Modules from the MBS will have
identifying keys there.

The code should have full coverage.  I introduced some unit tests of new
functions and methods I added and I also tested full request/reply paths
through the affected cornice services.

### DIFF
--- a/bodhi/server/models.py
+++ b/bodhi/server/models.py
@@ -265,6 +265,38 @@ class BodhiBase(object):
             columns.append(col.name)
         return columns
 
+    @classmethod
+    def find_polymorphic_child(cls, identity):
+        """ Find a child of a polymorphic base class.
+
+        For example, given the base Package class and the 'rpm' identity, this
+        class method should return the RpmPackage class.
+
+        This is accomplished by iterating over all classes in scope.
+        Limiting that to only those which are an extension of the given base
+        class.  Among those, return the one whose polymorphic_identity matches
+        the value given.  If none are found, then raise a NameError.
+
+        Arguments:
+        identity -- An instance of EnumSymbol used to identify the child.
+        """
+
+        if not isinstance(identity, EnumSymbol):
+            raise TypeError("%r is not an instance of EnumSymbol" % identity)
+
+        if 'polymorphic_on' not in getattr(cls, '__mapper_args__', {}):
+            raise KeyError("%r is not a polymorphic model." % cls)
+
+        classes = (c for c in globals().values() if isinstance(c, type))
+        children = (c for c in classes if issubclass(c, cls))
+        for child in children:
+            candidate = child.__mapper_args__.get('polymorphic_identity')
+            if candidate is identity:
+                return child
+
+        error = "Found no child of %r with identity %r"
+        raise NameError(error % (cls, identity))
+
     def update_relationship(self, name, model, data, db):
         """Add items to or remove items from a many-to-many relationship
 
@@ -311,6 +343,32 @@ class ContentType(DeclEnum):
     base = 'base', 'Base'
     rpm = 'rpm', 'RPM'
     module = 'module', 'Module'
+
+    @classmethod
+    def infer_content_class(cls, base, build):
+        """ Given a base class and a build from koji, identify and return the child
+        class associated with the appropriate ContentType.
+
+        For example, given the Package base class and a normal koji build, return
+        the RpmPackage model class. Or, given the Build base class and a container
+        build, return the ContainerBuild model class.
+
+        Arguments:
+        base -- A base model class.
+        build -- A dict of information from the build system (koji).
+        """
+
+        # Default value.  Overridden below if we find markers in the build info
+        identity = cls.rpm
+
+        extra = build.get('extra') or {}
+        if 'module' in extra.get('typeinfo', {}):
+            identity = cls.module
+        elif 'container_koji_task_id' in extra:
+            # TODO - implement containers as yet another content type someday.
+            raise NotImplementedError("Inferred type 'container' is unhandled.")
+
+        return base.find_polymorphic_child(identity)
 
 
 class UpdateStatus(DeclEnum):
@@ -1308,6 +1366,15 @@ class Update(Base):
         if not self.release.pending_signing_tag:
             return True
         return all([build.signed for build in self.builds])
+
+    @property
+    def content_type(self):
+        """ Return the ContentType associated with this Update.
+
+        If the update has no builds, this evaluates to `None`.
+        """
+        if self.builds:
+            return self.builds[0].type.value
 
     def obsolete_older_updates(self, db):
         """Obsolete any older pending/testing updates.
@@ -2311,6 +2378,8 @@ class Update(Base):
         result['submitter'] = result['user']['name']
         # Include the karma total in the results
         result['karma'] = self.karma
+        # Also, the Update content_type (derived from the builds content_types)
+        result['content_type'] = self.content_type
 
         # For https://github.com/fedora-infra/bodhi/issues/270, throw the JSON
         # of the test cases in our output as well but take extra care to
--- a/bodhi/server/services/updates.py
+++ b/bodhi/server/services/updates.py
@@ -21,8 +21,16 @@ from sqlalchemy.sql import or_
 
 from bodhi.server import log
 from bodhi.server.exceptions import BodhiException, LockedUpdateException
-from bodhi.server.models import (Update, Build, Bug, CVE, RpmPackage, UpdateRequest, ReleaseState,
-                                 RpmBuild)
+from bodhi.server.models import (
+    Update,
+    Bug,
+    ContentType,
+    CVE,
+    UpdateRequest,
+    ReleaseState,
+    Build,
+    Package,
+)
 import bodhi.server.schemas
 import bodhi.server.security
 import bodhi.server.services.errors
@@ -217,7 +225,7 @@ def query_updates(request):
     packages = data.get('packages')
     if packages is not None:
         query = query.join(Update.builds).join(Build.package)
-        query = query.filter(or_(*[RpmPackage.name == pkg for pkg in packages]))
+        query = query.filter(or_(*[Package.name == pkg for pkg in packages]))
 
     package = None
     if packages and len(packages):
@@ -226,7 +234,7 @@ def query_updates(request):
     builds = data.get('builds')
     if builds is not None:
         query = query.join(Update.builds)
-        query = query.filter(or_(*[RpmBuild.nvr == build for build in builds]))
+        query = query.filter(or_(*[Build.nvr == build for build in builds]))
 
     pushed = data.get('pushed')
     if pushed is not None:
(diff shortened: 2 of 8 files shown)


bodhi/server/util.py
```diff
@@ -684,3 +684,35 @@ def sort_severity(value):
...
+    if 'extra' not in build:
+        raise KeyError("No extra block found for build %r" % build)
+
+    if build['extra'] is None:
+        identity = ContentType.rpm
+    elif 'module' in build['extra'].get('typeinfo', {}):
+        identity = ContentType.module
+    elif 'container_koji_task_id' in build['extra']:
+        # TODO - implement containers as yet another content type someday.
+        raise NotImplementedError("Inferred type 'container' is unhandled.")
```
sochotnicky: So what happens when someone inputs container build NVR after this? Will they get some semi-verbose error?

Separately from this - while it would be outside of current implementations, there's no reason a single build couldn't have both rpms and containers (or other mixed content). Might be worth keeping in mind for future at least

ralphbean: > So what happens when someone inputs container build NVR after this? Will they get some semi-verbose error?

Yeah, this should be improved.  They'll get a 500 error.

Perhaps they should get a [501 error](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#5xx_Server_error)?

ralphbean: Adjusted in 4bf91b1.

bodhi/server/models.py
```diff
@@ -265,6 +265,38 @@ def grid_columns(cls):
...
+        Limiting that to only those which are an extension of the given base
+        class.  Among those, return the one whose polymorphic_identity matches
+        the value given.  If none are found, then raise a NameError.
+
+        Arguments:
+        identity -- An instance of EnumSymbol used to identify the child.
+        """
+
+        if not isinstance(identity, EnumSymbol):
+            raise ValueError("%r is not an instance of EnumSymbol" % identity)
```
bowlofeggs: This line is missing test coverage.

bodhi/server/models.py
```diff
@@ -265,6 +265,38 @@ def grid_columns(cls):
...
+    def find_polymorphic_child(cls, identity):
+        """ Find a child of a polymorphic base class.
+
+        For example, given the base Package class and the 'rpm' identity, this
+        class method should return the RpmPackage class.
+
+        This is accomplished by iterating over all classes in scope.
+        Limiting that to only those which are an extension of the given base
+        class.  Among those, return the one whose polymorphic_identity matches
+        the value given.  If none are found, then raise a NameError.
```
bowlofeggs, line 278: Fancy.

bodhi/server/models.py
```diff
@@ -265,6 +265,38 @@ def grid_columns(cls):
...
+
+        classes = (c for c in globals().values() if isinstance(c, type))
+        children = (c for c in classes if issubclass(c, cls))
+        for child in children:
+            candidate = child.__mapper_args__.get('polymorphic_identity')
+            if candidate is identity:
+                return child
+
+        error = "Found no child of %r with identity %r"
+        raise NameError(error % (cls, identity))
```
bowlofeggs, line 298: This method seems useful, nice!

bodhi/server/models.py
```diff
@@ -1309,6 +1341,19 @@ def signed(self):
             return True
         return all([build.signed for build in self.builds])
 
+    @property
+    def content_type(self):
+        """ Return the ContentType associated with this Update. """
+
+        if not self.builds:
+            raise ValueError("%r has no associated builds" % self)
```
bowlofeggs: I think ```ValueError``` usually means that a parameter to the function/method was an unacceptable value:

https://docs.python.org/2/library/exceptions.html#exceptions.ValueError

I think an ```IndexError``` could strangely make some kind of sense here, because we need there to be at least a self.builds[0] and there isn't one (and trying to access that would raise an ```IndexError```).

If we wanted to be non-strange, we could just go for the catch-all [```RuntimeError```](https://docs.python.org/2/library/exceptions.html#exceptions.RuntimeError).

bodhi/server/models.py
```diff
@@ -1309,6 +1341,19 @@ def signed(self):
...
+
+        content_types = set([build.type for build in self.builds])
+        if len(content_types) > 1:
+            raise ValueError("%r is a multi-type update (%r)." % self)
```
bowlofeggs: We do have validation in place to ensure that this state *should* never happen. I'm fine with the above if you want a belt-and-suspender approach, but I also think we can make this method simpler if you want to rely on the existing validation. I also think a ```RuntimeError``` probably makes more sense here.

FWIW, I bet you could replace this entire method with something much simpler that would even raise a TypeError for you (though I think it'd be best to catch and re-raise it with a better error message):

```
@property
def content_type(self):
    try:
        return self.builds[0].type
    except IndexError:
        raise IndexError("%r has no associated builds" % self)
```

What do you think? If you want the belt-and-suspenders I think that what you have is fine too, though I recommend changing the error type.

bowlofeggs: Oooh, another option is to return ```None```, since an Update without builds can rightfully be thought to have no type. I think raising an ```Exception``` is fine too though.

bodhi/server/models.py
bowlofeggs: This line is missing test coverage.

bodhi/server/models.py
bowlofeggs: This line is missing test coverage.

ralphbean:
@sochotnicky, looking better?

@bowlofeggs, any feedback?

bodhi/server/services/updates.py
```diff
@@ -349,17 +362,29 @@ def new_update(request):
...
+            try:
+                build_class = bodhi.server.util.infer_content_class(
+                    base=Build, build=request.buildinfo[nvr]['info'])
+            except Exception as e:
+                error = 'Unable to infer content_type.  %r' % e.message
+                log.exception(error)
+                request.errors.add('body', 'builds', error)
+                if isinstance(e, NotImplementedError):
+                    request.errors.status = HTTPNotImplemented.code
+                return
```
bowlofeggs: This exception handling block is missing coverage.

bodhi/server/util.py
```diff
@@ -684,3 +684,35 @@ def sort_severity(value):
...
+    Arguments:
+    base -- A base model class.
+    build -- A dict of information from the build system (koji).
+    """
+
+    # Import this here to avoid circular import
+    from bodhi.server.models import ContentType
+
+    if 'extra' not in build:
+        raise KeyError("No extra block found for build %r" % build)
```
bowlofeggs: This line is missing test coverage.

bodhi/server/util.py
```diff
@@ -684,3 +684,35 @@ def sort_severity(value):
...
+
+    if build['extra'] is None:
+        identity = ContentType.rpm
+    elif 'module' in build['extra'].get('typeinfo', {}):
+        identity = ContentType.module
+    elif 'container_koji_task_id' in build['extra']:
+        # TODO - implement containers as yet another content type someday.
+        raise NotImplementedError("Inferred type 'container' is unhandled.")
+    else:
+        identity = ContentType.rpm
```
bowlofeggs: This line is missing test coverage.

bodhi/server/util.py
```diff
@@ -684,3 +684,35 @@ def sort_severity(value):
...
+    the RpmPackage model class. Or, given the Build base class and a container
+    build, return the ContainerBuild model class.
+
+    Arguments:
+    base -- A base model class.
+    build -- A dict of information from the build system (koji).
+    """
+
+    # Import this here to avoid circular import
+    from bodhi.server.models import ContentType
```
bowlofeggs: Circular imports are usually a sign that there is an organization problem. I.e., if this module uses the models class, and the models class uses this module, then these things should probably be the same. In other words, we should move this function to the models module.

bodhi/server/validators.py
```diff
@@ -261,11 +288,26 @@ def validate_acls(request):
...
+                request.errors.add('body', 'builds', error)
+                if isinstance(e, NotImplementedError):
+                    request.errors.status = HTTPNotImplemented.code
+                return
+
+            # Get the Package object
             package_name = buildinfo['nvr'][0]
-            package = db.query(RpmPackage).filter_by(name=package_name).first()
+            package = db.query(package_class)\
+                .filter_by(name=package_name).first()
```
bowlofeggs: FWIW, it is now possible to do ```package_class.query.filter_by()```, thanks to some fancy pants work by @jeremycline ☺

bowlofeggs (changes requested):

bodhi/server/util.py
ralphbean: :vulcan_salute: can do!

bodhi/server/models.py
ralphbean: Let's go with your simple `None`-return approach.  :)

bodhi/server/services/updates.py
ralphbean: Removed this, since it is unreachable.

bodhi/server/util.py
ralphbean: It's my understanding that *old* versions of koji may not return an `extra` field... but the current versions do always return it.

I'll just remove the conditional, since testing it would require gymnastic mocking, and without it we'll still raise a KeyError anyways.

ralphbean:
OK, all your comments should be addressed now @bowlofeggs.

bodhi/server/services/updates.py
```diff
@@ -349,17 +362,27 @@ def new_update(request):
...
+
+            # Figure out what kind of build this should be
+            try:
+                build_class = ContentType.infer_content_class(
+                    base=Build, build=request.buildinfo[nvr]['info'])
+            except Exception as e:
+                error = 'Unable to infer content_type.  %r' % e.message
+                log.exception(error)
+                request.errors.add('body', 'builds', error)
+                return
```
bowlofeggs: This exception handler is still missing test coverage.

bowlofeggs (changes requested):

bodhi/server/services/updates.py
ralphbean: I just removed the handler rather than try to mock things out.

The issue is that the error *should* always be caught in the validator before hand when inferring the Package type.

ralphbean:
Any new feedback after ad3d4f3d83d591e1443954879d7c91d711917063 ?

bowlofeggs (approved):
LGTM! Please do squash down to one commit before merging. Also, it looks like you need to rebase, but no need for re-review after that as long as the tests still pass. Nice work!

ralphbean:
Thanks all!

ralphbean:
For posterity:  https://fedorapeople.org/groups/factory2/sprint-031/threebean-modular-bodhi-updates.ogv

PR #1544, opened 2017-05-24
Give Bodhi central configuration validation.

### LINKED ISSUE #1489: config.py should do more for us

bowlofeggs:
This pull request has two commits. The first is the primary goal of the PR, and gives Bodhi a central config validation system. This is a significant refactor, and a technical debt payment. I recommend reading its commit message for further details.

The first commit happened to raise test coverage to 85%, so the second commit requires test coverage to remain at 85%.

fixes #1489

### DIFF
--- a/bodhi/server/mail.py
+++ b/bodhi/server/mail.py
@@ -182,7 +182,7 @@ To reply to this comment, please visit the URL at the bottom of this mail
 """,
         'fields': lambda agent, x: {
             'package': x.title,
-            'comment': x.comments[-1].text,
+            'comment': x.comments[-1],
             'updatestr': unicode(x)
         }
     },
--- a/bodhi/server/mail.py
+++ b/bodhi/server/mail.py
@@ -182,7 +182,7 @@ To reply to this comment, please visit the URL at the bottom of this mail
 """,
         'fields': lambda agent, x: {
             'package': x.title,
-            'comment': x.comments[-1],
+            'comment': x.comments[-1].text,
             'updatestr': unicode(x)
         }
     },
@@ -463,7 +463,7 @@ def send_mail(from_addr, to_addr, subject, body_text, headers=None):
     if headers:
         for key, value in headers.items():
             msg.append('%s: %s' % (key, to_bytes(value)))
-    msg.append('X-Bodhi: %s' % config.get('default_email_domain'))
+    msg.append('X-Bodhi: %s' % to_bytes(config.get('default_email_domain')))
     msg += ['Subject: %s' % subject, '', body_text]
     body = to_bytes('\r\n'.join(msg))
 
@@ -500,7 +500,7 @@ def send(to, msg_type, update, sender=None, agent=None):
             headers["References"] = initial_message_id
             headers["In-Reply-To"] = initial_message_id
 
-    subject_template = '[Fedora Update] %s[%s] %s'
+    subject_template = u'[Fedora Update] %s[%s] %s'
     for person in iterate(to):
         subject = subject_template % (critpath, msg_type, update.title)
         fields = MESSAGES[msg_type]['fields'](agent, update)
--- a/bodhi/server/models.py
+++ b/bodhi/server/models.py
@@ -27,7 +27,6 @@ import time
 import uuid
 
 from pkgdb2client import PkgDB
-from pyramid.settings import asbool
 from simplemediawiki import MediaWiki
 from six.moves.urllib.parse import quote
 from sqlalchemy import (and_, Boolean, Column, DateTime, ForeignKey, Integer, or_, Table, Unicode,
@@ -621,8 +620,7 @@ class Package(Base):
         * The first list of the tuple is for committers. The second is for
           watchers.
         """
-        pagure_url = config.get(
-            'pagure_url', 'https://src.fedoraproject.org/pagure/')
+        pagure_url = config.get('pagure_url')
         # Pagure uses plural names for its namespaces such as "rpms"
         namespace = self.type.name + 's'
         package_pagure_url = '{0}/api/0/{1}/{2}'.format(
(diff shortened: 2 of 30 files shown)


pypingou:
Looks like it is conflicting now :(

bowlofeggs:
@pypingou I have resolved the conflict, and I have also altered the second commit to require 86% test coverage since @ryanlerch recently also merged another commit that increased test coverage to 85%. That means we've gotten it raised by 2% in a month!

pypingou:
Pretty big one, but the bodhi_config file looks good

bowlofeggs:
I have now rebased this PR on top of #1536, incorporating its new settings and adjusting it to work the new way that's introduced in this PR.

bodhi/server/mail.py
```diff
@@ -182,7 +182,7 @@
 """,
         'fields': lambda agent, x: {
             'package': x.title,
-            'comment': x.comments[-1],
+            'comment': x.comments[-1].text,
```
ralphbean: This is a surprising change to find in this PR.

What's this about?

bodhi/server/models.py
```diff
@@ -2208,8 +2206,7 @@ def check_karma_thresholds(self, db, agent):
         if self.autokarma and self._composite_karma[1] != 0:
             log.info("Disabling Auto Push since the update has received negative karma")
             self.autokarma = False
-            text = unicode(config.get(
-                'disable_automatic_push_to_stable', DEFAULT_DISABLE_AUTOPUSH_MESSAGE))
+            text = config.get('disable_automatic_push_to_stable', DEFAULT_DISABLE_AUTOPUSH_MESSAGE)
```
ralphbean: Is this default still needed here?

setup.cfg
```diff
@@ -2,7 +2,7 @@
 cover-erase=TRUE
 cover-html=TRUE
 cover-inclusive=TRUE
-cover-min-percentage=84
+cover-min-percentage=87
```
ralphbean, line 5: Woohoo!!

ralphbean:
In general, this change looks great.

I left two questions where I was genuinely uncertain about what should be done.

Basic :+1: from me to merge at your discretion.

bodhi/server/mail.py
bowlofeggs: @ralphbean I had been having some encoding related issues when I was developing this patch early on, and I think I had done this to test that. I reverted this change and all tests pass so I think it's not needed. I'll back it out.

bodhi/server/models.py
bowlofeggs: @ralphbean It is not, good catch! I'll remove it too.

bowlofeggs:
@ralphbean In response to your comments, I am applying this patch to my commit:

```
diff --git a/bodhi/server/mail.py b/bodhi/server/mail.py
index 6650014..8846400 100644
--- a/bodhi/server/mail.py
+++ b/bodhi/server/mail.py
@@ -182,7 +182,7 @@ To reply to this comment, please visit the URL at the bottom of this mail
 """,
         'fields': lambda agent, x: {
             'package': x.title,
-            'comment': x.comments[-1].text,
+            'comment': x.comments[-1],
             'updatestr': unicode(x)
         }
     },
index 00a8039..6507686 100644
--- a/bodhi/server/models.py
+++ b/bodhi/server/models.py
@@ -47,11 +47,6 @@ from bodhi.server.util import (
 import bodhi.server.util
 
 
-DEFAULT_DISABLE_AUTOPUSH_MESSAGE = (
-    u'Bodhi is disabling automatic push to stable due to negative karma. '
-    u'The maintainer may push manually if they determine that the issue is not severe.')
-
-
 # http://techspot.zzzeek.org/2011/01/14/the-enum-recipe
 
 class EnumSymbol(object):
@@ -2206,7 +2201,7 @@ class Update(Base):
         if self.autokarma and self._composite_karma[1] != 0:
             log.info("Disabling Auto Push since the update has received negative karma")
             self.autokarma = False
-            text = config.get('disable_automatic_push_to_stable', DEFAULT_DISABLE_AUTOPUSH_MESSAGE)
+            text = config.get('disable_automatic_push_to_stable')
             self.comment(db, text, author=u'bodhi')
         elif self.stable_karma and self.karma >= self.stable_karma:
             if self.autokarma:
index 95486c6..6994252 100644
@@ -28,7 +28,7 @@ import mock
 from bodhi.server import main
 from bodhi.server.models import (
-    BuildrootOverride, DEFAULT_DISABLE_AUTOPUSH_MESSAGE, Group, RpmPackage, Release,
+    BuildrootOverride, Group, RpmPackage, Release,
     ReleaseState, RpmBuild, Update, UpdateRequest, UpdateStatus, UpdateType, User)
 import bodhi.tests.server.functional.base
 
@@ -3246,8 +3246,7 @@ class TestUpdatesService(bodhi.tests.server.functional.base.BaseWSGICase):
         up.comment(self.db, u'Failed to work', author=u'ralph', karma=-1)
 
-        expected_comment = unicode(config.get(
-            'disable_automatic_push_to_stable', DEFAULT_DISABLE_AUTOPUSH_MESSAGE))
+        expected_comment = config.get('disable_automatic_push_to_stable')
         self.assertEquals(up.comments[2].text, expected_comment)
 
         up.comment(self.db, u'LGTM Now', author=u'ralph', karma=1)
```

I will rebase and merge this PR now - thanks for the review!

ralphbean:
:unicorn:

PR #1546, opened 2017-05-25
remove old get_age methods

ryanlerch:
We now use the "age" method in util.py to get relative
dates and times (i.e. 3h hours ago), which uses the arrow
library.

There are two methods in util.py, get_age() and get_age_in_days()
that we dont use anymore, as we use the age() method instead.

pluralize() in util.py is also removed, as the only place we
used this was in the get_ages, and we do other pluralizations
in JS now anyway.

The only other places that get_age() and get_age_in_days() were used
were a handful of methods in models.py, but these were not used anymore
either.


ralphbean (approved):
:rainbow: :rainbow: :rainbow:  for deleting code :x: 

Tests pass locally for me.

bowlofeggs:
```deleting_code++``` thanks!

PR #1547, opened 2017-05-25
PEP-257 bodhi.server.scripts.

bowlofeggs:


ralphbean (approved):
:rice_ball:

pypingou:
Looks like it is conflicting now :(

PR #1549, opened 2017-05-26
Add tests for the admin view

ryanlerch:
This adds a couple of tests for the admin view, which
is a json call.

### DIFF
--- a/bodhi/tests/server/functional/test_admin.py
+++ b/bodhi/tests/server/functional/test_admin.py
@@ -1,0 +1,48 @@
+# Copyright 2017 Red Hat, Inc.
+#
+# This file is part of Bodhi.
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+
+import copy
+
+from webtest import TestApp
+
+from bodhi.server import main
+import bodhi.tests.server.functional.base
+
+
+class TestAdminView(bodhi.tests.server.functional.base.BaseWSGICase):
+
+    def test_admin(self):
+        """Test that authticated user can see the Admin panel"""
+        res = self.app.get('/admin/')
+        body = res.json_body
+        self.assertEquals(body['principals'][0], 'system.Everyone')
+        self.assertEquals(body['principals'][1], 'system.Authenticated')
+        self.assertEquals(body['principals'][2], 'guest')
+        self.assertEquals(body['user'], 'guest')
+
+    def test_admin_unauthed(self):
+        """Test that an unauthed user cannot see the admin endpoint"""
+        anonymous_settings = copy.copy(self.app_settings)
+        anonymous_settings.update({
+            'authtkt.secret': 'whatever',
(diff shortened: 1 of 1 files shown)


bodhi/tests/server/functional/test_admin.py
```diff
@@ -0,0 +1,44 @@
+# This program is free software; you can redistribute it and/or
```
bowlofeggs, line 5: Above this line, let's add:

```
# Copyright 2017 Red Hat, Inc.
#
# This file is part of Bodhi.
#
```

I've been meaning to add these declarations to the existing files, but I'll need to do some research to find out what years to claim on each file so it's not super easy.

bodhi/tests/server/functional/test_admin.py
```diff
@@ -0,0 +1,44 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+
+import copy
+from webtest import TestApp
```
bowlofeggs: Webtest is a third party library, so let's add a space between copy and webtest (copy is stdlib).

bodhi/tests/server/functional/test_admin.py
```diff
@@ -0,0 +1,44 @@
...
+
+from bodhi.server import main
+from bodhi.server.models import (User, Group)
```
bowlofeggs: You don't need these parenthesis, though they aren't harmful either.

bodhi/tests/server/functional/test_admin.py
```diff
@@ -0,0 +1,44 @@
...
+class TestAdminView(bodhi.tests.server.functional.base.BaseWSGICase):
+
+    def test_admin(self):
+        """Test that authticated user can see the Admin panel"""
+        res = self.app.get('/admin/')
+        body = res.json_body
+        self.assertEquals(body['principals'][0], 'system.Everyone')
+        self.assertEquals(body['principals'][1], 'system.Authenticated')
+        self.assertEquals(body['principals'][2], 'guest')
+        self.assertEquals(body['user'], 'guest')
```
bowlofeggs, line 36: This is a weird view. I've not messed with it before.

bodhi/tests/server/functional/test_admin.py
bowlofeggs: Actually, it turns out that you also don't need this import and it is causing one of the tests to fail:

```
/home/vagrant/bodhi/bodhi/tests/server/functional/test_admin.py:19:1: F401 'bodhi.server.models.Group' imported but unused
/home/vagrant/bodhi/bodhi/tests/server/functional/test_admin.py:19:1: F401 'bodhi.server.models.User' imported but unused
F.
AssertionError: 1 != 0
```

bowlofeggs (approved):
Be sure to fix the failing test and the copyright before merging. No need for re-review, so merge when ready. Thanks for writing tests!

bodhi/tests/server/functional/test_admin.py
ryanlerch: yeah, not really sure of the merit of it, but i added the tests anyway, just in case we make it more useful in the future...

bodhi/tests/server/functional/test_admin.py
ryanlerch: added! thanks!

bodhi/tests/server/functional/test_admin.py
ryanlerch: space added. Thanks!

bodhi/tests/server/functional/test_admin.py
ryanlerch: Whoops, yeah i had it in there for an earlier iteration of my tests when i was hacking on it, where i was using it. Forgot to take it out. Not sure why i didnt notice the test failing when I ran my final tests before pushing though :/ 

thanks for picking this up! cheers!

PR #1550, opened 2017-05-26
switch requirements from depcheck to rpmdeplint, rework resultsdb queries

kparal:
Depcheck is replaced by rpmdeplint, update the code. Also fix queries to
resultsdb:
* use API 2.0
* ask for both bodhi_update and koji_build results
* do very specific queries (only request required testcases, use
timestamps)
* use results/latest endpoint by default
* give up if there are too many results for a given item (doesn't affect
results/latest endpoint, only results endpoint)
* use larger pages when retrieving testcases

There are known issues with the current results/latest endpoint
approach, in particular it doesn't display results for all architectures
(for rpmdeplint). Since rpmdeplint is currently only executed for
x86_64, it isn't a huge problem atm, but will need to be resolved in the
future.

----

Hello, this is a preliminary patch for testing and discussion. Rpmdeplint is only run in dev at the moment, we'll be pushing it into production after Beta freeze (so this patch doesn't make sense before that). Furthermore, this patch is completely untested. I used a development VM using vagrant, but I have no idea how to actually test this backend stuff. (I probably broke the unit tests as well). Can you please test it?

Is the code in `check_requirements()` even used at all? And when it is used? How can I trigger it when using vagrant dev env?

I assume that `check_requirements()` should block karma autopush in case any of those testcases specified in `site_requirements` are not passed. Is that correct? Is it being used at the moment? (Because it seems to me that this must've been broken for some time).

Thanks.

### DIFF
--- a/bodhi/server/buildsys.py
+++ b/bodhi/server/buildsys.py
@@ -15,6 +15,7 @@
 from threading import Lock
 import logging
 import time
+from functools import wraps
 
 import koji
 
@@ -25,7 +26,23 @@ _buildsystem = None
 _koji_hub = None
 
 
-class Buildsystem:
+def multicall_enabled(func):
+    '''Decorator to enable multicall handling for DevBuildsys methods'''
+    @wraps(func)
+    def wrapper(self, *args, **kwargs):
+        if not self.multicall:
+            return func(self, *args, **kwargs)
+
+        # disable multicall during execution, so that inner func calls to other
+        # methods don't append their results as well
+        self._multicall = False
+        result = func(self, *args, **kwargs)
+        self.multicall_result.append([result])
+        self._multicall = True
+    return wrapper
+
+
+class Buildsystem(object):
     """
     The parent for our buildsystem.  Not only does this help us keep track of
     the functionality that we expect from our buildsystem, but it also alows
@@ -80,8 +97,17 @@ class DevBuildsys(Buildsystem):
     __rpms__ = []
 
     def __init__(self):
-        self.multicall = False
-        self.multicallresult = []
+        self._multicall = False
+        self.multicall_result = []
+
+    @property
+    def multicall(self):
+        return self._multicall
+
+    @multicall.setter
+    def multicall(self, value):
+        self._multicall = value
+        self.multicall_result = []
 
     @classmethod
     def clear(cls):
@@ -92,7 +118,9 @@ class DevBuildsys(Buildsystem):
         cls.__rpms__ = []
 
     def multiCall(self):
-        return self.multicallresult
+        result = self.multicall_result
+        self.multicall = False
+        return result
 
     def moveBuild(self, from_tag, to_tag, build, *args, **kw):
         log.debug("moveBuild(%s, %s, %s)" % (from_tag, to_tag, build))
@@ -120,6 +148,7 @@ class DevBuildsys(Buildsystem):
             {'package_id': 2625, 'package_name': 'nethack'},
         ]
 
+    @multicall_enabled
     def getBuild(self, build='TurboGears-1.0.2.2-2.fc17', other=False, testing=False):
         theid = 16058
         if other and not testing:
@@ -128,6 +157,7 @@ class DevBuildsys(Buildsystem):
             theid = 16060
         data = {'build_id': 16058,
                 'completion_time': '2007-08-24 23:26:10.890319',
+                'completion_ts': 1187997970,
                 'creation_event_id': 151517,
                 'creation_time': '2007-08-24 19:38:29.422344',
                 'extra': None,
@@ -258,6 +288,7 @@ class DevBuildsys(Buildsystem):
                 result += [{'name': tag}]
         return result
 
+    @multicall_enabled
     def listTagged(self, tag, *args, **kw):
         builds = []
         for build in [self.getBuild(),
@@ -273,11 +304,7 @@ class DevBuildsys(Buildsystem):
                 if tag_ == tag:
                     builds.append(self.getBuild(build))
         log.debug(builds)
-        if self.multicall:
-            for build in builds:
-                self.multicallresult.append([[build]])
-        else:
-            return builds
+        return builds
 
     def getLatestBuilds(self, *args, **kw):
         return [self.getBuild()]
--- a/bodhi/server/models.py
+++ b/bodhi/server/models.py
@@ -2258,27 +2258,54 @@ class Update(Base):
             # https://github.com/fedora-infra/bodhi/issues/362
             since = self.last_modified.isoformat().rsplit('.', 1)[0]
         except Exception as e:
+            log.exception("Failed to determine last_modified from %r : %r",
+                          self.last_modified, e.message)
             return False, "Failed to determine last_modified: %r" % e.message
 
         try:
-            query = dict(title=self.title, since=since)
+            # query results for this update
+            query = dict(type='bodhi_update', item=self.alias, since=since,
+                         testcases=','.join(requirements))
             results = bodhi.server.util.taskotron_results(settings, **query)
-        except IOError as e:
-            return False, "Failed to talk to taskotron: %r" % e.message
+
+            # query results for each build
+            # retrieve timestamp for each build so that queries can be optimized
+            koji = buildsys.get_session()
+            koji.multicall = True
+            for build in self.builds:
+                koji.getBuild(build.nvr)
+            buildinfos = koji.multiCall()
+
+            for index, build in enumerate(self.builds):
+                buildinfo = buildinfos[index]
+                if not isinstance(buildinfo, dict):
+                    log.warn("Koji build info doesn't seem correct: %r",
+                             buildinfo)
+                ts = datetime.utcfromtimestamp(buildinfo['completion_ts']).isoformat()
+
+                query = dict(type='koji_build', item=build.nvr, since=ts,
+                             testcases=','.join(requirements))
+                build_results = bodhi.server.util.taskotron_results(settings, **query)
+                results.extend(build_results)
+
+        except Exception as e:
+            log.exception("Failed retrieving requirements results: %r", e.message)
+            return False, "Failed retrieving requirements results: %r" % e.message
 
         for testcase in requirements:
             relevant = [result for result in results
                         if result['testcase']['name'] == testcase]
 
             if not relevant:
-                return False, 'No result found for required %s' % testcase
+                return False, 'No result found for required testcase %s' % testcase
 
             by_arch = defaultdict(list)
-            for r in relevant:
-                by_arch[r['result_data'].get('arch', ['noarch'])[0]].append(r)
+            for result in relevant:
+                arch = result['data'].get('arch', ['noarch'])[0]
+                by_arch[arch].append(result)
 
             for arch, results in by_arch.items():
-                latest = results[0]  # TODO - do these need to be sorted still?
+                latest = results[0]  # resultsdb results are ordered chronologically
                 if latest['outcome'] not in ['PASSED', 'INFO']:
                     return False, "Required task %s returned %s" % (
                         latest['testcase']['name'], latest['outcome'])
@@ -2342,7 +2369,7 @@ class Update(Base):
         """ Return the last time this update was edited or created.
 
         This gets used specifically by taskotron/resultsdb queries so we only
-        query for depcheck runs that occur *after* the last time this update
+        query for test runs that occur *after* the last time this update
         (in its current form) was in play.
         """
 
--- a/bodhi/server/models.py
+++ b/bodhi/server/models.py
@@ -2255,7 +2255,7 @@ class Update(Base):
         requirements = list(requirements)
 
         if not requirements:
-            return True, "All checks pass."
+            return True, "No checks required."
 
         try:
             # https://github.com/fedora-infra/bodhi/issues/362
@@ -2284,7 +2284,7 @@ class Update(Base):
                 if (not isinstance(multicall_response, list) or
                         not isinstance(multicall_response[0], dict)):
                     msg = ("Error retrieving data from Koji for %r: %r" %
-                           build.nvr, multicall_response)
+                           (build.nvr, multicall_response))
                     log.error(msg)
                     raise TypeError(msg)
 
@@ -2312,8 +2312,8 @@ class Update(Base):
                 arch = result['data'].get('arch', ['noarch'])[0]
                 by_arch[arch].append(result)
 
-            for arch, results in by_arch.items():
-                latest = results[0]  # resultsdb results are ordered chronologically
+            for arch, result in by_arch.items():
+                latest = relevant[0]  # resultsdb results are ordered chronologically
                 if latest['outcome'] not in ['PASSED', 'INFO']:
                     return False, "Required task %s returned %s" % (
                         latest['testcase']['name'], latest['outcome'])
--- a/bodhi/server/models.py
+++ b/bodhi/server/models.py
@@ -2254,6 +2254,9 @@ class Update(Base):
         requirements = tokenize(self.requirements or '')
         requirements = list(requirements)
 
+        if not requirements:
+            return True, "All checks pass."
+
         try:
             # https://github.com/fedora-infra/bodhi/issues/362
             since = self.last_modified.isoformat().rsplit('.', 1)[0]
@@ -2266,7 +2269,7 @@ class Update(Base):
             # query results for this update
             query = dict(type='bodhi_update', item=self.alias, since=since,
                          testcases=','.join(requirements))
-            results = bodhi.server.util.taskotron_results(settings, **query)
+            results = list(bodhi.server.util.taskotron_results(settings, **query))
 
             # query results for each build
             # retrieve timestamp for each build so that queries can be optimized
@@ -2277,15 +2280,20 @@ class Update(Base):
             buildinfos = koji.multiCall()
 
             for index, build in enumerate(self.builds):
-                buildinfo = buildinfos[index]
-                if not isinstance(buildinfo, dict):
-                    log.warn("Koji build info doesn't seem correct: %r",
-                             buildinfo)
+                multicall_response = buildinfos[index]
+                if (not isinstance(multicall_response, list) or
+                        not isinstance(multicall_response[0], dict)):
+                    msg = ("Error retrieving data from Koji for %r: %r" %
+                           build.nvr, multicall_response)
+                    log.error(msg)
+                    raise TypeError(msg)
+
+                buildinfo = multicall_response[0]
                 ts = datetime.utcfromtimestamp(buildinfo['completion_ts']).isoformat()
 
                 query = dict(type='koji_build', item=build.nvr, since=ts,
                              testcases=','.join(requirements))
-                build_results = bodhi.server.util.taskotron_results(settings, **query)
+                build_results = list(bodhi.server.util.taskotron_results(settings, **query))
                 results.extend(build_results)
 
         except Exception as e:
--- a/bodhi/server/templates/update.html
+++ b/bodhi/server/templates/update.html
@@ -103,10 +103,6 @@ ${update.type} update for ${update.beautify_title(amp=True) | n}
       // the latest ones for each arch.  So, prune like this:
       $.each(data.data, function(i, result) {
 
-        // However, we want to skip over any ABORTED results:
-        // https://github.com/fedora-infra/bodhi/issues/167
-        if (result.outcome == 'ABORTED') return;
-
         var name = result.testcase.name;
         // we want to stash results by scenario when there is one, use name
         // when there isn't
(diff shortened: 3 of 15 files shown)


bowlofeggs:
Hey @kparal!

This pull request does cause quite a few test failures and errors (13 and 3, respectively), as you guessed ☺ When you are in the vagrant box you can run the tests with ```btest```. You can also use handy flags, like ```-x``` to exit upon the first failure. When you ```vagrant ssh``` into the box, you should see a handy motd printed to your terminal that tells you about btest and some other handy aliases.

Here's a snippet of one of the errors:

```
FAIL: Test submitting a stable request, but with bad taskotron results
  File "/home/vagrant/bodhi/bodhi/tests/server/services/test_updates.py", line 2330, in test_stable_request_failed_taskotron_results
    self.assertIn('Required task', resp)
AssertionError: 'Required task' not found in <400 Bad Request application/json body='{"status"..."}]}'/176>
bodhi: INFO: Using the FakeBugTracker
bodhi.server: DEBUG: Creating all models for Engine(sqlite://)
bodhi.server: DEBUG: Setting alias for bodhi-2.0-1.fc17 to FEDORA-2017-a3bbe1a8f2
bodhi: INFO: Using the FakeBugTracker
dogpile.lock: DEBUG: NeedRegenerationException
dogpile.lock: DEBUG: no value, waiting for create lock
dogpile.lock: DEBUG: value creation lock <dogpile.cache.region._LockWrapper object at 0x7f50cc581290> acquired
dogpile.lock: DEBUG: Calling creation function
dogpile.lock: DEBUG: Released creation lock
bodhi.server: DEBUG: validated = {'close_bugs': True, 'stable_karma': 3, 'csrf_token': u'f6a9de0b52ec76952a0400e69586b0421fed9952', 'requirements': u'rpmlint', 'builds': [u'bodhi-2.0.0-3.fc17'], 'autokarma': True, 'edited': '', 'suggest'
: <unspecified>, 'notes': u'this is a test update', 'request': <testing>, 'bugs': [], 'unstable_karma': -3, 'require_bugs': False, 'require_testcases': True, 'type': <bugfix>, 'severity': <unspecified>}
bodhi.server: DEBUG: Adding nvr bodhi-2.0.0-3.fc17
bodhi.server: INFO: Creating new update: [<RpmBuild {'epoch': 0, 'type': u'rpm', 'nvr': u'bodhi-2.0.0-3.fc17', 'signed': False}>]
bodhi.server: DEBUG: Creating new Update(**data) object.
bodhi.server: DEBUG: Assigning alias for new update..
bodhi.server: DEBUG: Setting alias for bodhi-2.0.0-3.fc17 to FEDORA-2017-48150de3bc
bodhi.server: DEBUG: Setting request for new update.
bodhi.server: DEBUG: Attempting to set request <testing>
bodhi.server: DEBUG: Adding tag f17-updates-testing-signing to bodhi-2.0.0-3.fc17
bodhi: DEBUG: tagBuild(f17-updates-testing-signing, bodhi-2.0.0-3.fc17)
bodhi.server: DEBUG: bodhi-2.0.0-3.fc17 has been submitted for testing. 
bodhi.server: INFO: Sending mail to guest: [Fedora Update] [comment] bodhi-2.0.0-3.fc17
bodhi.server: INFO: Not sending email: No smtp_server defined
bodhi.server: DEBUG: Adding new update to the db.
bodhi.server: DEBUG: Triggering db flush for new update.
bodhi.server: DEBUG: Done with Update.new(...)
bodhi.server: DEBUG: bodhi-2.0.0-3.fc17 update created
bodhi.server: INFO: Sending mail to guest: [Fedora Update] [comment] bodhi-2.0.0-3.fc17
bodhi.server: INFO: Not sending email: No smtp_server defined
dogpile.lock: DEBUG: NeedRegenerationException
dogpile.lock: DEBUG: no value, waiting for create lock
dogpile.lock: DEBUG: value creation lock <dogpile.cache.region._LockWrapper object at 0x7f50ccbf1890> acquired
dogpile.lock: DEBUG: Calling creation function
dogpile.lock: DEBUG: Released creation lock
bodhi.server: DEBUG: validating request
```

As for ```check_requirements()```, I do see calls to it in two places:

0. [The masher seems to call it](https://github.com/fedora-infra/bodhi/blob/2.6.2/bodhi/server/consumers/masher.py#L434-L437) before sending the update out, and seems to reject it if result doesn't evaluate "truthy".
1. The [set_request update handler seems to call it](https://github.com/fedora-infra/bodhi/blob/2.6.2/bodhi/server/services/updates.py#L127-L133) when the update is requested for stable via the API.

Weirdly, it doesn't seem to get used when an update gets set for stable based on karma, though the masher theoretically would still reject it.

So to trigger it in the vagrant environment, I'd say make a call with the CLI (which is automatically configured to talk to localhost in the vagrant box) to request stable. In order to do that, you will also need to fake the update meeting testing requirements. You can make it look like an update has been in testing for ```n``` days with bshell:

```
$ bshell
<snip>
In [1]: u = m.Update.query.filter(m.Update.alias=="FEDORA-2017-afd3f53b9d").one()
/usr/lib64/python2.7/site-packages/sqlalchemy/sql/sqltypes.py:185: SAWarning: Unicode type received non-unicode bind param value 'FEDORA-2017-afd3f53b9d'. (this warning may be suppressed after 10 occurrences)
  (util.ellipses_string(value),))

In [3]: import datetime

In [4]: u.date_testing = datetime.datetime.now() - datetime.timedelta(days=15)

In [5]: s().add(u)

In [6]: s().commit()
```

That will make the update be there long enough, and then you can just do the "normal" things to try to push it to stable and that should call ```check_requirements()```.

Let me know if you need further help getting the tests to pass, or if you think I need to explain something more clearly!

bodhi/server/models.py
```diff
@@ -2056,24 +2056,48 @@ def check_requirements(self, session, settings):
...
+            # query results for each build
+            # retrieve timestamp for each build so that queries can be optimized
+            koji = buildsys.get_session()
+            koji.multicall = True
+            for build in self.builds:
+                koji.getBuild(build.nvr)
+            buildinfos = koji.multiCall()
+
+            for index, build in enumerate(self.builds):
+                buildinfo = buildinfos[index]
```
bowlofeggs: I'm not very familiar with the Koji API - are the results of the multCall() guaranteed to be in the same order as the nvrs we passed above? Thus making this buildinfo definitely correlate to the build?

bodhi/server/models.py
```diff
@@ -2056,24 +2056,48 @@ def check_requirements(self, session, settings):
...
+                if not isinstance(buildinfo, dict):
+                    log.warn("Koji build info doesn't seem correct: %r",
+                             buildinfo)
```
bowlofeggs: It seems like we'd want to raise an Exception here, perhaps a ```TypeError```.

bodhi/server/models.py
```diff
@@ -2056,24 +2056,48 @@ def check_requirements(self, session, settings):
...
+                ts = datetime.utcfromtimestamp(buildinfo['completion_ts']).isoformat()
+
+                query = dict(type='koji_build', item=build.nvr, since=ts,
+                             testcases=','.join(requirements))
+                build_results = bodhi.server.util.taskotron_results(settings, **query)
+                results.extend(build_results)
+
+        except Exception as e:
+            return False, "Failed retrieving requirements results: %r" % e.message
```
bowlofeggs: We should probably do a log.exception() here too so we can have a record of what happened.

bodhi/server/models.py
```diff
@@ -2056,24 +2056,48 @@ def check_requirements(self, session, settings):
...
             by_arch = defaultdict(list)
-            for r in relevant:
-                by_arch[r['result_data'].get('arch', ['noarch'])[0]].append(r)
+            for result in relevant:
+                arch = result['data'].get('arch', ['noarch'])[0]
+                by_arch[arch].append(result)
 
             for arch, results in by_arch.items():
-                latest = results[0]  # TODO - do these need to be sorted still?
+                latest = results[0]  # resultsdb results are ordered chronologically
```
bowlofeggs: Thanks for answering the question ☺

bodhi/server/util.py
```diff
@@ -594,27 +594,44 @@ def tokenize(string):
...
+    """ Yield resultsdb results using query arguments.
+
+    :param str entity: The API endpoint to use, see resultsdb documentation.
+    :param int max_queries: The maximum number of queries to perform (pages
+        to retrieve). So ``1`` means just a single page. ``None`` or ``0``
+        means no limit. Please note some tests might have thousands of results
+        in the database and it's very reasonable to limit queries (thus the
+        default value).
+    :param kwargs: Any args that will be passed to resultsdb to specify what
+        results to retrieve.
```
bowlofeggs, line 636: +1 I love documented parameters!

bodhi/server/util.py
```diff
@@ -594,27 +594,44 @@ def tokenize(string):
...
                 # https://github.com/fedora-infra/bodhi/issues/167
-                if entity == 'results' and datum.get('outcome') == 'ABORTED':
+                if 'results' in entity and datum.get('outcome') == 'ABORTED':
                     continue
                 yield datum
+
+            url = json.get('next')
+            queries += 1
+            if queries >= max_queries and url:
+                log.warn('Too many result pages, aborting at: %r' % url)
```
bowlofeggs: Should this be a warning, or maybe just an info or even a debug? After all, the caller asked for ```max_queries``` so I'm not sure we should warn that we hit it.

production.ini
```diff
@@ -204,9 +204,7 @@ stacks_enabled = False
 # updates.  Users have free-reign to override them for each kind of entity.  At
 # the end of the day, we only consider the requirements defined by single
 # updates themselves when gating in the backend masher process.
-site_requirements = depcheck upgradepath
-## Some day we'll have rpmgrill, and that will be cool.  Ask tflink.
-#site_requirements = depcheck upgradepath rpmgrill
+site_requirements = dist.rpmdeplint dist.upgradepath
```
bowlofeggs: I've got a PR open to rework basically everything about how settings in Bodhi work: #1544. If that gets merged first, we should make sure we make these changes in the new ```bodhi.server.config.BodhiConfig._defaults```. If not, I'll have to be sure to add this there.

bowlofeggs (changes requested):
I've got a few notes for what you have so far too. Thanks for the PR!

bodhi/server/models.py
kparal: Yes, this is the docstring:

>     def multiCall(self, strict=False):
>         """Execute a multicall (multiple function calls passed to the server
>         and executed at the same time, with results being returned in a batch).
>         Before calling this method, the self.multicall field must have
>         been set to True, and then one or more methods must have been called on
>         the current session (those method calls will return None).  On executing
>         the multicall, the self.multicall field will be reset to False
>         (so subsequent method calls will be executed immediately)
>         and results will be returned in a list.  The list will contain one element
>         for each method added to the multicall, in the order it was added to the multicall.
>         Each element of the list will be either a one-element list containing the result of the
>         method call, or a map containing "faultCode" and "faultString" keys, describing the
>         error that occurred during the method call."""
>

bodhi/server/models.py
kparal: It will fail on the next step, probably raising `TypeError`. Do you think I should raise that immediately?

bodhi/server/models.py
kparal: Good idea, will do.

bodhi/server/util.py
kparal: You're right, I wrote this before I added `max_queries` as a method argument. Now it's much better to have it as debug.

kparal:
> When you are in the vagrant box you can run the tests with btest. You can also use handy flags, like -x to exit upon the first failure.

I was hoping you do the tests for me :-P OK, will try.

> As for check_requirements(), I do see calls to it in two places:
> 
>     The masher seems to call it before sending the update out, and seems to reject it if result doesn't evaluate "truthy".
>     The set_request update handler seems to call it when the update is requested for stable via the API.
> 
> Weirdly, it doesn't seem to get used when an update gets set for stable based on karma, though the masher theoretically would still reject it.

When the masher rejects it, is it somehow visible in Bodhi UI (a comment added to the update)? I remember seeing a comment like "autokarma push disabled, because an important test failed" (or something along those lines). Is that it, or is it something else? And where does that code live?

There's one important piece of information. Due to how our tests are currently written, `depcheck` is executed immediately for everything in `fxx-updates-testing-pending` and `fxx-updates-pending`, but `upgradepath` is only executed for packages in `fxx-updates-pending`. So until the user manually requests a push to stable (or autokarma limit is reached), the package has no `upgradepath` result. As I currently see the code, that would prevent manual pushing to updates. And it could also make a race condition with the masher, if the result doesn't arrive in time before the masher processes it. But that doesn't seem to actually happen (updates can be pushed stable manually), so either the code is not really being used, or I'm missing something. Thoughts?

> You can make it look like an update has been in testing for n days with bshell:

Is it possible to give myself superuser permissions to do anything and everything, without tinkering with the database? I added a group I'm in into `admin_packager_groups` in `development.ini`, and that allowed me to create updates for any package, but I still can't push it until the requirements are met (days in testing, karma -- and I can't give karma to my updates). Thanks for that bshell help, though!

kparal:
It was a tough fight, but the test suite is working now :) I had to do some major changes, please review. I also added more optimizations.

Bodhi UI is working for me, I can set required testcases (name verification works), I can't push stable until they pass (nice). I don't know how to verify the masher, but the test suite runs...

What I think doesn't work and hasn't worked are `site_requirements`. It seems they are really not used anywhere, or at least I don't see it. If I create a new update, by default the testcases field is empty, and it means "no requirements". So the site defaults doesn't really seem to be used. I don't know if that's a problem or not, as I said, we still haven't fine-tuned all the processes (especially the upgradepath is getting run late), so it might be fine having it just like this for now.

What I'd like to see and verify is that karma autopush is disabled if any of `site_requirements` fail. I think Bodhi should have such a feature already, but I can't locate it.

production.ini
bowlofeggs: #1544 is now merged.

bodhi/server/buildsys.py
```diff
@@ -25,7 +26,23 @@
...
+        # disable multicall during execution, so that inner func calls to other
+        # methods don't append their results as well
+        self._multicall = False
+        result = func(self, *args, **kwargs)
+        self.multicall_result.append([result])
+        self._multicall = True
+    return wrapper
+
+
+class Buildsystem(object):
```
bowlofeggs, line 45: Is this change needed for the goal at hand? If not, I suggest making it a separate pull request.

bodhi/server/models.py
```diff
@@ -2049,31 +2049,66 @@ def check_requirements(self, session, settings):
         requirements = tokenize(self.requirements or '')
         requirements = list(requirements)
 
+        if not requirements:
+            return True, "All checks pass."
```
bowlofeggs: Maybe change "All checks pass." to "No checks required."

bodhi/server/templates/update.html
```diff
@@ -128,10 +128,6 @@
       // the latest ones for each arch.  So, prune like this:
       $.each(data.data, function(i, result) {
 
-        // However, we want to skip over any ABORTED results:
-        // https://github.com/fedora-infra/bodhi/issues/167
-        if (result.outcome == 'ABORTED') return;
```
bowlofeggs, line 108: Do we not want to skip aborted results anymore?

development.ini.example
```diff
@@ -216,9 +216,7 @@ stacks_enabled = False
 # updates.  Users have free-reign to override them for each kind of entity.  At
 # the end of the day, we only consider the requirements defined by single
 # updates themselves when gating in the backend masher process.
-site_requirements = depcheck upgradepath
-## Some day we'll have rpmgrill, and that will be cool.  Ask tflink.
-#site_requirements = depcheck upgradepath rpmgrill
+site_requirements = dist.rpmdeplint dist.upgradepath
```
bowlofeggs: Please rebase this PR on the develop branch - once you've done that, please also make this setting commented by default (similar to most of the rest of the file) and then add this default to ```bodhi.server.config.BodhiConfig._defaults```. Thanks!

bowlofeggs (changes requested):
The tests pass, but there are a number of lines of code that are missing test coverage:

```
$ diff-cover coverage.xml --compare-branch=develop --fail-under=100
-------------
Diff Coverage
-------------
bodhi/server/buildsys.py (100%)
bodhi/server/models.py (75.9%): Missing lines 2059,2081,2083-2084,2094-2096
bodhi/server/util.py (9.1%): Missing lines 609-610,614,617,623,626-630
bodhi/server/validators.py (42.9%): Missing lines 929-930,932-933
bodhi/server/views/generic.py (100%)
-------------
Total:   75 lines
Missing: 21 lines
Coverage: 72%
```

bodhi/server/buildsys.py
bowlofeggs: By "this change" I mean explicitly subclassing object here.

kparal: It's needed for a class to have properties. I made `multicall` a property, in order to fix previous incorrect multicall implementation (it wasn't possible to test multicall on methods which called each other).

bodhi/server/templates/update.html
kparal: We no longer allow resultsdb to store them.

kparal:
After updating `develop`, I can't reprovision vagrant anymore, even if I do `vagrant destroy && vagrant up`. Seems something broke:
```
TASK [dev : Apply database migrations] *****************************************
fatal: [bodhi]: FAILED! => {"changed": true, "cmd": ["alembic", "upgrade", "head"], "delta": "0:00:00.920123", "end": "2017-06-07 13:08:01.694655", "failed": true, "rc": 1, "start": "2017-06-07 13:08:00.774532", "stderr": "Traceback (most recent call last):\n  File \"/bin/alembic\", line 9, in <module>\n    load_entry_point('alembic==0.8.6', 'console_scripts', 'alembic')()\n  File \"/usr/lib/python2.7/site-packages/alembic/config.py\", line 479, in main\n    CommandLine(prog=prog).main(argv=argv)\n  File \"/usr/lib/python2.7/site-packages/alembic/config.py\", line 473, in main\n    self.run_cmd(cfg, options)\n  File \"/usr/lib/python2.7/site-packages/alembic/config.py\", line 456, in run_cmd\n    **dict((k, getattr(options, k)) for k in kwarg)\n  File \"/usr/lib/python2.7/site-packages/alembic/command.py\", line 174, in upgrade\n    script.run_env()\n  File \"/usr/lib/python2.7/site-packages/alembic/script/base.py\", line 397, in run_env\n    util.load_python_file(self.dir, 'env.py')\n  File \"/usr/lib/python2.7/site-packages/alembic/util/pyfiles.py\", line 93, in load_python_file\n    module = load_module_py(module_id, path)\n  File \"/usr/lib/python2.7/site-packages/alembic/util/compat.py\", line 79, in load_module_py\n    mod = imp.load_source(module_id, path, fp)\n  File \"alembic/env.py\", line 17, in <module>\n    config.set_main_option(\"sqlalchemy.url\", bodhi_config[\"sqlalchemy.url\"])\n  File \"/home/vagrant/bodhi/bodhi/server/config.py\", line 551, in __getitem__\n    self.load_config()\n  File \"/home/vagrant/bodhi/bodhi/server/config.py\", line 584, in load_config\n    self._validate()\n  File \"/home/vagrant/bodhi/bodhi/server/config.py\", line 602, in _validate\n    'Invalid config values were set: \\n{}'.format('\\n'.join(errors)))\nValueError: Invalid config values were set: \n\tcompose_atomic_trees: \"\" cannot be interpreted as a boolean value.", "stderr_lines": ["Traceback (most recent call last):", "  File \"/bin/alembic\", line 9, in <module>", "    load_entry_point('alembic==0.8.6', 'console_scripts', 'alembic')()", "  File \"/usr/lib/python2.7/site-packages/alembic/config.py\", line 479, in main", "    CommandLine(prog=prog).main(argv=argv)", "  File \"/usr/lib/python2.7/site-packages/alembic/config.py\", line 473, in main", "    self.run_cmd(cfg, options)", "  File \"/usr/lib/python2.7/site-packages/alembic/config.py\", line 456, in run_cmd", "    **dict((k, getattr(options, k)) for k in kwarg)", "  File \"/usr/lib/python2.7/site-packages/alembic/command.py\", line 174, in upgrade", "    script.run_env()", "  File \"/usr/lib/python2.7/site-packages/alembic/script/base.py\", line 397, in run_env", "    util.load_python_file(self.dir, 'env.py')", "  File \"/usr/lib/python2.7/site-packages/alembic/util/pyfiles.py\", line 93, in load_python_file", "    module = load_module_py(module_id, path)", "  File \"/usr/lib/python2.7/site-packages/alembic/util/compat.py\", line 79, in load_module_py", "    mod = imp.load_source(module_id, path, fp)", "  File \"alembic/env.py\", line 17, in <module>", "    config.set_main_option(\"sqlalchemy.url\", bodhi_config[\"sqlalchemy.url\"])", "  File \"/home/vagrant/bodhi/bodhi/server/config.py\", line 551, in __getitem__", "    self.load_config()", "  File \"/home/vagrant/bodhi/bodhi/server/config.py\", line 584, in load_config", "    self._validate()", "  File \"/home/vagrant/bodhi/bodhi/server/config.py\", line 602, in _validate", "    'Invalid config values were set: \\n{}'.format('\\n'.join(errors)))", "ValueError: Invalid config values were set: ", "\tcompose_atomic_trees: \"\" cannot be interpreted as a boolean value."], "stdout": "", "stdout_lines": []}
	to retry, use: --limit @/home/kparal/devel/forks/bodhi/devel/ansible/playbook.retry
```

bowlofeggs:
On Wed, 2017-06-07 at 05:20 -0700, Kamil Páral wrote:
> It's needed for a class to have properties. I made multicall a
> property, in order to fix previous incorrect multicall implementation
> (it wasn't possible to test multicall on methods which called each
> other).

Cool.

bowlofeggs:
On Wed, 2017-06-07 at 05:22 -0700, Kamil Páral wrote:
> We no longer allow resultsdb to store them.

OK, thanks!

kparal:
OK, tests added. I hope this is the last iteration, because I can't really spend more time on this.

bowlofeggs (approved):
Looks great - very nice tests. I'm going to squash the commits down to one and merge. Thanks!

kparal:
Thanks @bowlofeggs .

Issue #1515, opened 2017-05-12
Cron job update CI status in addition to fedmsg


pypingou:
We would like a cron job that would be checking and updating the CI status of all the updates for which we are missing this information.

This would help mitigate potential network issues (ie: missed fedmsg message)

bowlofeggs:
This depends on #1499, and the commit hash patch that I've been working on depends on this.

bowlofeggs:
We plan to use [Greenwave](https://pagure.io/greenwave) as the "source of truth" for whether updates pass CI tests. There is [one issue](https://pagure.io/greenwave/issue/26) that blocks work on this ticket. Also, since the [Greenwave response's schema is undocumented](https://pagure.io/greenwave/issue/25), I am including a link to a demo that shows an example call/response or two:

https://fedorapeople.org/groups/factory2/sprint-033/dcallagh-greenwave-multiple-products.ogv

bowlofeggs:
The CI status feature got removed recently.

PR #1555, opened 2017-05-26
Add a commit field to the Build model.

bowlofeggs:

### DIFF
--- a/bodhi/server/scripts/babysit_ci_status.py
+++ b/bodhi/server/scripts/babysit_ci_status.py
@@ -1,0 +1,40 @@
+# -*- coding: utf-8 -*-
+# Copyright ® 2017 Red Hat, Inc.
+#
+# This file is part of Bodhi.
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+"""Iterate updates that don't have final CI statuses to see if we missed messages about them."""
+import click
+
+from bodhi.server import config, initialize_db, models, Session
+
+
+@click.command()
+@click.version_option(message='%(version)s')
+def babysit():
+    """Iterate over updates that don't have final CI statuses to see if we missed messages."""
+    initialize_db(config.config)
+    session = Session()
+
+    for b in models.Build.query.filter(models.Build.ci_status.in_(
+            [models.CiStatus.waiting, models.CiStatus.queued, models.CiStatus.running])):
+        # We might have failed to look up the scm_url when the update was created, so let's make
+        # sure we know one now so we can use it to look up test results with it.
+        if not b.scm_url:
+            b.scm_url = b.get_scm_url()
+
+        if session.is_modified(b):
+            session.commit()
--- a/bodhi/tests/server/consumers/test_updates.py
+++ b/bodhi/tests/server/consumers/test_updates.py
@@ -21,7 +21,7 @@ import unittest
 import mock
 import sqlalchemy
 
-from bodhi.server import exceptions, util
+from bodhi.server import exceptions, models, util
 from bodhi.server.consumers import updates
 from bodhi.tests.server import base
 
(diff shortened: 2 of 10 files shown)


alembic/versions/0e105f6e36b4_add_a_commit_column_to_the_builds_table.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+# revision identifiers, used by Alembic.
+revision = '0e105f6e36b4'
+down_revision = 'c6a5e2849ca4'
+
+
+def upgrade():
+    """Add a commit column to the builds table with a uniqueness constraint."""
+    op.add_column('builds', sa.Column('commit', sa.Unicode(length=100), nullable=True))
```
pypingou: This is likely going to be too short:
````
>>> len('git://pkgs.fedoraproject.org/rpms/rubygem-activerecord-deprecated_finde
rs?#34e14931464563b7d549816d1a2324c9af444d8e')
115
````

**Comment by pypingou** [comment]
One remark about the length planned at the moment.

Also, when is ``get_commit()`` supposed to be called?

File: alembic/versions/0e105f6e36b4_add_a_commit_column_to_the_builds_table.py (hunk 1 above)
Comment (pypingou) [reply alembic/versions/0e105f6e36b4_add_a_commit_column_to_the_builds_table.py]: FTR currently:
````
pkgdb2=# SELECT length(name) as l, name from package order by l DESC limit 1;
 l  |                        name                         
----+-----------------------------------------------------
 51 | golang-github-matttproud-golang_protobuf_extensions
(1 row)
````

**Comment by bowlofeggs** [comment]
@pypingou Thanks for the note about the commit hash length - I'll raise the size of it when I get back to working on this. As for ```get_commit()``` — I plan for it to get used in ```bodhi.server.consumers.updates``` as one more background task for Bodhi to do when new updates get created. I also plan for there to be code in the background cron task (See #1515) that can add the hash if the updates consumer wasn't able to get it for some reason.

**Comment by pypingou** [comment]
Ok, so now that #1499 is in, how can I help to get this one in ? :)

**Comment by bowlofeggs** [comment]
@pypingou #1515 is sort of a blocker for this one, because I want to add code to the cron job for when this code fails to get the commit from koji. So once #1515 is merged, I'll rebase this one on it and I'll modify that cron job in light of this code.

**Comment by pypingou** [comment]
#1515 says that it depends on #1499 which has not gotten merged, so I guess we can rebase this PR and start working on the cron that will fix #1515 and allow merging this PR :)

**Comment by bowlofeggs** [comment]
On Thu, 2017-06-08 at 08:19 -0700, Pierre-Yves Chibon wrote:
> #1515 says that it depends on #1499 which has not gotten merged

I assume you mean "now" gotten merged? So yeah, let's work on #1515 and
get it merged, and then I will rebase this one and finish it!

File: alembic/versions/0e105f6e36b4_add_a_commit_column_to_the_builds_table.py (hunk 1 above)
Comment (bowlofeggs) [reply alembic/versions/0e105f6e36b4_add_a_commit_column_to_the_builds_table.py]: OK, so the longest project name we have with a current git hash would be 127 characters:

```
>>> len("git://pkgs.fedoraproject.org/rpms/golang-github-matttproud-golang_protobuf_extensions?#34e14931464563b7d549816d1a2324c9af444d8e")
127
```

If git switches to a sha512 later, that same long name would be 215 characters:

```
>>> len("git://pkgs.fedoraproject.org/rpms/golang-github-matttproud-golang_protobuf_extensions?#e7c22b994c59d9cf2b48e549b1e24666636045930d3da7c1acb299d1c3b7f931f94aae41edda2c2b207a36e10f8bcb8d45223e54878f5b316e7ce3b6bc019629")
215
```

I propose we use 256 characters since that's a pleasing power of 2 and gives even a little more room for growth? Or we could just use 150 or so and bump it as needed, but I say why not just go ahead and make room for sha512.

Comment (pypingou) [reply alembic/versions/0e105f6e36b4_add_a_commit_column_to_the_builds_table.py]: +1 for 256, I wonder if MySQL doesn't support varchar up to 255 so we may want to use this instead but either one is fine for me

**Comment by bowlofeggs** [comment]
This is now ready for review.

File: bodhi/tests/server/consumers/test_updates.py [hunk 2]
```diff
@@ -117,6 +117,70 @@ def test_request_testing(self, work_on_bugs, fetch_test_cases):
...
+        """Test that the scm_url is left alone when it's already set in the database."""
+        b = models.Build.query.filter(models.Build.nvr == u'bodhi-2.0-1.fc17').one()
+        b.scm_url = u'git://some.cool.host/rpms/neat#master'
+        self.db.commit()
+        hub = mock.MagicMock()
+        hub.config = {'environment': 'environment',
+                      'topic_prefix': 'topic_prefix'}
+        h = updates.UpdatesHandler(hub)
+        h.db_factory = base.TransactionalSessionMaker(self.Session)
+        # Throw a bogus bug id in there to ensure it doesn't raise AssertionError.
```
jeremycline: This comment is confusing to me. I don't see a bug id.

jeremycline:
This generally looks fine to me.

One thing I thought about a little (and it's probably over-engineering or something) is whether it makes sense to store the URL vs. the commit alone: 
```'git://pkgs.fedoraproject.org/rpms/bodhi?#2e994ca8b3296e62e8b0aadee1c5c0649559625a'```
on a build object seems a bit redundant. It's really 5 pieces of information:
1. The protocol (``git://``)
2. The host (``pkgs.fedoraproject.org``)
3. The build type (``/rpms/``)
4. The package name (``bodhi``)
5. The commit hash.

Will 1 and 2 ever vary between updates? Are we okay writing a migration if/when the protocol or hostname changes? Would it be better for that to just be a configuration? I'm not crazy about weird configuration options so I'm on the fence about this, just thought I'd bring it up.

3 is already stored on the Build model as the ``type`` and 4 is part of the ``NVR``, also stored on the model.

File: bodhi/server/scripts/babysit_ci_status.py [hunk 3]
```diff
@@ -0,0 +1,40 @@
...
+
+@click.command()
+@click.version_option(message='%(version)s')
+def babysit():
+    """Iterate over updates that don't have final CI statuses to see if we missed messages."""
+    initialize_db(config.config)
+    session = Session()
+
+    for b in models.Build.query.filter(models.Build.ci_status.in_(
+            [models.CiStatus.waiting, models.CiStatus.queued, models.CiStatus.running])):
```
pypingou, line 33: Should we just filter by ``scm_url == NULL``?

pypingou (commented):
One comment otherwise looks good I think

bodhi/server/scripts/babysit_ci_status.py
bowlofeggs: @pypingou This function is intended to be extended for #1515, so it needs to search for these build specifically. Thus, there are two reasons not to query by ```scm_url``` being ```NULL```:

0. Even if ```scm_url``` is known, we will want to check in on these updates to see if we missed messages about them.
1. We don't really want to query Koji for all historic Build scm_urls - we only really need to know them for builds that we're waiting on CI results.

bowlofeggs:
@jeremycline I had also considered storing just the commit hash. I went with this just because it doesn't require massaging the data out of Koji at all (just take it and store it exactly as is). I don't think we will ever do a migration to update this data - the only reason we are storing the commit hash in the first place is so that we can match CI test results up to builds. Once that is done, the data is never used again. Thus, it isn't very important that it remain "correct" or useful in the long term - it just needs to be right until we know if the CI results passed, failed, or were skipped.

bowlofeggs:
@jeremycline @pypingou Were either of your commented intended to be an LGTM?

jeremycline:
> @jeremycline I had also considered storing just the commit hash. I went with this just because it doesn't require massaging the data out of Koji at all (just take it and store it exactly as is). I don't think we will ever do a migration to update this data - the only reason we are storing the commit hash in the first place is so that we can match CI test results up to builds. Once that is done, the data is never used again. Thus, it isn't very important that it remain "correct" or useful in the long term - it just needs to be right until we know if the CI results passed, failed, or were skipped.

Ah okay, makes sense. Looks good to me, :shipit:

bodhi/tests/server/consumers/test_updates.py
bowlofeggs: @jeremycline you caught me in a copypaste crime! I'm applying this patch to fix it:

```
diff --git a/bodhi/tests/server/consumers/test_updates.py b/bodhi/tests/server/consumers/test_updates.py
index 99db898..abacc7a 100644
--- a/bodhi/tests/server/consumers/test_updates.py
+++ b/bodhi/tests/server/consumers/test_updates.py
@@ -125,7 +125,6 @@ class TestUpdatesHandlerConsume(base.BaseTestCase):
                       'topic_prefix': 'topic_prefix'}
         h = updates.UpdatesHandler(hub)
         h.db_factory = base.TransactionalSessionMaker(self.Session)
-        # Throw a bogus bug id in there to ensure it doesn't raise AssertionError.
         message = {
             'topic': 'bodhi.update.request.testing',
             'body': {'msg': {'update': {'alias': u'bodhi-2.0-1.fc17'}}}}
@@ -148,7 +147,6 @@ class TestUpdatesHandlerConsume(base.BaseTestCase):
                       'topic_prefix': 'topic_prefix'}
         h = updates.UpdatesHandler(hub)
         h.db_factory = base.TransactionalSessionMaker(self.Session)
-        # Throw a bogus bug id in there to ensure it doesn't raise AssertionError.
         message = {
             'topic': 'bodhi.update.request.testing',
             'body': {'msg': {'update': {'alias': u'bodhi-2.0-1.fc17'}}}}
@@ -168,7 +166,6 @@ class TestUpdatesHandlerConsume(base.BaseTestCase):
                       'topic_prefix': 'topic_prefix'}
         h = updates.UpdatesHandler(hub)
         h.db_factory = base.TransactionalSessionMaker(self.Session)
-        # Throw a bogus bug id in there to ensure it doesn't raise AssertionError.
         message = {
             'topic': 'bodhi.update.request.testing',
             'body': {'msg': {'update': {'alias': u'bodhi-2.0-1.fc17'}}}}
```

PR #1556, opened 2017-05-27
Pending update directly requests for stable if it hits stable karma

### LINKED ISSUE #632: should push to stable immediately if will be automatically set to request stable at time of push to testing
Example:
https://bodhi.fedoraproject.org/updates/FEDORA-2015-16569

This update was pushed to testing at 2015-09-27 00:42:54
It was then automatically requested for stable based upon karma 42 seconds later.
Then the package had to wait 15 hours to be pushed to stable, even though it already met the karma requirements for automatic request for stable before the push to testing.

Instead of (or in addition to?) pushing it to testing, it should have been automatically pushed to stable at the time it pushed to testing, rather than having to wait until the next time updates get pushed, which in this case was 15 hours later (in some cases the wait seems to be longer still).

Probably the best way to approach this is for Bodhi to automatically change the request from testing to stable if the Karma requirements for automatic request for stable have been reached before the package makes it to testing.

Timeliness of updates is important for security, for example. In this case, the sqlite package was holding up for several days a security update for firefox which depended upon it (https://bodhi.fedoraproject.org/updates/FEDORA-2015-16455). (Actually, it looks like this firefox package is another example where the push to stable was delayed for the same reason.)

trishnaguha:
Signed-off-by: Trishna Guha <trishnaguha17@gmail.com>
Fixes #632 
- Requests for stable if pending update hits stable karma on Autopush enabled/disabled.

### DIFF
--- a/bodhi/server/models.py
+++ b/bodhi/server/models.py
@@ -2299,11 +2299,11 @@ class Update(Base):
         if self.locked:
             log.debug('%s locked. Ignoring karma thresholds.' % self.title)
             raise LockedUpdateException
-        # Return if the status of the update is not in testing
-        if self.status != UpdateStatus.testing:
+        # Return if the status of the update is not in testing or pending
+        if self.status not in (UpdateStatus.testing, UpdateStatus.pending):
             return
         # If an update receives negative karma disable autopush
-        if self.autokarma and self._composite_karma[1] != 0:
+        if self.autokarma and self._composite_karma[1] != 0 and self.status is UpdateStatus.testing:
             log.info("Disabling Auto Push since the update has received negative karma")
             self.autokarma = False
             text = config.get('disable_automatic_push_to_stable')
@@ -2323,11 +2323,14 @@ class Update(Base):
                     "%s update has reached the stable karma threshold and can be pushed to "
                     "stable now if the maintainer wishes") % self.title)
         elif self.unstable_karma and self.karma <= self.unstable_karma:
-            log.info("Automatically unpushing %s" % self.title)
-            self.obsolete(db)
-            notifications.publish(
-                topic='update.karma.threshold.reach',
-                msg=dict(update=self, status='unstable'))
+            if self.status is UpdateStatus.pending and not self.autokarma:
+                pass
+            else:
+                log.info("Automatically unpushing %s" % self.title)
+                self.obsolete(db)
+                notifications.publish(
+                    topic='update.karma.threshold.reach',
+                    msg=dict(update=self, status='unstable'))
 
     @property
     def builds_json(self):
--- a/bodhi/tests/server/services/test_updates.py
+++ b/bodhi/tests/server/services/test_updates.py
@@ -2191,6 +2191,83 @@ class TestUpdatesService(bodhi.tests.server.functional.base.BaseWSGICase):
 
     @mock.patch(**mock_valid_requirements)
     @mock.patch('bodhi.server.notifications.publish')
+    def test_pending_update_on_stable_karma_reached_autopush_enabled(self, publish, *args):
+        """ Ensure that pending update directly requests for stable if
+        it hits stable karma before reaching testing state """
+        nvr = u'bodhi-2.0.0-2.fc17'
+        args = self.get_update(nvr)
+        args['autokarma'] = True
+        args['stable_karma'] = 2
+        args['unstable_karma'] = -2
+        self.app.post_json('/updates/', args)
+
+        up = self.db.query(Update).filter_by(title=nvr).one()
+        up.status = UpdateStatus.pending
+        self.db.flush()
+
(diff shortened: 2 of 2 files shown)


bodhi/server/models.py
```diff
@@ -2092,10 +2092,10 @@ def check_karma_thresholds(self, db, agent):
             log.debug('%s locked. Ignoring karma thresholds.' % self.title)
             raise LockedUpdateException
         # Return if the status of the update is not in testing
```
bowlofeggs: This comment should be updated.

bodhi/server/models.py
```diff
@@ -2092,10 +2092,10 @@ def check_karma_thresholds(self, db, agent):
...
-        if self.status != UpdateStatus.testing:
+        if self.status not in (UpdateStatus.testing, UpdateStatus.pending):
             return
         # If an update receives negative karma disable autopush
-        if self.autokarma and self._composite_karma[1] != 0:
+        if self.autokarma and self._composite_karma[1] != 0 and self.status is UpdateStatus.testing:
```
bowlofeggs, line 2306: Why would we only want to disable autokarma for testing updates? It seems to me that we should treat the pending updates the same way.

bodhi/server/models.py
```diff
@@ -2115,7 +2115,8 @@ def check_karma_thresholds(self, db, agent):
                 log.info((
                     "%s update has reached the stable karma threshold and can be pushed to "
                     "stable now if the maintainer wishes") % self.title)
-        elif self.unstable_karma and self.karma <= self.unstable_karma:
+        elif self.unstable_karma and self.karma <= self.unstable_karma \
+                and self.status is UpdateStatus.testing:
```
bowlofeggs: Similarly here - why treat the testing update differently than the pending one?

bodhi/tests/server/services/test_updates.py
```diff
@@ -1866,6 +1866,84 @@ def test_edit_locked_update(self, publish, *args):
...
+
+        self.assertEquals(up.karma, 2)
+        self.assertEquals(up.status, UpdateStatus.pending)
+
+        id = 'bodhi-2.0.0-2.fc17'
+        resp = self.app.get('/updates/%s' % id,
+                            headers={'Accept': 'text/html'})
+        self.assertIn('text/html', resp.headers['Content-Type'])
+        self.assertIn(id, resp)
+        self.assertNotIn('Push to Stable', resp)
```
bowlofeggs: Wouldn't we want ```self.assertIn('Push to Stable', resp)``` here instead?

bowlofeggs (changes requested):
There seem to be 5 failing tests. Here's one of them:

```
FAIL: test_commenting_twice_with_positive_then_negative_karma (bodhi.tests.server.functional.test_comments.TestCommentsService)
  File "/home/vagrant/bodhi/bodhi/tests/server/functional/test_comments.py", line 193, in test_commenting_twice_with_positive_then_negative_karma
    self.assertEquals(res.json_body['comment']['text'], 'Test')
AssertionError: u'Bodhi is disabling automatic push to stable due to negative karma. The maintainer may push manually if they determine that the issue is not severe.' != 'Test'
```

bowlofeggs:
ok test

bodhi/server/models.py
trishnaguha: @bowlofeggs We would want to disable autokarma **only** for testing updates as this is conflicting with https://github.com/fedora-infra/bodhi/blob/develop/bodhi/server/models.py#L2380 for pending updates.
As we would not like to disable autopush for pending updates on unstable karma, rather we want to send it to `obsolete` state https://github.com/fedora-infra/bodhi/blob/develop/bodhi/tests/server/services/test_updates.py#L2397

bodhi/server/models.py
```diff
@@ -2323,11 +2323,14 @@ def check_karma_thresholds(self, db, agent):
                     "%s update has reached the stable karma threshold and can be pushed to "
                     "stable now if the maintainer wishes") % self.title)
         elif self.unstable_karma and self.karma <= self.unstable_karma:
-            log.info("Automatically unpushing %s" % self.title)
-            self.obsolete(db)
-            notifications.publish(
-                topic='update.karma.threshold.reach',
-                msg=dict(update=self, status='unstable'))
+            if self.status is UpdateStatus.pending and not self.autokarma:
+                pass
```
trishnaguha, line 2327: For pending updates with unstable karma (but with disabled autopush), we would not like to send it to `obsolete` state automatically https://github.com/fedora-infra/bodhi/blob/develop/bodhi/tests/server/services/test_updates.py#L2423

bowlofeggs:
test

bowlofeggs (approved):

bowlofeggs:
I opened https://github.com/fedora-infra/bodhi/pull/1651 to add one more assertion to one of these tests.

PR #1560, opened 2017-05-30
Use bleach instead of python-markdown's Safe Mode

ryanlerch:
Previously, we used the Safe Mode of python-markdown to
replace HTML tags in the markdown input from the user. It
replaced all text with a Raw HTML not allowed message.

However, the Safe Mode feature in python-markdown was deprecated,
and they recomended that Bleach be used instead.

This commit passes the markdown input from the user through
Bleach -- before it renders the markdown to html -- escaping all
HTML tags.

### DIFF
--- a/bodhi/server/util.py
+++ b/bodhi/server/util.py
@@ -33,6 +33,7 @@ from kitchen.iterutils import iterate
 from pyramid.i18n import TranslationStringFactory
 from pyramid.settings import asbool
 import arrow
+import bleach
 import colander
 import libravatar
 import markdown
@@ -297,9 +298,23 @@ def hostname(context=None):
 
 
 def markup(context, text):
-    return markdown.markdown(text, safe_mode="replace",
-                             html_replacement_text="--RAW HTML NOT ALLOWED--",
-                             extensions=['markdown.extensions.fenced_code'])
+    """
+    Return HTML from a markdown string
+
+    Args:
+        context (mako.runtime.Context)
+        text (basestring): markdown text to be converted to HTML
+
+    Return:
+        basestring: HTML representation of the markdown text
+    """
+
+    # previously, we used the Safe Mode in python-markdown to strip all HTML
+    # tags. Safe Mode is deprecated, so we now use Bleach to sanitize all HTML
+    # tags before running it through the markdown parser
+    sanitized_text = bleach.clean(text, tags=[])
+
+    return markdown.markdown(sanitized_text, extensions=['markdown.extensions.fenced_code'])
 
 
 def status2html(context, status):
--- a/bodhi/tests/server/functional/test_generic.py
+++ b/bodhi/tests/server/functional/test_generic.py
@@ -198,7 +198,7 @@ class TestGenericViews(bodhi.tests.server.functional.base.BaseWSGICase):
         self.assertEquals(
             res.json_body['html'],
             "<div class='markdown'>"
-            '<p>--RAW HTML NOT ALLOWED--bold--RAW HTML NOT ALLOWED--</p>'
+            '<p>&lt;b&gt;bold&lt;/b&gt;</p>'
             "</div>"
         )
 
--- a/bodhi/tests/server/test_utils.py
+++ b/bodhi/tests/server/test_utils.py
@@ -52,7 +52,7 @@ class TestUtils(object):
         html = util.markup(None, text)
         assert html == (
             "<div class='markdown'>"
-            '<p>--RAW HTML NOT ALLOWED--bold--RAW HTML NOT ALLOWED--</p>'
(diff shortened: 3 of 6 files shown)


bodhi/server/util.py
```diff
@@ -297,9 +298,23 @@ def hostname(context=None):
...
 
 def markup(context, text):
-    return markdown.markdown(text, safe_mode="replace",
-                             html_replacement_text="--RAW HTML NOT ALLOWED--",
-                             extensions=['markdown.extensions.fenced_code'])
+    """
+    Return HTML from a markdown string
+
+    Args:
+        context:
```
ryanlerch: Did not know how to describe `context` here. Any thoughts?

Do we even need to pass context to this method?

bodhi/server/util.py
```diff
@@ -297,9 +298,23 @@ def hostname(context=None):
...
+        text (basestring): markdown text to be converted to HTML
+
+    Return:
+        basestring: HTML representation of the markdown text
+    """
+
+    # previously, we used the Safe Mode in python-markdown to strip all HTML
+    # tags. Safe Mode is deprecated, so we now use Bleach to sanitize all HTML
+    # tags before running it through the markdown parser
+    sanitized_text = bleach.clean(text, tags=[])
```
pypingou, line 315: Note that bleach broke it's API in 2.0, so in pagure we've had to do some tricks to work pre and pose 2.0: https://pagure.io/pagure/pull-request/2269

bodhi/server/util.py
pypingou: I believe passing it is required, but we could s/context/_/ in the function definition

pypingou (commented):
Couple of comments but overall 👍 for me

bodhi/server/util.py
ryanlerch: I wonder what version of bleach is default in prod where we run bodhi... Hopefully we can avoid this dance...

bodhi/server/util.py
ryanlerch: ah! forgot that this method was used in the templates too, where context is passed automatically!

do you think we should just leave it as context, since all the others use it?

bodhi/server/util.py
ralphbean: Looks like python-bleach-1.4.1-2.el7 is in epel7, which is what bodhi runs on.  (Confirmed, that's exactly the version available to the `bodhi03` VM).

(Someone could always put the latest bleach > 2.0 in the epel7-infra tag).

bodhi/server/util.py
pypingou: +1 to leave it

bodhi/server/util.py
ryanlerch: python-bleach is still version 1.x in Fedora too:

https://apps.fedoraproject.org/packages/python-bleach

are we able to just add `bleach<2` in the requirements.txt to solve this?

bowlofeggs: Since bodhi itself is in EPEL, we shouldn't require a newer bleach than is available in EPEL. However, Bodhi is also in Rawhide. Rawhide currently has bleach 1.5, but I would assume it could get updated at any time.

bodhi/server/util.py
bowlofeggs: ```context``` is a ```mako.runtime.Context```. I just stuck a log statement here to find out the type, like this:

```
log.info('type(context): %s' % type(context))
```

bodhi/tests/server/functional/test_generic.py
```diff
@@ -195,10 +195,11 @@ def test_markdown(self):
 
     def test_markdown_with_html(self):
         res = self.app.get('/markdown', {'text': '<b>bold</b>'}, status=200)
+        print res
```

bowlofeggs (changes requested):
A couple more things:

* Let's put the type on the context (see my earlier comment).
* Let's put a note in the release_notes.rst file that bleach is a new dependency.

bowlofeggs:
On Tue, 2017-05-30 at 20:02 -0700, Ryan Lerch wrote:
> are we able to just add bleach<2 in the requirements.txt to solve
> this?

Sure, that works for me. We can burn the bleach == 2 bridge when we get
to it.

ryanlerch:
* added a note in the release notes
* updated requirements.txt to require bleach<2
* added the type to the context in the docblock
* removed one of my now infamous print statements i like to leave lying around.

pypingou:
Looks fine to me

bowlofeggs (approved):
@ryanlerch you are a hero

bodhi/server/util.py
ryanlerch: Just FYI, we didnt avoid this dance for long... Rawhide bleach is now updated to 2. see #1718

PR #1562, opened 2017-05-30
Use specific package subclass in query.

ralphbean:
This fixes a MultipleResultsFound error I got from sqlalchemy when trying to
create my demo of #1543:  https://fedorapeople.org/groups/factory2/sprint-031/threebean-modular-bodhi-updates.ogv

The "package" in question here was "nginx"... but do I want the nginx
RpmPackage or the nginx ModulePackage?  Specificity required!

### DIFF
--- a/bodhi/server/services/updates.py
+++ b/bodhi/server/services/updates.py
@@ -424,13 +424,16 @@ def new_update(request):
         # Create the Package and Build entities
         for nvr in data['builds']:
             name, version, release = request.buildinfo[nvr]['nvr']
-            # The package will have been created by validate_acls
-            package = request.db.query(Package).filter_by(name=name).one()
 
-            # Figure out what kind of build this should be.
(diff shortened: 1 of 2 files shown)


mcurlej:
LGFM

bowlofeggs (changes requested):
It would be good to write a test for this change. Perhaps a test that submits a new update here, and we can mock the Koji data to make it be a module and assert that the right Package subclass was used?

bowlofeggs:
@ralphbean Do you still want to work on this?

ryanlerch:
jenkies test

ralphbean:
I forgot all about this change!  I'll try to revisit it in the next few days.

ralphbean:
Rebased on `develop`.

> It would be good to write a test for this change.

Added test in 6b51ce1.

bowlofeggs (approved):
Thanks!

PR #1565, opened 2017-05-31
Add diff-cover to setup.py and add documentation on how to use it

mprahl:
I'm proposing we add `diff-cover` to Bodhi's development environment and contribution documentation so that developers can easily tell if their code changes are covered by tests.

We can consider putting this as part of `btest`, however, we would have to always assume that the branch to compare it to is `origin/develop` and that may not always be the case. I will leave that up to the maintainers of Bodhi to decide.

### DIFF
--- a/devel/ansible/roles/dev/tasks/main.yml
+++ b/devel/ansible/roles/dev/tasks/main.yml
@@ -30,6 +30,7 @@
       - python-cornice
       - python-cornice-sphinx
       - python-devel
+      - python3-diff-cover
       - python-dogpile-cache
       - python-fedora
       - python-ipdb
--- a/setup.py
+++ b/setup.py
@@ -87,7 +87,8 @@ setup(
         'nose',
         'nose-cov',
         'webtest',
-        'mock'
+        'mock',
+        'diff-cover'
     ],
     test_suite="nose.collector")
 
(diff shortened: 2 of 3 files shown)


setup.py
```diff
@@ -87,7 +87,8 @@ def get_requirements(requirements_file='requirements.txt'):
         'nose',
         'nose-cov',
         'webtest',
-        'mock'
+        'mock',
+        'diff-cover'
```
bowlofeggs, line 91: In addition to this, let's also add ```python2-diff-cover``` [here](https://github.com/fedora-infra/bodhi/blob/2.6.2/devel/ansible/roles/dev/tasks/main.yml#L30) so that the RPM gets installed in the Vagrant box automatically.

bowlofeggs (changes requested):

setup.py
mprahl: Just added it. Thanks for the review.

devel/ansible/roles/dev/tasks/main.yml
```diff
@@ -30,6 +30,7 @@
       - python-cornice
       - python-cornice-sphinx
       - python-devel
+      - python2-diff-cover
```
bowlofeggs: I think I lied to you - apparently this package doesn't provide ```/usr/bin/diff-cover``` - only ```python3-diff-cover``` seems to provide that. The tool does seem to work as you documented with the python3 version though.

mprahl: Okay, fixed!

bowlofeggs (approved):
Thanks!

PR #1566, opened 2017-05-31
Move acl parameters out of the method definitions

ryanlerch:
Cornice has deprecated the use of the acl parameter in method
definitions. This caused the following warnings to show when
running the tests:

/usr/lib/python2.7/site-packages/cornice/service.py:295:
DeprecationWarning: Warning: Using acl in method definitions is
deprecated.Use service or resource instead.

This commit moves the acls into the Service calls.

Related: #1522


mcurlej:
LGFM

bowlofeggs (approved):
Thanks!

PR #1567, opened 2017-05-31
Add tests for server/renderers.py

ryanlerch:
Adds a new test for server/renderers.py, testing the JPEG
renderer.

### DIFF
--- a/bodhi/tests/server/test_renderers.py
+++ b/bodhi/tests/server/test_renderers.py
@@ -1,0 +1,61 @@
+# Copyright 2017 Red Hat, Inc.
+#
+# This file is part of Bodhi.
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+
+import copy
+import re
+import StringIO
+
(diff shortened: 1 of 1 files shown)


bodhi/tests/server/test_renderers.py
```diff
@@ -0,0 +1,54 @@
...
+            'captcha.background_color': '#ffffff',
+            'captcha.padding': '5',
+            'captcha.ttl': '300',
+        })
+        app = TestApp(main({}, session=self.db, **settings))
+
+        res = app.get('/updates/FEDORA-2017-a3bbe1a8f2', headers=dict(accept='text/html'))
+        captcha_url = re.search(r'"http://localhost(/captcha/[^"]*)"', str(res)).groups()[0]
+        resp = app.get(captcha_url)
+        self.assertIn('image/jpeg', resp.headers['Content-Type'])
```
bowlofeggs: We could probably use Pillow to assert that the image is a valid JPEG. Perhaps we could use [```pillow```](https://pillow.readthedocs.io/en/4.1.x/) to load the image and assert that it's is 300x80 pixels?

I'd say this is optional if you don't feel like doing it right now - we can always do it later.

bodhi/tests/server/test_renderers.py
```diff
@@ -0,0 +1,54 @@
...
+            'captcha.font_color': '#000000',
+            'captcha.background_color': '#ffffff',
+            'captcha.padding': '5',
+            'captcha.ttl': '300',
+        })
+        app = TestApp(main({}, session=self.db, **settings))
+
+        res = app.get('/updates/FEDORA-2017-a3bbe1a8f2', headers=dict(accept='text/html'))
+        captcha_url = re.search(r'"http://localhost(/captcha/[^"]*)"', str(res)).groups()[0]
+        resp = app.get(captcha_url)
```
bowlofeggs: On both calls to ```app.get()```, you can put a ```status=200``` to also assert that there aren't error codes.

bowlofeggs (approved):
Let's put the status=200 calls on there, and I'll leave it to you if you want to try to load the jpeg with pillow or not.

ryanlerch:
Thanks for the review, @bowlofeggs !

I have added the status checks, and also checking the imagesize using PIL.

merging now.

cheers,
ryanlerch

PR #1568, opened 2017-05-31
Improve test coverage for util.py

ryanlerch:
* Adds tests for getting RPM headers from koji
* Adds tests for the splitter() method in util.py
* Remove unused Singleton class
* Remove unused get_repo_tag() method

### DIFF
--- a/bodhi/tests/server/test_utils.py
+++ b/bodhi/tests/server/test_utils.py
@@ -57,8 +57,22 @@ class TestUtils(object):
         ), html
 
     def test_rpm_header(self):
-        h = util.get_rpm_header('')
+        h = util.get_rpm_header('libseccomp')
         assert h['name'] == 'libseccomp', h
+
+    def test_rpm_header_exception(self):
+        try:
+            util.get_rpm_header('raise-exception')
+            assert False
+        except Exception:
+            pass
+
+    def test_rpm_header_not_found(self):
+        try:
+            util.get_rpm_header("do-not-find-anything")
+            assert False
+        except ValueError:
+            pass
 
     def test_cmd_failure(self):
         try:
--- a/bodhi/tests/server/test_utils.py
+++ b/bodhi/tests/server/test_utils.py
@@ -21,7 +21,7 @@ from bodhi.server.buildsys import setup_buildsystem, teardown_buildsystem
 from bodhi.server.config import config
 
 
-class TestUtils(object):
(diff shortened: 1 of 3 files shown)


bodhi/tests/server/test_utils.py
```diff
@@ -88,6 +88,21 @@ def test_sorted_builds(self):
...
+        assert splitlist[0] == "build-0.1"
+        assert splitlist[1] == "build-0.2"
+
+        splitcommastring = util.splitter("build-0.1, build-0.2")
+        print splitcommastring
+        assert splitcommastring[0] == "build-0.1"
+        assert splitcommastring[1] == "build-0.2"
+
+        splitspacestring = util.splitter("build-0.1 build-0.2")
+        print splitspacestring
```
bowlofeggs: Let's drop the two print statements.

bodhi/tests/server/test_utils.py
```diff
@@ -88,6 +88,21 @@ def test_sorted_builds(self):
         assert b1 == new, b1
         assert b2 == old, b2
 
+    def test_splitter(self):
+        splitlist = util.splitter(["build-0.1", "build-0.2"])
+        assert splitlist[0] == "build-0.1"
+        assert splitlist[1] == "build-0.2"
```
bowlofeggs: For all three of these test blocks, you could replace the assertions with a one liner that also asserts that the length is 2:

```
self.assertEqual(splitlist, ['build-0.1', 'build-0.2'])
```

bowlofeggs (changes requested):
```deleting_unused_code++```

bodhi/tests/server/test_utils.py
ryanlerch: Done. I had to make the class in this file extend unittest.TestCase to be able to use assertEqual though.

bodhi/tests/server/test_utils.py
ryanlerch: Done! Sorry about leaving debugging stuff in there :/

bodhi/tests/server/test_utils.py
bowlofeggs: Oh right, I'd forgotten about that. Some day I hope to unify the tests a bit more so they at least all subclass ```unittest.TestCase```.

bowlofeggs (approved):
Awesome, thanks!

PR #1573, opened 2017-06-01
Fix a failing unit test by adjusting its assertion to be correct.

### LINKED ISSUE #1572: Test failure: Test a logged in User without permissions on the update can't see the form
```
FAIL: Test a logged in User without permissions on the update can't see the form
  File "/home/vagrant/bodhi/bodhi/tests/server/services/test_updates.py", line 513, in test_edit_without_permission
    self.assertIn('anonymous does not have commit access to bodhi', resp)
AssertionError: 'anonymous does not have commit access to bodhi' not found in <400 Bad Request text/html body='\n\n\n\n\...l>\n'/9016>
```

bowlofeggs:
fixes #1572


cverna (approved):
👍

PR #1574, opened 2017-06-01
PEP-257 bodhi.server.views.search.

bowlofeggs:

### DIFF
--- a/bodhi/server/views/search.py
+++ b/bodhi/server/views/search.py
@@ -1,3 +1,7 @@
+# Copyright 2014-2017 Red Hat, Inc. and others
+#
+# This file is part of Bodhi.
+#
 # This program is free software; you can redistribute it and/or
 # modify it under the terms of the GNU General Public License
 # as published by the Free Software Foundation; either version 2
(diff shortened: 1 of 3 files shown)


bodhi/server/views/search.py
```diff
@@ -27,7 +37,19 @@ def get_all_packages():
...
+    Deprecated: This view is unused by the web UI and the CLI and might be a good candidate for
+    removal in a future major release.
+
+    Args:
+        request (pyramid.request): The current web request.
+    Returns:
+        list: A list of dictionaries with keys 'id', 'label', and 'value' all indexing koji
+            listPackages() results that match the search term.
+    """
+    log.info('BOWLOFEGGS')
```
ryanlerch: is this log entry still needed? or was this for debugging?

ryanlerch (approved):
Good find! It looks like we get this info for the search widget straight from the Fedora Packages app now.

This feature also has no test coverage either. Should we add some, or just exclude it from the tests now it is deprecated?

bowlofeggs:
On Thu, 2017-06-01 at 14:13 -0700, Ryan Lerch wrote:
> is this log entry still needed? or was this for debugging?

Nope, good catch! I was trying to determine if this was used or not by
the bindings or the web UI (it isn't).

PR #1577, opened 2017-06-02
PEP-257: bodhi.server.views.generic

ryanlerch:
Adds a few docblocks to bodhi.server.views.generic


cverna (approved):
👍 looks good ;)

bowlofeggs (approved):
Very awesome, thanks so much!

PR #1578, opened 2017-06-02
Remove several unused settings from the settings files.

bowlofeggs:


pypingou:
👍 for me

PR #1580, opened 2017-06-02
Add a new bdiff-cover development shell alias.

bowlofeggs:

### DIFF
--- a/devel/ansible/roles/dev/files/.bashrc
+++ b/devel/ansible/roles/dev/files/.bashrc
@@ -9,6 +9,7 @@ fi
 # export SYSTEMD_PAGER=
 
 shopt -s expand_aliases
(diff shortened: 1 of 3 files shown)


devel/ansible/roles/dev/files/.bashrc
```diff
@@ -9,6 +9,7 @@ fi
 # export SYSTEMD_PAGER=
 
 shopt -s expand_aliases
+alias bdiff-cover="btest; diff-cover /home/vagrant/bodhi/coverage.xml --compare-branch=develop --fail-under=100"
```
cverna, line 12: diff-cover needs to be added to the ansible playbook too, so that the command is available

bowlofeggs:
On Fri, 2017-06-02 at 12:16 -0700, Clement Verna wrote:
> diff-cover needs to be added to the ansible playbook too, so that the
> command is available

It's actually already there from an earlier pull request.

jeremycline (approved):
:shipit:

PR #1582, opened 2017-06-02
Use bodhi bindings client

### LINKED ISSUE #1338: bodhi.server.scripts.manage_releases still uses fedora.client.bodhi
```bodhi.server.scripts.manage_releases``` imports ```fedora.client.bodhi```, but it should be using ```bodhi.client.bindings``` instead.

cverna:
Use the bodhi bindings client instead of the one in python-fedora for the manage_releases.py script

Fixes #1338

### DIFF
--- a/bodhi/server/scripts/manage_releases.py
+++ b/bodhi/server/scripts/manage_releases.py
@@ -27,7 +27,7 @@ import munch
 
 import click
 
-from fedora.client.bodhi import Bodhi2Client
+from bodhi.client.bindings import BodhiClient
 
 
 @click.group()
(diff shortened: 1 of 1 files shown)


ryanlerch:
Just a note here too, AFAICT, we don't have any test coverage at all on the manage_releases.py script. 

I do like that this will remove another deprecation warning that shows up during the tests :D

bowlofeggs:
I am adding the ```--url``` flag to the manage_releases script, and in doing so I also need this patch and I need to write tests, so don't worry about the tests. I'll write the tests for you, and I'll submit a new PR that has this commit, with a commit from me on top that adds the flag and tests. Once that's open, we'll close this one and just have a dual-authored PR. Sound good?

cverna:
Sounds Good ;)

bowlofeggs:
I've opened https://github.com/fedora-infra/bodhi/pull/1586 with the commit from this pull request. Thanks!

bowlofeggs:
This has been merged as https://github.com/fedora-infra/bodhi/commit/98fb4f394de5c1550cccba466425773c1bacd123.

PR #1583, opened 2017-06-03
Fix the logic of --autokarma option.

### LINKED ISSUE #1378: Creating an update with --autokarma and without the --stable and --unstable-karma flags sets autokarma off
Today I created an update with the CLI like this:

```
$ bodhi updates new --user bowlofeggs --type bugfix --notes "The bodhi-client package no longer depends on yum." --bugs 1135681 --close-bugs --autokarma bodhi-2.4.0-3.fc26
Password:
================================================================================
     bodhi-2.4.0-3.fc26
================================================================================
  Update ID: FEDORA-2017-e2e877073e
    Release: Fedora 26
     Status: pending
       Type: bugfix
      Karma: 0
  Autokarma: False  [-3, 3]
    Request: testing
       Bugs: 1135681 - None
      Notes: The bodhi-client package no longer depends on yum.
  Submitter: bowlofeggs
  Submitted: 2017-03-22 22:30:16
   Comments: bodhi - 2017-03-22 22:30:16 (karma 0)
             This update has been submitted for testing by
             bowlofeggs.

  https://bodhi.fedoraproject.org/updates/FEDORA-2017-e2e877073e
```

Note that the Autokarma is False in the resulting text, and if I go to that update Autokarma is displayed as being off there as well. I would expect that the ```--autokarma``` flag would set it to be on.

#1298 may be related to this (and may even be the same underlying problem) - it reports that setting ```--stable-karma``` without specifying ```--autokarma``` sets autokarma to be on. Given this, my guess is that something is using the karma thresholds to determine whether to enable autokarma instead of the ```--autokarma``` flag.

cverna:
Fix the logic of --autokarma option. autokarma is enabled by default.
Fixes #1378

### DIFF
--- a/bodhi/client/__init__.py
+++ b/bodhi/client/__init__.py
@@ -60,7 +60,7 @@ new_edit_options = [
     click.option('--close-bugs', default=True, is_flag=True, help='Automatically close bugs'),
     click.option('--request', help='Requested repository',
                  type=click.Choice(['testing', 'stable', 'unpush'])),
-    click.option('--autokarma', default=True, is_flag=True, help='Enable karma automatism'),
+    click.option('--autokarma', is_flag=True, help='Enable karma automatism'),
     click.option('--stable-karma', type=click.INT, help='Stable karma threshold'),
     click.option('--unstable-karma', type=click.INT, help='Unstable karma threshold'),
     click.option('--suggest', help='Post-update user suggestion',
--- a/bodhi/tests/client/test___init__.py
+++ b/bodhi/tests/client/test___init__.py
@@ -109,7 +109,7 @@ class TestNew(unittest.TestCase):
 
         result = runner.invoke(
             client.new,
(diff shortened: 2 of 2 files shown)


bodhi/client/__init__.py
```diff
@@ -60,7 +60,7 @@ def _warn_if_url_and_staging_set(ctx, param, value):
     click.option('--close-bugs', default=True, is_flag=True, help='Automatically close bugs'),
     click.option('--request', help='Requested repository',
                  type=click.Choice(['testing', 'stable', 'unpush'])),
-    click.option('--autokarma', default=True, is_flag=True, help='Enable karma automatism'),
+    click.option('--autokarma', is_flag=True, help='Enable karma automatism'),
```
bowlofeggs, line 63: Oh man, I didn't realize the problem was this simple. Nice find!

bowlofeggs (approved):
Looks good and seems to work. Thanks!

PR #1581, opened 2017-06-02
Add support for content_type to the Python bindings.

bowlofeggs:
fixes #1327

Signed-off-by: Randy Barlow <randy@electronsweatshop.com>

PR #1587, opened 2017-06-06
Add a content_type query switch to the CLI.

### LINKED ISSUE #1328: Modify the CLI to work with multiple content types
We will need to modify the CLI to handle multiple content types, once #1327 is done.

ralphbean:
Fixes #1328 (and complements #1581).

### DIFF
--- a/bodhi/client/__init__.py
+++ b/bodhi/client/__init__.py
@@ -236,6 +236,8 @@ def edit(user, password, url, **kwargs):
               help='Query only critical path packages')
 @click.option('--cves', help='Query by CVE id')
 @click.option('--packages', help='Query by package name(s)')
+@click.option('--content-type', help='Query updates based on content type',
+              type=click.Choice(['rpm', 'module']))  # And someday, container.
 @click.option('--pushed', is_flag=True, default=None,
               help='Filter by pushed updates')
 @click.option('--pushed-since',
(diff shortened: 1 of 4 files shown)


bodhi/client/__init__.py
```diff
@@ -236,6 +236,8 @@ def edit(user, password, url, **kwargs):
               help='Query only critical path packages')
 @click.option('--cves', help='Query by CVE id')
 @click.option('--packages', help='Query by package name(s)')
+@click.option('--content-type', help='Query updates based on content type',
+              type=click.Choice(['rpm', 'module']))  # And someday, container.
```
pypingou, line 240: Should it default to rpm?

pypingou:
One question, otherwise 👍 for me

bodhi/client/__init__.py
ralphbean: I just don't know, but am open to opinions.

By defaulting to `None`, a query for `--packages=nginx` would return all updates with the nginx package and those with the nginx module.

pypingou: Sounds also sane, let's see what @bowlofeggs prefers

ralphbean: Hm, a thought.  If the default is anything other than `None` (e.g., if it is 'rpm'), then there is no way to query for *all* updates of *any* content type.

bowlofeggs: Yeah I think I like letting it default to showing all content types.

bowlofeggs (approved):
Seems to work!

PR #1591, opened 2017-06-07
PEP-257 bodhi.server.views.__init__.

bowlofeggs:


pypingou:
👍 as well

pypingou:
Tests are passing, let's merge :)

Issue #1441, opened 2017-04-14
Sort the builds on the update page alphabetically


bowlofeggs:
It would be nice if the list of builds on the update page's Builds tab were sorted alphabetically.

PR #1592, opened 2017-06-09
Sort builds in the update by the NVR of the build

ryanlerch:
Adds sorting to the model for updates, sorting the builds by the NVR.

Note that this also changes the output of the titles that we generate for the update, as well as the order in the builds tab.

Before:
before

After:
after

Fixes: #1441


bowlofeggs (approved):
Thanks!

Issue #946, opened 2016-09-21
Add test coverage to ensure the override.html template is covered


bowlofeggs:
Due to lack of test coverage, we had a problem in production where the override.html template could not be rendered due to an incorrect import. If we had test coverage on rendering that template, we would have caught that problem before it reached production (or even develop!)

https://github.com/fedora-infra/bodhi/pull/944/ fixed the issue.

Issue #1594, opened 2017-06-13
Any logged in user can create or expire a buildroot override for any package


ryanlerch:
Currently, any logged in user is able to create a buildroot override for any package. AFAICT, we don't check any ACLs from packageDB when doing buildroot overrides. Note too that this also allows any logged in user to expire a buildroot override that someone else has created.

The only reason why the non-logged in user couldnt create a buildroot override was due to the exception raised in #1541

bowlofeggs:
I do believe that due to your recent fix, Bodhi does now enforce that the user be a packager (i.e., not just logged in but also a member of a packager group).

I think the permissions might be intentionally this way. A lot of packagers depend on packages that they don't have ACLs on, and they may need to build their packages against a version of a dependency that isn't yet stable. Allowing them to create a BuildrootOverride for those packages gives them a "self service" option, rather than having to ask the package maintainer to do it for them. It's probably a bit of an unfortunate side effect that packagers can expire other user's overrides, but I haven't seen people doing that maliciously so far. We *could* do something special to disallow that, but I would probably recommend waiting until it's a real problem before doing that.

PR #1595, opened 2017-06-13
not show the override form if not logged in

### LINKED ISSUE #1541: clicking on expire buildroot override when not logged in gives poor messaging
if you try to expire a buildroot override when not logged in you end up with a message
"Override : Unable to save buildroot override: 'NoneType' object has no attribute 'name'" which is really confusing and gives no hint that its just that you are not logged in.  instread you should be redirected to login, or at the least be given a message that you are not logged in and to do so first

ryanlerch:
Previously, the full edit override form was shown when
looking at an override, even if the user was not logged
in. This commit changes the override template so the
form is not shown to the non-logged in user. Instead
they get just the info about the override.

Fixes: #1541

This also adds some tests to test the above funtionality,
as well as a few extra tests on the overrides views in
html where we didnt test in the past, fixing #946

### DIFF
--- a/bodhi/server/templates/override.html
+++ b/bodhi/server/templates/override.html
@@ -1,6 +1,7 @@
 <%namespace name="util" module="bodhi.server.util"/>
 <%inherit file="master.html"/>
 <div class="container">
+%if request.user:
 <div class="row">
   <div id='js-warning' class="col-md-6 col-md-offset-3 m-t-3">
     <div class="alert alert-warning">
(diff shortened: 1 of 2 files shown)


ryanlerch:
Note that this only addresses the non-logged in user, as described in issue #1541 

I was not sure if the ability for any logged in user to create or expire an override was an oversight or a features, so i left it alone for this PR, and opened Issue #1594

bowlofeggs:
test

bodhi/server/templates/override.html
```diff
@@ -20,14 +21,26 @@
   </div>
 
   <script>$("#js-warning").addClass('hidden');</script>
-
-  <div id="new-override-form" class="col-md-12 hidden m-t-3">
+%endif
+  %if override is UNDEFINED:
+  <h2 class="pull-left m-t-3">New Override</h2>
```
bowlofeggs, line 26: When would the override be undefined?

bowlofeggs (approved):
Just one question, but not a merge blocker (so merge as you please).

bodhi/server/templates/override.html
ryanlerch: this page is also shown for creating a new override. The override is undefined in this case, and this is how we check in the template if we are looking at the new override form or not.

PR #1596, opened 2017-06-13
Add tests for bodhi.client.bindings.parse_file()

ryanlerch:


bowlofeggs:
test

bodhi/tests/client/test_bindings.py
```diff
@@ -987,3 +987,71 @@ def test___unicode__(self):
...
+             "\n"
+             "# Automatically close bugs when this marked as stable\n"
+             "close_bugs=True\n"
+             "\n"
+             "# Suggest that users restart after update\n"
+             "suggest_reboot=False")
+
+        update_template = open("/tmp/bodhi-test-parsefile1", "w")
+        update_template.write(s)
+        update_template.close()
```
bowlofeggs: You could do this without writing a real file to the disk by mocking the builtin ```open()``` function. Here's an example where I did that to simulate opening an empty file:

https://github.com/fedora-infra/bodhi/blob/2.7.0/bodhi/tests/client/test_bindings.py#L98-L118

Some of the other tests in that class substitute other data in the file. This is a nice way to do it because you don't have to worry about cleaning up the file if the test fails, and it performs a little better too!

bodhi/tests/client/test_bindings.py
```diff
@@ -987,3 +987,71 @@ def test___unicode__(self):
...
+        self.assertEqual(updates[0]['severity'], 'unspecified')
+        self.assertEqual(updates[0]['stable_karma'], '3')
+        self.assertEqual(updates[0]['builds'], 'fedora-workstation-backgrounds-1.1-1.fc26')
+        self.assertEqual(updates[0]['autokarma'], 'True')
+        self.assertEqual(updates[0]['suggest'], 'unspecified')
+        self.assertEqual(updates[0]['notes'], 'Initial Release')
+        self.assertEqual(updates[0]['request'], 'testing')
+        self.assertEqual(updates[0]['bugs'], '123456,43212')
+        self.assertEqual(updates[0]['type_'], 'bugfix')
+        self.assertEqual(updates[0]['type'], 'bugfix')
```
bowlofeggs, line 1047: Nice!

bowlofeggs (approved):
It's fine as-is if you prefer this, but I suggest using mock to avoid writing/reading a real file during the test.

bowlofeggs:
Hey @ryanlerch!

I spent some time looking into the paste you sent me today about trying to use mock and hitting the ```StopIteration```. I sat down and did some quality time today, and it turns out that you found a bug in Mock! Here's the upstream and downstream bugs about it:

https://github.com/testing-cabal/mock/issues/382
http://bugs.python.org/issue26807

bowlofeggs:
@ryanlerch I've been able to get your test to work by mocking ```readline()``` by hand:

```
class TestBodhiClient_parse_file(unittest.TestCase):
    """
    Test the BodhiClient.parse_file() method.
    """
    @mock.patch('__builtin__.open', create=True)
    @mock.patch('bodhi.client.bindings.BodhiClient._load_cookies')
    @mock.patch('bodhi.client.bindings.os.path.exists')
    def test_parsing_valid_file(self, exists, _load_cookies, mock_open):
        """
        Test parsing a valid update template file
        """
        s = [
            "[fedora-workstation-backgrounds-1.1-1.fc26]\n",
            "# bugfix, security, enhancement, newpackage (required)\n",
            "type=bugfix\n",
            "\n",
            "# testing, stable\n",
            "request=testing\n",
            "\n",
            "# Bug numbers: 1234,9876\n",
            "bugs=123456,43212\n",
            "\n",
            "# Here is where you give an explanation of your update.\n",
            "notes=Initial Release\n",
            "\n",
            "# Enable request automation based on the stable/unstable karma thresholds\n",
            "autokarma=True\n",
            "stable_karma=3\n",
            "unstable_karma=-3\n",
            "\n",
            "# Automatically close bugs when this marked as stable\n",
            "close_bugs=True\n",
            "\n",
            "# Suggest that users restart after update\n",
            "suggest_reboot=False\n",
            ""]

        exists.return_value = True
        mock_open.return_value.readline.side_effect = s * 1000
        client = bindings.BodhiClient()
        updates = client.parse_file("sad")
        self.assertEqual(len(updates), 1)
        self.assertEqual(len(updates[0]), 12)
        self.assertEqual(updates[0]['close_bugs'], True)
        self.assertEqual(updates[0]['unstable_karma'], '-3')
        self.assertEqual(updates[0]['severity'], 'unspecified')
        self.assertEqual(updates[0]['stable_karma'], '3')
        self.assertEqual(updates[0]['builds'], 'fedora-workstation-backgrounds-1.1-1.fc26')
        self.assertEqual(updates[0]['autokarma'], 'True')
        self.assertEqual(updates[0]['suggest'], 'unspecified')
        self.assertEqual(updates[0]['notes'], 'Initial Release')
        self.assertEqual(updates[0]['request'], 'testing')
        self.assertEqual(updates[0]['bugs'], '123456,43212')
        self.assertEqual(updates[0]['type_'], 'bugfix')
        self.assertEqual(updates[0]['type'], 'bugfix')

    def test_parsing_nonexistant_file(self):
        """
        Test trying to parse a file that doesnt exist
        """
        client = bindings.BodhiClient()

        with self.assertRaises(ValueError) as exc:
            client.parse_file("/tmp/bodhi-test-parsefile2")

        self.assertEqual(unicode(exc.exception),
                         'No such file or directory: /tmp/bodhi-test-parsefile2')
```

ryanlerch:
🎉 🎉 🎉 

Thanks *SO MUCH* @bowlofeggs for taking the time to figuring that out for me!

I have updated the commit with your changes!

bodhi/tests/client/test_bindings.py
```diff
@@ -987,3 +987,72 @@ def test___unicode__(self):
...
+            "\n",
+            "# Automatically close bugs when this marked as stable\n",
+            "close_bugs=True\n",
+            "\n",
+            "# Suggest that users restart after update\n",
+            "suggest_reboot=False\n",
+            ""]
+
+        exists.return_value = True
+        mock_open.return_value.readline.side_effect = s * 1000
```
bowlofeggs: Ooops - this line was here from some debugging I was doing when I was trying to figure out why Mock wasn't working. Can you replace this line with ```mock_open.return_value.readline.side_effect = s``` before merging? Thanks!

bowlofeggs (approved):
Please make the one edit I noted below before merging, but no need for further review. Thanks for the patch!

PR #1605, opened 2017-06-13
Remove an unneeded conditional import.

### LINKED ISSUE #1166: bodhi.server.util conditionally imports rpm
I don't see a reason for ```bodhi.server.util``` to conditionally import rpm, as opposed to hard depending on it. We should investigate why it is this way, and make it a hard dependency if it makes sense. If this change is made, release notes will need to be written as well.

bowlofeggs:
Other parts of bodhi already hard-depend on rpm being available
(such as bodhi.server.models), so there's no reason to
conditionally import it.

fixes #1166

### DIFF
--- a/bodhi/server/util.py
+++ b/bodhi/server/util.py
@@ -36,15 +36,11 @@ import colander
 import libravatar
 import markdown
 import requests
+import rpm
 
 from bodhi.server import log, buildsys, Session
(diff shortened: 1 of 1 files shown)


ryanlerch (approved):
Awesome thanks!

This import wasnt tested either, so another couple lines towards 100% test coverage! 😄

PR #1606, opened 2017-06-15
Clean up the --help output in bodhi-client

### LINKED ISSUE #1457: The CLI docblocks end up in the --help text and are noisy

ryanlerch:
python-click, the library we use to create the CLI for bodhi-client
takes the first docblock of a method and uses it to help create the
--help docs. Previously, these docblocks contained developer docs
that we of little use to the end user. This commit splits these
docblocks into two seperate ones, the first for user docs, (which is
nommed and used by click), and the second for developer docs.

The user docs are also expanded to include documentation on the
arguments a command requires, which was not done in the past.
click does not have a way to document the arguments of a command, like
it does for options, and they suggest that it just be done in the
main docblock for a command.

fixes #1457 

For quick reference too, here is the output of --help for all commands before:
[help-before.txt](https://github.com/fedora-infra/bodhi/files/1076339/help-before.txt)

And after:
[help-after.txt](https://github.com/fedora-infra/bodhi/files/1076340/help-after.txt)

### DIFF
--- a/bodhi/client/__init__.py
+++ b/bodhi/client/__init__.py
@@ -125,14 +125,26 @@ def _save_override(url, user, password, staging, edit=False, **kwargs):
 @click.group()
 @click.version_option(message='%(version)s')
 def cli():
+    # Docs that show in the --help
     """
-    Create the main CLI group.
+    Command line tool for interacting with Bodhi
+    """
+
+    # Developer Docs
+    """
+    Create the main CLI group
     """
     pass
 
 
 @cli.group()
 def updates():
+    # Docs that show in the --help
+    """
+    Interact with updates on Bodhi.
+    """
+
+    # Developer Docs
     """
     Create the updates group.
     """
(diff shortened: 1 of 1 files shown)


bowlofeggs (approved):
Thanks!

PR #1608, opened 2017-06-15
Provide messenging if user tries to create a new override that already exists

### LINKED ISSUE #1377: The CLI should not traceback when trying to save an already active override
I tried to use the CLI to alter the duration on an override I made today, and it crashes:

```
$ bodhi overrides save --duration 3 --notes "This is needed to build Bodhi." --user bowlofeggs python-pyramid-1.8.2-2.fc26
Password:
  File "/usr/bin/bodhi", line 11, in <module>
    load_entry_point('bodhi-client==2.4.0', 'console_scripts', 'bodhi')()
  File "/usr/lib/python2.7/site-packages/click/core.py", line 722, in __call__
  File "/usr/lib/python2.7/site-packages/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/usr/lib/python2.7/site-packages/click/core.py", line 1066, in invoke
  File "/usr/lib/python2.7/site-packages/click/core.py", line 1066, in invoke
  File "/usr/lib/python2.7/site-packages/click/core.py", line 895, in invoke
  File "/usr/lib/python2.7/site-packages/click/core.py", line 535, in invoke
  File "/usr/lib/python2.7/site-packages/bodhi/client/__init__.py", line 229, in save_buildroot_overrides
    resp = client.save_override(nvr=nvr, duration=duration, notes=notes)
  File "/usr/lib/python2.7/site-packages/bodhi/client/bindings.py", line 78, in wrapper
    raise BodhiClientException(problems)
bodhi.client.bindings.BodhiClientException: Buildroot override for python-pyramid-1.8.2-2.fc26 already exists
```

It should be possible to use the CLI to edit active overrides, just like the web UI can do.

ryanlerch:
This commit provides better error messages to the user if they try
to create an override with bodhi-client that already exists.

fixes #1377

### DIFF
--- a/bodhi/client/__init__.py
+++ b/bodhi/client/__init__.py
@@ -537,7 +537,16 @@ def save_buildroot_overrides(user, password, url, staging, **kwargs):
         staging (bool): Whether to use the staging server or not.
         kwargs (dict): Other keyword arguments passed to us by click.
     """
-    _save_override(url=url, user=user, password=password, staging=staging, **kwargs)
+
+    try:
+        _save_override(url=url, user=user, password=password, staging=staging, **kwargs)
+    except bindings.BodhiClientException as e:
+        if str(e) == "Buildroot override for %s already exists" % (kwargs['nvr']):
+            click.echo(str(e))
+            click.echo("The `overrides save` command is used for creating a new override.")
+            click.echo("Use `overrides edit` to edit an existing override.")
+        else:
+            raise
 
 
 @overrides.command('edit')
--- a/bodhi/tests/client/test___init__.py
+++ b/bodhi/tests/client/test___init__.py
@@ -310,6 +310,44 @@ class TestSaveBuilrootOverrides(unittest.TestCase):
                 'csrf_token': 'a_csrf_token'})
         self.assertEqual(bindings_client.base_url, 'http://localhost:6543/')
 
+    @mock.patch('bodhi.client.bindings.BodhiClient.csrf',
+                mock.MagicMock(return_value='a_csrf_token'))
+    @mock.patch('bodhi.client.bindings.BodhiClient.send_request', autospec=True)
+    def test_existing_override_error_message(self, send_request):
+        """
+        Assert that the error message is provided if we try to save an existing override
+        """
+        exception_message = "Buildroot override for js-tag-it-2.0-1.fc25 already exists"
+        send_request.side_effect = bindings.BodhiClientException(exception_message)
+        runner = testing.CliRunner()
+
+        result = runner.invoke(
+            client.save_buildroot_overrides,
+            ['--user', 'bowlofeggs', '--password', 's3kr3t', 'js-tag-it-2.0-1.fc25'])
+
+        self.assertEqual(result.exit_code, 0)
+        self.assertIn("The `overrides save` command is used for creating a new override",
+                      result.output)
+        self.assertIn("Use `overrides edit` to edit an existing override",
+                      result.output)
+
+    @mock.patch('bodhi.client.bindings.BodhiClient.csrf',
+                mock.MagicMock(return_value='a_csrf_token'))
+    @mock.patch('bodhi.client.bindings.BodhiClient.send_request', autospec=True)
+    def test_other_client_exception(self, send_request):
+        """
+        Assert that any other BodhiClientExceptions are raised as expected
+        """
+        send_request.side_effect = bindings.BodhiClientException("Pants Exception")
+        runner = testing.CliRunner()
+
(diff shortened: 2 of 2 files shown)


bodhi/client/__init__.py
```diff
@@ -456,7 +456,14 @@ def save_buildroot_overrides(user, password, url, staging, **kwargs):
...
     """
-    _save_override(url=url, user=user, password=password, staging=staging, **kwargs)
+
+    try:
+        _save_override(url=url, user=user, password=password, staging=staging, **kwargs)
+    except bindings.BodhiClientException as e:
+        click.echo(str(e))
+        if str(e) == "Buildroot override for %s already exists" % (kwargs['nvr']):
+            click.echo("The `overrides save` command is used for creating a new override")
+            click.echo("Use `overrides edit` to edit an existing override")
```
bowlofeggs: I would suggest moving the ```click.echo()``` into the ```if``` block, and adding an ```else``` with ```raise```, so other errors still print a traceback. That way we can get better debugging info if users experience a different problem than this one.

bowlofeggs (changes requested):
The new lines are missing coverage:

```
-------------
Diff Coverage
-------------
bodhi/client/__init__.py (28.6%): Missing lines 462-466
-------------
Total:   7 lines
Missing: 5 lines
Coverage: 28%
-------------
```

bodhi/client/__init__.py
ryanlerch: Did this, thanks!

ryanlerch:
Added the additional tests as well.

bdiffcover is pretty neat!

bodhi/client/__init__.py
```diff
@@ -456,7 +456,16 @@ def save_buildroot_overrides(user, password, url, staging, **kwargs):
...
     """
-    _save_override(url=url, user=user, password=password, staging=staging, **kwargs)
+
+    try:
+        _save_override(url=url, user=user, password=password, staging=staging, **kwargs)
+    except bindings.BodhiClientException as e:
+        if str(e) == "Buildroot override for %s already exists" % (kwargs['nvr']):
+            click.echo(str(e))
+            click.echo("The `overrides save` command is used for creating a new override")
+            click.echo("Use `overrides edit` to edit an existing override")
```
bowlofeggs: I recommend adding a period on the end of each of these sentences.

bodhi/tests/client/test___init__.py
```diff
@@ -310,6 +310,43 @@ def test_url_flag(self, send_request):
...
+        """
+        Assert that the error message is provided if we try to save an existing override
+        """
+        exception_message = "Buildroot override for js-tag-it-2.0-1.fc25 already exists"
+        send_request.side_effect = bindings.BodhiClientException(exception_message)
+        runner = testing.CliRunner()
+
+        result = runner.invoke(
+            client.save_buildroot_overrides,
+            ['--user', 'bowlofeggs', '--password', 's3kr3t', 'js-tag-it-2.0-1.fc25'])
```
bowlofeggs, line 326: Just a recommendation (feel free to ignore): I suggest adding a space here. I like to group my tests into three groups visually, using spaces to separate them. The first group is code that sets up the test (like you have above this line). The middle group is usually a 1-liner that runs the thing being tested (i.e., this line). And the last group is the assertions that the right thing happened. It's just a personal preference, and just a suggestion ☺

bodhi/tests/client/test___init__.py
```diff
@@ -310,6 +310,43 @@ def test_url_flag(self, send_request):
...
+                      result.output)
+
+    @mock.patch('bodhi.client.bindings.BodhiClient.csrf',
+                mock.MagicMock(return_value='a_csrf_token'))
+    @mock.patch('bodhi.client.bindings.BodhiClient.send_request', autospec=True)
+    def test_other_client_exception(self, send_request):
+        """
+        Assert that any other BodhiClientExceptions are raised as expected
+        """
+        send_request.side_effect = bindings.BodhiClientException("Pants Exception")
```
bowlofeggs, line 341: haha

bowlofeggs (approved):
Both comments are just suggestions at your option. No need for re-review if you decide to go with them, so merge as you please!

bodhi/client/__init__.py
ryanlerch: Added the fullstops. Thanks! ;)

bodhi/tests/client/test___init__.py
ryanlerch: I like this, it makes sense to me! Added!

PR #1609, opened 2017-06-15
PEP-257 bodhi.server.views.admin.

bowlofeggs:

### DIFF
--- a/bodhi/server/views/admin.py
+++ b/bodhi/server/views/admin.py
@@ -1,3 +1,8 @@
+# -*- coding: utf-8 -*-
+# Copyright © 2014-2017 Red Hat, Inc. and others
+#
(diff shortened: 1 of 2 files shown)


bodhi/server/views/admin.py
```diff
@@ -24,6 +30,17 @@
 
 @admin_service.get(permission='admin')
 def admin(request):
+    """
+    Return a dictionary with keys "user" and "principal".
```
ryanlerch: Tiny nit: should this be "principals"?

ryanlerch (approved):

bodhi/server/views/admin.py
bowlofeggs: Corrected, thanks!

PR #1614, opened 2017-06-15
Remove diff-cover from test requirements, since it is not required.

bowlofeggs:
The tests pass just fine without diff-cover. It is merely a tool to
aid in comparing test coverage between branches. Thus, it is not a
requirement.


ryanlerch (approved):
bdiffcover still works out-of-the-box on a fresh vagrant env too, woot!

thanks @bowlofeggs

Issue #1597, opened 2017-06-13
bodhi.server.util.taskotron_results doesn't use a timeout on its call to requests.get()


bowlofeggs:
bodhi.server.util.taskotron_results should use a timeout on its call to requests.get() so that it won't hang indefinitely if there are issues communicating with resultsdb. 60 seconds is probably reasonable.

rimshaakhan:
Hello @bowlofeggs !

I would like to work on this issue, can you please assign it to me. Thanks!

bowlofeggs:
Hello @rimshaakhan,

I don't think GitHub allows assigning issues to users who are not in the fedora-infra group, but you are certainly welcome to work on this issue (you don't have to be officially assigned to an issue to work on it). Just make sure you end your commit message with a line that says "fixes #1597" when you send your pull request. Thanks!

bowlofeggs:
The fix for this issue will appear in the upcoming 2.11.0 release:

https://github.com/fedora-infra/bodhi/pull/1790

PR #1621, opened 2017-06-18
Add a timeout in taskotron_results

rimshaakhan:
Add a timeout to requests.get() , now prevents bodhi.server.util.taskotron_results from hanging indefinitely.

Closes: https://github.com/fedora-infra/bodhi/issues/1597

### DIFF
--- a/bodhi/server/util.py
+++ b/bodhi/server/util.py
@@ -641,7 +641,7 @@ def taskotron_results(settings, entity='results/latest', max_queries=10, **kwarg
     try:
         while data and url:
             log.debug("Grabbing %r" % url)
-            response = requests.get(url)
+            response = requests.get(url, timeout=60)
             if response.status_code != 200:
                 raise IOError("status code was %r" % response.status_code)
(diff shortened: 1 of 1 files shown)


pypingou:
Looks good to me :)

bowlofeggs:
Let's add a test assertion for this change. I think we could probably just tack on something to one of the existing tests (or you can make your own new test if you prefer). For example, you could probably do something like this at the end of [this test](https://github.com/fedora-infra/bodhi/blob/619e008ffbb84e8a7c3259dcb26a883379200c4d/bodhi/tests/server/test_utils.py#L462-L479):

```
self.assertEqual(mock_get.call_args[1]['timeout'], 60)
```

bowlofeggs:
Hello @rimshaakhan, do you still plan to work on this?

bowlofeggs:
I have rebased this commit and added one of my own on top with the test I proposed above at https://github.com/fedora-infra/bodhi/pull/1746.

PR #1622, opened 2017-06-19
add mine flag in cli for updates & overrides query

### LINKED ISSUE #811: bodhi --mine gives KeyError
Running "bodhi --mine --username=puiterwijk" results in a KeyError: 'title'.

Also, not all update information is displayed:
....
 python-openid-cla-1.2-1.fc22                 enhancement  stable    2016-03-02 
 python-fedora-0.7.1-4.fc22
 python-openid-cla-1.2-1.el6                  enhancement  stable    2016-03-10 
 python-openid-cla-1.2-1.el7                  enhancement  stable    2016-03-10 
 python-fedora-0.7.1-4.fc23                   enhancement  stable    2016-02-22 
 python-openid-cla-1.2-1.fc23
 devtodo2-2.1-14.20120711git8dee6.fc23        bugfix       stable    2016-02-21 
.....

ryanlerch:
Implement a --mine flag for `bodhi updates query` & `bodhi overrides query`. The flag uses the USERNAME envvar as the user parameter for the query so a user can quickly view their own updates or overrides.

Fixes #811 #1382

### DIFF
--- a/bodhi/client/__init__.py
+++ b/bodhi/client/__init__.py
@@ -285,10 +285,11 @@ def edit(user, password, url, **kwargs):
 @click.option('--type', default=None, help='Filter by update type',
               type=click.Choice(['newpackage', 'security', 'bugfix', 'enhancement']))
 @click.option('--user', help='Updates submitted by a specific user')
+@click.option('--mine', is_flag=True, help='Show only your updates')
 @click.option('--staging', help='Use the staging bodhi instance',
               is_flag=True, default=False)
 @url_option
-def query(url, **kwargs):
+def query(url, mine=False, **kwargs):
     # User Docs that show in the --help
     """
     Query updates on Bodhi
@@ -301,9 +302,16 @@ def query(url, **kwargs):
     Args:
         url (unicode): The URL of a Bodhi server to create the update on. Ignored if staging is
                        True.
+        mine (Boolean): If the --mine flag was set
         kwargs (dict): Other keyword arguments passed to us by click.
     """
     client = bindings.BodhiClient(base_url=url, staging=kwargs['staging'])
+    if mine:
+        if 'USERNAME' in os.environ:
+            kwargs['user'] = os.environ['USERNAME']
+        else:
+            client.init_username()
+            kwargs['user'] = client.username
     resp = client.query(**kwargs)
     print_resp(resp, client)
 
@@ -494,8 +502,10 @@ def overrides():
               help='Updates submitted by a specific user')
 @click.option('--staging', help='Use the staging bodhi instance',
               is_flag=True, default=False)
+@click.option('--mine', is_flag=True,
+              help='Show only your overrides.')
 @url_option
-def query_buildroot_overrides(url, user=None, **kwargs):
+def query_buildroot_overrides(url, user=None, mine=False, **kwargs):
     # Docs that show in the --help
     """
     Query the buildroot overrides.
@@ -508,11 +518,18 @@ def query_buildroot_overrides(url, user=None, **kwargs):
     Args:
         user (unicode): If supplied, overrides for this user will be queried.
         staging (bool): Whether to use the staging server or not.
+        mine (bool): Whether to use the --mine flag was given.
         url (unicode): The URL of a Bodhi server to create the update on. Ignored if staging is
                        True.
         kwargs (dict): Other keyword arguments passed to us by click.
     """
     client = bindings.BodhiClient(base_url=url, staging=kwargs['staging'])
+    if mine:
+        if 'USERNAME' in os.environ:
+            user = os.environ['USERNAME']
+        else:
+            client.init_username()
+            user = client.username
     resp = client.list_overrides(user=user)
     print_resp(resp, client)
 
--- a/bodhi/tests/client/test_bindings.py
+++ b/bodhi/tests/client/test_bindings.py
@@ -88,8 +88,8 @@ class TestBodhiClient_comment(unittest.TestCase):
                   'email': True, 'csrf_token': 'a token'})
 
 
-class TestBodhiClient__init_username(unittest.TestCase):
-    """Test the BodhiClient._init_username() method."""
+class TestBodhiClient_init_username(unittest.TestCase):
+    """Test the BodhiClient.init_username() method."""
     TEST_EMPTY_OBJECT_SESSION_CACHE = '{}'
     TEST_FAILED_SESSION_CACHE = '{"https://bodhi.fedoraproject.org/:bowlofeggs": []}'
     TEST_HOT_SESSION_CACHE = '{"https://bodhi.fedoraproject.org/:bowlofeggs": [["stuff", "login"]]}'
@@ -109,7 +109,7 @@ class TestBodhiClient__init_username(unittest.TestCase):
         mock_raw_input.return_value = 'pongou'
         client = bindings.BodhiClient()
 
-        client._init_username()
+        client.init_username()
 
         exists.assert_called_once_with(fedora.client.openidbaseclient.b_SESSION_FILE)
         self.assertEqual(_load_cookies.mock_calls, [mock.call(), mock.call()])
@@ -131,7 +131,7 @@ class TestBodhiClient__init_username(unittest.TestCase):
         mock_raw_input.return_value = 'pongou'
         client = bindings.BodhiClient()
 
-        client._init_username()
+        client.init_username()
 
         exists.assert_called_once_with(fedora.client.openidbaseclient.b_SESSION_FILE)
         self.assertEqual(_load_cookies.mock_calls, [mock.call(), mock.call()])
@@ -153,7 +153,7 @@ class TestBodhiClient__init_username(unittest.TestCase):
         mock_raw_input.return_value = 'pongou'
         client = bindings.BodhiClient()
 
-        client._init_username()
+        client.init_username()
 
         exists.assert_called_once_with(fedora.client.openidbaseclient.b_SESSION_FILE)
         self.assertEqual(_load_cookies.mock_calls, [mock.call(), mock.call()])
@@ -174,7 +174,7 @@ class TestBodhiClient__init_username(unittest.TestCase):
         mock_open.side_effect = mock.mock_open(read_data=self.TEST_HOT_SESSION_CACHE)
         client = bindings.BodhiClient()
 
-        client._init_username()
+        client.init_username()
 
         exists.assert_called_once_with(fedora.client.openidbaseclient.b_SESSION_FILE)
         self.assertEqual(_load_cookies.mock_calls, [mock.call(), mock.call()])
@@ -203,7 +203,7 @@ class TestBodhiClient__init_username(unittest.TestCase):
         mock_open.side_effect = mock.mock_open(read_data=self.TEST_HOT_SESSION_CACHE)
         client = bindings.BodhiClient()
 
-        client._init_username()
+        client.init_username()
 
         exists.assert_called_once_with(fedora.client.openidbaseclient.b_SESSION_FILE)
         loads.assert_called_once_with(self.TEST_HOT_SESSION_CACHE)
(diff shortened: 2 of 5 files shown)


bowlofeggs:
test

bodhi/client/__init__.py
```diff
@@ -43,6 +43,48 @@ def _warn_if_url_and_staging_set(ctx, param, value):
...
+    Args:
+        ctx (click.core.Context): The Click context, used to find out if the --user flag is set.
+        param (click.core.Option): The option being handled. Unused.
+        value (Boolean): True if the --mine flag was given.
+
+    Returns:
+        Boolean: The value of the --mine flag.
+    """
+    if value and not os.environ['USERNAME']:
+        raise click.BadParameter('USERNAME environment variable not set.')
```
bowlofeggs: I recently created an [```_init_username()```](https://github.com/fedora-infra/bodhi/blob/2.7.0/bodhi/client/bindings.py#L335-L363) method that can be used to get the user's username automatically. It uses their saved cookies to find if they've ever authenticated before and uses that name if they have. Otherwise, it falls back to prompting the user. The environment variable will override what this does since it will set username when the BodhiClient is initialized.

What if we make that a public method (by renaming it from ```_init_username``` to ```init_username```) and use that to ensure that the username will work as people expect? This way, they don't have to set an environment variable (though they still can).

bodhi/client/__init__.py
```diff
@@ -43,6 +43,48 @@ def _warn_if_url_and_staging_set(ctx, param, value):
...
+    elif value and ctx.params.get('user'):
+        # also check in this callback, in case the --mine was given after --user
```
bowlofeggs: In your testing, does the order of the parameters affect whether each of the callbacks see the other setting?

bodhi/client/__init__.py
```diff
@@ -299,9 +344,12 @@ def query(url, **kwargs):
     Args:
         url (unicode): The URL of a Bodhi server to create the update on. Ignored if staging is
                        True.
+        mine (Boolean): If the --mine flag was set
         kwargs (dict): Other keyword arguments passed to us by click.
     """
     client = bindings.BodhiClient(base_url=url, staging=kwargs['staging'])
+    if mine:
+        kwargs['user'] = os.environ['USERNAME']
```
bowlofeggs: If you put a call to ```client.init_username``` above this, I think you can just do ```kwargs['user'] = client.username``` here.

bodhi/client/__init__.py
```diff
@@ -506,11 +557,14 @@ def query_buildroot_overrides(url, user=None, **kwargs):
...
         user (unicode): If supplied, overrides for this user will be queried.
         staging (bool): Whether to use the staging server or not.
+        mine (bool): Whether to use the --mine flag was given.
         url (unicode): The URL of a Bodhi server to create the update on. Ignored if staging is
                        True.
         kwargs (dict): Other keyword arguments passed to us by click.
     """
     client = bindings.BodhiClient(base_url=url, staging=kwargs['staging'])
+    if mine:
+        user = os.environ['USERNAME']
```
bowlofeggs: Similar recommendations here.

bodhi/client/__init__.py
```diff
@@ -43,6 +43,48 @@ def _warn_if_url_and_staging_set(ctx, param, value):
...
+        param (click.core.Option): The option being handled. Unused.
+        value (unicode): The value of the --user flag.
+
+    Returns:
+        unicode: The value of the --user flag.
+    """
+    if value and ctx.params.get('mine'):
+        click.echo('Warning: --mine and --user options are both set. --user will be ignored.\n',
+                   err=True)
+    return value
```
bowlofeggs: If both validators are really needed, and since this validator only prints a warning, you can probably save writing this code twice (since the other validator has the same check) and move this warning into the bodies of the functions that use them.

bowlofeggs (changes requested):

bodhi/client/__init__.py
ryanlerch: Bit confused with this here. I like the init_username idea, but with the current check here in place, the validator won't pass, and we wont get down to running the init_username.

IMHO, we could just use the init_username for the --mine, and not worry about the environment variable at all. This means that a user that has logged in before, --mine will just work, otherwise, they will need to type in their username every time --mine is run.

Of course, once we enable openidconnect, this will make a little more sense i think.

bowlofeggs: I think the validator is still OK - if the user supplies ```--username``` or the environment variable, those things override what ```init_username``` does (it'll just no-op if ```self.username``` is already set). So the validator checks are fine and I think The Right Thing™ will happen whether or not the user uses the flag or env var. ```init_username()``` only takes action when the user doesn't tell us who they are through those mechanisms.

ryanlerch: I had the init_username in the main method for the query methods, so if the envvar was not set, it never got down to the init_username() call, becasue the validator would fail on not having an envvar set. going to move the init_username() call up to the validator, and call it only if the envvar is not set.

bowlofeggs:
On Mon, 2017-06-26 at 21:32 -0700, Ryan Lerch wrote:
> I had the init_username in the main method for the query methods, so
> if the envvar was not set, it never got down to the init_username()
> call, becasue the validator would fail on not having an envvar set.
> going to move the init_username() call up to the validator, and call
> it only if the envvar is not set.

Ah I think I see why it's confusing a little better now - I don't think
you need to validate that --user or USER is set when --mine is used. If
those aren't set, init_username() will prompt the user for a username.
So I suggest not worrying about validating and just let it prompt when
needed. Does that help?

ryanlerch:
Okies, updated the code. Just need to fix up the tests -- i started on it, but need to finish it off.

ryanlerch:
Okies, fixed the tests, this one should be right for a re-review!

bodhi/client/__init__.py
```diff
@@ -285,10 +285,12 @@ def edit(user, password, url, **kwargs):
 @click.option('--type', default=None, help='Filter by update type',
               type=click.Choice(['newpackage', 'security', 'bugfix', 'enhancement']))
 @click.option('--user', help='Updates submitted by a specific user')
+@click.option('--mine', is_flag=True,
+              help='Show only your updates. User is defined by the $USERNAME envvar')
 @click.option('--staging', help='Use the staging bodhi instance',
               is_flag=True, default=False)
 @url_option
-def query(url, **kwargs):
+def query(url, mine=None, **kwargs):
```
bowlofeggs: Optional: For consistency with the type, you could make this default to ```False``` instead of ```None```. No big deal though, as it is functionally the same.

bodhi/client/__init__.py
```diff
@@ -285,10 +285,12 @@ def edit(user, password, url, **kwargs):
 @click.option('--type', default=None, help='Filter by update type',
               type=click.Choice(['newpackage', 'security', 'bugfix', 'enhancement']))
 @click.option('--user', help='Updates submitted by a specific user')
+@click.option('--mine', is_flag=True,
+              help='Show only your updates. User is defined by the $USERNAME envvar')
```
bowlofeggs: Technically, if the user has logged in before the user is also defined by their login cookies. And if they both don't set ```USERNAME``` and haven't logged in, ```init_username()``` will prompt them for a username. Thus, I think we should either just not mention the ```USERNAME``` bit, or mention the other ways too. Since the man page covers the existence of ```USERNAME```, I think we could just leave the second sentence off here.

bodhi/client/__init__.py
```diff
@@ -494,8 +506,10 @@ def overrides():
               help='Updates submitted by a specific user')
 @click.option('--staging', help='Use the staging bodhi instance',
               is_flag=True, default=False)
+@click.option('--mine', is_flag=True,
+              help='Show only your overrides. User is defined by the $USERNAME envvar')
```
bowlofeggs: Same thing here - the user can be defined three ways. It might be best to just leave it at "Show only your overrides." so the helptext isn't too verbose. The USERNAME variable is defined in the man page.

bodhi/client/__init__.py
```diff
@@ -301,9 +303,19 @@ def query(url, **kwargs):
     Args:
         url (unicode): The URL of a Bodhi server to create the update on. Ignored if staging is
                        True.
+        mine (Boolean): If the --mine flag was set
```
bowlofeggs, line 305: Let's add this and the other ```--mine``` flag to the man page at ```docs/man_pages/bodhi.rst```.

bodhi/client/__init__.py
bowlofeggs: Optional: You could also use the man page for the expanded details about how the flag finds the username if you like.

bodhi/tests/client/test_bindings.py
```diff
@@ -88,8 +88,8 @@ def test_comment(self):
                   'email': True, 'csrf_token': 'a token'})
 
 
-class TestBodhiClient__init_username(unittest.TestCase):
-    """Test the BodhiClient._init_username() method."""
+class TestBodhiClientinit_username(unittest.TestCase):
```
bowlofeggs: Optional: might be nice to keep one _ between BodhiClient and init_username.

bodhi/client/__init__.py
```diff
@@ -301,9 +303,19 @@ def query(url, **kwargs):
...
         kwargs (dict): Other keyword arguments passed to us by click.
     """
     client = bindings.BodhiClient(base_url=url, staging=kwargs['staging'])
+    if mine:
+        if 'user' in kwargs:
```
bowlofeggs: I think this ```if``` statement might always be True. In my testing, I neither set the environment variable nor set the --user flag, but saw:

```
Warning: --mine and --user options both set. --user will be ignored.
```

In these cases, ```user``` is in ```kwargs``` but is set to ```None```. Perhaps the ```if``` statement here and elsewhere should be:

```
if kwargs.get('user', None):
```

bodhi/client/__init__.py
```diff
@@ -508,11 +522,21 @@ def query_buildroot_overrides(url, user=None, **kwargs):
...
+        mine (bool): Whether to use the --mine flag was given.
         url (unicode): The URL of a Bodhi server to create the update on. Ignored if staging is
                        True.
         kwargs (dict): Other keyword arguments passed to us by click.
     """
     client = bindings.BodhiClient(base_url=url, staging=kwargs['staging'])
+    if mine:
+        if user:
+            click.echo('Warning: --mine and --user options both set. --user will be ignored.\n',
+                       err=True)
```
bowlofeggs: So one problem with this warning is that it will be displayed if the user uses ```USERNAME``` - that effectively comes in as ```--user```, so this causes the warning:

```
USERNAME="bowlofeggs" bodhi overrides query --mine
```

I wonder if it would be best to just not do these warnings? I struggle to think of a better way to handle it without falsely warning the user.

bowlofeggs (changes requested):

bodhi/client/__init__.py
bowlofeggs: As I mentioned later, these warnings will also happen if the user uses the ```USERNAME``` environment variable, so perhaps we just don't need to warn here at all and can just drop the ```if``` statement?

bodhi/client/__init__.py
ryanlerch: Dropped the warning!

bodhi/client/__init__.py
ryanlerch: removed the second sentence

bodhi/client/__init__.py
ryanlerch: done

bodhi/client/__init__.py
ryanlerch: done. We were not documenting any options for overrides query, so i also added the --user option here too.

bodhi/client/__init__.py
ryanlerch: done

bodhi/client/__init__.py
ryanlerch: done.

bowlofeggs (approved):
Very nice!

Issue #1619, opened 2017-06-17
The home page doesn't show stats for branched releases


bowlofeggs:
The home page doesn't show stats for branched releases, for example, Fedora 26 at the time of this writing.

keszybz:
I think this *might* be on purpose: in the drop-down menu, F26 is separated at the bottom. Nevertheless, I think that's a bad choice, because most of the users of that page are Fedora packagers and testers, for whom branched is of primary interest.

PR #1623, opened 2017-06-19
add pending releases to the frontpage

ryanlerch:
Previously, the new frontpage only included 'current' releases. This commit adds pending releases also, with pending releases having a label 'prerelease' in next to the release name.

Fixes: #1619

### DIFF
--- a/bodhi/server/views/generic.py
+++ b/bodhi/server/views/generic.py
@@ -208,7 +208,7 @@ def home(request):
         critpath_updates = get_latest_updates(request, True, False)
         security_updates = get_latest_updates(request, False, True)
         release_updates_counts = {}
-        for release in request.releases['current']:
+        for release in request.releases['current'] + request.releases['pending']:
             release_updates_counts[release["name"]] = get_update_counts(request, release["name"])
 
         return {
--- a/bodhi/tests/server/functional/test_generic.py
+++ b/bodhi/tests/server/functional/test_generic.py
@@ -19,7 +19,8 @@ from pyramid.testing import DummyRequest
 from webtest import TestApp
 
 from bodhi.server import main, util
-from bodhi.server.models import (Group, User, Update, Release, UpdateStatus, UpdateType)
+from bodhi.server.models import (
+    Group, User, Update, Release, ReleaseState, UpdateStatus, UpdateType)
 from bodhi.server.security import remember_me
 from bodhi.tests.server import base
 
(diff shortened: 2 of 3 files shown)


pypingou:
Change in itself looks fine, screenshot ? :)

bowlofeggs:
test

bodhi/server/views/generic.py
```diff
@@ -208,7 +208,7 @@ def work():
         critpath_updates = get_latest_updates(request, True, False)
         security_updates = get_latest_updates(request, False, True)
         release_updates_counts = {}
-        for release in request.releases['current']:
+        for release in request.releases['current'] + request.releases['pending']:
```
bowlofeggs, line 211: We should test that these pending releases get included here.

bowlofeggs (changes requested):
Don't worry about the CI - CI won't work until #1631 is merged.

ryanlerch:
Added the test.

However, there is an issue. For some reason, when i run the tests by themselves, i.e:
```
btest --tests bodhi/tests/server/functional/test_generic.py
```

They pass fine. But the same test fails when trying to run the whole test suite (i.e. `btest`). The tests fail because the new release that i add to test against is not showing up on the frontpage when running the test. I tried to figure out what was going on here, but it really stumped me. There is something about adding the extra release in the test that is not being picked up.

I thought it might be caching something from a previous test, but could not figure out how to disable caching, or force reload in the test.

bowlofeggs:
test

bodhi/server/views/generic.py
```diff
@@ -208,7 +208,7 @@ def work():
         critpath_updates = get_latest_updates(request, True, False)
```
bowlofeggs, line 208: This function is cached - that might be the cause of the test isolation problems.

bowlofeggs: I tried disabling the cache and it didn't help, so I was wrong.

jeremycline:
In my database test refactoring work, I've also found bodhi/tests/server/functional/test_generic.py to be very fragile and order-dependent. I plan to track the problem down as part of that refactoring work and fix it. I'm not sure how long it will take though, I spent a good bit of yesterday investigating a different set of order-dependent tests :(.

Maybe by the end of the week?

bowlofeggs:
@ryanlerch Why don't we wait and see what @jeremycline comes up with - perhaps you will be able to simply rebase on a commit from him and everything will be fine?

ryanlerch:
@jeremycline @bowlofeggs thanks for looking into this!

ryanlerch:
Just rebased this PR to include the recent changes to the test infrastruture that @jeremycline implemented in #1664 , and the test now works when running all the tests. Thanks @jeremycline & @bowlofeggs !

bowlofeggs (changes requested):
The releases visually cascade with this change:

delme

The code looks good though - can we get the releases to line up?

ryanlerch:
@bowlofeggs thanks for catching the lining up issue, mucked up the rebase there, sorry!

Updated the commit to fix an extra div that was causing the cascading issue, and this one should be right to go!

cheers,
ryanlerch

bodhi/tests/server/functional/test_generic.py
```diff
@@ -407,3 +330,246 @@ def test_api_version(self):
...
+                           [UpdateType.enhancement, 4],
+                           [UpdateType.newpackage, 4]]]]
+        _add_updates(addedupdates2, user2, pendingrelease, "fc18")
+        self.db.flush()
+        Release._tag_cache = None
+
+    def test_home_counts(self):
+        """Test the frontpage update counts"""
+        res = self.app.get('/', status=200)
+        print res
```

bowlofeggs (changes requested):
Just one tiny fix left, looks good!

bowlofeggs:
Feel free to merge once that print statement is dropped!

ryanlerch:
Hi Jenkies!

It was failing because i hadn't changed the base test class to the new one that @jeremycline pushed recently. Just updated the PR to use the new testing base class.

ryanlerch:
Thanks for testing it again Jenkies. :D 

@bowlofeggs i also removed the print statement (thanks for catching that one!)

merging now!

PR #1629, opened 2017-06-20
Install epel-release for CI.

bowlofeggs:


jeremycline (approved):
Well that little green checkmark is a lie because the tests look to be failing (missing koji?), but the change looks good :shipit:

bowlofeggs:
Yeah I think https://github.com/fedora-infra/bodhi/pull/1630 would have fixed the green check mark, but I think this job ran before that was merged.

PR #1630, opened 2017-06-20
Fix a typo so CI sees the true return code.

bowlofeggs:


pypingou:
👍 for me

PR #1631, opened 2017-06-20
Get CI working

bowlofeggs:
This pull request has three commits. The first commit reworks the new CI run_tests.sh script so that it does everything needed for the tests to pass. It also enforces diff-coverage now. The next two commits fix test failures that happen in CentOS CI that don't happen in the Vagrant dev environment.

### DIFF
--- a/bodhi/tests/server/scripts/test_approve_testing.py
+++ b/bodhi/tests/server/scripts/test_approve_testing.py
@@ -38,8 +38,7 @@ class TestMain(BaseTestCase):
         update.request = None
         update.stable_karma = 10
         update.status = models.UpdateStatus.testing
-        update.date_testing = datetime.now() - timedelta(days=7)
-        self.db.flush()
+        update.date_testing = datetime.utcnow() - timedelta(days=7)
 
         with patch('bodhi.server.scripts.approve_testing.initialize_db'):
             with patch('bodhi.server.scripts.approve_testing.get_appsettings', return_value=''):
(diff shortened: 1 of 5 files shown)


jeremycline:
Hmm, still looks like it's failing because (among other things) it's missing alembic

bowlofeggs:
@jeremycline Yeah I'm doing the grind of repeatedly running the tests and fixing my commit and force pushing, over and over. Also, I'm pretty sure that only my changes to run_tests are being reflected in the test run and not my changes to the yaml file. I think (for good reason) Jenkies always uses the yaml file from the develop branch. So no idea what effect those have (and I still don't know why the exit code isn't causing Jenkins to think the tests failed.)

bowlofeggs:
I finally have CI passing! I need to go for the day, but for the record I do plan to break this commit up into more atomic commits before asking for review.

bowlofeggs:
This pull request is now ready for review.

bodhi/tests/server/scripts/test_approve_testing.py
```diff
@@ -38,8 +38,7 @@ def test_autokarma_update_meeting_time_requirements_gets_one_comment(self):
         update.request = None
         update.stable_karma = 10
         update.status = models.UpdateStatus.testing
-        update.date_testing = datetime.now() - timedelta(days=7)
-        self.db.flush()
+        update.date_testing = datetime.utcnow() - timedelta(days=7)
```
jeremycline, line 41: ✨

jeremycline (approved):
:shipit: so good

bowlofeggs:
Let the integration be continuous.

PR #1633, opened 2017-06-21
Fix the pkgtags_url default

puiterwijk:
Before the new config system, not specifying a configuration option would set it
to an empty string or None.

The pkgtags code is entirely broken, since it overwrites the updateinfo entry in
repomd.xml because modifyrpeo only takes a filename, and then update updateinfo with
that.
Before, this was not a problem since we just did not configure pkgtags_url, which
means that pagure.server.metadata's insert_pkgtags would not do anything.
With the new config code, this code now gets used, which broke all repos.

NOTE: We should make *VERY* sure there are no other such places where code was not
used because config.get(foobar) returned None, and now returns the default value,
with an incorrect default value.

### DIFF
--- a/bodhi/server/config.py
+++ b/bodhi/server/config.py
@@ -483,7 +483,7 @@ class BodhiConfig(dict):
             'value': 'https://admin.fedoraproject.org/pkgdb',
             'validator': unicode},
         'pkgtags_url': {
-            'value': 'https://apps.fedoraproject.org/tagger/api/v1/tag/sqlitebuildtags/',
+            'value': '',
             'validator': unicode},
         'query_wiki_test_cases': {
             'value': False,
--- a/development.ini.example
+++ b/development.ini.example
@@ -252,7 +252,8 @@ dogpile.cache.arguments.filename = %(here)s/dogpile-cache.dbm
 
 # We used to get our package tags from pkgdb, but they come from tagger now.
 # https://github.com/fedora-infra/fedora-tagger/pull/74
-# pkgtags_url = https://apps.fedoraproject.org/tagger/api/v1/tag/sqlitebuildtags/
+# Fedora's pkgtags can be found at https://apps.fedoraproject.org/tagger/api/v1/tag/sqlitebuildtags/
+# pkgtags_url =
 
 ##
 ## Bug tracker settings
--- a/production.ini
+++ b/production.ini
@@ -252,7 +252,8 @@ use = egg:bodhi-server
 
 # We used to get our package tags from pkgdb, but they come from tagger now.
 # https://github.com/fedora-infra/fedora-tagger/pull/74
-# pkgtags_url = https://apps.fedoraproject.org/tagger/api/v1/tag/sqlitebuildtags/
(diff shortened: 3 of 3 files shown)


bodhi/server/config.py
```diff
@@ -483,7 +483,7 @@ class BodhiConfig(dict):
             'value': 'https://admin.fedoraproject.org/pkgdb',
             'validator': unicode},
         'pkgtags_url': {
-            'value': 'https://apps.fedoraproject.org/tagger/api/v1/tag/sqlitebuildtags/',
+            'value': '',
```
bowlofeggs, line 486: Ah, this was definitely a mistake. I had seen that this value was commented in [production.ini](https://github.com/fedora-infra/bodhi/blob/2.7.0/production.ini#L266-L268):

```
# We used to get our package tags from pkgdb, but they come from tagger now.
# https://github.com/fedora-infra/fedora-tagger/pull/74
#pkgtags_url = https://apps.fedoraproject.org/tagger/api/v1/tag/sqlitebuildtags/
```

and mistakenly made it the default. This probably happened as a result of how I made the change - I had been going through the file and the code for each setting, and as I encoded each setting I also made the default be in the ```production.ini``` file commented like the above. My guess is that I saw that commented value and assumed that I commented it, and then put the value to match it. D'oh.

Thanks so much for figuring this out and for fixing it. You are a huge help!

bowlofeggs (approved):

bowlofeggs:
I filed https://github.com/fedora-infra/bodhi/issues/1634 about considering removing the code, as you suggested.

bowlofeggs:
Thanks so much for the patch! I am going to amend your commit with this patch and then I'll merge it:

```
diff --git a/development.ini.example b/development.ini.example
index 381be96..d603c79 100644
--- a/development.ini.example
+++ b/development.ini.example
@@ -252,7 +252,8 @@ dogpile.cache.arguments.filename = %(here)s/dogpile-cache.dbm
 
 # We used to get our package tags from pkgdb, but they come from tagger now.
 # https://github.com/fedora-infra/fedora-tagger/pull/74
-# pkgtags_url = https://apps.fedoraproject.org/tagger/api/v1/tag/sqlitebuildtags/
+# Fedora's pkgtags can be found at https://apps.fedoraproject.org/tagger/api/v1/tag/sqlitebuildtags/
+# pkgtags_url =
 
 ##
 ## Bug tracker settings
index 483e4ab..7c611db 100644
--- a/production.ini
+++ b/production.ini
@@ -252,7 +252,8 @@ use = egg:bodhi-server
 
 # We used to get our package tags from pkgdb, but they come from tagger now.
 # https://github.com/fedora-infra/fedora-tagger/pull/74
-# pkgtags_url = https://apps.fedoraproject.org/tagger/api/v1/tag/sqlitebuildtags/
+# Fedora's pkgtags can be found at https://apps.fedoraproject.org/tagger/api/v1/tag/sqlitebuildtags/
+# pkgtags_url =
 
 ##
 ## Bug tracker settings
```

bowlofeggs:
This has been cherry picked to the 2.8 branch as 9d95a003b7a1f8df5859cc8e0e2aa7bd2bbc144c and will be included in bodhi-2.8.1.

PR #1637, opened 2017-06-21
Fix test warnings that happen due to calling exception.message.

bowlofeggs:


jeremycline (approved):
Looks good :shipit:

Is CI still a work in progress? I see this one failed because there was no koji, but I remember you fixed that.

bowlofeggs:
@jeremycline yeah, CI won't work until #1631 is merged (and it's finally ready for review).

PR #1638, opened 2017-06-21
Add '--requirements' flag to CLI

### LINKED ISSUE #1319: It should be possible to set required tests with the cli
When creating or editing an update with the web interface, it is possible to set required tests. For example, you can type ```dist.depcheck``` and Bodhi will make sure that taskotron test passed for the update. The CLI doesn't provide a way to specify this, we should add it.

crungehottman:
Permits users to set required Taskatron tasks in the CLI via a '--requirements' flag.

fixes #1319


cverna:
Looks good, it would be good to add the new option to the man page too. See https://github.com/fedora-infra/bodhi/blob/develop/docs/man_pages/bodhi.rst

👍

docs/man_pages/bodhi.rst
```diff
@@ -179,6 +179,11 @@ The ``updates`` command allows users to interact with bodhi updates.
 
         A path to a file containing all the update details.
 
+    ``--requirements <Taskatron tasks>``
```
bowlofeggs: It's spelled Taskotron (though the [cool looking] logo font makes it a little ambiguous): https://taskotron.fedoraproject.org/

docs/man_pages/bodhi.rst
```diff
@@ -179,6 +179,11 @@ The ``updates`` command allows users to interact with bodhi updates.
...
+
+        A comma or space-separated list of required Taskatron tasks that must pass for this update
```
bowlofeggs: Taskotron here too

docs/man_pages/bodhi.rst
```diff
@@ -179,6 +179,11 @@ The ``updates`` command allows users to interact with bodhi updates.
...
+        to reach stable.
```
bowlofeggs: @kparal is the intention for these tests to stop broken updates from reaching stable, or updates-testing?

docs/man_pages/bodhi.rst
```diff
@@ -225,6 +230,11 @@ The ``updates`` command allows users to interact with bodhi updates.
 
         Suggest that the user logout or reboot upon applying the update.
 
+    ``--requirements <Taskatron tasks>``
+
+        A comma or space-separated list of required Taskatron tasks that must pass for this update
```
bowlofeggs: Taskotron here and the line above too.

bowlofeggs (changes requested):

bowlofeggs:
test

docs/man_pages/bodhi.rst
kparal: @bowlofeggs All tests are targeting stable so far, but in theory we could add tests that target updates-testing in the future. That would likely require more changes, though, and is not probably anything short-term.

bowlofeggs: Cool, then it sounds like this docblock is correct.

bowlofeggs:
test

bowlofeggs (approved):
Thanks!

Issue #1188, opened 2016-12-23
consider to implement bash completion for bodhi-client


bowlofeggs:
[Lukas Slebodnik originally filed this RFE in Bugzilla](https://bugzilla.redhat.com/show_bug.cgi?id=1408205):

> Description of problem:
> The bodhi client 2.0+ has two many nested subcomands. I do not use bodhi very often and therefore I do not remember the correct syntax. It is possible to type --help for each subcomand but it is slow. It is also possible to check syntax in manual page but then I need to copy&paste from man page to a command line. 
>
> Version-Release number of selected component (if applicable):
> sh$ rpm -q bodhi-client
> bodhi-client-2.3.3-1.fc26.noarch
>
> How reproducible:
> Deterministic
> 
> Steps to Reproduce:
> 1. //ensure that you are testing with bash and installed bash-completion
>    bash
> 2. // type bodhi<tab>
> 
> 
> Actual results:
> bash completion try to use local files as parameters which is not very usefull
> 
> Expected results:
> overrides updates
> 
> Additional info:
> $ bodhi updates<tab>
>  comment download new query request
> 
> //the similar output is printed with --help
> sh$ bodhi updates --help
> Usage: bodhi updates [OPTIONS] COMMAND [ARGS]...
>
> Options:
>   --help  Show this message and exit.
>
> Commands:
>   comment
>   download
>   new
>   query
>   request

ryanlerch:
bodhi-client has built in support for bash-completion via click.

currently, if you enable it via this line in your bashrc:

```
eval "$(_BODHI_COMPLETE=source bodhi)"
```

it works. However, when testing on the vagrant install, every completion throws up a bunch of warnings:

```
overrides  updates
[vagrant@bodhi-dev bodhi]$ bodhi ov/usr/lib64/python2.7/pkgutil.py:186: ImportWarning: Not importing directory '/usr/lib64/python2.7/site-packages/zope': missing __init__.py
  file, filename, etc = imp.find_module(subname, path)
/usr/lib64/python2.7/pkgutil.py:186: ImportWarning: Not importing directory '/usr/lib/python2.7/site-packages/zope': missing __init__.py
  file, filename, etc = imp.find_module(subname, path)
/usr/lib64/python2.7/pkgutil.py:186: ImportWarning: Not importing directory '/usr/lib/python2.7/site-packages/paste': missing __init__.py
  file, filename, etc = imp.find_module(subname, path)
[...]

PR #1639, opened 2017-06-21
Add python-click's bash completion activation script

crungehottman:
This commit will allow users to enable bash completion via shipping an activation script. 

To activate the activation script, the user would simply need to place the following line somewhere in his/her/their vagrant .bashrc:
`. ~/bodhi/bodhi-complete.sh`

If the user does not add the previous line to his/her/their .bashrc, no errors will occur, bash completion will simply not be enabled.

For a flawless UI experience, it should be recommended that the user also ignore python warnings by changing the following line in the .bashrc

`export PYTHONWARNINGS="once"
`

to

`export PYTHONWARNINGS="ignore"
`

Example:
image

_Please note that this PR does not intend to provide total and dynamic bash completion, as python-click has yet to implement the ability of dynamic completion._

Examples of Bodhi using python-click's available bash completion:

image

image

source: http://click.pocoo.org/5/bashcomplete/

### DIFF
--- a/bodhi-complete.sh
+++ b/bodhi-complete.sh
@@ -1,0 +1,8 @@
+_bodhi_completion() {
+    COMPREPLY=( $( env COMP_WORDS="${COMP_WORDS[*]}" \
+                   COMP_CWORD=$COMP_CWORD \
+                   _BODHI_COMPLETE=complete $1 ) )
+    return 0
+}
+
+complete -F _bodhi_completion -o default bodhi;
--- a/devel/ansible/roles/dev/tasks/main.yml
+++ b/devel/ansible/roles/dev/tasks/main.yml
@@ -178,3 +178,9 @@
       name: fedmsg-tail
       state: started
       enabled: yes
+
+- name: Link the bodhi bash completion script
+  file:
+    src: /home/vagrant/bodhi/bodhi-complete.sh
+    dest: /etc/bash_completion.d/bodhi-complete.sh
+    state: link


ryanlerch:
FYI - here is the open issue for adding bash completion to the bodhi-cli -- #1188

bowlofeggs:
@crungehottman Can you amend your commit message to have it's own line with ```fixes #1188```?

bowlofeggs (changes requested):
The script looks good (I guess that's just what click autogenerates.) How about we modify the dev role to drop a symlink to this file in ```/etc/bash_completion.d/``` on the Vagrant box? You could add that in [this file](https://github.com/fedora-infra/bodhi/blob/2.8.1/devel/ansible/roles/dev/tasks/main.yml).

I personally think the warnings are good in the dev environment, even though they kinda mess up the bash completion. There may be a way to set PYTHONWARNINGS to the default in the dev environment when bash completion is being used. I don't immediately have an idea for that and it's not a requirement to merge this patch, just a thought.

Once this is in a release, I can modify the downstream spec files to drop this file into the rpms so they work in Fedora and EPEL!

crungehottman:
I agree re: thinking warnings are good in the dev environment. In my limited experience with bodhi, I have only encountered python warnings while using btest and bash completion. Besides during btest, when else might someone want python warnings?

I ask because I found a simple way to enable warnings in btest and disable them otherwise:
image
image

bowlofeggs:
On Fri, 2017-06-23 at 12:22 -0700, Caleigh Runge-Hottman wrote:
> I agree re: thinking warnings are good in the dev environment. In my
> limited experience with bodhi, I have only encountered python
> warnings while using btest and bash completion. Besides during btest,
> when else might someone want python warnings?

That's a good point - so we could consider only setting PYTHONWARNINGS
on certain commands. I'd say it's probably useful in the server logs
(which is run in systemd, and I *think* I might have also set it there?
Can't remember, but it'd be easy to do if I didn't…), btest as you
said, and probably fedmsg-hub. So yeah, if you want to make it so I say
go 4 it.

bowlofeggs (approved):
Thanks!

PR #1643, opened 2017-06-22
Archive the CI results so they can be analyzed by Jenkies.

bowlofeggs:


jeremycline (approved):
:shipit:

PR #1645, opened 2017-06-23
added option (default=false) to email subject beautifier to include nvr

### LINKED ISSUE #1635: Put versions back into e-mail subjects
In https://github.com/fedora-infra/bodhi/pull/1561 we made it so update e-mail subjects use the new beautify title code (which is very nice), but inadvertently we also stopped including version info on the packages and now only list the package names. It has been requested that we bring back the versions into e-mail subjects, so we should modify the new beautify title to have an ```nvr``` flag (defaulting to ```False```) that includes the ```nvr``` of the Build if ```True```, and uses the current behavior if ```False```.

brandongray:
fixes #1635

### DIFF
--- a/bodhi/server/models.py
+++ b/bodhi/server/models.py
@@ -2290,8 +2290,8 @@ class Update(Base):
             since = self.last_modified.isoformat().rsplit('.', 1)[0]
         except Exception as e:
             log.exception("Failed to determine last_modified from %r : %r",
-                          self.last_modified, e.message)
-            return False, "Failed to determine last_modified: %r" % e.message
+                          self.last_modified, str(e))
+            return False, "Failed to determine last_modified: %r" % str(e)
 
         try:
             # query results for this update
@@ -2325,8 +2325,8 @@ class Update(Base):
                 results.extend(build_results)
 
         except Exception as e:
-            log.exception("Failed retrieving requirements results: %r", e.message)
-            return False, "Failed retrieving requirements results: %r" % e.message
+            log.exception("Failed retrieving requirements results: %r", str(e))
+            return False, "Failed retrieving requirements results: %r" % str(e)
 
         for testcase in requirements:
             relevant = [result for result in results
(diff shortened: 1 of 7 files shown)


bodhi/server/models.py
```diff
@@ -1653,7 +1653,11 @@ def beautify_title(self, amp=False):
         &amp; html entity
         """
```
bowlofeggs: Let's document the new nvr parameter in this docblock.

bodhi/server/models.py
```diff
@@ -1653,7 +1653,11 @@ def beautify_title(self, amp=False):
...
         if len(self.builds) > 2:
```
bowlofeggs: You could simplify this a bit by introducing an inline function for how to pick the label to do a little less if/elsing. Something like this:

```
def build_label():
    return build.nvr if nvr else packagename_from_nvr(self, build.nvr)
```

Then you can just change the calls to ```packagename_from_nvr``` to ```build_label```.

It's fine to keep it the way it is if you prefer, just a suggestion.

bodhi/tests/server/test_models.py
```diff
@@ -1078,15 +1078,21 @@ def test_beautify_title(self):
...
 
         update.builds.append(rpm_build)
         eq_(update.beautify_title(), u'TurboGears, TurboGears, and 1 more')
+        eq_(update.beautify_title(nvr=True),
+            u'TurboGears-1.0.8-3.fc11, TurboGears-1.0.8-3.fc11, and 1 more')
 
         p = HTMLParser()
         eq_(p.unescape(update.beautify_title(amp=True)), u'TurboGears, TurboGears, & 1 more')
+        eq_(p.unescape(update.beautify_title(amp=True, nvr=True)),
+            u'TurboGears-1.0.8-3.fc11, TurboGears-1.0.8-3.fc11, & 1 more')
```
bowlofeggs, line 1095: Nice!

bowlofeggs (changes requested):
The only required change to merge is to document the new parameter in the docblock. Thanks for the patch!

bodhi/server/models.py
brandongray: I like this idea!

brandongray:
Ah, rebased wrong! I'm going to close this PR and try again.

PR #1646, opened 2017-06-23
added option (default=false) to email subject beautifier to include nvr

### LINKED ISSUE #1635: Put versions back into e-mail subjects

brandongray:
Let's try this again - fixes #1635

### DIFF
--- a/bodhi/server/models.py
+++ b/bodhi/server/models.py
@@ -1640,7 +1640,7 @@ class Update(Base):
                         bad += 1
         return bad * -1, good
 
-    def beautify_title(self, amp=False):
+    def beautify_title(self, amp=False, nvr=False):
         """Will shorten the title according to its length.
 
         This is used mostly in subject of a update notification email and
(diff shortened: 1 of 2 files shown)


bowlofeggs:
In the future, please don't open new pull requests for the same code - this makes it harder to find the conversation about the change. The history of a pull request is important to preserve and it's much simpler to keep it in one place. You can use ```git commit --amend``` to amend an existing commit and ```git push -f origin <branch_name>``` to force push the amended commit to your existing branch in your fork. This will update the existing pull request, while preserving the comments and history.

For the record, this pull request started out as https://github.com/fedora-infra/bodhi/pull/1645.

bowlofeggs:
test

PR #1647, opened 2017-06-23
Change a heading in the release notes to be the right level.

bowlofeggs:
A Bugs heading in the release notes as at the same level as the
versions and so was rendering next to the versions in the table of
contents.


pypingou:
Sure 👍

PR #1653, opened 2017-06-27
fix showing taskotron results with scenario

kparal:
When two results used the same scenario key, only one of them got
displayed in Bodhi web UI. This fixes that and shows all such results.

It also sorts the results by name as the first sorting key, which
massively improves readability of the web UI.

-----

You can verify the fix e.g. with:
https://bodhi.fedoraproject.org/updates/FEDORA-2017-46e3c100fe
(depcheck x86_64 is not shown, even though present in the retrieved data)
or
https://bodhi.fedoraproject.org/updates/FEDORA-2017-eba47499b6
(rpmdeplint x86_64 is not shown, even though present in the retrieved data)

You can see the sorting improvements here:
https://bodhi.fedoraproject.org/updates/FEDORA-2017-1448fa492e
going from 
1-fullpage
to
2-fullpage
So much better!

CC @AdamWill since I'm adjusting his code.

### DIFF
--- a/bodhi/server/templates/update.html
+++ b/bodhi/server/templates/update.html
@@ -103,7 +103,7 @@ ${update.type} update for ${update.beautify_title(amp=True) | n}
       // the latest ones for each arch.  So, prune like this:
       $.each(data.data, function(i, result) {
 
-        var name = result.testcase.name;
+        var testcase = result.testcase.name;
         var scenario = 'no_scenario';
         if (result.data.scenario) {
             scenario = result.data.scenario[0];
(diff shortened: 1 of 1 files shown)


bowlofeggs (dismissed):
Looks reasonable to me. @AdamWill, would you like to take a look too?

bowlofeggs:
jenkies test

bodhi/server/templates/update.html
```diff
@@ -104,9 +104,7 @@
       $.each(data.data, function(i, result) {
 
         var name = result.testcase.name;
-        // we want to stash results by scenario when there is one, use name
-        // when there isn't
-        var scenario = result.testcase.name;
+        var scenario = 'no_scenario';
```
AdamWill: edit: nevermind, saw the change below

AdamWill:
well, in my initial conception of 'scenario', the test name is *part* of the scenario, so this should never arise. But I don't think I see any downside to doing it this way 'just in case', so I think it's fine.

kparal:
> well, in my initial conception of 'scenario', the test name is part of the scenario, so this should never arise. But I don't think I see any downside to doing it this way 'just in case', so I think it's fine.

We seem to be going in circles :-) Yes, your initial concept worked like that, and I explained why it I believe it's not a good idea to rely on it, and IIRC we agreed on it being just another key to differentiate, but with no internal meaning. I.e. the testcase+type combination is always the primary identification, and you can rely on that. Scenario can extend it, but doesn't replace it.

We still need to finalize scenario concept and handling. But I don't intend to add unnecessary data into scenario for our tasks (e.g. copying the whole testcase name into it), so we definitely need to always take testcase name into account. I'm quite sure future changes will be needed (for example, with the current implementation, we can't distinguish results with and without scenario), but for this moment, it fixes displaying of our tasks as they're currently written.

I added another commit with a better variable name, so that it's clearer what we do there.

kparal:
If there are no more concerns let's get this pushed ;-)

bowlofeggs (approved):

PR #1658, opened 2017-06-28
Fix issues with the search

### LINKED ISSUE #436: Search box should display a placeholder while the package search is running
The Bodhi 2.0 web UI search box searches 'as you type', but the package search takes rather a long time and there's no indication at all that it's happening. It should probably have a 'searching for updates...' placeholder or something to make it clear that if you wait, results will show up.

### LINKED ISSUE #461: Cannot middle-click search result links
When you run a search in the Bodhi 2 web UI, you can't middle-click a result to open it in a new tab (at least, not in Firefox you can't). This is annoying, because sometimes I want to open a search result without navigating away from the Bodhi page I have open in the current tab (e.g.: while editing the description for one update, I want to open another update for reference).

### LINKED ISSUE #1015: Search does not work when package name has special characters
[Andre Robatino reported this issue](https://bugzilla.redhat.com/show_bug.cgi?id=1304119):

> Description of problem:
> If I do a search for "memtest86+" at https://bodhi.fedoraproject.org , no results are returned, even
> though memtest86+ is listed as one of the matching packages before entering. The search result
> should include https://bodhi.fedoraproject.org/updates/FEDORA-2016-1a840cc07e , for example.
> 
> How reproducible:
> always
> 
> Steps to Reproduce:
> 1. Search for "memtest86+".
> 
> Actual results:
> No results.
> 
> Expected results:
> At least one result.

ryanlerch:
This PR contains 4 commits. 

First, updates the typeahead.js library to the latest upstream version 0.11.1. We need the newer version of this library for the 3 fixes in this PR. The typeahead API changed between the 0.10.x release we were using before, so this commit also updates our code to work with the new API.

Secondly, a spinner now shows in the search-results dropdown when we are fetching results from the bodhi server. Each of the 4 results sections has their own spinner. Fixes #436

Thirdly, middle clicking on search results now opens the result in a new tab as expected. This was acheived by wrapping all search results in an a element. Fixes #461

Finally, we escape the packagename that we pass as a URI for cases like memtest86+ where the package name includes a special character that should not be in the URI. Fixes #1015

### DIFF
--- a/bodhi/server/static/css/site.css
+++ b/bodhi/server/static/css/site.css
@@ -154,6 +154,15 @@ form#search, form#search input, .twitter-typeahead {
   cursor: pointer;
 }
 
+.tt-suggestion.tt-cursor a, .tt-suggestion:hover a{
+  color:#fff;
+}
+
+.tt-suggestion a{
+  color:black;
+  display:block;
+}
+
 .tt-suggestion p {
   margin: 0;
 }
--- a/bodhi/server/static/css/site.css
+++ b/bodhi/server/static/css/site.css
@@ -123,7 +123,7 @@ form#search, form#search input, .twitter-typeahead {
   color: #999
 }
 
-.tt-dropdown-menu {
+.tt-menu {
   width: 100%;
   margin-top: 0px;
   padding: 8px;
@@ -148,10 +148,10 @@ form#search, form#search input, .twitter-typeahead {
   line-height: 18px;
 }
 
-.tt-suggestion.tt-cursor {
+.tt-suggestion.tt-cursor, .tt-suggestion:hover {
   color: #fff;
   background-color: #0097cf;
-
+  cursor: pointer;
 }
 
 .tt-suggestion p {
--- a/bodhi/server/static/js/search.js
+++ b/bodhi/server/static/js/search.js
@@ -50,7 +50,7 @@ $(document).ready(function() {
       } else if (data.title != undefined ) {
           return '/updates/' + data.title;
       } else if (data.hasOwnProperty('stack')) {
-          return'/updates/?packages=' + data.name;
+          return'/updates/?packages=' + encodeURIComponent(data.name);
       } else if (data.name != undefined) {
           return '/users/' + data.name;
       } else if (data.nvr != undefined) {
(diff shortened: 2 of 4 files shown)


bowlofeggs:
jenkies test

bowlofeggs (changes requested):
Please write a release note mentioning the new version of the typeahead library so I remember to update the spec file to declare that new version.

I did notice that searching for "memtest86+" did fix the package search, but the update search still shows no results. Additionally, even "memtest8" doesn't find update results, but "memtest" does find the memtest86+ results. On production, "memtest8" does find the updates, so I think this might be a small regression.

ryanlerch:
@bowlofeggs after a lot of digging, finally figured out that your issue with 'memtest8' as a search was actually a bug in 0.11.1 of typeahead.js. However, this is where this gets messy, it appears that the main typeahead.js (https://github.com/twitter/typeahead.js) is unmaintained. There are a list of bugs on this issue (https://github.com/twitter/typeahead.js/issues/1381) of people having similar issues. 

There is however, a maintained fork of typeahead,js that contains the fix for the issue we were having: https://github.com/corejavascript/typeahead.js . In this PR i am now using their 1.1.1 release which solves the issue.

ryanlerch:
Also have updated the release notes with a note about the updated typeahead.js library.

bowlofeggs:
jenkies test

bowlofeggs (dismissed):
Cool, this looks good to merge!

I did notice one remaining issue, but it's not introduced by this change (so it can be fixed separately if you prefer): If you type "memtest86+" and hit enter (without scrolling down to the search results), it still takes you to a URL with an unescaped "+" in it, which doesn't work. Again, this is how it is in production too, so we can fix it later if you prefer. Feel free to merge this as you please!

ryanlerch:
@bowlofeggs thanks for catching this one. Just added an extra commit to this PR to fix this. 

the search.js script has a special method that catches if the user presses enter on the search bar without choosing an item from the typeahead dropdown, and just searches for the package name. WE previously escaped the search value for items chosen in the dropdown, but missed the special case when the user presses enter. This escapes that value, so results for packagenames with special URI chars in them (e.g. memtest86+) now show properly when just typing a search string and pressing enter.

bowlofeggs (approved):
Looks great, thanks @ryanlerch!

PR #1664, opened 2017-06-29
Convert Bodhi to pytest as the test runner

jeremycline:
This removes the dependency on the abandoned nose project and switches
the runner to pytest. It also fixes the tests that broke due to run
order changes. Configuration has been added to a new tox.ini file.


devel/ci/run_tests.sh
```diff
@@ -59,6 +59,7 @@ mv development.ini.example development.ini
 
 sudo /usr/bin/python setup.py develop
 
-/usr/bin/python setup.py nosetests
+/usr/bin/pytest
+/usr/bin/tox -e flake8,pydocstyle
```
bowlofeggs: It might be nice to swap the order of these two, since the style tests run more quickly - we could give the user faster feedback on style errors. Just a suggestion.

tox.ini
```diff
@@ -0,0 +1,67 @@
...
+whitelist_externals =
+    mkdir
+    rm
+    sphinx-build
+commands=
+    rm -rf _build images
+    mkdir -p _static images
+    ./generate_db_schema
+    sphinx-build -W -b html -d {envtmpdir}/doctrees .  _build/html
+    sphinx-build -W -b man -d {envtmpdir}/doctrees .  _build/man
```
bowlofeggs, line 27: Very nice! You could add docs/_build/* as a Jenkies artifact too if you want. That would go here:

https://github.com/fedora-infra/bodhi/blob/93fad0c227539bfb6a759d7bcca8bb726619353a/devel/ci/githubprb-project.yml#L47-L48

If you do that, you'll also need to rsync them back after the tests run. You'd do that here:

https://github.com/fedora-infra/bodhi/blob/93fad0c227539bfb6a759d7bcca8bb726619353a/devel/ci/githubprb-project.yml#L88-L89

Completely optional for this pull request, but I'll probably do it if you don't because it seems like it'd be nice ☺

bowlofeggs: I'm also assuming that it'll be possible to view the docs in Jenkies. It might not actually render though, in which case it might not be that valuable. I was just thinking that it could be cool to see what a docs PR would look like without having to check out and build it myself. Pie in the sky.

tox.ini
```diff
@@ -0,0 +1,67 @@
...
+inherit = false
+ignore = D413
+match =
+    bodhi/server/views/*.py
+    bodhi/server/consumers/__init__.py
+    bodhi/server/consumers/signed.py
+    bodhi/server/consumers/updates.py
+    bodhi/server/scripts/
+    bodhi/server/services/stacks.py
+    bodhi/server/views/
```
bowlofeggs, line 56: Do we need this and also the ```bodhi/server/views/*.py``` line? If so, what about ```scripts```?

tox.ini
```diff
@@ -0,0 +1,67 @@
...
+
+[flake8]
+show-source = True
+max-line-length = 100
+exclude = .git,.tox,dist,*egg
+ignore = E712
+
+
+[pytest]
+addopts = --junit-xml=nosetests.xml --cov-config .coveragerc --cov=bodhi --cov-report term --cov-report xml --cov-report html
```
bowlofeggs, line 67: You could rename ```nosetests.xml``` to whatever you like (maybe just ```test_results.xml```), but it'll require coordination of the JJB service job (which only runs when things are merged to develop) and changing this. Thus, I recommend making a separate PR after this one is merged that does that and only that, because any PR that makes that change is guaranteed to fail CI. We can just make an exception. We can also just leave it named this way forever too.

bowlofeggs (changes requested):
Looks very nice!

However, btest fails on my Vagrant box:

```
flake8 runtests: PYTHONHASHSEED='3920989412'
flake8 runtests: commands[0] | python -m flake8
./build/lib/bodhi/server/util.py:230:1: E305 expected 2 blank lines after class or function definition, found 1
hardcoded_avatars = {
^
./build/lib/bodhi/server/__init__.py:122:1: E305 expected 2 blank lines after class or function definition, found 1
DEFAULT_FILTERS.insert(0, exception_filter)
^
ERROR: InvocationError: '/home/vagrant/bodhi/.tox/flake8/bin/python -m flake8'
```

tox.ini
jeremycline: I was having trouble with the docs environment in tox, but I'm happy to poke at that in a separate PR

devel/ci/run_tests.sh
jeremycline: Done

tox.ini
jeremycline: Nope, I'm not sure how the ``*.py`` line slipped in.

tox.ini
jeremycline: Yeah, I tried that initially and realized it would cause the tests to fail.

jeremycline:
It looks like you've got a build directory with stale files (which is causing the flake8 problem). I've added build/ to the ignore list for flake8.

bowlofeggs:
@jeremycline oh my, hahaha I didn't even notice the paths. Sorry, my mistake.

bowlofeggs (approved):
This is really great, thanks so much @jeremycline!

PR #1668, opened 2017-07-03
Unify the base test classes

jeremycline:
This unifies the base test classes by making a single base
class. This base class sets up the test database with nested
transactions so that it's simple to roll the changes a test makes back
at the end of the test. It also creates a /single/ TestApp for tests to
use.

Creating a TestApp leaks a substantial amount of memory for each test
case. Tests that require a different configuration should create their
own TestApp (although this is costly as it still leaks memory at this
time).


bodhi/tests/server/base.py
```diff
@@ -19,70 +19,198 @@
...
 
 
 original_config = config.config.copy()
+engine = None
+_app = None
 
-DB_PATH = 'sqlite://'
-DB_NAME = None
 
+def _configure_test_db(db_uri='sqlite:////var/tmp/bodhi-tests.sqlite'):
```
bowlofeggs: I think this may fail to run in Koji. Should we make it a relative path to the source code instead so it will run on systems that restrict paths that can be written? Another option might be to make it a setting in development.ini that the spec file could adjust as needed, and it can default to this value. I could be wrong - maybe it'll be just fine on Koji as is…

bodhi/tests/server/base.py
```diff
@@ -19,70 +19,198 @@
...
-        req = requests.get('%s/new' % FAITOUT, timeout=60)
-        if req.status_code == 200:
-            DB_PATH = req.text
-            DB_NAME = DB_PATH.rsplit('/', 1)[1]
-            print 'Using faitout at: %s' % DB_PATH
-    except:
-        pass
+    Args:
+        db_uri (str): The URI to use when creating the database engine. Defaults to an
+            in-memory SQLite database.
```
bowlofeggs, line 47: It defaults to a FS path SQLite db now…

bodhi/tests/server/base.py
```diff
@@ -19,70 +19,198 @@
...
+    slate for each test. Tests may call both ``commit`` and ``rollback``
+    on the database session they acquire from ``bodhi.server.Session``.
+    """
     _populate_db = True
 
-    def setUp(self):
+    app_settings = {
+        'authtkt.secret': 'sssshhhhhh',
+        'authtkt.secure': False,
+        'sqlalchemy.url': 'sqlite:////var/tmp/bodhi-tests.sqlite',
```
bowlofeggs: It might be good to make this one configurable or relative to the source checkout too.

bowlofeggs (dismissed):
This is great! I saw a peak total system memory usage of 863 MB with a background memory usage of 452 MB when the tests weren't running, so the tests now use about 411 MB of memory where they used to use over 4 GB.

bodhi/tests/server/base.py
jeremycline: We could definitely make it relative to the source and that would be more portable. I wonder if pytest has some hook at the end of the tests to delete it... Nothing except the schema is written to the database, but it'd still be nice to be neat.

bodhi/tests/server/base.py
jeremycline: Remnants of my aspirations :(

bodhi/tests/server/base.py
jeremycline: I just copied this from the old functional test class - it turns out it's not necessary and I've removed it.

jeremycline:
It looks like it might be possible to add a one-time cleanup to remove the database after the tests run, but I'm going to leave that for a later time since nothing is written to the database and it's removed during test initialization.

The default database is now created in the repository root rather than ``/var/tmp/``

(sorry for amending such a large commit, here's the diff):
```
diff --git a/bodhi/tests/server/base.py b/bodhi/tests/server/base.py
index a54204e..19209b2 100644
--- a/bodhi/tests/server/base.py
+++ b/bodhi/tests/server/base.py
@@ -30,9 +30,11 @@ from bodhi.tests.server import create_update, populate
 original_config = config.config.copy()
 engine = None
 _app = None
+DEFAULT_DB = 'sqlite:///' + os.path.abspath(
+    os.path.join(os.path.dirname(__file__), '../../../bodhi-tests.sqlite'))
 
 
-def _configure_test_db(db_uri='sqlite:////var/tmp/bodhi-tests.sqlite'):
+def _configure_test_db(db_uri=DEFAULT_DB):
     """
     Creates and configures a test database for Bodhi.
 
@@ -93,7 +95,6 @@ class BaseTestCase(unittest.TestCase):
     app_settings = {
         'authtkt.secret': 'sssshhhhhh',
         'authtkt.secure': False,
-        'sqlalchemy.url': 'sqlite:////var/tmp/bodhi-tests.sqlite',
         'mako.directories': 'bodhi:server/templates',
         'session.type': 'memory',
         'session.key': 'testing',
```

bowlofeggs (approved):
Cool, this looks great. It'd be nice to get that pytest cleanup thing running (I know nose can do that too), but I wouldn't say it's a requirement to merge. Plenty of other files get created as you develop Bodhi so it's not so different from those (pycs, the stuff from setup.py develop, the test xml files, etc.)

bowlofeggs:
I filed https://github.com/fedora-infra/bodhi/pull/1671 to lower the default RAM in the Vagrantfile in response to this awesome refactor.

PR #1669, opened 2017-07-05
Fixes #1304 calls clean_up after mash is done

### LINKED ISSUE #1304: The masher should clean up old mashes at the end of every run

kushaldas:
Also have assert statement to make sure that clean_up is called.

### DIFF
--- a/bodhi/server/consumers/masher.py
+++ b/bodhi/server/consumers/masher.py
@@ -43,6 +43,7 @@ from bodhi.server.metadata import ExtendedMetadata
 from bodhi.server.models import (Update, UpdateRequest, UpdateType, Release,
                                  UpdateStatus, ReleaseState, Base)
 from bodhi.server.util import sorted_updates, sanity_check_repodata, transactional_session_maker
+from bodhi.server.scripts.clean_old_mashes import clean_up
 
 
 def checkpoint(method):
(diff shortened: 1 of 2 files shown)


ryanlerch:
jenkies test

kushaldas:
Broke too many things :( Have to fix my testing environment first.

bowlofeggs (changes requested):
Can you adjust your commit message so that "fixes #1304" is on its own line? That way GitHub will automatically close the issue when this is merged. Something like this would do:

```
Call clean_up() after mash is done.

fixes #1304
```

bowlofeggs:
@kushaldas Do you still wish to work on this?

bowlofeggs:
I'm going to close this for now due to inactivity. Feel free to reopen if you do still want to work on this!

PR #1676, opened 2017-07-10
[WIP] [DO NOT MERGE] Adds usage of pungi for mashing

kushaldas:
Now to actually use this, we will require a few more changes.

* Release Engineering should first write the set of punig config files which we can use for  this use case
* The bodhi config should have `pungi_config` variable, which should point to the right location of the config file based on the current tag we are mashing.
* The bodhi config should also have a new variable `use_pungi_in_bodhi`, if only this is true, then we will use pungi, or else we can safely use the old code.

I have already asked for the pungi configs at https://pagure.io/releng/issue/6893 from rel-eng.

According to the lsedlar, we will also have to pass  `--label=somevalue` or `--no-label` to the command line. But, I will need more guidance about that argument.

The current patch will not remove any old code, we will do that only after everything is well tested with pungi.

### DIFF
--- a/bodhi/server/consumers/masher.py
+++ b/bodhi/server/consumers/masher.py
@@ -333,9 +333,11 @@ class MasherThread(threading.Thread):
                 uinfo.insert_pkgtags()
                 uinfo.cache_repodata()
 
-            # Compose OSTrees from our freshly mashed repos
-            if config.get('compose_atomic_trees'):
-                self.compose_atomic_trees()
+            # Do the following if we are not using pungi
+            if not config.get('use_pungi_in_bodhi'):
+                # Compose OSTrees from our freshly mashed repos
+                if config.get('compose_atomic_trees'):
+                    self.compose_atomic_trees()
 
             if not self.skip_mash:
                 self.sanity_check_repo()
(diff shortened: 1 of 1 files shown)


bodhi/server/consumers/masher.py
```diff
@@ -1027,12 +1029,21 @@ def __init__(self, tag, outputdir, comps, previous, log):
...
+
+            mash_cmd = 'mash -o {outputdir} -c {config} -f {compsfile} {tag}'
+            mash_conf = config.get('mash_conf')
+            if os.path.exists(previous):
+                mash_cmd += ' -p {}'.format(previous)
+            self.mash_cmd = mash_cmd.format(outputdir=outputdir, config=mash_conf,
+                                            compsfile=comps, tag=self.tag).split()
+
+        else:  # We are going to use pungi in this run
+            pungi_cmd = "pungi-koji  --config={config} --old-composes={outputdir}"
```
ralphbean, line 1042: FYI, @mcurlej has done some work on a python api for pungi so you can `import pungi` and call python functions, instead of shelling out to a subprocess like this.

bowlofeggs:
On Mon, 2017-07-10 at 02:43 -0700, Kushal Das wrote:
> The bodhi config should have pungi_config variable, which should
> point to the right location of the config file based on the current
> tag we are mashing.
> The bodhi config should also have a new variable use_pungi_in_bodhi,
> if only this is true, then we will use pungi, or else we can safely
> use the old code.

You could simplify a bit by using pungi if there is a pungi_config
defined for the given tag that is being mashed, and fall back to the
old way if it isn't defined.

bodhi/server/consumers/masher.py
puiterwijk: @ralphbean One major advantage of shelling out is that we can continue to create the updateinfo.xml, since otherwise you would be getting into the GIL.

lubomir:
> According to lsedlar, we will also have to pass --label=somevalue or --no-label to the command line. But, I will need more guidance about that argument.

Pungi has a concept of a *compose type*, which indicates what kind of compose it is. The options are *production*, *nightly*, *test* and *ci*. For the use case in Bodhi production seems to be best match. All production composes should have a label which indicates a milestone for the compose (Alpha, Beta, RC etc.). Alternatively you can add the `--no-label` option to have a production compose without a label. That would probably be a good match here.

---

As for the PR itself, I'm fairly sure the structure generated by Pungi will not match what is generated by Mash. The actual repos will be created under `{outputdir}/<COMPOSE_ID>/compose/<VARIANT>/$arch/os/` (with `repodata` and `Packages` subdirectories there). `VARIANT` will be a static string determined by the configuration. `COMPOSE_ID` is partly determined by configuration, the rest is current date and a number to make it unique.

The `stage_repo` method looks like a good place to handle this.

bowlofeggs:
What's the status on this PR?

igor-raits:
any news here? I would like to get bodhi to use pungi ASAP

bowlofeggs:
I am closing this due to inactivity. Feel free to reopen if you still want to work on it!

dustymabe:
I've been working on this - it's been slow going because of conference and PTO. Will open a new PR next week.

PR #1681, opened 2017-07-12
Fix misc spelling errors

crungehottman:


sayanchowdhury (approved):
:+1:

bodhi/server/consumers/updates.py
```diff
@@ -199,4 +199,4 @@ def work_on_bugs(self, session, update, bugs):
                 log.info("Modifying %r" % bug.bug_id)
                 bug.modified(update)
             except Exception:
-                log.warning('Error occured during updating single bug', exc_info=True)
+                log.warning('Error occurred during updating single bug', exc_info=True)
```
bowlofeggs, line 202: The CI test failed due to there not being coverage on this line. Since it's a spelling error I'd be slightly OK with merging it, but my spider senses say we should add a test to cover this line since it is a real line of code. I'm happy to write a test to hit this line for you if you want, just let me know!

bowlofeggs (approved):
Fantastic, thanks so much for taking care of that test. Now we have even higher test coverage. 100% test coverage is on the horizon!

Issue #1682, opened 2017-07-12
500 Server Error when enabling CI


matt8754:
If ci.required is set True, visiting one of the existing update will occur 500 Server Error

500 Server Error

Traceback (most recent call last): File "/usr/lib/python2.7/site-packages/pyramid_mako/__init__.py", line 148, in __call__ result = template.render_unicode(**system) File "/usr/lib/python2.7/site-packages/mako/template.py", line 471, in render_unicode as_unicode=True) File "/usr/lib/python2.7/site-packages/mako/runtime.py", line 838, in _render **_kwargs_for_callable(callable_, data)) File "/usr/lib/python2.7/site-packages/mako/runtime.py", line 873, in _render_context _exec_template(inherit, lclcontext, args=args, kwargs=kwargs) File "/usr/lib/python2.7/site-packages/mako/runtime.py", line 899, in _exec_template callable_(context, *args, **kwargs) File "/home/vagrant/bodhi/bodhi/server/templates/master.html", line 185, in render_body ${self.body()} File "/home/vagrant/bodhi/bodhi/server/templates/update.html", line 700, in render_body ${self.util.ci_status2html(update.ci_status.description) | n} AttributeError: 'NoneType' object has no attribute 'description'

This is because the ci_status could be None and this case is not covered on the UI.

pypingou:
I got the patch for this

matt8754:
@pypingou , you are so quick, I was about to write a patch for this, :-)

pypingou:
I had it from before, it was just laying on my git repo and I was reminding myself for a few days that I should submit it, you're just the trigger I needed :)

matt8754:
Nice, 👍

bowlofeggs:
In https://github.com/fedora-infra/bodhi/pull/1733 I am removing the ```ci.enabled``` setting. Thus, I believe we can close this and the associated PR.

PR #1683, opened 2017-07-12
Check if there is a status before trying to access its description

pypingou:
Fixes https://github.com/fedora-infra/bodhi/issues/1682

### DIFF
--- a/bodhi/server/templates/update.html
+++ b/bodhi/server/templates/update.html
@@ -697,7 +697,8 @@ $(document).ready(function(){
                   % endif
                 </div>
                 <div>
-                  ${self.util.ci_status2html(update.ci_status.description) | n}
+                  ${self.util.ci_status2html(update.ci_status.description
+                    if update.ci_status else None) | n}
                 </div>
               </div>
               % endif
@@ -884,7 +885,8 @@ $(document).ready(function(){
               % if build.ci_url:
                 <a href="${build.ci_url}">${self.util.ci_status2html(build.ci_status.description) | n}</a>
               % else:
-                ${self.util.ci_status2html(build.ci_status.description) | n}
+                ${self.util.ci_status2html(build.ci_status.description
+                  if build.ci_status else None) | n}
               % endif
             </td>
           % endif


pypingou:
I know it needs unit-tests but at least this way @matt8754 can keep working :)

sayanchowdhury (approved):

bodhi/server/templates/update.html
```diff
@@ -884,7 +885,8 @@ <h5 class="m-t-2">
               % if build.ci_url:
                 <a href="${build.ci_url}">${self.util.ci_status2html(build.ci_status.description) | n}</a>
               % else:
-                ${self.util.ci_status2html(build.ci_status.description) | n}
+                ${self.util.ci_status2html(build.ci_status.description
+                  if build.ci_status else None) | n}
```
bowlofeggs, line 889: Another way that could be nice to do it may be to alter ```ci_status2html()``` so that it accepts the status and knows to use the description attribute if it is not None. This is a fine way to do it too though.

bowlofeggs (changes requested):
Do you plan to write tests for this change, or do you need me to do it?

bodhi/server/templates/update.html
pypingou: Would you like me to change approach?

Btw, if you have time for the tests, go for it, I'm not sure when I'll find some

pypingou:
Looks like I answered your question while replying to your comment ^^

bowlofeggs:
On Wed, 2017-07-12 at 23:00 +0000, Pierre-Yves Chibon wrote:
> Would you like me to change approach?
> Btw, if you have time for the tests, go for it, I'm not sure when
> I'll find some

I'm willing to write tests. Would you object to me changing the
approach when I write tests? I'll probably just make a commit on top of
yours so you still get credit for your patch, if that works for you.

pypingou:
>I'm willing to write tests. Would you object to me changing the approach when I write tests? 

Go right ahead, thanks! :)

bowlofeggs:
In #1733 I am removing the ci.enabled setting. Thus, I believe we can close this and the associated issue.

PR #1684, opened 2017-07-12
PEP-257 bodhi.server.services.updates.

bowlofeggs:


sayanchowdhury (approved):
:+1:

PR #1686, opened 2017-07-12
Recreate the flake8 env to avoid webob sadness

jeremycline:
tox would fail to install bodhi on the second run because of
pkg_resources and webob.

@crungehottman this should address your issue with btest.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -28,6 +28,7 @@ commands=
 
 
 [testenv:flake8]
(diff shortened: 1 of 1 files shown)


bowlofeggs (approved):

crungehottman:
:thumbsup: This fixed my btest error! Thank you, @jeremycline !

PR #1687, opened 2017-07-17
Implement new search api

ryanlerch:
As recommended in #997 this implements a new API item 'search' for updates/, packages/, users/, and overrides/. It utilises `ilike` rather than `like` to make the search case insensitive. The search JS is also updated to use the new search API

Additionally, this PR adds the ability to search for an update by update alias (e.g. FEDORA-2017-a3bbe1a8f2) 

Finally, when searching packages, if there is an exact match to the package name, that package is always returned first. So when searching for "python", the "python" package is returned first, then all the other python packages.

### DIFF
--- a/bodhi/server/services/overrides.py
+++ b/bodhi/server/services/overrides.py
@@ -126,6 +126,11 @@ def query_overrides(request):
             RpmBuild.nvr.like('%%%s%%' % like)
         ]))
 
+    search = data.get('search')
+    if search is not None:
+        query = query.join(BuildrootOverride.build)
+        query = query.filter(Build.nvr.ilike('%%%s%%' % search))
+
     submitter = data.get('user')
     if submitter is not None:
         query = query.filter(BuildrootOverride.submitter == submitter)
--- a/bodhi/server/services/packages.py
+++ b/bodhi/server/services/packages.py
@@ -15,6 +15,7 @@ import math
 
 from cornice import Service
 from sqlalchemy import func, distinct
+from sqlalchemy.sql.expression import case
 
 from bodhi.server.models import RpmPackage, Package
 import bodhi.server.schemas
@@ -48,6 +49,7 @@ def query_packages(request):
     search = data.get('search')
     if search is not None:
         query = query.filter(Package.name.ilike('%%%s%%' % search))
+        query = query.order_by(case([(Package.name == search, Package.name)]))
 
     # We can't use ``query.count()`` here because it is naive with respect to
     # all the joins that we're doing above.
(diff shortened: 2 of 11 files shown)


bodhi/server/services/overrides.py
```diff
@@ -126,6 +126,11 @@ def query_overrides(request):
             RpmBuild.nvr.like('%%%s%%' % like)
         ]))
 
+    search = data.get('search')
+    if search is not None:
+        query = query.join(BuildrootOverride.build)
+        query = query.filter(RpmBuild.nvr.ilike('%%%s%%' % search))
```
bowlofeggs: The ```Build``` superclass actually has the ```nvr``` field, so let's use that here so it will also search other types.

bodhi/server/services/packages.py
```diff
@@ -15,13 +15,13 @@
...
 from cornice import Service
 from sqlalchemy import func, distinct
+from sqlalchemy.sql.expression import case
 
 from bodhi.server.models import RpmPackage
 import bodhi.server.schemas
 import bodhi.server.security
 import bodhi.server.services.errors
 
-
```
bowlofeggs: This newline is here for PEP-8 compliance, so we should keep it (2 spaces between top level blocks).

bodhi/server/services/packages.py
```diff
@@ -36,7 +36,6 @@ def query_packages(request):
     db = request.db
     data = request.validated
     query = db.query(RpmPackage)
-
```
bowlofeggs: This newline seems to logically separate the "name" processing block below from the code above this. Since it's not needed for the search change I recommend restoring it.

bodhi/server/services/packages.py
```diff
@@ -45,6 +44,11 @@ def query_packages(request):
     if like is not None:
         query = query.filter(RpmPackage.name.like('%%%s%%' % like))
 
+    search = data.get('search')
+    if search is not None:
+        query = query.filter(RpmPackage.name.ilike('%%%s%%' % search))
+        query = query.order_by(case([(RpmPackage.name == search, RpmPackage.name)]))
```
bowlofeggs: Similarly here, let's use ```Package``` rather than ```RpmPackage```. I see above that other places used ```RpmPackage``` so I'll note that I should come back through here and fix those (I wouldn't recommend fixing them in this commit since they are related to this change, but feel free to fix them in a separate PR if you like!)

bowlofeggs: Also, big +1 to ordering by the name match!

bowlofeggs (changes requested):
Very nice, it seems to work well! Just a few adjustments and we'll be good to merge.

PR #1693, opened 2017-07-19
make the autokarma checkbox on edit update persistent

### LINKED ISSUE #1308: autopush checkbox get enabled on update edit even though it was disabled before
When I create a new update with autopush disabled (the checkbox unticked), and then I edit the update, the autopush checkbox gets re-ticked automatically. That's annoying, because for each update I edit, I need to untick the checkbox every time, so that autopush doesn't get enabled when I don't want it. That's quite error-prone.

This might only happen on first update edit, but not on subsequent ones, I'm not sure.

Please honor the existing update configuration (autopush enabled/disabled) and make the checkbox reflect the current state, when editing an update. Thanks.

### LINKED ISSUE #1482: "Auto-request stable" cannot be disabled once enabled.
Hi,
I created errata and checked "Auto-request stable", next day I changed my mind and wanted to disable this option. So I in WebUI clicked on Edit and unchecked "Auto-request stable" and hit "Save". However the "Auto-request stable" is still enabled.

The temporary workaround is to raise karma treshold to unsane high number so the treshold is never reached.

### LINKED ISSUE #1692: Editing a bodhi update re-enables the auto-push checkbox
If I've turned off auto-push on and update and go to edit it, I notice the auto-push checkbox is checked again. I need to explicitly uncheck it each time, otherwise the update will have auto-push enabled.

I expect the update to keep the auto-push off even with edit.

ryanlerch:
Previously, the autokarma checkbox in the edit update form
was always set to on, regardless of the value set in the DB.
This commit fixes this, so the checkboxis set with the value
in the db for the update.

fixes #1692
fixes #1482
fixes #1308

### DIFF
--- a/bodhi/server/templates/new_update.html
+++ b/bodhi/server/templates/new_update.html
@@ -231,7 +231,7 @@ ${update.notes}
                   <dt data-toggle="tooltip" title="If checked, this option allows bodhi to automatically move your update from testing to stable once enough positive karma has been given.">
                   Auto-request stable?</dt>
                   <dd> <input type="checkbox" name="autokarma" data-singleton="true"
-                      % if update and update.stable_karma != None and update.unstable_karma != None:
+                      % if update and update.autokarma:
                       checked
                       % elif not update:
                       checked


bowlofeggs (approved):
Looks good to me, but I have a suggestion wrt the commit message: I suggest adding this to it:

```
fixes #1482
fixes #1308
```

Technically, I think the three bugs are all duplicates, but we can just mark them all as fixed. Or another option is just to close the others as duplicates. Either way, feel free to merge!

PR #1694, opened 2017-07-19
Greenwave integration

matt8754:
Highlights for this PR:
* enabling Greenwave integration in bodhi will make the application gate on the automated test resutls
* CI integration will no longer make the application gate on CI results
* display ‘Gating Status’ on the update page on the UI and link people to the automated test tab in order to help debug which tests are blocking if the policies are not satisfied.
* a new command is added to get a decision for each opening update(an update is not pushed and the status of the update is either pending or testing). If we have a message-bus driven process to cache Greenwave's decision, we could use this command in a cron job to update the cache frequently in case messages were missed along the way.

### DIFF
--- a/alembic/versions/33a416e03f50_add_greenwave_fields.py
+++ b/alembic/versions/33a416e03f50_add_greenwave_fields.py
@@ -1,0 +1,41 @@
+"""Add the fields for caching the Greenwave decision for each update.
+
+Revision ID: 33a416e03f50
+Revises: 2a10629168e4
+Create Date: 2017-07-06 13:52:50.892504
+"""
+from alembic import op
+from sqlalchemy import Column, Enum, Unicode, UnicodeText
+
+
+# revision identifiers, used by Alembic.
+revision = '33a416e03f50'
+down_revision = '2a10629168e4'
+
+
+def upgrade():
+    op.execute(
+        "CREATE TYPE ck_test_gating_status AS ENUM "
+        "('ignored', 'queued', 'running', 'passed', 'failed', 'waiting')")
+    op.add_column(
+        'updates',
+        Column(
+            'test_gating_status',
+            Enum(
+                'ignored', 'queued', 'running', 'passed', 'failed', 'waiting',
+                name='ck_test_gating_status'),
+            nullable=True
+        )
+    )
+    op.add_column('updates', Column('greenwave_summary_string', Unicode(255)))
+    op.add_column(
+        'updates',
+        Column('test_gating_url', UnicodeText, nullable=True)
+    )
+
+
+def downgrade():
+    op.drop_column('updates', 'test_gating_status')
+    op.execute("DROP TYPE ck_test_gating_status")
+    op.drop_column('updates', 'greenwave_summary_string')
+    op.drop_column('updates', 'test_gating_url')
--- a/bodhi/server/config.py
+++ b/bodhi/server/config.py
@@ -407,6 +407,9 @@ class BodhiConfig(dict):
         'initial_bug_msg': {
             'value': '%s has been submitted as an update to %s. %s',
             'validator': unicode},
+        'greenwave_api_url': {
+            'value': 'https://greenwave.fedoraproject.org/api/v1.0',
+            'validator': unicode},
         'koji_hub': {
             'value': 'https://koji.stg.fedoraproject.org/kojihub',
             'validator': str},
--- a/bodhi/server/config.py
+++ b/bodhi/server/config.py
@@ -547,6 +547,12 @@ class BodhiConfig(dict):
         'top_testers_timeframe': {
             'value': 7,
             'validator': int},
+        'test_gating.required': {
+            'value': False,
+            'validator': _validate_bool},
+        'test_gating.url': {
+            'value': '',
+            'validator': unicode},
         'updateinfo_rights': {
             'value': 'Copyright (C) {} Red Hat, Inc. and others.'.format(datetime.now().year),
             'validator': unicode},
--- a/bodhi/server/models.py
+++ b/bodhi/server/models.py
@@ -1804,6 +1804,8 @@ class Update(Base):
                         (config.get('critpath.min_karma') -
                             config.get('critpath.num_admin_approvals')),
                         config.get('critpath.stable_after_days_without_negative_karma'))
+                    if config.get('test_gating.required'):
+                        stern_note += ' Additionally, it must pass automated tests.'
                     notes.append(stern_note)
 
                     if self.status is UpdateStatus.testing:
--- a/bodhi/server/models.py
+++ b/bodhi/server/models.py
@@ -1341,11 +1341,10 @@ class Update(Base):
         log.debug("Triggering db flush for new update.")
         db.flush()
 
-        if config.get('ci.required'):
+        if config.get('test_gating.required'):
             log.debug(
-                'CI required is enforced, marking the build as waiting on CI')
-            for build in up.builds:
-                build.ci_status = CiStatus.waiting
+                'Test gating required is enforced, marking the update as waiting on test gating')
+            up.test_gating_status = TestGatingStatus.waiting
 
         log.debug("Done with Update.new(...)")
         return up, caveats
@@ -1492,6 +1491,20 @@ class Update(Base):
         """
         if self.builds:
             return self.builds[0].type
+
+    @property
+    def test_gating_passed(self):
+        """
+        Returns a boolean representing if this update has passed the test gating.
+
+        Returns:
+            bool: Returns True if the Update's test_gating_status property is None, ignored,
+                or passed. Otherwise it returns False.
+        """
+        if self.test_gating_status in (
+                None, TestGatingStatus.ignored, TestGatingStatus.passed):
+            return True
+        return False
 
     @property
     def ci_passed(self):
@@ -2378,8 +2391,8 @@ class Update(Base):
                     return False, "Required task %s returned %s" % (
                         latest['testcase']['name'], latest['outcome'])
 
-        if config.get('ci.required') and not self.ci_passed:
-            return False, "CI did not pass on this update"
+        if config.get('test_gating.required') and not self.test_gating_passed:
+            return (False, "Required tests did not pass on this update")
 
         # TODO - check require_bugs and require_testcases also?
 
@@ -2484,7 +2497,7 @@ class Update(Base):
         """
         num_days = self.mandatory_days_in_testing
 
-        if config.get('ci.required') and not self.ci_passed:
+        if config.get('test_gating.required') and not self.test_gating_passed:
             return False
 
         if self.critpath:
--- a/bodhi/server/models.py
+++ b/bodhi/server/models.py
@@ -391,6 +391,25 @@ class CiStatus(DeclEnum):
     failed = 'failed', 'Failed'
 
 
+class TestGatingStatus(DeclEnum):
+    """ This class lists the different status the test_gating_status flag can have. """
+    # None: Updates created while bodhi has not enabled Greenwave integration
+    # waiting: Updates created while bodhi has enabled Greenwave integration
+    waiting = 'waiting', 'Waiting'
+    # ignored: a decision from Greenwave said that this update does not require any tests
+    ignored = 'ignored', 'Ignored'
+    # queued: a decision from Greenwave said that the required tests for this update have been
+    # queued
+    queued = 'queued', 'Queued'
+    # running: a decision from Greenwave said that the required tests for this update are running
+    running = 'running', 'Running'
+    # passed: a decision from Greenwave said that the required tests for this update have passed
+    passed = 'passed', 'Passed'
+    # failed: a decision from the Greenwave said that the required tests for this update have
+    # been failed
+    failed = 'failed', 'Failed'
+
+
 class UpdateType(DeclEnum):
     bugfix = 'bugfix', 'bugfix'
     security = 'security', 'security'
@@ -1084,6 +1103,11 @@ class Update(Base):
         cves (sqlalchemy.orm.collections.InstrumentedList): A list of :class:`CVE` objects
             associated with this update.
         user_id (int): A foreign key to the :class:`User` that created this update.
+        test_gating_status (EnumSymbol): The test gating status of the update. This must be one
+            of the values defined in :class:`TestGatingStatus` or ``None``. None indicates that
+            Greenwave integration was not enabled when the update was created.
+        greenwave_summary_string (unicode): A short summary of the outcome from Greenwave
+            (e.g. 2 of 32 required tests failed).
     """
     __tablename__ = 'updates'
     __exclude_columns__ = ('id', 'user_id', 'release_id', 'cves')
@@ -1148,6 +1172,10 @@ class Update(Base):
     cves = relationship('CVE', secondary=update_cve_table, backref='updates')
 
     user_id = Column(Integer, ForeignKey('users.id'))
+
+    # Greenwave
+    test_gating_status = Column(TestGatingStatus.db_type(), default=None, nullable=True)
+    greenwave_summary_string = Column(Unicode(255))
 
     @validates('builds')
     def validate_builds(self, key, build):
@@ -2271,6 +2299,10 @@ class Update(Base):
                 for committer in build.package.committers:
                     people.add(committer)
         return list(people)
+
+    @property
+    def product_version(self):
+        return self.release.long_name.lower().replace(' ', '-')
 
     def check_requirements(self, session, settings):
         """ Check that an update meets its self-prescribed policy to be pushed
--- a/bodhi/server/scripts/check_policies.py
+++ b/bodhi/server/scripts/check_policies.py
@@ -1,0 +1,72 @@
+# -*- coding: utf-8 -*-
+# Copyright © 2017 Red Hat, Inc.
+#
+# This file is part of Bodhi.
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+
+"""
+Check the enforced policies by Greenwave for each open update.
+
+Ideally, this should be done in a fedmsg consumer but we currently do not have any
+messages in the message bus yet.
+"""
+import click
+from sqlalchemy.sql.expression import false
+
+from bodhi.server import config, initialize_db, models, Session
+from bodhi.server.util import greenwave_api_post
+
+
+@click.command()
+@click.version_option(message='%(version)s')
+def check():
+    """Check the enforced policies by Greenwave for each open update."""
+    initialize_db(config.config)
+    session = Session()
+    updates = models.Update.query.filter(models.Update.pushed == false())\
+        .filter(models.Update.status.in_(
+                [models.UpdateStatus.pending, models.UpdateStatus.testing]))
+    for update in updates:
+        subjects = [build.nvr for build in update.builds]
+        data = {
+            'product_version': update.product_version,
+            'decision_context': u'bodhi_update_push_stable',
+            'subject': subjects
+        }
+        api_url = '{}/decision'.format(
+            config.config.get('greenwave_api_url').rstrip('/'))
+        try:
+            decision = greenwave_api_post(api_url, data)
+            if decision['policies_satisified']:
+                # If an unrestricted policy is applied and no tests are required
+                # on this update, let's set the test gating as ignored in Bodhi.
+                if decision['summary'] == 'no tests are required':
+                    update.test_gating_status = models.TestGatingStatus.ignored
+                else:
+                    update.test_gating_status = models.TestGatingStatus.passed
+            else:
+                update.test_gating_status = models.TestGatingStatus.failed
+            update.greenwave_summary_string = decision['summary']
+            if session.is_modified(update):
+                session.commit()
+        except Exception as e:
+            # If there is a problem talking to Greenwave server, print the error.
+            click.echo(str(e))
+            session.rollback()
+
+
+if __name__ == '__main__':
+    check()
--- a/bodhi/server/templates/update.html
+++ b/bodhi/server/templates/update.html
@@ -25,7 +25,7 @@ ${update.type} update for ${update.beautify_title(amp=True) | n}
       NEEDS_INSPECTION: 'warning',
       ABORTED: 'warning',
       CRASHED: 'warning',
-      ABSENT: 'warning',
+      ABSENT: 'danger',
     }
     var icons = {
       PASSED: 'check-circle',
@@ -48,6 +48,51 @@ ${update.type} update for ${update.beautify_title(amp=True) | n}
 
     // These are the required taskotron tests
     var requirements = ${update.requirements_json | n};
+    // show the Greenwave decision
+    var greenwave_api_url = '${request.registry.settings["greenwave_api_url"]}';
+    var missing_tests = {};
+    // handle Greenwave decision
+    var handle_unsatisfied_reqs = function(data){
+        $.each(data['unsatisfied_requirements'], function(i, req) {
+            if (req.type == 'test-result-missing') {
+                if (missing_tests[req.item] === undefined)
+                    missing_tests[req.item]= [];
+                missing_tests[req.item].push(req);
+            }
+            // the user may have already specified this in the required taskotron tests
+            if ($.inArray(req.testcase, requirements) == -1) {
+                requirements.push(req.testcase);
+            }
+        });
+    };
+    var get_unsatisfied_reqs = function(data) {
+        $.ajax({
+            url: greenwave_api_url + '/decision',
+            type: 'POST',
+            contentType: 'application/json',
+            data: JSON.stringify(data),
+            success: function(data) {
+                handle_unsatisfied_reqs(data);
+            },
+            error: function (jqxhr, status, error) {
+               $('#resultsdb h3').after(
+                  '<h4 class="text-danger">Failed to talk to Greenwave.</h4>');
+            },
+        });
+    };
+
+    var test_gating_passed = '${update.test_gating_passed}';
+    var test_gating_required = '${request.registry.settings.get("test_gating.required")}';
+    if (test_gating_required == 'True' && test_gating_passed == 'False') {
+        var summary = '${update.greenwave_summary_string}';
+        $('#resultsdb h3').after(
+           '<div class="alert alert-danger">The update can not be pushed: ' + summary + '</div>');
+        get_unsatisfied_reqs({
+            'product_version': '${update.product_version}',
+            'decision_context': 'bodhi_update_push_stable',
+            'subject': builds
+        });
+    }
 
     var make_row = function(outcome, testcase, note, arch, time, url, flavor) {
       var icon = '<span data-toggle="tooltip" data-placement="top" ' +
@@ -82,15 +127,20 @@ ${update.type} update for ${update.beautify_title(amp=True) | n}
       if (note == null){
         note = '';
       }
-      return '<tr class="row-automatedtests table-' + classes[outcome] + '" ' +
-        'style="cursor: pointer;"' +
-        'data-href="' + url + '">' +
+      html = '<tr class="row-automatedtests table-' + classes[outcome];
+      if (outcome == 'ABSENT') {
+          html += ' absent-automatedtest"';
+      } else {
+          html += '" style="cursor: pointer;" data-href="' + url + '"';
+      }
+      html += '>' +
         '<td>' + required + '</td>' +
         '<td>' + icon + '</td>' +
         '<td class="nowrap">' + testcase + '</td>' +
         '<td class="stretch-table-column text-xs-right">' + note + '</td>' +
         '<td class="nowrap">' + age + '</td>' +
         '</tr>';
+      return html;
     };
 
     var latest = {};
@@ -163,36 +213,72 @@ ${update.type} update for ${update.beautify_title(amp=True) | n}
       if (activeResultsAjaxCalls == 0){
       var data = [];
       var count = 0;
-      $.each(latest, function(buildname, obj1) {
-        count = count +1;
-        $("#resultsdb").append("<h4 class='p-t-1'>"+buildname+
-        "</h4><table class='table table-hover' id='buildresults"+count+
-        "'></table>")
-        $.each(obj1, function(testcase, obj2) {
-          $.each(obj2, function(arch, obj3) {
-            $.each(obj3, function(scenario, result) {
-              var flavor;
-              if (scenario.includes("updates-server")) {
-                flavor = "server";
+      // if there are no results in ResultsDB yet and we have already got the Greenwave decision
+      // to say that some tests are missing, let's just show them.
+      if (Object.keys(latest).length == 0) {
+          $.each(builds, function(i, buildname) {
+              count = count +1;
+              if (buildname in missing_tests) {
+                  $("#resultsdb").append("<h4 class='p-t-1'>"+buildname+
+                  "</h4><table class='table table-hover' id='buildresults"+count+
+                  "'></table>");
+                  $.each(missing_tests[buildname], function(i, result) {
+                      $('#buildresults'+count).append(make_row(
+                          'ABSENT',
+                          result.testcase,
+                          '',
+                          '',
+                          ''
+                      ));
+                  });
+              } else {
+                  $("#resultsdb").append("<h4 class='p-t-1'>"+buildname+
+                  "</h4><p>No results reported for this build.</p>");
               }
-              else if (scenario.includes("updates-workstation")) {
-                flavor = "workstation";
-              }
-              $('#buildresults'+count).append(make_row(
-                  result.outcome,
-                  result.testcase.name,
-                  result.note,
-                  // note: this is a single-item array
-                  result.data.arch,
-                  result.submit_time,
-                  result.ref_url,
-                  flavor
-              ));
+          });
+      } else {
+          $.each(latest, function(buildname, obj1) {
+            count = count +1;
+            $("#resultsdb").append("<h4 class='p-t-1'>"+buildname+
+            "</h4><table class='table table-hover' id='buildresults"+count+
+            "'></table>")
+            // We show missing tests first
+            if (buildname in missing_tests) {
+                $.each(missing_tests[buildname], function(i, result) {
+                    $('#buildresults'+count).append(make_row(
+                        'ABSENT',
+                        result.testcase,
+                        '',
+                        '',
+                        ''
+                    ));
+                });
+            }
+            $.each(obj1, function(testcase, obj2) {
+              $.each(obj2, function(arch, obj3) {
+                $.each(obj3, function(scenario, result) {
+                  var flavor;
+                  if (scenario.includes("updates-server")) {
+                    flavor = "server";
+                  }
+                  else if (scenario.includes("updates-workstation")) {
+                    flavor = "workstation";
+                  }
+                  $('#buildresults'+count).append(make_row(
+                      result.outcome,
+                      result.testcase.name,
+                      result.note,
+                      // note: this is a single-item array
+                      result.data.arch,
+                      result.submit_time,
+                      result.ref_url,
+                      flavor
+                  ));
+                });
+              });
             });
           });
-        });
-      });
-
+      }
       finish();
     };
     };
@@ -226,16 +312,14 @@ ${update.type} update for ${update.beautify_title(amp=True) | n}
         $("#tab-automatedtests").append(" <span class='label label-danger'><span class='fa fa-minus-circle'></span> "+failedcount+"</span>");
       }
 
-      // Make each cell clickable and awesome
-      $('#resultsdb tr').off().click(function(event, row) {
+      // Make each cell clickable and awesome except the missing tests
+      $('#resultsdb tr').not('.absent-automatedtest').off().click(function(event, row) {
         window.open($(this).attr('data-href'));
       });
 
       // And, re-do tooltips for newly created spans.
       $('#resultsdb span').tooltip();
     }
-
-
 
     var request_results = function(url) {
       $.ajax(url, {
@@ -369,6 +453,12 @@ $(document).ready(function(){
       $("#comment-textarea").hide();
     }
      $(this).button('toggle');
+  });
+
+  /*open the automated tests tab*/
+  $('.gating-summary').click(function(e){
+    e.preventDefault();
+    $('.nav-tabs a[href="' + this.getAttribute('href') + '"]').tab('show');
   });
 
 });
@@ -685,19 +775,20 @@ $(document).ready(function(){
                 </div>
               </div>
 
-              % if request.registry.settings.get('ci.required'):
+              % if request.registry.settings.get('test_gating.required'):
               <div class="p-b-1">
                 <div>
-                  % if request.registry.settings.get('ci.url'):
-                    <a href="${request.registry.settings.get('ci.url')}" title="More information about CI">
-                      <strong>CI Status</strong>
+                  % if request.registry.settings.get('test_gating.url'):
+                    <a href="${request.registry.settings.get('test_gating.url')}" title="More information about test gating">
+                      <strong>Test Gating Status</strong>
                     </a>
                   % else:
-                    <strong>CI Status</strong>
+                    <strong>Test Gating Status</strong>
                   % endif
                 </div>
                 <div>
-                  ${self.util.ci_status2html(update.ci_status.description) | n}
+                  ${self.util.test_gating_status2html(update.test_gating_status,
+                    update.greenwave_summary_string) | n}
                 </div>
               </div>
               % endif
@@ -879,15 +970,6 @@ $(document).ready(function(){
             <a href="https://koji.fedoraproject.org/koji/search?terms=${build.nvr}&type=build&match=glob" target="_blank">
                 ${build.nvr}</a>
           </td>
-          % if request.registry.settings.get('ci.required'):
-            <td>
-              % if build.ci_url:
-                <a href="${build.ci_url}">${self.util.ci_status2html(build.ci_status.description) | n}</a>
-              % else:
-                ${self.util.ci_status2html(build.ci_status.description) | n}
-              % endif
-            </td>
-          % endif
           <td class="pull-right">
             % if build.signed:
               <span class="fa fa-key text-muted" aria-hidden="true" data-toggle="tooltip" data-placement="top" title="Build signed"></span>
--- a/bodhi/server/util.py
+++ b/bodhi/server/util.py
@@ -37,6 +37,7 @@ import libravatar
 import markdown
 import requests
 import rpm
+import json
 
 from bodhi.server import log, buildsys, Session
 from bodhi.server.config import config
@@ -730,21 +731,32 @@ def get_critpath_components_from_pdc(branch, component_type='rpm'):
     return list(critpath_pkgs_set)
 
 
-def api_get(api_url, service_name, error_key=None):
+def call_api(api_url, service_name, error_key=None, method='GET', data={}):
     """
-    A wrapper to get API resources with some error handling.
-    :param api_url: a string of the URL to query
+    A wrapper to send HTTP request with some error handling.
+    :param api_url: a string of the API URL.
     :param service_name: a string of the service name to use in error messages
     :param error_key: a string of the key that holds the error message in the
     JSON body. If this is set to None, the whole JSON will be used as the
     error message.
+    :param error_key: a string of the HTTP method to use to request. The defaut
+    value is ``GET``.
+    :param data: a dict of data will be sent to the API server.
     :return: dictionary of the returned JSON or a RuntimeError
     """
-    base_error_msg = (
-        'Bodhi failed to get a resource from {0} at the following URL '
-        '"{1}". The status code was "{2}".')
-    rv = requests.get(api_url, timeout=60)
-
+    if method == 'POST':
+        base_error_msg = (
+            'Bodhi failed to send POST request to {0} at the following URL '
+            '"{1}". The status code was "{2}".')
+        rv = requests.post(api_url,
+                           headers={'Content-Type': 'application/json'},
+                           data=json.dumps(data),
+                           timeout=60)
+    else:
+        base_error_msg = (
+            'Bodhi failed to get a resource from {0} at the following URL '
+            '"{1}". The status code was "{2}".')
+        rv = requests.get(api_url, timeout=60)
     if rv.status_code == 200:
         return rv.json()
     elif rv.status_code == 500:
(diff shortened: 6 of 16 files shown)


bodhi/server/config.py
```diff
@@ -407,6 +407,12 @@ class BodhiConfig(dict):
         'initial_bug_msg': {
             'value': '%s has been submitted as an update to %s. %s',
             'validator': unicode},
+        'greenwave_api_url': {
+            'value': 'https://greenwave.fedoraproject.org/api/v1.0',
+            'validator': unicode},
+        'greenwave_enabled': {
+            'value': False,
+            'validator': _validate_bool},
```
bowlofeggs: I had a talk with @ralphbean today about the confusion between the existing CI code that @pypingou wrote and the new plans for adding Greenwave support to Bodhi. We came to a conclusion together that we should try to resolve the differences in this patch rather than adding another way to gate updates (because without this patch there are already two ways to gate - ci_status and the code that blocks updates based on test results).

All that to say - there is already a ```ci.enabled``` setting - rather than having a ```greenwave_enabled``` setting, why not just make ```ci_enabled``` mean that we will try to talk to Greenwave? Later in this pull request I see that you have replaced usages of ```ci_enabled``` anyway, so it probably doesn't make sense to keep both settings. I prefer the ```ci_enabled``` name just for being more generic and meaningful to admins that might not be familiar with Greenwave but will know what CI means.

bodhi/server/models.py
```diff
@@ -1149,6 +1153,10 @@ class Update(Base):
 
     user_id = Column(Integer, ForeignKey('users.id'))
 
+    # Greenwave
+    greenwave_policies_satisfied = Column(Boolean, default=True, nullable=False)
+    greenwave_summary_string = Column(Unicode(255))
```
bowlofeggs: Is this guaranteed by Greenwave to be 255 characters or less?

bodhi/server/models.py
```diff
@@ -1756,13 +1764,15 @@ def set_request(self, db, action, username):
                         'stable repository.  It must first reach a karma of %s, consisting of %s '
                         'positive karma from proventesters, along with %d additional karma from '
                         'the community. Or, it must spend %s days in testing without any negative '
-                        'feedback')
+                        'feedback.')
```
bowlofeggs: I'm +1 to fixing the grammar here, but please do it in a separate pull request since it is not related to adding Greenwave functionality to Bodhi.

bodhi/server/models.py
```diff
@@ -1756,13 +1764,15 @@ def set_request(self, db, action, username):
...
                     stern_note = stern_note % (
                         config.get('critpath.min_karma'),
                         config.get('critpath.num_admin_approvals'),
                         (config.get('critpath.min_karma') -
                             config.get('critpath.num_admin_approvals')),
                         config.get('critpath.stable_after_days_without_negative_karma'))
+                    if config.get('greenwave_enabled'):
+                        stern_note += ' Or, it must pass gating.'
```
bowlofeggs: Shouldn't this be " and, it must pass gating."?

bodhi/server/models.py
```diff
@@ -2272,6 +2282,11 @@ def get_maintainers(self):
                     people.add(committer)
         return list(people)
 
+    @property
+    # FIXME: query PDC to get this
```
bowlofeggs: Let's not put todo's in the code - we can file issues for things like this.

bodhi/server/models.py
```diff
@@ -2272,6 +2282,11 @@ def get_maintainers(self):
...
+    def product_version(self):
```
bowlofeggs: Let's give this method a docblock that documents the return type as a ```basestring```.

bodhi/server/models.py
```diff
@@ -2346,8 +2361,8 @@ def check_requirements(self, session, settings):
                     return False, "Required task %s returned %s" % (
                         latest['testcase']['name'], latest['outcome'])
 
-        if config.get('ci.required') and not self.ci_passed:
-            return False, "CI did not pass on this update"
+        if config.get('greenwave_enabled') and not self.greenwave_policies_satisfied:
+            return False, "Greenwave policies are not satisfied on this update"
```
bowlofeggs: Let's s/Greenwave/Testing/, so that users not familiar with Greenwave will understand what this means.

bodhi/server/scripts/check_policies.py
```diff
@@ -0,0 +1,67 @@
+# -*- coding: utf-8 -*-
+# Copyright ® 2017 Red Hat, Inc.
```
bowlofeggs: This should be a © instead of a ®. I see that I made that mistake in some of my headers, so I've filed https://github.com/fedora-infra/bodhi/pull/1695 to fix that. Sorry!

bodhi/server/scripts/check_policies.py
```diff
@@ -0,0 +1,67 @@
...
+
+"""
+Check the enforced policies by Greenwave for each open update.
+
+Ideally, this should be done in a fedmsg consumer but we currently do not have any
+messages in the message bus yet.
+"""
+import click
+
+from sqlalchemy.sql.expression import false
```
bowlofeggs: For PEP-8, the sqlalchemy import should be grouped with the click import, and the bodhi imports should be on their own. i.e., add a space here, and delete the space above this line.

bodhi/server/models.py
```diff
@@ -1149,6 +1153,10 @@ class Update(Base):
 
     user_id = Column(Integer, ForeignKey('users.id'))
 
+    # Greenwave
+    greenwave_policies_satisfied = Column(Boolean, default=True, nullable=False)
```
bowlofeggs: Rather than keeping a boolean on whether the greenwave policies are satisfied, why not take advantage of the more rich states available with the ```ci_status``` field? That way, we can express more states than just "yes nor no". We have many states like "waiting", "failed", "skipped", "passed", "running", and maybe others I'm forgetting.

I propose that we just use that existing field and allow Greenwave to be the source of information we use to set that state.

bodhi/server/scripts/check_policies.py
```diff
@@ -0,0 +1,67 @@
...
+
+@click.command()
+@click.version_option(message='%(version)s')
+def check():
+    """Check the enforced policies by Greenwave for each open update."""
+    initialize_db(config.config)
+    session = Session()
+    updates = models.Update.query.filter(models.Update.pushed == false())\
+        .filter(models.Update.status.in_(
+                [models.UpdateStatus.pending, models.UpdateStatus.testing]))
```
bowlofeggs, line 41: If we use the ```ci_status``` field as I am suggesting, this query could be more targeted by querying for updates that are not in statuses passed, failed, and skipped. That way we don't check over and over again on updates that have already been spoken for.

bodhi/server/scripts/check_policies.py
```diff
@@ -0,0 +1,67 @@
...
+        subjects = [build.nvr for build in update.builds]
+        data = {
+            'product_version': update.product_version,
+            'decision_context': u'bodhi_update_push_stable',
+            'subject': subjects
+        }
+        api_url = '{}/decision'.format(
+            config.config.get('greenwave_api_url').rstrip('/'))
+        try:
+            click.echo("Get the decision for update %s" % update.title)
```
bowlofeggs: Presumably this will be run in cron. We probably don't want this script to print to the terminal as that could result in e-mails getting sent.

bodhi/server/scripts/check_policies.py
```diff
@@ -0,0 +1,67 @@
...
+            decision = greenwave_api_post(api_url, data)
+            update.greenwave_policies_satisfied = decision['policies_satisified']
```
bowlofeggs: We could use this to set the ```ci_status``` to passed, failed, or skipped. @ralphbean told me today that Greenwave will give a True response here for skipped updates, and that we can tell the difference between skipped and passed by inspecting the summary.

bodhi/server/scripts/check_policies.py
```diff
@@ -0,0 +1,67 @@
...
+            update.greenwave_summary_string = decision['summary']
+            if session.is_modified(update):
+                session.commit()
+        except Exception as e:
+            # If the Greenwave server is down or no policies are enforced, we just skip
+            # this one and print the error.
+            click.echo(str(e))
```
bowlofeggs: Since this is an error, I'm +1 to printing it to the terminal for cron to e-mail. You could also consider writing it to stderr, but this is fine too.

bodhi/server/scripts/check_policies.py
```diff
@@ -0,0 +1,67 @@
...
+            session.rollback()
+            pass
```
bowlofeggs: Let's drop this pass since it is not needed.

bodhi/server/templates/update.html
```diff
@@ -48,6 +48,51 @@
...
+                if (missing_tests[req.item] === undefined)
+                    missing_tests[req.item]= [];
+                missing_tests[req.item].push(req);
+            }
+            // the user may have already specified this in the required taskotron tests
+            if ($.inArray(req.testcase, requirements) == -1) {
+                requirements.push(req.testcase);
+            }
+        });
+    };
```
bowlofeggs, line 67: Why have the JavaScript query Greenwave when Bodhi will already know the Greenwave response? There are a few negatives to doing this:

1) Since Bodhi will poll Greenwave (at least initially until we have fedmsg), this will introduce a possibility of confusion if users see the JavaScript saying that the tests pass, but Bodhi's backend doesn't know that yet. This will lead people to ask questions in IRC. If we instead just display ```ci_status``` like we do today, it will be much less confusing for users.
2) This will put more load on the Greenwave server. ResultsDB already struggles a bit with the load that Bodhi's JS puts on it (and we've been wanting to make Bodhi cache the results there too to help with that).

I think we should avoid having the UI talk to Greenwave and just stick to using the ```ci_status``` field together with the "Automated tests" tab. I do think it would be important to display Greenwave's summary data somewhere. Perhaps @ryanlerch has ideas for the UI there?

bodhi/server/templates/update.html
```diff
@@ -685,19 +775,14 @@ <h5 class="m-t-2">
                 </div>
               </div>
 
-              % if request.registry.settings.get('ci.required'):
+              % if request.registry.settings.get('greenwave_enabled'):
               <div class="p-b-1">
                 <div>
-                  % if request.registry.settings.get('ci.url'):
-                    <a href="${request.registry.settings.get('ci.url')}" title="More information about CI">
```
bowlofeggs: @ralphbean had requested us to use this ```ci.url``` here. I don't have strong feelings about it, but he wanted to be able to link people to information about the tests. If we do decide that we don't need this anymore, please remove ```ci.url``` from the settings as well (I think this was the only place it was used).

bodhi/server/templates/update.html
```diff
@@ -685,19 +775,14 @@ <h5 class="m-t-2">
...
-                    </a>
-                  % else:
-                    <strong>CI Status</strong>
-                  % endif
+                  <strong>Gating Status</strong>
                 </div>
                 <div>
-                  ${self.util.ci_status2html(update.ci_status.description) | n}
+                  ${self.util.gating_status2html(update.greenwave_policies_satisfied,
+                    update.greenwave_summary_string) | n}
```
bowlofeggs: Again, I think the existing field is best to keep here too as it allows more states than a bool.

bodhi/server/util.py
```diff
@@ -343,6 +344,28 @@ def ci_status2html(context, status):
...
+    Returns:
+        unicode: A human friendly HTML label of the gating status
+
+    """
+    if status:
+        return u'<span class="fa fa-check-circle text-success"></span>'
+    else:
+        html = u'<span class="fa fa-minus-circle text-danger"></span>'
+        if reason:
+            html += u'<a class="gating-summary" href="#automatedtests">%s</a>' % reason
```
bowlofeggs: Is there no reason to display the summary if the tests passed? I would imagine there isn't a reason, just wanted to double check.

bodhi/server/util.py
```diff
@@ -730,21 +753,31 @@ def get_critpath_components_from_pdc(branch, component_type='rpm'):
...
-        '"{1}". The status code was "{2}".')
-    rv = requests.get(api_url, timeout=60)
-
+    if method == 'POST':
+        base_error_msg = (
+            'Bodhi failed to send POST request to {0} at the following URL '
+            '"{1}". The status code was "{2}".')
+        rv = requests.post(api_url,
+                           headers={'Content-Type': 'application/json'},
+                           data=json.dumps(kwargs.get('data', {})),
```
bowlofeggs: Since kwargs is only meaningful if it has a ```data``` in it, why not avoid the generic and undocumented ```kwargs``` variable, and instead give the method signature a ```post_params=None``` (or the name you are using ```data``` is fine too), which gets used here if set? Please do document this in the docblock.

bodhi/tests/server/scripts/test_check_policies.py
```diff
@@ -0,0 +1,93 @@
+# -*- coding: utf-8 -*-
+# Copyright ® 2017 Red Hat, Inc.
```
bowlofeggs: Watch out for the ® here too.

bodhi/tests/server/scripts/test_check_policies.py
```diff
@@ -0,0 +1,93 @@
...
+    """This class contains tests for the check_policies() function."""
+    @patch.dict(config, [('greenwave_api_url', 'http://domain.local')])
+    def test_policies_satisfied(self):
+        """Assert correct behavior when the policies enforced by Greenwave are satisfied"""
+        runner = testing.CliRunner()
+        update = self.db.query(models.Update).all()[0]
+        update.status = models.UpdateStatus.testing
+        self.db.commit()
+        with patch('bodhi.server.scripts.check_policies.greenwave_api_post') as mock_greenwave:
+            expected_json = {
```
bowlofeggs: The name "expected" makes it sound like we are going to use it in an assertion, but really it's just mock data. Why not name it "greenwave_response" or similar?

bodhi/tests/server/scripts/test_check_policies.py
```diff
@@ -0,0 +1,93 @@
...
+
+    @patch.dict(config, [('greenwave_api_url', 'http://domain.local')])
+    def test_policies_unsatisfied(self):
+        """Assert correct behavior when the policies enforced by Greenwave are unsatisfied"""
+        runner = testing.CliRunner()
+        update = self.db.query(models.Update).all()[0]
+        update.status = models.UpdateStatus.testing
+        self.db.commit()
+        with patch('bodhi.server.scripts.check_policies.greenwave_api_post') as mock_greenwave:
+            expected_json = {
```
bowlofeggs: Similar here.

bodhi/tests/server/services/test_updates.py
```diff
@@ -719,7 +719,7 @@ def test_provenpackager_request_privs(self, publish, *args):
     @mock.patch(**mock_valid_requirements)
     @mock.patch('bodhi.server.notifications.publish')
     def test_provenpackager_request_update_queued_in_ci(self, publish, *args):
-        """Ensure provenpackagers cannot request changes for any update which
+        """Ensure provenpackagers can request changes for any update which
```
bowlofeggs: Isn't "cannot" correct here? The test seems to assert that the provenpackager is unable to do this.

bodhi/tests/server/services/test_updates.py
```diff
@@ -745,17 +745,17 @@ def test_provenpackager_request_update_queued_in_ci(self, publish, *args):
...
             res.json_body['errors'][0]['description'],
-            'Requirement not met CI did not pass on this update')
+            config.get('not_yet_tested_msg'))
 
     @mock.patch(**mock_taskotron_results)
     @mock.patch(**mock_valid_requirements)
     @mock.patch('bodhi.server.notifications.publish')
     def test_provenpackager_request_update_running_in_ci(self, publish, *args):
-        """Ensure provenpackagers cannot request changes for any update which
+        """Ensure provenpackagers can request changes for any update which
```
bowlofeggs: Similar here?

bodhi/tests/server/services/test_updates.py
```diff
@@ -781,17 +781,17 @@ def test_provenpackager_request_update_running_in_ci(self, publish, *args):
...
             res.json_body['errors'][0]['description'],
-            'Requirement not met CI did not pass on this update')
+            config.get('not_yet_tested_msg'))
 
     @mock.patch(**mock_taskotron_results)
     @mock.patch(**mock_valid_requirements)
     @mock.patch('bodhi.server.notifications.publish')
     def test_provenpackager_request_update_failed_ci(self, publish, *args):
-        """Ensure provenpackagers cannot request changes for any update which
+        """Ensure provenpackagers can request changes for any update which
```
bowlofeggs: And here?

bodhi/tests/server/services/test_updates.py
```diff
@@ -932,6 +932,83 @@ def test_provenpackager_request_update_with_none_ci(self, publish, *args):
...
+        nvr = u'bodhi-2.1-1.fc17'
+        user = User(name=u'bob')
+        self.db.add(user)
+        group = self.db.query(Group).filter_by(name=u'provenpackager').one()
+        user.groups.append(group)
+        settings = self.app_settings.copy()
+        settings['greenwave_enabled'] = True
+        self.db.commit()
+
+        app = TestApp(main({}, testing=u'bob', session=self.db, **settings))
```
bowlofeggs: It seems that you might be creating an app to get a setting to be a particular way here. The ```TestApp``` is known to have a memory leak in it and is the source of much of the memory usage when running Bodhi's test suite. Instead, it would be better to use ```mock.patch.dict``` on ```bodhi.server.config.config``` to get settings to be a particular way.

There should be a ```self.app``` that you can use that is a singleton for the test suite to help keep memory usage down. There are still a few places in the code that create apps like this and we hope to eliminate the rest of them in the future.

bodhi/tests/server/services/test_updates.py
```diff
@@ -932,6 +932,83 @@ def test_provenpackager_request_update_with_none_ci(self, publish, *args):
             res.json_body['errors'][0]['description'],
             config.get('not_yet_tested_msg'))
 
+    @mock.patch(**mock_taskotron_results)
+    @mock.patch(**mock_valid_requirements)
+    @mock.patch('bodhi.server.notifications.publish')
+    def test_provenpackager_request_update_with_gw_satisfied(self, publish, *args):
+        """Ensure provenpackagers can request changes for any update which
+        satisfies Greenwave policies"""
```
bowlofeggs: This one says "can", but then asserts that the user cannot.

bodhi/tests/server/services/test_updates.py
```diff
@@ -932,6 +932,83 @@ def test_provenpackager_request_update_with_none_ci(self, publish, *args):
...
+        nvr = u'bodhi-2.1-1.fc17'
+        user = User(name=u'bob')
+        self.db.add(user)
+        group = self.db.query(Group).filter_by(name=u'provenpackager').one()
+        user.groups.append(group)
+        settings = self.app_settings.copy()
+        settings['greenwave_enabled'] = True
+        self.db.commit()
+
+        app = TestApp(main({}, testing=u'bob', session=self.db, **settings))
```
bowlofeggs: Similar here.

production.ini
```diff
@@ -66,10 +66,14 @@ use = egg:bodhi-server
 # wiki_url = https://fedoraproject.org/w/api.php
 # test_case_base_url = https://fedoraproject.org/wiki/
 
-# URL of the resultsdb for integrating checks and stuff
-# resultsdb_api_url = https://taskotron.fedoraproject.org/resultsdb_api/
```
bowlofeggs: This gets used to display the test results in the UI. I think we need to keep it for now.

bowlofeggs (changes requested):
Let's work to merge the greenwave idea with the existing ```ci_status``` code. One thing I think we want to do is to move ```ci_status``` from the ```Build``` model to the ```Update``` model since Greenwave makes decisions on updates rather than builds.

Also, if you think you might send future patches to Bodhi, feel free to send me a separate PR that adds your GitHub username to the list of trusted patch writers in ```devel/ci/githubprb-project.yml``` (see the top of the file).

bowlofeggs:
jenkies test

bodhi/server/config.py
bowlofeggs: We could also use a different name than CI if we want, like test_gating_status, or something. If we want a more neutral name. What do you think @ralphbean?

bodhi/server/models.py
bowlofeggs: Also, the CI failed because this line is missing test coverage. If you are using the Vagrant development environment, ```bdiff-cover``` can tell you if you have full coverage on your patch.

matt8754:
FYI, I made a demo showing this in action. 
https://fedorapeople.org/groups/factory2/sprint-035/mjia-bodhi-greenwave.mp4

bodhi/server/config.py
matt8754: The reason I keep the code that blocks updates based on test results is that I assume someone must have used it and I do not want to break them. Eventually, we will get rid of it for sure. 
Yeah, I agree that we should reuse the ci.enabled setting, but I would suggest using a different name to make it more clearly.  What about 'gating.enabled' or 'test_gating.enabled'?

bodhi/server/models.py
matt8754: I do not think this filed suits for Greenwave as Greenwave can't tell anything about "waiting","running", etc.

bodhi/server/models.py
matt8754: Yeah, this is a last minute change and I forgot to run bdiff-cover, :-)

matt8754: Well, if Greenwave is not enabled, the application will not block the update, right? So in this case, I think we should use 'Or, it must pass gating'.

bodhi/server/models.py
matt8754: Yeah, I could take this out from this PR.

bodhi/server/models.py
matt8754: ACK.

bodhi/server/models.py
matt8754: Ack.

bodhi/server/models.py
matt8754: May be a better message like "This update did not pass gating"?

bodhi/server/scripts/check_policies.py
matt8754: No worries. I will fix it here as well, :-)

bodhi/server/scripts/check_policies.py
matt8754: Ack. This is a new thing I learn today.

bodhi/server/scripts/check_policies.py
matt8754: It is intentional to check the pending updates over and over again. This is because Greewave could say yes after it says no, so I'm not sure using ci_status filed could cover this case.

bodhi/server/scripts/check_policies.py
matt8754: Yeah, we can get rid of it.

bodhi/server/scripts/check_policies.py
matt8754: Okay, that makes sense now. If we want to use the ci_status field, should we take out 'waiting','queued' and 'running' as those three do not make sense to Greenwave and Taskotron tasks?

bodhi/server/scripts/check_policies.py
matt8754: Agreed.

bodhi/server/templates/update.html
matt8754: Because I want to show unsatisfied requirements on the web UI which we do not store it in the database.  This is important to let people know why the update is blocked.  Yeah, the first negative point is valid and I'm not sure how to handle this. Should we add a new field into the update table to store unsatisfied requirements? For the second negative point, I think it would not put more load on the Greenwave server since there will be only one request to be sent.

bodhi/server/templates/update.html
matt8754: I guess the value of ci.url would be the ci centos, right? I do not think it makes sense to show it if we decide to change it to 'Gating Status'.

bodhi/server/util.py
matt8754: I just do not see any value of showing it as I think no one would care the summary if all the tests are passing.

bodhi/server/util.py
matt8754: Sounds reasonable to me.

bodhi/tests/server/scripts/test_check_policies.py
matt8754: Agreed.

bodhi/tests/server/services/test_updates.py
matt8754: The thing is CI integration no longer gates on the test results, so I change this test case to assert that the provenpackager now can do this. Maybe I should add a test to assert that the provenpackager can not push a package update when the gating status is failed. I might already have a test case for this.

bodhi/tests/server/services/test_updates.py
matt8754: This one actually asserts that the reason the user can't push is not due to Greenwave policy unsatisfied. I forgot to update the comment above line 967, :-)

bodhi/tests/server/services/test_updates.py
matt8754: It seems like if I change the settings in self.app for one test, it will interfere with other tests in the same class.  Maybe I should set it to something at the start and then set it back at the end of the test?

production.ini
matt8754: Yeah, I accidentally deleted it, sorry.

bodhi/server/models.py
matt8754: I think so as we want to keep it in a short summary.

bowlofeggs:
On Thu, 2017-07-20 at 00:21 +0000, mjia wrote:
> FYI, I made a demo showing this in action.
> https://fedorapeople.org/groups/factory2/sprint-035/mjia-bodhi-greenw
> ave.mp4

Very nice!

bodhi/server/config.py
bowlofeggs: @matt8754 I'm fine with keeping the existing code that blocks on resultsdb results if you like - I think that's a fine reason.

I like ```test_gating.enabled```. Bodhi in general does gating, so a setting that just says "gating" could sound like it's disabling all gating. How about we rename ```ci.enabled``` to ```test_gating.enabled``` and use that everywhere?

bowlofeggs: Nothing could ever have meaningfully used ```ci.enabled``` yet since it wasn't hooked up to anything, so I don't think we have a backwards incompatibility concern to worry about.

bodhi/server/models.py
bowlofeggs: @matt8754 Greenwave can tell about the "final" states though - "passed", "failed", and "skipped". I think we should get this patch to use the ```ci_status``` field (we can rename it to ```test_gating_status``` or something if you like that better, to be consistent with the setting name), but this patch can just worry about those three states.

I can write more code later that can figure out how to do "waiting" and "running" and "queued" and make that a separate concern. Those states are really just for user convenience/feedback and aren't necessary for Bodhi to function correctly. As long as we can reliably land in "passed", "failed", or "skipped" then I think this patch will be good to go, and Greenwave is able to tell us about those states.

bowlofeggs: I expect that I'll know about the convenience states by watching fedmsg for messages from taskotron and CentOS CI. So there will probably be a new fedmsg listener that sets those three in the future. This is in the "nice to have" category in my mind, so don't concern yourself about that for now ☺

bodhi/server/models.py
bowlofeggs: The "or" makes it sound like passing gating will allow the developer to push the update to stable without karma if it passes gating. With this code, it produces this message:

> This critical path update has not yet been approved for pushing to the stable repository. It must first reach a karma of 2, consisting of 2 positive karma from proventesters, along with 0 additional karma from the community. Or, it must spend 14 days in testing without any negative feedback. Or, it must pass gating.

As I understand it, the automated tests do not supercede the other requirements (karma or time), but are in addition to the other requirements. Thus, I suggest "Additionally, it must pass automated tests." so that it reflects the Fedora policy.

bodhi/server/models.py
bowlofeggs: I'm not sure that "gating" will be meaningful to the developers - I think we should talk about it not passing the required tests. Maybe "Required tests did not pass on this update"?

bodhi/server/scripts/check_policies.py
bowlofeggs: Can it also say "no" after it says "yes"? If so, then I think this code is fine. If not, I think the ```ci_status``` field (potentially renamed if you want) is still useful, because we can just query for updates that are not in "passed" or "skipped" (i.e., just leave off "failed").

bowlofeggs:
On Thu, 2017-07-20 at 03:11 +0000, mjia wrote:
> Okay, that makes sense now. If we want to use the ci_status field,
> should we take out 'waiting','queued' and 'running' as those three do
> not make sense to Greenwave and Taskotron tasks?

I think we should leave those states, but just not use them in this
pull request. I have ideas for how to make those work by watching
fedmsg for taskotron messages, and I'll just do that in the future.

bodhi/server/templates/update.html
bowlofeggs: @matt8754 after watching your [demo](https://fedorapeople.org/groups/factory2/sprint-035/mjia-bodhi-greenwave.mp4) I see what you mean. It seems that you've done some nice user feedback there that I do think is helpful and useful. I suppose we could take the approach you've done here, and then refactor it later when we refactor the resultsdb code to keep the data in Bodhi's backend rather than live querying resultsdb.

@ryanlerch @kparal @AdamWill, what do you think?

So yeah, point #1 that I made above will still be confusing until we do that refactor. As for point #2, what I meant is that every time anyone loads a Bodhi update, Greenwave will be queried where it would only be queried once and cached on Bodhi's server if we keep this information on Bodhi's backend. For example, say there's a Firefox update and people load the page about it 100 times. With this code, that will cause 100 queries on the Greenwave server. If we refactor Bodhi to be able to store this information in its backend, that would result in 1 query instead. This has caused considerable load on ResultsDB in the past (though I think some of that problem has been addressed by making this JS more efficient than it used to be).

Anyway, I'm not opposed to doing it in JavaScript for now, but I would like feedback from @ryanlerch, @kparal, or @AdamWill since they have been heavy contributors to this code recently.

bowlofeggs:
On Thu, 2017-07-20 at 03:32 +0000, mjia wrote:
> I guess the value of ci.url would be the ci centos, right? I do not
> think it makes sense to show it if we decide to change it to 'Gating
> Status'.

@ralphbean had requested it to be a URL for end-user information about
Fedora's testing policies. It's really only used to insert a
configurable link that users can click on to see what all the testing
stuff is about. We can also rename it to ```testing_policy.url``` or
something if you like that better. Or if @ralphbean doesn't think we
need this anymore, we can also drop it.

bowlofeggs:
On Thu, 2017-07-20 at 03:34 +0000, mjia wrote:
> I just do not see any value of showing it as I think no one would
> care the summary if all the tests are passing.

Cool, works for me.

ralphbean:
>  It's really only used to insert a configurable link that users can click on to see what all the testing
stuff is about. We can also rename it to ```testing_policy.url``` or something if you like that better.

Yeah - it's supposed to be a link to a wiki page so people can go read something to understand what is being changed under the hood.

Calling it `testing_policy.url` or `whatever` works for me.  :)

bowlofeggs:
On Thu, 2017-07-20 at 03:46 +0000, mjia wrote:
> This one actually asserts that the reason the user can't push is not
> Greenwave. I forgot to update the comment above line 967, :-)

@ralphbean and I talked yesterday and we both thought it was reasonable
to remove the other test gating from Bodhi. I know I said earlier than
we don't have to do that in this PR, and I still think we don't *have*
to, but it might help make the code cleaner and easier to understand if
we let Greenwave be the only test gating mechanism in Bodhi rather than
having two ways to do it. We could just name the next release Bodhi 3
and just drop the old feature.

I'm also fine with leaving the old way in and removing it in the future
- I'm just suggesting that removing it may simplify things for you.

bowlofeggs:
On Thu, 2017-07-20 at 03:55 +0000, mjia wrote:
> It seems like if I change the settings in self.app, it will interfere
> with other tests in the class. Maybe I should set it at start and
> then set it back at the end of the test?

You should be able to use ```mock.patch.dict``` so that the change is
only present in the mock context manager. So you could do:

```
@mock.patch.dict(bodhi.server.config.config, {'test_gating.enabled': True})
```

or even:

```
with mock.patch.dict(bodhi.server.config.config, {'test_gating.enabled': True}):
    # Test code that needs this setting here
```

This way you can still use ```self.app``` but that setting will not be
changed for other tests.

bowlofeggs:
On Thu, 2017-07-20 at 04:01 +0000, mjia wrote:
> I think so as we want to keep it in a short summary.

Cool.

bodhi/server/models.py
matt8754: Alright, that sounds good to me.

bodhi/server/models.py
matt8754: Yes, you are right, we're just adding one requirement based on the automated tests. Suggestion accepted.

bodhi/server/models.py
matt8754: Yeah, I like it.

bodhi/server/scripts/check_policies.py
matt8754: It could be if people revoke the waiver of a failed test result and Greewave will say 'no' in this case.

matt8754:
Thanks, Randy, I will update this patch to address the comments.

bodhi/server/models.py
matt8754: @bowlofeggs , to be clear, if Greenwave says "yes", should I mark the ci_status field "passed" for all the builds of the update? As you have mentioned before, you want to move the ci_status field from the Build model to the Update model, have you started yet? If not. I can add this as a part of this PR.

bowlofeggs:
On Fri, 2017-07-21 at 04:59 +0000, mjia wrote:
> @bowlofeggs , to be clear, if Greenwave says "yes", should I mark the
> ci_status field "passed" for all the builds of the update? As you
> have mentioned before, you want to move the ci_status field from the
> Build model to the Update model, have you started yet? If not. I can
> add this as a part of this PR.

I think it would make sense to move that field from the Build to the
Update in this PR.

matt8754:
Thanks, I will give it try. I have a feeling that doing this might make this PR more

complicated, :-)


发送自 Outlook<http://aka.ms/weboutlook>


________________________________
发件人: Randy Barlow <notifications@github.com>
发送时间: 2017年7月24日 18:44
收件人: fedora-infra/bodhi
抄送: mjia; Mention
主题: Re: [fedora-infra/bodhi] Greenwave integration (#1694)

On Fri, 2017-07-21 at 04:59 +0000, mjia wrote:
> @bowlofeggs , to be clear, if Greenwave says "yes", should I mark the
> ci_status field "passed" for all the builds of the update? As you
> have mentioned before, you want to move the ci_status field from the
> Build model to the Update model, have you started yet? If not. I can
> add this as a part of this PR.

I think it would make sense to move that field from the Build to the
Update in this PR.

―

matt8754:
With the latest update, I think I've addressed all the comments. I've spent some time trying to move ci_status field from the Build model to the Update model and then it turns out this is not an easy task which basically gets rid of the whole CI integration.
To make this PR simple, I've removed the test gating on CI integation and kept the rest. We can remove it in another PR. 
I've attached some screenshots since I've slightly modified the UI.

matt8754:
screenshot from 2017-07-26 17-11-51

matt8754:
screenshot from 2017-07-26 17-12-03

matt8754:
screenshot from 2017-07-26 17-12-15

matt8754:
screenshot from 2017-07-26 17-12-20

bowlofeggs:
I was talking with @pypingou today about this, and he raised a concern about abstracting the CI pipeline with Greenwave: the CI pipeline doesn't have uniqueness on NVRs, and instead knows results based on commit hashes. Bodhi does know about commit hashes, but the Greenwave API seems to be receiving NVRs from Bodhi instead of commit hashes. Thus, as is this PR isn't using enough information to get test results from the CI pipeline.

bowlofeggs:
I filed https://pagure.io/greenwave/issue/34 about needing the ability to query for test results by commit hash in addition to NVR.

bodhi/server/util.py
```diff
@@ -786,4 +806,17 @@ def pdc_api_get(pdc_api_url):
...
+def greenwave_api_post(greenwave_api_url, data):
+    """
+    A wrapper to post request to Greenwave.
+    :param greenwave_api_url: a string of the URL to query
+    :param data: A dictionary of data to send along the request
+    :return: JSON of the API response or a RuntimeError
+    """
+    # There is no error_key specified because the error key is not consistent
+    # based on the error message
+    return call_api(greenwave_api_url, service_name='Greenwave', method='POST',
```
puiterwijk, line 821: **Please be VERY CAUTIOUS HERE!**
This code, via call_api, ASSUMES that any non-500 error returned a JSON results, which is DEFINITELY not the result with greenwave, errors are either HTML or plain text, see: https://pagure.io/greenwave/issue/37.

puiterwijk (changes requested):

bodhi/server/util.py
matt8754: Yes, I have already filed an issue like yours a while ago, :-)

matt8754:
Today, I have been trying to implement the fedmsg support in Greenwave. When going through the design doc of Greenwave, I've noticed one thing that Greenwave will listen for new results from ResultsDB and new waivers from WaiverDB. When a new result or waiver is received, Greenwave will check for all the applicable polices for that build and publish a message about the new satisfied/unsatisfied policy.  And then Bodhi can consume this message and set the test_gating_status on that build.  This makes me think we should have the test_gating_status on the Build model rather the Update model because eventually we want to have fedmsg drive all of this. As to decide whether an update has passed the test gating, Bodhi will look at all the builds in this update and see if they have all satisfied the policies, as if how CI integration decide whether an update has passed CI. However, this approach does not work for the OpenQA results because they are for updates, not for builds.  I'm wondering in addition to have the test_gating_status on the builds, may be we can have a new table update_alias with the test_gating_status to store the decisions made on the OpenQA results?

bowlofeggs:
On Fri, 2017-07-28 at 07:11 +0000, mjia wrote:
> This makes me think we should have the test_gating_status on the
> Build model rather the Update model because eventually we want to
> have fedmsg drive all of this.

This has been a significant point of confusion to all of us so far, I
think ☺

If you recall, I had filed an issue against Greenwave to ask for it to
use HTTP GET instead of POST for decisions[0], but the reply was that
Greenwave was intended to give feedback about updates, not builds. That
was confusing to me at the time because we had designed Bodhi to care
about builds instead of updates with the CI status work that Pierre-
Yves did.

Then we decided that it did make sense to say that the gating decision
should be on the update instead of the build, because there are real
world scenarios where updates would fail independently, but would pass
together (or worse, vice-versa!) So that convinced me that we should
gate based on the collection of builds in an update rather than on each
build individually.

Fast forward to now, and you raise a good point - the fedmsg results
will likely know about builds and not updates (bundles of builds).

Perhaps @AdamWill, @kparal, or @tflink will correct me, but I think
that Taskotron results can be about individual builds or about updates
(is that right?)

We need to make a decision about all of this before we proceed. A few
ideas:

* We could give Greenwave knowledge about updates. If Greenwave knew
  that a particular update had a collection of builds in it and saw
  that there were partial but not complete results, it could wait for
  more information before sending a fedmsg. This way Bodhi would only
  hear results about the thing it most cares about - updates.

* We could use fedmsg to know that a decision might be ready. Bodhi
  could see that new test results became available for a build on one
  of its updates, and could then poll Greenwave via its REST API to see
  what it thinks about the complete update. A potential downside
  appears here though: what if an update has 2 builds, and the first
  one's results come in, but greenwave's tests only pass if both builds
  are done? In this case, Bodhi would mark the update as failed after
  the first result comes in. When the second one comes, Bodhi could
  mark it as passed, but it would be confusing for users in the
  meantime. Keep in mind that people will probably also want to watch
  CI gating decisions here, so that intermediate failure state would be
  very confusing. We probably really don't want to send "failed" or
  "passed" messages about an update until the final results are in.

I think I prefer the first idea because it better represents what
really needs to be tested, and it avoids the complications I mentioned
on the second. Thoughts? Other ideas?


[0] https://pagure.io/greenwave/issue/24

pypingou:
>  * We could use fedmsg to know that a decision might be ready. Bodhi   could see that new test results became available for a build on one   of its updates, and could then poll Greenwave via its REST API to see   what it thinks about the complete update. A potential downside   appears here though: what if an update has 2 builds, and the first   one's results come in, but greenwave's tests only pass if both builds   are done?

Then greenwave should return a PENDING, because we're asking for two builds and it only has enough info to make a decision for one of them. So it shouldn't return a "fail" but a "come back later", imho.

bowlofeggs:
What will happen for builds that can only pass (or fail) together? A common example of this is python-requests and python-urllib3. They often get released together, and testing and update of either one by itself would cause a failure where testing them together would be a success. In Fedora these are generally put into a Bodhi update together so they go out atomically. Will we not have a way to deal with situations like this with the CI pipeline? If not, I worry that Greenwave will always tell Bodhi that the tests failed for updates like this, where they wouldn't fail if they had been tested together.

Perhaps it makes more sense to test and gate proposed updates in Bodhi, and let the CI pipeline results be used only for Pagure pull request automated feedback? Bodhi could still use Greenwave to make decisions about Taskotron test results (which I do believe can handle these complex multi-build scenarios).

Alternatively, we could create a way to relate pull requests in Pagure so the CI pipeline knows to test them together?

bowlofeggs:
On Fri, 2017-07-28 at 13:54 -0700, Pierre-Yves Chibon wrote:
> Then greenwave should return a PENDING, because we're asking for two
> builds and it only has enough info to make a decision for one of
> them. So it shouldn't return a "fail" but a "come back later", imho.

Yeah that could work.

matt8754:
> Then we decided that it did make sense to say that the gating decision should be on the update instead of the build, because there are real world scenarios where updates would fail independently, but would pass together (or worse, vice-versa!) So that convinced me that we should gate based on the collection of builds in an update rather than on each build individually.

Hmm, this is something I was not aware of.


> We could give Greenwave knowledge about updates. If Greenwave knew   that a particular update had a collection of builds in it and saw   that there were partial but not complete results, it could wait for   more information before sending a fedmsg. This way Bodhi would only   hear results about the thing it most cares about - updates.

If I understand correctly, does this mean Bodhi would publish a message saying "All the tests are complete and the update is waiting for decision"? So Greenwave could listen to this message and publish a corresponding decision?  That sounds like a good approach to me.

matt8754:
> Then greenwave should return a PENDING, because we're asking for two builds and it only has enough info to make a decision for one of them. So it shouldn't return a "fail" but a "come back later", imho.

The problem is how could Greenwave know the update has completed all the testings or not? ResulstsDB can't tell about this. On the other hand, Greenwave only answers yes/no. As to "SKIPPING", it returns a message as "No tests are required" and the caller can use it to know that no tests are reuired on the update so the test gating is skipped. Now, if we want it to return a "PENDING",  the message could be sth like "Tests are not complete". In order to do this, Greenwave needs to know how many of tests are complete/incomplete on the update.

pypingou:
> The problem is how could Greenwave know the update has completed all the testings or not?

GW knows what are the test required so as long as it doesn't have information about all of these, it knows it is missing information. If you require CI tests + taskotron tests A and B but you only have results about taskotron test A, you know you're missing CI and test B. Does that make sense?

pypingou:
> What will happen for builds that can only pass (or fail) together? A common example of this is python-requests and python-urllib3. 

This is something that was thought of and discussed early in the process, but I don't want to reply wrongly, so I would defer to @arilivigni to give more insights about how the pipeline is planning to handle this (he comes back to work only today so I'd expect him to be swamped in emails and notifications so it may take a day or two for him to reach this one).

bowlofeggs:
> If I understand correctly, does this mean Bodhi would publish a message saying "All the tests are complete and the update is waiting for decision"? So Greenwave could listen to this message and publish a corresponding decision? That sounds like a good approach to me.

I don't think Bodhi would know that the tests are complete - it would need to be Greenwave that knew that the tests were complete and that a decision was ready.

matt8754:
> GW knows what are the test required so as long as it doesn't have information about all of these, it knows it is missing information. If you require CI tests + taskotron tests A and B but you only have results about taskotron test A, you know you're missing CI and test B. Does that make sense?

Yes, but this does not handle a situation where a test is *really* missing. For example, Greewave requires test A + test B to be passed, but only test A has run not test B. That could be because we forgot to configure the testing system to run test B. Am I overthinking in this case?

matt8754:
> I don't think Bodhi would know that the tests are complete - it would need to be Greenwave that knew that the tests were complete and that a decision was ready.

Okay, I have come out an approach. For instance, when a new result/waiver for glibc is published and the test results are incomplete, Greenwave will publish a message like this:

{ i': 1,
  'msg': { 
      'decision': {
          'policies_satisified': false,
         'summary': '1 of 15 required tests failed',
         'applicable_policies': ['1'],
         'unsatisfied_requirements': [
             {
                   'item': "glibc-1.0-1.f26",
                   'result_id': "123",
                   'testcase': 'dist.depcheck',
                   'type': 'test-result-missing'
            }
        ]
      },
    # the context on which this decision is made.
     'context': {
         'item': 'glibc-1.0-1.f26',
         'result_id': 1,
         'waiver_id': 1
    }
  },
  ....
}

When Bodhi sees this message, it knows that the test results for the glibc are incomplete by checking if there is any item having 'test-result-missing' in the 'unsatisfied_requirements' field. We probably could add a filed to make this step simple, indicating the results are not complete. And then, Bodhi will ignore the decision and set the test_gating_status as ' PENDING'. How does this sound?

pypingou:
> That could be because we forgot to configure the testing system to run test B. Am I overthinking in this case?

My take on this is: as long as we document clearly and thoroughly the behavior, this is a problem for the test-pipeline not for greenwave.

bowlofeggs:
> When Bodhi sees this message, it knows that the test results for the glibc are incomplete by checking if there is any item having 'test-result-missing' in the 'unsatisfied_requirements' field. We probably could add a filed to make this step simple, indicating the results are not complete. And then, Bodhi will ignore the decision and set the test_gating_status as ' PENDING'. How does this sound?

This sounds reasonable to me.

bowlofeggs:
On Tue, 2017-08-01 at 01:17 -0700, Pierre-Yves Chibon wrote:
> My take on this is: as long as we document clearly and thoroughly the
> behavior, this is a problem for the test-pipeline not for greenwave.

I agree - Greenwave won't be blamed if such a thing happens ☺

alembic/versions/33a416e03f50_add_greenwave_fields.py
```diff
@@ -0,0 +1,41 @@
...
+            Enum(
+                'ignored', 'queued', 'running', 'passed', 'failed', 'waiting',
+                name='ck_test_gating_status'),
+            nullable=True
+        )
+    )
+    op.add_column('updates', Column('greenwave_summary_string', Unicode(255)))
+    op.add_column(
+        'updates',
+        Column('test_gating_url', UnicodeText, nullable=True)
```
bowlofeggs, line 33: This column does not appear to be used in the rest of the code. Should we drop it?

matt8754: Yeah, this was a mistake.

bowlofeggs:
I have opened #1733 that has the commits from this PR and some of my own as well. It's still a WIP as I have a few more things to do before it is ready to merge, but feel free to take a look at what I've been up to.

PR #1698, opened 2017-07-24
Add more query parameters to client overrides query

### LINKED ISSUE #1679: The CLI seems to have lost the ability to query overrides by attributes other than user
It used to be possible to query overrides by package name, and possibly other attributes. The current CLI only allows querying by user:

```
$ bodhi overrides query --help                                  
Usage: bodhi overrides query [OPTIONS]                     

  Query the buildroot overrides.                           

Options:                                                   
  --user TEXT  Updates submitted by a specific user        
  --staging    Use the staging bodhi instance              
  --url TEXT   URL of a Bodhi server. Ignored if --staging is set. Can be set                                          
               with BODHI_URL environment variable         
  --help       Show this message and exit.
```

I'm going to mark this as high priority since it is a regression.

ryanlerch:
This adds support in the Bodhi CLI for querying buildroot overrides by expired/active status, release, and package name.

Fixes #1679

### DIFF
--- a/bodhi/client/__init__.py
+++ b/bodhi/client/__init__.py
@@ -499,13 +499,20 @@ def overrides():
 
 @overrides.command('query')
 @click.option('--user', default=None,
-              help='Updates submitted by a specific user')
+              help='Overrides submitted by a specific user')
 @click.option('--staging', help='Use the staging bodhi instance',
               is_flag=True, default=False)
 @click.option('--mine', is_flag=True,
               help='Show only your overrides.')
+@click.option('--packages', default=None,
+              help='Query by comma-separated package name(s)')
+@click.option('--expired/--active', default=None,
+              help='show only expired or active overrides')
+@click.option('--releases', default=None,
+              help='Query by release shortname(s). e.g. F26')
 @url_option
-def query_buildroot_overrides(url, user=None, mine=False, **kwargs):
+def query_buildroot_overrides(url, user=None, mine=False, packages=None,
+                              expired=None, releases=None, **kwargs):
     # Docs that show in the --help
     """
     Query the buildroot overrides.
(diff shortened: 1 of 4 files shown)


bodhi/client/__init__.py
```diff
@@ -499,13 +499,20 @@ def overrides():
...
 @overrides.command('query')
 @click.option('--user', default=None,
-              help='Updates submitted by a specific user')
+              help='Overrides submitted by a specific user')
 @click.option('--staging', help='Use the staging bodhi instance',
               is_flag=True, default=False)
 @click.option('--mine', is_flag=True,
               help='Show only your overrides.')
+@click.option('--packages', default=None,
+              help='Query by package name(s)')
```
bowlofeggs: How about "Query by comma-separated package name(s)"?

bowlofeggs (approved):
Looks great as is, so feel free to merge (even if you decide to mention the comma-separated thing, no need for re-review).

I do have a suggestion on one more thing that could be queried that I think could be useful (and I'd recommend opening a new PR for it after this is merged): what if we added a ```--builds``` so it'd be easy to query for a specific build? Right now, you can get it to display a single override (it gives more info when the query matches a single override) by doing something like this:

```
$ bodhi overrides query --active --packages nspr --releases f26
============================================================
     nspr-4.15.0-1.fc26
============================================================
  Submitter: ueno
  Expiration Date: 2017-07-25 00:00:00
  Notes: for nss-util
  Expired: False
1 overrides found (1 shown)
```

But it'd be nice to just say:

```
$ bodhi overrides query --builds nspr-4.15.0-1.fc26
============================================================
     nspr-4.15.0-1.fc26
============================================================
  Submitter: ueno
  Expiration Date: 2017-07-25 00:00:00
  Notes: for nss-util
  Expired: False
1 overrides found (1 shown)
```

ryanlerch:
@bowlofeggs just FYI, implemented the --builds flag you mentioned above in #1707

PR #1699, opened 2017-07-24
PEP-257 bodhi.server.services.user.

bowlofeggs:


ryanlerch (approved):
Looks great to me, @bowlofeggs ! more docs!!! 🎉

PR #1703, opened 2017-07-24
Fix inline literals in developers_docs.rst

mprahl:


ryanlerch:
jenkies test

ryanlerch (approved):
Looks good to me!

Thanks for the patch!

PR #1705, opened 2017-07-24
Add the ability for 'get_pkg_committers_from_pagure' to return FAS groups

mprahl:
* Add the ability for `get_pkg_committers_from_pagure` to return FAS groups
* Fix the docstring of `get_pkg_committers_from_pagure`

### DIFF
--- a/bodhi/server/models.py
+++ b/bodhi/server/models.py
@@ -633,8 +633,12 @@ class Package(Base):
         * The second list contains FAS group names that have commit access.
         """
         pagure_url = config.get('pagure_url')
-        # Pagure uses plural names for its namespaces such as "rpms"
-        namespace = self.type.name + 's'
+        # Pagure uses plural names for its namespaces such as "rpms" except for
+        # container
+        if self.type.name == 'container':
+            namespace = self.type.name
+        else:
+            namespace = self.type.name + 's'
         package_pagure_url = '{0}/api/0/{1}/{2}'.format(
             pagure_url.rstrip('/'), namespace, self.name)
         package_json = pagure_api_get(package_pagure_url)
--- a/bodhi/server/models.py
+++ b/bodhi/server/models.py
@@ -626,12 +626,11 @@ class Package(Base):
 
     def get_pkg_committers_from_pagure(self):
         """
-        Pull users who can commit and are watching a package in Pagure.
+        Pull users and groups who can commit on a package in Pagure.
 
-        Return two two-tuples of lists:
-        * The first tuple is for usernames.  The second tuple is for groups.
-        * The first list of the tuple is for committers. The second is for
-          watchers.
+        Returns a tuple with two lists:
+        * The first list contains usernames that have commit access.
+        * The second list contains FAS group names that have commit access.
         """
         pagure_url = config.get('pagure_url')
         # Pagure uses plural names for its namespaces such as "rpms"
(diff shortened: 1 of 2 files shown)


ryanlerch:
jenkies test

mprahl:
@bowlofeggs can you please review this when you get a chance?

bodhi/server/models.py
```diff
@@ -626,14 +626,17 @@ def get_pkg_pushers(self, branch, settings):
...
+        Pull users and groups who can commit on a package in Pagure.
 
-        Return two two-tuples of lists:
-        * The first tuple is for usernames.  The second tuple is for groups.
-        * The first list of the tuple is for committers. The second is for
-          watchers.
+        Returns a tuple with two lists:
+        * The first list is for usernames that have commit access. This list
+        will contain members from Pagure groups if `pagure_uses_fas_groups` is
+        set to True.
```
bowlofeggs: Is this correct? The code below does not seem to consider the ```pagure_uses_fas_groups``` setting when adding users to the first list.

bodhi/server/models.py
```diff
@@ -626,14 +626,17 @@ def get_pkg_pushers(self, branch, settings):
...
+        * The second list is for FAS groups that have commit access. This will
+        be empty if `pagure_uses_fas_groups` is set to False.
         """
         pagure_url = config.get('pagure_url')
+        pagure_uses_fas_groups = config.get('pagure_uses_fas_groups')
```
bowlofeggs: It seems odd to me to make this a setting. It seems that Pagure either would or would not use FAS groups, and IMO we should just code Bodhi to expect whichever behaviour Pagure has.

bowlofeggs (changes requested):

bodhi/server/models.py
mprahl: @bowlofeggs the setting itself is configurable in Pagure, but I can remove the setting and just assume they are FAS groups.

bodhi/server/models.py
mprahl: You are right. Let me fix that.

mprahl:
@bowlofeggs can you review again please? I also added a second commit with a non-related fix to that function.

mprahl:
Tests are passing now.

bodhi/server/models.py
```diff
@@ -633,8 +633,14 @@ def get_pkg_committers_from_pagure(self):
         * The second list contains FAS group names that have commit access.
         """
         pagure_url = config.get('pagure_url')
-        # Pagure uses plural names for its namespaces such as "rpms"
-        namespace = self.type.name + 's'
+        # Pagure uses plural names for its namespaces such as "rpms" except for
+        # container
+        # This if statement is not tested because there isn't direct support
+        # for containers in Bodhi. This is just future proofing.
+        if self.type.name == 'container':  # pragma: no cover
```
bowlofeggs: We should use mock to test this code to ensure that it uses the correct namespace. Even though ```self.type``` doesn't know about containers yet, it wouldn't be hard in a test to construct a Package obect and then set it's type attribute to a mock object that have a name attribute that is set to "container".

Also, this commit doesn't seem related to the first commit in the pull request. I recommend splitting this out into a separate pull request to separate concerns. It'll make it easier for you to respond to review comments, and it will also make it easier to review the changes for me. Splitting it is not required, just a recommendation.

bowlofeggs (changes requested):
The first commit looks OK to me, but we need test coverage on the second one.

bodhi/server/models.py
mprahl: Okay, I added a test for this. Thanks for the review!

mprahl:
@bowlofeggs can you review again please?

bowlofeggs (approved):

Issue #870, opened 2016-07-14
nothing on search results page for "selinux-policy" unless entered in full


oishommau:
entering "selinux" produces no results (https://bodhi.fedoraproject.org/updates/?packages=selinux)

entering up to "selinux-polic" still produces no results (https://bodhi.fedoraproject.org/updates/?packages=selinux-polic)

you can only get results for selinux-policy if you type the whole thing in (https://bodhi.fedoraproject.org/updates/?packages=selinux-policy)

note that it does however show up in the suggestions dropdown for any of the above cases

oishommau:
originally mentioned in #692 (but apparently unrelated to that bug, since not addressed by the fix) -- it's also not possible to find selinux-policy-targeted at all

looking at this again now, it seems like selinux-policy-targeted is a phantom package tied to the selinux-policy package? can anyone explain what's going on here?

should this be a separate bug, or is it related to the issue originally described?

if it's some kind of "alias", perhaps bodhi should handle it accordingly.

bowlofeggs:
Hello @oishommau!

I think what you might be experiencing is the difference between what RPM would think of as a "source package" and a "subpackage". A source package has a 1:1 relationship with a spec file. If you look for selinux-policy in [pkgdb](https://admin.fedoraproject.org/pkgdb/package/rpms/selinux-policy/), you can see that pkgdb is like Bodhi in that it knows about source packages.

However, a single spec file can generate more than one RPM that end users can install. And if you look in Koji at a [build](https://koji.fedoraproject.org/koji/buildinfo?buildID=919436) from selinux-policy, you can see that it results in several RPM packages for end-users to interact with, including selinux-policy-targeted.

Unfortunately, Bodhi does not currently know about subpackages and concerns itself only with source packages. This is why the search doesn't find what you are looking for. I could see adding a feature to give Bodhi the ability to find source packages based on subpackage name searches, but I would consider that more of an RFE than a bug.

In the meantime, it is possible to ask dnf to tell you the source package than generated an RPM:

```
$ dnf repoquery -s selinux-policy-targeted
selinux-policy-3.13.1-190.fc24.src.rpm
selinux-policy-3.13.1-191.24.fc24.src.rpm
```

Sorry for the confusion, but I hope that dnf trick helps!

oishommau:
thanks for the explanation, however, that still doesn't seem to explain the original issue described -- that a search for "selinux" doesn't produce any results at all. should i split one of these issues into a separate issue?

[...]

PR #1706, opened 2017-07-25
use the ilike search for search on enter in searchbar

ryanlerch:
Previously, if the user skipped the search suggestions
in the searchbar in the webui, and just pressed enter,
we returned results for whole packages only
(i.e. /updates/?packages=).

This commit changes the endpoint we point to when the
user presses enter on the search to be /updates/?search=
and now more obvious results are returned.

Fixes: #870

### DIFF
--- a/bodhi/server/static/js/search.js
+++ b/bodhi/server/static/js/search.js
@@ -176,7 +176,7 @@ $(document).ready(function() {
         if (e.which == 13) {
             cabbage.spin();
             var value = $("form#search .tt-input").val();
-            window.location.href = '/updates/?packages=' + encodeURIComponent(value);
+            window.location.href = '/updates/?search=' + encodeURIComponent(value);
         }
     });
 });


oishommau:
unless there's some other relevant change, this appears not to work at all -- the ?search= results just mirror the list of updates added to bodhi (https://bodhi.fedoraproject.org/updates/)

See that all of the following output the same irrelevant results:
https://bodhi.fedoraproject.org/updates/?search=selinux
https://bodhi.fedoraproject.org/updates/?search=firefox
https://bodhi.fedoraproject.org/updates/?search=pluma

bowlofeggs:
Hello @oishommau,

I think this pull request depends on another pull request that was recently merged but not yet deployed to production: https://github.com/fedora-infra/bodhi/pull/1687. Thus, the only way to test this pull request right now is to use the development environment.

ryanlerch:
@bowlofeggs is correct. #1687 implemented the ?search= parameter.

Note that if you use just parameter that bodhi doesnt know, you are just returned all the updates in the list, so this just returns all updates (just like using search= on prod right now.):

https://bodhi.fedoraproject.org/updates/?pants=selinux

bowlofeggs:
This does not seem to make the search box use search when I try it - it still seems to use ```packages=```. I've tried ctrl-f5 to ensure that my browser cache isn't to blame. When I search "bOdhi", I am taken to http://localhost:6543/updates/?packages=bOdhi instead of http://localhost:6543/updates/?search=bOdhi

bowlofeggs:
Nevermind my previous comment - I had checked out your branch incorrectly. I have a git config that lets me check out pull requests with ```git checkout pr/1706```. I absentmindedly had added a ```-b``` flag on there, so I didn't actually have your change. So I'm a doofus.

With this change I see that it does work!

PR #1707, opened 2017-07-25
Add builds parameter for overrides list

ryanlerch:
This commit adds the new `builds` parameter when querying overrides.
This allows the overrides to be queried by a comma-seperated list of
build NVRs. For example:
`/overrides/?builds=embree-2.16.4-1.fc24,embree-2.16.4-1.fc25`

This commit also adds support for using this new parameter in the client
CLI, adding the `--builds` parameter to `bodhi overrides query`


bowlofeggs (approved):
Fantastic, thanks!

Issue #1626, opened 2017-06-19
'NoneType' object has no attribute 'name' when editing an expired override with the CLI


bowlofeggs:
I received the following traceback when trying to edit an expired override (to extend/unexpire it):

```
$ bodhi overrides edit python-fedora-0.9.0-3.el7 --duration 2 --notes "Needed for Bodhi 2.8.0."
Traceback (most recent call last):
  File "/usr/bin/bodhi", line 11, in <module>
    load_entry_point('bodhi-client==2.8.0b1', 'console_scripts', 'bodhi')()
  File "/usr/lib/python2.7/site-packages/click/core.py", line 722, in __call__
    return self.main(*args, **kwargs)
  File "/usr/lib/python2.7/site-packages/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/usr/lib/python2.7/site-packages/click/core.py", line 1066, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/lib/python2.7/site-packages/click/core.py", line 1066, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/lib/python2.7/site-packages/click/core.py", line 895, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/lib/python2.7/site-packages/click/core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File "/usr/lib/python2.7/site-packages/bodhi/client/__init__.py", line 477, in edit_buildroot_overrides
    _save_override(url=url, user=user, password=password, staging=staging, edit=True, **kwargs)
  File "/usr/lib/python2.7/site-packages/bodhi/client/__init__.py", line 121, in _save_override
    expired=kwargs.get('expire', False))
  File "/usr/lib/python2.7/site-packages/bodhi/client/bindings.py", line 97, in wrapper
    raise BodhiClientException(problems)
bodhi.client.bindings.BodhiClientException: Unable to save buildroot override: 'NoneType' object has no attribute 'name'
```

The server also had a traceback, which revealed that I had not authenticated:

```
2017-06-19 18:45:58,003 ERROR [bodhi.server][Dummy-5] 'NoneType' object has no attribute 'name'
Traceback (most recent call last):
  File "/usr/lib/python2.7/site-packages/bodhi/server/services/overrides.py", line 186, in save_override
    submitter = User.get(request.user.name, request.db)
[...]

PR #1709, opened 2017-07-25
Make errors nicer for users in Bodhi CLI

### LINKED ISSUE #1126: Authentication errors print a traceback in the bodhi CLI

ryanlerch:
This patch makes errors nicer to read for users on the Bodhi Client CLI.
Instead of tracing back on AuthError and BodhiClientExceptions, we now
just print the error message.

Fixes #1126
Related #1626

### DIFF
--- a/bodhi/client/__init__.py
+++ b/bodhi/client/__init__.py
@@ -18,10 +18,12 @@ import subprocess
 import sys
 import traceback
 import re
+import functools
 
 import click
 
 from bodhi.client import bindings
+from fedora.client import AuthError
 
 
 def _warn_if_url_and_staging_set(ctx, param, value):
@@ -101,6 +103,21 @@ def add_options(options):
     return _add_options
 
 
+def handle_errors(method):
+    """ A decorator that echoes neat error messages on AuthError or BodhiClientException. """
+    @functools.wraps(method)
+    def wrapper(*args, **kwargs):
+        try:
+            method(*args, **kwargs)
+        except AuthError as e:
+            click.echo("%s: Check your FAS username & password" % (e))
+            sys.exit(1)
+        except bindings.BodhiClientException as e:
+            click.echo(str(e))
+            sys.exit(2)
+    return wrapper
+
+
 def _save_override(url, user, password, staging, edit=False, **kwargs):
     """
     Helper function to create or edit a buildroot override.
@@ -157,6 +174,7 @@ def updates():
 @add_options(new_edit_options)
 @click.argument('builds')
 @click.option('--file', help='A text file containing all the update details')
+@handle_errors
 @url_option
 def new(user, password, url, **kwargs):
     # User Docs that show in the --help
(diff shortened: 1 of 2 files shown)


bodhi/client/__init__.py
```diff
@@ -101,6 +103,21 @@ def _add_options(func):
...
+    @functools.wraps(method)
+    def wrapper(*args, **kwargs):
+        try:
+            method(*args, **kwargs)
+        except AuthError as e:
+            click.echo("%s: Check your FAS username & password" % (e))
+            sys.exit(1)
+        except bindings.BodhiClientException as e:
+            click.echo(str(e))
+            sys.exit(-1)
```
bowlofeggs: Technically, the exit code is an unsigned int. Though this does not cause a real failure, the resulting exit code is also not -1:

```
$ python -c "import sys; sys.exit(-1)"
$ echo $?
255
```

If you want to use a unique exit code here, you could just pick any unused code. I think only 1 is used here, so you could use 2.

bowlofeggs (changes requested):
Looks good, but before you merge please change the exit code to a positive 8-bit number (between 1 and 255). No need for re-review once you've made that change - feel free to merge!

Idea: you could make the error text be printed in bold red text. If you like that idea, I suggest doing it in a new PR ☺

bodhi/client/__init__.py
ryanlerch: thanks! changed it to 2 in the test as well. The reason i set it to -1 was that the test was asserting the error code be -1, so i just assumed it was a valid exit code.

PR #1712, opened 2017-07-26
Make updates on users page only show current

### LINKED ISSUE #963: Do not show updates from unsupported Fedora versions in users' profile pages
jskarvad reported that Bodhi is displaying updates for Fedora versions that are no longer supported on his user profile. In his case, these are Fedora 21 updates that were never pushed to stable, and Fedora 21 went end of life many months ago. He requests that Bodhi not display updates for non-supported Fedora releases in user profiles, which I think is reasonable.

This was originally reported in [Fedora Infrastructure Trac ticket #5489](https://fedorahosted.org/fedora-infrastructure/ticket/5489).

ryanlerch:
This commit makes the recent updates list on a users page only show
updates from current releases.

Fixes #963

### DIFF
--- a/bodhi/server/templates/user.html
+++ b/bodhi/server/templates/user.html
@@ -21,15 +21,29 @@
               },
           });
       $.ajax(
-          "${urls['recent_updates']}" + "&chrome=False&rows_per_page=12&display_user=False",
+          'releases/',
           {
-              dataType: 'html',
-              success: function(html) {
-                  $('#recent_updates').parent().append(html);
-                  $('#recent_updates').remove();
-              },
+              dataType: 'json',
+              success: function(e) {
+                  var releases_string = ""
+                  for (var i = 0; i < e.releases.length; i++) {
+                      if (e.releases[i].state == 'current'){
+                        releases_string = releases_string + ","+e.releases[i].name;
+                      }
+                  }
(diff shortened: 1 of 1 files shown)


bowlofeggs (changes requested):
The change works for the user profile page, but I'll note that if the user clicks on the testing link in the upper right to expand and see all current testing updates, they will still see the old releases there. For example, that URL on the original reporter's profile will take you to http://localhost:6543/updates/?user=jskarvad&status=testing which still displays F21 updates.

I think we should do something to the objects to remove them from "testing" when a release gets archived, since updates are no longer truly being tested for F21. IMO, that should be the true fix for the issue.

There is a CLI tool called ```bodhi-manage-releases``` that releng uses when they want to archive a release. That tool makes API calls to Bodhi. I think we should have that API kill all non-stable updates when a release is archived, perhaps by revoking them.

As for the existing objects, I think we can write a database migration to do the same. Fixing the API will fix the issue for the future, and the migration will get the existing data to be consistent.

I think this approach will ensure that the user's profile page shows the right data while also making all other queries show the right data.

Sound good to you?

ryanlerch:
okies, will close this PR, since the real way to fix is pretty different to my workaround type solution here.

PR #1715, opened 2017-07-26
Test bodhi.server.metadata

bowlofeggs:
This pull request has three commits on ```bodhi.server.metadata```. The first removes some EL 5 specific code that is no longer needed since EPEL 5 is retired. The second adds some test coverage to the module. The third removes a feature that breaks Bodhi's repositories in a devastating way if it is enabled.

Overall, all three of these also increase Bodhi's test coverage (either by adding tests or removing untested code.


jeremycline (approved):
Nice, tidy commits.

:shipit:

PR #1713, opened 2017-07-26
fix bindings.get_releases()

ryanlerch:
previously, bodhi.client.bindings.get_releases() was failing, always
returning a 403. This commit fixes this issue

Fixes #784

Issue #784, opened 2016-02-25
bodhi.get_releases() is now broken


kparal:
Using `python-fedora-0.7.1-4.fc23.noarch`:

```
In [1]: import fedora.client

In [2]: bodhi =  fedora.client.BodhiClient()

In [3]: bodhi.get_releases()
---------------------------------------------------------------------------
BodhiClientException                      Traceback (most recent call last)
<ipython-input-3-edab0102d5c5> in <module>()
----> 1 bodhi.get_releases()

/usr/lib/python2.7/site-packages/fedora/client/bodhi.pyc in wrapper(*args, **kwargs)
     94         except Exception:
     95             pass
---> 96         raise BodhiClientException(problems)
     97     return wrapper
     98 

BodhiClientException: csrf_token is missing
name is missing
long_name is missing
branch is missing
id_prefix is missing
dist_tag is missing
stable_tag is missing
testing_tag is missing
candidate_tag is missing
override_tag is missing
```

This works when I downgrade to `python-fedora-0.5.5-1.fc23.noarch`.

lmacken:
Very weird, since `curl https://bodhi.fedoraproject.org/releases/` works fine without any arguments or a CSRF token...

ryanlerch:
fedora.client.BodhiClient() is now deprecated. however, using the replacement bodhi.client.bindings still fails, returning the following server error:

```
ServerError: ServerError(https://bodhi.fedoraproject.org/releases/, 403, Error returned from json module while processing https://bodhi.fedoraproject.org/releases/: Expecting value: line 6 column 1 (char 5) 
```

The steps to get the error are:

```
In [1]: import bodhi.client                                                                     
In [2]: client = bodhi.client.bindings.BodhiClient()
In [3]: client.get_releases()
```

bowlofeggs:
The patch for this issue should be included in the upcoming 2.10.0 release:

https://github.com/fedora-infra/bodhi/pull/1744

kparal:
Should it be fixed? I still see it with `bodhi-2.10.0b0`:

```
In [10]: import bodhi.client
In [11]: client = bodhi.client.bindings.BodhiClient(staging=True)
In [12]: client.get_releases()
...
ServerError: ServerError(https://bodhi.stg.fedoraproject.org/releases/, 403, Error returned from json module while processing https://bodhi.stg.fedoraproject.org/releases/: Expecting value: line 6 column 1 (char 5)
```

[...]

PR #1716, opened 2017-07-27
Test bodhi.client.bindings

bowlofeggs:
This pull request has 5 atomic commits that focus on adding test coverage to ```bodhi.client.bindings```. They also remove some unusable code, and fix two issues. The commit messages give more detail.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -6,7 +6,7 @@ omit =
 
 [report]
 precision = 2
-fail_under = 88
+fail_under = 89
 exclude_lines =
     pragma: no cover
     def __repr__
--- a/bodhi/client/bindings.py
+++ b/bodhi/client/bindings.py
@@ -600,8 +600,7 @@ class BodhiClient(OpenIdBaseClient):
         This method is a generator that returns a list of koji builds that
         could potentially be pushed as updates.
         """
-        if not self.username:
-            raise BodhiClientException('You must specify a username')
+        self.init_username()
         builds = []
         data = self.get_releases()
         koji = self.get_koji_session()
--- a/bodhi/client/bindings.py
+++ b/bodhi/client/bindings.py
@@ -36,6 +36,7 @@ import re
 import textwrap
 
 from six.moves import configparser
+import dnf
 import six
 
 from fedora.client import AuthError, OpenIdBaseClient, FedoraClientError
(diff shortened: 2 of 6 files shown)


bowlofeggs:
Ah, I seem to have forgotten that dnf is only installed on Fedora systems. It's bed time now, but I'll make dnf an optional dependency and fix this PR up soon.

Or, we could make dnf a dependency as it is available on EL 7…

ryanlerch (commented):
closed #1713 which did the same thing as this commit, but are you able to add Fixed #784 to this commit?

bowlofeggs:
@ryanlerch Haha yeah I noticed after submitting this PR that we had made tests and fix for the same code, but it was midnight and I figured I'd inspect the two to see what to do later. It seems like an unlikely coincidence! Whenever I go on test writing rampages, I use the random module to pick some code that isn't tested and it just happened to pick testing that code. I noticed pretty quickly that it didn't work (with the 403) and that it just needed a GET.

Sure, I'll amend the commit message to reference #784.

On the topic of the CI failure - I've slept on it and I think I lean towards making dnf a requirement. Technically, it was a requirement for that function before anyway and was just undeclared. So if EPEL 7 users tried to use that function, they'd get a traceback. Since dnf is available in EPEL 7 I think it's better to just make it a requirement and call it a day. I'm going to make that change too, but feel free to counterpoint if you disagree ☺

bowlofeggs:
This most recent failure was due to https://bugzilla.redhat.com/show_bug.cgi?id=1476364. I will adjust the commit so that dnf isn't listed as a dependency in ```setup.py``` (even though it really is a dependency).

bowlofeggs:
Ugh, and now a mock that works on F25 doesn't work on EL 7:

```
AssertionError: Expected call: listTagged('f26-updates-testing', latest=True) Actual call: listTagged("f<MagicMock name='open().readlines().__getitem__().split().__getitem__()' id='83276240'>-updates-testing", latest=True)
```

bowlofeggs:
Ah, this is actually a mock bug that's been fixed upstream but not in EPEL 7 ☹ I started thinking that it felt familiar and then I remembered that I've been down this exact road before. ```mock_open()```'s ```readlines()``` implementation was not correct and hasn't been fixed. I'll have to write my own mock instead of using ```mock_open()```.

bowlofeggs:
I have now resolved the CI issues with this pull request and it is ready for review.

ryanlerch (approved):
This looks good to me, @bowlofeggs! +1 for adding dnf as a formal dep.

also +100 for another percentage point in testing coverage!!! 🌮 🌮 🌮

PR #1720, opened 2017-07-31
Make errors in BodhiCLI red

ryanlerch:
Makes errors in bodhi cli print in red to the user.

Looked into adding tests for this change, but there doesn't seem to be an easy way to check the clirunner output for the color escape codes.


bowlofeggs (approved):
Looks nice! IMO, it's OK not to assert that the color is red since that's not critical behavior. The critical behavior is that it prints the error message (already tested), the color is a nice touch.

I might suggest trying a ```bold=True``` on there too, but that's up to you. Feel free to merge either way!

PR #1728, opened 2017-08-02
Add hovers to the graphs on /metrics/

### LINKED ISSUE #209: Metrics enhancement to show also count/percentage
On page https://bodhi.fedoraproject.org/metrics , is it possible to show total exact count/percentage for various colors by just taking mouse pointer on top of that colored section in the given graph?

thanks.

ryanlerch:
Previously, on the metrics page there was no easy way
for a user to break down the values for each update type
per release. This commit adds hover labels that show the
values for each update type.

Fixes #209

### DIFF
--- a/bodhi/server/templates/metrics.html
+++ b/bodhi/server/templates/metrics.html
@@ -13,6 +13,27 @@
 
 <script type="text/javascript">
   $(document).ready(function(){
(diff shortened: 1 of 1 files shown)


bowlofeggs (approved):
I tested this with QupZilla and it works there too. Nice!

Issue #1406, opened 2017-04-03
Bugzilla links need a space before "rhbz" to be recognized


jsynacek:
When typing an update message, bugzilla links are not correctly recognized:

```(rhbz#12345)``` is **not** recognized as a link in the preview window.

```( rhbz#12345)``` **is** recognized as a link - notice the space between '(' and 'r'.

Expected result:
The link is recognized without the space between the markup string and a word separator.

**Comment by bowlofeggs** [comment]
[This regex](https://github.com/fedora-infra/bodhi/blob/2.5.0/bodhi/server/ffmarkdown.py#L85) is supposed to match bz links:

```
BUGZILLA_RE = r'([\S]+)(#[0-9]{5,})'
```

I don't see anything immediately wrong with it, but this area of the code is probably where the issue is.

bowlofeggs:
The fix for this issue will appear in the upcoming 2.11.0 release:

https://github.com/fedora-infra/bodhi/pull/1790

PR #1730, opened 2017-08-02
fix regex for matching bugtrackers in markdown input

ryanlerch:
Previously, the regex to match bugtracker keywords (e.g. RHBZ#12345)
was using /S to match the bugtracker keyword. Consequently, if the
keyword was wrapped in braces, it would not match e.g. (RHBZ#12345).
This commit makes the regex use [a-zA-Z] to match the keyword now.

Fixes: #1406

### DIFF
--- a/bodhi/server/ffmarkdown.py
+++ b/bodhi/server/ffmarkdown.py
@@ -74,7 +74,7 @@ def inject():
             return el
 
     MENTION_RE = r'(?<!\S)(@\w+)'
-    BUGZILLA_RE = r'([\S]+)(#[0-9]{5,})'
+    BUGZILLA_RE = r'([a-zA-Z]+)(#[0-9]{5,})'
 
     class SurroundProcessor(markdown.postprocessors.Postprocessor):
         def run(self, text):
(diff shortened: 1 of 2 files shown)


bodhi/server/ffmarkdown.py
```diff
@@ -74,7 +74,7 @@ def handleMatch(self, m):
             return el
 
     MENTION_RE = r'(?<!\S)(@\w+)'
-    BUGZILLA_RE = r'([\S]+)(#[0-9]{5,})'
+    BUGZILLA_RE = r'([a-zA-Z]+)(#[0-9]{5,})'
```
jeremycline, line 77: I recommend adding a test case or two using the strings from the bug report to demonstrate the fix

jeremycline (approved):
It looks good to me, although a couple tests wouldn't hurt

:shipit:

bowlofeggs (approved):
LGTM too, but I'd also like a test to demonstrate the fix. I tested it by hand it it appears to work as expected.

ryanlerch:
Thanks for the reviews @jeremycline and @bowlofeggs !

I have added a couple of tests to the commit now, one checking the case of `(RHBZ#12345)`, and one checking a valid prefix that is a substring in a longer string e.g. `aRHBZa#12345`.

bowlofeggs (approved):
Thanks!

PR #1733, opened 2017-08-08
Greenwave integration

bowlofeggs:
This pull request has the original commits from #1694, with some commits from me that adjust those commits and remove CI integration.

There are still a few things that need to be done before this is ready for review.


matt8754:
LGTM, 👍

bowlofeggs:
@jeremycline @pypingou @matt8754: I propose that we go ahead and review this PR and get it merged, despite one flaw that I know of in it. The JavaScript has not yet been updated to use the new Greenwave API, but I want to merge anyway for two reasons:

1. The JS won't be used unless the new test_gating_enabled setting is set to True (defaults to False), so it won't be a regression if we merge it as is.
2. We still don't have Greenwave in staging to test against, so I can't run the JavaScript code to test it to write it the way it needs to be.
3. This PR is already getting too huge. I think we should iterate rather than make this bigger and take longer.

What do you think?

pypingou:
:+1: on reviewing now to keep the diff smaller and easier to review

bowlofeggs:
jenkies test

docs/man_pages/index.rst
```diff
@@ -10,3 +10,4 @@ Man pages
    bodhi-babysit-ci
```
jeremycline: I think you need to remove this entry

bodhi/server/scripts/check_policies.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+@click.command()
+@click.version_option(message='%(version)s')
+def check():
+    """Check the enforced policies by Greenwave for each open update."""
+    initialize_db(config.config)
+    session = Session()
+
+    updates = models.Update.query.filter(models.Update.pushed == false())\
```
jeremycline, line 40: This is fine, but just a heads-up in case you don't know, there's a [filter_by](http://docs.sqlalchemy.org/en/latest/orm/query.html#sqlalchemy.orm.query.Query.filter_by) API as well so you can avoid the odd equality check returning non-bools. It's only really useful when you're checking for equality, but still, it's nice.

bodhi/server/scripts/check_policies.py
```diff
@@ -0,0 +1,76 @@
...
+                # If an unrestricted policy is applied and no tests are required
+                # on this update, let's set the test gating as ignored in Bodhi.
+                if decision['summary'] == 'no tests are required':
+                    update.test_gating_status = models.TestGatingStatus.ignored
+                else:
+                    update.test_gating_status = models.TestGatingStatus.passed
+            else:
+                update.test_gating_status = models.TestGatingStatus.failed
+            update.greenwave_summary_string = decision['summary']
+            if session.is_modified(update):
```
jeremycline, line 67: I would expect the session API to make a commit a no-op if there aren't any objects in a dirty state. It doesn't _hurt_, but I just don't know why it's necessary. What made you choose to do this?

bodhi/server/util.py
```diff
@@ -766,21 +775,32 @@ def get_critpath_components_from_pdc(branch, component_type='rpm'):
...
-    base_error_msg = (
-        'Bodhi failed to get a resource from {0} at the following URL '
-        '"{1}". The status code was "{2}".')
-    rv = requests.get(api_url, timeout=60)
-
+    if method == 'POST':
+        base_error_msg = (
+            'Bodhi failed to send POST request to {0} at the following URL '
+            '"{1}". The status code was "{2}".')
+        rv = requests.post(api_url,
```
jeremycline, line 795: If this code path is even a little bit hot (e.g. it's in the realm of possibility that it would be called twice in 5-30 seconds) you'll benefit from connection pooling via a requests Session object.

jeremycline (approved):
Just a few notes/questions, nothing blocking.

:shipit:

bodhi/tests/server/scripts/test_check_policies.py
```diff
@@ -0,0 +1,146 @@
...
+
+        When test gating is disabled, each Update's test_gating_status will be None.
+        """
+        runner = testing.CliRunner()
+        update = self.db.query(models.Update).all()[0]
+        update.status = models.UpdateStatus.testing
+        update.test_gating_status = None
+        self.db.commit()
+        with patch('bodhi.server.scripts.check_policies.greenwave_api_post') as mock_greenwave:
+            mock_greenwave.return_value = RuntimeError('The error was blablabla')
```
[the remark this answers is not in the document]
bowlofeggs: lol

bodhi/server/util.py
bowlofeggs: I like this suggestion. I do think it will be called many times in a row since a cron job will loop open updates to ask Greenwave for decisions. I guess this will perform suboptimally, but at least it's a background cron job and not a web/response cycle. Since this PR didn't add this code but rather modified it to be able to also POST, I'm inclined to file this as an RFE:

https://github.com/fedora-infra/bodhi/issues/1753

bodhi/server/scripts/check_policies.py
bowlofeggs: @matt8754 originally wrote this code, but may have based it on some code I wrote. I can remove it.

bowlofeggs: I think I'll remove this in a follow-on PR, since it's in @matt8754's commits and I don't want to have to wait for more review on these commits.

bowlofeggs: @jeremycline I filed https://github.com/fedora-infra/bodhi/pull/1754 with this change.

matt8754: Yeah, the code is heriited from babysit_ci.py. I was not aware of the commit() is a no-op, :-P

bodhi/tests/server/scripts/test_check_policies.py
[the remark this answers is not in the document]
matt8754: 😄

bodhi/server/util.py
matt8754: Right, this was suggested in Greenwave before and I somehow forgot using it when writing this part of code, :-)

jeremycline:
This has been cherry-picked back to the 2.10 branch as:
1. 9053c0546fd2ec95dd828c4dae499e4f7b88521f
2. 400c2f80bc95f0ac2bd13a8328bb5529080ae7b3
3. db5b9d7c10abea661a6dd66fea7cf1d5b2a4de13
4. a0286d58add7a2c712f0e9856576f0cfe3ce86ce
5. 1c3239f05713d85b7330a1e95c6727e6b03f79fb
6. 107377537fa8fc1e14edd2ef9e86b8b7ffe0b2d8
7. 80e57917f770bd89728f6b3512ab17e9ca112793
8. 189923b15d1d6b464dd532b1789e70319c2a4620
9. 22ae952ff62bf8e059f8b17159f76a84c0b0f413
10. 04dab0f80bfefaf9c08c14481af539ce4401e177

PR #1742, opened 2017-08-11
Make update form package search use the new typeahead APIs

### LINKED ISSUE #1731: The "Packages" search in the new update form errors with: unable to find any packages that match the current query
It was reported today that the "Packages" field in the new update form does not work correctly. Instead of offering autocomplete matches, an error message "unable to find any packages that match the current query" appears. It seems that the browser is experiencing a CSRF error.

I was unable to reproduce this error at first, but I found that issuing a hard refresh (ctrl-f5) to my browser enabled me to experience the issue. This leads me to believe that it may be due to JS changes in Bodhi 2.9.0.

ryanlerch:
We recently updated to a newer version of the typeahead JS
library, and the API changed. The package search in the
new / edit update form was still using the old calls, so this
commit updates it to use the new API.

Fixes #1731

### DIFF
--- a/bodhi/server/static/js/update_form.js
+++ b/bodhi/server/static/js/update_form.js
@@ -41,8 +41,9 @@ $(document).ready(function() {
         datumTokenizer: Bloodhound.tokenizers.obj.whitespace('value'),
         queryTokenizer: Bloodhound.tokenizers.whitespace,
         remote: {
+            wildcard: '%QUERY',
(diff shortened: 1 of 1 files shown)


bowlofeggs (approved):
Thanks so much!

jeremycline:
This was cherry-picked to the 2.10 branch as 27f41483807673a92dd09dd32112132cde2025e9

PR #1744, opened 2017-08-11
Write release notes for Bodhi 2.10.0.

bowlofeggs:


bowlofeggs:
jenkies test

docs/release_notes.rst
```diff
@@ -1,16 +1,115 @@
...
+* Exact matches are now sorted first in search results
+  (`#692 <https://github.com/fedora-infra/bodhi/issues/692>`_).
+* The CLI now has a ``--mine`` flag when searching for updates or overrides
+  (`#811 <https://github.com/fedora-infra/bodhi/issues/811>`_,
+  `#1382 <https://github.com/fedora-infra/bodhi/issues/1382>`_).
+* The CLI now has more search parameters when querying overrides
+  (`#1679 <https://github.com/fedora-infra/bodhi/issues/1679>`_).
+* The new case insensitive search is also used when hitting enter in the search box in the web UI
+  (`#870 <https://github.com/fedora-infra/bodhi/issues/870>`_).
+* Bodhi is now able to query Pagure for FAS groups for ACL info
```
ryanlerch, line 50: Would it be worth explicitly stating the dist-git pagure instance here, rather than Pagure generally?

ryanlerch (approved):
Looks good to me!

Just one comment inline, but by no means a blocker.

docs/release_notes.rst
bowlofeggs: @ryanlerch Technically Bodhi can work with any Pagure instance. Since Bodhi is used by more than just Fedora, it's probably better to just say we can integrate with Pagure.

jeremycline:
This was cherry-picked to the 2.10 branch as b032bed5596c47376046b098cdd04ea435198828

PR #1745, opened 2017-08-11
Build the docs as part of CI

jeremycline:


bowlofeggs:
Looks like we need a cornice sphinx dep for the CI to pass.

bowlofeggs:
@jeremycline Do you still want to finish this PR?

bowlofeggs:
@jeremycline I inadvertently added docs building as part of https://github.com/fedora-infra/bodhi/pull/1843, because it just calls tox when it runs the tests (and that builds the docs). I think we can close this as a result.

PR #1748, opened 2017-08-12
Add additional test for push to stable button

crungehottman:
Re: the suggestion on pr #1678 

> Write one more test to ensure the "push to stable" button appears when an update is already batched


bowlofeggs (approved):
Fantastic, thanks!

PR #1749, opened 2017-08-12
Correct docblock to read "newpackage update" instead of "security upd…

crungehottman:
…ate"

re: suggestion in #1678


bowlofeggs (approved):
Thanks!

PR #1751, opened 2017-08-12
Don't disable autopush when the update is already requested for stable

### LINKED ISSUE #1570: Don't disable autopush when the update is already requested for stable
@ignatenkobrain reported today that [an update](https://bodhi.fedoraproject.org/updates/FEDORA-2017-92e5a01da8) didn't have the push to stable button, even though Bodhi had commented on the update to say "Bodhi is disabling automatic push to stable due to negative karma. The maintainer may push manually if they determine that the issue is not severe." It turned out that the update had already been submitted for stable by reaching the karma threshold earlier. The comment from Bodhi that autopush was being disabled was confusing and misleading. Bodhi should not disable autopush if the update is already requested for stable, or if the update is already stable.

crungehottman:
fixes #1570

### DIFF
--- a/bodhi/server/models.py
+++ b/bodhi/server/models.py
@@ -2361,7 +2361,8 @@ class Update(Base):
         if self.status not in (UpdateStatus.testing, UpdateStatus.pending):
             return
         # If an update receives negative karma disable autopush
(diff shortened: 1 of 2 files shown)


bowlofeggs (approved):
Thanks!

PR #1752, opened 2017-08-12
Write docblocks for bodhi.server.services.csrf.

bowlofeggs:


ryanlerch (approved):
Looks good to me!

PR #1754, opened 2017-08-14
Remove an unnecessary guard against a call to commit().

bowlofeggs:

This ```if``` statement was unnecessary, as the ```commit()``` will be a no-op if nothing is changed. I am not aware of any testing that would make sense for this change, since the effective behavior of the script is not being changed.


matt8754:
👍

jeremycline (approved):
:shipit:

PR #1759, opened 2017-08-16
Update the Greenwave JavaScript to use the new API.

### LINKED ISSUE #1758: The Greenwave JavaScript uses the old Greenwave API
We need to refactor the [Greenwave JavaScript subject](https://github.com/fedora-infra/bodhi/blob/893a5d225fc941ec1d60c3d76e3912c6428994ba/bodhi/server/templates/update.html#L93) so that it passes objects for the NVRs and and update id, rather than just a plain list of build NVRs.

This [Python code](https://github.com/fedora-infra/bodhi/blob/893a5d225fc941ec1d60c3d76e3912c6428994ba/bodhi/server/scripts/check_policies.py#L44-L46) forms the correct subject for the current Greenwave API as an example of what the JavaScript needs to do:

```
        # See discussion on https://pagure.io/greenwave/issue/34 for why we use these subjects.
        subjects = [{'item': build.nvr, 'type': 'koji_build'} for build in update.builds]
        subjects.append({'item': update.alias, 'type': 'bodhi_update'})
```

bowlofeggs:
The Greenwave JavaScript code needed to be updated to use the new
Greenwave API for update decisions. This commit refactors the
existing Python code to move the subject generating code into the
Update model so that it can be shared between the check_policies
script and the JavaScript, and updates the JavaScript to use a
special _json version of it.

fixes #1758

### DIFF
--- a/bodhi/server/models.py
+++ b/bodhi/server/models.py
@@ -1,5 +1,8 @@
-# encoding: utf-8
-
+# -*- coding: utf-8 -*-
+# Copyright © 2011-2017 Red Hat, Inc. and others.
+#
+# This file is part of Bodhi.
+#
 # This program is free software; you can redistribute it and/or
 # modify it under the terms of the GNU General Public License
 # as published by the Free Software Foundation; either version 2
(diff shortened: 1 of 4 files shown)


pypingou:
Does look as bad as you said it does :)

👍 for me

jeremycline (approved):
:shipit:

jeremycline:
Cherry-picked to 2.10 as 42ef226fc6ed46d787e5a6ac28fa1534e3ee33d3

PR #1760, opened 2017-08-17
PEP-257 bodhi.server.services.overrides.

bowlofeggs:

### DIFF
--- a/bodhi/server/services/overrides.py
+++ b/bodhi/server/services/overrides.py
@@ -1,3 +1,8 @@
+# -*- coding: utf-8 -*-
+# Copyright © 2014-2017 Red Hat, Inc. and others.
+#
+# This file is part of Bodhi.
+#
 # This program is free software; you can redistribute it and/or
 # modify it under the terms of the GNU General Public License
 # as published by the Free Software Foundation; either version 2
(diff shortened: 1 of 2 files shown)


bodhi/server/services/overrides.py
```diff
@@ -182,11 +220,17 @@ def query_overrides(request):
...
+    Create or edit a buildroot override.
 
     This entails either creating a new buildroot override, or editing an
     existing one. To edit an existing buildroot override, the buildroot
     override's original id needs to be specified in the ``edited`` parameter.
+
+    Args:
+        request (pyramid.request): The current web request.
+    Returns:
+        basestring: A JSON response describing the new or edited override.
```
jeremycline: It looks to me like this is returning a dictionary

jeremycline (changes requested):
There's one docblock that looks slightly incorrect (dict rather than some JSON-serialized string). Other than that, this looks good.

:shipit:

bowlofeggs:
@jeremycline I've corrected the docblock and rebased to resolve a merge conflict. How does it look now?

jeremycline (approved):
Go forth and :shipit: with my blessing

PR #1762, opened 2017-08-18
add docstring for client's latest_builds()

ktdreyer:
This method and its parameter was undocumented. Give developers some clues about what it can do.


bowlofeggs:
jenkies test

bowlofeggs (approved):

PR #1767, opened 2017-08-25
bodhi.server.util.call_api now uses connection pooling

### LINKED ISSUE #1753: bodhi.server.util.call_api should use connection pooling
The ```bodhi.server.util.call_api()``` function should use requests connection pooling via a Session object. This will make repeated calls to it more efficient.

Thanks to @jeremycline for the suggestion!

crungehottman:
fixes #1753

### DIFF
--- a/bodhi/server/util.py
+++ b/bodhi/server/util.py
@@ -48,6 +48,8 @@ from bodhi.server.exceptions import RepodataException
 
 _ = TranslationStringFactory('bodhi')
 
+http_session = requests.Session()
+
 
 def header(x):
     """Display a given message as a heading."""
@@ -1162,15 +1164,15 @@ def call_api(api_url, service_name, error_key=None, method='GET', data=None):
         base_error_msg = (
             'Bodhi failed to send POST request to {0} at the following URL '
             '"{1}". The status code was "{2}".')
-        rv = requests.post(api_url,
-                           headers={'Content-Type': 'application/json'},
-                           data=json.dumps(data),
-                           timeout=60)
+        rv = http_session.post(api_url,
+                               headers={'Content-Type': 'application/json'},
+                               data=json.dumps(data),
+                               timeout=60)
     else:
         base_error_msg = (
             'Bodhi failed to get a resource from {0} at the following URL '
             '"{1}". The status code was "{2}".')
-        rv = requests.get(api_url, timeout=60)
+        rv = http_session.get(api_url, timeout=60)
     if rv.status_code == 200:
         return rv.json()
     elif rv.status_code == 500:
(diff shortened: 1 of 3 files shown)


bodhi/server/util.py
```diff
@@ -789,19 +789,21 @@ def call_api(api_url, service_name, error_key=None, method='GET', data={}):
     :param data: a dict of data will be sent to the API server.
     :return: dictionary of the returned JSON or a RuntimeError
     """
+    session = requests.Session()
```
jeremycline: This will have a session (which contain the connection pools) created each time the function is called. This will need to be moved outside the function into the module-level so it's created once at import-time.

jeremycline (changes requested):
There's one minor change that needs to happen here, but it looks good!

✨

bodhi/tests/server/test_util.py
```diff
@@ -232,17 +234,18 @@ def test_get_critpath_components_pdc_success(self, mock_get):
         # We can't verify all the calls made because the URL GET parameters
         # in the URL may have different orders based on the system/Python
         # version.
-        mock_get.assert_called_with(pdc_next_url, timeout=60)
+        (session.return_value.get).assert_called_with(pdc_next_url, timeout=60)
```
bowlofeggs: Are these parenthesis needed?

bowlofeggs (changes requested):
I agree with @jeremycline's suggestion. Additionally, it looks like this needs to be rebased to resolve a merge conflict. Looks good otherwise though!

bodhi/tests/server/test_models.py
```diff
@@ -409,8 +409,9 @@ def test_same_build_types(self):
         # This should not raise any Exception.
         self.db.flush()
 
+    @mock.patch('bodhi.server.util.http_session')
     @mock.patch('bodhi.server.util.requests.get')
-    def test_get_pkg_committers_from_pagure(self, mock_get):
+    def test_get_pkg_committers_from_pagure(self, mock_get, session):
```
bowlofeggs, line 414: Can we drop the mock on ```requests.get``` here? It seems like it isn't needed anymore, unless I am missing something which is possible :)

bowlofeggs: I notice this mock is in place in the other bits of diff in this pull request, so same question on those.

bowlofeggs (changes requested):
I think we can drop all the test mocks on ```request.get``` and ```request.post```, to keep the tests simpler. Looks good otherwise!

bowlofeggs:
I've opened https://github.com/fedora-infra/bodhi/pull/1816 with the commit from this PR and a commit of my own that addresses my comment about removing the old mocks. We can close this PR and track the work over there. Thanks for the contribution!

PR #1769, opened 2017-08-26
Make bodhi distinguish about updates going to testing vs stable when asking greenwave

pypingou:
This will allow us and greenwave to apply different policies for
depending on which repo is update is going (updates-testing vs updates)
which is required to integrate some of the more useful taskotron tests.

### DIFF
--- a/bodhi/server/scripts/check_policies.py
+++ b/bodhi/server/scripts/check_policies.py
@@ -41,9 +41,17 @@ def check():
         .filter(models.Update.status.in_(
                 [models.UpdateStatus.pending, models.UpdateStatus.testing]))
     for update in updates:
+        # We retrieve updates going to testing (status=pending) and updates
+        # (status=testing) going to stable.
+        # If the update is pending, we want to know if it can go to testing
+        decision_context = u'bodhi_update_push_testing'
+        if update.status == models.UpdateStatus.testing:
+            # Update is already in testing, let's ask if it can go to stable
+            decision_context = u'bodhi_update_push_stable'
+
         data = {
             'product_version': update.product_version,
-            'decision_context': u'bodhi_update_push_stable',
+            'decision_context': decision_context,
             'subject': update.greenwave_subject
         }
         api_url = '{}/decision'.format(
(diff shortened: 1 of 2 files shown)


ralphbean:
:+1:

Can you update the greenwave policy in ansible too to add a new decision context?

pypingou:
Sure thing though I was thinking we may want to discuss some more what we want for each context

bodhi/server/scripts/check_policies.py
```diff
@@ -41,9 +41,12 @@ def check():
         .filter(models.Update.status.in_(
                 [models.UpdateStatus.pending, models.UpdateStatus.testing]))
     for update in updates:
+        decision_context = u'bodhi_update_push_testing'
+        if update.status == models.UpdateStatus.testing:
+            decision_context = u'bodhi_update_push_stable'
```
bowlofeggs: I think this code is right, but it is also confusing initially that we'd mix "testing" and "stable" like this. Can we add a comment explaining why we want to check the stable context when it is in testing? Maybe something like this:

```
# We want to know a decision about pending updates going to testing, and about testing updates going to stable.
```

bodhi/tests/server/scripts/test_check_policies.py
```diff
@@ -59,6 +59,35 @@ def test_policies_satisfied(self):
                                                expected_query)
 
     @patch.dict(config, [('greenwave_api_url', 'http://domain.local')])
+    def test_policies_pending_satisfied(self):
+        """Assert correct behavior when the policies enforced by Greenwave are satisfied"""
```
bowlofeggs: This comment doesn't quite capture the purpose of the test as I understand it - can we reword it to state that we are ensuring that pending updates use the ```bodhi_update_push_testing``` decision context? That way it'll be clearer what this test is for later.

bowlofeggs (approved):
Code looks good, I just have a couple of comment changes I'd like to see to clarify things. Merge when ready (no need for re-review)!

pypingou:
Both comment adjusted :)

Will let jenkins run and merge afterward

bowlofeggs (approved):
Thanks!

PR #1775, opened 2017-08-29
PEP-257 bodhi.server.static.

bowlofeggs:


jeremycline (approved):
Tough review

:shipit:

PR #1777, opened 2017-08-29
Move the migrations into the Bodhi package

jeremycline:
Alembic makes it easy to distribute the migrations as part of your
Python package. It also makes configuration easy since it uses
pkg_resources to find the migrations, rather than having the user
configure the path as part of packaging.


jeremycline:
I just moved it to ``bodhi.server.migrations``, but  we could have ``bodhi.server.db`` with the migrations sub-package and the models. I don't have a strong opinion about which is nicer.

.coveragerc
```diff
@@ -3,6 +3,7 @@ source = bodhi
 omit =
     bodhi/scripts/*
```
bowlofeggs, line 4: Not related to this PR, but this path doesn't seem to exist.

alembic.ini
```diff
@@ -2,7 +2,7 @@
 
 [alembic]
 # path to migration scripts
-script_location = alembic
+script_location = bodhi:server/migrations
```
bowlofeggs, line 5: This could be a backwards-incompatible change for the RPM, since alembic.ini is probably marked as a configuration file. This means that users who upgrade will not get this change and the migrations will fail. Two possible solutions I can think of quickly:

1. Don't mark this as a config file. Does it have anything an end user should want to change in it? I'm not sure.
2. Provide a symlink from the old location to the new location so things work smoothly, and drop the symlink in Rawhide. We already do this with another symlink on F26.

I'd probably opt for the second one.

docs/release_notes.rst
```diff
@@ -1,6 +1,16 @@
...
 =============
 
+develop
+-------
+
+Development improvements
+^^^^^^^^^^^^^^^^^^^^^^^^
+
+* The database migrations are now shipped as part of the Python distribution
+  (`#1777 <https://github.com/fedora-infra/bodhi/pull/1777>`_).
```
bowlofeggs, line 18: It would be good to mention the ```alembic.ini``` change here.

bowlofeggs (approved):
Please do mention the ```alembic.ini``` change in the release notes before merging.

PR #1778, opened 2017-08-29
Fix an instance in the tests where os.environ needed to be mocked.

ralphbean:
Worked on this with Eric Kuzmenko and Caleigh Runge-Hottman at Flock!


bowlofeggs (approved):
Thanks!

PR #1780, opened 2017-08-30
Separate primary and secondary arches into two different mash tags

### LINKED ISSUE #1685: Bodhi needs to support alternative arches in mash config files
Currently Bodhi has a mash config file per release per repo (updates, updates-testing). This has worked historically, but starting with Fedora 26 there are alternative arches in Bodhi. Alternative arches are stored in a different filesystem path on ```bodhi-backend01```. That means that having a single [mash config file](https://infrastructure.fedoraproject.org/cgit/ansible.git/tree/roles/bodhi2/backend/files/f26-updates-testing.mash?id=c239ceaa451f47258554554f47550bb878c41b1b) for primary and alternative arches will be a problem, because it defines paths on the filesystem for where mash can find the repos it needs to operate on:

```
delta_dirs = /mnt/koji/compose/branched/latest-Fedora-26/compose/Everything/%(arch)s/os/
parent_repos = https://kojipkgs.fedoraproject.org/compose/branched/latest-Fedora-26/compose/Everything/%(arch)s/os/
```

To address this, we could support separate config files for primary and alternative arches, or we might be able to parameterize the config file so that the paths are substituted at runtime.

jdieter:
This patch uses separate mash files for primary and secondary arches.
This allows us to point to completely different GA paths, which means we
can enable deltarpm generation again for F26.

Secondary arches are required to have the tag <primary name>-secondary,
(i.e. if the primary tag is f26-updates, the secondary tag must be
f26-updates-secondary).

If bodhi's .ini file doesn't specify any primary arches, then all arches
are considered primary and bodhi will not attempt to mash secondary
arches.

This fixes #1685.

### DIFF
--- a/bodhi/server/consumers/masher.py
+++ b/bodhi/server/consumers/masher.py
@@ -68,6 +68,14 @@ def checkpoint(method):
     return wrapper
 
 
+def get_primary_arches(release):
+    release_num = release.id_prefix.lower().replace('-', '_')
+
+    return config.get(
+        '{release}_{version}_primary_arches'.format(
+            release=release_num, version=release.version))
+
+
 class Masher(fedmsg.consumers.FedmsgConsumer):
     """The Bodhi Masher.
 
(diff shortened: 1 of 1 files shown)


bowlofeggs:
Hi @jdieter!

Would you agree that we no longer need this patch since @ausil fixed drpms last week?

jdieter:
Sorry, I was traveling back to Beirut over the last couple of days.  Yes, I think this patch is no longer needed.

PR #1782, opened 2017-08-31
add titles to all updates types on the front page

### LINKED ISSUE #1620: Mouseovers on home page could tell users what the update type counts mean
langdon requested that we add mouseovers on the home page update-type counts to explain what they mean. We could write something like "There are 511 stable security updates in Fedora 25.", for example.

ryanlerch:
previously we had labels for only the middle column of
upadates on the front page. This adds labels for all columns.

Fixes #1620


bowlofeggs (approved):
LGTM!

PR #1784, opened 2017-09-03
PEP-257 bodhi.server.schemas.

bowlofeggs:

### DIFF
--- a/bodhi/server/schemas.py
+++ b/bodhi/server/schemas.py
@@ -1,3 +1,8 @@
+# -*- coding: utf-8 -*-
+# Copyright © 2013-2017 Red Hat, Inc. and others.
+#
+# This file is part of Bodhi.
+#
 # This program is free software; you can redistribute it and/or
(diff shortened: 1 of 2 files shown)


bodhi/server/schemas.py
```diff
@@ -33,6 +38,8 @@
 
 
 class CSRFProtectedSchema(colander.MappingSchema):
+    """Define a mixin class to validate csrf tokens."""
```
jeremycline: The docblock sounds less weird to me if you say "A mixin class..." rather than "Define a mixin class..".

jeremycline (approved):
Looks fine to me. I prefer to drop all the "define a/an" bits, but it's just my opinion. Do what you think is right.

:shipit:

bodhi/server/schemas.py
bowlofeggs: I've removed all the "Define"'s.

PR #1785, opened 2017-09-03
PEP-257 bodhi.server.util.

bowlofeggs:

### DIFF
--- a/bodhi/server/util.py
+++ b/bodhi/server/util.py
@@ -1,3 +1,8 @@
+# -*- coding: utf-8 -*-
+# Copyright © 2007-2017 Red Hat, Inc. and others.
+#
+# This file is part of Bodhi.
+#
 # This program is free software; you can redistribute it and/or
 # modify it under the terms of the GNU General Public License
 # as published by the Free Software Foundation; either version 2
@@ -11,10 +16,7 @@
 # You should have received a copy of the GNU General Public License
 # along with this program; if not, write to the Free Software
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
-
-"""
-Random functions that don't fit elsewhere
-"""
+"""Random functions that don't fit elsewhere."""
 
 from collections import defaultdict
 from contextlib import contextmanager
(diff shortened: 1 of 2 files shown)


bodhi/server/util.py
```diff
@@ -53,8 +55,16 @@ def header(x):
 
 
 def get_rpm_header(nvr, tries=0):
-    """ Get the rpm header for a given build """
+    """
+    Get the rpm header for a given build.
 
+    Args:
+        nvr (basestring): The name-version-release string of the build you want headers for.
```
jeremycline, line 62: Do you really mean basestring? This will get wonky when you want to use Python 3.

I often use ``six.text_type`` and ``six.binary_type`` so it's clear what I mean and also note in the comment whether or not it must be unicode.

bodhi/server/util.py
```diff
@@ -81,53 +91,119 @@ def get_rpm_header(nvr, tries=0):
 
 
 def get_nvr(nvr):
-    """ Return the [ name, version, release ] a given name-ver-rel. """
+    """
+    Return the (name, version, release) from a given name-ver-rel string.
+
+    Args:
+        nvr (basestring): The name-version-release string that you wish to know the split nvr from.
```
jeremycline, line 98: Same comment about the basestring as above. I see it's used throughout, so I'll stop pointing them out.

bodhi/server/util.py
```diff
@@ -81,53 +91,119 @@ def get_rpm_header(nvr, tries=0):
...
+
+    If the function is called later with the same arguments, the cached value is returned (not
     reevaluated).
 
     http://wiki.python.org/moin/PythonDecoratorLibrary#Memoize
-    '''
+
+    Attributes:
+        func (callable): The wrapped function.
+        cache (dict): The cacne, mapping arguments to the cached response.
```
jeremycline: s/cacne/cache

jeremycline (approved):
Oof, that was a lot of functions. Looks good.

My one note is that basestring is not particularly clear. I understand if you don't want to work through and audit each function at this time, though.

:shipit:

bodhi/server/util.py
bowlofeggs: I'm kind of on the fence about ```basestring``` vs. ```six.text_type```. I don't intend to support Python 2 and Python 3 at the same time server-side whenever I do switch to Python 3, so I'll probably just do a search and replace on basestring at that time.

I certainly don't want to audit, because even if it *is* being used with only one of ```str``` or ```unicode```, that doesn't mean it couldn't work with the other. I mostly want to express the "duck type" that the function expects to be able to work with.

So I think I'll leave it as-is for now, since "basestring" is unique enough to be easily sed'able later when we actually start supporting Python 3.

bodhi/server/util.py
bowlofeggs: Fixed.

PR #1797, opened 2017-09-09
Developer docs improvements

alenavolk:
I just finished building a Bodhi development environment on a clean Fedora 26. Here are some tweaks I made to get it up and running. I hope they will save other newcomers some time.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -18,7 +18,6 @@ progressbar
 pyDNS
 pylibravatar
 pyramid
-pyramid_debugtoolbar
 pyramid_fas_openid
 pyramid_mako
 pyramid_tm
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,4 @@
+alembic
 arrow
 bunch
 bleach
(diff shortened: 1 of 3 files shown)


bowlofeggs:
jenkies test

requirements.txt
```diff
@@ -17,6 +18,7 @@ progressbar
 pyDNS
 pylibravatar
 pyramid
+pyramid_debugtoolbar
```
bowlofeggs: This is not a production requirement, and in fact you wouldn't want it present on production. If we add this to this file, then the production package won't install without this dependency so we should drop this line.

bowlofeggs (changes requested):
Thanks for the patches! Just one issue to fix and then we can merge.

requirements.txt
alenavolk: Thanks for the feedback! I moved it to developer docs.

bowlofeggs:
jenkies test

bowlofeggs:
jenkies test

bowlofeggs:
This patch is included in the bodhi-3.1.0-0.1.beta, which is deployed to [staging](https://bodhi.stg.fedoraproject.org/) and is also available via [Copr](https://copr.fedorainfracloud.org/coprs/bowlofeggs/bodhi-pre-release/).

PR #1799, opened 2017-09-11
When updating comps, only update the file we need.

ralphbean:
Running blind make will update everything present, even the files unused by
this run of mash.  Passing the particular file we need should (very) marginally
cut down on mash time.

### DIFF
--- a/bodhi/server/consumers/masher.py
+++ b/bodhi/server/consumers/masher.py
@@ -636,7 +636,8 @@ class MasherThread(threading.Thread):
             util.cmd(['git', 'clone', comps_url, comps_dir], os.path.dirname(comps_dir))
 
         util.cmd(['git', 'pull'], comps_dir)
-        util.cmd(['make'], comps_dir)
+        target = 'comps-%s.xml' % self.release.branch
+        util.cmd(['make', target], comps_dir)
 
     def mash(self):
         if self.path in self.state['completed_repos']:


bodhi/server/consumers/masher.py
```diff
@@ -636,7 +636,8 @@ def update_comps(self):
             util.cmd(['git', 'clone', comps_url, comps_dir], os.path.dirname(comps_dir))
 
         util.cmd(['git', 'pull'], comps_dir)
-        util.cmd(['make'], comps_dir)
+        target = 'target-%s.xml' % self.release.branch
```
bowlofeggs: It seems like some unit tests are setting ```self.release``` to a ```unicode``` object. I presume that in production ```self.release``` is a ```bodhi.server.models.Release``` instance, but I'm not 100% sure of that. In any case, this line seems to cause a handful of unit tests to fail.

bowlofeggs (changes requested):

bodhi/server/consumers/masher.py
ralphbean: Yeah, I goofed it.  Just rebased.

ralphbean:
Super confused by the test failure here.

This is an unimportant change, fwiw.

bowlofeggs:
I think the conflict is probably with https://github.com/fedora-infra/bodhi/pull/1885

bowlofeggs:
I think the tests are just asserting that some particular calls are made, and now some different calls are being made, so we probably just need to update the assertions.

puiterwijk:
I think this PR can be dropped in light of #1900.
We are dropping all comps handling from Bodhi.

bowlofeggs:
Agreed, the word "comps" will barely appear in the repo when we are done with #1900 ☺

PR #1801, opened 2017-09-12
Replace usages of RpmBuild and RpmPackage with superclasses.

### LINKED ISSUE #1793: Several places in the code use RpmBuild and should use Build
When we first converted the models to be multi-type capable, the ```Build.nvr``` field made sense to move into the new ```RpmBuild``` model. As a result, many usages of ```Build``` were replaced with ```RpmBuild``` througout the codebase because many places used the ```nvr``` field.

It was later determined that the field maps to Koji's unique build identifier, and that this identifier is still used for other types (like modules and containers). We decided to move the ```nvr``` field back to the ```Build``` superclass, but we forgot to replace all usages of ```RpmBuild``` in the code with ```Build```.

It is important to fix this, as many queries will not operate on modules since they are querying for ```RpmBuild```s.

bowlofeggs:
fixes #1793

### DIFF
--- a/bodhi/server/consumers/signed.py
+++ b/bodhi/server/consumers/signed.py
@@ -30,7 +30,7 @@ import fedmsg.consumers
 
 from bodhi.server import initialize_db
 from bodhi.server.config import config
-from bodhi.server.models import RpmBuild
+from bodhi.server.models import Build
 from bodhi.server.util import transactional_session_maker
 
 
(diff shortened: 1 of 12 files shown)


ralphbean (approved):
Looks good to me.

bowlofeggs:
This has been cherry-picked to the 2.12 branch as bc5dc5d0b4f09ca3d638d971350d51ac4c736710, and will be included in the upcoming 2.12.0 release.

PR #1805, opened 2017-09-13
CLI prints a hint about how to use koji wait-repo on overrides.

### LINKED ISSUE #1376: The CLI should give the koji wait-repo hint that the web ui gives when creating an override
If a developer uses the web UI to create an override, it will give them a hint about how to use koji wait-repo to know when the override is active in Koji, like this:

```
Use the following to ensure the override is active:

$ koji wait-repo f26-build --build=python-pyramid-1.8.2-2.fc26
```

When I use the CLI to save this override it does not give me such a hint and it would be really nice if it did since I struggle to remember the tag name to wait for. It could probably just use the same text the web UI uses after its current output:

```
$ bodhi overrides save --duration 2 --notes "This is needed to build Bodhi." --user bowlofeggs python-pyramid-1.8.2-2.fc26
<snip>

Use the following to ensure the override is active:

    $ koji wait-repo f26-build --build=python-pyramid-1.8.2-2.fc26
```

bowlofeggs:
The web interface has long printed a message to users about how to
use ``koji wait-repo`` to monitor the progress of created buildroot
overrides, which was a feature gap in the CLI. This commit gives
the CLI the same ability when users create or edit an override, or
when they query for overrides and a single result is matched.

In order to accomplish this, it was necessary to be able to match
the override Build object with a release so the CLI could print the
correct tag to wait for. To accomplish this, Builds now serialize
their release_id field, and releases can now be queried with a new
ids parameter.

fixes #1376

### DIFF
--- a/bodhi/server/models.py
+++ b/bodhi/server/models.py
@@ -791,7 +791,7 @@ class Build(Base):
     """
     __tablename__ = 'builds'
     __exclude_columns__ = ('id', 'package', 'package_id', 'release',
-                           'release_id', 'update_id', 'update', 'override')
+                           'update_id', 'update', 'override')
     __get_by__ = ('nvr',)
 
     nvr = Column(Unicode(100), unique=True, nullable=False)
--- a/bodhi/client/__init__.py
+++ b/bodhi/client/__init__.py
@@ -634,6 +634,28 @@ def edit_buildroot_overrides(user, password, url, staging, **kwargs):
     _save_override(url=url, user=user, password=password, staging=staging, edit=True, **kwargs)
 
 
+def _print_override_koji_hint(override, client):
+    """
+    Print a human readable hint about how to use koji wait-repo to monitor an override, if possible.
+
+    Note: The hint can only be generated if the server provides a 'release_id' on the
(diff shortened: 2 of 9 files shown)


bodhi/server/models.py
```diff
@@ -791,7 +791,7 @@ class Build(Base):
     """
     __tablename__ = 'builds'
     __exclude_columns__ = ('id', 'package', 'package_id', 'release',
-                           'release_id', 'update_id', 'update', 'override')
+                           'update_id', 'update', 'override')
```
bowlofeggs, line 794: I am a bit on the fence about doing this. The downside is that it is exposing what might be considered internal primary keys to API users. The alternative would be to either pass the release along with the ```Build``` object (which would result in a lot more data when users query for Builds) or to add a ```release_name``` property (with the ```release.name```, the client could use the existing ```releases/{{ name }}``` API to get the needed information). Both of these alternatives would cause a join to happen any time Builds are serialized.

In this pull request I opted to give the ```releases/``` API the ability to query by primary key to avoid the join every time Builds are serialized.

jeremycline (approved):
:shipit:

PR #1816, opened 2017-09-19
bodhi.server.util.call_api now uses connection pooling

bowlofeggs:
This PR contains two commits. The first is from @crungehottman, and is approved by me and taken from #1767. The second is from me and removes a few mocks that are no longer needed.

### DIFF
--- a/bodhi/server/util.py
+++ b/bodhi/server/util.py
@@ -48,6 +48,8 @@ from bodhi.server.exceptions import RepodataException
 
 _ = TranslationStringFactory('bodhi')
 
+http_session = requests.Session()
+
 
 def header(x):
     """Display a given message as a heading."""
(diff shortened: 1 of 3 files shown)


bodhi/server/util.py
```diff
@@ -1162,15 +1164,15 @@ def call_api(api_url, service_name, error_key=None, method='GET', data=None):
         base_error_msg = (
             'Bodhi failed to send POST request to {0} at the following URL '
             '"{1}". The status code was "{2}".')
-        rv = requests.post(api_url,
-                           headers={'Content-Type': 'application/json'},
-                           data=json.dumps(data),
-                           timeout=60)
+        rv = http_session.post(api_url,
+                               headers={'Content-Type': 'application/json'},
+                               data=json.dumps(data),
```
jeremycline, line 1169: FYI there's a shortcut for this in requests, the ``json=`` kwarg[0], which will handle serialization, headers, etc. This should be fine as is, just so you know.

[0] http://docs.python-requests.org/en/master/user/quickstart/#more-complicated-post-requests

jeremycline (approved):
:shipit:

bodhi/server/util.py
bowlofeggs: Ah nice! I'm inclined to leave it as-is since this PR is more about using the pooling, but thanks for letting me know! Feel free to send a PR that adjusts it if you like, or I might do it later myself :)

PR #1820, opened 2017-09-20
delete obsolete symlink check

### LINKED ISSUE #1819: symlink check is not valid
https://github.com/fedora-infra/bodhi/blob/33d1e733d88bc891e00bf9f0e09d471360253ab1/bodhi/server/consumers/masher.py#L723-L729

The code is checking `.rpm` files under `mash_path, arches[0]` but you can see from [this example](https://kojipkgs.fedoraproject.org/mash//updates/f26-updates-170804.1348/f26-updates/x86_64/) that no rpm files exist. They are just directories.

dustymabe:
Fixes #1819

### DIFF
--- a/bodhi/server/consumers/masher.py
+++ b/bodhi/server/consumers/masher.py
@@ -704,7 +704,6 @@ class MasherThread(threading.Thread):
     def sanity_check_repo(self):
         """Sanity check our repo.
 
-            - make sure we didn't compose a repo full of symlinks
             - sanity check our repodata
         """
         mash_path = os.path.join(self.path, self.id)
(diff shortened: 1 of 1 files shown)


bowlofeggs:
This code seems to be here for sanity checking purposes. It was added in https://github.com/fedora-infra/bodhi/commit/724cd2df.

On the other hand, I don't recall ever seeing this particular error happen in the ~1 year I've been charged with Bodhi, which makes me think you are probably right that it isn't needed. Additionally, it does seem that mash organizes the repo into folders that collate RPMs by the first letter of their name.

bowlofeggs:
jenkies test

PR #1824, opened 2017-09-22
Add several more tests for the models and raise coverage to 90%.

bowlofeggs:


cverna (approved):
Sweet :) 👍

PR #1830, opened 2017-09-25
Optionally wait for repo signatures before promoting

puiterwijk:

### DIFF
--- a/bodhi/server/consumers/masher.py
+++ b/bodhi/server/consumers/masher.py
@@ -972,6 +972,7 @@ class PungiComposerThread(ComposerThread):
             uinfo.insert_updateinfo(self.path)
 
             self._sanity_check_repo()
+            self._wait_for_repo_signature()
             self._stage_repo()
 
             # Wait for the repo to hit the master mirror
@@ -1239,6 +1240,42 @@ class PungiComposerThread(ComposerThread):
 
         self.log.debug('Path: %s', self.path)
         self._checkpoints['completed_repo'] = self.path
+
+    def _wait_for_repo_signature(self):
+        """Wait for a repo signature to appear."""
+        # This message indicates to consumers that the repos are fully created and ready to be
+        # signed or otherwise processed.
+        notifications.publish(
+            topic="repo.done",
+            msg=dict(repo=self.id, agent=self.agent, path=self.path),
+            force=True,
+        )
+        if config.get('wait_for_repo_sig'):
+            self.save_state(ComposeState.signing_repo)
+            sigpaths = []
+            repopath = os.path.join(self.path, 'compose', 'Everything')
+            for arch in os.listdir(repopath):
+                if arch == 'source':
+                    sigpaths.append(os.path.join(repopath, arch, 'tree', 'repodata',
+                                                 'repomd.xml.asc'))
+                else:
+                    sigpaths.append(os.path.join(repopath, arch, 'os', 'repodata',
+                                                 'repomd.xml.asc'))
+
+            self.log.info('Waiting for signatures in %s', ', '.join(sigpaths))
+            while True:
+                missing = []
+                for path in sigpaths:
+                    if not os.path.exists(path):
+                        missing.append(path)
+                if len(missing) == 0:
+                    self.log.info('All signatures were created')
+                    break
+                else:
+                    self.log.info('Waiting on %s', ', '.join(missing))
+                    time.sleep(300)
+        else:
+            self.log.info('Not waiting for a repo signature')
 
     def _wait_for_sync(self):
         """
--- a/bodhi/server/models.py
+++ b/bodhi/server/models.py
@@ -740,6 +740,7 @@ class ComposeState(DeclEnum):
             forms of notifications, such as e-mail, fedmsgs, and bugzilla.
         success (EnumSymbol): The Compose has completed successfully.
         failed (EnumSymbol): The compose has failed, abandon hope.
+        signing_repo (EnumSymbol): Waiting for the repo to be signed.
     """
 
     requested = 'requested', 'Requested'
@@ -750,6 +751,7 @@ class ComposeState(DeclEnum):
     notifying = 'notifying', 'Sending notifications'
     success = 'success', 'Success'
     failed = 'failed', 'Failed'
+    signing_repo = 'signing_repo', 'Signing repo'
 
 
 ##
--- a/bodhi/tests/server/consumers/test_masher.py
+++ b/bodhi/tests/server/consumers/test_masher.py
@@ -372,6 +372,7 @@ That was the actual one''' % mash_dir
     @mock.patch('bodhi.server.consumers.masher.PungiComposerThread._wait_for_pungi')
     @mock.patch('bodhi.server.consumers.masher.PungiComposerThread._sanity_check_repo')
     @mock.patch('bodhi.server.consumers.masher.PungiComposerThread._stage_repo')
+    @mock.patch('bodhi.server.consumers.masher.PungiComposerThread._wait_for_repo_signature')
     @mock.patch('bodhi.server.consumers.masher.PungiComposerThread._generate_updateinfo')
     @mock.patch('bodhi.server.consumers.masher.PungiComposerThread._wait_for_sync')
     @mock.patch.object(ComposerThread, 'determine_and_perform_tag_actions', mock_exc)
@@ -413,6 +414,7 @@ That was the actual one''' % mash_dir
     @mock.patch('bodhi.server.consumers.masher.PungiComposerThread._wait_for_pungi')
     @mock.patch('bodhi.server.consumers.masher.PungiComposerThread._sanity_check_repo')
     @mock.patch('bodhi.server.consumers.masher.PungiComposerThread._stage_repo')
+    @mock.patch('bodhi.server.consumers.masher.PungiComposerThread._wait_for_repo_signature')
     @mock.patch('bodhi.server.consumers.masher.PungiComposerThread._generate_updateinfo')
     @mock.patch('bodhi.server.consumers.masher.PungiComposerThread._wait_for_sync')
     @mock.patch('bodhi.server.notifications.publish')
@@ -483,6 +485,7 @@ That was the actual one''' % mash_dir
     @mock.patch('bodhi.server.consumers.masher.PungiComposerThread._wait_for_pungi')
     @mock.patch('bodhi.server.consumers.masher.PungiComposerThread._sanity_check_repo')
     @mock.patch('bodhi.server.consumers.masher.PungiComposerThread._stage_repo')
+    @mock.patch('bodhi.server.consumers.masher.PungiComposerThread._wait_for_repo_signature')
     @mock.patch('bodhi.server.consumers.masher.PungiComposerThread._generate_updateinfo')
     @mock.patch('bodhi.server.consumers.masher.PungiComposerThread._wait_for_sync')
     @mock.patch('bodhi.server.notifications.publish')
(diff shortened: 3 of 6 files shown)


bodhi/server/consumers/masher.py
```diff
@@ -658,9 +659,22 @@ def wait_for_mash(self, mash_thread):
...
+                msg=dict(path=self.path,
+                         id=self.id,
+                         agent=self.agent),
+                force=True)
+
             self.state['completed_repos'].append(self.path)
             self.save_state()
         else:
+            notifications.publish(
+                topic='repo.mash.failed',
```
bowlofeggs: Are these ```repo.mash.{complete,failed}``` topics new? I don't see them when git grepping, so I presume they are. It would be good if we add them to these docs as well:

https://fedora-fedmsg.readthedocs.io/en/latest/topics.html#bodhi

bodhi/server/consumers/masher.py
```diff
@@ -721,6 +735,37 @@ def sanity_check_repo(self):
 
         return True
 
+    def wait_for_repo_signature(self):
+        """Wait for a repo signature to appear if configured as such."""
+        if not config.get('wait_for_repo_sig'):
+            self.log.info('Not waiting for a repo signature')
+            return
```
bowlofeggs: Optional suggestion: Why not move this if statement to the place where ```wait_for_repo_signature()``` is called? That way, this method always waits if called, which is more like what its name sounds like.

If you like this better I'm not opposed, just a suggestion.

bodhi/server/consumers/masher.py
```diff
@@ -721,6 +735,37 @@ def sanity_check_repo(self):
...
+        sigpaths = {}
+        for arch in os.listdir(os.path.join(self.path, self.id)):
+            sigpaths.append(os.path.join(self.path, self.id, arch, 'repodata',
+                                         'repomd.xml.asc'))
+        self.log.info('Waiting for signatures in %s' % sigpaths)
+        notifications.publish(
+            topic="reposig.wait",
+            msg=dict(repo=self.id, agent=self.agent),
+            force=True,
+        )
```
bowlofeggs: This one would be good to add to the fedmsg docs too.

bodhi/server/consumers/masher.py
```diff
@@ -721,6 +735,37 @@ def sanity_check_repo(self):
...
+        while True:
+            got_all = True
+            for path in sigpaths:
+                if not os.path.exists(path):
+                    got_all = False
+            if got_all:
+                self.log.info('All signatures were created')
+            else:
+                self.log.info('Waiting')
+                time.sleep(5)
```
bowlofeggs: Should we have some huge amount of time that we give up and error out after? Like, if it hasn't happened in 10 hours or something arbitrary?

bodhi/server/consumers/masher.py
```diff
@@ -721,6 +735,37 @@ def sanity_check_repo(self):
...
+        notifications.publish(
+            topic="reposig.done",
+            msg=dict(repo=self.id, agent=self.agent),
+            force=True,
+        )
```
bowlofeggs: This one would be good to document too.

bowlofeggs (changes requested):
Looks pretty good to me, I just have a few suggestions here and there.

The main thing we need is tests. I can help write them sometime soon, but of course feel free to write them too :)

bodhi/server/consumers/masher.py
puiterwijk: They are new, yes. And after you agree with the message structure, I would be adding them to fedmsg-meta yes.

bodhi/server/consumers/masher.py
puiterwijk: I primarily went with it here because the place where it gets called is really a single series of statements, with just the "if mashing;" conditionals. I'll move it over.

bodhi/server/consumers/masher.py
puiterwijk: Well, if we at some point enable this on clients, and we send any repo without signature, that repo will be useless.
So when we turn this on, we should make it required to proceed.

puiterwijk: Also, during the initial time while it's not yet enabled for clients, if for whatever reason the signature doesn't get generated, we can just create the files by hand.

puiterwijk:
Given the massive changes to masher.py in #1900, I'm going to wait for that to get merged, after that I'll write the tests for this.

bowlofeggs:
There are again massive changes in #2052, so you might want to either try to beat it to merge or wait until it merges to pick this back up.

bowlofeggs:
I am going to work on rebasing this. Might not finish it today though.

bowlofeggs:
Alright, I've gotten it rebased, and I also added a new ComposeState for this so we can see if it gets stuck here. I've not yet tested the migration so we might need some adjustments there, and we still need some tests.

I'll be gone for a while, but I hope to return to this when I get back.

bodhi/server/consumers/masher.py
```diff
@@ -358,12 +358,23 @@ def run(self):
...
                 self.db = session
                 self.compose = Compose.from_dict(session, self._compose)
                 self.compose.error_message = unicode(e)
                 self.save_state(ComposeState.failed)
+                notifications.publish(
+                    topic='repo.mash.failed',
+                    msg=dict(path=self.path,
+                             id=self.id,
+                             agent=self.agent),
+                    force=True)
```
bowlofeggs: Are these two messages related to waiting on repo signatures? If not, I'd prefer not to include this change in the patch to keep the patch more targeted. Also, there already is a message sent upon completion and it indicates success/failure, so is this redundant? Lastly, if we do keep this, should we keep using the word "mash" since it's composes now? We can't rename the existing messages without bumping Bodhi's X version, but I've files a [ticket](https://github.com/fedora-infra/bodhi/issues/2151) to do that later.

puiterwijk: So, those two messages can be dropped, but I'll need `self.path` as argument to the `repo.signature.wait` message.

bowlofeggs:
Here's a patch to fix a small issue with the migration and to add the new setting to ```production.ini```:

```
diff --git a/bodhi/server/migrations/versions/c21dd18b161a_add_a_new_signing_repo_composestate_enum.py b/bodhi/server/migrations/versions/c21dd18b161a_add_a_new_signing_repo_composestate_enum.py
index a4901546..32d3eddc 100644                            
--- a/bodhi/server/migrations/versions/c21dd18b161a_add_a_new_signing_repo_composestate_enum.py                       
+++ b/bodhi/server/migrations/versions/c21dd18b161a_add_a_new_signing_repo_composestate_enum.py                       
@@ -23,8 +23,7 @@ Revises: 2616c86d8ac6                    
 Create Date: 2018-01-29 16:18:00.441595                   
 """                         
 from alembic import op      
-import sqlalchemy as sa     
-                            
+from sqlalchemy import exc  
                             
                             
diff --git a/bodhi/tests/server/consumers/test_masher.py b/bodhi/tests/server/consumers/test_masher.py                
index 80467094..d81ce594 100644                            
--- a/bodhi/tests/server/consumers/test_masher.py          
+++ b/bodhi/tests/server/consumers/test_masher.py          
@@ -1,5 +1,5 @@              
 # -*- coding: utf-8 -*-     
-# Copyright © 2007-2017 Red Hat, Inc.                     
+# Copyright © 2007-2018 Red Hat, Inc.                     
 #                           
 # This file is part of Bodhi.                             
 #                           
index 81dce3c9..cbe60deb 100644                            
--- a/production.ini         
+++ b/production.ini         
@@ -116,6 +116,9 @@ use = egg:bodhi-server                 
 # location of this file.    
 # mash_stage_dir =          
                             
+# Whether to wait for .asc signature files in the repo when composing updates or not.                                
+# wait_for_repo_sig = False 
+                            
 # The following jinja2 template variables are available for use to customize the Pungi configs and                   
 # variants files to the Release and Updates:              
 #
```

bodhi/server/models.py
```diff
@@ -710,6 +710,7 @@ class ComposeState(DeclEnum):
             forms of notifications, such as e-mail, fedmsgs, and bugzilla.
         success (EnumSymbol): The Compose has completed successfully.
         failed (EnumSymbol): The compose has failed, abandon hope.
+        signing_repo (EnumSymbol): Waiting for the repo to be siged.
```
mscherer: Shouldn't it be "signed" rather than "siged" ?

bodhi/server/consumers/masher.py
```diff
@@ -963,6 +964,40 @@ def sanity_check_repo(self):
...
+            for arch in os.listdir(repopath):
+                if arch == 'source':
+                    sigpaths.append(os.path.join(repopath, arch, 'tree', 'repodata',
+                                                 'repomd.xml.asc'))
+                else:
+                    sigpaths.append(os.path.join(repopath, arch, 'os', 'repodata',
+                                                 'repomd.xml.asc'))
+
+            self.log.info('Waiting for signatures in %s' % sigpaths)
+            while True:
```
mscherer: Is it safe to wait basically "forever" and never fail ? 
I could imagine a few race conditions where this would slowly exhaust ressources, so maybe there should be some limit somewhere, unless that's already handled externally.

bowlofeggs: @mscherer Yeah I had a similar thought - we probably want it to fail if it's not signed after some timeout, just so it doesn't sit around forever.

puiterwijk: I would say that if there's something wrong to the point where we can't get a repo signed, releng should just abort the push by hand and disable the option.
The problem is that it might be that there's a backlog for signing up to multiple hours, and then there's some time needed for releng to debug it.
If we at some point decide to turn this on and we start shipping configurations with repo_gpgcheck=1, we just cannot ship unsigned repos anymore.

So, I personally think that we should not have an automated timeout, and any timing out should be done by releng.

bowlofeggs:
On 02/28/2018 08:18 PM, Patrick Uiterwijk wrote:
> If we at some point decide to turn this on and we start shipping
> configurations with repo_gpgcheck=1, we just cannot ship unsigned repos
> anymore.

I think the suggestion is that the compose would fail after the timeout
(i.e., we would not ship unsigned repos to users, we would just mark the
compose as failed automatically.)

puiterwijk:
Right, and I tried to say that even that is not something we'd probably want (or need) I think.
There are not many resources accumulated during the wait, but not having an automated timeout means that we don't inadvertently abort the push just because robosignatory was busy processing a mass rebuild.
If we want the push to fail because of signing issues, releng can do that.

puiterwijk:
Tests have been added.

bodhi/server/consumers/masher.py
```diff
@@ -431,7 +431,6 @@ def work(self):
             self.remove_pending_tags()
 
             self._compose_updates()
-
```
bowlofeggs: This change is unrelated to the patch at hand, and I also would prefer to keep this line here.

bodhi/server/consumers/masher.py
```diff
@@ -1240,6 +1240,40 @@ def _wait_for_pungi(self, pungi_process):
         self.log.debug('Path: %s', self.path)
         self._checkpoints['completed_repo'] = self.path
 
+    def _wait_for_repo_signature(self):
+        """Wait for a repo signature to appear."""
+        notifications.publish(
+            topic="repo.signature.wait",
+            msg=dict(repo=self.id, agent=self.agent, path=self.path),
+            force=True,
+        )
```
bowlofeggs: Shouldn't we only send this message if we are going to wait (i.e., do this inside the if block below)?

bodhi/tests/server/consumers/test_masher.py
```diff
@@ -2756,6 +2780,86 @@ def test_stage_dir_dne(self):
...
+        t.log.info = mock.MagicMock()
+
+        with mock.patch.dict(config, {'wait_for_repo_sig': False}):
+            t._wait_for_repo_signature()
+
+        publish.assert_called_once_with(topic='repo.signature.wait',
+                                        force=True,
+                                        msg={'repo': t.id,
+                                             'path': t.path,
+                                             'agent': 'ralph'})
```
bowlofeggs: I suggest moving that message into the if block and changing this assertion to:

```
self.assertEqual(publish.call_count, 0)
```

bowlofeggs (approved):
Only a few small changes - feel free to merge without re-review, and thanks!

PR #1833, opened 2017-09-25
Turn everything into py3 syntax

puiterwijk:


puiterwijk:
The changes from `datetime(1984, 11, 02)` to `datetime(1984, 11, 2)` are needed because in py2, "02" was interpreted as the octal number 2 (lucky you it coincides with decimal 2!), but py3 disallows this because of this exact confusion :).  (for octal numbers, you'd need 0o2 now).

ryanlerch:
It appears that one of the lines you touched in bodhi/server/models.py (line 2977) did not have test coverage before your patch, so for some reason (i think) the CI thinks you added a line that does not have testing, and that is why the CI is failing... :(

puiterwijk:
Yeah, it's just that not all lines I've changed have test coverage, and the CI is configured so that it'll fail if a changed line doesn't have test coverage.
Which normally is a good check, but in this case... meh.

ryanlerch (approved):

PR #1834, opened 2017-09-27
100% test coverage for client/__init__.py

ryanlerch:

### DIFF
--- a/bodhi/client/__init__.py
+++ b/bodhi/client/__init__.py
@@ -153,7 +153,7 @@ def cli():
     """
     Create the main CLI group
     """
-    pass
+    pass  # pragma: no cover
 
 
 @cli.group()
@@ -167,7 +167,7 @@ def updates():
     """
     Create the updates group.
     """
-    pass
+    pass  # pragma: no cover
 
 
 @updates.command()
@@ -517,7 +517,7 @@ def overrides():
     """
     Create the overrides CLI group.
     """
-    pass
+    pass  # pragma: no cover
 
 
 @overrides.command('query')
@@ -685,13 +685,8 @@ def print_resp(resp, client):
     elif 'comment' in resp:
         click.echo('The following comment was added to %s' % resp.comment['update'].title)
         click.echo(resp.comment.text)
-    elif 'errors' in resp:
-        for error in resp['errors']:
-            click.echo("ERROR: %s" % error['description'])
-        sys.exit(1)
     else:
         click.echo(resp)
-
     if resp.get('caveats', None):
         click.echo('Caveats:')
         for caveat in resp.caveats:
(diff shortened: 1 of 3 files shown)


bodhi/client/__init__.py
```diff
@@ -661,13 +661,8 @@ def print_resp(resp, client):
     elif 'comment' in resp:
         click.echo('The following comment was added to %s' % resp.comment['update'].title)
         click.echo(resp.comment.text)
-    elif 'errors' in resp:
-        for error in resp['errors']:
-            click.echo("ERROR: %s" % error['description'])
-        sys.exit(1)
```
bowlofeggs, line 691: At first this made me nervous, but I realized that we already have code that handles this in the bindings so this would never happen:

https://github.com/fedora-infra/bodhi/blob/2.11.0/bodhi/client/bindings.py#L92-L101

bodhi/tests/client/test___init__.py
```diff
@@ -134,7 +273,65 @@ class TestQuery(unittest.TestCase):
...
+        Assert we display correctly when the query returns a single update.
+        """
+        runner = testing.CliRunner()
+
+        result = runner.invoke(
+            client.query,
+            ['--builds', 'nodejs-grunt-wrap-0.3.0-2.fc25', '--url', 'http://localhost:6543'])
+
+        self.assertEqual(result.exit_code, 0)
+        self.maxDiff = 2000
```
bowlofeggs: This is probably a debug artifact - can you remove it before merging?

bodhi/tests/client/test___init__.py
```diff
@@ -134,7 +273,65 @@ class TestQuery(unittest.TestCase):
...
+        Assert we display correctly when the query returns a single update.
+        """
+        runner = testing.CliRunner()
+
+        result = runner.invoke(
+            client.query,
+            ['--builds', 'nodejs-grunt-wrap-0.3.0-2.fc25'])
+
+        self.assertEqual(result.exit_code, 0)
+        self.maxDiff = 2000
```
bowlofeggs: Same here.

bowlofeggs (approved):
Looks good! Can you drop the two maxDiff's before merging? Thanks for the test coverage, it's fantastic!

bodhi/client/__init__.py
ryanlerch: Sorry! i meant to mention this in the PR description. :)

I was a little hard to write a test for something that we would never hit :D

ryanlerch:
this is now failing CI after rebasing due to failing tests in develop. we can merge this after #1837

bowlofeggs:
jenkies test

bowlofeggs (approved):
LGTM!

PR #1835, opened 2017-09-27
100% test coverage for client/bindings.py

ryanlerch:
This also fixes the following issue i found while adding the new tests:

Add updates/ to url of updates w/out alias in CLI. Previously, if an update did not have an alias,
the bodhi CLI rendered the url in the updates output as ```bodhi.fedoraproject.org/bodhi-2.2.4-1.el7```
which would result in a 404 if viewed. This update changes the URL so it renders as: ```bodhi.fedoraproject.org/updates/bodhi-2.2.4-1.el7```

### DIFF
--- a/bodhi/client/bindings.py
+++ b/bodhi/client/bindings.py
@@ -592,7 +592,7 @@ class BodhiClient(OpenIdBaseClient):
             val += "\n  %s\n" % ('%supdates/%s' % (self.base_url,
                                                    update['alias']))
         else:
-            val += "\n  %s\n" % ('%s%s' % (self.base_url, update['title']))
+            val += "\n  %s\n" % ('%supdates/%s' % (self.base_url, update['title']))
         return val
 
     @errorhandled
--- a/bodhi/tests/client/test_bindings.py
+++ b/bodhi/tests/client/test_bindings.py
@@ -768,6 +768,60 @@ class TestBodhiClient_save_override(unittest.TestCase):
         expected_expiration = now + timedelta(days=2)
         self.assertTrue((actual_expiration - expected_expiration) < timedelta(minutes=5))
 
+    def test_save_override_edit(self):
+        """
+        Test the save_override() method with the edit argument.
+        """
+        client = bindings.BodhiClient()
+        client.send_request = mock.MagicMock(return_value='return_value')
+        client.csrf_token = 'a token'
+        now = datetime.utcnow()
+        response = client.save_override(nvr='python-pyramid-1.5.6-3.el7',
+                                        duration=2,
+                                        notes='This is needed to build bodhi-2.4.0.',
+                                        edit=True)
+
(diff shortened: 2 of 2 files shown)


bodhi/client/bindings.py
```diff
@@ -592,7 +592,7 @@ def update_str(self, update, minimal=False):
             val += "\n  %s\n" % ('%supdates/%s' % (self.base_url,
                                                    update['alias']))
         else:
-            val += "\n  %s\n" % ('%s%s' % (self.base_url, update['title']))
+            val += "\n  %s\n" % ('%supdates/%s' % (self.base_url, update['title']))
```
bowlofeggs, line 595: Oh good catch. I bet this ```else``` branch never happens in practice, but this was def wrong.

bodhi/tests/client/test_bindings.py
```diff
@@ -919,6 +987,48 @@ def test_with_autokarma_unset(self):
...
+            "Comments: bodhi ",
+            "Comments: bodhi (unauthenticated) "))
+
+    @mock.patch.dict(
+        client_test_data.EXAMPLE_UPDATE_MUNCH, {u'alias': None})
+    def test_update_with_no_alias(self):
+        """Ensure we render an update with no alias set properly"""
+        client = bindings.BodhiClient()
+        client.base_url = 'http://example.com/tests/'
+        self.maxDiff = 2000
```
bowlofeggs: We can drop this line I think.

bodhi/tests/client/test_bindings.py
```diff
@@ -1316,3 +1426,44 @@ def test_testable_no_dnf(self):
...
+    @mock.patch('__builtin__.open', create=True)
+    @mock.patch('bodhi.client.bindings.ConfigParser.readfp')
+    @mock.patch("os.path.exists")
+    @mock.patch("os.path.expanduser", return_value="/home/dudemcpants/")
+    @mock.patch('bodhi.client.bindings.BodhiClient._load_cookies')
+    @mock.patch('bodhi.client.bindings.koji.ClientSession')
+    @mock.patch('bodhi.client.bindings.ConfigParser.get')
+    def test_koji_conf_not_in_home_directory(self, get, koji, cookies,
+                                             expanduser, exists, readfp, mock_open):
+        """Test that if ~/.koji/config exists, we read the config from there first"""
```
bowlofeggs: Maybe update this docblock to say "Test that if ~/.koji.config doesn't exist, we read config from /etc/koji.conf".

bowlofeggs (approved):
Just a couple small things, feel free to merge once they are fixed. Thanks for the tests, it's so helpful!

ryanlerch:
this is now failing CI after rebasing due to failing tests in develop. we can merge this after #1837

bowlofeggs:
jenkies test

bowlofeggs (approved):

PR #1836, opened 2017-09-27
Expand the groups when querying pagure

pypingou:
With this commit pagure is asked to provide membership information
in addition to its usual data.
This feature was added in: https://pagure.io/pagure/pull-request/2627

The members of the group with commit or admin access are then simply
added to the list of committers.

### DIFF
--- a/bodhi/server/models.py
+++ b/bodhi/server/models.py
@@ -644,7 +644,7 @@ class Package(Base):
             namespace = self.type.name
         else:
             namespace = self.type.name + 's'
-        package_pagure_url = '{0}/api/0/{1}/{2}'.format(
+        package_pagure_url = '{0}/api/0/{1}/{2}?expand_group=1'.format(
             pagure_url.rstrip('/'), namespace, self.name)
         package_json = pagure_api_get(package_pagure_url)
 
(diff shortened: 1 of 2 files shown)


jeremycline (commented):
It'd be good to have a test or two that demonstrated how this alters/fixes the API.

pypingou:
Well, since this is something changing on the pagure response, I don't really know how you would want to test that. Any pointers?

jeremycline:
Maybe it's time to set up vcrpy/betamax for bodhi. Then your test can make an actual request and make sure it processes the json response correctly. Alternately you can deserialized the JSON response from a request and include it in the test.

pypingou:
I will not be able to dedicate the time to add this, so if someone wants to work on the tests, please go ahead.

bowlofeggs:
@puiterwijk suggested that we can use these two URLs as nice example data for the tests:

https://src.fedoraproject.org/api/0/rpms/dnf?expand_group=0 and https://src.fedoraproject.org/api/0/rpms/dnf?expand_group=1

I like the suggestion of using vcrpy, but we can also just copy those JSON blobs right into the test file for now if you want. Either way is fine by me.

bowlofeggs:
I'm going to write tests for this.

bowlofeggs:
@pypingou I rebased this PR and added a commit that brings some tests in to ensure the code can handle the data structure that Pagure provides.

I approve your commit, but now I need someone to review my commit and then we can merge this and make a release.

pypingou:
Your commit looks good to me, thanks for tackling it!

bowlofeggs:
@pypingou's commit has been cherry-picked to the 2.12 branch as 68dfedceda0587df8bcd7eaa457f2bc9c3bdd74e, and will be included in the upcoming 2.12.0 release.

The tests were not simple to backport, so they will remain only on the develop branch for now.

PR #1829, opened 2017-09-25
Added Link to IRC Freenode 

amanharitsh123:

PR #1842, opened 2017-10-01
Add IRC freenode link

amanharitsh123:
Signed-off-by: amanharitsh123<amanharitsh123@gmail.com>

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -30,7 +30,7 @@ Links
 ^^^^^
 
 * `Online documentation <https://bodhi.fedoraproject.org/docs/>`_
-*  IRC: #bodhi on Freenode
+* `IRC: #bodhi on Freenode <https://webchat.freenode.net/>`_
 * `Mailing list <https://lists.fedoraproject.org/archives/list/bodhi@lists.fedorahosted.org/>`_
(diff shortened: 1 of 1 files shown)


amanharitsh123:
I have attached the DCO..Why is it showing DCO missing

amanharitsh123:
Can anyone please review this PR?

amanharitsh123:
@bowlofeggs Please review

bowlofeggs:
Please don't open new pull requests to make changes, because that causes us to lose the context of the conversation around the code. You can use ```git commit --amend``` and ```git push -f``` to edit your existing branch and pull request, without needing to open new ones.

I had a requested change in the original PR to make only the word Freenode a link. I'll copy that here too.

For reference, this code was originally proposed in https://github.com/fedora-infra/bodhi/pull/1829.

bowlofeggs (changes requested):
Since the link doesn't actually take the person to the #bodhi channel but just to Freenode, can we adjust it so the link is only on the word "Freenode"? Like this:

```
* IRC: #bodhi on `Freenode <https://webchat.freenode.net/>`_
```

bowlofeggs:
jenkies test

bowlofeggs:
This was done as part of https://github.com/fedora-infra/bodhi/pull/1883. Thanks for the patch!

PR #1843, opened 2017-10-01
Run CI on Fedora containers as well.

bowlofeggs:

### DIFF
--- a/devel/ci/Dockerfile-footer
+++ b/devel/ci/Dockerfile-footer
@@ -1,0 +1,7 @@
+
+VOLUME ["/results"]
+WORKDIR /bodhi
+ENTRYPOINT ["/bodhi/devel/ci/run_tests_fedora.sh"]
+CMD [""]
+
+COPY . /bodhi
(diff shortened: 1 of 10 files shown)


bowlofeggs:
jenkies test

puiterwijk:
I would strongly suggest to combine the different dockerfile's into a collective part and separate parts where needed, to make sure the list of packages stays in sync.

bowlofeggs:
@puiterwijk Yeah I considered doing that, but there are a few issues - the Dockerfiles are different in subtle ways, and I don't know a great clean way to reuse code in Dockerfiles (afaik, you can't import other Dockerfiles in them) other than maybe using templating. For example, the F25 Dockerfile needs to install a bunch of fedmsg packages that don't exist anymore in F26+, and the Rawhide one uses all python2- names. Since they are almost entirely just a list of packages and since that list of packages is actually pretty specific to each Fedora release, I didn't feel too bad about the repetition.

Given that, do you think I should still try to consolidate the bits that are in common? Do you know a good technique for doing that with a Dockerfile, or is templating the way to go?

bowlofeggs:
jenkies test

bowlofeggs:
jenkies test

bowlofeggs:
I have reworked the PR to share Dockerfile code where possible, and the tests pass now. This PR is ready for review!

jeremycline (approved):
Very nice.

:shipit:

puiterwijk (approved):
:golf:

PR #1844, opened 2017-10-02
Adjust the timeout for CI.

bowlofeggs:
This is needed for
https://github.com/fedora-infra/bodhi/pull/1843, as it causes tests
to run for a longer period.


jeremycline (approved):
:shipit:

Issue #752, opened 2015-12-29
Don't show EOL releases in user's view


kalev:
https://bodhi.fedoraproject.org/updates/?user=kalev&status=testing still shows a F21 gom-0.3.1-1.fc21 update, even though F21 is long EOL. I think it would be better to hide all EOL updates as there's nothing the user can do with these any more.

ralphbean:
Hm.  It's a bit tricky, because that same endpoint is used to query for updates using *any* criteria (both UI and API).. and there's already a mechanism to control which releases get returned.  Take [this query](https://bodhi.fedoraproject.org/updates/?user=kalev&status=testing&release=f23) for example, which shows only your F23 testing updates.

One option would be to default the ``release`` argument to the list of active releases.  But then, how would you query for all updates, including old EOL updates (if you wanted them for some kind of report-generating script)?  Would you have to explicitly list all releases in the query?

ralphbean:
Another idea:  I'm assuming you got to that page by clicking the ``Testing >`` link from [your profile page](https://bodhi.fedoraproject.org/users/kalev).  We could make *that link* include an explicit list of only the active branches... that way users could look at the URL and see explicitly what they're requesting.. and then modify it if they wanted to base a script off of the page.

kalev:
Yup, that's exactly where I got the link. Sounds like a good solution to me, thanks.

Fale:
Hi,
I'm experiencing the same problem.
See:
- https://bodhi.fedoraproject.org/updates/?user=fale&status=testing
- https://bodhi.fedoraproject.org/updates/FEDORA-2016-42c5c95430
- https://fedorahosted.org/rel-eng/ticket/6474

deekej:
Same issue here:
https://bodhi.fedoraproject.org/updates/FEDORA-2016-6ecaf251f6

Would it be possible to actually delete the updates which are stuck in *testing* for EOL releases manually? I don't expect those updates to ever make it further, and if people need some data gathered they can do it before manually deleting it.

I'm not sure if this is not against some Bodhi policy at the moment, but we have the ability to *unpush* an update for current releases. IMHO we should have some similar (*delete*) option for an update in the EOL releases... :)

mcepl:
me too … https://bodhi.fedoraproject.org/updates/FEDORA-2016-eb0392d623

ryanlerch:
[...]

PR #1845, opened 2017-10-02
Only list the user's updates that are still actionable

pypingou:
Today when listing the user's updates we also include updates that
are part of releases that have been archived or disabled.
With this commit only the updates for releases that the user can still
influence will be shown.

### DIFF
--- a/bodhi/server/schemas.py
+++ b/bodhi/server/schemas.py
@@ -522,6 +522,12 @@ class ListUpdateSchema(PaginatedSchema, SearchableSchema, Cosmetics):
         missing=None,
     )
 
+    active_releases = Releases(
+        colander.Boolean(true_choices=('true', '1')),
+        location="querystring",
+        missing=False,
+    )
+
     packages = Packages(
         colander.Sequence(accept_scalar=True),
         location="querystring",
--- a/bodhi/server/services/updates.py
+++ b/bodhi/server/services/updates.py
@@ -34,6 +34,7 @@ from bodhi.server.models import (
     ContentType,
     CVE,
     UpdateRequest,
+    Release,
     ReleaseState,
     Build,
     Package,
@@ -320,8 +321,15 @@ def query_updates(request):
         query = query.filter(Update.date_pushed < pushed_before)
 
     releases = data.get('releases')
+    active_releases = data.get('active_releases')
     if releases is not None:
         query = query.filter(or_(*[Update.release == r for r in releases]))
+    elif active_releases:
+        query = query.filter(
+            Update.release_id == Release.id
+        ).filter(
+            Release.state.in_([ReleaseState.current, ReleaseState.pending])
+        )
 
     # This singular version of the plural "releases" is purely for bodhi1
     # backwards compat (mostly for RSS feeds) - threebean
@@ -397,6 +405,7 @@ def query_updates(request):
         display_user=data.get('display_user', False),
         display_request=data.get('display_request', True),
         package=package,
+        active_releases=active_releases,
     )
 
 
--- a/bodhi/server/templates/updates.html
+++ b/bodhi/server/templates/updates.html
@@ -27,6 +27,23 @@ def inherit(context):
       ${tables.updates(updates, display_user, display_request)}
   % if chrome:
       ${self.pager.render(page, pages)}
+  % if active_releases:
+    <p>
+      <a href="${request.route_url('updates') + '?' \
+        + request.query_string.split('active_releases', 1)[0] \
+        + (request.query_string.split('active_releases', 1)[1].partition('&')[2] or '')}"
+         class="btn btn-sm btn-secondary">
+        Include old releases
+      </a>
+    </p>
+  % else:
+    <p>
+      <a href="${request.route_url('updates') + '?active_releases=true&' + request.query_string}"
+         class="btn btn-sm btn-secondary">
+        Exclude old releases
+      </a>
+    </p>
+  % endif
     </div>
   </div>
 </div>
--- a/bodhi/server/templates/user.html
+++ b/bodhi/server/templates/user.html
@@ -91,7 +91,7 @@
         <a href="${urls['recent_updates_rss']}">
           RSS <span class="fa fa-rss"></span>
         </a>
-        <a href="${urls['recent_updates']}&status=testing">
+        <a href="${urls['recent_updates']}&status=testing&active_releases=true">
           Testing<span class="glyphicon glyphicon-chevron-right"></span>
         </a>
         <a href="${urls['recent_updates']}&status=stable">
--- a/bodhi/tests/server/services/test_updates.py
+++ b/bodhi/tests/server/services/test_updates.py
@@ -2079,6 +2079,41 @@ class TestUpdatesService(base.BaseTestCase):
         self.assertEquals(up['alias'], u'FEDORA-%s-a3bbe1a8f2' % YEAR)
         self.assertEquals(up['karma'], 1)
 
+    def test_list_updates_by_status_active_release(self):
+        res = self.app.get(
+            '/updates/', {"status": "pending", "active_releases": 'true'})
+        body = res.json_body
+        self.assertEquals(len(body['updates']), 1)
+
+        up = body['updates'][0]
+        self.assertEquals(up['title'], u'bodhi-2.0-1.fc17')
+        self.assertEquals(up['status'], u'pending')
+        self.assertEquals(up['request'], u'testing')
+        self.assertEquals(up['user']['name'], u'guest')
+        self.assertEquals(up['release']['name'], u'F17')
+        self.assertEquals(up['type'], u'bugfix')
+        self.assertEquals(up['severity'], u'medium')
+        self.assertEquals(up['suggest'], u'unspecified')
+        self.assertEquals(up['close_bugs'], True)
+        self.assertEquals(up['notes'], u'Useful details!')
+        self.assertEquals(up['date_submitted'], u'1984-11-02 00:00:00')
+        self.assertEquals(up['date_modified'], None)
+        self.assertEquals(up['date_approved'], None)
+        self.assertEquals(up['date_pushed'], None)
+        self.assertEquals(up['locked'], False)
+        self.assertEquals(up['alias'], u'FEDORA-%s-a3bbe1a8f2' % YEAR)
+        self.assertEquals(up['karma'], 1)
+
+    def test_list_updates_by_status_inactive_release(self):
+        rel = self.db.query(Release).filter_by(name='F17').one()
+        rel.state = ReleaseState.archived
+        self.db.commit()
+
+        res = self.app.get(
+            '/updates/', {"status": "pending", "active_releases": 'true'})
+        body = res.json_body
+        self.assertEquals(len(body['updates']), 0)
+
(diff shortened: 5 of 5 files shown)


pypingou:
The underlying idea is to make this page https://bodhi.fedoraproject.org/updates/?user=pingou&status=testing a little simpler as it's not really useful to include the F21/F24 updates.

bodhi/server/services/updates.py
```diff
@@ -308,8 +309,15 @@ def query_updates(request):
         query = query.filter(Update.date_pushed < pushed_before)
 
     releases = data.get('releases')
+    old_updates = data.get('old_updates')
     if releases is not None:
         query = query.filter(or_(*[Update.release == r for r in releases]))
+    elif old_updates is None:
```
bowlofeggs: This will default to a backwards-incompatible change in the API - let's make the new behavior opt-in instead of opt-out so the API is preserved. Perhaps the new parameter should be called "exclude_old_updates" or something like that.

Also, ```old_updates``` is always ```None``` so it is currently not possible to opt out and the UI button does not seem to work as a result. I think this is because we need to add this new parameter to the scemas here so it shows up in ```data```:

https://github.com/fedora-infra/bodhi/blob/2.11.0/bodhi/server/schemas.py#L388-L558

bowlofeggs (changes requested):
We should add tests for this change as well.

bowlofeggs:
You might want to reference https://github.com/fedora-infra/bodhi/issues/752 in your commit message.

bodhi/server/services/updates.py
pypingou: > This will default to a backwards-incompatible change in the API

Do we have a way to know when something is asked via the API or via the UI? I kinda really think displaying non-actionable updates in the UI isn't really useful.

> UI button does not seem to work as a result.

This is odd, it worked for me yesterday when I tried it :/
Indeed, adding it to ``schemas.py`` makes it work as well now, odd.

pypingou:
Unit-tests aside, I think this addresses your comment :)

pypingou:
Looks like jenkins failed but not on the tests :(

bowlofeggs:
On 10/03/2017 04:48 AM, Pierre-Yves Chibon wrote:
> Do we have a way to know when something is asked via the API or via the
> UI? I kinda really think displaying non-actionable updates in the UI
> isn't really useful.

We could use headers to figure that out, but I'd recommend just having
the UI put the parameter into the request so it's explicit.

bowlofeggs:
On 10/03/2017 11:19 AM, Pierre-Yves Chibon wrote:
> Looks like jenkins failed but not on the tests :(

I believe that is fixed now - if you rebase I think it'll work. Sorry
about that - I've been changing a lot of stuff with Bodhi's CI this week.

pypingou:
Still failing :(

bowlofeggs:
Ugh, sorry - I think it's some problem with code I wrote to pull the docs out:

```
+ parallel -v mv '/root/payload/test_results/{}/docs' 'docs-{}' ::: f25 f26 f27 rawhide
Academic tradition requires you to cite works you base your article on.
When using programs that use GNU Parallel to process data for publication
please cite:

  O. Tange (2011): GNU Parallel - The Command-Line Power Tool,
  ;login: The USENIX Magazine, February 2011:42-47.

This helps funding further development; AND IT WON'T COST YOU A CENT.
If you pay 10000 EUR you should feel free to use GNU Parallel without citing.

To silence the citation notice: run 'parallel --bibtex'.

mv /root/payload/test_results/f25/docs docs-f25
mv /root/payload/test_results/f26/docs docs-f26
mv /root/payload/test_results/f27/docs docs-f27
mv /root/payload/test_results/rawhide/docs docs-rawhide
+ exit 1
```

I've filed https://github.com/fedora-infra/bodhi/issues/1859 about that problem. Once I get bodhi-2.11.0 deployed I'll see if I can figure it out. Probably it's a path or permission problem or something.

bowlofeggs:
@pypingou Ah, I was actually incorrect - it seems that there is some code missing coverage in this patch:

```
-------------
Diff Coverage
Diff: origin/develop...HEAD, staged, and unstaged changes
-------------
bodhi/server/schemas.py (100%)
bodhi/server/services/updates.py (75.0%): Missing lines 316
-------------
Total:   5 lines
Missing: 1 line
Coverage: 80%
-------------
```

I forgot that the ```exit 1``` happens after the ```mv```'s happen - it wasn't the ```mv```'s that exited one. Sorry for the confusion.

pypingou:
Ok that makes sense since I didn't add any tests, before I do that I was curious to know if you like or agree with the current approach :)

bodhi/server/schemas.py
```diff
@@ -522,6 +522,12 @@ class ListUpdateSchema(PaginatedSchema, SearchableSchema, Cosmetics):
         missing=None,
     )
 
+    current_release = Releases(
```
bowlofeggs: I wonder if "active_release" would be more appropriate? This is OK too, just trying to think of a word that might more clearly map to the behavior.

bodhi/server/services/updates.py
```diff
@@ -308,8 +309,15 @@ def query_updates(request):
         query = query.filter(Update.date_pushed < pushed_before)
 
     releases = data.get('releases')
+    current_release = data.get('current_release')
     if releases is not None:
         query = query.filter(or_(*[Update.release == r for r in releases]))
+    elif str(current_release).lower() not in ['0', 'false']:
```
bowlofeggs: I think colander should already have turned this into a ```bool``` for you, since you added a schema for it.

bodhi/server/services/updates.py
```diff
@@ -308,8 +309,15 @@ def query_updates(request):
...
+        query = query.filter(
+            Update.release_id == Release.id
+        ).filter(
+            Release.state.in_([ReleaseState.current, ReleaseState.pending])
```
bowlofeggs: Since pending is also included (and I think reasonably so), I think we really should use the word 'active' instead of 'current'.

bodhi/server/schemas.py
bowlofeggs: Let's also pluralize this, so "active_releases".

bowlofeggs (changes requested):
Yeah the general approach seems good, and this does seem to work as I expect too. Let's just alter the parameter name and add some tests!

Oh, and let's fix the bool/str thing - I think it's already a bool (but you should verify that too).

bowlofeggs:
@pypingou Do you still want to work on this?

pypingou:
Sure, I'll be doing the adjustments :)

pypingou:
Adjusted and rebased :)

pypingou:
jenkies test

pypingou:
Ok, I'm not quite understanding why the line 320 isn't covered :(

bowlofeggs:
@pypingou I believe it is because the schema uses the name "active_releases", but the service looks for the name "active_release". I think you want to change ```active_release = data.get('active_release')``` to ```active_releases = data.get('active_releases')```.

Also, let's make the test create a second inactive release with an update to verify that the method only shows active releases when asked to. Additionally, let's add another test that also adds the second inactive release but that doesn't pass the new flag to verify that it is returned.

pypingou:
On Wed, Dec 06, 2017 at 06:36:28PM +0000, Randy Barlow wrote:
>    @pypingou I believe it is because the schema uses the name
>    "active_releases", but the service looks for the name "active_release". I
>    think you want to change active_release = data.get('active_release') to
>    active_releases = data.get('active_releases').

I'll fix that :)


>    Also, let's make the test create a second inactive release with an update
>    to verify that the method only shows active releases when asked to.
>    Additionally, let's add another test that also adds the second inactive
>    release but that doesn't pass the new flag to verify that it is returned.

Do you think you'd have time for these two?

bowlofeggs:
Yeah I can write the tests, but it might be a while before I have time.

pypingou:
jenkies test

pypingou:
Now without the debugging prints :)

bodhi/server/services/updates.py
```diff
@@ -364,7 +372,9 @@ def query_updates(request):
 
     user = data.get('user')
     if user is not None:
-        query = query.filter(Update.user == user)
+        query = query.filter(
+            Update.user == user
+        )
```
bowlofeggs: This change isn't related to the purpose of the patch, and I also happen to prefer the original styling.

bodhi/server/templates/user.html
```diff
@@ -91,7 +91,7 @@ <h3 class="pull-left nomargin nopadding">${user['name']}'s latest updates</h3>
         <a href="${urls['recent_updates_rss']}">
           RSS <span class="fa fa-rss"></span>
         </a>
-        <a href="${urls['recent_updates']}&status=testing">
+        <a href="${urls['recent_updates']}&status=testing&active_release=true">
```
bowlofeggs: When I visit http://localhost:6543/updates/?user=bowlofeggs&status=testing&active_release=true (as linked from "Testing" at http://localhost:6543/users/bowlofeggs), I see an "Exclude old releases" button, but the active_release=true *should* mean that they are already excluded. I think the problem is that this should be ```active_releases``` instead of ```active_release```.

bodhi/server/templates/updates.html
```diff
@@ -27,6 +27,23 @@ <h3>Updates <small>
...
+      <a href="${request.route_url('updates') + '?' \
+        + request.query_string.split('active_releases', 1)[0] \
+        + (request.query_string.split('active_releases', 1)[1].partition('&')[2] or '')}"
+         class="btn btn-sm btn-secondary">
+        Include old releases
+      </a>
+    </p>
+  % else:
+    <p>
+      <a href="${request.route_url('updates') + '?active_release=true&' + request.query_string}"
```
bowlofeggs: Similar to what I noted below, I think this should be ```active_releases``` instead of ```active_release```. Otherwise it piles up in the URL as you click the button repeatedly.

bowlofeggs (changes requested):

bowlofeggs (changes requested):
This still doesn't work correctly. If I visit [my updates page](http://localhost:6543/updates/?user=bowlofeggs&status=testing&active_releases=true) I am presented with "Exclude old releases" instead of the inclusion button. If I click on that button, the URL becomes http://localhost:6543/updates/?active_releases=true&user=bowlofeggs&status=testing&active_releases=true which has ```active_releases=true``` twice.

It seems like you might not be testing this patch before submitting it for review. If so, please test the patch yourself before submitting it here, because it takes a lot of time to go back and forth between us. Also, a patch that is tested by two people is better than a patch that is tested by one person.

pypingou:
On http://0.0.0.0:6543/users/pingou if I click on the ``Testing`` link I get http://0.0.0.0:6543/updates/?user=pingou&status=testing&active_releases=true which does exclude the old release and offers to include the new ones.
Navigate back and forth between include/exclude does seem to work for me.

Where did you find that first link?

pypingou:
Even when I visit these URLs in my local clone, I'm not seeing the behavior you're describing :(

bowlofeggs:
Hmm, it seems to work for me now. I wonder if I was using a stale branch or something. Sorry for the complaint, but it seems that it does work.

bowlofeggs (approved):

PR #1847, opened 2017-10-02
Query resultsdb for the results of the Atomic CI pipeline and display them

pypingou:
The way bodhi was querying resultsdb for test results was relying on the
specific structure of taskotron's results which lead to the results from
the Atomic CI pipeline to not be included.
With this commit we're querying for the specific structure of the Atomic
CI results so they are included in the "Automated tests" tab of the
update page.

### DIFF
--- a/bodhi/server/templates/update.html
+++ b/bodhi/server/templates/update.html
@@ -162,6 +162,9 @@ ${update.type} update for ${update.beautify_title(amp=True) | n}
         // seem to break anything
         var arch = result.data.arch;
         var item = result.data.item;
+        if (item === undefined && testcase == 'org.centos.prod.ci.pipeline.complete'){
+            item = result.data.original_spec_nvr[0] + '#' + result.data.rev;
+        }
         var submit_time = new Date(result.submit_time);
 
 
@@ -341,6 +344,8 @@ ${update.type} update for ${update.beautify_title(amp=True) | n}
     $("#tab-automatedtests").append(" <span class='fa fa-spinner fa-spin'></span>");
 
     for (index = 0, len = builds.length; index < len; ++index) {
+      request_results(base_url+"results/latest?original_spec_nvr="
+        +builds[index]+"&testcases=org.centos.prod.ci.pipeline.complete");
       request_results(base_url+"results/latest?item="+builds[index]+"&type=koji_build&testcases:like=dist.*");
     }
 


bowlofeggs:
The change looks OK to me, though I don't know a lot about this area of the code so I'd appreciate feedback from @kparal, @AdamWill, or @ryanlerch.

@pypingou, do you know of an update that has these kinds of tests that we can use to see how it renders? I looked at some recent kernel updates and they didn't seem to show these tests.

bodhi/server/templates/update.html
```diff
@@ -341,6 +347,8 @@
     $("#tab-automatedtests").append(" <span class='fa fa-spinner fa-spin'></span>");
 
     for (index = 0, len = builds.length; index < len; ++index) {
+      request_results(base_url+"results/latest?original_spec_nvr="
+        +builds[index]+"&testcases=org.centos.prod.ci.pipeline.complete");
```
kparal, line 348: This will not work, because in Koji the builds are named e.g. [cockpit-151-2.fc26](https://koji.fedoraproject.org/koji/buildinfo?buildID=975210) while centos pipeline results use [cockpit-151-2.f26](https://taskotron.fedoraproject.org/resultsdb/results/17433016) (notice the difference).

bodhi/server/templates/update.html
```diff
@@ -161,7 +161,13 @@
         // note: these are actually single-item arrays, though this doesn't
         // seem to break anything
         var arch = result.data.arch;
+        if (arch === undefined){
+            arch = 'noarch';
```
kparal: Interesting thing is that some of our tasks like `dist.rpmlint` don't fill out the `arch` field, and nothing breaks. Is this really necessary?

bodhi/server/templates/update.html
```diff
@@ -161,7 +161,13 @@
...
+        }
         var item = result.data.item;
+        if (item === undefined){
+            item = 'AtomicCi';
```
kparal: If I'm reading the code correctly, this will not work, since `item` is supposed to be NVR, so that if an update contains multiple NVRs, we compute the latest one for each of them. If you merge all item-less results under `AtomicCi`, you get just a single result for multi-build updates. Has this been tested? I might be wrong.

What you really want to do here is to use `original_spec_nvr` if it exists, I think. Ideally just when the testcase matches. Btw, is there a good reason that centosci tasks don't fill out `item`? At least for those testcases where it makes sense (like this one)?

bodhi/server/templates/update.html
pypingou: The pipeline is being fixed in https://github.com/CentOS-PaaS-SIG/ci-pipeline/pull/400

bodhi/server/templates/update.html
pypingou: Without this change and the one under I was getting a object whose first attribute was ``undefined`` making it rather hard to access what's underneath :)

bodhi/server/templates/update.html
pypingou: I have tested it with an update with a single build indeed, I'll see if I can adjust this to be ``original_spec_nvr`` as well.

As for why there is no ``item`` I have no idea, because @tflink designed it this way? What is ``item`` supposed to be?

pypingou:
> @pypingou, do you know of an update that has these kinds of tests that we can use to see how it renders?

I was testing this with an update for ``criu-3.3-5.fc26`` and I had this change as well to test it:
````
+      request_results(base_url+"results/latest?original_spec_nvr="
+        +builds[index].replace('.f', '.fc')+"&testcases=org.centos.prod.ci.pipeline.complete");
````

File: bodhi/server/templates/update.html (hunk 3 above)
Comment (kparal) [reply bodhi/server/templates/update.html]: `item` is supposed to be the common identifier of the thing that got tested. So for `type=koji_build`, it's NVR. For `type=bodhi_update`, it's update ID. See the usual `item+type` combinations here:
https://qa.fedoraproject.org/docs/libtaskotron/latest/library.html#module-libtaskotron.main

I don't exactly know what `org.centos.prod.ci.pipeline.complete` tests, can you explain? I don't know if there's a `type` that would match this (we can introduce a new one), but so far the `item` seems to look like Koji NVR (i.e. in [results/17433016](https://taskotron.fedoraproject.org/resultsdb/results/17433016) it seems that `cockpit-151-2.f26` is the thing that got tested). Does that sound correct?

If @tflink designed this, it might be a good idea to wait for him to provide details in here.

Comment (pypingou) [reply bodhi/server/templates/update.html]: If it is the thing that got tested, then I guess it should be the git hash of the commit that got tested. Basically the Atomic CI pipeline triggers for every commit made in dist-git. The  ``org.centos.prod.ci.pipeline.complete`` is then sent at the end of every run, successful or not, and it is then stored in resultsdb (as well as intermediary results).

But the git hash isn't really human friendly either :/
I'll try to put the ``original_spec_nvr`` instead so the UI is prettier

Comment (kparal) [reply bodhi/server/templates/update.html]: Well then, we already have exactly this `item+type` combination defined, an example:
```
type=dist_git_commit
item=modules/dhcp#5c6331592b6c92bf7c52a2ac735f92cb8fc11304
```
That's the right thing to submit to resultsdb, because then it will be easy to find results for dist git changes regardless of a testcase.

You see, the item+type standard is a way to make different testcases play along together nicely, to have a common standard. If we don't have that, each testcase will have its own approach for specifying what got tested, and the test consumers will be extremely difficult to write, having to know special cases for each testcase (you can see it here). So I'd very much recommend to use the aforementioned approach here. If all distgit tests use the same item+type definition, then we can query all the results, regardless whether they are centosci, taskotron or anything else.

One further complication here is that bodhi seems a bit unfit to display distgit commit results. Bodhi shows koji builds grouped to updates. Do I understand correctly that you want to show the distgit commit result for a specfile version that corresponds to the koji build NVR? And what happens if somebody pushed several commits without changing specfile version, you just take the latest one (which might or might not be correct)? Sure, this can be done this way, but wouldn't it make more sense to run the centosci test suite on actual koji builds (like we do with rpmlint etc), display those in bodhi, and make a different frontend for displaying distgit commit results, really after each commit (I'm assuming our resultsdb frontend is not sufficient)?

**Comment by pypingou** [comment]
>    Well then, we already have exactly this item+type combination defined, an
>    example:
> 
>  type=dist_git_commit
>  item=modules/dhcp#5c6331592b6c92bf7c52a2ac735f92cb8fc11304
> 
>    That's the right thing to submit to resultsdb, because then it will be
>    easy to find results for dist git changes regardless of a testcase.
> 
>    You see, the item+type standard is a way to make different testcases play
>    along together nicely, to have a common standard. If we don't have that,
>    each testcase will have its own approach for specifying what got tested,
>    and the test consumers will be extremely difficult to write, having to
>    know special cases for each testcase (you can see it here). So I'd very
>    much recommend to use the aforementioned approach here. If all distgit
>    tests use the same item+type definition, then we can query all the
>    results, regardless whether they are centosci, taskotron or anything else.

This sounds reasonable and fairly easy to implement.

I'll submit a PR on the listener for @tflink to review :)

 
>    One further complication here is that bodhi seems a bit unfit to display
>    distgit commit results. Bodhi shows koji builds grouped to updates. Do I
>    understand correctly that you want to show the distgit commit result for a
>    specfile version that corresponds to the koji build NVR? And what happens
>    if somebody pushed several commits without changing specfile version, you
>    just take the latest one (which might or might not be correct)?

We're querying for the latest yes.

> Sure, this
>    can be done this way, but wouldn't it make more sense to run the centosci
>    test suite on actual koji builds (like we do with rpmlint etc), display
>    those in bodhi, and make a different frontend for displaying distgit
>    commit results, really after each commit (I'm assuming our resultsdb
>    frontend is not sufficient)?

We've been discussing this pretty much since the start of the Atomic CI
pipeline, but the idea is the integrate the Atomic CI results in the packager's
workflow and to gate updates based on these results, so gating in bodhi a build
without explaining why, or sending the packager to a new interface, isn't ideal.

To me the best solution would be to get: https://pagure.io/koji/issue/550 fixed
so that there is a way to map a build to a git hash (and would also solve a
number of other problems, such as how to re-trigger a test run).

File: bodhi/server/templates/update.html (hunk 2 above)
Comment (pypingou) [reply bodhi/server/templates/update.html]: Ok apparently this is no longer needed, dropping it.

File: bodhi/server/templates/update.html (hunk 3 above)
Comment (pypingou) [reply bodhi/server/templates/update.html]: I managed to put the ``original_spec_nvr`` as well as the commit hash in there instead, was pretty straight forward at the end :)

This is going to be the good middle group I believe, but I'm willing to revisit if people feel otherwise.

**Comment by pypingou** [comment]
Small screenshot to give an idea of what it looks now:

screenshot from 2017-10-03 15-41-09

File: bodhi/server/templates/update.html [hunk 4]
```diff
@@ -162,6 +162,9 @@
         // seem to break anything
         var arch = result.data.arch;
         var item = result.data.item;
+        if (item === undefined){
```
kparal: I think it would be better to do
```
if (item === undefined && testcase == 'org.centos.prod.ci.pipeline.complete') {
```
it would then be clear that this is for this specific testcase.

bodhi/server/templates/update.html
```diff
@@ -162,6 +162,9 @@
...
+            item = result.data.original_spec_nvr[0] + '#' + result.data.rev;
```
kparal: Otherwise the code would crash here, if another result without item appears and doesn't have the required fields (`original_spec_nvr` and `rev`).

kparal (commented):
If I understand this correctly from the screenshot, for every NVR section there will be also NVR#hash section containing a single centosci testcase (if some result is present).

This makes the bodhi result page a bit longer, but on the other hand it's more correct - the specific commit was tested using internal centosci build, and not the regular koji build. So I think it's even better this way.

kparal:
One further note. If you add the `item+type` to results soon, please don't forget to adjust this code. The queries should be faster with `item` than with `original_spec_nvr`, I believe, because that's the common case we've optimized for.

bodhi/server/templates/update.html
pypingou: Good call, adjusted :)

pypingou:
> One further note. If you add the item+type to results soon, please don't forget to adjust this code

I'll wait for @tflink to comment on this before I change the current code, he definitely knows more than me so I suspect there might have been a good reason for this design choice.

kparal (approved):
Ack from me for the current code

tflink:
item+type is a taskotron-ism and the reporter code wasn't written with taskotron in mind. Unless @kparal knows something I don't, there is no difference in query optimization for any of the key-value pair fields (item, item_type, original_spec_nvr etc.).

I don't really see a reason why item+item_type is needed but I don't think it'd actively hurt anything if there's a good reason to make that change.

pypingou:
I was thinking about this a little more:
````
type=dist_git_commit
item=modules/dhcp#5c6331592b6c92bf7c52a2ac735f92cb8fc11304
````

Seems like the best actual item+type combination except that if we uses this, bodhi won't be able to query for it anyway since it doesn't know the hash.

Tbh, at this stage, I'm tempted to let the structure as is unless there is a real benefit to change it (such as optimization or so).

**Comment by jskladan** [comment]
There are no optimizations for `item/type` searches in resultsdb (or at least I don't know of any). The only reason to have those would be consistency (aka taskotronism), but if the `item` is unique anyway, I don't see any added benefit in doing that.

File: bodhi/server/templates/update.html [hunk 6]
```diff
@@ -341,6 +344,8 @@
     $("#tab-automatedtests").append(" <span class='fa fa-spinner fa-spin'></span>");
 
     for (index = 0, len = builds.length; index < len; ++index) {
+      request_results(base_url+"results/latest?original_spec_nvr="
+        +builds[index]+"&testcases=org.centos.prod.ci.pipeline.complete");
       request_results(base_url+"results/latest?item="+builds[index]+"&type=koji_build&testcases:like=dist.*");
```
jskladan, line 349: Just a small thing - please add `since=` parameter in ISO8601 format (UTC without tz) to limit the search for the `request_results`. Bodhi IIRC knows when the update was created (or something like that), so set it to that (or any other reasonable) value, so resultsdb does not need to crawl the whole db in order to find that "for some testcases, there is no result". Thanks!

kparal:
> Unless @kparal knows something I don't, there is no difference in query optimization for any of the key-value pair fields (item, item_type, original_spec_nvr etc.).

I was wrong, jskladan replied above.

> I don't really see a reason why item+item_type is needed but I don't think it'd actively hurt anything if there's a good reason to make that change.

The reason is that the results consumer then needs to know peculiarities of each and every testcase, instead of relying on conventions common for Fedora results. Compare this approach:
```
results/latest?item=NVR&type=koji_build&testcases:like=dist.*
```
with this approach:
```
results/latest?testcases:like=dist.rpmlint&nvr=NVR
results/latest?testcases:like=dist.rpmgrill&item=NVR
results/latest?testcases:like=dist.abicheck&spec_nvr=NVR
results/latest?testcases:like=dist.rpmdeplint&tested=NVR
results/latest?testcases:like=dist.upgradepath&subject=NVR
...
```
This is harder to maintain for consumer tools, and slower (I won't speculate about DB load again, but slower it is). That's why I think Bodhi should insist on keeping the Fedora conventions for results that should be shown in Bodhi, just to keep the code sane.

However, all my arguments are moot at the moment, because:

> Seems like the best actual item+type combination except that if we uses this, bodhi won't be able to query for it anyway since it doesn't know the hash.

More or less. It would improve things for any consumer wanting to show `dist_git_commit` type results (and it would allow maintainers to use our current resultsdb frontend to see per-commit results for their packages), and it wouldn't do any harm for Bodhi (you would still query by `original_spec_nvr`, or by using `item` prefix, e.g. `item:like=rpms/dhcp#*`), but it doesn't improve things for Bodhi either, for this moment. But once https://pagure.io/koji/issue/550 is completed, you could easily switch to querying by `item=rpms/name#hash` and receive exactly the one result you want. But it doesn't have to be done now, yes.

bodhi/server/templates/update.html
kparal: The builds were tested in centosci *before* the update was created, so I'm afraid there's no easy value to pass to `since=`. We could limit showing results to builds tested in the past 2 weeks, or a month, but it might be... limiting. Probably best to introduce that if we have performance issues.

We use `since=` when requesting results for `type=bodhi_update`, but we don't use it when requesting results for `type=koji_build` (see, another nice example of why conventions are good). And this is the same case as koji builds, I believe.

jskladan: Sure, that's where "something reasonable" goes. We already do have perf issues, and I'll be adding a "data since last month" defualt constraint to resultsdb anyway real-time-soon-now.

But if you don't mind sometimes waiting for a couple of minutes, before the data loads (if it loads), we could go with the unlimited approach.

pypingou: Doesn't using the /latest endpoint limit the results by default? (which should mean it's not scrolling the entire DB)

jskladan: @pypingou I understand the confusion, but no. `/latest` limits the data _output_, not the database contents. Say you query for `?item=foo&testcases=moo,bar`. It does something like `select * from results where item=foo and testcase=moo` + `select * from results where item=foo and testcase=bar` (disclaimer: this is far more complicated in fact, but on a high level is what is happening).
In this example, if the `foo/bar` pair is not present, the only way (for any db) to say "there is none" is to go through all the data (or a subset of data limited by index).
So the queries are limited in some way, but between the amount of data in resultsdb, the insufficient hardware the db has, and the postgres' query planner, looking for sparse data is quite slow. Thus narrowing the interval can help.
This is not needed, nor required, but take it as a tip from an insider :)

kparal: @rajcze So what do you think it's the reasonable value here? Is it better to fall back to the default value you'll be soon adding (so e.g. a month old bodhi updates will stop displaying results), or is it better to use a custom `since=` with a timestamp e.g. 2-4 weeks prior to the update creation? (The builds might be available and tested easily 2 weeks before the maintainer decides to submit them as an update).

Btw, we should use the same approach for `type=koji_build` query, not just this centosci one.

pypingou: While I think this is a good question and a chance we should look into (indeed not just for the Atomic CI but likely for other), do we want to bring this in another PR?

Also, the since = 2 weeks before the update means that in a year, resultsdb may still have to go through 1 year + 2 weeks of data. Would an upper bound help?
Something like since <update - 2 weeks> until <update + 1 week> ?

bowlofeggs:
On 10/03/2017 05:05 AM, Pierre-Yves Chibon wrote:
> To me the best solution would be to get:
> https://pagure.io/koji/issue/550 fixed
> so that there is a way to map a build to a git hash (and would also solve a
> number of other problems, such as how to re-trigger a test run).

I think the best solution is to do the gating in Pagure rather than in
Bodhi. Pagure is the natural place to display test results, and the
natural place for the developer to override them. Pagure also knows the
commit hashes. It doesn't make sense to me to invest engineering effort
for Bodhi to know the commit hashes when the end result is a system that
still isn't ideal.

TomasTomecek:
> I think the best solution is to do the gating in Pagure

Do you suggest that pull request workflow would be then mandatory?

pypingou:
> I think the best solution is to do the gating in Pagure rather than in Bodhi.

Well this PR isn't about gating but showing the test results :)

> It doesn't make sense to me to invest engineering effort for Bodhi to know the commit hashes when the end result is a system that still isn't ideal.

And we're not doing that presently :)

bowlofeggs:
@TomasTomecek I don't suggest making the PR process mandatory. Consider how projects on GitHub work, such as Bodhi. I *can* just make direct commits on the ```develop``` branch but I don't do that because I want my CI system to tell me if my change passes tests. (Technically it is against fedora-infra policy too, because we do actually require PR review, but that's tangential to my point - we don't require it for CI purposes.) The CI system itself is motivation for me to use the PRs. I think the same would be true for pull requests in Pagure.

bowlofeggs:
@pypingou Glad that we aren't planning to get commit hashes into Bodhi ☺

bowlofeggs (approved):
I don't personally see a problem with this PR though I also am not particularly familiar with this code. Since @kparal is a +1 that satisfies me, as he *is* pretty familiar with this code. Thanks @pypingou, @kparal and @rajcze!

bowlofeggs:
I'm going to rebase this, as I think the test failure has been fixed (and is unrelated anyway).

bowlofeggs:
Thanks @pypingou!

pypingou:
Thank you :)

bowlofeggs:
This has been cherry-picked to the 2.12 branch as c67d4ce1190e57bfaabded87d7ca0b1a81d61787, and will be included in the upcoming 2.12.0 release.

PR #1856, opened 2017-10-04
Collect test results and docs in their new locations.

bowlofeggs:


jeremycline (approved):
:shipit:

bowlofeggs:
I'm going to merge this without waiting on CI because the code changed here actually can't affect the results on its own job (because it only touches the service file). Thus a failure or success on this PR is meaningless.

PR #1858, opened 2017-10-04
Refuse starting a new push if one is in progress

puiterwijk:

### DIFF
--- a/bodhi/server/push.py
+++ b/bodhi/server/push.py
@@ -74,6 +74,10 @@ def push(username, cert_prefix, **kwargs):
                     update = session.query(Update).filter(Update.title == update).first()
                     updates.append(update)
         else:
+            if locked_updates:
+                click.echo('\nExisting push in progress. Please wait or resume')
+                return
+
             # Accept both comma and space separated request list
             requests = kwargs['request'].replace(',', ' ').split(' ')
             requests = [UpdateRequest.from_string(val) for val in requests]


bodhi/server/push.py
```diff
@@ -80,6 +80,10 @@ def push(username, cert_prefix, **kwargs):
                     update = session.query(Update).filter(Update.title == update).first()
                     updates.append(update)
         else:
+            if locked_updates:
+                click.echo('\nExisting push in progress. Please wait or resume')
+                return
```
bowlofeggs, line 79: Maybe an ```exit(1)``` would make sense?

bowlofeggs (changes requested):
We could use a test for this new if statement, but looks good otherwise.

bowlofeggs:

bowlofeggs:
I'm going to write a second patch for this PR with tests and my suggested ```exit(1)``` call.

bowlofeggs:
I started working on a patch for this, but then @puiterwijk and I decided that it would not be worth putting further effort into ```push.py``` since we plan to delete it in https://github.com/fedora-infra/bodhi/issues/2017. Thus, I am going to close this pull request.

For the record, here is the patch I came up with for this PR:

```
diff --git a/bodhi/server/push.py b/bodhi/server/push.py
index 2dd255d..0d04dd8 100644
--- a/bodhi/server/push.py
+++ b/bodhi/server/push.py
@@ -50,8 +50,8 @@ def push(username, cert_prefix, **kwargs):
 
     lockfiles = defaultdict(list)
     locked_updates = []
-    locks = '%s/MASHING-*' % config.get('mash_dir')
-    for lockfile in glob.glob(locks):
+    locks = glob.glob('%s/MASHING-*' % config.get('mash_dir'))
+    for lockfile in locks:
         with file(lockfile) as lock:
             state = json.load(lock)
         for update in state['updates']:
@@ -74,10 +74,6 @@ def push(username, cert_prefix, **kwargs):
                     update = session.query(Update).filter(Update.title == update).first()
                     updates.append(update)
         else:
-            if locked_updates:
-                click.echo('\nExisting push in progress. Please wait or resume')
-                return
-
             # Accept both comma and space separated request list
             requests = kwargs['request'].replace(',', ' ').split(' ')
             requests = [UpdateRequest.from_string(val) for val in requests]
@@ -115,6 +111,12 @@ def push(username, cert_prefix, **kwargs):
 
         if updates:
             click.confirm('Push these {:d} updates?'.format(len(updates)), abort=True)
+
+            if locked_updates:
+                msg = ('\nExisting push in progress. Lockfiles found: {}. Do you wish to push '
+                       'anyway?').format(', '.join(locks))
+                click.confirm(msg, abort=True)
+
             click.echo('\nLocking updates...')
             for update in updates:
                 update.locked = True
diff --git a/bodhi/tests/server/test_push.py b/bodhi/tests/server/test_push.py
index 5d944c1..8051724 100644
--- a/bodhi/tests/server/test_push.py
+++ b/bodhi/tests/server/test_push.py
@@ -413,7 +413,7 @@ class TestPush(base.BaseTestCase):
                         return_value=base.TransactionalSessionMaker(self.Session)):
             result = cli.invoke(push.push, ['--username', 'bowlofeggs'], input='y')
 
-        self.assertEqual(result.exit_code, 0)
         mock_init.assert_called_once_with(active=True, cert_prefix=u'shell')
         self.assertEqual(result.output, TEST_LOCKED_UPDATES_EXPECTED_OUTPUT)
         glob.assert_called_once_with('/mashdir//MASHING-*')
```

PR #1863, opened 2017-10-06
100% Test Coverage for server/services/comments.py

ryanlerch:
Signed-off-by: Ryan Lerch <rlerch@redhat.com>

PR #1864, opened 2017-10-06
100% Test Coverage for server/services/stacks.py

ryanlerch:

### DIFF
--- a/bodhi/tests/server/functional/test_stacks.py
+++ b/bodhi/tests/server/functional/test_stacks.py
@@ -12,8 +12,12 @@
 # along with this program; if not, write to the Free Software
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
 
-import mock
+import copy
 
+import mock
+from webtest import TestApp
+
+from bodhi.server import main
 from bodhi.server.models import Group, RpmPackage, Stack, User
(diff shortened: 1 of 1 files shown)


bodhi/tests/server/functional/test_stacks.py
```diff
@@ -13,7 +13,11 @@
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
 
 import mock
+import copy
 
+from webtest import TestApp
```
bowlofeggs: Similar import groupings here - let's put ```copy``` by itself in the first group, and let's group ```mock``` and ```webtest``` together.

bowlofeggs (approved):
Thanks for testing this! I've actually been thinking about dropping the stacks feature since it's unused. @ralphbean, do you know why the stacks feature exists, and why it's not used?

No need for re-review, so merge when you are ready.

bowlofeggs:
Odd. This one had the exact same error as #1863, also on F25. I wouldn't expect the tox failure to be Fedora release specific since it should just be pip installing its deps.

bowlofeggs:
jenkies test

bowlofeggs:
jenkies test

bowlofeggs (approved):
Thanks!

PR #1866, opened 2017-10-06
100% Test Coverage for server/config.py

ryanlerch:

### DIFF
--- a/bodhi/server/config.py
+++ b/bodhi/server/config.py
@@ -26,18 +26,13 @@ log = logging.getLogger('bodhi')
 def get_configfile():
     configfile = None
     setupdir = os.path.join(os.path.dirname(os.path.dirname(__file__)), '..')
-    if configfile:
-        if not os.path.exists(configfile):
-            log.error("Cannot find config: %s" % configfile)
-            return
+    for cfg in (os.path.join(setupdir, 'development.ini'),
+                '/etc/bodhi/production.ini'):
+        if os.path.exists(cfg):
+            configfile = cfg
+            break
     else:
-        for cfg in (os.path.join(setupdir, 'development.ini'),
-                    '/etc/bodhi/production.ini'):
-            if os.path.exists(cfg):
-                configfile = cfg
-                break
-        else:
-            log.error("Unable to find configuration to load!")
+        log.error("Unable to find configuration to load!")
     return configfile
 
 
--- a/bodhi/tests/server/test_config.py
+++ b/bodhi/tests/server/test_config.py
@@ -176,6 +176,14 @@ class BodhiConfigLoadConfig(unittest.TestCase):
         self.assertEqual(c['wiki_url'], 'test')
         self.assertEqual(get_appsettings.call_count, 0)
 
+    @mock.patch('bodhi.server.config.log.error')
+    @mock.patch('os.path.exists', return_value=False)
+    def test_get_config_unable_to_find_file(self, exists, log_error):
+        """Test we log an error if get_configfile() doenst find a config file"""
+        config.get_configfile()
+
+        log_error.assert_called_once_with("Unable to find configuration to load!")
+
 
 class BodhiConfigLoadDefaultsTests(unittest.TestCase):
     """Test the BodhiConfig._load_defaults() method."""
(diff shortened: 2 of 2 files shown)


bodhi/server/config.py
```diff
@@ -157,7 +152,7 @@ def _validate_fernet_key(value):
     """
     _validate_secret(value)
 
-    if isinstance(value, unicode):
+    if isinstance(value, str):
```
bowlofeggs: I'm curious why this change? The purpose of the original code was to turn a ```unicode``` object into a utf-8 encoded str. I think the code as-is will attempt to re-encode a str as UTF-8, but would only work if the str were correctly encoded in the env's locale. Additionally, it will not re-encode a ```unicode``` as an str which I think will fail if run through the Fernet engine later.

Perhaps it would be good to add a comment to the code here to explain that this is intended to ensure the value is an ```str```. In fact, I just noticed that the docblock I wrote for this function is incorrect because it says it returns a ```unicode``` :/ Would you mind adding such a comment and fixing the docblock? I can do it too, if you prefer.

bodhi/tests/server/test_config.py
```diff
@@ -370,7 +378,7 @@ def test_valid_key(self):
         result = config._validate_fernet_key(key)
 
         self.assertEqual(result, key)
-        self.assertTrue(type(result), str)
+        self.assertTrue(type(result), unicode)
```
bowlofeggs: I think it's actually important that the Fernet key be an str. IIRC, the cryptoengine will raise an Exception if it receives a ```unicode``` object.

bowlofeggs (changes requested):
Let's keep the Fernet validator using str's, but looks good otherwise!

bodhi/server/config.py
ryanlerch: Sorry, yeah this is my bad, my (now i realise was incorrect assumption) was that utf-8 == unicode, and with the mismatch with the docs, thought it was an error in the code (rather than an error in the docs...) -- will fix!

bowlofeggs: A good way to think about Python's ```str``` and ```unicode``` types is that a ```str``` is "dumb" - it's just an array of bytes and Python doesn't know its encoding. In fact, it might not even be a string at all in any encoding! A ```unicode``` is indeed text in a particular encoding that Python knows.

When you ```encode()``` a ```unicode``` you get a ```str``` in the requested encoding (if possible - some characters don't have code points in all encodings, like ☺ - it can't be encoded as ASCII). In the reverse, if you take an ```str``` that you know the encoding of, you can ```decode()``` it and get a ```unicode``` of it.

Here's some examples:

```
>>> unicode(u'☺').encode('utf-8')  # This is an str that contains utf-8 bytes that represent the smiley - note that the smiley is actually three bytes here
'\xe2\x98\xba'
>>> unicode(u'☺').encode('ascii')  # Oops - you can't do this because smiley isn't defined in the ascii code table
  File "<stdin>", line 1, in <module>
UnicodeEncodeError: 'ascii' codec can't encode character u'\u263a' in position 0: ordinal not in range(128)
>>> '☺'.decode('utf-8')  # My terminal is UTF-8, so i can type a smiley as UTF-8 and then decode it by telling python to expect it to be utf-8, and i'll get a nice u'' string
u'\u263a'
>>> 'é'.decode('utf-8').encode('latin-1')  # You can see é encoded as latin-1. Note that it is a single byte character in latin-1 (latin-1 only uses single byte codepoints)
'\xe9'
>>> 'é'.decode('utf-8').encode('utf-8')  # Or as utf-8, where it is a two-byte character!
'\xc3\xa9'
```

Anyways, encoding is frequently a point of confusion for developers. I cut my teeth on it at my previous employer where I did a lot of natural language processing on Spanish text, and Spanish text can come in a variety of encodings so we had to do a lot of guesswork sometimes!

ryanlerch:
@bowlofeggs this one is ready for a re-review! i think i got the encoding logic back to normal now! sprry about the mixup!

bodhi/server/config.py
```diff
@@ -26,18 +26,13 @@
 def get_configfile():
     configfile = None
     setupdir = os.path.join(os.path.dirname(os.path.dirname(__file__)), '..')
-    if configfile:
```
bowlofeggs, line 29: I didn't notice how hiliarious this code used to be when I first reviewed this PR ☺

bowlofeggs (approved):
Thanks!

PR #1867, opened 2017-10-06
100% Test Coverage for server/services/errors.py

ryanlerch:

### DIFF
--- a/bodhi/tests/server/services/test_errors.py
+++ b/bodhi/tests/server/services/test_errors.py
@@ -1,0 +1,42 @@
+# -*- coding: utf-8 -*-
+# Copyright © 2017 Red Hat, Inc.
+#
+# This file is part of Bodhi.
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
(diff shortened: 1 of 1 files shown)


bodhi/tests/server/services/test_errors.py
```diff
@@ -0,0 +1,36 @@
...
+    @mock.patch('mako.template.Template.render',
+                side_effect=[IOError('random error'), mock.DEFAULT])
+    def test_template_render_exception(self, template_exception):
+        """
+        Assert that we log an error if the error template renderer raises an exception
+        """
+        with self.assertRaises(IOError) as exc:
+            self.app.get('/pants', headers={'Accept': 'text/html'}, status=404)
+
+        self.assertEqual(str(exc.exception), 'random error')
```
bowlofeggs, line 37: Let's add an assertion that ```log.error()``` gets called with the output of the ```mako``` ```render()``` function.

bowlofeggs (approved):
Let's add one more assertion, but no need for re-review. Thanks!

bowlofeggs (approved):

PR #1873, opened 2017-10-08
Tar the test results so the docs can be retrieved.

bowlofeggs:

This pull request will necessarily fail CI testing because it changes the location of the nosetests.xml files. Once it is merged, all open PRs will need to be rebased upon this one to pass.


jeremycline (approved):
:shipit:

PR #1876, opened 2017-10-09
Make the modifyrepo accept a filetype for updates to repomd

puiterwijk:
Without this, any call to modifyrepo would overwrite updateinfo.xml, often resulting
in totally unusable repositories.

### DIFF
--- a/bodhi/server/metadata.py
+++ b/bodhi/server/metadata.py
@@ -274,26 +274,26 @@ class ExtendedMetadata(object):
         fd, name = tempfile.mkstemp()
         os.write(fd, self.uinfo.xml_dump().encode('utf-8'))
         os.close(fd)
-        self.modifyrepo(name)
+        self.modifyrepo('updateinfo', name)
         os.unlink(name)
 
-    def modifyrepo(self, filename):
+    def modifyrepo(self, filetype, filename):
         """Inject a file into the repodata for each architecture"""
         for arch in os.listdir(self.repo_path):
             repodata = os.path.join(self.repo_path, arch, 'repodata')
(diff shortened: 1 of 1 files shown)


bodhi/server/metadata.py
```diff
@@ -274,26 +274,26 @@ def insert_updateinfo(self):
         fd, name = tempfile.mkstemp()
         os.write(fd, self.uinfo.xml_dump().encode('utf-8'))
         os.close(fd)
-        self.modifyrepo(name)
+        self.modifyrepo('updateinfo', name)
         os.unlink(name)
 
-    def modifyrepo(self, filename):
+    def modifyrepo(self, filetype, filename):
         """Inject a file into the repodata for each architecture"""
```
bowlofeggs: Let's add an args section to this docblock:

```
Args:
    filetype (basestring): The type of repodata file to be injected. Valid values include 'updateinfo'.
    filename (basestring): The name of the file to inject, such as 'updateinfo.xml'.
```

If you know of more valid options for ```filetype```, please add them to that doc section.

bowlofeggs (changes requested):
The change seems reasonable to me, but we'll need the tests to pass. It seems like some other code might just need to start passing the ```filetype```.

It also seems this PR is failing due to the mysterious F25 pydocstyle issue. I'm working on fixing that in https://github.com/fedora-infra/bodhi/pull/1877 so you'll probably just need to rebase once I get a fix merged.

bowlofeggs (approved):
Excellent, thanks for the patches! Once Jenkins passes it, we can merge. We will need this for the modular mashing release, which is currently planned for 2.13.0.

PR #1877, opened 2017-10-09
Add configparser to the f25-packages for CI for pydocstyle.

bowlofeggs:


bowlofeggs:
Somehow tests started passing again without this change, so I think we can close this PR.

PR #1880, opened 2017-10-10
Tag all CI output by container release.

bowlofeggs:
This commit adjusts the CI script so that it prepends the Fedora
release version on each line of output from the test script. This
makes it easier to identify which container tests's output a
developer is looking at during debugging.


jeremycline (approved):
:shipit:

PR #1882, opened 2017-10-10
 Use the Alabaster themes to make Bodhi docs prettier

jeremycline:
I did this during Flock, but forgot about it.

Adjust some of the Alabaster theme settings to make Bodhi's docs look nicer. This adds a sidebar with docs-wide navigation (and the logo) as well as widens the docs slightly to 1040px.

### DIFF
--- a/docs/_static/logo.svg
+++ b/docs/_static/logo.svg
@@ -1,0 +1,1 @@
+../../bodhi/server/static/img/bodhi-logo.svg
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -99,7 +99,13 @@ html_theme = 'alabaster'
 # Theme options are theme-specific and customize the look and feel of a theme
 # further.  For a list of options available for each theme, see the
 # documentation.
-# html_theme_options = {}
+html_theme_options = {
+    'logo': 'logo.svg',
+    'logo_name': True,
+    'github_user': 'fedora-infra',
+    'github_repo': 'bodhi',
+    'page_width': '1040px',
+}
 
 # Add any paths that contain custom themes here, relative to this directory.
 # html_theme_path = []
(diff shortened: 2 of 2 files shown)


docs/_static/logo.svg
```diff
@@ -0,0 +1,55 @@
...
+      <path sodipodi:nodetypes="cssssc" id="path2162" d="M 23.895199,20.752745 C 22.568192,22.726992 21.112309,23.64611 18.451414,25.729754 C 15.300944,28.196766 15.178276,36.780379 16.483434,38.98804 C 19.24814,43.664511 29.483434,43.929216 31.659905,39.458628 C 32.719433,37.282301 32.59065,28.15039 29.21612,25.516519 C 25.171133,22.359345 25.685839,23.006404 23.895199,20.752745 z " style="fill:url(#linearGradient3321);fill-opacity:1;fill-rule:evenodd;stroke:url(#linearGradient3323);stroke-width:3;stroke-linecap:butt;stroke-linejoin:miter;stroke-miterlimit:4;stroke-dasharray:none;stroke-opacity:1"/>
+    </g>
+  </g>
+</svg>
```
bowlofeggs: This file appears to be identical to the one at ```bodhi/server/static/bodhi-logo.svg``` - should we use a symlink instead of copying it?

bowlofeggs (changes requested):
Looks very nice! Let's just see if we can use a symlink on that svg file. Thanks!

bowlofeggs (approved):
I meant to mark it as approved with one tiny change request, but I don't see how to change my last one. So feel free to make that svg be a symlink to the existing one and merge without re-review. Sorry for confusion!

docs/_static/logo.svg
jeremycline: I don't know why I copied it before, but I have made it so

PR #1885, opened 2017-10-10
Fix multiple mashes racing for the comps directory

puiterwijk:
Without this, multiple running mashes might get in each others way with the comps
repository.
If one mash is just performing the "make", and another starts with the "git pull"
phase, the "make" might hit on filees that the "git pull" is still writing out,
leading to incorrect XML files being parsed, and masher erroring out.

This patch makes the comps repos be cloned into a directory per mash.


bowlofeggs (approved):

bowlofeggs:
I approve @puiterwijk's commit, but have added three of my own on top that I need review on from somebody. This will become a 2.12.1 release.

puiterwijk:
@bowlofeggs The extra commits look good to me. (I can't GH-approve the PR).

PR #1890, opened 2017-10-11
Make the Update Edit button green for pushed Updates.

bowlofeggs:

### DIFF
--- a/bodhi/server/templates/update.html
+++ b/bodhi/server/templates/update.html
@@ -514,7 +514,7 @@ $(document).ready(function(){
               <a id='revoke' class="btn btn-sm btn-danger"><span class="fa fa-fw fa-arrow-circle-left"></span> Revoke</a>
             % endif
           % elif update.pushed and (update.status.description != 'stable' or (update.status.description == 'stable' and 'releng' in [group.name for group in request.user.groups])):
-            <a id='edit' class="btn btn-sm btn-default"><span class="fa fa-fw fa-pencil"></span> Edit</a>
+            <a id='edit' class="btn btn-sm btn-primary"><span class="fa fa-fw fa-pencil"></span> Edit</a>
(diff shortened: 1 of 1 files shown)


ryanlerch:
Thanks for the PR @bowlofeggs !

Just wondering what the thinking is behind changing this button to green?

bowlofeggs:
On 10/12/2017 07:31 PM, Ryan Lerch wrote:
> Just wondering what the thinking is behind changing this button to green?
Hey Ryan!

I only changed it so that it would be consistent with how the button looks when the update hasn't been pushed yet. My goal wasn't actually to make it green - it was just to make it the same. The template currently looks like [this](https://github.com/fedora-infra/bodhi/blob/3cd80ea3d998a14585f5279148f604c65da6f92e/bodhi/server/templates/update.html#L499-L517):

```
      % if can_edit:
        % if not update.locked:
          <div id='updatebuttons' class="btn-group pull-right" role="group" aria-label="...">
          % if not update.pushed:
            <a id='edit' class="btn btn-sm btn-primary"><span class="fa fa-fw fa-pencil"></span> Edit</a>
            
           ...<SNIP>...

          % elif update.pushed and (update.status.description != 'stable' or (update.status.description == 'stable' and 'releng' in [group.name for group in request.user.groups])):
            <a id='edit' class="btn btn-sm btn-default"><span class="fa fa-fw fa-pencil"></span> Edit</a>
```

Note that an unpushed update has an Edit button that is "primary", but a pushed one has "default". Primary leads to green, and default leads to white (or clear?).

I actually wasn't so much wanting it to be green as I was thinking it'd make sense to look the same on the two sides of the if statement.

What do you think?

ryanlerch (approved):

ryanlerch:
@bowlofeggs makes sense to me! thanks for providing the justification! :D

PR #1891, opened 2017-10-12
Update for fixupdaterepo branch

mcurlej:
Hi @puiterwijk ,

so this is an update for https://github.com/fedora-infra/bodhi/pull/1876. I looked into why the tests are not passing. The problem was on line https://github.com/fedora-infra/bodhi/blob/fixupdaterepo/bodhi/server/metadata.py#L285. Both arguments for join where full paths. So the outcome was always the second path (filename var) and on line 286 target_fname had the value of filename var. So it wanted to copy on its self. So i did little bit of refactoring and added needed dockblock and comments so its clear what every line does. Also discussed some stuff with the autor of createrepo_c so we know more about the types that can be inserted into repomd.xml. All test of bodhi now pass. So WDYT?

### DIFF
--- a/bodhi/server/metadata.py
+++ b/bodhi/server/metadata.py
@@ -271,25 +271,45 @@ class ExtendedMetadata(object):
         self.uinfo.append(rec)
 
     def insert_updateinfo(self):
-        fd, name = tempfile.mkstemp()
+        fd, tmp_file_path = tempfile.mkstemp()
         os.write(fd, self.uinfo.xml_dump().encode('utf-8'))
         os.close(fd)
-        self.modifyrepo('updateinfo', name)
-        os.unlink(name)
+        self.modifyrepo('updateinfo', 'updateinfo.xml', tmp_file_path)
+        os.unlink(tmp_file_path)
 
-    def modifyrepo(self, filetype, filename):
-        """Inject a file into the repodata for each architecture"""
+    def modifyrepo(self, filetype, filename, tempfile):
+        """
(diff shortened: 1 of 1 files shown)


bodhi/server/metadata.py
```diff
@@ -271,25 +271,45 @@ def add_update(self, update):
...
+        Inject a file into the repodata for each architecture with the help of createrepo_c.
+
+        Args:
+            filetype (basestring): what type of metadata will be inserted by createrepo_c.
+                This does allow any string to be inserted (custom types). There are some
+                types which are used with dnf repos as primary, updateinfo, comps, filelist etc.
+            filename (basestring): the actual name of the metadata file which will be inserted
+                (createrepo_c takes this as one of its arguments so the ouput file will be
+                {hash}-{filename}.xz)
+            tempfile (basestring): a temp file path. File holds the dump of metadata untill
```
bowlofeggs, line 291: s/File/The file/
s/untill/until/

bowlofeggs (approved):
Just a few suggested grammatical changes, but the code looks good to me. I'd also like @puiterwijk to review it, since it's a patch on his patch.

bodhi/server/metadata.py
bowlofeggs: Since it's late in the day in CZ, I'm going to go ahead and merge this PR as-is - I'll file a new PR to correct these two typos myself - they're low priority anyway. Thanks for the patch!

PR #1896, opened 2017-10-16
PEP-257 bodhi.server.mail.

bowlofeggs:


bowlofeggs:
Ah, looks like the test failure is due to bzip2 not being installed when CI results are gathered. This is odd, since it used to be installed when CI was running (and didn't have a problem when the patch that started bzipping the results was merged).

I'm going to submit a new PR to explicitly install bzip2 in the containers.

bowlofeggs:
I filed https://github.com/fedora-infra/bodhi/pull/1903 to fix the bzip2 issue.

ryanlerch (approved):

PR #1901, opened 2017-10-16
Remove support for USERNAME environment varaible from CLI tools.

### LINKED ISSUE #1789: The client should not use USERNAME
The CLI can use the ```USERNAME``` environment variable to identify a user's username. This name is problematic because GDM sets ```USERNAME```, and not everyone uses the same username on their desktop machine as they do for their FAS username.

We have a few options:

1. Don't use environment variables at all to identify usernames. This was used to work around the way that the bodhi 2 client used to do auth - it didn't have a good way before to know the username other than requiring users to type ```--username bowlofeggs``` on each command, which was a pain. Now it has the ability to use past successful authentications to determine the username, and can also prompt for the username if there isn't an existing cookie. Unfortunately, this would be a backwards-incompatible change.
2. Add a new ```BODHI_USER``` environment variable. The namespacing would make this unique enough not to conflict with other things. If this is present, it would supercede the ```USERNAME``` environment variable, and we could remove support for ```USERNAME``` in a future release of Bodhi.
3. Same as 2, but go ahead and also drop support for ```USERNAME```.

I suppose the main question in my mind is whether it is time to go ahead and release a Bodhi 3. I'm not afraid to rev the X bit, but I'd prefer not to do it too often. Perhaps 2a would be good to do for now?

bowlofeggs:
This is a backwards incompatible change, so this commit also
adjusts the bodhi version to be 3.0.0.

fixes #1789


bowlofeggs:
This needs to be rebased on https://github.com/fedora-infra/bodhi/pull/1903 for the CI tests to pass.

ryanlerch (approved):

igor-raits:
lgtm, probably would be nice to add support of using auto-detecting username from kerberos ticket as koji / whatsoever does. Although this is just RFE and has nothing to do with this PR.

igor-raits (approved):

bowlofeggs:
@ignatenkobrain feel free to file an RFE, though I'll point out that Bodhi doesn't use kerberos tickets and there are not plans for it to use them in the future.

Also note that Bodhi does remember your username once you've authenticated one time.

bowlofeggs:
This has been cherry-picked to the 3.0 branch as a7a86cf403ec79f968414a524916cb369b21f280.

PR #1903, opened 2017-10-16
Install bzip2 in all CI containers as it is needed to tar results.

bowlofeggs:
Signed-off-by: Randy Barlow <randy@electronsweatshop.com>

PR #1904, opened 2017-10-16
Copy the test_results/ folder back to the Jenkins master when done.

bowlofeggs:
This one is needed to get CI passing on https://github.com/fedora-infra/bodhi/pull/1903 which is needed to get CI passing on all other PRs.

Basically, I had forgotten to copy the test_results/ folder back to the Jenkins master after the tests finished on the slave, and as a result Jenkins fails to find all the archived results.

This PR will necessarily fail CI because of the problem it solves (the JJB template is being changed by this commit, but Jenkins will run the tests against with with the old template.)


puiterwijk (approved):

bowlofeggs:
Thanks for the review - I'm going to override the protected branch feature and go ahead and merge so we can get the JJB template updated.

PR #1910, opened 2017-10-19
Get rid of Bunch dependency and use collections.namedtuple instead

frenzymadness:
As a first step of porting Bodhi to Python 3 compatible form, we have to deal with dependencies.

Bunch is a really obsolete package with the last release in December 2011 and it probably won't be ported to Python 3. Because it is used only in FakeBugTracker, I think it can be easily replaced by namedtuple from the standard library.

If it's not necessary to access `bug_id` as an attribute, I can make it more simple and use a dictionary instead namedtuple.

### DIFF
--- a/bodhi/server/bugs.py
+++ b/bodhi/server/bugs.py
@@ -15,7 +15,7 @@
 import logging
 import xmlrpclib
 
-from bunch import Bunch
+from collections import namedtuple
 from kitchen.text.converters import to_unicode
 import bugzilla
 
@@ -24,6 +24,7 @@ from bodhi.server.config import config
 
 bugtracker = None
 log = logging.getLogger('bodhi')
+FakeBug = namedtuple('FakeBug', ['bug_id'])
 
 
 class BugTracker(object):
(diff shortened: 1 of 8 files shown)


puiterwijk:
@centos-ci test this please

bodhi/server/bugs.py
```diff
@@ -24,6 +24,7 @@
 
 bugtracker = None
 log = logging.getLogger('bodhi')
+FakeBug = namedtuple('FakeBug', 'bug_id')
```
puiterwijk: Personally, I would prefer `FakeBug = namedtuple('FakeBug', ['bug_id'])`.

puiterwijk:
@centos-ci test this please

puiterwijk (approved):
Looks good to me, thanks!

bodhi/server/bugs.py
frenzymadness: Ok, I used `string` because there is only one attribute but I have nothing against using `list` here. Fixed and rebased.

PR #1911, opened 2017-10-19
Build and use caches for the CI containers

puiterwijk:

### DIFF
--- a/devel/refresh_cache.sh
+++ b/devel/refresh_cache.sh
@@ -25,22 +25,7 @@
 
 RELEASES=${RELEASES:="f25 f26 f27 rawhide pip"}
 
-if [[ $@ == *"-x"* ]]; then
-    FAILFAST="--halt now,fail=1"
-    PYTEST_ARGS="-x"
-else
-    FAILFAST=""
-    PYTEST_ARGS=""
-fi
-
-PARALLEL="parallel -v $FAILFAST --tag"
-
-
-tar_results() {
-    tar cjf test_results.tar.bz2 test_results/
-    mv test_results.tar.bz2 test_results/
-}
-
+PARALLEL="parallel -v --tag"
 
 # Assemble the Dockerfile snippets into Dockerfiles for each release.
 pushd devel/ci/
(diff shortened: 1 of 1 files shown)


devel/refresh_cache.sh
```diff
@@ -0,0 +1,55 @@
...
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+
+# This script is used to run the tests in a variety of parallel
+# containers to test Bodhi across supported Fedora releases (See the RELEASES variable below). You
+# can set the RELEASES environment variable to override the RELEASES for a given test run if you
+# like. You can pass a -x flag to it to get it to exit early if a build or test run fails.
+# It is intended to be run with sudo, since it needs to use docker.
+
+RELEASES=${RELEASES:="f25 f26 f27 rawhide pip"}
```
bowlofeggs, line 26: I wonder if we should just create a file in here called ```RELEASES``` with this string so that the two scripts can share. I could see myself forgetting to update it.

devel/refresh_cache.sh
```diff
@@ -0,0 +1,55 @@
...
+        # Add the common rpm packages for the non-pip releases.
+        cat rpm-packages >> Dockerfile-$r
+    else
+        # Let's use F26 for the pip tests.
+        sed -i "s/FEDORA_RELEASE/26/" Dockerfile-$r
+    fi
+    cat $r-packages Dockerfile-footer >> Dockerfile-$r
+    echo "COPY . /bodhi" >> Dockerfile-$r
+done
+popd
```
bowlofeggs, line 44: This block seems the same as the one in ```run_tests.sh``` - should we break it out into something shared? I was actually thinking about using a ```Makefile``` to do this as you had suggested to me.

devel/refresh_cache.sh
```diff
@@ -0,0 +1,55 @@
...
+
+# Insert the container tag to pull for each release. There's a substitution in the parallel {}'s
+# that will remove the "f" from the releases since Fedora uses just the number of the release in its
+# tags.
+$PARALLEL sed -i "s/FEDORA_RELEASE/{= s:f:: =}/" devel/ci/Dockerfile-{} ::: $RELEASES
+# Try to pull existing containers
+$PARALLEL "docker pull docker.io/puiterwijk/bodhi-cache:{} || :" ::: $RELEASES
```
bowlofeggs, line 51: Since we're building the cache here, why not just pull from fedora's registry?

devel/refresh_cache.sh
```diff
@@ -0,0 +1,55 @@
...
+# Build the containers.
+$PARALLEL "docker build --pull -t docker.io/puiterwijk/bodhi-cache:{} -f devel/ci/Dockerfile-{} . || (echo \"JENKIES FAIL\"; exit 1)" ::: $RELEASES || (echo -e "\n\n\033[0;31mFAILED TO BUILD IMAGE(S)\033[0m\n\n"; exit 1)
```
bowlofeggs, line 53: What do you think of naming it bodhi-test instead of bodhi-cache? It is for caching, but even more than that it's for testing. Since it'll be named this way in devs' local docker runtimes, it might be good to make its purpose for testing clearer.

bowlofeggs (changes requested):

bowlofeggs:

bowlofeggs:
I'm going to close this due to inactivity, but feel free to reopen if you want to finish it!

PR #1912, opened 2017-10-19
Document the features and bugs for 3.0.0.

bowlofeggs:


jeremycline (approved):
I have one request which is that the mashing process now be referred to as "punging" now that mash is not being used.

:shipit:
 3️⃣ 0️⃣ 0️⃣

PR #1913, opened 2017-10-19
Remove the dependency on fedmsg-atomic-composer.

bowlofeggs:
This will need to be backported to the 3.0 branch as well, as 3.0 will not build as-is.


puiterwijk (approved):

bowlofeggs:
This was cherry picked to the 3.0 branch as 62217f1ab838c54d61422e3a116832424c94b112.

PR #1916, opened 2017-10-20
Don't assume the variant name.

ralphbean:
I ran into this while testing in staging.

The variant name we're using for the modular stuff is just "Server".  We could
change that to be "Everything" in the modular variants config, and all of this
would work with no code change to Bodhi.

However, mirrormanager has already been configured to look for the Server
folder (aiui).  Is it easier to just do this in Bodhi?


ralphbean:
@bowlofeggs said he'd think about solving this some other way.  So, closing for now.

PR #1917, opened 2017-10-20
Process the prefer_ssl setting.

bowlofeggs:


puiterwijk (approved):
I would personally suggest renaming this setting to avatar_enforce_ssl or something. But this patch itself looks good to me.

bowlofeggs:
Renaming the setting would unfortunately be a backwards incompatible change, so would require another X release. I agree with the suggestion though, so I've filed a ticket to remember to do this the next time we have an X release: https://github.com/fedora-infra/bodhi/issues/1921

PR #1924, opened 2017-10-23
Fix PEP8 style failures on E722

ryanlerch:
After re-provisioning my vagrant box today, i started getting failures on E722 complaining about blank except: lines.

So this PR Changes a bunch of bare excepts to except Exception: so the style tests pass where they were previously failing on E722

### DIFF
--- a/bodhi/tests/server/functional/test_generic.py
+++ b/bodhi/tests/server/functional/test_generic.py
@@ -15,6 +15,7 @@
 from datetime import datetime
 import copy
 
+import mock
 from pyramid.testing import DummyRequest
 from webtest import TestApp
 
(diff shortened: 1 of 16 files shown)


bowlofeggs:
You could add a "fixes #1927" into your commit message.

Looks like there were just a couple lines touched that are missing test coverage:

```
pip	bodhi/client/bindings.py (100%)
pip	bodhi/server/bugs.py (50.0%): Missing lines 162,183
pip	bodhi/server/consumers/masher.py (66.7%): Missing lines 672
pip	bodhi/server/mail.py (0.0%): Missing lines 460
pip	bodhi/server/services/errors.py (100%)
pip	bodhi/server/util.py (100%)
pip	bodhi/server/views/generic.py (0.0%): Missing lines 335
```

I'm willing to write tests to cover those lines for you, but it might take a few days.

bowlofeggs:
BTW, I do have a commit on develop that does temporarily work around this problem by disabling the check. This commit is definitely better, I was just looking for a quick fix at the time.

https://github.com/fedora-infra/bodhi/commit/af7662e6461193b8433aee498d712ce0659ed159

bowlofeggs:
Oh also, it would be good to undo the change I made in my commit with this commit, since it fixes the real problem.

ryanlerch:
jenkies test

ryanlerch:
@bowlofeggs done!

also, this commit makes the test coverage over 93%, so i tweaked that too. :D

bodhi/tests/server/functional/test_generic.py
```diff
@@ -342,6 +343,22 @@ def test_latest_builds(self):
...
 
+    @mock.patch("bodhi.server.buildsys.DevBuildsys.getLatestBuilds", side_effect=Exception())
+    def test_latest_builds_exception(self, mock_getlatestbuilds):
+        """
+        Test that the latest_builds() just passes if it hits an exception
+        the result here is that there should be no builds returned
+        """
+        res = self.app.get('/latest_builds')
+        body = res.json_body
+        print body
```

bowlofeggs (approved):
Looks great! Let's drop that one print statement and merge this bad boy. Thanks for the tests, you've been killing it lately!

bowlofeggs:
Oh, don't forget to put "fixes #1927" into your commit message so the bug gets linked and closed.

bodhi/tests/server/functional/test_generic.py
ryanlerch: gah! thanks for spotting -- it seems i am destined to always leave one of these lying around!

PR #1929, opened 2017-10-24
Use feedgen instead of webhelpers.feedgenerator

frenzymadness:
Webhelpers package is obsolete and without Python 3 support so I replaced it with the new feedgen package.

There are a few changes:
* FeedGenerator has to be initialized empty and all attributes added via methods later.
* Link has to be initialized from a dictionary or kwargs with at least `href` kwarg. It's the reason why I changed `linker` function.
* Pubdate has to contain timezone info so I replaced itemgetter with the lambda function.

I've tested everything manually and all tests are passing as well. This PR is still WIP because I'll create review request for python-feedgen later (today or tomorrow) and I'll add a link to the comments in main.yml, but all code is ready for a review.


frenzymadness:
Package review for python-feedgen is here: https://bugzilla.redhat.com/show_bug.cgi?id=1505993 Link also added to comment in `main.yml` file.

This PR is no longer WIP.

bowlofeggs:
jenkies test

bowlofeggs:
I've approved your package review, but since that'll take some time to stabilize we can keep the pip installs for now and I'll just switch them to rpm installs when they hit the repos.

bowlofeggs:
Cool, this refactor does seem to work. I tested it with Thunderbird and it seems to process the updates RSS feed just fine. Let's wait and see what Jenkies thinks.

bowlofeggs (changes requested):
Cool, this looks great!

Can you edit the devel section at the top of ```docs/release_notes.rst``` to mention that webhelpers is no longer a required dep and that feedgen is a newly required dep? I think that's all we need to do to get this merged. Thanks!

frenzymadness:
@bowlofeggs Release notes added.

bowlofeggs (approved):
This looks great, thanks! I'll merge once CI passes.

bowlofeggs:
jenkies test

PR #1931, opened 2017-10-24
Fix the path to the lock files in push.py

puiterwijk:

### DIFF
--- a/bodhi/server/push.py
+++ b/bodhi/server/push.py
@@ -42,21 +42,15 @@ import bodhi.server.notifications
               help='Push updates with a specific request (default: testing,stable)')
 @click.option('--resume', help='Resume one or more previously failed pushes',
               is_flag=True, default=False)
-@click.option('--staging', help='Use the staging bodhi instance',
-              is_flag=True, default=False)
 @click.option('--username', prompt=True)
 @click.version_option(message='%(version)s')
 def push(username, cert_prefix, **kwargs):
     """Push builds out to the repositories."""
-    staging = kwargs.pop('staging')
     resume = kwargs.pop('resume')
 
     lockfiles = defaultdict(list)
     locked_updates = []
-    if staging:
-        locks = '/var/cache/bodhi/mashing/MASHING-*'
-    else:
-        locks = '/mnt/koji/mash/updates/MASHING-*'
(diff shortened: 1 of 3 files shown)


puiterwijk:
I considered keeping the `--staging` flag as a no-op for backwards compatibility, but given that we're making this a Bodhi 3.0 anyway, I figured I'd just do away with it.

bodhi/server/push.py
```diff
@@ -42,21 +42,15 @@
...
-    staging = kwargs.pop('staging')
     resume = kwargs.pop('resume')
 
     lockfiles = defaultdict(list)
     locked_updates = []
-    if staging:
-        locks = '/var/cache/bodhi/mashing/MASHING-*'
-    else:
-        locks = '/mnt/koji/mash/updates/MASHING-*'
+    locks = '%s/MASHING-*' % config.get('mash_dir')
```
bowlofeggs, line 53: Yeah good call on killing ```--staging``` - since ```mash_dir``` is part of staging's config, it's not needed anyway.

bowlofeggs (approved):

bodhi/server/push.py
```diff
@@ -42,21 +42,15 @@
               help='Push updates with a specific request (default: testing,stable)')
 @click.option('--resume', help='Resume one or more previously failed pushes',
               is_flag=True, default=False)
-@click.option('--staging', help='Use the staging bodhi instance',
-              is_flag=True, default=False)
```
bowlofeggs, line 46: Can we add a note to ```docs/release_notes.rst``` about this?

bowlofeggs (changes requested):

bowlofeggs (approved):

puiterwijk:
Release notes changes look good. :+1:

bowlofeggs:
jenkies test

bowlofeggs:
In the interest of time, @puiterwijk and I have agreed that running the ```devel/run_tests.sh``` script locally would be an acceptable subtitute for waiting on CI to pass this build. The tests pass, and there is minimal risk that Jenkies would disgree so I am going to merge this.

bowlofeggs:
This was cherry picked to the 3.0 branch as c95ba78580c3b70224146c4752c387319d9e6eab.

PR #1933, opened 2017-10-24
Let CI work with concurrent builds

puiterwijk:


bowlofeggs (approved):
OMG how did not know this before.

PR #1948, opened 2017-10-27
Use plain ASCII for alembic template

tyll:
For unknown reasons alembic strips the first line of the template when
creating a migration script. Then the utf-8 characters leads to an
encoding error.

### DIFF
--- a/bodhi/server/migrations/script.py.mako
+++ b/bodhi/server/migrations/script.py.mako
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-# Copyright © YEAR AUTHOR
+# Copyright (c) YEAR AUTHOR
 #
 # This file is part of Bodhi.
 #


puiterwijk:
@centos-ci please test

puiterwijk:
@centos-ci test this please

bodhi/server/migrations/script.py.mako
```diff
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
```
bowlofeggs, line 1: Let's drop this line too, since it was only here to accomodate the © character.

bowlofeggs (changes requested):
Thanks!

bowlofeggs:
I just hit this bug so I'm going to merge this. I'll make a new PR with my requested change to drop that encoding header.

bodhi/server/migrations/script.py.mako
bowlofeggs: I opened #2037 with this suggested change.

PR #1950, opened 2017-10-29
Run a max number of mash threads at the same time

puiterwijk:


puiterwijk:
This is needed because seemingly NFS does not like 4 createrepo's on the same volume at the same time.

puiterwijk:
The test failures are errors installing packages in the containers

puiterwijk:
@centos-ci retest this please

production.ini
```diff
@@ -223,6 +223,9 @@ use = egg:bodhi-server
 # A list of non-person users
 # system_users = bodhi autoqa taskotron
 
+# The max number of mash threads running at the same time
+# max_concurrent_mashes = 2
```
bowlofeggs: I think there's a "Mash settings" section (or something similarly named) where the pungi settings are - can we move this there? It's not a formal ini section, but there's a comment block to group them together.

bodhi/server/consumers/masher.py
```diff
@@ -256,14 +256,16 @@ def work(self, msg):
         last_key = None
         threads = []
         for batch in batches:
-            if last_key is not None and request_order_key(batch) != last_key:
+            if ((last_key is not None and request_order_key(batch) != last_key) or
+                    (len(threads) >= int(config.get('max_concurrent_mashes', 2)))):
```
bowlofeggs: Since we have the config validator on this setting, we don't need to convert it to an int or have a default here, so let's replace the above line with:

```
(len(threads) >= config['max_concurent_mashes'])
```

bodhi/server/consumers/masher.py
```diff
@@ -256,14 +256,16 @@ def work(self, msg):
...
                 # This means that after we submit all Stable+Security updates, we wait with kicking
                 # off the next series of mashes until that finishes.
                 self.log.info('All mashes for priority %s running, waiting', last_key)
```
bowlofeggs: This log statement might not be entirely accurate now. Maybe we can set it to say something like this instead?

```
self.log.info('Waiting on %d mashes of priority %s.', len(threads), last_key)
```

bowlofeggs (changes requested):
If we change that log statement, it becomes easier to write a test for this because we can launch a mash that would have two concurrent mashes with that setting set to 1 via mock, and then we can assert that the log statement says it's waiting on 1 mash.

bowlofeggs (changes requested):
Cool, the code looks great now. All we need is a test. I can write it for you if you don't have time.

bowlofeggs:
I am working on writing tests for this.

bowlofeggs:
@puiterwijk Do you approve of the commit I've added on top of yours?

puiterwijk:
Tests look good to me.

bowlofeggs:
jenkies test

bowlofeggs:
jenkies test

bowlofeggs (approved):

PR #1954, opened 2017-10-30
Instead of predicting the output dir, let Pungi tell us

puiterwijk:

### DIFF
--- a/bodhi/tests/server/consumers/test_masher.py
+++ b/bodhi/tests/server/consumers/test_masher.py
@@ -237,8 +237,17 @@ class TestMasher(base.BaseTestCase):
             with open(os.path.join(repodata, 'repomd.xml'), 'w') as repomd:
                 repomd.write(fake_repodata)
 
+            os.makedirs(os.path.join(mash_dir, 'compose', 'metadata'))
+            with open(os.path.join(mash_dir, 'compose', 'metadata', 'composeinfo.json'), 'w') as f:
+                f.write('{}')
+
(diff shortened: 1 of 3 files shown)


puiterwijk:
This needs test updates. I'll see about doing those later, as this is not a priority.

bowlofeggs:

bodhi/tests/server/consumers/test_masher.py
```diff
@@ -554,6 +563,36 @@ def test_testing_digest(self, mail, *args):
...
+                fake_popen.returncode = 0
+                t._startyear = datetime.datetime.utcnow().year
+                t._wait_for_pungi(fake_popen)
+                assert False, "Mash with invalid directory did not crash"
+            except Exception as ex:
+                expected_error = ('Directory at /tmp/nonsensical_directory does not look like a '
+                                  'compose')
+                expected_error = expected_error.format(datetime.datetime.utcnow().year)
+                assert str(ex) == expected_error
+            t.db = None
```
bowlofeggs, line 594: There's no problem leaving this line here, but you shouldn't need it because ```t``` will go out of scope once this test is done.

bowlofeggs (approved):
Before merging, please make one more change (which I don't feel the need to review, so merge without review): please add a note in ```docs/user/release_notes.rst``` about which minimum version of Pungi is needed to get this ```--print-output-dir``` flag. You can just drop a new section right at the top called "Dependency changes" or something like that. Thanks!

Issue #1922, opened 2017-10-23
Bodhi is not compatible with the latest release of Cornice


bowlofeggs:
Cornice 3.0.0 has been released, and Bodhi still has a requirement on Cornice < 2 in its requirements.txt. It would be nice if Bodhi could work with any version of Cornice, but it may be difficult to support Cornice 1, 2, and 3 with the same codebase. If that turns out to be difficult, we should rework Bodhi to use Cornice 3. This will introduce CI issues since only Rawhide would have Cornice 3. To work around that, we should allow pip installing Cornice for CI on Fedora versions less than Rawhide. For production, we can carry a newer Cornice package in the Fedora infrastructure repository.

bowlofeggs:
This makes Cornice packaging awkward in Rawhide https://bugzilla.redhat.com/show_bug.cgi?id=1430178

bowlofeggs:
I'm using this document to aid in the transition:

https://cornice.readthedocs.io/en/latest/upgrading.html#x-to-3-x

I'm starting to believe it will not be feasible to support Cornice 1, 2, and 3 in the same Bodhi release.

bowlofeggs:
Bodhi now works with ```Cornice~=2```. Now it is time to make it work with ```Cornice~=3```.

bowlofeggs:
The patch for this will be included in the upcoming 3.2.0 release:

https://github.com/fedora-infra/bodhi/pull/2092

PR #1958, opened 2017-10-30
Add support for Cornice 2, dropping support for Cornice 1.

bowlofeggs:
Cornice 2 requires pyramid >= 1.7 and Fedora 25 only has
pyramid-1.5, so this commit also adjusts the Vagrantfile.example
to use Fedora 26 by default instead of 25.

re #1922

### DIFF
--- a/bodhi/server/validators.py
+++ b/bodhi/server/validators.py
@@ -14,6 +14,7 @@
 
 from datetime import datetime, timedelta
 
+from cornice.validators import _colander
 from pyramid.exceptions import HTTPNotFound, HTTPBadRequest
 from pyramid.httpexceptions import HTTPFound, HTTPNotImplemented
 from sqlalchemy.sql import or_, and_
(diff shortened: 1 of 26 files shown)


bowlofeggs:
jenkies test

bowlofeggs:
jenkies test

bowlofeggs:
jenkies test

bowlofeggs:
jenkies test

bowlofeggs:
jenkies test

bowlofeggs:
jenkies test

bodhi/server/validators.py
```diff
@@ -14,6 +14,7 @@
 
 from datetime import datetime, timedelta
 
+from cornice.validators import _colander
```
jeremycline, line 17: Oof. Private APIs :(

devel/ansible/roles/dev/tasks/main.yml
```diff
@@ -25,63 +25,71 @@
...
+      - python2-pyramid-mako
+      - python2-pyramid-tm
       - python2-pytest-cov
+      - python2-rpdb
       - python2-sphinx
+      - python2-sqlalchemy
       - python2-sqlalchemy_schemadisplay
       - python2-waitress
       - python2-yaml
+      - python2-zmq
```
jeremycline: Out of curiosity, why do you need to directly depend on zmq? fedmsg should depend on it.

jeremycline (approved):
I'm not familiar with cornice or the differences between 1 and 2, but the change here looks okay.

bodhi/server/validators.py
bowlofeggs: @jeremycline Perhaps it will assuage your fears to know that I've only used this private API temporarily until there is an upstream release that has the ```_generate_colander_validator()``` that I wrote below.

I've pitched that code upstream and I have the PR approved, but it needs some tests:

https://github.com/Cornices/cornice/pull/465

So it's only a private API because I wanted to use the exact same thing that will eventually be in an upstream Cornice release.

bowlofeggs: Oh and actually, the public API that will ultimately be upstream will actually be the ```colander_querystring_validator()``` that is made by ```_generate_colander_validator()```. So once this is upstream, I will remove the code from this file and just use it directly from cornice.

devel/ansible/roles/dev/tasks/main.yml
bowlofeggs: I'm not sure why ```zmq``` is here, but I just tested the Vagrant box without it and it worked fine, so I'll drop it.

PR #1959, opened 2017-10-31
Remove python-pydns from dependencies

frenzymadness:
Python-pydns package can be removed from the list of dependencies and also from specfile. It is not used in Bodhi codebase.

It'll be still available in the testing environment because some other packages depend on it.
```
$ dnf repoquery --whatrequires python-pydns
python-fmn-web-0:0.8.1-2.fc25.noarch
python-pylibravatar-0:1.6-9.fc25.noarch
python-pymilter-0:1.0-4.fc25.x86_64
python-pyspf-0:2.0.11-5.fc25.noarch
python2-pyspf-0:2.0.12-1.fc25.noarch
```

### DIFF
--- a/devel/ansible/roles/dev/tasks/main.yml
+++ b/devel/ansible/roles/dev/tasks/main.yml
@@ -28,7 +28,6 @@
       - python-bugzilla
       - python-openid
       - python-progressbar
-      - python-pydns
       - python-pylibravatar
       - python-pyramid-fas-openid
       - python-simplemediawiki
--- a/requirements.txt
+++ b/requirements.txt
@@ -15,7 +15,7 @@ packagedb-cli
 # for captchas
 Pillow
 progressbar
-pyDNS
+pyDNS  # Should be required by pyLibravatar https://bugs.launchpad.net/pylibravatar/+bug/1661218
 pylibravatar
 pyramid~=1.7
 pyramid_fas_openid


frenzymadness:
CI build failed but it doesn't seem to be related to changes in this PR. Could anybody please check CI?

bowlofeggs:
Yeah I'm not sure what is going on with the container tests. I can run them locally without a problem, but they fail to dnf install things in the CentOS CI environment. I've confirmed that I have the same base layer locally as Jenkins is pulling.

bowlofeggs:
jenkies test

bowlofeggs:
It sounds like we need to have the f26 container base image rebuilt by the releng group. The glibc update combined with the dnf update seems to be causing this issue.

Also, I ran the container tests locally and found that removing this dependency does cause a problem:

```
pip     docs runtests: PYTHONHASHSEED='1791836502'         
pip     docs runtests: commands[0] | rm -rf _build images  
pip     docs runtests: commands[1] | mkdir -p _static images                                                           
pip     docs runtests: commands[2] | ./generate_db_schema  
pip     Traceback (most recent call last):                 
pip       File "/bodhi/docs/generate_db_schema", line 7, in <module>                                                   
pip         from bodhi.server import models                
pip       File "/bodhi/bodhi/server/models.py", line 44, in <module>                                                   
pip         from bodhi.server import bugs, buildsys, log, mail, notifications, Session                                 
pip       File "/bodhi/bodhi/server/mail.py", line 29, in <module>                                                     
pip         from bodhi.server.util import get_rpm_header   
pip       File "/bodhi/bodhi/server/util.py", line 39, in <module>                                                     
pip         import libravatar
pip       File "/usr/lib/python2.7/site-packages/libravatar.py", line 27, in <module>                                  
pip         import DNS       
pip     ImportError: No module named DNS                   
pip     ERROR: InvocationError: '/bodhi/docs/generate_db_schema'
```

My unverified guess is that the libravatar package upstream does not declare its dependency on PyDNS in its own setup.py and so it doesn't get pip installed with it? In any case, this PR does seem to cause the pip container tests to fail. Ideally we'd fix this upstream with pylibravatar, but we can pip install PyDNS in the pip container as a workaround until it's addressed upstream if we want.

frenzymadness:
Yes, you are right, thanks for investigation.

There is a bug in pylibravatar setup.py file and there is also a bug for this since February [0]. I've added a comment with a proposed fix and we will see.

Let this PR open and if the pylibravarat won't be fixed in some time, we can implement mentioned workaround for pip container.

[0] https://bugs.launchpad.net/pylibravatar/+bug/1661218

bowlofeggs:
That sounds like a plan to me. How about if they haven't responded within ~10 days, let's just pip install PyDNS in the pip container?

frenzymadness:
Because there is no response from pyLibravatar developers, I decided to keep pyDNS in requirements.txt (pip based container) with comment and remove it from ansible playbook.

Tests are passing now.

bowlofeggs (approved):

bowlofeggs:

PR #1960, opened 2017-10-31
Do not dnf upgrade when building the CI containers.

bowlofeggs:
This commit has one pro, one speculative pro, and one con:

Pro: Bodhi's CI tests will run faster.

Speculative pro: Sometimes we get what seem like race conditions
in the Docker storage system. This *may* be due to how we have two
f26 containers trying to write the same layer (the pip container is
f26) at the same time due to the dnf upgrade.

Con: If Fedora updates to the base layer packages break Bodhi in
some way, we won't know until the next container base layer
rebuild, which happens every two weeks in Fedora.

### DIFF
--- a/devel/ci/Dockerfile-header
+++ b/devel/ci/Dockerfile-header
@@ -1,9 +1,6 @@
 FROM registry.fedoraproject.org/fedora:FEDORA_RELEASE
 MAINTAINER "Randy Barlow" <bowlofeggs@fedoraproject.org>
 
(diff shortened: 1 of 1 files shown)


pypingou:
Assuming this doesn't conflict with https://github.com/fedora-infra/bodhi/pull/1980 somehow, 👍 for me

linux-modder:
With same contingence +1 from me as well.

bowlofeggs:
I'm going to close this for now, as the speculative pro was acheived by running fewer container builds in parallel.

PR #1961, opened 2017-10-31
Prepare for the 3.1.0 release.

bowlofeggs:
This commit prepares for the 3.1.0 release by adding release notes,
trimming release notes for releases that are older than a year, and
setting the version strings to 3.1.0.


docs/conf.py
```diff
@@ -273,3 +273,7 @@
 
 # Example configuration for intersphinx: refer to the Python standard library.
 intersphinx_mapping = {'http://docs.python.org/': None}
+
+extlinks = {
```
jeremycline, line 277: Ooooh nice, I didn't know about this.

jeremycline (approved):
Looks good to me. Are you trimming the changelog because it's getting so long? Just curious.

:shipit:

bowlofeggs:
@jeremycline Yeah, just because it's fairly long. I figured "releases older than one year" was a good place to draw the line.

I considered an alternative of creating release notes files per X version, but I think I prefer just to trim them. They remain in git history for those curious about ancient history.

PR #1965, opened 2017-11-01
The popup_toggle view no longer asserts that the user exists.

bowlofeggs:
Since the popup_toggle() view asserts that the user is logged in,
it is redundant to also check that the user exists as the act of
logging in ensures that the user exists.


bowlofeggs:
jenkies test

pypingou:
Makes sense to me! :)

ryanlerch (approved):

ryanlerch:
Yeah, i wondered what this was when i wan in here a couple months ago :) sorry i didnt mention it!

PR #1967, opened 2017-11-01
Remove an unreachable RSS type check.

bowlofeggs:
The RSS handler has a series of if statements to check which type
of thing is being rendered. If none of the things match, it raised
a 404 error stating that the item cannot be rendered. This check
was not reachable by end users, since the decorator must be present
on views that support RSS for the code to be hit, and no views have
the decorator that aren't supported. It would be an error (and not
a 404) if a view had the RSS decorator without support in this
function, so it would truly be better to blow up than to hand back
a false 404. Thus, this commit removes that unreachable check.

### DIFF
--- a/bodhi/server/renderers.py
+++ b/bodhi/server/renderers.py
@@ -2,7 +2,6 @@ import io
 import operator
 
 from pytz import utc
-from pyramid.exceptions import HTTPNotFound
 from feedgen.feed import FeedGenerator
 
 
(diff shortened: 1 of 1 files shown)


bowlofeggs:
jenkies test

pypingou:
looks fine :)

ryanlerch (approved):
LGTM also! 

Thanks for removing this! I poked around in here a while ago trying to write a test for this exception. Now i know why i couldnt hit it :D

PR #1968, opened 2017-11-01
Tests bodhi.server.services.overrides

bowlofeggs:


pypingou:
👍 for me

ryanlerch (approved):
Perfect! yay for another 100-percented file!!!

PR #1975, opened 2017-11-03
PEP-257 bodhi.server.metadata.

bowlofeggs:

### DIFF
--- a/bodhi/server/metadata.py
+++ b/bodhi/server/metadata.py
@@ -1,3 +1,8 @@
+# -*- coding: utf-8 -*-
+# Copyright 2007-2017 Red Hat, Inc. and others.
+#
+# This file is part of Bodhi.
+#
 # This program is free software; you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
 # the Free Software Foundation; either version 2 of the License, or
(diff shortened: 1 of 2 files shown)


bodhi/server/metadata.py
```diff
@@ -69,14 +75,26 @@ def modifyrepo(comp_type, compose_path, filetype, extension, source):
...
     def __init__(self, release, request, db, mashdir, close_shelf=True):
+        """
+        Initialize the UpdateInfoMetadata object.
+
+        Args:
+            release (bodhi.server.models.Release): The Release that is being mashed.
+            request (bodhi.server.models.UpdateRequest): The Request that is being mashed.
+            db (): A database session to be used for queries.
+            mashdir (basestring): A path to the mashdir.
+            close_shelf (bool): Whether to close the shelve, which is used to cache updateinfo
```
stickster, line 95: shelf? And a closing full stop.

stickster (changes requested):

bodhi/server/metadata.py
bowlofeggs: @stickster The spelling discrepancy is actually a mistake in the argument's name. In Python, the object is actually called a "[shelve](https://docs.python.org/3/library/shelve.html)". I thought it would be better to use the Python spelling here even though the argument is not spelled consistently with the Python class name. We could write a PR to change the argument name, but I'd prefer to do that separately since this PR is just about documenting the module.

The full stop is on the next line (the text was too wide to fit on one line so I continued with indentation on the next line.)

stickster: Ah, I get it. Thanks for explaining. And not sure how I missed the next line, sorry.

stickster (approved):

PR #1976, opened 2017-11-04
PEP-257 bodhi.server.config.

bowlofeggs:

### DIFF
--- a/bodhi/server/config.py
+++ b/bodhi/server/config.py
@@ -1,3 +1,8 @@
+# -*- coding: utf-8 -*-
+# Copyright © 2013-2017 Red Hat, Inc. and others.
+#
+# This file is part of Bodhi.
+#
 # This program is free software; you can redistribute it and/or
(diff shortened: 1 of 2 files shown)


bodhi/server/config.py
```diff
@@ -563,21 +586,25 @@ class BodhiConfig(dict):
     }
 
     def __getitem__(self, *args, **kw):
+        """Ensure the config is loaded, and then call the superclass __getiten__."""
```
stickster: getitem?

stickster (changes requested):
Just a simple typo fix

bodhi/server/config.py
bowlofeggs: @stickster I've fixed the typo in a new commit I've pushed - look good to you?

bowlofeggs:
jenkies test

stickster (approved):

PR #1978, opened 2017-11-07
Mock open so the CLI tests pass when the dev is logged in.

bowlofeggs:

### DIFF
--- a/bodhi/tests/client/test___init__.py
+++ b/bodhi/tests/client/test___init__.py
@@ -1,5 +1,8 @@
 # -*- coding: utf-8 -*-
-
+# Copyright © 2016-2017 Red Hat, Inc. and others.
+#
+# This file is part of Bodhi.
+#
 # This program is free software; you can redistribute it and/or
 # modify it under the terms of the GNU General Public License
 # as published by the Free Software Foundation; either version 2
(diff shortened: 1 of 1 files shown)


bodhi/tests/client/test___init__.py
```diff
@@ -501,17 +508,27 @@ def test_url_flag(self, send_request):
...
         mock_raw_input.return_value = 'dudemcpants'
-        runner = testing.CliRunner()
+        real_open = open
 
-        runner.invoke(client.query_buildroot_overrides, ['--mine'])
+        def fake_open(*args, **kwargs):
+            """Fake open so that it looks like we have no session cache."""
+            if args[0] == fedora.client.openidbaseclient.b_SESSION_FILE:
+                return mock.mock_open(read_data='{}')(*args, **kwargs)
+            return real_open(*args, **kwargs)
```
pypingou: Do you want to see about re-using the function defined above?

pypingou:
Looks fine to me :)

bodhi/tests/client/test___init__.py
bowlofeggs: @pypingou Sure, I'm pushing a different commit that shares the function. Thanks for the suggestion!

PR #1979, opened 2017-11-07
Do not allow pending updates to get request: batched.

### LINKED ISSUE #1930: batched regression: updates are not present in any repo if karma is supplied quick enough
Before batched: If karma was supplied quick enough, the package went stable immediately.
With batched: If karma is supplied quick enough, the package goes to batched, and waits for 2 weeks before going stable. During this time, it's not available in any repository. Instead, it should be pushed to updates-testing repo for the duration of the batched tag.

Example:
https://bodhi.fedoraproject.org/updates/FEDORA-2017-6db5c0829b

bowlofeggs:
Previously, if pending updates received enough karma they would get
their request set to batched. This was bad because the update would
then not be mashed into the testing repository and would sit in
Koji only until the batched updates were set to stable, which means
that further testing oppotunities were missed.

Four different ideas were considered for fixing this bug:

1. We could make bodhi-push and the masher also select
   pending:batched updates when mashing testing updates. This
   seemed like the natural thing to do, but it might actually be
   the trickiest solution because so many if statements in the code
   need to be touched to say something like "or update.status is
   pending and update.request is batched".
2. The state:pending should always imply request:testing, and so we
   could adjust the code that looks to mash request:testing mash
   state:pending instead, which would pick up these updates too.
   However, this solution suffers from a similar problem to the
   first solution - a whole lot of places in the code would need to
   be touched and so it's not a simple fix. (The more places we
   have to touch to fix this, the higher chance we have of
   introducing new bugs ☺).
3. We could make it so pending updates don't get request:batched
   with karma, but stay at request:testing. Then after the mash
   runs, it checks the mashed updates to see if any of them are
   eligible to become batched and go ahead and do it. This seems
   simpler than #1 and #2, except for non-autokarma updates - we
   need to make sure that developers can't manually mark their
   updates to be batched when they are state:pending or those will
   end up in this same situation. That last caveat makes this
   solution a little less simple than it would otherwise be.
4. Refactor the state machine so that it is expressed in one place
   and is therefore much easier to modify and verify. This is a
   large undertaking.

After some deliberation solution #3 was chosen, though it was
tempting to go with #4.

fixes #1930

### DIFF
--- a/bodhi/server/models.py
+++ b/bodhi/server/models.py
@@ -1781,6 +1781,12 @@ class Update(Base):
                         else:
                             action = UpdateRequest.testing
 
+        if action is UpdateRequest.batched and self.status is not UpdateStatus.testing:
+            # We don't want to allow updates to go to batched if they haven't been mashed into the
+            # testing repository yet.
+            raise BodhiException('This update is not in the testing repository yet. It cannot be '
+                                 'requested for batching until it is in testing.')
+
         # Add the appropriate 'pending' koji tag to this update, so tools like
         # AutoQA can mash repositories of them for testing.
         if action is UpdateRequest.testing:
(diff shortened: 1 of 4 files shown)


bowlofeggs:
jenkies test

bowlofeggs:
jenkies test

puiterwijk (approved):

Issue #1945, opened 2017-10-26
There are several validators for username that are essentially copied code


bowlofeggs:
There are several validators that make sure that various kwargs reference existing username. For example, check out [```validate_username()```](https://github.com/fedora-infra/bodhi/blob/3.0.0/bodhi/server/validators.py#L619-L632), [```validate_update_owner()```](https://github.com/fedora-infra/bodhi/blob/3.0.0/bodhi/server/validators.py#L648-L661), and [```validate_ignore_user()```](https://github.com/fedora-infra/bodhi/blob/3.0.0/bodhi/server/validators.py#L664-L677). These validators are nearly identical to the point where it'd be easy to write a common method they all use, or a validator generator function.

There may be others, so check for that too.

Yadnyawalkya:
hey @bowlofeggs, do you like to share your thoughts on this? https://github.com/fedora-infra/bodhi/pull/1982

PR #1982, opened 2017-11-08
Issue #1945 fix: Minimized copied code usage by common function for v…

Yadnyawalkya:
This PR is a fix for [#1945](https://github.com/fedora-infra/bodhi/issues/1945) issue.

> There are several validators that make sure that various kwargs reference existing username. For example, check out validate_username(), validate_update_owner(), and validate_ignore_user(). These validators are nearly identical to the point where it'd be easy to write a common method they all use, or a validator generator function.

Copied code of validate_username(), validate_update_owner(), and validate_ignore_user() functions can be minimized with having common function for all. Please feel free to add comment if there are any validation functions are left or not covered in this PR. 

Peace out !✌

### DIFF
--- a/bodhi/server/validators.py
+++ b/bodhi/server/validators.py
@@ -777,7 +777,7 @@ def validate_update(request, **kwargs):
 
 def ensure_user_exists(param, request):
     """
-    Ensure the user referenced by param exists and if it does replace it with the User object
+    Ensure the user referenced by param exists and if it does replace it with the User object.
 
     Args:
         param (string): Request parameter that references a username to be validated.
--- a/bodhi/server/validators.py
+++ b/bodhi/server/validators.py
@@ -756,28 +756,6 @@ def validate_bugs(request, **kwargs):
                                "Invalid bug ID specified: {}".format(bugs))
 
 
-def validate_username(request, **kwargs):
-    """
-    Make sure the referenced user exists.
-
-    Args:
-        request (pyramid.util.Request): The current request.
-        kwargs (dict): The kwargs of the related service definition. Unused.
-    """
-    username = request.validated.get("user")
-    if username is None:
-        return
-
-    db = request.db
-    user = db.query(User).filter_by(name=username).first()
-
-    if user:
-        request.validated["user"] = user
-    else:
-        request.errors.add("querystring", "user",
-                           "Invalid user specified: {}".format(username))
-
-
 def validate_update(request, **kwargs):
     """
     Make sure the requested update exists.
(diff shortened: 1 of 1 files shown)


bodhi/server/validators.py
```diff
@@ -800,48 +778,60 @@ def validate_update(request, **kwargs):
         request.errors.status = HTTPNotFound.code
 
 
-def validate_update_owner(request, **kwargs):
+def common_validate(param, request):
```
bowlofeggs: Let's give this a more descriptive name, perhaps something like "```ensure_user_exists()```.

bodhi/server/validators.py
```diff
@@ -800,48 +778,60 @@ def validate_update(request, **kwargs):
...
     """
-    Make sure the referenced update owner is an existing user.
+    Common fuction to be called for validatation.
```
bowlofeggs: Let's specify here that this is specifically for ensuring the referenced user exists. Perhaps we could say "Ensure the user referenced by param exists, and if it does replace it with the User object."""

bodhi/server/validators.py
```diff
@@ -800,48 +778,60 @@ def validate_update(request, **kwargs):
...
 
     Args:
+        param (string): Variable to pass as parameter.
```
bowlofeggs: Let's make this a little more specific too - perhaps "Request parameter that references a username to be validated."""

bowlofeggs (changes requested):
Looks nice! Just a few documentation fixes and a function rename, and we'll be good to merge. Thanks!

bowlofeggs:
jenkies test

bowlofeggs:
Please also adjust your commit message to have this on its own line after the commit message body:

```
fixes #1945
```

That will cause GitHub to close #1945 automatically when this is merged.

bodhi/server/validators.py
Yadnyawalkya: Nice, make sense!

Yadnyawalkya (commented):
Thanks for review! :) 👍  
@bowlofeggs, please have a look at recent changes.
Let me know if anything required !

bowlofeggs (approved):
Looks good to me! The CI system will likely not pass this PR today because Fedora is moving its servers and the container registry was not available earlier today. If the tests don't pass I'll retry them again soon. Once they pass, we are good to merge. Thanks for the patch!

bowlofeggs:
jenkies test

bowlofeggs:
jenkies test

bodhi/server/validators.py
```diff
@@ -800,48 +778,60 @@ def validate_update(request, **kwargs):
         request.errors.status = HTTPNotFound.code
 
 
-def validate_update_owner(request, **kwargs):
+def ensure_user_exists(param, request):
     """
-    Make sure the referenced update owner is an existing user.
+    Ensure the user referenced by param exists and if it does replace it with the User object
```
bowlofeggs: The test failure wants you to end this sentence with a period:

```
bodhi/server/validators.py:807 in public function `ensure_user_exists`:
        D400: First line should end with a period (not 't')
```

You can manually run pydocstyle on the code to make sure it passes with tox, or by running this:

```
$ pydocstyle  --add-ignore D413 bodhi/server/validators.py
```

bowlofeggs (changes requested):

bowlofeggs:
@Yadnyawalkya do you still want to work on this?

Yadnyawalkya:
Hey @bowlofeggs , Thanks you very much for kind reminder, I turned off my notifications and forgot to update this changes! 

Just updated that, surely like to contribute more on this side! Best regards!:)

bowlofeggs:
jenkies test

bodhi/server/validators.py
Yadnyawalkya: @bowlofeggs  Added period.

bowlofeggs (approved):
Thanks!

bowlofeggs:
This patch has been included in the [```3.3.0b0``` build](https://bodhi.stg.fedoraproject.org/docs/user/release_notes.html), which is currently deployed to [staging](https://bodhi.stg.fedoraproject.org).

PR #1987, opened 2017-11-11
Store a single completed repo

puiterwijk:
This also fixes resuming with completed repository, since we no longer care about the entire repo path, so "self.id" is very likely not to be in there.

### DIFF
--- a/bodhi/tests/server/consumers/test_masher.py
+++ b/bodhi/tests/server/consumers/test_masher.py
@@ -449,7 +449,7 @@ class TestMasher(unittest.TestCase):
         with file(t.mash_lock) as f:
             state = json.load(f)
         try:
-            self.assertEquals(state, {u'updates': [u'bodhi-2.0-1.fc17'], u'completed_repos': []})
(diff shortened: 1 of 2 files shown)


bodhi/tests/server/consumers/test_masher.py
```diff
@@ -1201,7 +1201,7 @@ def test_absent_gating(self, publish, *args):
                 cwd=t.mash_dir, shell=False, stderr=-1,
                 stdin=mock.ANY,
                 stdout=mock.ANY)])
-        self.assertEquals(len(t.state['completed_repos']), 1)
+        self.assertIsNotNone(t.state['completed_repo'])
```
bowlofeggs, line 1204: Can we change these ```self.assertIsNotNone```'s to ```self.assertEqual```'s? That way we have stronger assertions and can make sure that they are the right value too.

bowlofeggs (changes requested):

bowlofeggs:
jenkies test

bowlofeggs:
This test failure appears to be real.

bowlofeggs:
I think I've incorporated this change into #2052. If you agree, shall we close this PR?

puiterwijk:
@bowlofeggs++

PR #1988, opened 2017-11-11
Checkpoint adding comments, so we do not send those twice

puiterwijk:


bowlofeggs (approved):
Thanks!

bowlofeggs:

PR #1990, opened 2017-11-11
Use a semaphore for max_concurrent_mashes

puiterwijk:

### DIFF
--- a/bodhi/server/consumers/masher.py
+++ b/bodhi/server/consumers/masher.py
@@ -182,6 +182,7 @@ class Masher(fedmsg.consumers.FedmsgConsumer):
         if not self.valid_signer:
             log.warn('No releng_fedmsg_certname defined'
                      'Cert validation disabled')
+        self.max_mashes_sem = threading.BoundedSemaphore(config.get('max_concurrent_mashes'))
         super(Masher, self).__init__(hub, *args, **kw)
         log.info('Bodhi masher listening on topic: %s' % self.topic)
 
@@ -258,8 +259,7 @@ class Masher(fedmsg.consumers.FedmsgConsumer):
         last_key = None
         threads = []
         for compose in self._get_composes(body):
-            if ((last_key is not None and request_order_key(compose) != last_key) or
-                    (len(threads) >= config.get('max_concurrent_mashes'))):
+            if (last_key is not None and request_order_key(compose) != last_key):
                 # This means that after we submit all Stable+Security updates, we wait with kicking
                 # off the next series of mashes until that finishes.
                 self.log.info('Waiting on %d mashes for priority %s', len(threads), last_key)
@@ -278,7 +278,8 @@ class Masher(fedmsg.consumers.FedmsgConsumer):
                                compose['content_type'])
                 continue
 
-            thread = masher(compose, agent, self.log, self.db_factory, self.mash_dir, resume)
+            thread = masher(self.max_mashes_sem, compose, agent, self.log, self.db_factory,
+                            self.mash_dir, resume)
             threads.append(thread)
             thread.start()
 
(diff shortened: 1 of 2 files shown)


bodhi/server/consumers/masher.py
```diff
@@ -388,16 +389,22 @@ def __init__(self, release, request, updates, agent,
...
+        self.log.info('Acquired semaphore, starting')
         try:
             with self.db_factory() as session:
                 self.db = session
                 self.work()
                 self.db = None
         except Exception:
             self.log.exception('MasherThread failed. Transaction rolled back.')
+        self.max_concur_sem.release()
+        self.log.info('Released semaphore')
```
bowlofeggs: It might be good to put the above two lines in a ```finally``` block, even though the ```Exception``` handler currently catches all. That way it's a bit more future proof in case the ```Exception``` handler is altered. It can also guard against the call to ```log.exception()``` raising an ```Exception``` of its own.

bodhi/server/consumers/masher.py
```diff
@@ -350,7 +351,7 @@ class MasherThread(threading.Thread):
     ctype = None
     pungi_template_config_key = None
 
-    def __init__(self, release, request, updates, agent,
+    def __init__(self, max_concur_sem, release, request, updates, agent,
                  log, db_factory, mash_dir, resume=False):
```
bowlofeggs: Let's add the new parameter to the docblock below.

bowlofeggs (changes requested):
This looks really nice! I wrote some tests that try to ensure that no more than ```max_concurrent_mashes``` are run, so theoretically we don't need to write new tests for this change.

bowlofeggs:
There are quite a few test failures that we'll need to fix to merge this, but it looks like many of them just don't know about the new required parameter on the object.

bowlofeggs:
We may need to refactor this in light of #2052.

bodhi/server/consumers/masher.py
```diff
@@ -315,11 +316,13 @@ class ComposerThread(threading.Thread):
...
     ctype = None
 
-    def __init__(self, compose, agent, log, db_factory, mash_dir, resume=False):
+    def __init__(self, max_concur_sem, compose, agent, log, db_factory, mash_dir, resume=False):
         """
         Initialize the ComposerThread.
 
         Args:
+            max_concur_sem (threading.BoundedSemaphore): Semaphore making sure only a limited
+                number of ComposerThread's run at the same time.
```
bowlofeggs: This should be "ComposerThreads" since it is plural and not possessive.

bodhi/server/consumers/masher.py
```diff
@@ -856,7 +865,7 @@ class PungiComposerThread(ComposerThread):
 
     pungi_template_config_key = None
 
-    def __init__(self, compose, agent, log, db_factory, mash_dir, resume=False):
+    def __init__(self, max_concur_sem, compose, agent, log, db_factory, mash_dir, resume=False):
         """
         Initialize the ComposerThread.
```
bowlofeggs, line 870: We should also doc the new parameter below.

bodhi/tests/server/consumers/test_masher.py
```diff
@@ -175,11 +175,22 @@ def setUp(self):
...
         super(TestMasher, self).tearDown()
+
+        self.assert_sems(self.expected_sems)
+
         shutil.rmtree(self.tempdir)
         shutil.rmtree(self._new_mash_stage_dir)
 
+    def assert_sems(self, nr_expected):
+        assert self.semmock.acquire.call_count == nr_expected
+        assert self.semmock.release.call_count == self.semmock.acquire.call_count
```
bowlofeggs, line 192: Nice.

bowlofeggs (approved):
Just a couple of simple doc changes, so feel free to merge once they are fixed. No need for re-review.

PR #1991, opened 2017-11-13
Convert Bodhi to use and require Cornice>=3.

### LINKED ISSUE #1922: Bodhi is not compatible with the latest release of Cornice

bowlofeggs:
fixes #1922

### DIFF
--- a/bodhi/server/security.py
+++ b/bodhi/server/security.py
@@ -31,35 +31,51 @@ from .models import User, Group
 #
 # Pyramid ACL factories
 #
+class ACLFactory(object):
+    """Define an ACL factory base class to share the __init__()."""
 
-def admin_only_acl(request):
-    """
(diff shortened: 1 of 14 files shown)


pypingou:
Assuming this isn't a problem for the environment in which we're running bodhi, 👍 for me

PR #1992, opened 2017-11-13
Delay 16 seconds between each CI container build.

bowlofeggs:
Bodhi's CI tests have been experiencing a high rate of what seem
like race conditions that lead to the parallel container test
builds failing. This commit is an experiment to see if staggering
the container builds by 8 seconds each will help us to win the race
more often.


bowlofeggs:
jenkies test

puiterwijk (approved):
Let's hope this helps...

PR #2000, opened 2017-11-14
PEP-257 bodhi.server.models.

bowlofeggs:
Signed-off-by: Randy Barlow <randy@electronsweatshop.com>

PR #2001, opened 2017-11-14
Use autosummary to generate model and bindings docs.

bowlofeggs:
This pull request will combine nicely with #2000.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,6 +27,8 @@ devel/ci/Dockerfile-rawhide
 development.ini
 dist
 docs/_build
+docs/docblocks
+docs/developer/docblocks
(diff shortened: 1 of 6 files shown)


bowlofeggs:
Apparently this is a real failure, as it seems that on rawhide a warning is now treated as an error. This is OK though, as #2000 actually fixes that warning anyway, so I should be able to rebase this pull request once #2000 is merged and things should pass.

bowlofeggs:
jenkies test

bowlofeggs:
jenkies test

puiterwijk (approved):
Looks good to me

bowlofeggs:

PR #2010, opened 2017-11-15
Use the diff-cover package from Rawhide in CI.

bowlofeggs:


bowlofeggs:
jenkies test

puiterwijk (approved):

PR #2037, opened 2017-11-21
Remove an unneeded UTF-8 header from script.py.mako.

bowlofeggs:
re #1948


bowlofeggs:
jenkies test

pypingou (approved):

PR #2039, opened 2017-11-21
Add a uniqueness constraint on packages.{name,type}.

### LINKED ISSUE #2038: Bodhi is missing a uniqueness constraint on packages(name, type) due to a migration error
@tyll and I have noticed that Bodhi is missing a compound uniqueness constraint on the packages table for the ```name``` and ```type``` fields, which is apparent when you ask alembic to autogenerate a migration.

I looked into the history and I think I found out what happened. It seems that there was a migration added to create this field in f404e248872e6a99ef5dc703f496dd729780f3b2. However, the field's type was changed in cb8b2ab4d14c8ce2e269843162219811216990ef and when I did that I didn't noticed that I had effectively dropped the constraint when I deleted the type column.

Thus, we simply need a migration to add it back.

bowlofeggs:
fixes #2038

### DIFF
--- a/bodhi/server/migrations/versions/74cfbc6116ad_add_packages_type_constraint.py
+++ b/bodhi/server/migrations/versions/74cfbc6116ad_add_packages_type_constraint.py
@@ -1,0 +1,40 @@
+# Copyright (c) 2017 Red Hat, Inc.
+#
+# This file is part of Bodhi.
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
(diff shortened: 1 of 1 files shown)


pypingou:
No changes in the model itself?

Otherwise 👍 for me

bowlofeggs:
@pypingou I wrote more about it in #2038, but this migration is needed due to a model change that happened long ago. The original change actually did have a migration, but a later migration accidentally removed this so we need to add it back again. Oops!

PR #2042, opened 2017-11-22
Run only one container build at a time in CI.

bowlofeggs:
We've been trying to figure out a problem with the device mapper in
CentOS CI. We had previously tried to stagger the container builds
to see if it was a race condition and it did not work. This commit
instead tries to run only one docker build at a time.

### DIFF
--- a/devel/run_tests.sh
+++ b/devel/run_tests.sh
@@ -63,7 +63,7 @@ popd
 # tags.
 $PARALLEL sed -i "s/FEDORA_RELEASE/{= s:f:: =}/" devel/ci/Dockerfile-{} ::: $RELEASES
 # Build the containers.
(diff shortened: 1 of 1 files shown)


puiterwijk (approved):

bowlofeggs:
Looks like we hit the job timeout for this. It made it from f25 to Rawhide, so it actually wasn't too bad. I'll open a new PR to double the timeout.

bowlofeggs:
I opened https://github.com/fedora-infra/bodhi/pull/2043 to raise the timeout.

bowlofeggs:
jenkies test

bowlofeggs:
The CI is now very slow, but it did pass on the first try (well, the first one that allowed it this long). It took 83 minutes.

Issue #1514, opened 2017-05-12
Modify the updates fedmsg-client to check/set the CI status for new updates


pypingou:
We need a fedmsg-client listening to messages from bodhi about new updates and edits to updates and checks the CI status of the builds associated with this update and adjust the CI status accordingly

bowlofeggs:
There is an existing "updates" consumer that listens for new updates to do background processing - I think it makes sense to tack this feature onto it:

https://github.com/fedora-infra/bodhi/blob/2.6.2/bodhi/server/consumers/updates.py

bowlofeggs:
This is blocked on https://pagure.io/greenwave/issue/26.

ralphbean:
Are you unblocked now with https://pagure.io/greenwave/pull-request/27 ?

bowlofeggs:
@ralphbean Indeed!

bowlofeggs:
@pypingou You had assigned this to yourself, but I can probably do it if you don't still intend to. What say you?

pypingou:
@bowlofeggs go for it, thanks :)

bowlofeggs:
The patch for this feature has been included in the [```3.3.0b0``` build](https://bodhi.stg.fedoraproject.org/docs/user/release_notes.html), which is currently deployed to [staging](https://bodhi.stg.fedoraproject.org).

PR #2047, opened 2017-11-28
introduce a new endpoint for waiving test results on an update

matt8754:
A new endpoint `/updates/{id}/waive-test-results` for waiving test results. This is part of allowing packagers to be able to waive test results from the web UI.

### DIFF
--- a/bodhi/server/config.py
+++ b/bodhi/server/config.py
@@ -412,7 +412,13 @@ class BodhiConfig(dict):
             'value': '%s has been submitted as an update to %s. %s',
             'validator': six.text_type},
         'greenwave_api_url': {
-            'value': 'https://greenwave.fedoraproject.org/api/v1.0',
+            'value': 'https://greenwave-web-greenwave.app.os.fedoraproject.org/api/v1.0',
+            'validator': six.text_type},
+        'waiverdb_api_url': {
+            'value': 'https://waiverdb-web-waiverdb.app.os.fedoraproject.org/api/v1.0',
+            'validator': six.text_type},
+        'waiverdb.access_token': {
+            'value': None,
             'validator': six.text_type},
         'koji_hub': {
             'value': 'https://koji.stg.fedoraproject.org/kojihub',
--- a/bodhi/server/models.py
+++ b/bodhi/server/models.py
@@ -48,7 +48,8 @@ from bodhi.server.config import config
 from bodhi.server.exceptions import BodhiException, LockedUpdateException
 from bodhi.server.util import (
     avatar as get_avatar, build_evr, flash_log, get_critpath_components,
-    get_nvr, get_rpm_header, header, packagename_from_nvr, tokenize, pagure_api_get)
+    get_nvr, get_rpm_header, header, packagename_from_nvr, tokenize, pagure_api_get,
+    greenwave_api_post, waiverdb_api_post)
 import bodhi.server.util
 
 
@@ -2350,6 +2351,51 @@ class Update(Base):
             action.description, username, notes), author=u'bodhi')
         topic = u'update.request.%s' % action
         notifications.publish(topic=topic, msg=dict(update=self, agent=username))
+
+    def waive_test_results(self, username, comment=None):
+        """
+        Attempt to waive test results for this update.
+
+        Args:
+            username (basestring): The name of the user who is waiving the test results.
+            comment (basestring): A comment from the user describing their decision.
+        Raises:
+            LockedUpdateException: If the Update is locked.
+            BodhiException: If test gating is not enabled in this Bodhi instance,
+                            or if the tests have passed.
+        """
+        log.debug('Attempting to waive test results for this update %s' % self.alias)
+
+        if self.locked:
+            raise LockedUpdateException("Can't waive test results on a "
+                                        "locked update")
+
+        if not config.get('test_gating.required'):
+            raise BodhiException('Test gating is not enabled')
+
+        if self.test_gating_passed:
+            raise BodhiException("Can't waive test resuts on an update that passes test gating")
+
+        decision_context = u'bodhi_update_push_testing'
+        if self.status == UpdateStatus.testing:
+            decision_context = u'bodhi_update_push_stable'
+        data = {
+            'product_version': self.product_version,
+            'decision_context': decision_context,
+            'subject': self.greenwave_subject
+        }
+        decision = greenwave_api_post('{}/decision'.format(config.get('greenwave_api_url')), data)
+        results = [dict(subject=req['item'], testcase=req['testcase']) for req in
+                   decision['unsatisfied_requirements']]
+        log.debug('Waiving test results: %s' % results)
+        data = {
+            'results': results,
+            'product_version': self.product_version,
+            'waived': True,
+            'proxy_user': username,
+            'comment': comment
+        }
+        waiverdb_api_post('{}/waivers/'.format(config.get('waiverdb_api_url')), data)
 
     def add_tag(self, tag):
         """
--- a/bodhi/server/services/updates.py
+++ b/bodhi/server/services/updates.py
@@ -85,6 +85,14 @@ update_request = Service(name='update_request', path='/updates/{id}/request',
                          factory=security.PackagerACLFactory,
                          cors_origins=bodhi.server.security.cors_origins_rw)
 
+update_waive_test_results = Service(
+    name='update_waive_test_results',
+    path='/updates/{id}/waive-test-results',
+    description='Waive test results that block transitioning the update to next state',
+    factory=security.PackagerACLFactory,
+    cors_origins=bodhi.server.security.cors_origins_rw
+)
+
 
 @update.get(accept=('application/json', 'text/json'), renderer='json',
             error_handler=bodhi.server.services.errors.json_handler)
@@ -541,3 +549,35 @@ def new_update(request):
     result['caveats'] = caveats
 
     return result
+
+
+@update_waive_test_results.post(schema=bodhi.server.schemas.WaiveTestResultsSchema,
+                                validators=(colander_body_validator,
+                                            validate_update_id,
+                                            validate_acls),
+                                permission='edit', renderer='json',
+                                error_handler=bodhi.server.services.errors.json_handler)
+def waive_test_results(request):
+    """
+    Waive test results on a given update when gating is on.
+
+    Args:
+        request (pyramid.request): The current request.
+    Returns:
+        dict: A dictionary mapping the key "update" to the update.
+    """
+    update = request.validated['update']
+    comment = request.validated.pop('comment', None)
+
+    try:
+        update.waive_test_results(request.user.name, comment)
+    except LockedUpdateException as e:
+        request.errors.add('body', 'request', str(e))
+    except BodhiException as e:
+        log.exception("Failed to waive the test results")
+        request.errors.add('body', 'request', str(e))
+    except Exception as e:
+        log.exception("Unhandled exception in waive_test_results")
+        request.errors.add('body', 'request', str(e))
+
+    return dict(update=update)
--- a/bodhi/server/schemas.py
+++ b/bodhi/server/schemas.py
@@ -790,3 +790,12 @@ class SaveOverrideSchema(CSRFProtectedSchema, colander.MappingSchema):
         colander.String(),
         missing=None,
     )
+
+
+class WaiveTestResultsSchema(CSRFProtectedSchema, colander.MappingSchema):
+    """An API schema for bodhi.server.services.updates.waive_test_results()."""
+
+    comment = colander.SchemaNode(
+        colander.String(),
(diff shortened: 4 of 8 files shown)


pypingou:
I thought waivers were for rel-eng, so this PR makes me curious.

What's the point of enforcing tests if anyone can just waive them?

ralphbean:
/cc @AdamWill, @mohanboddu - any opinion here?

Although greenwave doesn't support it yet, we were figuring we would allow anyone to submit a waiver for a test result, but that greenwave could optionally specify that "in the context of policy XXX, waivers of failures will be considered valid no matter who submitted them, but in the context of policy YYY, waivers of failures will only be considered if submitted by a member of groups AAA or BBB"  (i.e., releng, qa, or provenpackager).

What the *actual* policy should be for gating updates is another story.

Regardless, right now greenwave doesn't take into consideration *who* submitted the waiver.  A waiver from anyone will do.

pypingou:
Do we have an actual use-case or are we building this "in case" ?

I really think it's a neat idea but the more moving parts the more changes of breakages so, wondering if this is backed up by an use-case :)

ralphbean:
Sure.  I was figuring that we let all packagers waive their own failures in the beginning, so that people can self service their way out of a snag.  As we identify things we decide should be more tightly controlled, we add the auth/group checks.

It's all about trying to ease in CI without introducing show-stoppers.

pypingou:
That makes sense, but thinking more about this yesterday I got wondering: does it make sense to allow waiving test results that do not matter anyway?

As a concrete example: does it make sense to allow waiving rpmlint errors while rpmlint isn't being considered for gating?

Thought reading the code, I wonder: does the current implement waive all results at once? Though a little radical that would surely avoid the issue I'm describing above :)

bowlofeggs:
Shouldn't there be a separate UI than Bodhi for waiving test results? In the end, we want pagure to gate PRs based on test results. If we add the waiving UI to Bodhi, we will have to also add it to Pagure. If we instead add it to WaiverDB then we don't have to do it twice.

What do you think?

ralphbean:
There's logic there.  :)

I was thinking we would add it to bodhi, and add it to pagure, but not give waiverdb its own UI (with the thought that less UIs is more.  It is one less place to have to remember where it is, what its called, etc.).

For function though, we have a waiverdb-cli tool people can use to waive things directly (if they want to get into it).

bowlofeggs:
Another thought I had about Pagure and Bodhi wanting some similar functionality is that there might be some possibility for Pagure and Bodhi to share some library that does this for us. The library could have themeable HTML/CSS/JS code (so that Bodhi and Pagure can each maintain their independent appearances) to retrieve and display the test results. For waiving, I guess that's trickier to do with JavaScript, but maybe we could make the shared JS lib configurable enough that it would know how to ask whichever system it is used on to waive (i.e., maybe Pagure's waive API is subtly different than Bodhi's, but we could just write a JS function to have to the lib that knows how to do it for each system).

Anyways, we can consider all of the above an idea for the future, no need to hold off this immediate work on it, but I'd like to hear feedback from you all on it.

bowlofeggs:
P.S. I'm bad at JS/HTML/CSS.

bowlofeggs:
P.S.

MongoDB

bowlofeggs:
P.S. I propose that we name the shared library ```neck.js```.

bodhi/server/models.py
```diff
@@ -1817,6 +1818,45 @@ def set_request(self, db, action, username):
         topic = u'update.request.%s' % action
         notifications.publish(topic=topic, msg=dict(update=self, agent=username))
 
+    def waive_test_results(self, username, comment=None):
+        """ Attempt to waive test results for this update """
```
bowlofeggs: We need to document the two parameters, their types, and the conditions under which this method can raise the exceptions. Something like this:

```
"""
Attempt to waive test results for this update.

Args:
    username (basestring): The name of the user who is waiving the test results.
    comment (basestring): A comment from the user describing their decision.
Raises:
    LockedUpdateException: If the Update is locked.
    BodhiException: If test gating is not enabled in this Bodhi instance, or if the tests have passed.
"""
```

bodhi/server/models.py
```diff
@@ -1817,6 +1818,45 @@ def set_request(self, db, action, username):
...
+
+        decision_context = u'bodhi_update_push_testing'
+        if self.status == UpdateStatus.testing:
+            decision_context = u'bodhi_update_push_stable'
+        data = {
+            'product_version': self.product_version,
+            'decision_context': decision_context,
+            'subject': self.greenwave_subject
+        }
+        api_url = '{}/decision'.format(config.get('greenwave_api_url').rstrip('/'))
```
bowlofeggs: Instead of putting this ```rstrip()``` here, let's do that in ```bodhi.server.config```. If you look there, there are facilities to validate every setting in Bodhi, which is designed for things like this. It's useful to help you not need to remember to do this everywhere (i.e., the DRY principle) and it also makes the code cleaner because you can assume the setting is already formatted and typed the way you need it to be.

bodhi/server/models.py
```diff
@@ -1817,6 +1818,45 @@ def set_request(self, db, action, username):
...
+        decision = greenwave_api_post(api_url, data)
+        for req in decision['unsatisfied_requirements']:
+            # TODO: create waivers for missing results once this https://pagure.io/waiverdb/issue/80
+            # is resolved.
```
bowlofeggs: Let's not put TODO's in the code. If we need to do this later, let's just file a new issue so we remember to do it.

bodhi/server/models.py
```diff
@@ -1817,6 +1818,45 @@ def set_request(self, db, action, username):
...
+            if req['type'] == 'test-result-failed':
+                log.debug('Waiving test result %s' % req['result_id'])
+                wapi_url = '{}/waivers/'.format(config.get('waiverdb_api_url').rstrip('/'))
```
bowlofeggs: Let's also move this ```rstrip()``` over to ```config.py```.

bodhi/server/models.py
```diff
@@ -1817,6 +1818,45 @@ def set_request(self, db, action, username):
...
+                wdata = {
+                    'result_id': req['result_id'],
+                    'product_version': self.product_version,
+                    'waived': True,
+                    'proxy_user': username,
+                    'comment': comment
+                }
+                waiverdb_api_post(wapi_url, wdata)
```
bowlofeggs: Since we've had performance issues with Greenwave and PDC, I'm a bit sensitive to the possibility of proxy server timeouts with this. Has there been load testing on waiverdb to ensure it can handle this load? i.e., what if there are 120 test failures and someone clicks waive? If it is slower than 0.5 s per POST, Fedora's proxy server will hang up on this request.

I have two ideas we could consider to help here:

1. Is it possible to ask WaiverDB in one post to waive n test results? If so, that would be pretty straightforward to do here.
2. We could do this as a background task instead of as a synchronous API call. I've been kind of wanting to set up Celery for Bodhi, which would be ideal for this, but even without that we can use a fedmsg listener like we do for the update create/edit events. That way these waivers can happen in the background and the httpd server won't timeout.

If you have more/better ideas, please share!

bodhi/server/services/updates.py
```diff
@@ -86,6 +86,14 @@
                          acl=bodhi.server.security.packagers_allowed_acl,
                          cors_origins=bodhi.server.security.cors_origins_rw)
 
+update_waive_test_results = Service(
+    name='update_waive_test_results',
+    path='/updates/{id}/waive-test-results',
+    description='Waive test results that block transitioning the update to next state',
+    acl=bodhi.server.security.packagers_allowed_acl,
```
bowlofeggs: I've just merged #1991 which does not allow this style of acl declaration anymore. Instead, you can replace this line with this:

```
factory=security.PackagerACLFactory,
```

bodhi/server/services/updates.py
```diff
@@ -541,3 +549,36 @@ def new_update(request):
     result['caveats'] = caveats
 
     return result
+
+
+@update_waive_test_results.post(validators=(validate_update_id,
+                                            validate_acls,
+                                            validate_request),
+                                permission='edit', renderer='json',
+                                error_handler=bodhi.server.services.errors.json_handler)
```
bowlofeggs: We should also set up a schema for this API, similar to what the other API endpoints in this file have. Without that, we won't get useful autogenerated API docs.

bodhi/server/services/updates.py
```diff
@@ -541,3 +549,36 @@ def new_update(request):
     result['caveats'] = caveats
 
     return result
+
+
+@update_waive_test_results.post(validators=(validate_update_id,
```
bowlofeggs: Also related to #1991, you will want to add ```colander_querystring_validator``` and ```colander_body_validator``` to the beginning of this tuple, or you won't have a validated schema.

bodhi/server/services/updates.py
```diff
@@ -541,3 +549,36 @@ def new_update(request):
...
+        request (pyramid.request): The current request.
+    Returns:
+        dict: A dictionary mapping the key "update" to the update.
+    """
+    update = request.validated['update']
+    comment = request.validated.pop('comment', None)
+
+    if update.locked:
+        request.errors.add('body', 'request',
+                           "Can't waive test results of a locked update")
```
bowlofeggs: Since ```update.waive_test_results()``` below will also check for this condition, why don't we drop the check here and just catch the ```LockedUpdateException``` exception below?

bodhi/server/services/updates.py
```diff
@@ -541,3 +549,36 @@ def new_update(request):
...
+
+    try:
+        update.waive_test_results(request.user.name, comment)
```
bowlofeggs: So here we can add:

```
except LockedUpdateException:
    request.errors.add('body', 'request', "Cannot waive test results on a locked update.")
```

bowlofeggs (changes requested):
I really like the thorough testing, nice work!

The main concerns I have are the performance question I raised, and also the integration with Cornice 3 that I mention in a couple places.

AdamWill:
Sorry for the belated reply to @ralphbean 's question; I think his suggestion looks reasonable, not having read *all* the follow-up discussion. I certainly think the general model ralph suggested is the sensible one *in principle*: handle waivers submitted by anyone, but enable the use of policies to decide which waivers will be considered/accepted.

matt8754:
> Another thought I had about Pagure and Bodhi wanting some similar functionality is that there might be some possibility for Pagure and Bodhi to share some library that does this for us. The library could have themeable HTML/CSS/JS code (so that Bodhi and Pagure can each maintain their independent appearances) to retrieve and display the test results. For waiving, I guess that's trickier to do with JavaScript, but maybe we could make the shared JS lib configurable enough that it would know how to ask whichever system it is used on to waive (i.e., maybe Pagure's waive API is subtly different than Bodhi's, but we could just write a JS function to have to the lib that knows how to do it for each system).

Yeah. I would like to see how the test results would be shown in pagure.

pypingou:
> Yeah. I would like to see how the test results would be shown in pagure.

Most likely in the same way as we do now: https://src.fedoraproject.org/rpms/python2-ipython/pull-request/1

bodhi/server/models.py
bowlofeggs: I need to do this as part of #1514, so you don't need to worry about how to get ```bodhi.server.config``` to do this - just remove the ```.rstrip()``` here and I'll take care of the rest for you.

Here's a WIP commit that shows what I'm thinking:

https://github.com/bowlofeggs/bodhi/commit/5d890925a5df01dc36660a1f3ae66e32a61d336e

matt8754:
Rebased to address the feedback. Ready for another look.

bodhi/server/config.py
```diff
@@ -412,7 +412,13 @@ class BodhiConfig(dict):
             'value': '%s has been submitted as an update to %s. %s',
             'validator': six.text_type},
         'greenwave_api_url': {
-            'value': 'https://greenwave.fedoraproject.org/api/v1.0',
+            'value': 'https://greenwave-web-greenwave.app.os.fedoraproject.org/api/v1.0',
```
bowlofeggs, line 415: Normally I wouldn't want to change a default setting without raising the X version of the release, but in this case it does appear that the default does not exist:

```
$ host greenwave.fedoraproject.org
Host greenwave.fedoraproject.org not found: 3(NXDOMAIN)
```

bowlofeggs (approved):

bowlofeggs:
Thanks!

bowlofeggs:
I filed https://github.com/fedora-infra/bodhi/pull/2098 to document the change to the default config.

bowlofeggs:
The patch for this feature has been included in the [```3.3.0b0``` build](https://bodhi.stg.fedoraproject.org/docs/user/release_notes.html), which is currently deployed to [staging](https://bodhi.stg.fedoraproject.org).

PR #2052, opened 2017-12-06
Define a Compose model to replace MASHER lockfiles.

### LINKED ISSUE #717: When an update is locked, feedback appears to be ignored but is apparently cached, which leads to multiple submissions from the same user
It appears when an update is locked, the form will allow users to submit feedback, however, the submit button just "spins" at the bottom, and after a while returns to the static "submit" button without any sign of anything happening. There's no indication the feedback was recorded/accepted. This causes users (myself included, until I figured this out just now) to try and submit feedback multiple times, since the functionality appears to be silently failing. Then when the update unlocks, it accepts all the form submissions, so the same user's comment is recorded however many times they tried to submit. It's smart enough not to count karma more than once from the same submitter, but not smart enough to pop up a warning saying "the package is locked, receipt of feedback will be delayed" or identify multiple form submissions with identical content.

### LINKED ISSUE #1245: bodhi-push should sort the updates by release
bodhi-push used to collate the list of packages it prints by release, and now it prints them in whatever order the database gives us. This makes it more difficult for the release engineer to review the list to approve or deny.

Instead, we could collate the printed list by release, and additionally it could be helpful to collate them by update (for updates with multiple packages).

### LINKED ISSUE #2014: Create a push status model
Bodhi currently creates lock files when mashing. These are used to ensure that no two processes attempt to mash the same repository at the same time, but they are also used to store state about the current mash. Here's an example from today's ```MASHING-f27-updates``` lock file:

```
{'completed_repos': [],
 'determine_and_perform_tag_actions': True,
 'updates': ['lightdm-gtk-2.0.3-4.fc27',
             'java-9-openjdk-9.0.1.11-1.fc27',
             'icedtea-web-1.7-6.fc27',
             'kanatest-0.4.10-0.1.D20170810git19dd1a7d.fc27',
             'hexalate-1.1.1-1.fc27',
             'crossfire-client-1.72.0-1.fc27',
             'tuxmath-2.0.3-1.fc27',
             'python3-typed_ast-1.1.0-1.fc27',
             'tuxpaint-stamps-2014.08.23-1.fc27',
             'knot-resolver-1.5.0-1.fc27 knot-2.6.1-1.fc27',
             'python-stem-1.6.0-1.fc27',
             'tanglet-1.5.0-1.fc27',
             'simsu-1.3.4-1.fc27',
             'fwupd-1.0.1-3.fc27 gnome-software-3.26.2-1.fc27 libappstream-glib-0.7.4-1.fc27',
             'qt5-qtwebengine-5.9.2-2.fc27',
             'libsmbios-2.3.3-1.fc27',
             'mint-x-icons-1.4.5-1.fc27',
             'mint-y-icons-1.1.0-1.fc27',
             'pyotherside-1.5.3-2.fc27',
             'libx86emu-1.11-3.fc27 hwinfo-21.47-3.fc27']}
```

The use of these lock files limits us in a few ways. For one, it is difficult to make a good human interface to see the current mashing status (the [```/masher```](https://bodhi.fedoraproject.org/masher/) view has poor performance and is also difficult to interpret) since the lock files are only available on the backend mashing server. It is also difficult to make ```bodhi-push``` work remotely, since it tries to read these lock files.

We should refactor bodhi to store its mash state in a new ```PushStatus``` model in the database. It should be flexible enough to handle storing the current status of the mash just as the lock files do today.

This will enable further work, like creating a REST API to control mashing and ultimately web and CLI interfaces to view the current status of mashing.

bowlofeggs:
Due to
[alembic #122](https://bitbucket.org/zzzeek/alembic/issues/122/),
it was not simple to write the included database migration in
a database backend agnostic fashion. Thus, this commit also
documents that Bodhi formally requires PostgreSQL.

This commit also happens to fix an issue (#717) where users could
not comment on Updates while they were mashing. This was due to
the extremely long transactions that Bodhi used to hold open prior
to this commit, and due to Bodhi modifying the state of the Update
early on during the process. The masher now commits whenever the
Compose's state changes and no longer modifies the Updates until
the end of the process. This means that the Update does not get
locked for hours in the database any longer.

This commit also addresses an RFE (#1245) to get bodhi-push to
collate the updates by release. It goes further by also collating
them by content type and stable/testing.

fixes #717
fixes #1245
fixes #2014

### DIFF
--- a/bodhi/server/consumers/masher.py
+++ b/bodhi/server/consumers/masher.py
@@ -46,7 +46,7 @@ from bodhi.server import bugs, initialize_db, log, buildsys, notifications, mail
 from bodhi.server.config import config
 from bodhi.server.exceptions import BodhiException
 from bodhi.server.metadata import UpdateInfoMetadata
-from bodhi.server.models import (Update, UpdateRequest, UpdateType, Release,
+from bodhi.server.models import (Compose, ComposeState, Update, UpdateRequest, UpdateType, Release,
                                  UpdateStatus, ReleaseState, ContentType)
 from bodhi.server.util import sorted_updates, sanity_check_repodata, transactional_session_maker
 
@@ -64,13 +64,13 @@ def checkpoint(method):
 
     @functools.wraps(method)
     def wrapper(self, *args, **kwargs):
-        if not self.resume or not self.state.get(key):
+        if not self.resume or not self._checkpoints.get(key):
             # Call it
             retval = method(self, *args, **kwargs)
             if retval is not None:
                 raise ValueError("checkpointed functions may not return stuff")
             # if it didn't raise an exception, mark the checkpoint
-            self.state[key] = True
+            self._checkpoints[key] = True
             self.save_state()
         else:
             # cool!  we don't need to do anything, since we ran last time
@@ -80,7 +80,7 @@ def checkpoint(method):
     return wrapper
 
 
-def request_order_key(requestblob):
+def request_order_key(compose):
     """
     Generate a sort key for the updates documents in generate_batches.
 
@@ -97,9 +97,9 @@ def request_order_key(requestblob):
         Testing: 0
     """
     value = 0
-    if requestblob['has_security']:
+    if compose['security']:
         value -= 2
-    if requestblob['phase'] == 'stable':
+    if compose['request'] == UpdateRequest.stable.value:
         value -= 1
     return value
 
@@ -110,7 +110,7 @@ class Masher(fedmsg.consumers.FedmsgConsumer):
 
     A fedmsg consumer that listens for messages from releng members.
 
-    An updates "push" consists of::
+    An updates "compose" consists of::
 
     - Verify that the message was sent by someone in releng
     - Determine which updates to push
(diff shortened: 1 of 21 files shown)


bowlofeggs:
The Jenkies failures are due to the registry being down due to the Fedora infrastructure move that is in progress.

bowlofeggs:
jenkies test

bowlofeggs:
Looks like another PR I merged caused this test failure - I'll have to rebase the alembic migration in this PR to use a new down_revision.

bodhi/server/consumers/masher.py
```diff
@@ -588,65 +525,59 @@ def eject_from_mash(self, update, reason):
...
             force=True,
         )
 
     def init_state(self):
         """
-        Create the mash_dir if it doesn't exist, and generate the lock file path within it.
-
-        This does not create the lock file itself, only the path that will be used for the lock
-        file, which is stored on self.mash_lock.
+        Create the mash_dir if it doesn't exist, and update the Compose state.
```
puiterwijk: It doesn't update the compose state?

bodhi/server/consumers/masher.py
```diff
@@ -588,65 +525,59 @@ def eject_from_mash(self, update, reason):
...
 
         Raises:
             Execption: If a lock file already exists and we are not resuming.
```
puiterwijk: 1. This is a typo (old one)
2. I don't think this applies anymore

bodhi/server/models.py
```diff
@@ -692,6 +692,33 @@ class ReleaseState(DeclEnum):
...
+        requested (EnumSymbol): A compose has been requested, but it has not started yet.
+        pending (EnumSymbol): The request for the compose has been received by the backend worker,
+            but the compose has not started yet.
+        initializing (EnumSymbol): The compose is initializing.
+        updateinfo (EnumSymbol): The updateinfo.xml is being generated.
+        punging (EnumSymbol): A Pungi soldier has been deployed to deal with the situation.
+        notifying (EnumSymbol): Pungi has finished successfully, and we are now sending out various
+            forms of notifications, such as e-mail, fedmsgs, and bugzilla.
+        success (EnumSymbol): The Compose has completed successfully.
+        failed (EnumSymbol): The compose has failed, abandon hope.
```
puiterwijk, line 709: I love this description :smiley:

bodhi/server/models.py
```diff
@@ -1529,6 +1564,36 @@ def validate_builds(self, key, build):
...
+
+        Args:
+            key (str): The field's key, which is un-used in this validator.
+            release (Release): The release object which is being associated with this update.
+        Raises:
+            ValueError: If the release being associated is not the same content type as the
+                update.
+        """
+        if release and self.content_type is not None:
+            for u in release.updates:
```
puiterwijk: Do we really want to go over all updates?
I'd say that if one of them is there and of the same type, we can break.

bodhi/server/models.py
```diff
@@ -3253,6 +3302,204 @@ def __json__(self, request=None, anonymize=False):
...
+            target (Compose): The compose that has had a change to its state attribute.
+            value (EnumSymbol): The new value of the state.
+            old (EnumSymbol): The old value of the state
+            initiator (sqlalchemy.orm.attributes.Event): The event object that is initiating this
+                transition.
+        """
+        if value != old:
+            target.state_date = datetime.utcnow()
+
+    def __eq__(self, other):
```
puiterwijk: I am honestly not sure that this is a good idea.
Since this might confuse people a lot, since normally `==` would say whether two objects are the actual same compose.

puiterwijk (approved):
Some nitpicks, but :+1:

puiterwijk:
Also, it looks like you never actually set the `error_message` outside of the tests?

bodhi/server/consumers/masher.py
bowlofeggs: Ah yes. Fixed!

bodhi/server/consumers/masher.py
bowlofeggs: Fixed.

bodhi/server/models.py
bowlofeggs: ☺

bodhi/server/models.py
bowlofeggs: Yeah I had briefly wondered about this too when I was doing this. I think I agree. In the long term I plan to formalize the Release and give it a DB enforced type anyway. Fixed.

bodhi/server/models.py
bowlofeggs: I had thought this would be important for the sorting to work, but I now believe that I only need the ```__le__()``` to do that. I'll drop this.

PR #2055, opened 2017-12-07
Make builds.package_id non-nullable.

bowlofeggs:
While writing tests for the validators, I came across some untested
code that was checking to make sure that Builds had Pacages
associated with them. Since it would be an error for a Build to be
unassociated with a Package, I opted to delete that check and
instead make the foreign key non-nullable in the database.

### DIFF
--- a/bodhi/tests/server/test_models.py
+++ b/bodhi/tests/server/test_models.py
@@ -742,6 +742,15 @@ class TestBuild(ModelTest):
     klass = model.Build
     attrs = dict(nvr=u"TurboGears-1.0.8-3.fc11")
 
+    def do_get_dependencies(self):
+        """
+        A Build needs a package to be associated with.
+
+        Returns:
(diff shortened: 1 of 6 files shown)


bodhi/tests/server/test_models.py
```diff
@@ -742,6 +742,15 @@ class TestBuild(ModelTest):
     klass = model.Build
     attrs = dict(nvr=u"TurboGears-1.0.8-3.fc11")
 
+    def do_get_dependencies(self):
```
jeremycline, line 745: I recommend calling this ``_do_real_get_fetch_dependencies`` 😄 (I see you don't have a choice in the matter, though)

jeremycline (approved):
Looks good :shipit:

bodhi/tests/server/test_models.py
bowlofeggs: Hahah

bowlofeggs:

PR #2056, opened 2017-12-08
Make updates.release_id non-nullable.

bowlofeggs:
While writing tests for the validators, I came across an untested
block of code that was raising an error if the release on an update
was null. Rather than writing a test for that code, I decided to
make it impossible for updates to be unassociated with releases, as
much of the codebase assumes that updates have releases as is (and
it doesn't make sense to have updates floating around out there
without releases).


jeremycline (approved):
:shipit:

bowlofeggs:

PR #2057, opened 2017-12-08
Add test coverage for bodhi.server.validators.

bowlofeggs:


jeremycline (approved):
Is this the year of the 100% bodhi test coverage?

:shipit:

bowlofeggs:
I don't think so - I still have 5% left to make it. But I think 2018 will be the year!

PR #2060, opened 2017-12-11
Add documentation of Bodhi's update states.

bowlofeggs:
This commit also reorganizes the user focused documentation into a
new docs/user folder since there is now enough user documentation
for such a folder to be useful.


ryanlerch (approved):

bowlofeggs:
The patch for this feature has been included in the [```3.3.0b0``` build](https://bodhi.stg.fedoraproject.org/docs/user/release_notes.html), which is currently deployed to [staging](https://bodhi.stg.fedoraproject.org).

PR #2075, opened 2017-12-13
Add a container value to the content_type enum.

### LINKED ISSUE #2026: Create Enums for Containers
In order for Bodhi to support containers, we will need to add enums for them.

bowlofeggs:
fixes #2026

### DIFF
--- a/bodhi/server/migrations/versions/da710ff02641_add_container_enum.py
+++ b/bodhi/server/migrations/versions/da710ff02641_add_container_enum.py
@@ -1,0 +1,54 @@
+# Copyright (c) 2017 Red Hat, Inc.
+#
+# This file is part of Bodhi.
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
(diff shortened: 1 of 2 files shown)


bodhi/server/migrations/versions/da710ff02641_add_container_enum.py
```diff
@@ -0,0 +1,54 @@
...
+    op.execute("ALTER TYPE ck_content_type ADD VALUE 'container' AFTER 'module'")
+
+
+def downgrade():
+    """
+    Alert user that we cannot downgrade this migration.
+
+    PostgreSQL does not allow enum values to be removed.
+    """
+    raise NotImplementedError("Downgrading this migration is not supported.")
```
puiterwijk, line 54: Would it be an idea to print a warning, but not error out? Given the fact that having this extra value there should still work for older codebases.

puiterwijk (approved):

bodhi/server/migrations/versions/da710ff02641_add_container_enum.py
bowlofeggs: I think that would cause a problem when the user tries to re-upgrade, because then adding the Enum would fail. Thus they'd get stuck in the downgraded state.

PR #2083, opened 2017-12-14
Configure CI to run two builds in parallel.

bowlofeggs:
We had been having stability issues with Bodhi's CI, with what
seemed like a race condition. As an experiment we tried reducing
the number of parallel container builds to 1 to see if that helped.
It did seem to solve the issue, so now I would like to experiment
with setting it to two.


bowlofeggs:
This PR cuts down on the testing time by a good bit - it went from about 1.5 hours to about 50 minutes.

jeremycline (approved):
:shipit:

Issue #2015, opened 2017-11-16
Create a REST API to control composes


bowlofeggs:
In order to enable container rebuild automation, we will need to provide a way for container rebuilding to launch container mashes. If we are going to provide a REST API to launch mashes, it would be best to generalize it and make it suitable for the regular mashing to use it too.

The API should accept parameters that specify what to mash (use ```push.py``` as inspiration for designing the API), and should generate the same fedmsg that ```push.py``` creates today. Of course, it should limit the users who are allowed to use it.

Depends on #2014.

bowlofeggs:
@puiterwijk and I discussed what the API would look like today, and we came up with this:

GET /composes/
- List of compose data.


GET /composes/release/request/


POST /composes/release/request/
- Can be used to start the mash, or to resume.


POST /composes/
- Post a list of updates to create a set of composes, which are returned to you.
- Allow all the kinds of flags that push.py allows.
- The composes are created, but not started. They are in a new ComposeState to indicate this.

bowlofeggs:
It's important that only admins can use the write API. For an example of how to do this, take a look at the way that ```bodhi.server.services.releases.releases``` uses ```security.AdminACLFactory```.

PR #2091, opened 2017-12-18
Add a new /composes/ service with an html renderer.

### LINKED ISSUE #2022: Create a web interface that will display the status of mashing
The ```/masher``` interface is currently very difficult to interpret and takes a very long time to load (and often times out without loading). Once we have a PushStatus object as proposed in #2014, we can create a new mash status web UI that shows the current state of the mashing processes. We can even display which step each process is at and display whether a maintenance mode is currently requested. Make it fancy.

Depends on #2014.

bowlofeggs:
fixes #2022 

Here are a couple of screenshots of the new pages:

screenshot from 2017-12-18 17-55-55
screenshot from 2017-12-18 17-55-28

@ryanlerch, I'd love your feedback on this one.

### DIFF
--- a/bodhi/server/services/composes.py
+++ b/bodhi/server/services/composes.py
@@ -1,0 +1,85 @@
+# -*- coding: utf-8 -*-
+# Copyright © 2017 Red Hat, Inc.
+#
+# This file is part of Bodhi.
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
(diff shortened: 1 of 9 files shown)


puiterwijk (approved):
Looks good to me!
Maybe we would want to add a json renderer at some point, but for now this is fine.

bowlofeggs:
Yeah I'm planning to add the JSON renderer as part https://github.com/fedora-infra/bodhi/issues/2015. Thanks for the review!

PR #2092, opened 2017-12-19
Write release notes for 3.2.0.

bowlofeggs:


bowlofeggs:
The CI test failure is due to https://bugzilla.redhat.com/show_bug.cgi?id=1527666

bowlofeggs:
jenkies test

docs/user/release_notes.rst
```diff
@@ -1129,58 +1191,3 @@ The following contributors submitted patches for Bodhi 2.4.0:
 * andreas
 * Ankit Raj Ojha
 * Randy Barlow
-
-
-2.3.3
```
pypingou, line 1135: Removing this entire section was intended?

pypingou:
One question, otherwise 👍

docs/user/release_notes.rst
bowlofeggs: Yeah I want to trim the changelog after a year so it doesn't grow without bound. I do similar things in my RPM spec files.

pypingou: I like keeping it, even if it grows since it allows going back to look for when a feature was added or a bug fixed, but up to you :)

My 👍 stands

PR #2095, opened 2017-12-20
UI for waiving test results on an update

matt8754:
This introduces a UI change for waving the test results on an update, built upon #2047 
I will add some screenshots to show how the change looks like.

### DIFF
--- a/bodhi/server/templates/update.html
+++ b/bodhi/server/templates/update.html
@@ -397,6 +397,43 @@ $(document).ready(function() {
       });
     });
 
+% if self.util.can_waive_test_results(update):
+    // handle for waiving test restuls.
+    // set focus on the comment textarea when opening the modal.
+    $("#waive_test_results").on('shown.bs.modal', function () {
+        $('#waive_comment').trigger('focus')
+    })
+    $("#waive_test_results").submit(function(evt) {
+      evt.preventDefault();
+      var comment = $('#waive_test_results [name=waive_comment]').val();
+      var url = '${request.route_url("update_waive_test_results", id=update.alias)}';
+      var $this = $(this);
+      $.ajax({
+          url: url,
+          data: {
+            comment: comment,
+            csrf_token: "${request.session.get_csrf_token()}",
+          },
+          method: 'POST',
+          dataType: 'json',
+          success: function(response) {
+            // Just reload the page to refresh the test gating status.
+            location.reload();
+          },
+          error: function(response) {
+            $this.modal('hide');
+            $.each(response.responseJSON.errors, function(i, error) {
+                msg = messenger.post({
+                  message: error.description,
+                  type: "error",
+                  hideAfter: false,
+                  showCloseButton: true,
+                });
+            });
+          },
+      });
+    })
+% endif
 });
 </script>
 
@@ -497,7 +534,6 @@ $(document).ready(function(){
       % if update.critpath:
       <span data-toggle="tooltip" title="This update is in the critical path" class="label label-danger"> <i class="fa fa-fw fa-fire"></i></span>
       % endif
-
       % if can_edit:
         % if not update.locked:
           <div id='updatebuttons' class="btn-group pull-right" role="group" aria-label="...">
@@ -522,6 +558,9 @@ $(document).ready(function(){
               ${self.util.push_to_batched_or_stable_button(update) | n}
             % endif
             <a id='unpush' class="btn btn-sm btn-danger"><span class="fa fa-fw fa-arrow-circle-left"></span> Unpush</a>
+          % endif
+          % if self.util.can_waive_test_results(update):
+            <a id='waive' class="btn btn-sm btn-primary" data-toggle="modal" data-target="#waive_test_results">Waive Test Results</a>
           % endif
           </div>
         % else:
@@ -1040,4 +1079,30 @@ $(document).ready(function(){
 <div class="row">
 
 </div>
+
+% if self.util.can_waive_test_results(update):
+<div class="modal fade" id="waive_test_results" tabindex="-1" role="dialog" aria-labelledby="waiveModalLabel" aria-hidden="true">
+  <form>
+  <div class="modal-dialog" role="document">
+    <div class="modal-content">
+      <div class="modal-header">
+        <button type="button" class="close" data-dismiss="modal" aria-label="Close"><span aria-hidden="true">&times;</span></button>
+        <h3 class="modal-title" id="waiveModalLabel">Waive Test Results</h3>
+        </button>
+      </div>
+      <div class="modal-body">
+          <div class="form-group">
+            <textarea class="form-control" id="waive_comment" name="waive_comment"></textarea>
+          </div>
+      </div>
+      <div class="modal-footer">
+        <button type="button" class="btn btn-secondary" data-dismiss="modal">Cancel</button>
+        <button type="submit" class="btn btn-primary">Waive Test Results</button>
+      </div>
+    </div>
+    </form>
+  </div>
+</div>
+% endif
+
 </div> <!-- end container -->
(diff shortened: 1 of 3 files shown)


matt8754:
screenshot from 2017-12-20 10-33-41

matt8754:
screenshot from 2017-12-20 10-35-11

matt8754:
screenshot from 2017-12-20 10-38-15

matt8754:
Note: the error in the last screenshot is expected as we haven't resolved these two issues yet.
[1] https://pagure.io/waiverdb/issue/80
[2] https://pagure.io/waiverdb/issue/98

In theory, the page should be reloaded once the test results are successfully waived and the test gating status will be updated to "Passed". I will test this again once we have those two waiverdb issues resolved.

bowlofeggs:
The test failure is due to https://bugzilla.redhat.com/show_bug.cgi?id=1527666 and should go away whenever the next compose happens. I'll go ahead and ask it to run again just in case it's composed now.

bowlofeggs:
jenkies test

bodhi/server/templates/update.html
```diff
@@ -397,6 +397,40 @@
       });
     });
 
+    // handle for waiving test restuls.
+    // set focus on the comment textarea when openning the modal.
```
bowlofeggs: s/openning/opening/

bodhi/server/templates/update.html
```diff
@@ -508,6 +542,31 @@ <h1>
             % else:
               <a id='revoke' class="btn btn-sm btn-danger"><span class="fa fa-fw fa-arrow-circle-left"></span> Revoke</a>
             % endif
+            % if request.registry.settings.get('test_gating.required') and not update.test_gating_passed:
```
bowlofeggs: This is in the 'unpushed' side of a big if block, but don't forget that we may want to also be able to waive pushed updates too. I recommend pulling this out of the larger if block if feasible. One day I'm going to refactor the code here because I really find it messy (lots of repeated code).

bodhi/server/templates/update.html
```diff
@@ -508,6 +542,31 @@ <h1>
...
+              <a id='waive' class="btn btn-sm btn-primary" data-toggle="modal" data-target="#waive_test_results">Waive Test Results</a>
+              <div class="modal fade" id="waive_test_results" tabindex="-1" role="dialog" aria-labelledby="waiveModalLabel" aria-hidden="true">
```
bowlofeggs: Can this ```<div>``` be moved to be somewhere else in the code? It's a lot of code and it makes it a little hard to read the already complicated if/else block that it's inside.

bowlofeggs (changes requested):
Can you rebase this PR to pull the first commit out, since it is now merged to develop?

Just a few other suggestions. Hopefully we'll be able to get the tests passing again soon.

bodhi/server/templates/update.html
matt8754: Hmm, can you help me recall what the use case is for waiving the updates that have been pushed?

bodhi/server/templates/update.html
matt8754: Sure, I guess I could move it down to the bottom.

bodhi/server/templates/update.html
bowlofeggs: Hi @matt8754!

So "pushed" in Bodhi only means that it is in a repository (not that it is necessarily stable). Therefore, testing updates are still considered "pushed". I would assume that you would want developers to be able to waive updates that are in the testing state?

bowlofeggs:
jenkies test

bodhi/server/templates/update.html
matt8754: Aha, I think I got it, thanks, :-D

matt8754:
Ready for another look, thank you.

ralphbean:
@bowlofeggs, @pypingou - are either of you able to take another look?

bodhi/server/templates/update.html
```diff
@@ -397,6 +397,42 @@
       });
     });
 
+    // handle for waiving test restuls.
```
pypingou: Should this code also only be included if tests are turned on as done below?

bowlofeggs:
jenkies test

bodhi/server/templates/update.html
bowlofeggs: I don't personally mind the code being here when tests are turned off, since it won't perform any actions without the form being present.

bowlofeggs:
This looks good to me. The tests don't pass right now due to #2109, but once the fix for that is merged we can rebase this to see the tests pass.

bodhi/server/templates/update.html
pypingou: Doesn't it make the page heavier ?

matt8754:
It seems like something unrelated is broken and tests failed again.

bowlofeggs:
jenkies test

bowlofeggs:
jenkies test

bodhi/server/templates/update.html
bowlofeggs: Yeah it does. I'm certainly not opposed to wrapping it in an if statement.

bowlofeggs: What do you think @matt8754?

bodhi/server/templates/update.html
```diff
@@ -523,6 +558,9 @@ <h1>
             % endif
             <a id='unpush' class="btn btn-sm btn-danger"><span class="fa fa-fw fa-arrow-circle-left"></span> Unpush</a>
           % endif
+          % if update.status.description != 'stable' and request.registry.settings.get('test_gating.required') and not update.test_gating_passed:
```
bowlofeggs: We could make a small function in util.py that does this if block and use that over the three blocks of code in this commit.

matt8754:
This is read y for another round, :-P

bodhi/server/util.py
```diff
@@ -936,6 +936,21 @@ def testcase_link(context, test, short=False):
...
+
+    Args:
+        context (mako.runtime.Context): The current template rendering context. Unused.
+        update (bodhi.server.models.Update): The Update on which we are going to waive test results.
+    Returns:
+        bool: Indicating if the test results can be waived on the given update.
+    """
+    return (config.get('test_gating.required') and
+            not update.test_gating_passed and
+            update.status.description != 'stable')
```
bowlofeggs, line 951: You can do ```update.status != UpdateStatus.stable``` to avoid the string comparison, but this is fine too.

bodhi/tests/server/test_util.py
```diff
@@ -247,6 +247,16 @@ def test_failed(self):
...
 
 
+class TestCanWaiveTestResults(base.BaseTestCase):
+    """Test the can_waive_test_results() function."""
+    @mock.patch.dict('bodhi.server.util.config', {'test_gating.required': True})
+    def test_can_waive_test_results(self):
+        u = Update.query.all()[0]
+        u.test_gating_status = TestGatingStatus.failed
+        u.status = models.UpdateStatus.testing
+        self.assertTrue(util.can_waive_test_results(None, u))
```
bowlofeggs, line 282: We should add a few more tests to hit the other branches of the if statement:

1. It should return ```False``` if the gating is disabled, even if other conditions are met.
2. It should return ```False``` if the tests passed, even if other conditions are met.
3. It should return ```False``` if the update is stable.

bowlofeggs (changes requested):
Let's add a few more tests. I can write them for you if you are tired of this PR, but I won't be able to get to it for a few weeks most likely.

matt8754:
More tests have been added.

bowlofeggs (approved):
Thanks for adding those tests. This looks good to me!

ralphbean:
Exciting!

PR #2100, opened 2017-12-21
Make PDC lookups limited to the set of components we need

puiterwijk:
This will do a few things to optimize:
1. Less requests in general, since most updates likely do not contain a lot of packages.
   This means we don't have to request all of the list, which takes a long time.
   If for whatever reason the update does contain a lot of packages, we fall back to
   just getting the full list, since that is likely just faster then.
2. Less set operations to perform in the few-components case.

### DIFF
--- a/bodhi/server/models.py
+++ b/bodhi/server/models.py
@@ -1705,12 +1705,16 @@ class Update(Base):
             bool: ``True`` if the update contains a critical path package, ``False`` otherwise.
         component
         """
+        relname = release_name.lower()
+        components = defaultdict(list)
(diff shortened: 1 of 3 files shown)


puiterwijk:
@bowlofeggs +1 to the additional patch

bowlofeggs:
Excellent. I will sqash the two commits, wait for CI to pass, and merge. Thanks for the first commit and review of mine!

bowlofeggs:
jenkies test

pypingou (approved):
Looks fine to me

PR #2101, opened 2017-12-21
Bodhi no longer requires progressbar.

bowlofeggs:


pypingou (approved):
\ó/ for less dependencies

PR #2102, opened 2017-12-24
Updated link for develepors docs Signed-off-by: Somesh Chaturvedi <so…

0xsomesh:
updated link for develepors_docs in readme


bowlofeggs:
The current link will be correct once Bodhi 3.2.0 is deployed to production, which should happen "soon". Since the README is distributed with Bodhi, there isn't a way to have a period of time where the README in git has a 404 link and have the distributed README be correct.

PR #2105, opened 2018-01-05
update ffmarkdown link

ktdreyer:
commit fb9f4dc2 moved ffmarkdown.py to the "server" directory.

### DIFF
--- a/bodhi/server/templates/fragments.html
+++ b/bodhi/server/templates/fragments.html
@@ -155,7 +155,7 @@
         </div>
         <div class="modal-body">
           <p> Text fields in Bodhi2 support an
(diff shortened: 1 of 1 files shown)


bowlofeggs (approved):
Good catch!

bowlofeggs:
jenkies test

bowlofeggs:
Don't worry about the CI failure - there is a known Fedora repo outage right now so the CI system wasn't able to get the files it needs.

bowlofeggs:
jenkies test

bowlofeggs:
The patch for this bug has been included in the [```3.3.0b0``` build](https://bodhi.stg.fedoraproject.org/docs/user/release_notes.html), which is currently deployed to [staging](https://bodhi.stg.fedoraproject.org).

PR #2110, opened 2018-01-10
Fix some tests that relied on it being 2017.

### LINKED ISSUE #2109: Some tests are failing due to asserting that it is 2017
Some tests assume that it is 2017 and are failing as a result.

bowlofeggs:
fixes #2109


jeremycline (approved):
:shipit:

Issue #2059, opened 2017-12-11
Python 3 compatibility - final phase


frenzymadness:
Hello.

This issue is here for tracking work on the final phase of making Bodhi Python 3 compatible.

Currently, Bodhi is somehow Python 3 compatible (PR #2006) but we cannot be sure everything is okay and we don't break Python 3 compatibility in the future until we enable tests with Python 3.

Known stuff currently keeping us from enabling tests with Python 3:
* pyramid-fas-openid - https://github.com/fedora-infra/pyramid_fas_openid/pull/7 - We need to merge existing PR, prepare a new specfile with Python 3 subpackage and ship it to Fedoras. Then we can install both versions to VM and start testing with Python 3.
* packagedb-cli - https://github.com/fedora-infra/bodhi/issues/1970 - Backward incompatible change where we have to use some `if six.PY3` conditions to disable this backend and all related tests.

The last big task will be to solve all issues which will appear in tests with Python 3 and then include Python 3 tests to CI.

If I'll have enough time, I'll try to prepare PR with enabled Python 3 tests and all tests depending on blocking stuff mentioned above skipped until they will be fixed.

frenzymadness:
There are last 20 tests skipped in Python 3 which need to be fixed.

bowlofeggs:
bodhi-3.11.0b5 has been deployed to Bodhi's staging environment and uses Python 3 only:

https://bodhi.stg.fedoraproject.org

Please test and let us know if you find issues!

bowlofeggs:
Bodhi 3.11.0b5 is also available via Copr:

https://copr.fedorainfracloud.org/coprs/bowlofeggs/bodhi-pre-release

PR #2115, opened 2018-01-17
Python 3 tests

frenzymadness:
Hello.

As I mentioned in https://github.com/fedora-infra/bodhi/issues/2059 I started with the final phase of Python 3 compatibility for Bodhi. The final phase is based on availability to run tests also with Python 3.

The changes proposed in this PR enables me to execute tests in VM provisioned by Ansible. There are a lot of failures and I'll fix it one by one but with these changes, you can at least try to run tests with Python 3.

After the first look, it seems that some errors are caused by a different wrapping of outputs in CLI.

Feel free to add feedback although it is still in progress.

See you soon :)

Lumír

### DIFF
--- a/bodhi/client/bindings.py
+++ b/bodhi/client/bindings.py
@@ -474,7 +474,7 @@ class BodhiClient(OpenIdBaseClient):
             raise ValueError("No such file or directory: %s" % input_file)
 
         defaults = dict(severity='unspecified', suggest='unspecified')
-        config = configparser.SafeConfigParser(defaults=defaults)
+        config = configparser.ConfigParser(defaults=defaults)
         read = config.read(input_file)
 
         if len(read) != 1 or read[0] != input_file:
--- a/bodhi/client/bindings.py
+++ b/bodhi/client/bindings.py
@@ -80,6 +80,9 @@ class UpdateNotFound(BodhiClientException):
         """
         return u'Update not found: {}'.format(self.update)
 
+    # Use __unicode__ method under __str__ name for Python 3
+    __str__ = __unicode__
+
 
 def errorhandled(method):
     """Raise exceptions on failure. Used as a decorator for BodhiClient methods."""
--- a/bodhi/tests/server/consumers/test_masher.py
+++ b/bodhi/tests/server/consumers/test_masher.py
@@ -16,7 +16,6 @@
 # You should have received a copy of the GNU General Public License
 # along with this program; if not, write to the Free Software
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
-from cStringIO import StringIO
 import datetime
 import dummy_threading
 import errno
@@ -26,12 +25,13 @@ import shutil
 import tempfile
 import time
 import unittest
-import urllib2
-import urlparse
 
 from click import testing
 import mock
 import six
+import six.moves.urllib.parse as urlparse
+from six.moves.urllib.error import HTTPError, URLError
+from six import StringIO
 
 from bodhi.server import buildsys, exceptions, log, push
 from bodhi.server.config import config
@@ -2247,7 +2247,7 @@ class TestPungiComposerThread__wait_for_sync(ComposerThreadBaseTestCase):
     @mock.patch('bodhi.server.consumers.masher.time.sleep')
     @mock.patch(
         'bodhi.server.consumers.masher.urllib2.urlopen',
-        side_effect=[urllib2.HTTPError('url', 404, 'Not found', {}, None),
+        side_effect=[HTTPError('url', 404, 'Not found', {}, None),
                      StringIO('---\nyaml: rules')])
     def test_httperror(self, urlopen, sleep, publish):
         """
@@ -2350,7 +2350,7 @@ class TestPungiComposerThread__wait_for_sync(ComposerThreadBaseTestCase):
     @mock.patch('bodhi.server.consumers.masher.time.sleep')
     @mock.patch(
         'bodhi.server.consumers.masher.urllib2.urlopen',
-        side_effect=[urllib2.URLError('it broke'),
+        side_effect=[URLError('it broke'),
                      StringIO('---\nyaml: rules')])
     def test_urlerror(self, urlopen, sleep, publish):
         """
--- a/bodhi/tests/server/consumers/test_masher.py
+++ b/bodhi/tests/server/consumers/test_masher.py
@@ -1107,7 +1107,7 @@ That was the actual one'''
 
         with self.db_factory() as session:
             with tempfile.NamedTemporaryFile(delete=False) as script:
-                script.write('#!/bin/bash\nsleep 5\nexit 1\n')
+                script.write(b'#!/bin/bash\nsleep 5\nexit 1\n')
                 script.close()
                 os.chmod(script.name, 0o755)
 
--- a/bodhi/tests/server/consumers/test_masher.py
+++ b/bodhi/tests/server/consumers/test_masher.py
@@ -528,6 +528,7 @@ That was the actual one''' % mash_dir
         self.assertEquals(self.koji.__moved__[1],
                           (u'f17-updates-candidate', u'f17-updates-testing', u'bodhi-2.0-2.fc17'))
 
+    @unittest.skipIf(six.PY3, 'Not working with Python 3 yet')
     @mock.patch(**mock_taskotron_results)
     @mock.patch('bodhi.server.consumers.masher.PungiComposerThread._wait_for_pungi')
     @mock.patch('bodhi.server.consumers.masher.PungiComposerThread._sanity_check_repo')
@@ -964,6 +965,7 @@ That was the actual one'''
             force=True,
             topic='mashtask.complete'))
 
+    @unittest.skipIf(six.PY3, 'Not working with Python 3 yet')
     @mock.patch(**mock_taskotron_results)
     @mock.patch('bodhi.server.consumers.masher.PungiComposerThread._wait_for_pungi')
     @mock.patch('bodhi.server.consumers.masher.PungiComposerThread._sanity_check_repo')
@@ -1173,6 +1175,7 @@ That was the actual one'''
              'status_comments': True})
         self.assertTrue(os.path.exists(mash_dir))
 
+    @unittest.skipIf(six.PY3, 'Not working with Python 3 yet')
     @mock.patch(**mock_taskotron_results)
     @mock.patch('bodhi.server.consumers.masher.PungiComposerThread._sanity_check_repo')
     @mock.patch('bodhi.server.consumers.masher.PungiComposerThread._stage_repo')
@@ -2123,6 +2126,7 @@ class TestComposerThread_save_state(ComposerThreadBaseTestCase):
 
 class TestPungiComposerThread__wait_for_sync(ComposerThreadBaseTestCase):
     """This test class contains tests for the PungiComposerThread._wait_for_sync() method."""
+    @unittest.skipIf(six.PY3, 'Not working with Python 3 yet')
     @mock.patch.dict(
         'bodhi.server.consumers.masher.config',
         {'fedora_testing_master_repomd':
@@ -2196,6 +2200,7 @@ class TestPungiComposerThread__wait_for_sync(ComposerThreadBaseTestCase):
         except Exception as ex:
             assert str(ex) == "Not found an arch to _wait_for_sync with"
 
+    @unittest.skipIf(six.PY3, 'Not working with Python 3 yet')
     @mock.patch.dict(
         'bodhi.server.consumers.masher.config',
         {'fedora_testing_master_repomd':
@@ -2239,6 +2244,7 @@ class TestPungiComposerThread__wait_for_sync(ComposerThreadBaseTestCase):
         urlopen.assert_has_calls(expected_calls)
         sleep.assert_has_calls([mock.call(200), mock.call(200)])
 
+    @unittest.skipIf(six.PY3, 'Not working with Python 3 yet')
     @mock.patch.dict(
         'bodhi.server.consumers.masher.config',
         {'fedora_testing_master_repomd':
@@ -2342,6 +2348,7 @@ class TestPungiComposerThread__wait_for_sync(ComposerThreadBaseTestCase):
         t.log.error.assert_called_once_with(
             'Cannot find local repomd: %s', os.path.join(repodata, 'repomd.xml'))
 
+    @unittest.skipIf(six.PY3, 'Not working with Python 3 yet')
     @mock.patch.dict(
         'bodhi.server.consumers.masher.config',
         {'fedora_testing_master_repomd':
@@ -2456,6 +2463,7 @@ class TestComposerThread_send_notifications(ComposerThreadBaseTestCase):
 class TestComposerThread_send_testing_digest(ComposerThreadBaseTestCase):
     """Test ComposerThread.send_testing_digest()."""
 
+    @unittest.skipIf(six.PY3, 'Not working with Python 3 yet')
     @mock.patch('bodhi.server.mail.smtplib.SMTP')
     def test_security_updates(self, SMTP):
         """If there are security updates, the maildata should mention it."""
--- a/bodhi/tests/server/test_util.py
+++ b/bodhi/tests/server/test_util.py
@@ -439,9 +439,9 @@ class TestUtils(base.BaseTestCase):
         self.assertEqual(
             session.get.mock_calls,
             [mock.call(
-                ('http://domain.local/rest_api/v1/component-branches/?name=f26'
-                 '&fields=global_component&global_component=gcc&page_size=100&critical_path=true'
-                 '&active=true&type=rpm'),
+                ('http://domain.local/rest_api/v1/component-branches/?active=true'
+                 '&critical_path=true&fields=global_component&name=f26&page_size=100&type=rpm'
+                 '&global_component=gcc'),
                 timeout=60),
              mock.call().json()])
 
@@ -469,9 +469,9 @@ class TestUtils(base.BaseTestCase):
         self.assertEqual(
             session.get.mock_calls,
             [mock.call(
-                ('http://domain.local/rest_api/v1/component-branches/?name=f26'
-                 '&fields=global_component&global_component=gcc&page_size=100&critical_path=true'
-                 '&active=true&type=rpm'),
+                ('http://domain.local/rest_api/v1/component-branches/?active=true'
+                 '&critical_path=true&fields=global_component&name=f26&page_size=100&type=rpm'
+                 '&global_component=gcc'),
                 timeout=60),
              mock.call().json()])
 
--- a/bodhi/tests/server/test_util.py
+++ b/bodhi/tests/server/test_util.py
@@ -20,7 +20,6 @@ import subprocess
 import unittest
 
 import mock
-import pkgdb2client
 import six
 
 from bodhi.server import util, models
@@ -29,6 +28,9 @@ from bodhi.server.config import config
 from bodhi.server.models import (ComposeState, TestGatingStatus, Update, UpdateRequest,
                                  UpdateSeverity)
 from bodhi.tests.server import base
+
+if six.PY2:
+    import pkgdb2client
 
 
 class TestBugLink(base.BaseTestCase):
--- a/bodhi/tests/server/test_util.py
+++ b/bodhi/tests/server/test_util.py
@@ -351,28 +351,29 @@ class TestUtils(base.BaseTestCase):
         """
         self.assertEqual(util.get_critpath_components(), ['kernel', 'glibc'])
 
-    @mock.patch.object(pkgdb2client.PkgDB, 'get_critpath_packages')
-    @mock.patch.dict(util.config, {
-        'critpath.type': 'pkgdb',
-        'pkgdb_url': 'http://domain.local'
-    })
-    def test_get_critpath_components_pkgdb_success(self, mock_get_critpath):
-        """ Ensure that critpath packages can be found using PkgDB.
-        """
-        # A subset of critpath packages
-        critpath_pkgs = [
-            'pth',
-            'xorg-x11-server-utils',
-            'giflib',
-            'basesystem'
-        ]
-        mock_get_critpath.return_value = {
-            'pkgs': {
-                'f20': critpath_pkgs
+    if six.PY2:
+        @mock.patch.object(pkgdb2client.PkgDB, 'get_critpath_packages')
+        @mock.patch.dict(util.config, {
+            'critpath.type': 'pkgdb',
+            'pkgdb_url': 'http://domain.local'
+        })
+        def test_get_critpath_components_pkgdb_success(self, mock_get_critpath):
+            """ Ensure that critpath packages can be found using PkgDB.
+            """
+            # A subset of critpath packages
+            critpath_pkgs = [
+                'pth',
+                'xorg-x11-server-utils',
+                'giflib',
+                'basesystem'
+            ]
+            mock_get_critpath.return_value = {
+                'pkgs': {
+                    'f20': critpath_pkgs
+                }
             }
-        }
-        pkgs = util.get_critpath_components('f20')
-        assert critpath_pkgs == pkgs, pkgs
+            pkgs = util.get_critpath_components('f20')
+            assert critpath_pkgs == pkgs, pkgs
 
     @mock.patch('bodhi.server.util.http_session')
     @mock.patch.dict(util.config, {
--- a/bodhi/tests/utils.py
+++ b/bodhi/tests/utils.py
@@ -1,0 +1,36 @@
+# -*- coding: utf-8 -*-
+# Copyright © 2018 Red Hat, Inc.
+#
+# This file is part of Bodhi.
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+"""This module contains useful functions that helps with testing."""
+
+import re
+
+
+def prepare_text_for_comparison(text):
+    """Transfer wrapped text output to one line."""
+    text = text.replace('\n', '')  # Remove newlines
+    text = re.sub(r' +:', '', text)  # Remove start of wrapped lines
+    text = "".join(text.split())  # Remove whitespaces
+    return text
+
+
+def compare_output(output, expected):
+    """Compare content of wrapped text outputs and show diff if enabled."""
+    output = prepare_text_for_comparison(output)
+    expected = prepare_text_for_comparison(expected)
+    return output == expected
--- a/bodhi/tests/utils.py
+++ b/bodhi/tests/utils.py
@@ -19,6 +19,7 @@
 """This module contains useful functions that helps with testing."""
 
 import re
+import difflib
 
 
 def prepare_text_for_comparison(text):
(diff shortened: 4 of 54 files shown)


frenzymadness:
I found out that a lot of tests are failing because tested command output looks different than expected. It is caused by the different behavior of `textwrap` module from Python stdlib.

I am able to reproduce it: https://paste.fedoraproject.org/paste/3H3s4BIIC2978r-4AtEiww

I am thinking about fill a bug in Python bug tracker and also about some way to fix it.

frenzymadness:
Ok, I was right. The problem comes from `textwrap` module. `textwrap` behaves differently in Python 2 and Python 3. See reproducer in my previous comment.

Options I see now:
1. Implement different way for text comparison in test
We can, for example, remove all newlines and whitespaces from outputs and compare them .. I know, it is not a good idea.
2. Use `textwrap3` module which is a backport of `textwrap` module from Python 3.6 to Python >= 2.6. This means that we can use `textwrap` in Python 3 and `textwrap3` installed from PyPI in Python 2 and have the same result for both. The downside of this solution is that we have to redefine some expected outputs because they will look different with newer `textwrap`.

If you have anything against point 2, let me know. I'll continue with `textwrap3` in Python 2.

bowlofeggs:
Hi @frenzymadness!

I think it's fine to go with option 1 - honestly we don't need to verify that the output is exactly the same, as long as it still looks reasonable in Python 3. The goal for me wouldn't be to perfectly preserve whitespace, so I think we can just make the tests more flexible in their assertions.

BTW, I'm in Brno!

devel/ansible/roles/dev/files/motd
```diff
@@ -11,6 +11,7 @@ bshell:      Get a handy python shell initialized with Bodhi models.
 bstart:      Start the Bodhi service.
 bstop:       Stop the Bodhi service.
 btest:       Run Bodhi's test suite.
+btest3:      Run Bodhi's test suite with Python 3.
```
bowlofeggs: We could make btest run Python 3 and just rely on CI for Python 2 if you like. This is also fine.

devel/ansible/roles/dev/tasks/main.yml
```diff
@@ -95,6 +124,16 @@
       chdir: /home/vagrant/bodhi
       creates: /usr/lib/python2.7/site-packages/bodhi.egg-link
 
+- name: Get Python 3 version
+  command: python3 -c "import sys; print('%s.%s' % sys.version_info[0:2])"
+  register: python3_version
```
bowlofeggs, line 134: Fancy!

requirements.txt
```diff
@@ -11,10 +11,11 @@ feedgen
 kitchen
 jinja2
 markdown
-packagedb-cli
+packagedb-cli; python_version <= '2.7'
```
bowlofeggs: Ah neat, I didn't know you could express this.

frenzymadness:
With another bunch of changes, I am introducing the new way how to compare strings in test cases where expected and real command output is compared. The rest of changes are related to Python 3 compatibility, unicode/bytes differences, one Python bug etc.

Feel free to review my changes. I'll continue going test by test and fixing all of them.

Current state with Python 3 is: 429 failed, 710 passed

bodhi/client/bindings.py
```diff
@@ -80,6 +80,9 @@ def __unicode__(self):
         """
         return u'Update not found: {}'.format(self.update)
 
+    # Use __unicode__ method under __str__ name for Python 3
+    __str__ = __unicode__
```
bowlofeggs, line 84: https://github.com/fedora-infra/bodhi/pull/2153 also adds a ```__str__()``` method here. Would you care to comment over there so the two PRs coordinate a bit?

bodhi/tests/utils.py
```diff
@@ -0,0 +1,31 @@
+# This program is free software; you can redistribute it and/or
```
bowlofeggs, line 6: Let's add a Copyright claim above this (I've been trying to do this consistently across Bodhi, but a lot of modules still don't…). Something like this:

```
# -*- coding: utf-8 -*-
# Copyright © 2018 Red Hat, Inc.
#
# This file is part of Bodhi.
#
```

bodhi/tests/utils.py
```diff
@@ -0,0 +1,31 @@
...
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+"""
+This module contains a useful functions that helps with testing
```
bowlofeggs: s/a useful/useful/

bodhi/tests/utils.py
```diff
@@ -0,0 +1,31 @@
...
+    text = text.replace('\n', '')  # Remove newlines
+    text = re.sub(r' +:', '', text)  # Remove start of wrapped lines
+    text = "".join(text.split())  # Remove whitespaces
+    return text
+
+
+def compare_output(output, expected):
+    output = prepare_text_for_comparison(output)
+    expected = prepare_text_for_comparison(expected)
+    return output == expected
```
bowlofeggs: I suggest making this raise an ```Exception``` when they don't match, rather than returning a ```bool```, so that it's more similar to how ```assertEqual()``` works in the test suite. This is OK too if you prefer it.

devel/ansible/roles/dev/files/.bashrc
```diff
@@ -37,6 +37,11 @@ function btest {
     pushd /home/vagrant/bodhi && tox && py.test $@; popd
 }
 
+function btest3 {
+    find /home/vagrant/bodhi -name "__pycache__" -type d -exec rm -r {} \;
+    pushd /home/vagrant/bodhi && py.test-3 $@ && tox; popd
+}
```
bowlofeggs: I would probably make ```btest``` run the Python 3 tests, and maybe make a ```btest2``` if you want. We can rely on CI to run them both. On the other hand, this is more similar to the Fedora packaging guidelines, so I can see why you did it this way. Either way is fine with me, whatever you prefer.

bowlofeggs (changes requested):
Looks great, I can't wait to have Bodhi working with Python 3. Let me know if I can assist you in any way!

bowlofeggs:
There are some conflicts to resolve. Also, be sure to sign the DCO in your commit message. Thanks again!

bodhi/client/bindings.py
frenzymadness: Comment added. I'll solve all conflicts at the end of porting process because if I would rebase frequently, it would be harder to see recent changes in this PR and do a review.

bodhi/tests/utils.py
frenzymadness: It depends on what we prefer here.
1. We can use this function inside `assertTrue()` method and let outer method raise exception
2. or we can use `compare_output()` alone and raise AssertionError but in this case `compare_output()` should be renamed to contain assert in the name and this can be confusing because it's not part of unittest module.

bodhi/tests/utils.py
frenzymadness: Fixed in fixup commit

bodhi/tests/utils.py
frenzymadness: Fixed in fixup commit

devel/ansible/roles/dev/files/.bashrc
frenzymadness: We can change it later and publish this change also in docs/readme etc.

frenzymadness:
@bowlofeggs See the last commit, please. I am facing the problem when some message contains randomly generated ID of an update. We can do it like I did in the last commit with comparing start and end of a message in separated asserts or second idea I have is implement some limit of different chars in `compare_output()` functions which can consider two string as equal even if they differ in some count of chars. What do you think?

bodhi/tests/utils.py
bowlofeggs: There's a third option that might be cleaner - we could create a ```BaseClientTestCase``` class that subclasses ```unittest.TestCase``` and defines ```assert_cli_output_similar()``` or some name like that. The server tests already have a unified base class so they can do some fancy common things like this, so it would be consistent with what we have over there.

bodhi/tests/server/consumers/test_masher.py
```diff
@@ -1296,16 +1296,19 @@ def test_absent_gating(self, publish, *args):
...
             u'https://fedoraproject.org/wiki/QA:Updates_Testing for\ninstructions on how to '
             u'install test updates.\nYou can provide feedback for this update here: {}')
-        expected_message = expected_message.format(
+        expected_message_start = expected_message_start.format(
             urlparse.urljoin(
                 config['base_address'],
-                '/updates/FEDORA-{}-a3bbe1a8f2'.format(datetime.datetime.now().year)))
-        on_qa.assert_called_once_with(12345, expected_message)
+                '/updates/FEDORA-{}-'.format(datetime.datetime.now().year)))
+        called_args = on_qa.mock_calls[0][1]
```
bowlofeggs: Hmm, I'm not sure what is going on here. It had seemed that the update ID was being generated in a predictable manner in the test suite before. Perhaps the predictable manner depends on the Python version (like Python 2 predicably makes this, but Python 3 makes something else?).

I'll see if I can find what is making these strings in the test suite.

bowlofeggs: I found it!

Remember that thing that was patching ```uuid.uuid4``` with "wat"? [```Update.assign_alias()```](https://github.com/fedora-infra/bodhi/blob/3.4.0/bodhi/server/models.py#L2238-L2249) was using that to basically make a predictable update alias. So if we can bring that mock back I think this string will be deterministic and identical to what we see here.

frenzymadness:
Hello.

As I promised, I am preparing this PR for a merge.

* Mocked UUID is back. I found another workaround to keep hashes predictable.
* All commits are sign-off.
* `flake8`, `pydocstyle` and `py.test-2` passing
* All incompatible tests have `@unittest.skipif` decorator to skip them in Python 3.
* I added `py.test-3` as a part of `btest` command and I also added it to CI script. Now I just need to figure out how to make tests in CI pass. Could you please help me with this? I don't understand the hierarchy of files with pip/rpm packages specified in CI folder.

But, I also found one bug I cannot solve right now. During the provision of the vagrant box via Ansible, the file `/usr/local/lib/python3.6/site-packages/python_openid-2.2.5-py3.6.egg` somehow appears and it causes a problem with Python 3 tests because for Python 3 `python3-openid` should be installed. `python-openid` is not Python 3 compatible and I cannot find out where this egg comes from and how it's possible to Python 2 only package/egg to be installed for Python 3.

I'll do m best to solve the last issue asap so we'll be able to merge the PR. In the meantime, you can do a review.

Thanks

bowlofeggs:
Hello @frenzymadness!

Sure, I can take a look at the remaining issues. I'll do some poking and see what I come up with. Thanks so much for all your efforts on this!

bowlofeggs:
@frenzymadness When I use this pull request to do a vagrant up, I have an empty ```/usr/local/lib```, but I do see openid at ```/usr/lib/python3.6/site-packages/python_openid-2.2.5-py3.6.egg```. I believe it is getting installed when ```python3 setup.py develop``` is called:

```
$ sudo python3 setup.py develop
<snip>
Installed /home/vagrant/bodhi                                     
Processing dependencies for bodhi-server==3.5.0          
Searching for python-openid     
Reading https://pypi.python.org/simple/python-openid/
Downloading https://pypi.python.org/packages/aa/85/0bac9096d208f04c9de9d49f3a4e27fec8bfd56ee841c185f5cf420a24d2/python-openid-2.2.5.zip#md5=f89d9d4f4dccfd33b5ce34eb4725f751
Best match: python-openid 2.2.5                                        
Processing python-openid-2.2.5.zip                     
Writing /tmp/easy_install-ydswl49_/python-openid-2.2.5/setup.cfg  
Running python-openid-2.2.5/setup.py -q bdist_egg --dist-dir /tmp/easy_install-ydswl49_/python-openid-2.2.5/egg-dist-tmp-a_ec1aui
<snip>
Moving python_openid-2.2.5-py3.6.egg to /usr/lib/python3.6/site-packages
Adding python-openid 2.2.5 to easy-install.pth file

Installed /usr/lib/python3.6/site-packages/python_openid-2.2.5-py3.6.egg
```

```python-openid``` is listed in ```requirements.txt``` as a requirement. However, I don't see it being used directly in Bodhi's code. I don't have time to test without it today, but tomorrow I will make a Vagrant box with that removed from requirements.txt and do some testing to make sure the login still works. We might be able to just drop it.

The tests do seem to pass even with that package present (with 346 skipped, which is fine - I can deal with those in follow up PRs).

As for the CI stuff, yeah it's kinda weird. Basically, ```devel/run_tests.sh``` cats those Dockerfile snippets together for each Fedora release. Since packages have different names in different Fedora release, we made Dockerfile snippets for each release (and one for pip installing too). For packages that have the same name in all Fedora releases, there's a common rpm snippet. It's all kinda weird. If you want, I can take care of that for you tomorrow.

**Comment by frenzymadness** [comment]
Great, you are right. Removing python-openid from requirements.txt fixed the problem. Python-openid is still installed as a dependency of python-fas-openid but in that case the right version in right folder.

I'll run all tests in fresh VM once again and then I'll push everything here and I'll start fixing problems in CI.

**Comment by bowlofeggs** [comment]
@frenzymadness I removed python-openid from ```requirements.txt``` on the develop branch, and I am able to log into the web UI and CLI just fine, and the tests pass. I say let's just drop it from ```requirements.txt```.

**Comment by bowlofeggs** [comment]
@frenzymadness An idea: IMO we don't need to have CI running Python 3 stuff in this same pull request. The PR is already pretty huge - if you want, we can focus on getting it merged without CI and I can make it work with CI in a follow up PR. Since you have btest running Python 3 tests and passing I think that'll do.

I do see one current problem that we do need to solve however - the Vagrant box doesn't seem to serve Bodhi - it has some strange dogpile cache related crash. For some reason, it is running Python 3 code even though the development server is still launched with /usr/bin/pserve (not /usr/bin/pserve-3). No idea why it's calling into the Python 3 stack, or why dogpile cache is failing, but you can see it if you visit http://localhost:6543

```
Mar 15 13:54:28 bodhi-dev.example.com pserve[3309]: 2018-03-15 13:54:28,943 ERROR [bodhi.server][waitress] Error caught.  Handling HTML response.                                                                                   [362/1796]
Mar 15 13:54:28 bodhi-dev.example.com pserve[3309]: Traceback (most recent call last):                                                
Mar 15 13:54:28 bodhi-dev.example.com pserve[3309]:   File "/usr/lib/python3.6/site-packages/pyramid/tweens.py", line 39, in excview_tween                                                                                                   
Mar 15 13:54:28 bodhi-dev.example.com pserve[3309]:     response = handler(request)                                                                       
Mar 15 13:54:28 bodhi-dev.example.com pserve[3309]:   File "/usr/lib/python3.6/site-packages/pyramid/router.py", line 156, in handle_request                                                                                                 
Mar 15 13:54:28 bodhi-dev.example.com pserve[3309]:     view_name                                                                          
Mar 15 13:54:28 bodhi-dev.example.com pserve[3309]:   File "/usr/lib/python3.6/site-packages/pyramid/view.py", line 642, in _call_view                                                                                                       
Mar 15 13:54:28 bodhi-dev.example.com pserve[3309]:     response = view_callable(context, request)                                     
Mar 15 13:54:28 bodhi-dev.example.com pserve[3309]:   File "/usr/lib/python3.6/site-packages/pyramid/config/views.py", line 181, in __call__                                                                                                 
Mar 15 13:54:28 bodhi-dev.example.com pserve[3309]:     return view(context, request)                                                  
Mar 15 13:54:28 bodhi-dev.example.com pserve[3309]:   File "/usr/lib/python3.6/site-packages/pyramid/viewderivers.py", line 390, in attr_view                                                                                                
Mar 15 13:54:28 bodhi-dev.example.com pserve[3309]:     return view(context, request)                                              
Mar 15 13:54:28 bodhi-dev.example.com pserve[3309]:   File "/usr/lib/python3.6/site-packages/pyramid/viewderivers.py", line 368, in predicate_wrapper                                                                                        
Mar 15 13:54:28 bodhi-dev.example.com pserve[3309]:     return view(context, request)                                                      
Mar 15 13:54:28 bodhi-dev.example.com pserve[3309]:   File "/usr/lib/python3.6/site-packages/pyramid/viewderivers.py", line 352, in authdebug_view       
Mar 15 13:54:28 bodhi-dev.example.com pserve[3309]:     return view(context, request)                                                      
Mar 15 13:54:28 bodhi-dev.example.com pserve[3309]:   File "/usr/lib/python3.6/site-packages/pyramid/viewderivers.py", line 439, in rendered_view                                                                                            
Mar 15 13:54:28 bodhi-dev.example.com pserve[3309]:     result = view(context, request)                                                   
Mar 15 13:54:28 bodhi-dev.example.com pserve[3309]:   File "/usr/lib/python3.6/site-packages/pyramid/viewderivers.py", line 148, in _requestonly_view           
Mar 15 13:54:28 bodhi-dev.example.com pserve[3309]:     response = view(request)                                                            
Mar 15 13:54:28 bodhi-dev.example.com pserve[3309]:   File "/usr/lib/python3.6/site-packages/cornice/service.py", line 496, in wrapper                              
Mar 15 13:54:28 bodhi-dev.example.com pserve[3309]:     response = view_(request)                                                      
Mar 15 13:54:28 bodhi-dev.example.com pserve[3309]:   File "/home/vagrant/bodhi/bodhi/server/services/updates.py", line 115, in get_update
Mar 15 13:54:28 bodhi-dev.example.com pserve[3309]:     proxy_request = bodhi.server.security.ProtectedRequest(request)
Mar 15 13:54:28 bodhi-dev.example.com pserve[3309]:   File "/home/vagrant/bodhi/bodhi/server/security.py", line 310, in __init__
Mar 15 13:54:28 bodhi-dev.example.com pserve[3309]:     setattr(self, attr, getattr(self.real_request, attr))
Mar 15 13:54:28 bodhi-dev.example.com pserve[3309]:   File "/usr/lib/python3.6/site-packages/pyramid/decorator.py", line 42, in __get__
Mar 15 13:54:28 bodhi-dev.example.com pserve[3309]:     val = self.wrapped(inst)
Mar 15 13:54:28 bodhi-dev.example.com pserve[3309]:   File "/usr/lib/python3.6/site-packages/pyramid/util.py", line 80, in <lambda>
Mar 15 13:54:28 bodhi-dev.example.com pserve[3309]:     fn = lambda this: callable(this)
Mar 15 13:54:28 bodhi-dev.example.com pserve[3309]:   File "/home/vagrant/bodhi/bodhi/server/__init__.py", line 118, in get_user
Mar 15 13:54:28 bodhi-dev.example.com pserve[3309]:     return munchify(user.__json__(request=request))
Mar 15 13:54:28 bodhi-dev.example.com pserve[3309]:   File "/home/vagrant/bodhi/bodhi/server/models.py", line 349, in __json__
Mar 15 13:54:28 bodhi-dev.example.com pserve[3309]:     return self._to_json(self, request=request, anonymize=anonymize)
Mar 15 13:54:28 bodhi-dev.example.com pserve[3309]:   File "/home/vagrant/bodhi/bodhi/server/models.py", line 369, in _to_json
Mar 15 13:54:28 bodhi-dev.example.com pserve[3309]:     attribute = attribute(request)
Mar 15 13:54:28 bodhi-dev.example.com pserve[3309]:   File "/home/vagrant/bodhi/bodhi/server/models.py", line 3988, in avatar
Mar 15 13:54:28 bodhi-dev.example.com pserve[3309]:     return get_avatar(context=context, username=self.name, size=24)
Mar 15 13:54:28 bodhi-dev.example.com pserve[3309]:   File "/home/vagrant/bodhi/bodhi/server/util.py", line 388, in avatar
Mar 15 13:54:28 bodhi-dev.example.com pserve[3309]:     return work(username, size)
Mar 15 13:54:28 bodhi-dev.example.com pserve[3309]:   File "/usr/lib/python3.6/site-packages/dogpile/cache/region.py", line 1220, in decorate
Mar 15 13:54:28 bodhi-dev.example.com pserve[3309]:     should_cache_fn)
Mar 15 13:54:28 bodhi-dev.example.com pserve[3309]:   File "/usr/lib/python3.6/site-packages/dogpile/cache/region.py", line 825, in get_or_create
Mar 15 13:54:28 bodhi-dev.example.com pserve[3309]:     async_creator) as value:
Mar 15 13:54:28 bodhi-dev.example.com pserve[3309]:   File "/usr/lib/python3.6/site-packages/dogpile/lock.py", line 154, in __enter__
Mar 15 13:54:28 bodhi-dev.example.com pserve[3309]:     return self._enter()
Mar 15 13:54:28 bodhi-dev.example.com pserve[3309]:   File "/usr/lib/python3.6/site-packages/dogpile/lock.py", line 87, in _enter
Mar 15 13:54:28 bodhi-dev.example.com pserve[3309]:     value = value_fn()
Mar 15 13:54:28 bodhi-dev.example.com pserve[3309]:   File "/usr/lib/python3.6/site-packages/dogpile/cache/region.py", line 780, in get_value
Mar 15 13:54:28 bodhi-dev.example.com pserve[3309]:     value = self.backend.get(key)
Mar 15 13:54:28 bodhi-dev.example.com pserve[3309]:   File "/usr/lib/python3.6/site-packages/dogpile/cache/backends/file.py", line 220, in get
Mar 15 13:54:28 bodhi-dev.example.com pserve[3309]:     with self._dbm_file(False) as dbm:
Mar 15 13:54:28 bodhi-dev.example.com pserve[3309]:   File "/usr/lib64/python3.6/contextlib.py", line 81, in __enter__
Mar 15 13:54:28 bodhi-dev.example.com pserve[3309]:     return next(self.gen)
Mar 15 13:54:28 bodhi-dev.example.com pserve[3309]:   File "/usr/lib/python3.6/site-packages/dogpile/cache/backends/file.py", line 215, in _dbm_file
Mar 15 13:54:28 bodhi-dev.example.com pserve[3309]:     "w" if write else "r")
Mar 15 13:54:28 bodhi-dev.example.com pserve[3309]:   File "/usr/lib64/python3.6/dbm/__init__.py", line 88, in open
Mar 15 13:54:28 bodhi-dev.example.com pserve[3309]:     raise error[0]("db type could not be determined")
Mar 15 13:54:28 bodhi-dev.example.com pserve[3309]: dbm.error: db type could not be determined
```

**Comment by bowlofeggs** [comment]
Btw, if you want to make the dev server use /usr/bin/pserve-3, you can edit ```devel/ansible/roles/dev/files/bodhi.service``` and do a fresh ```vagrant up``` (or a ```vagrant provision``` should work too).

**Comment by bowlofeggs** [comment]
Actually, if you do vagrant provision you will also need to do a ```systemctl daemon-reload``` inside the guest.

**Comment by bowlofeggs** [comment]
Ooooh, false alarm about that dogpile cache traceback! It turns out that all I needed to do was to ```rm dogpile-cache.dbm*``` in my checkout and everything is fine. I guess the dbm file format changed between Python 2 and Python 3, but deleting is fine since it's just a cache. So think do seem to be working.

We should make that change to the systemd unit file, so that Bodhi is launched with ```pserve-3``` instead of ```pserve```. Also, I had to install ```python3-pyramid``` to get that working, so we'll need to add that to the playbook as well.

File: bodhi/tests/server/test_util.py [hunk 10]
```diff
@@ -349,28 +351,29 @@ def test_get_critpath_components_dummy(self):
...
-        critpath_pkgs = [
-            'pth',
-            'xorg-x11-server-utils',
-            'giflib',
-            'basesystem'
-        ]
-        mock_get_critpath.return_value = {
-            'pkgs': {
-                'f20': critpath_pkgs
+    if six.PY2:
```
bowlofeggs, line 354: Should we use the ```@skipif``` decorator here too?

devel/ansible/roles/dev/tasks/main.yml
```diff
@@ -63,9 +63,30 @@
...
+      - python3-fedora
+      - python3-kitchen
+      - python3-markdown
+      - python3-mock
+      - python3-munch
+      - python3-openid
+      - python3-pillow
+      - python3-psycopg2
       - python3-pydocstyle
+      - python3-pylibravatar
```
bowlofeggs, line 83: We should put ```python3-pyramid``` here too.

bowlofeggs (changes requested):
Cool, this is looking great. I only see three issues left for merge. The two noted below, and we should adjust that systemd unit file I mentioned in an earlier comment to use ```pserve-3```.

If you want, I can make CI work in a follow on commit, and I can start working on the skipped tests too.

This is very exciting!

bodhi/tests/server/test_util.py
frenzymadness: No, because in this case, we need to skip also all the decorators and therefore we cannot use `@unittest.skipif`.

devel/ansible/roles/dev/tasks/main.yml
frenzymadness: Fixed in the next commit.

frenzymadness:
I'll take a look at the systemd file later (probably tomorrow) or if you want to merge this PR sooner it would be better if you can fix it.

Also feel free to add some commit to fix CI or let CI fix as a next PR. Then, it'll be better to fix tests in smaller groups or per file.

bowlofeggs:
FYI - I just rebased your PR on the latest HEAD from develop, because it has some commits that make it easier to run the CI tests locally, which will help me to get them to pass.

bowlofeggs:
For the record, you can run the CI tests inside the Vagrant box now (as of this week!) like this:

```
$ sudo dnf install docker parallel -y
$ sudo systemctl start docker
$ sudo devel/run_tests.sh
```

That last command also accepts a ```-x``` for failfast if you want, and you can also limit it to a list of specific releases like this:

```
$ RELEASES="pip rawhide" sudo -E devel/run_tests.sh -x
```

Anyways, I'll see if I can get the CI tests to be happy!

bowlofeggs:
There is a problem that I'm not sure what to do about:

The docs don't build on the pip container due to PkgDB, even though PkgDB's import is guarded by an "if six.py3":

```
/bodhi/docs/developer/docblocks/bodhi.server.models.rst:38:failed to import PkgDB
```

Strangely, the docs do build on the Rawhide container.

bowlofeggs:
I made some good progress today on this PR. I think it's pretty close to ready, but there are still some straggling issues to solve. I will work on it some more on Monday.

bowlofeggs:
P.S. I hacked around the pkgdb issue by moving the docs building to be last and seding it out. Kinda cheating but it works.

frenzymadness:
Is there anything I can help with? Honestly, I am kinda lost in your progress but I'll do anything to finish this PR asap.

bowlofeggs:
@frenzymadness The current issue seems to be some failing tests. I basically worked on this until late Friday night (in my TZ) and got too tired to continue, so I haven't looked into the current test failures to say what's wrong with them yet. The traceback isn't very useful because it's basically just the test suite being unable to deserialize JSON, so I suspect the real issue is in the serialization code somewhere. It'll take some debugging.

Strangely, the container tests are failing on Fedora 26 and 27. In my Fedora 26 Vagrant box I cannot reproduce the problems with py.test-3 directly, but I can run the container tests in Vagrant and see the failures. So it fails in the container but not in Vagrant… weird. No idea what's going on there.

Anyways, I can focus on this tomorrow. Feel free to poke at it too if you want, but you can also leave it to me if you are too busy. If you do poke at it, don't forget that I force pushed your branch to rebase it. Getting this merged is high on my priority list too, because the longer we wait to merge the more it'll diverge and keep needing new fixes. I'd love to have CI on new patches to ensure they are Python 3 compatible :)

Thanks again so much for all your diligent work here!

frenzymadness:
> The current issue seems to be some failing tests. I basically worked on this until late Friday night (in my TZ) and got too tired to continue, so I haven't looked into the current test failures to say what's wrong with them yet. The traceback isn't very useful because it's basically just the test suite being unable to deserialize JSON, so I suspect the real issue is in the serialization code somewhere. It'll take some debugging.

I am diving into this issue and it seems that the problem is probably caused by renderer because it returns HTML response in some cases and in other cases returns JSON. The renderer behavior is unpredictable. I tried to run one failing test (`TestStacksService.test_get_single_stack`) 50 times in a loop and it failed 11 times and passed 39 times.

It also seems that rendered is chosen based on some random thing at the beginning of tests run because they fail all or none of them.

frenzymadness:
Okay, I think I got it. It seems that is necessary to request JSON in request headers. Otherwise, the renderer is chosen randomly. I tried to add `headers={'Accept': 'application/json'}` to all `self.app.get()` calls in all failed tests and it seems that it solved the problem. I run all touched test files 50 times in a loop and all passed.

Unfortunately, I don't know how to add commit to this PR and I hopefully didn't screw it trying to do it so I am attaching a patch with the mentioned fix.
[0001-Add-request-header-to-request-JSON-response-in-tests.txt](https://github.com/fedora-infra/bodhi/files/1824841/0001-Add-request-header-to-request-JSON-response-in-tests.txt)

I'll try to run all CI tests as you described and inspect failures.

bowlofeggs:
That's an interesting find! I guess it was more deterministic in Python 2 than it is in Python 3. Strange.

bowlofeggs (approved):
So long as CI passes, I think we're good here.

PR #2119, opened 2018-01-17
Don't use autoflush when setting the release on a new Update.

bowlofeggs:
fixes #2117

Signed-off-by: Randy Barlow <randy@electronsweatshop.com>

Issue #2117, opened 2018-01-17
null value in column "release_id" violates not-null constraint when creating updates


bowlofeggs:
While testing the 3.2.0 beta today I noticed that creating updates causes a traceback:

```
2018-01-17 15:57:07,275 INFO  [bodhi.server][waitress] Creating new update: [<RpmBuild {'release_id': 17, 'nvr': u'erlang-fs-4.1-2.fc27', 'signed': False, 'epoch': 0, 'ci_url': N[50/1955]
e': u'rpm'}>]
2018-01-17 15:57:07,399 DEBUG [bodhi.server][waitress] Creating new Update(**data) object.
2018-01-17 15:57:07,407 ERROR [bodhi.server][waitress] Failed to create update
Traceback (most recent call last):
  File "/home/vagrant/bodhi/bodhi/server/services/updates.py", line 503, in new_update
    result, _caveats = Update.new(request, _data)
  File "/home/vagrant/bodhi/bodhi/server/models.py", line 1776, in new
    up.release = release
  File "/usr/lib64/python2.7/site-packages/sqlalchemy/orm/attributes.py", line 224, in __set__
    instance_dict(instance), value, None)
  File "/usr/lib64/python2.7/site-packages/sqlalchemy/orm/attributes.py", line 812, in set
    value = self.fire_replace_event(state, dict_, value, old, initiator)
  File "/usr/lib64/python2.7/site-packages/sqlalchemy/orm/attributes.py", line 834, in fire_replace_event
    self._replace_token or self._init_append_or_replace_token())
  File "/usr/lib64/python2.7/site-packages/sqlalchemy/orm/util.py", line 118, in set_
    return validator(state.obj(), key, value)
  File "/home/vagrant/bodhi/bodhi/server/models.py", line 1582, in validate_release
    u = Update.query.filter(Update.release_id == release.id, Update.id != self.id).first()
  File "/usr/lib64/python2.7/site-packages/sqlalchemy/orm/query.py", line 2755, in first
    ret = list(self[0:1])
  File "/usr/lib64/python2.7/site-packages/sqlalchemy/orm/query.py", line 2547, in __getitem__
    return list(res)
  File "/usr/lib64/python2.7/site-packages/sqlalchemy/orm/query.py", line 2854, in __iter__
    self.session._autoflush()
  File "/usr/lib64/python2.7/site-packages/sqlalchemy/orm/session.py", line 1407, in _autoflush
    util.raise_from_cause(e)
[...]

PR #2120, opened 2018-01-17
Backport the fix for #2117 to the 3.2 branch

bowlofeggs:
This PR is here for CI to run. The patch has already been approved in #2119 so this does not need review.


puiterwijk (approved):
It might not need a review, but you still get one :).  :goat:

Issue #2121, opened 2018-01-17
bodhi-check-policies doesn't filter updates by current releases


bowlofeggs:
While testing the 3.2.0 beta on stg today, I noticed that ```bodhi-check-policies``` asks greenwave about a bunch of EPEL 5 updates. It prints out error messages, similar to this:

```
Bodhi failed to send POST request to Greenwave at the following URL "https://greenwave-web-greenwave.app.os.stg.fedoraproject.org/api/v1.0/decision". The status code was "404". The error was "{u'message': u'Cannot find any applicable policies for fedora-epel-5 and bodhi_update_push_stable'}".
```

bowlofeggs:
I timed a run of this CLI on staging today and it took 125 minutes. I expect we could make it much faster if we skipped all the EOL releases.

PR #2125, opened 2018-01-19
Include check the updates for active releases when checking the gating policies

pypingou:
Fixes  https://github.com/fedora-infra/bodhi/issues/2121

### DIFF
--- a/bodhi/server/scripts/check_policies.py
+++ b/bodhi/server/scripts/check_policies.py
@@ -35,9 +35,20 @@ def check():
     initialize_db(config.config)
     session = Session()
 
-    updates = models.Update.query.filter(models.Update.status.in_(
-        [models.UpdateStatus.pending, models.UpdateStatus.testing])).order_by(
-        models.Update.id.desc())
+    updates = models.Update.query.filter(
+        models.Update.status.in_(
+            [models.UpdateStatus.pending, models.UpdateStatus.testing])
+    ).filter(
+        models.Update.release_id == models.Release.id
+    ).filter(
+        models.Release.state.in_(
+            [models.ReleaseState.current, models.ReleaseState.pending])
+    ).order_by(
+        # Check the older updates first so there is more time for the newer to
+        # get their test results
+        models.Update.id.asc()
+    )
+
     for update in updates:
         try:
             update.update_test_gating_status()
--- a/bodhi/tests/server/scripts/test_check_policies.py
+++ b/bodhi/tests/server/scripts/test_check_policies.py
@@ -220,3 +220,25 @@ class TestCheckPolicies(BaseTestCase):
                  'type': 'bodhi_update'}]}
         mock_greenwave.assert_called_once_with(config['greenwave_api_url'] + '/decision',
                                                expected_query)
+
+    @patch.dict(config, [('greenwave_api_url', 'http://domain.local')])
+    def test_archived_release_updates(self):
+        """Assert that updates for archived releases isn't being considered
+        by the script.
+        """
+        # Archive the F17 release
+        rel = self.db.query(models.Release).filter_by(name='F17').one()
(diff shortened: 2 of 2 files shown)


puiterwijk (approved):
Looks same to me

bowlofeggs (changes requested):
Let's add a test to ensure that this filters out the right updates.

bowlofeggs:
@pypingou Do you still want to work on this?

pypingou:
>    @pypingou Do you still want to work on this?

I do think this feature is useful, last time we reduce the running time to under
30 minutes.

pypingou:
Added some tests, very basic :)

pypingou:
Not seeing what's wrong in the tests in jenkins :(

pypingou:
jenkies test

bodhi/server/scripts/check_policies.py
```diff
@@ -35,9 +35,18 @@ def check():
...
+    updates = models.Update.query.filter(
+        models.Update.status.in_(
+            [models.UpdateStatus.pending, models.UpdateStatus.testing])
+    ).filter(
+        models.Update.release_id == models.Release.id
+    ).filter(
+        models.Release.state.in_(
+            [models.ReleaseState.current, models.ReleaseState.pending])
+    ).order_by(
+        models.Update.id.desc()
```
bowlofeggs: The most recent updates are the most likely to be untested. By checking those first, you increase the chances of getting the missing tests error from Greenwave - if we instead sort ascending, we move them to the end of the list we check and increase the chances of finding their tests.

bodhi/tests/server/scripts/test_check_policies.py
```diff
@@ -220,3 +220,33 @@ def test_unrestricted_policy(self):
...
+        update.status = models.UpdateStatus.testing
+        self.db.commit()
+        with patch('bodhi.server.models.util.greenwave_api_post') as mock_greenwave:
+            greenwave_response = {
+                'policies_satisfied': True,
+                'summary': 'All tests passed',
+                'applicable_policies': ['taskotron_release_critical_tasks'],
+                'unsatisfied_requirements': []
+            }
+            mock_greenwave.return_value = greenwave_response
```
bowlofeggs: You could do this instead:

```
mock_greenwave.side_effect = Exception('Greenwave should not be accessed for archived releases.')
```

bodhi/tests/server/scripts/test_check_policies.py
```diff
@@ -220,3 +220,33 @@ def test_unrestricted_policy(self):
...
+            result = runner.invoke(check_policies.check, [])
+            self.assertEqual(result.exit_code, 0)
+            update = self.db.query(models.Update).filter(models.Update.id == update.id).one()
+            self.assertEqual(update.test_gating_status, models.TestGatingStatus.passed)
+            self.assertEqual(update.greenwave_summary_string, None)
```
bowlofeggs: These three lines don't really check anything since this is both the starting state of the update, but is also the state the update would get to if greenwave were erroneously accessed. I suggest just making the greenwave access raise an Exception and then drop these three lines to keep the test shorter and more laser focused.

bowlofeggs (approved):
All comments are suggestions, so feel free to do as you please with them and merge when you are ready.

bodhi/server/scripts/check_policies.py
pypingou: I didn't add this piece, but I'm ok with changing this.

bodhi/tests/server/scripts/test_check_policies.py
pypingou: Thanks for the suggestion

pypingou:
All comments addressed

bowlofeggs (approved):
Thanks!

PR #2126, opened 2018-01-19
Ensure there is a summary to display before displaying it

pypingou:
If an update is waiting for its information about whether it passed
gating or not, then its gating status is False, but there is no
greenwave summary since greenwave hasn't been queried yet.

So ensure there is a greenwave summary to display before displaying it

### DIFF
--- a/bodhi/tests/server/services/test_updates.py
+++ b/bodhi/tests/server/services/test_updates.py
@@ -4548,6 +4548,84 @@ class TestUpdatesService(base.BaseTestCase):
         up = self.db.query(Update).filter_by(title=resp.json['title']).one()
         self.assertEquals(up.request, UpdateRequest.stable)
 
+    @mock.patch(**mock_taskotron_results)
+    @mock.patch(**mock_valid_requirements)
+    def test_alert_not_displayed_if_no_summary_present(self, publish, *args):
+        """ Ensure that the alert is not display if there are not summary
+        status (returned normaly by greenwave) set in the database for this
+        update.
+        """
+        nvr = u'bodhi-2.1-1.fc17'
+
+        up_data = self.get_update(nvr)
+        up_data['csrf_token'] = self.app.get('/csrf').json_body['csrf_token']
+        self.app.post_json('/updates/', up_data)
+
+        build = self.db.query(RpmBuild).filter_by(nvr=nvr).one()
+        build.update.test_gating_status = TestGatingStatus.failed
+        self.assertEqual(build.update.request, UpdateRequest.testing)
+
+        with mock.patch.dict(self.app.app.registry.settings, {'test_gating.required': True}):
+            resp = self.app.get('/updates/%s' % nvr,
+                                headers={'Accept': 'text/html'})
+            self.assertIn('text/html', resp.headers['Content-Type'])
+            self.assertIn(nvr, resp)
+            self.assertNotIn(
+                '<div class="alert alert-danger">The update can not be pushed:',
+                resp)
+
+    @mock.patch(**mock_taskotron_results)
+    @mock.patch(**mock_valid_requirements)
+    def test_alert_notdisplayed_if_summary_present_but_gating_off(
+            self, publish, *args):
+        """ Ensure that the alert is not being displayed even if there is a
+        summary status (returned normaly by greenwave) set in the database
+        for this update, because the configuration is set otherwise.
+        """
+        nvr = u'bodhi-2.1-1.fc17'
+
+        up_data = self.get_update(nvr)
+        up_data['csrf_token'] = self.app.get('/csrf').json_body['csrf_token']
+        self.app.post_json('/updates/', up_data)
+
+        build = self.db.query(RpmBuild).filter_by(nvr=nvr).one()
+        build.update.test_gating_status = TestGatingStatus.failed
+        build.update.greenwave_summary_string = 'Because Randy said so'
+        self.assertEqual(build.update.request, UpdateRequest.testing)
+
+        with mock.patch.dict(self.app.app.registry.settings, {'test_gating.required': False}):
+            resp = self.app.get('/updates/%s' % nvr,
+                                headers={'Accept': 'text/html'})
+            self.assertIn('text/html', resp.headers['Content-Type'])
+            self.assertIn(nvr, resp)
+            self.assertNotIn(
+                '<div class="alert alert-danger">The update can not be pushed: \' + summary ', resp)
+
+    @mock.patch(**mock_taskotron_results)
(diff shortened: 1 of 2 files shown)


bowlofeggs (changes requested):
Hello @pypingou, long time no see.

Let's add a test to make sure the text "The update can not be pushed: None" does not appear in this case.

Thanks for the patch!

pypingou:
jenkies test

pypingou:
Annoyingly the print call I added to help me debug doesn't seem to show up in the logs :(

pypingou:
jenkies test

pypingou:
Ok, now I see why the test is failing, it's basically because the check whether to display the message is done in JS and not by the template.
So as is, I can't in fact test that the message doesn't show.

options that I can think of:
- Drop the test that fail
- Have the test checks something else, any idea?
- Move the logic for displaying the message from JS code into the template itself.

What do you think?

pypingou:
jenkies tests

pypingou:
jenkies tests

pypingou:
jenkies test

bowlofeggs:
We don't really have tests for the JS (though that would be nice... maybe one day) so if the check is in JS then I think maybe it's OK to not worry about it.

bowlofeggs:
Also, sorry - I hadn't noticed before that the if statement was JS vs. template code - my mistake!

bodhi/tests/server/services/test_updates.py
```diff
@@ -4548,6 +4548,86 @@ def test_urgent_update_bypass_batched(self, publish, *args):
...
+        build = self.db.query(RpmBuild).filter_by(nvr=nvr).one()
+        build.update.test_gating_status = TestGatingStatus.failed
+        build.update.greenwave_summary_string = 'Because Randy said so'
+        self.assertEqual(build.update.request, UpdateRequest.testing)
+
+        resp = self.app.get('/updates/%s' % nvr,
+                            headers={'Accept': 'text/html'})
+        self.assertIn('text/html', resp.headers['Content-Type'])
+        self.assertIn(nvr, resp)
+        print(resp)
```
bowlofeggs: Let's drop this print statement if you go with the template boolean logic (and thus keep these tests).

bowlofeggs (changes requested):

bodhi/tests/server/services/test_updates.py
pypingou: It is/was there for debugging purposes until I got jenkins to be happy (which afaics, is still not the case)

pypingou:
Ok, I'm not managing to find the way to change the config for this call, any idea?

bowlofeggs:
@pypingou Since the settings are copied into the registry, you need to mock them there instead of mocking them at config (which happens after they are already copied). Also, I think there was a missing ```not``` in the template:

```
diff --git a/bodhi/server/templates/update.html b/bodhi/server/templates/update.html
index 2aa4a1b3..779c27e8 100644
--- a/bodhi/server/templates/update.html
+++ b/bodhi/server/templates/update.html
@@ -86,7 +86,7 @@ ${update.type} update for ${update.beautify_title(amp=True) | n}
     ${ update.test_gating_passed }
     ${ update.greenwave_summary_string }
     % if request.registry.settings.get("test_gating.required") is True and \
-            update.test_gating_passed and update.greenwave_summary_string:
+            not update.test_gating_passed and update.greenwave_summary_string:
         var summary = '${update.greenwave_summary_string}';
         $('#resultsdb h3').after(
            '<div class="alert alert-danger">The update can not be pushed: ' + summary + '</div>');
index d5195be4..abc40797 100644
@@ -4565,8 +4565,10 @@ class TestUpdatesService(base.BaseTestCase):
         build.update.test_gating_status = TestGatingStatus.failed
         self.assertEqual(build.update.request, UpdateRequest.testing)
 
-        resp = self.app.get('/updates/%s' % nvr,
-                            headers={'Accept': 'text/html'})
+        with mock.patch.dict(self.app.app.registry.settings, {'test_gating.required': True}):
+            resp = self.app.get('/updates/%s' % nvr,
+                                headers={'Accept': 'text/html'})
+
         self.assertIn('text/html', resp.headers['Content-Type'])
         self.assertIn(nvr, resp)
         self.assertNotIn(
@@ -4592,18 +4594,16 @@ class TestUpdatesService(base.BaseTestCase):
         build.update.greenwave_summary_string = 'Because Randy said so'
         self.assertEqual(build.update.request, UpdateRequest.testing)
 
-        with mock.patch.dict(config, {'test_gating.required': False}):
+        with mock.patch.dict(self.app.app.registry.settings, {'test_gating.required': False}):
             resp = self.app.get('/updates/%s' % nvr,
                                 headers={'Accept': 'text/html'})
             self.assertIn('text/html', resp.headers['Content-Type'])
             self.assertIn(nvr, resp)
             self.assertNotIn(
-                '<div class="alert alert-danger">The update can not be pushed: '
-                'Because Randy said so', resp)
+                '<div class="alert alert-danger">The update can not be pushed: \' + summary', resp)
 
-    @mock.patch.dict('bodhi.server.config.config', {'test_gating.required': True})
     def test_alert_displayed_if_summary_present(self, publish, *args):
         """ Ensure that the alert is displayed if there is a summary status
         (returned normaly by greenwave) set in the database for this update.
@@ -4619,15 +4619,13 @@ class TestUpdatesService(base.BaseTestCase):
         build.update.greenwave_summary_string = 'Because Randy said so'
         self.assertEqual(build.update.request, UpdateRequest.testing)
 
-        with mock.patch.dict(config, {'test_gating.required': True}):
+        with mock.patch.dict(self.app.app.registry.settings, {'test_gating.required': True}):
             resp = self.app.get('/updates/%s' % nvr,
                                 headers={'Accept': 'text/html'})
             self.assertIn('text/html', resp.headers['Content-Type'])
             self.assertIn(nvr, resp)
-            print(resp)
             self.assertIn(
-                '<div class="alert alert-danger">The update can not be pushed: '
-                'Because Randy said so', resp)
+                '<div class="alert alert-danger">The update can not be pushed: \' + summary', resp)
 
 
 class TestWaiveTestResults(base.BaseTestCase):
```

pypingou:
jenkies test

pypingou:
jenkies test

pypingou:
jenkies test

pypingou:
Finally!!!

bowlofeggs (approved):
Thanks @pypingou!

bowlofeggs:
This patch will be included in the [3.4.1 release](https://github.com/fedora-infra/bodhi/pull/2179).

PR #2129, opened 2018-01-19
PEP-257 bodhi.client.

bowlofeggs:

### DIFF
--- a/bodhi/client/__init__.py
+++ b/bodhi/client/__init__.py
@@ -1,3 +1,8 @@
+# -*- coding: utf-8 -*-
+# Copyright © 2014-2018 Red Hat, Inc. and others.
+#
+# This file is part of Bodhi.
+#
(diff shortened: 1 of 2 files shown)


bodhi/client/__init__.py
```diff
@@ -29,14 +35,14 @@
 
 def _warn_if_url_and_staging_set(ctx, param, value):
     """
-    This function will print a warning to stderr if the user has set both the --url and --staging
-    flags, so that the user is aware that --staging supercedes --url.
+    Print a warning to stderr if the user has set both the --url and --staging flags.
+
+    This ensures that the user is aware that --staging supercedes --url.
```
puiterwijk: supersedes

puiterwijk (approved):

bodhi/client/__init__.py
bowlofeggs: Fixed.

PR #2137, opened 2018-02-01
Insert utcnow for issued_date if not yet pushed

puiterwijk:
This works around a silly bug in Satellite where it really wants
an issued date and will crash if not there.

### DIFF
--- a/bodhi/server/metadata.py
+++ b/bodhi/server/metadata.py
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-# Copyright 2007-2017 Red Hat, Inc. and others.
+# Copyright 2007-2018 Red Hat, Inc. and others.
 #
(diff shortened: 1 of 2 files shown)


pypingou:
Looks fine to me but do we want to add test for this?

bowlofeggs:
@puiterwijk I've added another commit with tests for this change. Care to review them?

bowlofeggs (approved):

puiterwijk:
@bowlofeggs as said on IRC, +1 on your commit. Thanks!

bowlofeggs:
The patch for this bug has been included in the [```3.3.0b0``` build](https://bodhi.stg.fedoraproject.org/docs/user/release_notes.html), which is currently deployed to [staging](https://bodhi.stg.fedoraproject.org).

PR #2140, opened 2018-02-07
Prepare for the 3.3.0 release.

bowlofeggs:


pypingou:
The changes made to the script running as cron to pull the CI status are included, right?

bowlofeggs:
@pypingou Those aren't included since they aren't merged yet. Since they are really just performance related, IMO they should not hold up the 3.3.0 release.

pypingou:
I would really include them, they are hotfixed on the server and not including them would mean reverting these fixes

pypingou:
Note that they aren't just performance related, they also just allow the cron to finish, so it's a little more than just "faster" :)

bowlofeggs:
@pypingou I'm not aware of problems with the cron job failing to finish. Can you elaborate?

One of them has an LGTM if you would like to merge it. The other needs tests.

pypingou:
> @pypingou I'm not aware of problems with the cron job failing to finish. Can you elaborate?

When we turned on gating you told me that the cron job rans for hours and eventually crashed. We have not seen this since it was hotfixed with these two PRs.

bowlofeggs:
@pypingou I don't recall it ever crashing. It completed successfully after about 2 hours, and runs every 6 hours. That's why I don't think it's urgent to include the two patches (though I do of course think we should get tests written for the one and merged). I think it'll be fine if they wait until the next release, just a little inefficient.

bowlofeggs:
I added c8942556f5df47df840a29e18ad8af5926d544d1 to the release notes in this PR.

puiterwijk (approved):

PR #2146, opened 2018-02-09
Make it clearer that some settings don't have defaults.

bowlofeggs:

### DIFF
--- a/production.ini
+++ b/production.ini
@@ -162,10 +162,11 @@ use = egg:bodhi-server
 #     arches listed in {release}_{version}_primary_arches when it is defined, else used for all
 #     arches. You must put two %s's in this setting - the first will be replaced with the release
 #     version and the second will be replaced with the architecture.
-# fedora_stable_master_repomd = http://download01.phx2.fedoraproject.org/pub/fedora/linux/updates/%s/%s/repodata/repomd.xml
-# fedora_testing_master_repomd = http://download01.phx2.fedoraproject.org/pub/fedora/linux/updates/testing/%s/%s/repodata/repomd.xml
-# fedora_epel_stable_master_repomd = http://download01.phx2.fedoraproject.org/pub/epel/%s/%s/repodata/repomd.xml
-# fedora_epel_testing_master_repomd = http://download01.phx2.fedoraproject.org/pub/epel/testing/%s/%s/repodata/repomd.xml
+# examples (these settings do not have defaults):
+#     fedora_stable_master_repomd = http://download01.phx2.fedoraproject.org/pub/fedora/linux/updates/%s/%s/repodata/repomd.xml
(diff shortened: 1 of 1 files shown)


bowlofeggs:
jenkies test

production.ini
```diff
@@ -162,10 +162,11 @@ use = egg:bodhi-server
 #     arches listed in {release}_{version}_primary_arches when it is defined, else used for all
 #     arches. You must put two %s's in this setting - the first will be replaced with the release
 #     version and the second will be replaced with the architecture.
-# fedora_stable_master_repomd = http://download01.phx2.fedoraproject.org/pub/fedora/linux/updates/%s/%s/repodata/repomd.xml
-# fedora_testing_master_repomd = http://download01.phx2.fedoraproject.org/pub/fedora/linux/updates/testing/%s/%s/repodata/repomd.xml
-# fedora_epel_stable_master_repomd = http://download01.phx2.fedoraproject.org/pub/epel/%s/%s/repodata/repomd.xml
-# fedora_epel_testing_master_repomd = http://download01.phx2.fedoraproject.org/pub/epel/testing/%s/%s/repodata/repomd.xml
+# examples (these are not defaults):
```
pypingou: I think I would be even more obvious: ``these are no the defaults values, there are no default values``

pypingou (approved):

production.ini
bowlofeggs: Cool, I'll do this instead:

```
index 3c9d20b7..a45261bd 100644
--- a/production.ini
+++ b/production.ini
@@ -162,7 +162,7 @@ use = egg:bodhi-server
 #     arches listed in {release}_{version}_primary_arches when it is defined, else used for all
 #     arches. You must put two %s's in this setting - the first will be replaced with the release
 #     version and the second will be replaced with the architecture.
-# examples (these are not defaults):
+# examples (these settings do not have defaults):
 #     fedora_stable_master_repomd = http://download01.phx2.fedoraproject.org/pub/fedora/linux/updates/%s/%s/repodata/repomd.xml
 #     fedora_testing_master_repomd = http://download01.phx2.fedoraproject.org/pub/fedora/linux/updates/testing/%s/%s/repodata/repomd.xml
 #     fedora_epel_stable_master_repomd = http://download01.phx2.fedoraproject.org/pub/epel/%s/%s/repodata/repomd.xml
@@ -176,7 +176,7 @@ use = egg:bodhi-server
 #     arches not listed in {release}_{version}_primary_arches if it is defined. You must put two
 #     %s's in this setting - the first will be replaced with the release version and the second will
 #     be replaced with the architecture.
-# examples (these are not defaults):
+# examples (these settings do not have defaults):
 #     fedora_stable_alt_master_repomd = http://download01.phx2.fedoraproject.org/pub/fedora-secondary/updates/%s/%s/repodata/repomd.xml
 #     fedora_testing_alt_master_repomd = http://download01.phx2.fedoraproject.org/pub/fedora-secondary/updates/testing/%s/%s/repodata/repomd.xml
```

bowlofeggs:
jenkies test

PR #2149, opened 2018-02-13
Change the logic of --close-bugs flag.

### LINKED ISSUE #1818: The CLI --close-bugs flag does not work
I noticed today that two updates I created with the ```--close-bugs``` CLI flag did not close the bugs when they went to stable. For example, I used this command to create an update recently (according to bash history):

```
$ bodhi updates new --type bugfix --notes "Bodhi no longer depends on python-dnf." --bugs 1492610 --close-bugs --autokarma bodhi-2.10.1-2.el7
```

If you look at https://bugzilla.redhat.com/show_bug.cgi?id=1488223 you can see that the bug wasn't closed when the update went stable.

cverna:
This commit makes using the --close-bugs flag set
the boolean to True instead of False previously

Fixes #1818

### DIFF
--- a/docs/user/release_notes.rst
+++ b/docs/user/release_notes.rst
@@ -2,6 +2,18 @@
 Release notes
 =============
 
+Bugs
+^^^^
+
+* The CLI --close-bugs flag does not work (:issue:`1818`).
+
+
+Contributors
+^^^^^^^^^^^^
+
+* Clément Verna
+
+
 v3.4.0
 ------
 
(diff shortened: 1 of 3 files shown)


bowlofeggs:
@centos-ci add to whitelist

docs/user/release_notes.rst
```diff
@@ -50,6 +51,7 @@ The following developers contributed to Bodhi 3.3.0:
 * Till Maas
 * Ken Dreyer
 * Randy Barlow
+* Clément Verna
```
bowlofeggs: The 3.3.0 release is already out the door and into production - can you move the two lines you changed here into a new ```develop``` heading at the top?

bowlofeggs (changes requested):
Thanks so much for the patch! We just need a change to the release notes and this should be good to go (assuming CI also passes - it's running now).

bowlofeggs:
jenkies test

cverna:
I have updated the release notes and rebased the PR.

bowlofeggs (approved):
Thanks!

Issue #2151, opened 2018-02-14
Rename "masher" to "composer"


bowlofeggs:
With the next major version of Bodhi, we should rename the "masher" to "composer" since mash is no longer used.

sebwoj:
@bowlofeggs Since github has an issue closing "Bodhi issues", we should do it manualy ;/

bowlofeggs:
This change is planned for inclusion in the upcoming 4.0.0 release: #3221

bowlofeggs:
Bodhi 4.0.0 beta is built and deployed to staging:

https://copr.fedorainfracloud.org/coprs/bowlofeggs/bodhi-pre-release
https://bodhi.stg.fedoraproject.org/

PR #2152, opened 2018-02-14
Refactor the composer to split pungi code into a new base class.

### LINKED ISSUE #2028: Add support for containers to the masher
We need the masher to be able to promote containers into testing and stable. We will probably use skopeo for this.

Depends on #2027.

bowlofeggs:
In order to prepare the way for Bodhi to support composing
containers we need to split pungi-specific code out of the
MasherThread base class so the general composition code can be
reused by the soon-to-be-written container composing class.

This commit does that refactoring work, and renames some uses of
"mash" to "compose". Not all uses are renamed here, as some of them
would not be backwards compatible (see #2151).

re #2028

### DIFF
--- a/bodhi/server/consumers/masher.py
+++ b/bodhi/server/consumers/masher.py
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-# Copyright © 2007-2017 Red Hat, Inc. and others.
+# Copyright © 2007-2018 Red Hat, Inc. and others.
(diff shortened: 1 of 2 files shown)


puiterwijk (approved):

bowlofeggs:
This patch will be included in the [3.4.1 release](https://github.com/fedora-infra/bodhi/pull/2179).

PR #2153, opened 2018-02-15
Do not print traceback in updates-new command

tkdchen:
Just like other commands, do not catch and print traceback in method
new. Instead, let handle_errors decorator to handle exceptions raised.

The updated handle_errors ensure that 1 will be returned for AuthError,
and 2 will be returned for any other errors.

Major changes in this patch

* Exception handler of bindings.BodhiClientException is droppped.
  Handling Exception covers that.
* Handle EOFError that could be raised if user press Ctrl+D to
  stop entering password. bodhi just terminates and keeps quiet without
  any message output.
* Add __str__ to UpdateNotFound.
* Do not raise click.BadParameter from method request in order to ensure
  handle_errors is able to handle any type of errors raised from
  request. Meanwhile, "an update is not found" is not proper for a bad
  parameter error.
* Tests are updated.

Signed-off-by: Chenxiong Qi <cqi@redhat.com>

### DIFF
--- a/bodhi/client/__init__.py
+++ b/bodhi/client/__init__.py
@@ -22,12 +22,10 @@ import os
 import platform
 import subprocess
 import sys
-import traceback
 import re
 import functools
 
 import click
-import six
 
 from bodhi.client import bindings
 from fedora.client import AuthError
@@ -152,7 +150,12 @@ def handle_errors(method):
         except AuthError as e:
             click.secho("%s: Check your FAS username & password" % (e), fg='red', bold=True)
             sys.exit(1)
-        except bindings.BodhiClientException as e:
+        except EOFError as e:
+            # This error could be raised if user presses Ctrl+D while entering
+            # password. Treat this event as a normal thing that user wants
+            # instead of error. So, just ignore raised error and keep quiet.
+            pass
+        except Exception as e:
             click.secho(str(e), fg='red', bold=True)
             sys.exit(2)
     return wrapper
@@ -268,13 +271,8 @@ def new(user, password, url, **kwargs):
     kwargs['notes'] = _get_notes(**kwargs)
 
     for update in updates:
-        try:
-            resp = client.save(**update)
-            print_resp(resp, client)
-        except bindings.BodhiClientException as e:
-            click.echo(str(e))
-        except Exception as e:
-            traceback.print_exc()
+        resp = client.save(**update)
+        print_resp(resp, client)
 
 
 def _validate_edit_update(ctx, param, value):
@@ -440,11 +438,7 @@ def request(update, state, user, password, url, **kwargs):
     client = bindings.BodhiClient(base_url=url, username=user, password=password,
                                   staging=kwargs['staging'])
 
-    try:
-        resp = client.request(update, state)
-    except bindings.UpdateNotFound as exc:
-        raise click.BadParameter(six.text_type(exc), param_hint='UPDATE')
-
+    resp = client.request(update, state)
     print_resp(resp, client)
 
 
--- a/bodhi/tests/client/test___init__.py
+++ b/bodhi/tests/client/test___init__.py
@@ -377,7 +377,7 @@ class TestNew(unittest.TestCase):
             client.new,
             ['--user', 'bowlofeggs', '--password', 's3kr3t', '--autokarma', 'bodhi-2.2.4-1.el7'])
 
-        self.assertEqual(result.exit_code, 0)
+        self.assertEqual(result.exit_code, 2)
         self.assertIn("This is a BodhiClientException message", result.output)
 
     @mock.patch('bodhi.client.bindings.BodhiClient.csrf',
@@ -395,9 +395,22 @@ class TestNew(unittest.TestCase):
             client.new,
             ['--user', 'bowlofeggs', '--password', 's3kr3t', '--autokarma', 'bodhi-2.2.4-1.el7'])
 
+        self.assertEqual(result.exit_code, 2)
+        self.assertIn("This is an Exception message", result.output)
+
+    @mock.patch('bodhi.client.bindings.getpass.getpass', side_effect=EOFError)
+    def test_keep_quiet_if_ctrld_ends_password_input(self, getpass):
+        """
+        Assert keeping quiet if user press Ctrl+D to end password input
+        """
+        runner = testing.CliRunner()
+
+        result = runner.invoke(
+            client.new,
+            ['--user', 'bowlofeggs', '--autokarma', 'bodhi-2.2.4-1.el7'])
+
         self.assertEqual(result.exit_code, 0)
-        self.assertIn("Traceback (most recent call last):", result.output)
-        self.assertIn("Exception: This is an Exception message", result.output)
+        self.assertEqual('', result.output.strip())
 
     @mock.patch('bodhi.client.bindings.BodhiClient.csrf',
                 mock.MagicMock(return_value='a_csrf_token'))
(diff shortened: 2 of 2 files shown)


tkdchen:
Hi, what's wrong with the checks? That haven't completed yet for 7 days.

bowlofeggs:
@centos-ci add to whitelist

bodhi/client/__init__.py
```diff
@@ -150,7 +148,12 @@ def wrapper(*args, **kwargs):
...
             sys.exit(1)
-        except bindings.BodhiClientException as e:
+        except EOFError as e:
+            # This error could be raised if user presses Ctrl+D while entering
+            # password. Treat this event as a normal thing that user wants
+            # instead of error. So, just ignore raised error and keep quiet.
+            pass
+        except Exception as e:
             click.secho(str(e), fg='red', bold=True)
             sys.exit(2)
```
bowlofeggs, line 160: I don't think we want to do this for general ```Exception```s, because these would be unexpected errors and we won't have the traceback. The ```BodhiClientException``` is raised under specific conditions, for which we shouldn't need the traceback (i.e., expected errors). This error handler is designed to catch specific "normal" errors that are expected to happen from time to time and inform the user in a useful manner. Anything that isn't a normal error that a user might hit is something that should be considered a bug in Bodhi, and we would like to have the traceback when that happens.

bodhi/client/__init__.py
```diff
@@ -150,7 +148,12 @@ def wrapper(*args, **kwargs):
         except AuthError as e:
             click.secho("%s: Check your FAS username & password" % (e), fg='red', bold=True)
             sys.exit(1)
-        except bindings.BodhiClientException as e:
+        except EOFError as e:
+            # This error could be raised if user presses Ctrl+D while entering
+            # password. Treat this event as a normal thing that user wants
+            # instead of error. So, just ignore raised error and keep quiet.
+            pass
```
bowlofeggs, line 157: When would a user want to use Ctrl+D while typing a password?

bowlofeggs (changes requested):

bodhi/client/bindings.py
```diff
@@ -71,6 +71,9 @@ def __init__(self, update):
         """Initialize the Exception."""
         self.update = six.text_type(update)
 
+    def __str__(self):
+        return str(self.__unicode__())
```
bowlofeggs: ```pydocstyle``` wants this method to have a docblock:

```
rawhide	bodhi/client/bindings.py:74 in public method `__str__`:
rawhide	        D105: Missing docstring in magic method
```

I suggest something like this:

```
"""
Return a string representation of the error message.

Returns:
    str: A string representation of the error message.
"""
```

bowlofeggs (changes requested):

bodhi/client/bindings.py
frenzymadness: I am not sure that this will work in Py2.7 when `self.__unicode__()` returns a true unicode with more than just ascii chars.

I would prefer providing the same method `self.__unicode__()` under both names: `__unicode__` and `__str__`. This makes no difference in Python 3 and in Python 2.7, we don't need exception text in bytes so I think that's ok to return unicode object from `__str__` method.

What do you think?

bodhi/client/__init__.py
tkdchen: I actually don't have such a user case. A bug was reported to fedpkg that a user would press Ctrl+D to terminate entering password.
FYI: https://bugzilla.redhat.com/show_bug.cgi?id=1505326

bodhi/client/bindings.py
tkdchen: > I am not sure that this will work in Py2.7 when self.__unicode__() returns a true unicode with more than just ascii chars.

@frenzymadness I've no idea what problem would happen. I tried to recall my memory and I haven't remember a failure case related to issue.

> providing the same method self.__unicode__() under both names

Sorry, I don't understand what it means. Can you give an example?

frenzymadness: > Sorry, I don't understand what it means. Can you give an example?

I mean something like `__str__ = __unicode__` which do basically the same as your code but without converting the result of `__unicode__` back into `str` which does nothing in Python 3 and can cause problems in Python 2.

bowlofeggs:
Do you still want to work on this patch? It's gone a while without activity and is conflicting now.

bodhi/client/__init__.py
tkdchen: Hi @bowlofeggs Thanks for reminding. I would like to continue work on this. What do you think about this use case?

bowlofeggs:
Closing due to inactivity.

tkdchen:
> Hi @bowlofeggs Thanks for reminding. I would like to continue work on this. What do you think about this use case?

Hi @bowlofeggs I have commented out this above. Not sure if you noticed it.

bodhi/client/__init__.py
bowlofeggs: Hey @tkdchen!

Sorry I had missed your question earlier. It wouldn't occur to me to use CTRL-D to abort a password - I think I'd probably CTRL-C instead. However, I don't see any harm in it, so sure we can keep this if you want.

tkdchen:
@bowlofeggs I rebased my patch. Jenkins job fails and report message `ERROR: Step ‘Publish JUnit test result report’ failed: No test report files were found. Configuration error?`. Anything wrong with the configuration?

Anything else I need to update my patch?

bowlofeggs:
@tkdchen When looking at the Jenkins console output, search for the string "JENKIES FAIL" and you'll see the actual error:

```
pip	./bodhi/client/bindings.py:87:5: F811 redefinition of unused '__str__' from line 74
pip	    __str__ = __unicode__
pip	    ^
pip	JENKIES FAIL!
```

bodhi/client/bindings.py
```diff
@@ -71,6 +71,9 @@ def __init__(self, update):
         """Initialize the Exception."""
         self.update = six.text_type(update)
 
+    def __str__(self):
```
bowlofeggs: The failure might be due to @frenzymadness having added the same name in his patch, so now it's defined twice.

tkdchen:
@bowlofeggs Hi, fixed that duplicate`__str__`.

tkdchen:
@bowlofeggs ping for review.

bodhi/client/__init__.py
bowlofeggs: This comment still needs to be addressed.

bodhi/client/__init__.py
```diff
@@ -440,11 +438,7 @@ def request(update, state, user, password, url, **kwargs):
     client = bindings.BodhiClient(base_url=url, username=user, password=password,
                                   staging=kwargs['staging'])
 
-    try:
-        resp = client.request(update, state)
-    except bindings.UpdateNotFound as exc:
-        raise click.BadParameter(six.text_type(exc), param_hint='UPDATE')
-
+    resp = client.request(update, state)
```
bowlofeggs, line 441: I think we want to keep this try/except the way it was, because it lets us give a nice hint to the user that is specific to this command.

bowlofeggs (changes requested):

bowlofeggs:
Closing due to inactivity.

tkdchen:
@bowlofeggs  Sorry for late response. It is fine to close this issue now. I've rewrite fedpkg update command with Python bindings recently, which could give me more control of potential user behavior while creating update. Thanks for taking your time in the past to review this patch.

PR #2155, opened 2018-02-15
Link updates to package packages

abitrolly:
This adds a link to https://bodhi.fedoraproject.org/updates/?packages=qesteidutil here.

image

Signed-off-by: Anatoli Babenia <anatoli@rainforce.org>

### DIFF
--- a/bodhi/server/templates/update.html
+++ b/bodhi/server/templates/update.html
@@ -628,11 +628,9 @@ $(document).ready(function(){
     <h4 id='shortdesc'>
       ${update.type} update
       in ${update.release.long_name} for
-      <a href="${request.route_url('updates') + '?packages=' + build.package.name}">
-        ${update.beautify_title(amp=True) | n}
-      </a>
-      
-      
+      % for build in update.builds:
+        <a href="https://apps.fedoraproject.org/packages/{$build.package.name}">${build.package.name}</a>${'' if loop.last else ','}
+      % endfor
     </h4>
     <div>
       Status: ${self.util.status2html(update.status) | n}
--- a/bodhi/server/templates/update.html
+++ b/bodhi/server/templates/update.html
@@ -625,10 +625,14 @@ $(document).ready(function(){
       % endif
 
     </h1>
-    <h4>
+    <h4 id='shortdesc'>
       ${update.type} update
-      in ${update.release.long_name}
-      for ${update.beautify_title(amp=True) | n}
+      in ${update.release.long_name} for
+      <a href="${request.route_url('updates') + '?packages=' + build.package.name}">
+        ${update.beautify_title(amp=True) | n}
+      </a>
+      
+      
     </h4>
     <div>
       Status: ${self.util.status2html(update.status) | n}


bowlofeggs:
jenkies test

bodhi/server/templates/update.html
```diff
@@ -575,10 +575,14 @@ <h1>
...
 
     </h1>
-    <h4>
+    <h4 id='shortdesc'>
       ${update.type} update
-      in ${update.release.long_name}
-      for ${update.beautify_title(amp=True) | n}
+      in ${update.release.long_name} for
+      <a href="${request.route_url('updates') + '?packages=' + build.package.name}">
+        ${update.beautify_title(amp=True) | n}
```
bowlofeggs: There isn't a variable called ```build``` in this scope, which leads to a traceback:

```
  File "/usr/lib/python2.7/site-packages/pyramid_mako/__init__.py", line 148, in __call__ 
    result = template.render_unicode(**system) 
  File "/usr/lib/python2.7/site-packages/mako/template.py", line 471, in render_unicode 
    as_unicode=True) 
  File "/usr/lib/python2.7/site-packages/mako/runtime.py", line 838, in _render 
    **_kwargs_for_callable(callable_, data)) 
  File "/usr/lib/python2.7/site-packages/mako/runtime.py", line 873, in _render_context 
    _exec_template(inherit, lclcontext, args=args, kwargs=kwargs) 
  File "/usr/lib/python2.7/site-packages/mako/runtime.py", line 899, in _exec_template 
    callable_(context, *args, **kwargs) 
  File "/home/vagrant/bodhi/bodhi/server/templates/master.html", line 188, in render_body 
    ${self.body()} 
  File "/home/vagrant/bodhi/bodhi/server/templates/update.html", line 581, in render_body 
    <a href="${request.route_url('updates') + '?packages=' + build.package.name}"> 
UnboundLocalError: local variable 'build' referenced before assignment
```

Though many updates only reference a single build, updates can contain multiple builds.

The Builds tab already contains this for each of the builds. Does that not satisfy what you are trying to achieve?

bowlofeggs:
I'm going to close this due to inactivity.

bodhi/server/templates/update.html
abitrolly: Builds tab redirects to https://koji.fedoraproject.org/koji/buildinfo?buildID=1044924

How to see which variables exist at the given scope?

abitrolly:
So, what is the proper way to inspect variables at a given scope?

bowlofeggs:
Hey @abitrolly!

I apologize - I somehow missed your question.

I'm not actually all that familiar with mako, which is the library we use to do our templates in Bodhi, so I don't know the answer to your question in general, but I know that you can inject Python code into it, so I'd think you could do something like run a ```dir()``` to find out what's there. You can also add things you need to the scope. As an example:

https://github.com/fedora-infra/bodhi/blob/3.12.0/bodhi/server/templates/update.html#L5-L8

I'll reopen this in case you want to see it through. Thanks for your patience!

abitrolly:
No problem. It happens sometime. I will try to fix this PR from GitHub. No time to deploy development environment.

https://github.com/fedora-infra/bodhi/blob/ee1f4fcfee733486143823c70022d04b3854fd00/bodhi/server/templates/update.html#L5-L8

I would prefer Jinja2 filters that hide such Python differences. Should be faster, though.

abitrolly:
@bowlofeggs linking to https://apps.fedoraproject.org/packages/qesteidutil is more useful.

image

bodhi/server/templates/update.html
```diff
@@ -575,10 +575,12 @@ <h1>
...
 
     </h1>
-    <h4>
+    <h4 id='shortdesc'>
       ${update.type} update
-      in ${update.release.long_name}
-      for ${update.beautify_title(amp=True) | n}
+      in ${update.release.long_name} for
+      % for build in update.builds:
+        <a href="https://apps.fedoraproject.org/packages/{$build.package.name}">${build.package.name}</a>${'' if loop.last else ','}
```
bowlofeggs, line 632: We don't want to hardcode any URLs into Bodhi - this will cause issues for deployments that are not in Fedora, and will also be incorrect for our staging deployment.

bowlofeggs:
You will need to rebase this pull request on latest develop to get it to run the tests. It's based on a commit that's more than a year old right now.

bowlofeggs (changes requested):

bodhi/server/templates/update.html
abitrolly: It is an external link. Where do you propose to store it?

bowlofeggs: We should put it into Bodhi's settings. To do this, you add the setting to ```production.ini```, with the default commented out and some text over it explaining what the setting does. There is a proposal from @cverna to remove the Packages app, so I think we probably want the default to be the empty string and have this code only show the link if the setting is truthy.

The second place the setting needs to be added is ```bodhi.server.config```, and this is where the default is defined.

abitrolly: No time for investigation. How to reference config from template properly?

bowlofeggs: Here's an example:

https://github.com/fedora-infra/bodhi/blob/3.13.3/bodhi/server/templates/update.html#L54

bowlofeggs:
Closing due to inactivity.

PR #2166, opened 2018-02-21
Add an API for retrieving information about Composes.

bowlofeggs:
re #2015

The docs won't build for this PR until #2173 is merged, because the docs are upset that there are two classes named ```Compose```. I could make all docblock use the fully qualified name, but it's easier to just resolve the conflict in #2173, and the name there made more sense to be plural anyway.


puiterwijk (approved):
Cool! Annoying about cornice, but this works for now.

PR #2167, opened 2018-02-22
Use ':' as separator for name:stream:version in pungi variants.xml when mashing modules.

jankaluza:
Signed-off-by: Jan Kaluza <jkaluza@redhat.com>

### DIFF
--- a/bodhi/server/consumers/masher.py
+++ b/bodhi/server/consumers/masher.py
@@ -1293,10 +1293,15 @@ class ModuleComposerThread(PungiComposerThread):
         """
         template = template_env.get_template('variants.module.xml.j2')
 
-        module_list = self._generate_module_list()
+        # These are assigned to self to be testable
+        self._module_defs = self._generate_module_list()
+        # This is so as to not break existing Pungi configurations, but ideally they get updated.
+        self._module_list = ['%(name)s:%(stream)s:%(version)s' % mod for mod in self._module_defs]
 
         with open(os.path.join(pungi_conf_dir, 'module-variants.xml'), 'w') as variantsfile:
-            variantsfile.write(template.render(modules=module_list))
+            self._variants_file = template.render(modules=self._module_list,
+                                                  moduledefs=self._module_defs)
+            variantsfile.write(self._variants_file)
 
     def generate_testing_digest(self):
         """Temporarily disable testing digests for modules.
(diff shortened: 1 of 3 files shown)


puiterwijk:
@centos-ci add to whitelist

bodhi/server/consumers/masher.py
```diff
@@ -1289,10 +1320,18 @@ def _generate_module_list(self):
...
+            for update in self.compose.updates:
+                for build in update.builds:
+                    koji.getBuild(build.nvr)
+            results = koji.multiCall()
+            for result, build in zip(results, update.builds):
+                self._raise_on_get_build_multicall_error(result, build)
+                koji_build = result[0]
+                # name:stream:version maps to Koji's name-version-release.
+                ns = "%s:%s" % (koji_build["name"], koji_build["version"])
+                version = koji_build["release"]
```
bowlofeggs: This code block is almost identical to the code block above. Let's replace it with a method so we can adjust them together in the future if needed.

bowlofeggs (changes requested):
There are tests that assert that exceptions are raised by ```_raise_on_get_build_multicall_error```, but we will also need tests that assert that the correct data is returned (with :'s) from ```_generate_module_list``` when there are not koji errors.

bodhi/server/consumers/masher.py
```diff
@@ -1284,12 +1289,43 @@ def _raise_on_get_build_multicall_error(self, result, build):
             self.log.error(err)
             raise Exception(err)
 
+    def _add_build_to_newest_builds(self, newest_builds, koji_build, override=False):
+        """
+        Add Koji build to newest_builds dict if it's newer than the one there or override is set.
+
+        Args:
+            newest_builds (dict): Dict with name:stream as a key and version as a value.
```
bowlofeggs: It looks like the value is actually a dictionary with the key "version" (perhaps among others?) based on reading the code below:

```
curr_version = newest_builds[ns]['version']
```

bowlofeggs (commented):
Just a nitpick on a docblock.

I'm running the tests now on my laptop.

bowlofeggs (approved):
Alright, I've run CI for this on my laptop and it passes, so I think we are good to merge. Please correct that one docblock I mentioned earlier, but no need for re-review.

Issue #1349, opened 2017-03-15
Don't reopen BZ review ticket if "Close bugs on stable" is disabled


morxa:
As the title says, I would expect bodhi to change the ticket status only if "Close on stable" is selected for the update.

Currently, bodhi always adds comments to the tickets and changes the status to ON_QA, even if "Close on stable" is disabled and the ticket is already closed.

As an example, I recently submitted python-pyswip. I closed the [BZ review request ticket](https://bugzilla.redhat.com/show_bug.cgi?id=1413374) after building the package for rawhide. For the bodhi updates (e.g., [F24](https://bodhi.fedoraproject.org/updates/FEDORA-2017-cd4110b464)) I added the ticket number for completeness sake, but I disabled "Push on stable". After the package has been pushed to testing, the ticket was reopened and the status set to ON_QA.

bowlofeggs:
I think the language "Close on stable" might only relate to changing the status to CLOSED, and doesn't imply that Bodhi wouldn't (or would) make other state changes on tickets. At least, that's how I've always interpreted it. In other words, the behavior you described is what I would expect that boolean to do based on the language it uses.

It sounds like you are asking for an RFE to have Bodhi make no state changes at all on bugs. If we were to do that I would probably want to adjust the boolean to say something else, like "modify bug states" or similar.

morxa:
I agree that this is an RFE. Still, I think the current behavior is not consistent: In the example above, the bug was re-opened, but after submitting the package to stable, it is still ON_CA.

Maybe the issue is really that the current state is not checked (at least in this case), and the bug is set to ON_QA no matter what. I think the ticket shouldn't be re-opened even if "Close on stable" is enabled.

bowlofeggs:
@morxa That's a bit similar to the request from @adamwill in #1091 - he similarly didn't want Bodhi to "downgrade" the state of a bug.

AdamWill:
@bowlofeggs I'm pretty sure that in the past 'close on stable' has always meant 'don't make any status changes'. When it was unselected, Bodhi would add comments when the update was submitted and pushed, but that's *all* it would do. It would never make any state change. If it's been changed now and it makes state changes but just never makes a change to `CLOSED`, that's a significant behaviour change that people will *not* be expecting, and I'd suggest it should be reverted.

[...]

PR #2168, opened 2018-02-22
Don't downgrade bug status from VERIFIED or CLOSED to ON_QA

### LINKED ISSUE #1091: Bodhi should not erroneously 'downgrade' bug status from VERIFIED
We frequently play a game of pong with Bodhi, where we verify a fix when a bug is in `MODIFIED` state (i.e. not yet pushed to u-t) and change the state to `VERIFIED`, only for Bodhi to come along and pong the state back to `ON_QA` when it pushes the update. For e.g., every single damn F25 blocker fix:

https://bugzilla.redhat.com/show_bug.cgi?id=1393110
https://bugzilla.redhat.com/show_bug.cgi?id=1376471
https://bugzilla.redhat.com/show_bug.cgi?id=1390607
https://bugzilla.redhat.com/show_bug.cgi?id=1378156
https://bugzilla.redhat.com/show_bug.cgi?id=1392654
https://bugzilla.redhat.com/show_bug.cgi?id=1393109

when Bodhi is only doing the `ON_QA` change for an update being pushed, whose contents have not changed since it was submitted, it should avoid changing the bug's status if it's `VERIFIED`.

mattiaverga:
Fixes #1091 #1349 

This patch checks if a bug is already in VERIFIED or CLOSED status and avoids to downgrade it to ON_QA.
Also, if the bug is already in ON_QA it just add a new comment without set again the ON_QA state (for example when an update on multiple Fedora branches is linked to a bug)

### DIFF
--- a/bodhi/server/bugs.py
+++ b/bodhi/server/bugs.py
@@ -165,7 +165,7 @@ class Bugzilla(BugTracker):
 
     def on_qa(self, bug_id, comment):
         """
-        Change the status of this bug to ON_QA.
+        Change the status of this bug to ON_QA if it is not already ON_QA, VERIFIED, or CLOSED.
 
         This will also comment on the bug with some details on how to test and provide feedback for
         this update.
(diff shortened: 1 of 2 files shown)


bowlofeggs (changes requested):
The code change looks good, but let's add some tests to ensure it operates properly. There is a test module for this code at ```bodhi/tests/server/test_bugs.py```. Thanks!

bowlofeggs:
jenkies test

mattiaverga:
I've added a test case when a bug is already in 'CLOSED' status. Should I add another for the 'VERIFIED' status?

bowlofeggs:
On 02/26/2018 02:23 PM, Mattia Verga wrote:
> I've added a test case when a bug is already in 'CLOSED' status. Should
> I add another for the 'VERIFIED' status?

Yeah, I would also add one for the ```ON_QA``` status, so that you hit
all four of the branches in the if statement (there are already tests
for the fourth condition where we *do* want to change the status).

bowlofeggs:
jenkies test

bowlofeggs:
I've observed this CI failure a few times, and am not sure what causes it, but usually re-running it works.

bowlofeggs:
jenkies test

bowlofeggs (approved):
Thanks so much!

bowlofeggs:
jenkies test

PR #2173, opened 2018-02-23
Rename the Compose resource to Composes.

bowlofeggs:
It bugged me a bit that the resource was singular rather than
plural, so this commit adjusts it. This should not be noticeable
to users.


bowlofeggs:
jenkies test

cverna (approved):

PR #2181, opened 2018-02-27
Drop some stay code that breaks the update JS.

### LINKED ISSUE #2180: The edit/push/unpush buttons do not work
Due to some [extraneous code in the update template](https://github.com/fedora-infra/bodhi/blob/c07daf9/bodhi/server/templates/update.html#L84-L87), the Update page JavaScript fails to execute. This causes the edit/push/unpush buttons not to work and the test results not to load. It may cause other problems as well.

The fix is simple - drop the code.

bowlofeggs:
fixes #2180

### DIFF
--- a/bodhi/server/templates/update.html
+++ b/bodhi/server/templates/update.html
@@ -81,10 +81,6 @@ ${update.type} update for ${update.beautify_title(amp=True) | n}
         });
     };
 
-    // data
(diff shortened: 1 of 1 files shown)


pypingou (approved):
Oups, sorry about that :(

:+1:

PR #2184, opened 2018-03-01
Put the shelve in a cache_dir

puiterwijk:

### DIFF
--- a/bodhi/server/config.py
+++ b/bodhi/server/config.py
@@ -328,6 +328,9 @@ class BodhiConfig(dict):
         'bz_server': {
             'value': 'https://bugzilla.redhat.com/xmlrpc.cgi',
             'validator': six.text_type},
+        'cache_dir': {
+            'value': None,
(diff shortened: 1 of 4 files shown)


bodhi/server/config.py
```diff
@@ -328,6 +328,9 @@ class BodhiConfig(dict):
         'bz_server': {
             'value': 'https://bugzilla.redhat.com/xmlrpc.cgi',
             'validator': six.text_type},
+        'cache_dir': {
+            'value': None,
+            'validator': _validate_none_or(_validate_path)},
```
bowlofeggs, line 333: Can we also add this to ```production.ini``` so it's documented? ```production.ini``` actually automatically ends up in the Sphinx docs these days, so it's a handy way to make sure admin users know about it.

bowlofeggs (approved):
Please add the new setting to ```production.ini``` before merging - no need for re-review. Thanks!

PR #2187, opened 2018-03-02
Update name of documentation to include Randy

puiterwijk:
This should update things like the title page on the PDF docs: https://media.readthedocs.org/pdf/bodhi/2.3/bodhi.pdf


bowlofeggs:
jenkies test

PR #2188, opened 2018-03-02
Tests, bugfix, function removal, and RFE for bodhi.server.mail

bowlofeggs:
This pull request has four commits that build upon each other but are separate.

* Correct a comparison operator in RpmBuild.get_latest().
* Bring bodhi.server.mail to 100% test coverage.
* Instruct Fedora users to use dnf's --advisory flag in notices.
* Remove an unused and broken function.

### DIFF
--- a/bodhi/server/buildsys.py
+++ b/bodhi/server/buildsys.py
@@ -443,7 +443,7 @@ class DevBuildsys(Buildsystem):
         elif rpmID == 'do-not-find-anything.src':
             return None
         else:
-            return {
+            headers = {
                 'description': (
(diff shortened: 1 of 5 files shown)


bowlofeggs:
Seems like the Rawhide GPG key might not be set up correctly in the container, which is causing the CI to fail this build.

puiterwijk (approved):

bowlofeggs:
jenkies test

bowlofeggs:
It's weird that this one failed CI. Other ones today built from Rawhide just fine, but this one had the GPG issue again. Whaaa?

bowlofeggs:
jenkies test

PR #2193, opened 2018-03-06
Tests bodhi.server.validators

bowlofeggs:
This pull requests contains three commits that build upon each other. The first removes unreachable code, the second corrects an incompatibility with Cornice 3, and the last adds 100% coverage to the ```validate_captcha()``` function (including the changed code from the second commit).


bowlofeggs:
The tests are failing due to https://pagure.io/releng/issue/7377

puiterwijk (approved):

bowlofeggs:
jenkies test

PR #2179, opened 2018-02-27
Prepare the 3.5.0 release.

bowlofeggs:
Signed-off-by: Randy Barlow <randy@electronsweatshop.com>

PR #2196, opened 2018-03-06
bodhi-push no longer authenticates to Koji.

### LINKED ISSUE #2190: bodhi-push overrides the kerberos ticket cache when run as root
```bodhi-push``` is sometimes run as root instead of as ```apache```. When this happens, it overwrites ```/tmp/krb5cc_%{uid}``` and makes it only readable by root:

```
-rw-------. 1 root   root   1875 Mar  5 17:53 /tmp/krb5cc_%{uid}
```

Once this state is achieved, the composer fails because it cannot use the ticket cache (since the composer would typically be run as the ```apache``` user.

@puiterwijk pointed out that we don't really need ```bodhi-push``` to use Koji authenticated since it is only performing read queries, and suggested adding a new boolean parameter to ```setup_buildsys()``` to specify whether an authenticated session is needed. ```bodhi-push``` would then use an unauthenticated Koji session, which would then make things OK if a release engineer forgot to run ```bodhi-push``` as apache instead of as root.

bowlofeggs:
Sometimes users run bodhi-push as root. Typically, the backend
fedmsg-hub process is run as the apache user. When users ran
bodhi-push as root, this would cause the Kerberos ticket cache to
become root owned with permissions of 0600, which would prevent
the composer process from authenticating to Koji, causing the
composes to fail.

This commit adjusts the buildsys code to allow callers to request
whether they want an authenticated Koji session or not, and adjusts
bodhi-push to request an unauthenticated session.

fixes #2190

### DIFF
--- a/bodhi/server/buildsys.py
+++ b/bodhi/server/buildsys.py
@@ -473,12 +473,13 @@ class DevBuildsys(Buildsystem):
             }
 
 
-def koji_login(config):
+def koji_login(config, authenticate):
     """
     Login to Koji and return the session.
 
     Args:
         config (bodhi.server.config.BodhiConfig): Bodhi's configuration dictionary.
+        authenticate (bool): If True, establish an authenticated client session.
     Returns:
         koji.ClientSession: An authenticated Koji ClientSession that is ready to use.
     """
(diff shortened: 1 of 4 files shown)


bowlofeggs:
The CI test failures are due to https://pagure.io/releng/issue/7377.

puiterwijk (approved):

bowlofeggs:
Since CI is broken I am going to run these tests on my own machine, leaving off rawhide. If they pass I will merge (since this is needed for the 3.5.0 release).

bowlofeggs:
This patch will be included in 3.5.0: https://github.com/fedora-infra/bodhi/pull/2179

bowlofeggs:
The tests pass on f26, f27, and pip locally, so I am going to merge this now.

PR #2199, opened 2018-03-07
Allow version-specific repomd url override

puiterwijk:

### DIFF
--- a/bodhi/server/consumers/masher.py
+++ b/bodhi/server/consumers/masher.py
@@ -1005,24 +1005,30 @@ class PungiComposerThread(ComposerThread):
             basestring: A URL on the master mirror where the repomd.xml file should be synchronized.
         """
         release = self.compose.release.id_prefix.lower().replace('-', '_')
+        version = self.compose.release.version
         request = self.compose.request.value
 
+        # First check to see if there's an override for the current version, if not, fall back.
+        # This will first try fedora_28_stable_(suffix), and then fedora_stable_(suffix).
+        key_prefixes = ['%s_%s_%s' % (release, version, request),
+                        '%s_%s' % (release, request)]
         # If the release has primary_arches defined in the config, we need to consider whether to
         # use the release's *alt_master_repomd setting.
         primary_arches = config.get(
             '{release}_{version}_primary_arches'.format(
                 release=release, version=self.compose.release.version))
         if primary_arches and arch not in primary_arches.split():
-            key = '%s_%s_alt_master_repomd'
+            suffix = '_alt_master_repomd'
         else:
-            key = '%s_%s_master_repomd'
-        key = key % (release, request)
+            suffix = '_master_repomd'
 
-        master_repomd = config.get(key)
-        if not master_repomd:
-            raise ValueError("Could not find %s in the config file" % key)
+        keys = [key_prefix + suffix for key_prefix in key_prefixes]
 
-        return master_repomd % (self.compose.release.version, arch)
+        for key in keys:
+            val = config.get(key)
+            if val:
+                return val % (version, arch)
+        raise ValueError("Could not find any of %s in the config file" % ','.join(keys))
 
     def _punge(self):
         """
--- a/production.ini
+++ b/production.ini
@@ -163,7 +163,7 @@ use = egg:bodhi-server
 # file_url: Used in the repo metadata to set RPM URLs.
 # file_url = https://download.fedoraproject.org/pub/fedora/linux/updates
 
-# {release}_{request}_master_repomd: This is used by the masher to determine when a
+# {release}_({version}_){request}_master_repomd: This is used by the masher to determine when a
 #     primary architecture push has been synchronized to the master mirror for a given release and
 #     request. The masher will verify that the checksum of repomd.xml at the master URL matches the
 #     expected value, and will poll the URL until this test passes. Substitute release and request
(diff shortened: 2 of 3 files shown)


bowlofeggs:
This CI failure is due to https://pagure.io/releng/issue/7377.

bodhi/server/consumers/masher.py
```diff
@@ -1005,24 +1005,30 @@ def _get_master_repomd_url(self, arch):
             basestring: A URL on the master mirror where the repomd.xml file should be synchronized.
         """
         release = self.compose.release.id_prefix.lower().replace('-', '_')
+        version = self.compose.release.version
         request = self.compose.request.value
 
+        # First check to see if there's an override for the current version, if not, fall back.
+        # This will first try fedora_28_stable_(suffix), and then fedora_stable_(suffix).
+        key_prefixes = ['%s_%s_%s' % (release, version, request),
+                        '%s_%s' % (release, request)]
```
bowlofeggs, line 1014: Now is not the time for the following idea, but I really want to get this kinda stuff out of the settings file and onto the Release object. I really hate that we have settings where the name of the setting itself is dynamic. It's a nightmare to document :(

That said, this is in keeping with the status quo and things are broken today, and I don't know a better way to do it quickly.

bodhi/server/consumers/masher.py
```diff
@@ -1005,24 +1005,30 @@ def _get_master_repomd_url(self, arch):
...
-        if not master_repomd:
-            raise ValueError("Could not find %s in the config file" % key)
+        keys = [key_prefix + suffix for key_prefix in key_prefixes]
 
-        return master_repomd % (self.compose.release.version, arch)
+        for key in keys:
+            val = config.get(key)
+            if val:
+                return val % (version, arch)
+        raise ValueError("Could not find any of %s in the config file" % ','.join(keys))
```
bowlofeggs, line 1031: [```ValueError```](https://docs.python.org/3/library/exceptions.html#ValueError) is used to indicate that an argument to a function or method is the right type but an unacceptable value. In this case, it's really that the settings file was missing a particular setting, not that the arch passed to this function was invalid.

It might make sense to raise an ```IndexError``` since the config didn't have any of the requested keys, or you could just do a more general ```RuntimeError```.

bowlofeggs:
jenkies test

bodhi/server/consumers/masher.py
bowlofeggs: We should add a test to assert that this Exception is raised.

bowlofeggs (changes requested):

bodhi/server/consumers/masher.py
puiterwijk: https://github.com/puiterwijk/bodhi/blob/2efb5639c6d4d282cfe2ef90b697c9e223584ad5/bodhi/tests/server/consumers/test_masher.py#L1797

bowlofeggs: Nevermind, I didn't notice that there already is a test for the ValueError, or even that it was already raised, so don't worry about the exception type since it was already like this.

bowlofeggs (approved):
I think CI might pass now, so let's wait and see if it does. If it doesn't, I am also running the CI tests locally without rawhide and if that passes we can merge.

puiterwijk:
`rawhide	[91mPublic key for python2-markdown-2.4.1-11.module_0b083f9b.noarch.rpm is not installed. Failing package is: python2-markdown-2.4.1-11.module_0b083f9b.noarch` - modularity content is not yet signed with the f29 key, and also shouldn't be picked yet.

bowlofeggs (changes requested):
Just thought of something - we should document the new setting in ```production.ini```. There's already docs in there for the way it is now, so we can just amend the existing docs.

bowlofeggs:
Why is a module build getting installed?

production.ini
```diff
@@ -163,28 +163,30 @@ use = egg:bodhi-server
...
-# {release}_{request}_master_repomd: This is used by the masher to determine when a
+# {release}_({version}_){request}_master_repomd: This is used by the masher to determine when a
 #     primary architecture push has been synchronized to the master mirror for a given release and
 #     request. The masher will verify that the checksum of repomd.xml at the master URL matches the
 #     expected value, and will poll the URL until this test passes. Substitute release and request
 #     for each release id (replacing -'s with _'s) and request (stable, testing). Used for the
 #     arches listed in {release}_{version}_primary_arches when it is defined, else used for all
 #     arches. You must put two %s's in this setting - the first will be replaced with the release
 #     version and the second will be replaced with the architecture.
+# If a version of the option exists with a matching version, it has priority over one without.
```
bowlofeggs, line 174: Can we explicitly mention here that the ```()``` syntax means the version is optional?

production.ini
```diff
@@ -163,28 +163,30 @@ use = egg:bodhi-server
...
 # examples (these settings do not have defaults):
 #     fedora_stable_master_repomd = http://download01.phx2.fedoraproject.org/pub/fedora/linux/updates/%s/%s/repodata/repomd.xml
 #     fedora_testing_master_repomd = http://download01.phx2.fedoraproject.org/pub/fedora/linux/updates/testing/%s/%s/repodata/repomd.xml
 #     fedora_epel_stable_master_repomd = http://download01.phx2.fedoraproject.org/pub/epel/%s/%s/repodata/repomd.xml
 #     fedora_epel_testing_master_repomd = http://download01.phx2.fedoraproject.org/pub/epel/testing/%s/%s/repodata/repomd.xml
```
bowlofeggs: Maybe we could add a versioned example here too?

production.ini
```diff
@@ -163,28 +163,30 @@ use = egg:bodhi-server
...
 #     secondary architecture push has been synchronized to the master mirror for a given release and
 #     request. The masher will verify that the checksum of repomd.xml at the master URL matches the
 #     expected value, and will poll the URL until this test passes. Substitute release and request
 #     for each release id (replacing -'s with _'s) and request (stable, testing). Used for the
 #     arches not listed in {release}_{version}_primary_arches if it is defined. You must put two
 #     %s's in this setting - the first will be replaced with the release version and the second will
 #     be replaced with the architecture.
+# If a version of the option exists with a matching version, it has priority over one without.
 # examples (these settings do not have defaults):
 #     fedora_stable_alt_master_repomd = http://download01.phx2.fedoraproject.org/pub/fedora-secondary/updates/%s/%s/repodata/repomd.xml
```
bowlofeggs, line 192: Similar comments here.

bowlofeggs (approved):
Looks good, feel free to merge without re-review (if you want to take my suggestions).

bowlofeggs:
The CI tests pass on my local system, after an extreme amount of finagling.

bowlofeggs:
This patch will be included in the 3.5.0 release:

https://github.com/fedora-infra/bodhi/pull/2179

PR #2200, opened 2018-03-07
Add a script to run a docker test for a single release

puiterwijk:

### DIFF
--- a/devel/run_single_test.sh
+++ b/devel/run_single_test.sh
@@ -1,0 +1,59 @@
+#!/usr/bin/bash -ex
+# Copyright (c) 2017 Red Hat, Inc.
+#
+# This file is part of Bodhi.
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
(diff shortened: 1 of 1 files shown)


puiterwijk:
The major advantage of this over running the standard test script with a single release in RELEASES is that this doesn't use parallel, thus it prints out stdout/stderr as they happen, so you get quick test results.

bowlofeggs:
This duplicates some code with the other script. I wonder if we could instead make the other script not use parallel when len(RELEASES) is 1. Perhaps we could substitute the call to parallel with "" or something clever? That way we don't have to maintain two scripts.

bowlofeggs:
I just remembered that parallel has a flag that can make it output its text right away rather than buffering it until the end. We could use that flag when the length of RELEASES is 1 and that might be a real easy way to do it.

bowlofeggs:
See https://github.com/fedora-infra/bodhi/pull/2219 for an implementation of what I mean.

bowlofeggs:
I think we can close this since we have #2219 merged.

PR #2201, opened 2018-03-07
Do not archive CI results anymore.

bowlofeggs:
I originally archived these files because I thought it'd be useful,
but I've never even looked at them so they are a waste of
resources.

### DIFF
--- a/devel/ci/githubprb-project.yml
+++ b/devel/ci/githubprb-project.yml
@@ -50,8 +50,6 @@
     triggers:
         - githubprb
     publishers:
-        - archive:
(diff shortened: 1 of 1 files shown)


puiterwijk:
Maybe also drop the archiving lines from the scripts?

bowlofeggs:
@puiterwijk We can't do that in the same pull request, or it will fail CI. This PR just gets CI to stop looking for the files. Once it is merged and CI is no longer looking for those files, I can do a second PR that stops creating them.

puiterwijk (approved):

bowlofeggs:
jenkies test

PR #2165, opened 2018-02-20
Make sure fixedin field is lower than 256 chars

mattiaverga:
Make sure fixedin field is lower than 256 chars (fixes #1430 ).

Instead of adding the new version to the fixedin list and then check the length of the resultant string, check if the new version can be added to the string without going over the 255 chars limit.

PR #2205, opened 2018-03-08
Add a test to ensure we can close bugs with fixedin field at 254.

bowlofeggs:
In #2165 some patches are added to ensure that Bodhi won't set the
fixedin Bugzilla field to be longer than 254 characters. This test
ensures that exactly 254 characters is allowed.


pypingou (approved):
Jenkins isn't happy, but looks good to me

PR #2209, opened 2018-03-08
Add support for severity to the Bodhi CLI.

### LINKED ISSUE #1814: The CLI cannot set or view an update's severity
I noticed today when creating an update that I could not use the CLI to set or view my update's severity, which is a feature gap with the web UI.

bowlofeggs:
fixes #1814

### DIFF
--- a/bodhi/client/__init__.py
+++ b/bodhi/client/__init__.py
@@ -63,6 +63,8 @@ new_edit_options = [
     click.option('--password', hide_input=True),
     click.option('--type', default='bugfix', help='Update type', required=True,
                  type=click.Choice(['security', 'bugfix', 'enhancement', 'newpackage'])),
+    click.option('--severity', help='Update severity',
+                 type=click.Choice(['unspecified', 'low', 'medium', 'high', 'urgent'])),
     click.option('--notes', help='Update description'),
     click.option('--notes-file', help='Update description from a file'),
     click.option('--bugs', help='Comma-separated list of bug numbers', default=''),
--- a/bodhi/client/bindings.py
+++ b/bodhi/client/bindings.py
@@ -630,6 +630,7 @@ class BodhiClient(OpenIdBaseClient):
             line_formatter.format('Release', update['release']['long_name']),
             line_formatter.format('Status', update['status']),
             line_formatter.format('Type', update['type']),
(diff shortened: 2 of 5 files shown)


pypingou (approved):
Looks fine to me, just jenkins isn't happy :)

bowlofeggs:
Yeah the Jenkins failure was due to GPG issues that were pervasive in Fedora last week. It should pass the current run. /me crosses fingers.

bowlofeggs:
Hmm, the current test failure is a real failure, but I cannot reproduce it in the Vagrant environment so far. Perplexing! Hopefully I can reproduce it with the container test locally.

bowlofeggs:
Well, the good news is that I can reproduce the test failure by running the container tests locally. The weird news is... why? It's F27 in the container and F27 in Vagrant. So why would it fail it one but not the other? So weird. I'll see what I can do to get it to pass on both.

bowlofeggs:
Ah, found it. I'd forgotten that the Vagrant environment sets the BODHI_URL environment variable, which influences which host to use, and that host gets printed in the CLI output. Easy fix.

bowlofeggs:
This is the fix:

```
index 87a2efda..1511757d 100644
@@ -230,6 +230,7 @@ class TestNew(unittest.TestCase):
     """

     @mock.patch.dict(client_test_data.EXAMPLE_UPDATE_MUNCH, {'severity': 'urgent'})
+    @mock.patch.dict(os.environ, {'BODHI_URL': 'http://example.com/tests/'})
     @mock.patch('bodhi.client.bindings.BodhiClient.csrf',
                 mock.MagicMock(return_value='a_csrf_token'))
     @mock.patch('bodhi.client.bindings.BodhiClient.send_request',
@@ -245,7 +246,6 @@ class TestNew(unittest.TestCase):

         expected_output = client_test_data.EXPECTED_UPDATE_OUTPUT.replace('unspecified', 'urgent')
-        expected_output = expected_output.replace('example.com/tests', 'localhost:6543')
         self.assertEqual(result.output, expected_output + '\n')
         bindings_client = send_request.mock_calls[0][1][0]
         send_request.assert_called_once_with(
```

pypingou (approved):
Looks fine to me

PR #2218, opened 2018-03-14
Clean up some dev artifacts before running CI tests.

bowlofeggs:
I noticed that if I run the container tests inside the Vagrant
environment, they often fail to build due to various dev artifacts
that get copied into the container. This commit removes some of
those artifacts, so the CI tests can run inside the Vagrant box.


cverna (approved):
+1

PR #2219, opened 2018-03-15
Don't buffer the devel/run_tests.sh output for one release.

bowlofeggs:

### DIFF
--- a/devel/run_tests.sh
+++ b/devel/run_tests.sh
@@ -34,7 +34,14 @@ else
     PYTEST_ARGS=""
 fi
 
-PARALLEL="parallel -v $FAILFAST --tag"
+releases_array=( $RELEASES )
+if [[ ${#releases_array[@]} == 1 ]]; then
+    BUFFER="--ungroup"
+else
+    BUFFER="--tag"
+fi
+
+PARALLEL="parallel -v $FAILFAST $BUFFER"
 
 
 tar_results() {


puiterwijk (approved):

puiterwijk:
Thanks!

bowlofeggs:
I actually made an alteration since you reviewed it, because converting ```RELEASES``` to a bash array caused problems with how it was used later in the script:

```
diff --git a/devel/run_tests.sh b/devel/run_tests.sh
index dba99584..611716e7 100755
--- a/devel/run_tests.sh
+++ b/devel/run_tests.sh
@@ -24,7 +24,7 @@
 # It is intended to be run with sudo, since it needs to use docker.

 BUILD_PARALLEL=${BUILD_PARALLEL:=""}
-RELEASES=(${RELEASES:="f26 f27 rawhide pip"})
+RELEASES=${RELEASES:="f26 f27 rawhide pip"}

 if [[ $@ == *"-x"* ]]; then
     FAILFAST="--halt now,fail=1"
@@ -34,7 +34,8 @@ else
     PYTEST_ARGS=""
 fi

-if [[ ${#RELEASES[@]} == 1 ]]; then
+releases_array=( $RELEASES )
+if [[ ${#releases_array[@]} == 1 ]]; then
     BUFFER="--ungroup"
 else
     BUFFER="--tag"
```

PR #2220, opened 2018-03-15
Rename the test container image to bodhi-dev.

bowlofeggs:
Previously, the container image would get called "test", which is
too generic. If a developer is using other containers on their
system, it won't be obvious that a container called "test" is for
Bodhi development.


pypingou (approved):
good idea

PR #2226, opened 2018-03-20
Use correct N, V, R splitting for module builds and add stream support

puiterwijk:
This patch adds nvr_name, nvr_version and nvr_release to Build models that will
represent the correct n, v and r components for different build types.
This will also allow multiple builds of different module streams without them
overriding eachother.

### DIFF
--- a/bodhi/server/buildsys.py
+++ b/bodhi/server/buildsys.py
@@ -340,7 +340,7 @@ class DevBuildsys(Buildsystem):
                  'perm': None, 'perm_id': None},
                 {'arches': 'i386 x86_64 ppc ppc64', 'id': 5, 'locked': True,
                  'name': '%s-testing' % release, 'perm': None, 'perm_id': None}]
-        elif '-master-' in build:
+        elif '-master-' in build or build.startswith(('nodejs-6-', 'nodejs-8-', 'nodejs-9-')):
             # Hardcoding for modules in the dev buildsys
             result = [
                 {'arches': 'x86_64', 'id': 15, 'locked': True,
--- a/bodhi/server/models.py
+++ b/bodhi/server/models.py
@@ -47,7 +47,7 @@ from bodhi.server.config import config
 from bodhi.server.exceptions import BodhiException, LockedUpdateException
 from bodhi.server.util import (
     avatar as get_avatar, build_evr, flash_log, get_critpath_components,
-    get_nvr, get_rpm_header, header, packagename_from_nvr, tokenize, pagure_api_get,
+    get_rpm_header, header, tokenize, pagure_api_get,
     greenwave_api_post, waiverdb_api_post)
 import bodhi.server.util
 
@@ -1176,6 +1176,44 @@ class Package(Base):
         del states
         return x
 
+    @staticmethod
+    def _get_name(build):
+        """
+        Determine the package name for a particular build.
+
+        For most builds, this will return the RPM name, unless overridden in a specific
+        subclass.
+
+        Args:
+            build (dict): Information about the build from the build system (koji).
+        Returns:
+            str: The Package object identifier for this build.
+        """
+        name, _, _ = build['nvr']
+        return name
+
+    @staticmethod
+    def get_or_create(build):
+        """
+        Identify and return the Package instance associated with the build.
+
+        For example, given a normal koji build, return a RpmPackage instance.
+        Or, given a container, return a ContainerBuild instance.
+
+        Args:
+            build (dict): Information about the build from the build system (koji).
+        Returns:
+            Package: A type-specific instance of Package for the specific build requested.
+        """
+        base = ContentType.infer_content_class(Package, build['info'])
+        name = base._get_name(build)
+        package = base.query.filter_by(name=name).one_or_none()
+        if not package:
+            package = base(name=name)
+            Session().add(package)
+            Session().flush()
+        return package
+
 
 class ContainerPackage(Package):
     """Represents a Container package."""
@@ -1191,6 +1229,19 @@ class ModulePackage(Package):
     __mapper_args__ = {
         'polymorphic_identity': ContentType.module,
     }
+
+    @staticmethod
+    def _get_name(build):
+        """
+        Determine the name:stream for a particular module build.
+
+        Args:
+            build (dict): Information about the build from the build system (koji).
+        Returns:
+            str: The name:stream of this module build.
+        """
+        name, stream, _ = build['nvr']
+        return '%s:%s' % (name, stream)
 
 
 class RpmPackage(Package):
@@ -1247,6 +1298,68 @@ class Build(Base):
         'polymorphic_on': type,
         'polymorphic_identity': ContentType.base,
     }
+
+    def _get_kojiinfo(self):
+        """
+        Return Koji build info about this build, from a cache if possible.
+
+        Returns:
+            dict: The response from Koji's getBuild() for this Build.
+        """
+        if not hasattr(self, '_kojiinfo'):
+            koji_session = buildsys.get_session()
+            self._kojiinfo = koji_session.getBuild(self.nvr)
+        return self._kojiinfo
+
+    def _get_n_v_r(self):
+        """
+        Return the N, V and R components for a traditionally dash-separated build.
+
+        Returns:
+            tuple: A 3-tuple of name, version, release.
+        """
+        return self.nvr.rsplit('-', 2)
+
+    @property
+    def nvr_name(self):
+        """
+        Return the RPM name.
+
+        Returns:
+            str: The name of the Build.
+        """
+        return self._get_n_v_r()[0]
+
+    @property
+    def nvr_version(self):
+        """
+        Return the RPM version.
+
+        Returns:
+            str: The version of the Build.
+        """
+        return self._get_n_v_r()[1]
+
+    @property
+    def nvr_release(self):
+        """
+        Return the RPM release.
+
+        Returns:
+            str: The release of the Build.
+        """
+        return self._get_n_v_r()[2]
+
+    def get_n_v_r(self):
+        """
+        Return the (name, version, release) of this build.
+
+        Note: This does not directly return the Build's nvr attribute, which is a str.
+
+        Returns:
+            tuple: A 3-tuple representing the name, version and release from the build.
+        """
+        return (self.nvr_name, self.nvr_version, self.nvr_release)
 
     def get_url(self):
         """
@@ -1339,6 +1452,36 @@ class ModuleBuild(Build):
         'polymorphic_identity': ContentType.module,
     }
 
+    @property
+    def nvr_name(self):
+        """
+        Return the ModuleBuild's name.
+
+        Returns:
+            str: The name of the module.
+        """
+        return self._get_kojiinfo()['name']
+
+    @property
+    def nvr_version(self):
+        """
+        Return the the ModuleBuild's stream.
+
+        Returns:
+            str: The stream of the ModuleBuild.
+        """
+        return self._get_kojiinfo()['version']
+
+    @property
+    def nvr_release(self):
+        """
+        Return the ModuleBuild's version and context.
+
+        Returns:
+            str: The version of the ModuleBuild.
+        """
+        return self._get_kojiinfo()['release']
+
 
 class RpmBuild(Build):
     """
@@ -1366,15 +1509,11 @@ class RpmBuild(Build):
         Return:
             tuple: (epoch, version, release)
         """
-        if self.epoch:
-            name, version, release = get_nvr(self.nvr)
-            return (str(self.epoch), version, release)
-        else:
-            koji_session = buildsys.get_session()
-            build = koji_session.getBuild(self.nvr)
-            evr = build_evr(build)
-            self.epoch = int(evr[0])
-            return evr
+        if not self.epoch:
+            self.epoch = self._get_kojiinfo()['epoch']
+            if not self.epoch:
+                self.epoch = 0
+        return (str(self.epoch), str(self.nvr_version), str(self.nvr_release))
 
     def get_latest(self):
         """
@@ -1912,11 +2051,7 @@ class Update(Base):
                                                 "locked update")
 
                 new_builds.append(build)
-                name, version, release = buildinfo[build]['nvr']
-                package = db.query(Package).filter_by(name=name).first()
-                if not package:
-                    package = Package(name=name)
-                    db.add(package)
+                Package.get_or_create(buildinfo[build])
                 b = db.query(Build).filter_by(nvr=build).first()
 
                 up.builds.append(b)
@@ -2075,8 +2210,8 @@ class Update(Base):
                          Update.status == UpdateStatus.pending))
             ).all():
                 obsoletable = False
-                nvr = get_nvr(build.nvr)
-                if rpm.labelCompare(get_nvr(oldBuild.nvr), nvr) < 0:
+                nvr = build.get_n_v_r()
+                if rpm.labelCompare(oldBuild.get_n_v_r(), nvr) < 0:
                     log.debug("%s is newer than %s" % (nvr, oldBuild.nvr))
                     obsoletable = True
 
@@ -2250,7 +2385,7 @@ class Update(Base):
         release information in package labels.
         """
         def build_label(build):
-            return build.nvr if nvr else packagename_from_nvr(self, build.nvr)
+            return build.nvr if nvr else build.package.name
 
         if len(self.builds) > 2:
             title = ", ".join([build_label(build) for build in self.builds[:2]])
@@ -3969,7 +4104,7 @@ class Bug(Base):
         # Build a mapping of package names to build versions
         # so that .close() can figure out which build version fixes which bug.
         versions = dict([
-            (get_nvr(b.nvr)[0], b.nvr) for b in update.builds
+            (b.nvr_name, b.nvr) for b in update.builds
         ])
         bugs.bugtracker.close(self.bug_id, versions=versions, comment=self.default_message(update))
 
--- a/bodhi/server/services/updates.py
+++ b/bodhi/server/services/updates.py
@@ -416,8 +416,8 @@ def query_updates(request):
                   colander_body_validator,
                   validate_nvrs,
                   validate_builds,
-                  validate_uniqueness,
                   validate_build_tags,
+                  validate_uniqueness,
                   validate_acls,
                   validate_enums,
                   validate_requirements,
@@ -451,13 +451,7 @@ def new_update(request):
         for nvr in data['builds']:
             name, version, release = request.buildinfo[nvr]['nvr']
 
-            # Figure out what kind of package this should be.
-            # (Note, this can possibly raise a NotImplementedError, but the
-            # error should have been caught earlier in the validator when
-            # inferring the Package type and creating the Package in the validator.)
-            package_class = ContentType.infer_content_class(
-                base=Package, build=request.buildinfo[nvr]['info'])
-            package = request.db.query(package_class).filter_by(name=name).one()
+            package = Package.get_or_create(request.buildinfo[nvr])
 
             # Also figure out the build type and create the build if absent.
             build_class = ContentType.infer_content_class(
--- a/bodhi/server/util.py
+++ b/bodhi/server/util.py
@@ -96,33 +96,6 @@ def get_rpm_header(nvr, tries=0):
     raise ValueError("No rpm headers found in koji for %r" % nvr)
 
 
-def get_nvr(nvr):
-    """
-    Return the (name, version, release) from a given name-ver-rel string.
-
-    Args:
-        nvr (basestring): The name-version-release string that you wish to know the split nvr from.
-    Returns:
-        tuple: A 3-tuple representing the name, version, and release from the given nvr string.
-    """
-    x = list(map(six.text_type, nvr.split('-')))
-    return ('-'.join(x[:-2]), x[-2], x[-1])
-
-
-def packagename_from_nvr(context, nvr):
-    """
-    Extract and return the package name from the given nvr string.
-
-    Args:
-        context (mako.runtime.Context): The current template rendering context. Unused.
-        nvr (basestring): The name-version-release string that you wish to know the name from.
-    Returns:
-        basestring: The name from the nvr string.
-    """
-    x = list(map(six.text_type, nvr.split('-')))
-    return '-'.join(x[:-2])
-
-
 def mkmetadatadir(path):
     """
     Generate package metadata for a given directory.
@@ -963,7 +936,7 @@ def sorted_builds(builds):
     """
     key_function = functools.cmp_to_key(rpm.labelCompare)
     return sorted(builds,
-                  key=lambda x: key_function(get_nvr(x)),
+                  key=lambda x: key_function(x.get_n_v_r()),
                   reverse=True)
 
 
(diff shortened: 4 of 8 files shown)


bodhi/server/buildsys.py
```diff
@@ -340,7 +340,7 @@ def listTags(self, build, *args, **kw):
                  'perm': None, 'perm_id': None},
                 {'arches': 'i386 x86_64 ppc ppc64', 'id': 5, 'locked': True,
                  'name': '%s-testing' % release, 'perm': None, 'perm_id': None}]
-        elif '-master-' in build:
+        elif '-master-' in build or build.startswith(('nodejs-6-', 'nodejs-8-', 'nodejs-9-')):
```
pypingou, line 343: This seems oddly hardcoded, is it expected?

bodhi/server/models.py
```diff
@@ -1248,6 +1305,43 @@ class Build(Base):
...
+        return self._get_n_v_r()[2]
+
+    def get_n_v_r(self):
+        """
+        Return the (name, version, release) of this build.
+
+        Returns:
+            tuple: A 3-tuple representing the name, version and release from the build.
+        """
+        return (self.nvr_name, self.nvr_version, self.nvr_release)
```
pypingou, line 1362: I'd just call ``self._get_n_v_r()`` as otherwise, you're basically calling it 3 times in a row (and thus doing 3 times the same thing)

pypingou:
Looks like jenkisn is not happy :(

bodhi/server/buildsys.py
puiterwijk: Yes, that's how the whole test buildsys determines whether things are a module etc.
This is just the DevBuildsystem.

bodhi/server/models.py
puiterwijk: No, since _get_n_v_r() is invalid for module builds.
For those, the nvr_name, nvr_version and nvr_release are overridden.

puiterwijk:
Jenkins unhappiness was because of the new py3 stuff, the new tests needed to be marked as not py3 compatible yet.

bodhi/server/buildsys.py
ralphbean: Yeah, the DevBuildsystem is full of oddly hardcoded things.  :+1:

bodhi/server/models.py
```diff
@@ -1339,6 +1433,21 @@ class ModuleBuild(Build):
         'polymorphic_identity': ContentType.module,
     }
 
+    @property
```
hanzz: Can't you use the `['name']`, `['version']`, `['release']` even in the `RpmPackage` and therefore remove this code from `ModuleBuild` and also remove `_get_n_v_r()` in `RpmBuild`?

jankaluza, line 1455: Can't you use `['name']`, `['version']` and `['release']` also in these properties in the `Build` class and therefore you could remove these properties completely from `ModuleBuild` class and also remove `_get_nvr` from the `Build` class?

bodhi/server/models.py
```diff
@@ -1248,6 +1305,43 @@ class Build(Base):
...
+        if not self._kojiinfo:
+            koji_session = buildsys.get_session()
+            self._kojiinfo = koji_session.getBuild(self.nvr)
+        return self._kojiinfo
+
+    def _get_n_v_r(self):
+        """Return the N, V and R components of an RPM build from NVR."""
+        return self.nvr.rsplit('-', 2)
+
+    @property
```
jankaluza: Can't you use `self._kojiinfo['name']`, `self._kojiinfo['version']`, `self._kojiinfo['release']` also here? See my comment for `ModuleBuild`.

puiterwijk: Using the `_get_kojiinfo` function requires going out to koji, which for RPMs (and containers) is not required, since the `rsplit` method works just fine there.

bodhi/server/models.py
puiterwijk: No, because the way you get nvr_name etc are entirely different for modules versus everything else.

bodhi/server/models.py
jankaluza: Ah, that makes sense, I thought you are calling that at least once for these builds anyway. +1 then.

bodhi/server/models.py
```diff
@@ -1176,6 +1176,47 @@ def __str__(self):
...
+    @staticmethod
+    def get_or_create(db, build):
+        """
+        Identify and return the Package instance associated with the build.
+
+        For example, given a normal koji build, return a RpmPackage instance.
+        Or, given a container, return a ContainerBuild instance.
+
+        Args:
+            db (sqlalchemy.orm.session.Session): A database session to use to query for the compose.
```
bowlofeggs: I've actually been wanting to do away with this pattern of passing a db session from method to method. The model classes all have a ```.query``` attribute now which can be used instead.

bodhi/server/models.py
```diff
@@ -1176,6 +1176,47 @@ def __str__(self):
...
+            build (dict): Information about the build from the build system (koji).
+        Returns:
+            Package: A type-specific instance of Package for the specific build requested.
+        Raises:
+            NotImplementedError: If the build is a container.
+        """
+        base = ContentType.infer_content_class(Package, build['info'])
+        name = base._get_name(build)
+        package = db.query(base).filter_by(name=name).one_or_none()
```
bowlofeggs: To get rid of the db parameter, you can replace this line with:

```
package = base.query.filter_by(name=name).one_or_none()
```

bodhi/server/models.py
```diff
@@ -1176,6 +1176,47 @@ def __str__(self):
...
+        """
+        Determine the package name for a particular build.
+
+        For most builds, this will return the RPM name, unless overridden in a specific
+        subclass.
+
+        Args:
+            build (dict): Information about the build from the build system (koji).
+        Returns:
+            identifier (str): The Package object identifier for this build.
```
bowlofeggs: The docblock format used by Bodhi doesn't name the returned parameter, instead only documenting its type and a description. So basically, the above line should be like this:

```
str: The Package object identifier for this build.
```

bowlofeggs: Should we simplify this to state that it returns the build's package name?

bodhi/server/models.py
```diff
@@ -1176,6 +1176,47 @@ def __str__(self):
...
+            Package: A type-specific instance of Package for the specific build requested.
+        Raises:
+            NotImplementedError: If the build is a container.
+        """
+        base = ContentType.infer_content_class(Package, build['info'])
+        name = base._get_name(build)
+        package = db.query(base).filter_by(name=name).one_or_none()
+        if not package:
+            package = base(name=name)
+            db.add(package)
```
bowlofeggs: For this part we can avoid the ```db``` parameter two ways:

0. Since we are in ```models.py```, we have access to the module-level ```Session```, so we can just do ```Session().add(package)```. Since the db session is scoped, you are guaranteed to get the same session instance as ```db```.
1. You can also cheat and get the session from ```base.query.session```.

bodhi/server/models.py
```diff
@@ -1176,6 +1176,47 @@ def __str__(self):
...
+            db.flush()
```
bowlofeggs: Does something else query for this ```Package``` before the session is committed? If so, cool. If not, we probably don't need this ```flush()``` and it might slow the request down a bit.

bodhi/server/models.py
```diff
@@ -1192,6 +1233,22 @@ class ModulePackage(Package):
         'polymorphic_identity': ContentType.module,
     }
 
+    @staticmethod
+    def _get_name(build):
+        """
+        Determine the package name for a particular build.
```
bowlofeggs: "Determine the name:stream for a particular module build."

bodhi/server/models.py
```diff
@@ -1192,6 +1233,22 @@ class ModulePackage(Package):
...
+
+        For most builds, this will return the RPM name, unless overridden in a specific
+        subclass.
```
bowlofeggs: I think we can drop this sentence.

bodhi/server/models.py
```diff
@@ -1192,6 +1233,22 @@ class ModulePackage(Package):
...
+
+        Args:
+            build (dict): Information about the build from the build system (koji).
+        Returns:
+            identifier (str): The Package object identifier for this build.
```
```
str: The name:stream of this module build.
```

bodhi/server/models.py
```diff
@@ -1248,6 +1305,43 @@ class Build(Base):
         'polymorphic_identity': ContentType.base,
     }
 
+    # Local cache in case we need more koji info
+    _kojiinfo = None
```
bowlofeggs: We should probably make this an instance attribute instead of a class attribute, since it seems to pertain to a specific build instance and not to the collection of them.

bodhi/server/models.py
```diff
@@ -1248,6 +1305,43 @@ class Build(Base):
...
+
+    def _get_kojiinfo(self):
```
bowlofeggs: Let's add a docblock:

```
"""
Return Koji build info about this package, from a cache if possible.

Returns:
    dict: The response from Koji's getBuild() for this Build.
"""
```

bodhi/server/models.py
```diff
@@ -1248,6 +1305,43 @@ class Build(Base):
...
+            self._kojiinfo = koji_session.getBuild(self.nvr)
+        return self._kojiinfo
+
+    def _get_n_v_r(self):
+        """Return the N, V and R components of an RPM build from NVR."""
+        return self.nvr.rsplit('-', 2)
+
+    @property
+    def nvr_name(self):
+        """Return the RPM Name."""
```
bowlofeggs: Let's document the return type too:

```
"""
Return the RPM name.

Returns:
    str: The name of the RPM.
"""
```

bodhi/server/models.py
```diff
@@ -1248,6 +1305,43 @@ class Build(Base):
...
+        return self._get_n_v_r()[0]
+
+    @property
+    def nvr_version(self):
+        """Return the RPM Version."""
```
```
"""
Return the RPM version.

Returns:
    str: The version of the RPM.
"""
```

bodhi/server/models.py
```diff
@@ -1248,6 +1305,43 @@ class Build(Base):
...
+        return self._get_n_v_r()[1]
+
+    @property
+    def nvr_release(self):
+        """Return the RPM Release."""
```
```
"""
Return the release of the RPM.

Returns:
    str: The release of the RPM.
"""
```

bodhi/server/models.py
```diff
@@ -1248,6 +1305,43 @@ class Build(Base):
...
+    _kojiinfo = None
+
+    def _get_kojiinfo(self):
+        if not self._kojiinfo:
+            koji_session = buildsys.get_session()
+            self._kojiinfo = koji_session.getBuild(self.nvr)
+        return self._kojiinfo
+
+    def _get_n_v_r(self):
+        """Return the N, V and R components of an RPM build from NVR."""
```
bowlofeggs: This isn't an RPMBuild, it's the Build superclass. Let's also document the return type. How about something like:

```
"""
Return the N, V, and R components for a traditionally dash-separated build.

Returns:
    tuple: A 3-tuple of name, version, and release.
"""
```

bodhi/server/models.py
```diff
@@ -1248,6 +1305,43 @@ class Build(Base):
...
+
+    @property
+    def nvr_release(self):
+        """Return the RPM Release."""
+        return self._get_n_v_r()[2]
+
+    def get_n_v_r(self):
+        """
+        Return the (name, version, release) of this build.
+
```
bowlofeggs, line 1356: Maybe let's add this as a second sentence in this docblock:

Note: This does not directly return the Build's nvr attribute, which is a str.

bodhi/server/models.py
```diff
@@ -1339,6 +1433,21 @@ class ModuleBuild(Build):
         'polymorphic_identity': ContentType.module,
     }
 
+    @property
+    def nvr_name(self):
+        """Return the RPM Name."""
```
```
"""
Return the ModuleBuild's name.

Returns:
    str: The name of the module.
"""
```

(This is name, and not name:stream, right? If not, please amend the message above as appropriate.

bodhi/server/models.py
```diff
@@ -1339,6 +1433,21 @@ class ModuleBuild(Build):
...
+        return self._get_kojiinfo()['name']
+
+    @property
+    def nvr_version(self):
+        """Return the RPM Version."""
```
```
"""
Return the the ModuleBuild's version.

Returns:
    str: The version of the ModuleBuild.
"""
```

bowlofeggs: Or... is this the stream?

bodhi/server/models.py
```diff
@@ -1339,6 +1433,21 @@ class ModuleBuild(Build):
...
+        return self._get_kojiinfo()['version']
+
+    @property
+    def nvr_release(self):
+        """Return the RPM Release."""
```
```
"""
Return the ModuleBuild's <?>.

Returns:
    str: The ? of the ModuleBuild.
"""
```

bodhi/server/models.py
```diff
@@ -1913,10 +2018,7 @@ def edit(cls, request, data):
 
                 new_builds.append(build)
                 name, version, release = buildinfo[build]['nvr']
```
bowlofeggs: Looks like we can drop this line too?

bodhi/server/models.py
```diff
@@ -2075,8 +2177,8 @@ def obsolete_older_updates(self, db):
                          Update.status == UpdateStatus.pending))
             ).all():
                 obsoletable = False
-                nvr = get_nvr(build.nvr)
-                if rpm.labelCompare(get_nvr(oldBuild.nvr), nvr) < 0:
+                nvr = build.get_n_v_r()
+                if rpm.labelCompare(oldBuild.get_n_v_r(), nvr) < 0:
```
bowlofeggs, line 2214: Much nicer API :)

bodhi/server/services/updates.py
```diff
@@ -451,13 +451,7 @@ def new_update(request):
...
             name, version, release = request.buildinfo[nvr]['nvr']
 
-            # Figure out what kind of package this should be.
-            # (Note, this can possibly raise a NotImplementedError, but the
-            # error should have been caught earlier in the validator when
-            # inferring the Package type and creating the Package in the validator.)
-            package_class = ContentType.infer_content_class(
-                base=Package, build=request.buildinfo[nvr]['info'])
-            package = request.db.query(package_class).filter_by(name=name).one()
+            package = Package.get_or_create(request.db, request.buildinfo[nvr])
```
bowlofeggs: Nice.

bodhi/server/util.py
```diff
@@ -96,33 +96,6 @@ def get_rpm_header(nvr, tries=0):
...
-    Extract and return the package name from the given nvr string.
-
-    Args:
-        context (mako.runtime.Context): The current template rendering context. Unused.
-        nvr (basestring): The name-version-release string that you wish to know the name from.
-    Returns:
-        basestring: The name from the nvr string.
-    """
-    x = list(map(six.text_type, nvr.split('-')))
-    return '-'.join(x[:-2])
```
bowlofeggs, line 123: Oh yeah, deleting code like a boss.

bodhi/server/util.py
```diff
@@ -982,26 +955,22 @@ def sorted_updates(updates):
...
-    for package in builds:
+            builds[build.nvr_name].add(build)
+    # The sorted here is so we actually have a way to test this
+    # Otherwise, we would be depending on the way Python orders dict keys
+    for package in sorted(builds.keys()):
         if len(builds[package]) > 1:
-            log.info('Found multiple %s packages' % package)
             log.debug(builds[package])
             for build in sorted_builds(builds[package])[::-1]:
-                update = build_to_update[build]
```
bowlofeggs, line 997: Hahah what the heck was this here for? Nice of you to fix that to a simple ```build.update``` ☺

bodhi/server/validators.py
```diff
@@ -149,6 +148,66 @@ def validate_csrf_token(node, value):
...
 
+def cache_tags(request, build):
+    """
+    Cache the tags for a koji build.
+
+    Args:
+        request (pyramid.util.Request): The current request.
+        build (basestring): The NVR of the build to cache.
+    Returns:
+        The list of tags or None.
```
bowlofeggs:
```
list or None: The list of tags, or None if there are no tags, or if there was a failure to communicate with Koji.
```

bodhi/server/validators.py
```diff
@@ -149,6 +148,66 @@ def validate_csrf_token(node, value):
...
+        tags = [tag['name'] for tag in request.koji.listTags(build)]
+        if len(tags) == 0:
+            request.errors.add('body', 'builds',
+                               'Cannot find any tags associated with build: %s' % build)
+    except koji.GenericError:
+        request.errors.add('body', 'builds',
+                           'Invalid koji build: %s' % build)
+    # This might end up setting tags to None. That is expected, and indicates it failed.
+    request.buildinfo[build]['tags'] = tags
+    return tags
```
bowlofeggs, line 174: Nice.

bodhi/server/validators.py
```diff
@@ -149,6 +148,66 @@ def validate_csrf_token(node, value):
...
+
+def cache_release(request, build):
+    """
+    Cache the builds release from the request.
+
+    Args:
+        request (pyramid.util.Request): The current request.
+        build (basestring): The NVR of the build to cache.
+    Returns:
+        The release object or None.
```
bowlofeggs: Release or None: The release object, or None if no Release can be matched to the tags associated with the request.

bodhi/server/validators.py
```diff
@@ -163,14 +222,13 @@ def cache_nvrs(request, build):
...
-    if '' in (name, version, release):
-        raise ValueError
-
-    request.buildinfo[build]['nvr'] = name, version, release
-    # Cram some extra information in there, used later to infer type.
-    request.buildinfo[build]['info'] = request.koji.getBuild(build)
+    # Request info from koji, used to split NVR and determine type
+    # We use Koji's information to get the NVR split, because modules can have dashes in their
+    # stream.
+    kbinfo = request.koji.getBuild(build)
```
bowlofeggs, line 230: Might we want to do things the "old way" when they are not ModuleBuilds? I think this is fine too, and is consistent, though it does increase reliance on a network service. Either way is fine with me.

bodhi/server/validators.py
```diff
@@ -402,33 +443,10 @@ def validate_acls(request, **kwargs):
...
-            buildinfo['release'] = release
-            if not release:
-                msg = 'Cannot find release associated with ' + \
-                    'build: {}, tags: {}'.format(build, tags)
-                log.warn(msg)
-                request.errors.add('body', 'builds', msg)
+            # Get the Package and Release objects
+            package = Package.get_or_create(db, buildinfo)
+            release = cache_release(request, build)
+            if release is None:
```
bowlofeggs: Very nice cleanup.

bodhi/tests/server/services/test_updates.py
```diff
@@ -322,6 +322,51 @@ def test_new_module_update(self, publish, *args):
...
+    def test_multiple_updates_single_module_steam(self, *args):
+        # Ensure there are no module packages in the DB to begin with.
+        self.assertEquals(self.db.query(ModulePackage).count(), 0)
+        self.create_release(u'27M')
+
+        # First create an update for nodejs:6
+        self.app.post_json('/updates/', self.get_update(u'nodejs-6-20170101'))
+
+        # Next create a second update for nodejs:6
+        self.app.post_json('/updates/', self.get_update(u'nodejs-6-20170202'))
```
bowlofeggs, line 365: So this one should have obsoleted the former one, right? How about an assertion that the obsoletion happened? There should be two updates now. The first should be obsolete (and have a comment mentioning the new update), and the second should have a comment about inheriting the notes from the first.

bodhi/tests/server/services/test_updates.py
```diff
@@ -712,15 +757,16 @@ def test_edit_add_build_from_different_release(self):
         del update_json['requirements']
 
         res = self.app.post_json('/updates/', update_json, status=400)
+        print('Res: %s' % res)
```
bowlofeggs: Let's drop the print.

bodhi/tests/server/services/test_updates.py
```diff
@@ -768,16 +814,14 @@ def test_edit_koji_error(self):
...
-                                  u'tags: []'),
-                 u'location': u'body', u'name': u'builds'}]}
+                 u'name': u'builds'}]}
         self.assertEqual(res.json, expected_json)
         listTags.assert_called_once_with(update.title)
 
     @unittest.skipIf(six.PY3, 'Not working with Python 3 yet')
     def test_edit_untagged_build(self):
         """Editing an update that references untagged builds should raise an error."""
+        self.maxDiff = None
```
bowlofeggs: Let's drop this line.

bowlofeggs (changes requested):
This is a nice patch - thanks so much. I really love the code consolidation - many things are so much cleaner now. All my comments are pretty minor so we should be able to get this merged and released pretty soon. Thanks again!

bodhi/server/models.py
puiterwijk: Yes, there's a reasonable chance we get into the get_or_create for the same package.

bodhi/tests/server/services/test_updates.py
pypingou: Those I quite handy, they will give you a much more detailed output when the test fails

I tend to set them for the entire class rather than per method personally

bodhi/server/validators.py
puiterwijk: This was already there, also we need the buildinfo to determine whether it is a module build.

bodhi/server/util.py
puiterwijk: This was because in the original code, it was dealing with the nvr strings, not the actual build objects.

bodhi/server/models.py
bowlofeggs: Cool.

bodhi/server/models.py
```diff
@@ -1176,6 +1176,46 @@ def __str__(self):
...
+
+        For example, given a normal koji build, return a RpmPackage instance.
+        Or, given a container, return a ContainerBuild instance.
+
+        Args:
+            build (dict): Information about the build from the build system (koji).
+        Returns:
+            Package: A type-specific instance of Package for the specific build requested.
+        Raises:
+            NotImplementedError: If the build is a container.
```
bowlofeggs: Containers are supported by Bodhi now, so I don't think this will raise ```NotImplemented```.

bodhi/server/models.py
```diff
@@ -1248,6 +1301,68 @@ class Build(Base):
...
+        return self._get_n_v_r()[1]
+
+    @property
+    def nvr_release(self):
+        """
+        Return the RPM release.
+
+        Returns:
+            str: The release of the RPM.
+        """
```
bowlofeggs: On the three properties above, we should s/RPM/Build/.

bodhi/server/models.py
```diff
@@ -1339,6 +1454,36 @@ class ModuleBuild(Build):
...
+
+        Returns:
+            str: The stream of the ModuleBuild.
+        """
+        return self._get_kojiinfo()['version']
+
+    @property
+    def nvr_release(self):
+        """
+        Return the ModuleBuild's version.
```
bowlofeggs: Would the version also include the context? If so, might be good to mention here too.

bowlofeggs (approved):
Cool just a couple nits left. Feel free to merge without re-review. Thanks!

bowlofeggs:
This patch will be included in 3.5.1: https://github.com/fedora-infra/bodhi/pull/2231

Issue #2236, opened 2018-03-21
Invalid NVRs used in the update form result in a server-side traceback and no UI feedback


bowlofeggs:
I deployed 75095940601a80005eb0ee79ca39ec6956040ae3 to staging and found that I did not get a UI error when using an NVR that was unknown to stg koji, with this traceback:

```
2018-03-21 16:57:08,195 ERROR [bodhi.server][Dummy-1] Error caught.  Handling HTML response.
Traceback (most recent call last):
  File "/usr/lib/python2.7/site-packages/pyramid/tweens.py", line 39, in excview_tween
    response = handler(request)
  File "/usr/lib/python2.7/site-packages/pyramid/router.py", line 156, in handle_request
    view_name
  File "/usr/lib/python2.7/site-packages/pyramid/view.py", line 617, in _call_view
    response = view_callable(context, request)
  File "/usr/lib/python2.7/site-packages/pyramid/config/views.py", line 181, in __call__
    return view(context, request)
  File "/usr/lib/python2.7/site-packages/pyramid/viewderivers.py", line 389, in attr_view
    return view(context, request)
  File "/usr/lib/python2.7/site-packages/pyramid/viewderivers.py", line 367, in predicate_wrapper
    return view(context, request)
  File "/usr/lib/python2.7/site-packages/pyramid/viewderivers.py", line 300, in secured_view
    return view(context, request)
  File "/usr/lib/python2.7/site-packages/pyramid/viewderivers.py", line 438, in rendered_view
    result = view(context, request)
  File "/usr/lib/python2.7/site-packages/pyramid/viewderivers.py", line 147, in _requestonly_view
    response = view(request)
  File "/usr/lib/python2.7/site-packages/cornice/service.py", line 487, in wrapper
    validator(request, **args)
  File "/usr/lib/python2.7/site-packages/bodhi/server/validators.py", line 245, in validate_nvrs
    cache_nvrs(request, build)
  File "/usr/lib/python2.7/site-packages/bodhi/server/validators.py", line 232, in cache_nvrs
    request.buildinfo[build]['nvr'] = kbinfo['name'], kbinfo['version'], kbinfo['release']
TypeError: 'NoneType' object has no attribute '__getitem__'
```

PR #2237, opened 2018-03-21
Catch the case where a non-existing update is submitted

puiterwijk:
Fixes: #2236


bowlofeggs (approved):
Looks good. Once CI passes I think we'll be good to merge. Thanks!

PR #2240, opened 2018-03-22
Tests bodhi.server.util

bowlofeggs:
This pull requests brings ```bodhi.server.util``` to 100% test coverage. It contains three commits. The first two fix some bugs I discovered along the way and have detailed commit messages, and the third adds tests (including some that cover the first two commits).

### DIFF
--- a/bodhi/server/util.py
+++ b/bodhi/server/util.py
@@ -20,7 +20,6 @@
 
 from collections import defaultdict, OrderedDict
 from contextlib import contextmanager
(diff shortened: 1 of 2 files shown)


pypingou (approved):
+1 for me but jenkins seems unhappy

bowlofeggs:
@pypingou it was an unused import that displeased ```flake8```. I've fixed that and the tests are re-running now.

PR #2243, opened 2018-03-24
Fix loop variable leaking in sorted_updates

puiterwijk:
The recent N,V,R code introduced a bug in this, where the build and updates variables
were still the build and updates values from the builds population loop.

### DIFF
--- a/bodhi/tests/server/test_util.py
+++ b/bodhi/tests/server/test_util.py
@@ -933,6 +933,43 @@ class TestUtils(base.BaseTestCase):
         assert sync[0] == u2
         assert sync[1] == u1
 
+    def test_sorted_updates_general(self):
+        u1 = self.create_update(['bodhi-1.0-1.fc24', 'somepkg-2.0-3.fc24'])
+        u2 = self.create_update(['somepkg-1.0-3.fc24'])
+        u3 = self.create_update(['pkga-1.0-3.fc24'])
+        u4 = self.create_update(['pkgb-1.0-3.fc24'])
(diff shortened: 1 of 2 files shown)


puiterwijk:
Note that the `test_sorted_updates_insanity` test case is unrelated to the fix but I figured adding it might be useful anyway.
It makes sure that in the face of insanity, we maintain predictability.

Also, the `Exception` should, as the comment says, never be triggered, and there should be no way to trigger it, but it is a safeguard so that if a bug like this gets reintroduced ever again, we make doubly sure the composer doesn't return a success, even though not all updates had been pushed.

pypingou:
Looks fine to me, but I'll let @bowlofeggs make the final call

bodhi/tests/server/test_util.py
```diff
@@ -933,6 +933,43 @@ def test_sorted_updates_async_removal(self):
...
+        but we want to make sure that we at least don't crash, and produce predictable ordering.
+        """
+        u1 = self.create_update(['bodhi-1.0-1.fc24', 'somepkg-2.0-3.fc24'])
+        u2 = self.create_update(['pkga-1.0-3.fc24', 'pkgb-2.0-1.fc24'])  # Newer pkgb, thus >u3
+        u3 = self.create_update(['pkga-2.0-1.fc24', 'pkgb-1.0-3.fc24'])  # Newer pkga, thus >u2
+
+        us = [u1, u2, u3]
+        sync, async = util.sorted_updates(us)
+
+        # This ordering is actually insane, since both u2 and u3 contain a newer and an older build
```
bowlofeggs, line 968: lol

bowlofeggs (approved):

PR #2229, opened 2018-03-20
Add 100% test coverage for bodhi.server.config.

bowlofeggs:
Signed-off-by: Randy Barlow <randy@electronsweatshop.com>

PR #2245, opened 2018-03-24
Require 98% test coverage.

bowlofeggs:
This will not pass CI until #2229, #2230, #2239, #2240, and #2244 are merged.


pypingou (approved):
+1 for me on the change itself.

PR #2251, opened 2018-03-28
Fix ACL lookup for Module Packages

puiterwijk:
With the recent name:stream Package naming for modules, calling their ACLs
broke because externally, the package is still named without stream.

This patch makes sure to strip the :stream part when requesting information
from an external service about a package.


jankaluza:
+1

puiterwijk:
The impact of this is that currently only provenpackagers can submit module updates.

bowlofeggs (approved):
Thanks! I will include this in 3.6.0.

Issue #2018, opened 2017-11-16
Container buildroot overrides should be able to alter tags in the registry


bowlofeggs:
There is a suggestion that we should use build tags in the registry to define which container tags should be the build targets. This is currently a bit controversial, so this ticket should not proceed until we are sure it will work this way.

If we do end up using build tags in the container registry, Bodhi will need to be able to modify the build tags when developers submit container BuildrootOverrides.

Depends on #2019.

bowlofeggs:
I alllllmost have this working, but I hit a snag that is going to increase the complexity of my patch a bit. It turns out that skopeo cannot remove tags from a registry, but can only remove entire images:

https://github.com/projectatomic/skopeo/issues/489

To work around this, I think I will use requests to make the necessary API call to the registry until skopeo supports the call we need. This will unfortunately also require workaround configs in Bodhi so we can pass along TLS certificates to requests to access the registry.

PR #2259, opened 2018-04-04
Refactor the container compose to use a container copy function.

bowlofeggs:
In an effort to re-use the skopeo code between the Buildroot
Override container handler and the composer, this commit creates a
useful common function that can satisfy the needs of both use
cases.

re #2018

### DIFF
--- a/bodhi/server/util.py
+++ b/bodhi/server/util.py
@@ -990,7 +990,7 @@ def sorted_updates(updates):
     return sync, async
 
 
-def cmd(cmd, cwd=None):
+def cmd(cmd, cwd=None, raise_on_error=False):
     """
     Run the given command in a subprocess.
 
@@ -1000,26 +1000,28 @@ def cmd(cmd, cwd=None):
             basestring.split() to form the list to pass to Popen().
         cwd (basestring or None): The current working directory to use when launching the
             subprocess.
+        raise_on_error (bool): If True, raise a RuntimeError if the command's exit code is non-0.
+            Defaults to False.
     Returns:
         tuple: A 3-tuple of the standard output (basestring), standard error (basestring), and the
             process's return code (int).
+    Raises:
+        RuntimeError: If exception is True and the command's exit code is non-0.
     """
-    log.info('Running %r', cmd)
     if isinstance(cmd, six.string_types):
         cmd = cmd.split()
-    p = subprocess.Popen(cmd, cwd=cwd,
-                         stdout=subprocess.PIPE,
-                         stderr=subprocess.PIPE)
+    log.debug('Running {}'.format(' '.join(cmd)))
+    p = subprocess.Popen(cmd, cwd=cwd, shell=False, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
     out, err = p.communicate()
-    if out:
-        log.debug(out)
-    if err:
-        if p.returncode == 0:
-            log.debug(err)
-        else:
-            log.error(err)
+    output = '{}\n{}'.format(out, err)
     if p.returncode != 0:
-        log.error('return code %s', p.returncode)
+        msg = '{} returned a non-0 exit code: {}'.format(' '.join(cmd), p.returncode)
+        log.error(msg)
+        log.error(output)
+        if raise_on_error:
+            raise RuntimeError(msg)
+    elif out or err:
+        log.debug(output)
     return out, err, p.returncode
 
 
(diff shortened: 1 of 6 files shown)


bowlofeggs:
jenkies test

bodhi/server/util.py
```diff
@@ -999,26 +999,28 @@ def cmd(cmd, cwd=None):
...
+            Defaults to False.
     Returns:
         tuple: A 3-tuple of the standard output (basestring), standard error (basestring), and the
             process's return code (int).
+    Raises:
+        RuntimeError: If exception is True and the command's exit code is non-0.
     """
-    log.info('Running %r', cmd)
     if isinstance(cmd, six.string_types):
         cmd = cmd.split()
```
puiterwijk, line 1012: This would cause problems with commands containing spaces in arguments like: `cmd = 'mv "some folder/with spaces" somewhere/else`

bodhi/server/util.py
```diff
@@ -999,26 +999,28 @@ def cmd(cmd, cwd=None):
             basestring.split() to form the list to pass to Popen().
         cwd (basestring or None): The current working directory to use when launching the
             subprocess.
+        exception (bool): If True, raise a RuntimeError if the command's exit code is non-0.
```
puiterwijk: Maybe rename this to `raise_on_error` or something?
Since `exception` to me doesn't obviously represent a True/False.

puiterwijk (approved):
Looks reasonable, just a few remarks.

bodhi/server/util.py
bowlofeggs: Most certainly. However, this silly code predates this patch so I would argue we should deal with it separately of this PR. It would not be simple to have this function handle the string case properly. Perhaps it would be better and cleaner to just drop support for str cmds and force all callers to pass lists.

bodhi/server/util.py
bowlofeggs: Sure, I'll go with that name - it's more descriptive.

bowlofeggs: I'm applying this patch:

```
index d24048fe..1b3a72c2 100644
--- a/bodhi/server/util.py
+++ b/bodhi/server/util.py
@@ -990,7 +990,7 @@ def sorted_updates(updates):
     return sync, async


-def cmd(cmd, cwd=None, exception=False):
+def cmd(cmd, cwd=None, raise_on_error=False):
     """
     Run the given command in a subprocess.

@@ -1000,7 +1000,7 @@ def cmd(cmd, cwd=None, exception=False):
             basestring.split() to form the list to pass to Popen().
         cwd (basestring or None): The current working directory to use when launching the
             subprocess.
-        exception (bool): If True, raise a RuntimeError if the command's exit code is non-0.
+        raise_on_error (bool): If True, raise a RuntimeError if the command's exit code is non-0.
             Defaults to False.
     Returns:
         tuple: A 3-tuple of the standard output (basestring), standard error (basestring), and the
@@ -1018,7 +1018,7 @@ def cmd(cmd, cwd=None, exception=False):
         msg = '{} returned a non-0 exit code: {}'.format(' '.join(cmd), p.returncode)
         log.error(msg)
         log.error(output)
-        if exception:
+        if raise_on_error:
             raise RuntimeError(msg)
     elif out or err:
         log.debug(output)
@@ -1410,7 +1410,7 @@ def copy_container(build, destination_registry=None, destination_tag=None):
     if config.get('skopeo.extra_copy_flags'):
         for flag in reversed(config.get('skopeo.extra_copy_flags').split(',')):
             skopeo_cmd.insert(2, flag)
-    cmd(skopeo_cmd, exception=True)
+    cmd(skopeo_cmd, raise_on_error=True)


 def _container_image_url(build, registry, tag=None):
```

bodhi/server/util.py
bowlofeggs: Ha! And it appears that the ```cmd``` function wasn't used by anything at all until this PR:

```
$ git grep "cmd(.*)"
bodhi/server/util.py:def cmd(cmd, cwd=None):
bodhi/tests/server/test_util.py:            util.cmd('false')
bodhi/tests/server/test_util.py:        util.cmd('/bin/echo', '"home/imgs/catpix"')
bodhi/tests/server/test_util.py:        util.cmd('/bin/echo', '"home/imgs/catpix"')
bodhi/tests/server/test_util.py:        util.cmd('/bin/echo', '"home/imgs/catpix"')
```

So silly. I think maybe it was used in the old masher code and we forgot to drop it. I should do an analysis sometime of Bodhi's functions to make sure they are all used. I do usually do this when writing test coverage, because it's pointless to test coverage on code that is unused.

In any case, I'll send a PR that just deletes support for a str cmd and I think that'll solve the problem. Nothing uses that feature anyway.

bowlofeggs: I created https://github.com/fedora-infra/bodhi/pull/2332 with this change.

bowlofeggs:

PR #2262, opened 2018-04-07
Specify path to Vagrantfile in env variable

abitrolly:
This also adds source checkout command to the list

### DIFF
--- a/docs/developer/vagrant.rst
+++ b/docs/developer/vagrant.rst
@@ -10,9 +10,12 @@ get started, simply use these commands::
     $ sudo dnf install ansible libvirt vagrant-libvirt vagrant-sshfs
     $ sudo systemctl enable libvirtd
     $ sudo systemctl start libvirtd
-    $ cp devel/Vagrantfile.example Vagrantfile
-    # Make sure your bodhi checkout is your shell's cwd
-    $ vagrant up
+
+Checkout the code and run ``vagrant``::
+
+    $ git clone https://github.com/fedora-infra/bodhi
+    $ cd bodhi
+    $ VAGRANT_VAGRANTFILE=devel/Vagrantfile.example vagrant up
 
 ``Vagrantfile.example`` sets up a port forward from the host machine's port 6543 into the Vagrant
 guest's port 6543, so you can now visit http://localhost:6543 with your browser to see your Bodhi


docs/developer/vagrant.rst
```diff
@@ -10,9 +10,12 @@ get started, simply use these commands::
...
     $ sudo systemctl start libvirtd
-    $ cp devel/Vagrantfile.example Vagrantfile
-    # Make sure your bodhi checkout is your shell's cwd
-    $ vagrant up
+
+Checkout the code and run ``vagrant``::
+
+    $ git clone https://github.com/fedora-infra/bodhi
+    $ cd bodhi
+    $ VAGRANT_VAGRANTFILE=devel/Vagrantfile.example vagrant up
```
bowlofeggs, line 18: My goal in having an example Vagrantfile was to allow for customizations (the example Vagrant file has some optional commented items that a developer might want to uncomment, and of course they could also add any of their own customizations.

I guess we could do something where we suggest using your proposed method, while mentioning that the developer can also just copy the Vagrantfile if they want to customize it? What you think?

abitrolly: I think that having default Vagrantfile in root is convenient. Mention `VAGRANT_VAGRANTFILE` in docs if somebody needs alternative start from a copy is a second good thing.

bowlofeggs: Hello @abitrolly!

I've thought about this a bit and I still feel on the fence. In favor of your argument, I realized that the Ansible playbook is pretty tied to the specific Fedora version used in the example Vagrantfile, which means that whenever I update it developers might have to adjust their Vagrantfile to use the new Fedora version. However, I still find tweaking the various items in the Vagrantfile to be enticing - I regularly do this on my own copy (I often change the memory and use the caching or disk resizing as well, and these are things I wouldn't want to be the default for all developers.) I guess one thing we could do is make a simple settings file for the sorts of tweaks people want to do, and have a default Vagrantfile that reads that config file?

bowlofeggs:
@abitrolly I've opened https://github.com/fedora-infra/bodhi/pull/2422 which I think accomplishes what you wanted to do here (in its first commit - the second commit is unrelated). I'll close this PR, but feel free to comment over there if you think it should be different. Thanks for the inspiration!

docs/developer/vagrant.rst
abitrolly: Sorry for the late reply. Can you point me to docs that show how Ansible is connected to the Vagrantfile? I don't know about it. For me the purpose for default Vagrant file is to make everybody start `bodhi` as fast and easy as possible, and when people need to send multiple patches - they can setup their own customized environment.

abitrolly:
LGTM.

docs/developer/vagrant.rst
bowlofeggs: Vagrant documents this at https://www.vagrantup.com/docs/provisioning/ansible.html

PR #2275, opened 2018-04-11
Modify models to support side tags.

bowlofeggs:

### DIFF
--- a/bodhi/server/migrations/versions/be25565a1211_add_a_dislay_name_field_to_the_updates_.py
+++ b/bodhi/server/migrations/versions/be25565a1211_add_a_dislay_name_field_to_the_updates_.py
@@ -1,0 +1,41 @@
+# Copyright (c) 2018 Red Hat, Inc.
+#
+# This file is part of Bodhi.
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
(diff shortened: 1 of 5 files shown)


pypingou (approved):
Nice and simple

bodhi/server/migrations/versions/be25565a1211_add_a_dislay_name_field_to_the_updates_.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+# revision identifiers, used by Alembic.
+revision = 'be25565a1211'
+down_revision = '22858ba91115'
+
+
+def upgrade():
+    """Add a display_name column to the updates table."""
+    op.add_column('updates', sa.Column('display_name', sa.UnicodeText(), nullable=False))
```
elyscape, line 36: Because this has `nullable=False` but no `server_default`, it results in a migration failure if there are any rows in the `updates` table.

bodhi/server/models.py
```diff
@@ -1694,6 +1699,7 @@ class Update(Base):
     require_bugs = Column(Boolean, default=False)
     require_testcases = Column(Boolean, default=False)
 
+    display_name = Column(UnicodeText, nullable=False, default='')
```
elyscape, line 1702: Since this is a `UnicodeText` column, the value of `default` should be `u''`.

bowlofeggs:

PR #2279, opened 2018-04-12
Implement fake tags and addTag/deleteTag

puiterwijk:

### DIFF
--- a/bodhi/server/buildsys.py
+++ b/bodhi/server/buildsys.py
@@ -119,6 +119,14 @@ class Buildsystem(object):
         """Raise NotImplementedError."""
         raise NotImplementedError
 
+    def addTag(self, *args, **kw):
+        """Raise NotImplementedError."""
+        raise NotImplementedError
+
+    def deleteTag(self, *args, **kw):
+        """Raise NotImplementedError."""
+        raise NotImplementedError
+
 
 class DevBuildsys(Buildsystem):
     """A dummy buildsystem instance used during development and testing."""
@@ -128,6 +136,7 @@ class DevBuildsys(Buildsystem):
     __added__ = []
     __tagged__ = {}
     __rpms__ = []
+    __tags__ = []
 
     def __init__(self):
         """Initialize the DevBuildsys."""
@@ -414,6 +423,12 @@ class DevBuildsys(Buildsystem):
         Raises:
             koji.GenericError: If strict is True and epel is requested.
         """
+        for nr in self.__tags__:
+            if taginfo == self.__tags__[nr][0]:
+                toreturn = self.__tags__[nr][1]
+                toreturn['id'] = nr
+                return toreturn
+
         if isinstance(taginfo, int):
             taginfo = "f%d" % taginfo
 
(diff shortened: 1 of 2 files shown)


bowlofeggs:
The test failure was unrelated to this patch, but required the patch to be rebased onto the latest develop with the fix. I've force pushed your branch to fix that.

bowlofeggs (approved):

bodhi/server/buildsys.py
```diff
@@ -428,6 +443,26 @@ def getTag(self, taginfo, **kw):
                 'perm': None, 'id': 246, 'arches': None,
                 'maven_include_all': False, 'perm_id': None}
 
+    def addTag(self, tag, **opts):
+        """ Emulate tag adding. """
```
bowlofeggs: The test failure was due to the spaces around this sentence:

```
rawhide	bodhi/server/buildsys.py:446 in public method `addTag`:
rawhide	        D210: No whitespaces allowed surrounding docstring text
```

bodhi/server/buildsys.py
```diff
@@ -428,6 +443,26 @@ def getTag(self, taginfo, **kw):
...
+        opts['maven_support'] = False
+        opts['name'] = tag
+        opts['perm'] = 'admin'
+        opts['arches'] = None
+        opts['maven_include_all'] = False
+        opts['perm_id'] = 1
+        self.__tags__.append((tag, opts))
+
+    def deleteTag(self, tagid):
+        """ Emulate tag deletion. """
```
bowlofeggs:
```
rawhide	bodhi/server/buildsys.py:462 in public method `deleteTag`:
rawhide	        D210: No whitespaces allowed surrounding docstring text
```

bowlofeggs (changes requested):
Just two docblock changes needed to satisfy pydocstyle.

bodhi/server/buildsys.py
```diff
@@ -119,6 +119,14 @@ def getTag(self, *args, **kw):
         """Raise NotImplementedError."""
         raise NotImplementedError
 
+    def addTag(self, *args, **kw):
+        """Raise NotImplementedError."""
+        raise NotImplementedError
+
+    def deleteTag(self, *args, **kw):
+        """Raise NotImplementedError."""
+        raise NotImplementedError
```
bowlofeggs, line 128: Hmm, looks like CI is upset about missing coverage on the above two methods. I added a really [dumb test](https://github.com/fedora-infra/bodhi/blob/3.6.1/bodhi/tests/server/test_buildsys.py#L32-L43) to hit the other methods here, so you could just add to that.

I also think it's reallllly dumb that we have this class at all. We could also just kill the entire class if you want. I really don't see what purpose it serves - it's not like Koji subclasses it - only ```DevBuildsys``` does, and it has its own methods anyway.

Either of the above solutions are fine by me.

bowlofeggs (changes requested):

bodhi/server/buildsys.py
puiterwijk: Just to note: I've went for the dumb test method.

bowlofeggs (approved):

bowlofeggs:

PR #2280, opened 2018-04-12
Use the WaiverDB API correctly.

### LINKED ISSUE #2272: Bodhi does not use the WaiverDB API correctly
We gave Bodhi an auth token for WaiverDB today and it turned out that Bodhi's code for using WaiverDB's API is incorrect in two ways:

0. It uses "proxy_user" instead of "username" to specify the acting user.
1. It sends all waivers in one API call, but WaiverDB needs one waiver per API call.

bowlofeggs:
fixes #2272


pypingou (approved):
:+1: for me

pypingou:
jenkies test

PR #2285, opened 2018-04-16
Test of consumers-updates in Python 3

frenzymadness:
Another one with simple change leading to Python 3 compatibility.


bowlofeggs (approved):
Thanks!

Issue #2258, opened 2018-04-04
Bodhi is not compatible with Cornice 3.2.0


bowlofeggs:
I noticed today that the Vagrant environment will not serve Bodhi:

```
Traceback (most recent call last):
  File "/usr/lib64/python2.7/multiprocessing/process.py", line 267, in _bootstrap
    self.run()
  File "/usr/lib64/python2.7/multiprocessing/process.py", line 114, in run
    self._target(*self._args, **self._kwargs)
  File "/usr/lib/python2.7/site-packages/hupper/worker.py", line 253, in worker_main
    func(*spec_args, **spec_kwargs)
  File "/usr/lib/python2.7/site-packages/pyramid/scripts/pserve.py", line 32, in main
    return command.run()
  File "/usr/lib/python2.7/site-packages/pyramid/scripts/pserve.py", line 229, in run
    app = loader.get_wsgi_app(app_name, config_vars)
  File "/usr/lib/python2.7/site-packages/plaster_pastedeploy/__init__.py", line 131, in get_wsgi_app
    global_conf=defaults)
  File "/usr/lib/python2.7/site-packages/paste/deploy/loadwsgi.py", line 247, in loadapp
    return loadobj(APP, uri, name=name, **kw)
  File "/usr/lib/python2.7/site-packages/paste/deploy/loadwsgi.py", line 272, in loadobj
    return context.create()
  File "/usr/lib/python2.7/site-packages/paste/deploy/loadwsgi.py", line 710, in create
    return self.object_type.invoke(self)
  File "/usr/lib/python2.7/site-packages/paste/deploy/loadwsgi.py", line 146, in invoke
    return fix_call(context.object, context.global_conf, **context.local_conf)
  File "/usr/lib/python2.7/site-packages/paste/deploy/util.py", line 55, in fix_call
    val = callable(*args, **kw)
  File "/home/vagrant/bodhi/bodhi/server/__init__.py", line 358, in main
    return config.make_wsgi_app()
  File "/usr/lib/python2.7/site-packages/pyramid/config/__init__.py", line 1052, in make_wsgi_app
    self.commit()
  File "/usr/lib/python2.7/site-packages/pyramid/config/__init__.py", line 684, in commit
    self.action_state.execute_actions(introspector=self.introspector)
  File "/usr/lib/python2.7/site-packages/pyramid/config/__init__.py", line 1197, in execute_actions
    action = next(action_iter, None)
[...]

PR #2286, opened 2018-04-16
Cornice querystring validator + functional tests

frenzymadness:
There was a bug in previous query string validator which causes that default value for some params was None in Python 3. I tried to fix it but then I found a comment about PR in cornice project with query string validator change so I decided to switch to that version and now it works with both Pythons.

### DIFF
--- a/bodhi/server/services/builds.py
+++ b/bodhi/server/services/builds.py
@@ -20,12 +20,13 @@
 import math
 
 from cornice import Service
+from cornice.validators import colander_querystring_validator
 from pyramid.exceptions import HTTPNotFound
 from sqlalchemy import func, distinct
 from sqlalchemy.sql import or_
 
 from bodhi.server.models import Update, Build, Package, Release
-from bodhi.server.validators import (colander_querystring_validator, validate_updates,
+from bodhi.server.validators import (validate_updates,
                                      validate_packages, validate_releases)
 import bodhi.server.schemas
 import bodhi.server.security
--- a/bodhi/server/services/comments.py
+++ b/bodhi/server/services/comments.py
@@ -21,7 +21,7 @@
 import math
 
 from cornice import Service
-from cornice.validators import colander_body_validator
+from cornice.validators import colander_body_validator, colander_querystring_validator
 from pyramid.httpexceptions import HTTPBadRequest
 from sqlalchemy import func, distinct
 from sqlalchemy.sql import or_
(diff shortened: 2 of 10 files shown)


frenzymadness:
It seems that tests failed but not because of tests itself.

bowlofeggs:
Cool, I've been meaning to do this :)

We should alter ```requirements.txt``` to state ```cornice == 3.1.0```, as that code was introduced in that release. Bodhi has a [problem](https://github.com/fedora-infra/bodhi/issues/2258) with Cornice 3.2.0 so for now we should just require exactly that release.

bowlofeggs:
It seems that the docs fail to build due to a failure to import that validator:

```
f28	Exception occurred:
f28	  File "/bodhi/bodhi/server/services/builds.py", line 23, in <module>
f28	    from cornice.validators import colander_querystring_validator
f28	ImportError: cannot import name colander_querystring_validator
```

bowlofeggs:
Ah I see the problem - Fedora still has Cornice 3.0.0, which doesn't have that validator:

https://apps.fedoraproject.org/packages/python-cornice

We will need to update F28+ to 3.1.0 before we can merge this PR. It is a backwards compatible release, and in fact only has this change:

https://github.com/Cornices/cornice/blob/3.1.0/CHANGES.txt

bowlofeggs:
Here is a BZ about updating Cornice: https://bugzilla.redhat.com/show_bug.cgi?id=1543660

bowlofeggs:
I've updated Rawhide to 3.1.0, and there's an update for F28 in progress:

https://bodhi.fedoraproject.org/updates/FEDORA-2018-c2b59b9114

Once that is stable, this PR should pass CI I think.

frenzymadness:
`python-cornice` 3.1.0 is in stable and available in repos so I'll rerun CI and we'll see.

frenzymadness:
jenkies test

frenzymadness:
It works!

frenzymadness:
Hi. Is there any issue with the PR?

frenzymadness:
Ping ... :)

bowlofeggs (approved):
I'm going to merge this as-is, but we do want to update the requirements and release notes so they specify that cornice >= 3.1.0 is required. I'll do that in a separate PR.

bowlofeggs:
Sorry for taking so long to respond. Things have been on $FIRE for the past several months :(

bowlofeggs:

bowlofeggs:

bowlofeggs:
These patches were deployed to Fedora's production Bodhi instance yesterday.

PR #2288, opened 2018-04-18
Fix last instance of route_url('compose') to composes

puiterwijk:
This completes the change started in 25a312779

### DIFF
--- a/bodhi/server/templates/update.html
+++ b/bodhi/server/templates/update.html
@@ -511,7 +511,7 @@ $(document).ready(function(){
   <div class="col-md-12">
     <h1>
       % if update.locked and update.date_locked:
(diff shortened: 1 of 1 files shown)


bowlofeggs:
This is my mistake for sure, oops!

This also reveals that I must not have had a test that hit this condition in the template code. Unfortunately, I don't know a way to measure test coverage inside the templates so it's not easy to know that things like this weren't hit. In any case, I will write a test for this condition to ensure it stays fixed.

Thanks for the patch!

bowlofeggs (approved):
I added a test to ensure this stays fixed in #2292.

bowlofeggs:
This patch is being backported to the 3.6 branch in https://github.com/fedora-infra/bodhi/pull/2293.

PR #2289, opened 2018-04-18
Do not cache libravatar if DNS is not used

puiterwijk:

### DIFF
--- a/bodhi/server/util.py
+++ b/bodhi/server/util.py
@@ -341,16 +341,19 @@ def avatar(context, username, size):
     https = request.registry.settings.get('prefer_ssl')
 
     @request.cache.cache_on_arguments()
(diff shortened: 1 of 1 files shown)


bowlofeggs (approved):
I'm +1 to the change, but I'm curious why it was suddenly a problem in 3.6.0 and/or OpenShift.

IMO we don't need to test that the caching has moved to a more specific function and we do have coverage here, so I'm good with merging this as-is.

bowlofeggs:
There was a user I remember a few weeks ago who was wondering why his libravatar icon wasn't updating in Bodhi, so this change would have helped him out.

bowlofeggs:
This patch is being backported to the 3.6 branch in https://github.com/fedora-infra/bodhi/pull/2293.

PR #2290, opened 2018-04-18
Rollback a crashed database transaction

puiterwijk:

### DIFF
--- a/bodhi/server/__init__.py
+++ b/bodhi/server/__init__.py
@@ -75,7 +75,12 @@ def get_db_session_for_request(request=None):
         if request.exception is not None:
             session.rollback()
         else:
-            session.commit()
+            try:
+                session.commit()
+            except:
+                session.rollback()
+                session.close()
+                raise
         session.close()
 
     request.add_finished_callback(cleanup)


pypingou:
Code change looks good to me, but jenkins fails because the code coverage is going down :(

bowlofeggs:
@puiterwijk and I talked about this patch yesterday. We are going to try a few other things to address the issue this patch is intended to fix, which was that there were some entangled DB queries, likely due to Bodhi's strange Release caching. We're going to try to warm up caches in ```bodhi.server.main()``` so that they don't require a request to get warmed up. I believe the error happens due to a request warming the cache in the middle of another query.

I'll keep this PR open until we see how our other attempts to fix things work out.

bowlofeggs:
Ugh, I accidentally deleted @pypingou's comment - I had the page loaded from a week ago in a tab and clicked refresh, and as my browser rendered the page I clicked to delete the @centos-ci comment that I must have deleted in another tab a week ago and it swapped out for @pypingou's comment under my mouse. Sorry!

pypingou:
No biggy, iirc I was saying that the code looks fine to me and CI is unhappy because coverage lowers.
But your reasoning in your first comment makes sense and may moot this PR all together.

bowlofeggs:
I think we can safely say that Bodhi is now working in OpenShift without this patch, so I am going to close this PR.

PR #2296, opened 2018-04-19
Warm up the Release._all_releases cache during web app startup.

bowlofeggs:
We've had some problems with our web services flapping in
OpenShift, and it seems that there is some trouble around caching
and concurrent queries. This commit warms up the
bodhi.server.models.Release._all_releases cache in the web app's
main() function so that it does not need to wait for a request to
run the query.

It also drops the db argument from the function since it is not
necessary.

### DIFF
--- a/bodhi/server/__init__.py
+++ b/bodhi/server/__init__.py
@@ -176,13 +176,13 @@ def get_releases(request):
     Return a defaultdict describing all Releases keyed by state.
 
     Args:
-        request (pyramid.request.Request): The current web request.
+        request (pyramid.request.Request): The current web request. Unused.
     Returns:
         collections.defaultdict: A dictionary mapping release states to a list of JSON strings
             that describe the Releases that are in those states.
     """
     from bodhi.server.models import Release
-    return Release.all_releases(request.db)
+    return Release.all_releases()
 
 
 def exception_filter(response, request):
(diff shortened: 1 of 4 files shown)


pypingou (approved):

bowlofeggs:
@pypingou I am adding a test before merging:

```
diff --git a/bodhi/tests/server/test___init__.py b/bodhi/tests/server/test___init__.py
index f0ed380a..12b05abf 100644
--- a/bodhi/tests/server/test___init__.py
+++ b/bodhi/tests/server/test___init__.py
@@ -237,6 +237,16 @@ class TestMain(base.BaseTestCase):
 
         self.assertEqual(config['test'], 'setting')
 
+    def test_warms_up_releases_cache(self):
+        """main() should warm up the _all_releases cache."""
+        # Let's force the release cache to None
+        models.Release._all_releases = None
+
+        server.main({}, testing='guest', session=self.db)
+
+        # The cache should have a release in it now - let's just spot check it
+        self.assertEqual(models.Release._all_releases['current'][0]['name'], 'F17')
+
 
 class TestGetDbSessionForRequest(unittest.TestCase):
     """Test the get_db_session_for_request() function."""
```

pypingou:
:+1:

bowlofeggs:
This is being backported to 3.6 in https://github.com/fedora-infra/bodhi/pull/2305.

PR #2303, opened 2018-04-20
Do not require composer configs to be paths that exist in server.

bowlofeggs:
The goal of this patch is to allow us to split the Bodhi composer
out of the Bodhi server package so we can avoid installing Pungi on
the web server (or non-composer backends). This patch reworks the
configuration verification code so that it only ensures that
composer related path settings are strings, and now the composer
itself ensures the settings refer to existing paths upon startup.

### DIFF
--- a/bodhi/server/config.py
+++ b/bodhi/server/config.py
@@ -203,7 +203,7 @@ def _validate_none_or(validator):
     return _validate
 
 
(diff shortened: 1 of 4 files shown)


puiterwijk (approved):
:ship:

PR #2313, opened 2018-04-23
Notify fedmsg if update time requirements are met

### LINKED ISSUE #2281: bodhi-approve-updates does not publish fedmsg notifications
[The bodhi-approve-updates script][bodhi-approve-updates] runs periodically and comments on updates that have spent enough time in testing to be pushed to stable but have not received enough karma to be automatically pushed. However, since this comment is made as the bodhi user, [no fedmsg notification is sent][excluded]. It would be nice if either these comments were published to fedmsg anyway or the script published notifications to a new fedmsg topic.

[bodhi-approve-updates]: https://github.com/fedora-infra/bodhi/blob/develop/bodhi/server/scripts/approve_testing.py
[excluded]: https://github.com/fedora-infra/bodhi/blob/4253073628d7a17192f03c5d1bd95d6e69503431/bodhi/server/models.py#L3015-L3020

elyscape:
Resolves #2281.

See also fedora-infra/fedmsg_meta_fedora_infrastructure#478.

### DIFF
--- a/bodhi/server/scripts/approve_testing.py
+++ b/bodhi/server/scripts/approve_testing.py
@@ -31,7 +31,7 @@ import six
 
 from ..models import Update, UpdateStatus
 from ..config import config
-from bodhi.server import Session, initialize_db
+from bodhi.server import Session, initialize_db, notifications
 
 
 def usage(argv):
@@ -99,6 +99,10 @@ def main(argv=sys.argv):
                 text = six.text_type(
                     config.get('testing_approval_msg') % update.mandatory_days_in_testing)
                 update.comment(db, text, author=u'bodhi')
+
+                notifications.publish(
+                    topic='update.requirements_met.stable',
+                    msg=dict(update=update))
 
         db.commit()
     except Exception as e:
--- a/bodhi/tests/server/scripts/test_approve_testing.py
+++ b/bodhi/tests/server/scripts/test_approve_testing.py
@@ -34,7 +34,8 @@ class TestMain(BaseTestCase):
     """
     This class contains tests for the main() function.
     """
-    def test_autokarma_update_meeting_time_requirements_gets_one_comment(self):
+    @patch('bodhi.server.notifications.publish')
+    def test_autokarma_update_meeting_time_requirements_gets_one_comment(self, publish):
         """
         Ensure that an update that meets the required time in testing gets only one comment from
         Bodhi to that effect, even on subsequent runs of main().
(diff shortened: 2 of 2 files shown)


puiterwijk:
jenkins, test this please

bodhi/tests/server/scripts/test_approve_testing.py
```diff
@@ -60,6 +61,8 @@ def test_autokarma_update_meeting_time_requirements_gets_one_comment(self):
         self.assertEqual(
             comment_q[0].text,
             config.get('testing_approval_msg') % update.release.mandatory_days_in_testing)
+        publish.assert_called_once_with(
+            topic='update.requirements_met.stable', msg=ANY)
```
puiterwijk: I think it might be useful to make sure that more parts of the message are sent correctly, to prevent something like `<Update object at ...>` to be sent in the fedmsg.

bodhi/server/scripts/approve_testing.py
```diff
@@ -100,6 +100,10 @@ def main(argv=sys.argv):
                     config.get('testing_approval_msg') % update.mandatory_days_in_testing)
```
bowlofeggs, line 100: Note that there is another way an update can reach the stable threshold above at line 89. You probably should send the same message there too.

bodhi/tests/server/scripts/test_approve_testing.py
bowlofeggs: I agree with Patrick - it's best to assert that a specific message was sent. I like opinionated assertions :)

bowlofeggs (changes requested):
Looks like a great start - I think we probably want to send a message at the other place and test for that one too.

bodhi/server/scripts/approve_testing.py
elyscape: I didn't send the message there because that results in the `bodhi.update.karma.threshold` message being sent. Should I send it anyway? If so, it's probably a good idea to add a field to the message indicating the reason, right? Something like `reason='time'` or `reason='karma'`.

bodhi/tests/server/scripts/test_approve_testing.py
elyscape: Added a more specific test. Note that we don't need to worry about `<Update object at ...>` being sent because the `Update` class has a `__json__` method.

bowlofeggs:
On 04/24/2018 04:24 PM, Eli Young wrote:
> I didn't send the message there because that results in the
> |bodhi.update.karma.threshold| message being sent. Should I send it
> anyway? If so, it's probably a good idea to add a field to the message
> indicating the reason, right? Something like |reason='time'| or
> |reason='karma'|.

Ah in that case I agree, don't worry about sending it there.

bowlofeggs (approved):
Thanks!

bowlofeggs:
jenkies test

bowlofeggs:

PR #2336, opened 2018-04-28
Avoid changing status to bugs which aren't related to approved products.

### LINKED ISSUE #1043: Bugs should only be modified if they are Fedora/EPEL-only
Bodhi occasionally tries to update bugs it should not, such as this one:

https://bugzilla.redhat.com/show_bug.cgi?id=1383657#c6

This is a top-level CVE bug and its state should not be modified by Bodhi since it tracks the issue in products other than Fedora and EPEL. Bodhi could simply check that the component is not 'vulnerability' and product 'Security Response' and skip making any state changes if it is.

mattiaverga:
Fixes #1043 

Bodhi commments on bugs which are not Fedora or Fedora EPEL is still enabled since it's harmless. This however will only happen if the bug number is entered manually in the 'new update' form, since the javascript provides only Fedora and Epel bugs, or in the CLI. Or if a bug which was previously assigned to Fedora or Epel product has changed to something else during it's lifetime.

### DIFF
--- a/bodhi/server/bugs.py
+++ b/bodhi/server/bugs.py
@@ -165,7 +165,8 @@ class Bugzilla(BugTracker):
 
     def on_qa(self, bug_id, comment):
         """
-        Change the status of this bug to ON_QA if it is not already ON_QA, VERIFIED, or CLOSED.
+        Change the status of this bug to ON_QA if bug.product is in admitted products list
+        and if it is not already ON_QA, VERIFIED, or CLOSED.
 
         This will also comment on the bug with some details on how to test and provide feedback for
         this update.
@@ -176,6 +177,9 @@ class Bugzilla(BugTracker):
         """
         try:
             bug = self.bz.getbug(bug_id)
+            if bug.product not in config.get('bz_products'):
+                log.info("Skipping set on_qa on {0!r} bug #{1}".format(bug.product, bug_id))
+                return
             if bug.bug_status not in ('ON_QA', 'VERIFIED', 'CLOSED'):
                 log.debug("Setting Bug #%d to ON_QA" % bug_id)
                 bug.setstatus('ON_QA', comment=comment)
@@ -186,7 +190,8 @@ class Bugzilla(BugTracker):
 
     def close(self, bug_id, versions, comment):
         """
-        Close the bug given by bug_id, mark it as fixed in the given versions, and add a comment.
+        Close the bug given by bug_id if bug.product is in admitted products list,
+        mark it as fixed in the given versions, and add a comment.
 
         Args:
             bug_id (int): The ID of the bug you wish to close.
@@ -197,6 +202,9 @@ class Bugzilla(BugTracker):
         args = {'comment': comment}
         try:
             bug = self.bz.getbug(bug_id)
+            if bug.product not in config.get('bz_products'):
+                log.info("Skipping set closed on {0!r} bug #{1}".format(bug.product, bug_id))
+                return
             # If this bug is for one of these builds...
             if bug.component in versions:
                 version = versions[bug.component]
@@ -253,7 +261,7 @@ class Bugzilla(BugTracker):
 
     def modified(self, bug_id):
         """
-        Mark the given bug as MODIFIED.
+        Mark the given bug as MODIFIED if bug.product is in admitted products list.
 
         The bug will only be marked MODIFIED if it is not already MODIFIED, VERIFIED, or CLOSED.
 
@@ -263,7 +271,7 @@ class Bugzilla(BugTracker):
         try:
             bug = self.bz.getbug(bug_id)
             if bug.product not in config.get('bz_products'):
-                log.info("Skipping %r bug" % bug.product)
+                log.info("Skipping set modified on {0!r} bug #{1}".format(bug.product, bug_id))
                 return
             if bug.bug_status not in ('MODIFIED', 'VERIFIED', 'CLOSED'):
                 log.info('Setting bug #%d status to MODIFIED' % bug_id)
--- a/bodhi/server/bugs.py
+++ b/bodhi/server/bugs.py
@@ -165,8 +165,10 @@ class Bugzilla(BugTracker):
 
     def on_qa(self, bug_id, comment):
         """
-        Change the status of this bug to ON_QA if bug.product is in admitted products list
-        and if it is not already ON_QA, VERIFIED, or CLOSED.
+        Change the status of this bug to ON_QA if it is not already ON_QA, VERIFIED, or CLOSED.
+
+        This method will only operate on bugs that are associated with products listed
+        in the bz_products setting.
 
         This will also comment on the bug with some details on how to test and provide feedback for
         this update.
@@ -190,8 +192,10 @@ class Bugzilla(BugTracker):
 
     def close(self, bug_id, versions, comment):
         """
-        Close the bug given by bug_id if bug.product is in admitted products list,
-        mark it as fixed in the given versions, and add a comment.
+        Close the bug given by bug_id, mark it as fixed in the given versions, and add a comment.
+
+        This method will only operate on bugs that are associated with products listed
+        in the bz_products setting.
 
         Args:
             bug_id (int): The ID of the bug you wish to close.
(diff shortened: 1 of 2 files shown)


bowlofeggs:
jenkies test

bodhi/server/bugs.py
```diff
@@ -165,7 +165,8 @@ def comment(self, bug_id, comment):
 
     def on_qa(self, bug_id, comment):
         """
-        Change the status of this bug to ON_QA if it is not already ON_QA, VERIFIED, or CLOSED.
+        Change the status of this bug to ON_QA if bug.product is in admitted products list
+        and if it is not already ON_QA, VERIFIED, or CLOSED.
```
bowlofeggs: Bodhi uses [pydocstyle](http://www.pydocstyle.org) to enforce docblock style, and one of the rules it enforces is that the first paragraph of a docblock must be a single sentence.

```
f27	bodhi/server/bugs.py:166 in public method `on_qa`:
f27	        D205: 1 blank line required between summary line and description (found 0)
f27	bodhi/server/bugs.py:166 in public method `on_qa`:
f27	        D400: First line should end with a period (not 't')
```

To fix this, I suggest mentioning the product list restriction in its own paragraph. Perhaps something like this instead:

```
Change the status of this bug to ON_QA if it is not already ON_QA, VERIFIED, or CLOSED.

This method will only operate on bugs that are associated with products listed in the bz_products setting.
```

bodhi/server/bugs.py
```diff
@@ -186,7 +190,8 @@ def on_qa(self, bug_id, comment):
 
     def close(self, bug_id, versions, comment):
         """
-        Close the bug given by bug_id, mark it as fixed in the given versions, and add a comment.
+        Close the bug given by bug_id if bug.product is in admitted products list,
+        mark it as fixed in the given versions, and add a comment.
```
bowlofeggs: Similar issue here:

```
f27	bodhi/server/bugs.py:191 in public method `close`:
f27	        D205: 1 blank line required between summary line and description (found 0)
f27	bodhi/server/bugs.py:191 in public method `close`:
f27	        D400: First line should end with a period (not ',')
```

bowlofeggs (changes requested):
Cool, looks nice. We just need to get CI to be happy.

You've been sending quite a few PRs lately - if you think you'll be sending more, feel free to send a PR to add your name to [this list](https://github.com/fedora-infra/bodhi/blob/3.6.1/devel/ci/githubprb-project.yml#L6-L17) and Jenkies will automatically run the tests when you send PRs (no need to wait for an admin to type "jenkies test"). Thanks!

mattiaverga:
jenkies test

bowlofeggs (approved):
Thanks!

bowlofeggs:
A fix for this issue is planned to be included in the upcoming 3.9.0 release: #2477.

pcahyna:
Why does Bodhi close bugs that are against EPEL if the update is not for EPEL? See https://bugzilla.redhat.com/show_bug.cgi?id=1591906 . IIUC, the bz was mistakenly closed and gives now the impression that the issue was fixed in EPEL, while it is not.

bowlofeggs:
@pcahyna I think that's a separate (though related) issue - Bodhi does not have a mapping from releases to Bugzilla products. Feel free to file a new issue about that (or a PR!) if you like.

mattiaverga:
I think this is not yet fixed, since it's planned for 3.9.0?

bowlofeggs:
@mattiaverga correct, though I think the fix in 3.9.0 is also a bit different than what @pcahyna is discussing. The fix in 3.9.0 will assert that Bodhi doesn't edit bugs that aren't in EPEL or Fedora, but it will not assert that Fedora updates are not associated with EPEL bugs (or vice versa).

After thinking about it some more, I'm not sure such an enforcement would be popular with what I perceive as a common workflow - packagers often associate updates for EPEL with Fedora bugs if they make updates across all releases at once, and many people dislike cloning bugs in Fedora.

mattiaverga:
@bowlofeggs ah, I misunderstanded the problem.

I don't know how much hard can be to implement, but I was thinking about a general enforcement to check correspondence between Bugzilla product and Update builds.
I mean: if we have a bug opened on EPEL and a packager wants to associate an update in bugzilla, the update should carry at least one EPEL build. It doesn't make sense to have an update with builds for Fedora only that is associated to an EPEL bug (like the example reported), or an update with EPEL builds only which closes a Fedora bug. In these cases I would say it's likely an error by the packager.
Obviously, if the packager deploy in the same update a Fedora build and an EPEL build, the update can be associated with Fedora or EPEL products.

Shall we open a new bug for discussing more on that?

bowlofeggs:
@mattiaverga The problem is that a single Bodhi update cannot contain both EPEL and Fedora builds, because all builds in a Bodhi update must be for the same release.

mattiaverga:
@bowlofeggs ok, I'll try to explain my thoughts by example since it's more easy for me to write that in english.
Let's suppose that, from the web UI, I create a new update for 'package', I add builds for 'package.fc27', 'package.fc28' and 'package.el7'. I also add two related bugs, one is for component 'Fedora', the other for 'EPEL'.

Bodhi will now create three different updates, one for F27, one for F28 and one for EL7. But each update will be associated with both the bugs, so that the two Fedora updates will act also on the EPEL bug and viceversa.

In my opinion, bodhi lacks an enforcement when it splits the data from the user input to the three different updates. The EPEL bug should be associated to the EPEL update only, while the Fedora bug can be associated to both the Fedora updates (awful, but acceptable).

This would not break the packager common workflow, but it will help to keep things cleaner. There isn't any reason for a packager to fix an EPEL bug with a Fedora update, is there?

bowlofeggs:
On 08/22/2018 02:39 PM, Mattia Verga wrote:
> This would not break the packager common workflow, but it will help to
> keep things cleaner. There isn't any reason for a packager to fix an
> EPEL bug with a Fedora update, is there?

In my experience, the reverse happens quite often - packagers mark EPEL
bugs as fixing Fedora bugs. Most people don't clone bugs from Fedora to
EPEL even if they fix them there too, and some people are even morally
opposed to cloning bugs (@nirik ☺).

I haven't observed very many times when Fedora packages were marked as
fixing EPEL bugs, but I suppose that could happen too.

I personally think the current behavior is what packagers expect.

pcahyna:
> I personally think the current behavior is what packagers expect.

I certainly don't expect EPEL to carry a vulnerable package if the bug has been CLOSED/ERRATA, as it happened here. (I think the bug was not even matched by bz number but by a CVE id.)

bowlofeggs:
I agree that that particular bug being closed is bad.

Looking at the bug, it seems that [the packager was told to use a particular Bodhi template](https://bugzilla.redhat.com/show_bug.cgi?id=1591906#c1) when creating their update, and that template referenced that EPEL bug number. Thus, I don't think it was referenced by CVE ID in Bodhi.

If you feel strongly that Bodhi should refuse to cross-associate bugs, it could be helpful to get feedback from more Fedora packagers for additional perspectives. My position is based primarily on my perceptions of what people might expect, and perhaps additional feedback could be helpful to us. Perhaps a post on Fedora's devel list would be helpful?

Note that it isn't just EPEL vs. Fedora, but bugs can also be associated with particular versions of EPEL and Fedora.

nirik:
I'll note one case here that often causes me to be lazy about fedora vs epel bugs: If there's a security update and it affects both fedora and fedora epel, I build for all the needed branches, then use the web form to create seperate updates from that list. Since I am only filling in one update form I list both the fedora and fedora epel bug in there. If the code that splits things could be smart enough to attach fedora issues to fedora updates and epel issues to epel updates it would avoid that. Or I suppose if I was less lazy and did two updates forms.

pcahyna:
@bowlofeggs 
> Looking at the bug, it seems that the packager was told to use a particular Bodhi template when creating their update, and that template referenced that EPEL bug number. Thus, I don't think it was referenced by CVE ID in Bodhi.

The packager was told to do so when submitting an update for EPEL, but are you sure the suggested template has been ever used, given that the packager has apparently never submitted an update for EPEL? The bug (1591906) is listed among related bugs, and I wonder how it got there, was it by matching the CVE id or by explicit packager's decision?

Even if it was an explicit packager's decision, it appears to be a mistake and Bodhi should not honor the request.
 
@nirik 
> Since I am only filling in one update form I list both the fedora and fedora epel bug in there. If the code that splits things could be smart enough to attach fedora issues to fedora updates and epel issues to epel updates it would avoid that.

It sounds like the right thing to do.

> Or I suppose if I was less lazy and did two updates forms.

I would say the tool should save your time and allow you to be lazy.

bowlofeggs:
@pcahyna I cannot be sure what the packager did, but the OP for that BZ did tell the packager to associate their updates with that bug via the template pasted. If the packager followed those instructions, that would explain how this happened.

In any case, I still invite you to start a discussion on Fedora devel, or even file a new issue about this. This pull request isn't a great forum for discussing this since it's technically not about this issue and since it's not discoverable by others (since it's merged).

bowlofeggs:
The patch for this issue was deployed to Fedora's production Bodhi instance yesterday.

PR #2343, opened 2018-04-30
Add full timestamp as tooltip for comments

### LINKED ISSUE #2260: Add tooltips for comment timestamps with the full timestamp
The timestamps on comments are human-readable but somewhat vague:

Screenshot of a comment with a timestamp of "15 days ago"

The Dates section on the right side of the page is the same but, when you mouseover the timestamp, a tooltip pops up with the full timestamp:

Screenshot of the Dates section with a timestamp of "15 days ago" and a tooltip containing a full timestamp

It would be nice if this functionality was extended to the comment timestamps.

elyscape:
Resolves #2260.

Signed-off-by: Eli Young <elyscape@gmail.com>

### DIFF
--- a/bodhi/server/templates/fragments.html
+++ b/bodhi/server/templates/fragments.html
@@ -7,7 +7,9 @@
           ${util.markup(comment.text) | n}
         </div>
         <div class="col-md-2 text-xs-right text-muted">
-          <a href="${comment.url()}">${util.age(comment.timestamp)}</a>
+          <a href="${comment.url()}" data-toggle="tooltip" title="${comment.timestamp.strftime("%Y-%m-%d %H:%M:%S")} (UTC)">
+            ${util.age(comment.timestamp)}
+          </a>
         </div>
 </div>
 % else:
@@ -33,7 +35,7 @@
         </span>
 
         <span class="pull-xs-right">
-          <a href="${comment.url()}">
+          <a href="${comment.url()}" data-toggle="tooltip" title="${comment.timestamp.strftime("%Y-%m-%d %H:%M:%S")} (UTC)">
             ${util.age(comment.timestamp)}
           </a>
         </span>


bowlofeggs:
jenkies test

bodhi/server/templates/fragments.html
```diff
@@ -7,7 +7,9 @@
           ${util.markup(comment.text) | n}
         </div>
         <div class="col-md-2 text-xs-right text-muted">
-          <a href="${comment.url()}">${util.age(comment.timestamp)}</a>
+          <a href="${comment.url()}" data-toggle="tooltip" title="${comment.timestamp.strftime("%Y-%m-%d %H:%M:%S")}">
```
bowlofeggs: Some of the timestamps in the right hand box specify that the timestamp is UTC in their hovertexts (like the "Submitted" and "In testing" times") - would you like to add that here too?

Weirdly, it seems like the number of days to stable one doesn't have a hoverover.

Anyways, I'm fine with this patch as-is if you like this, but those are just some thoughts. Let me know!

bowlofeggs (approved):

bodhi/server/templates/fragments.html
elyscape: Done. Should I add a hovertext to the days-to-stable entry?

bowlofeggs: @elyscape it's up to you - if you want to go for it.

bowlofeggs:
jenkies test

bodhi/server/templates/fragments.html
elyscape: Actually, looks like this was done on purpose. `update.days_to_stable` is a property that returns the number of days left until the update is eligible for stable. To add a tooltip to it, I'd either need to use that to calculate the time until stable in the template or add another property to the `Update` class that returns that value. As such, I'll leave it out, and we can consider adding it in a separate pull request.

bowlofeggs:
On 05/07/2018 02:34 PM, Eli Young wrote:
> As such, I'll leave it out, and we can consider adding it in a separate
> pull request.

+1 - I'd forgotten about that. Thanks for looking into it, and I think
it's fine to not worry about it.

elyscape:
@bowlofeggs Just checking to make sure this didn't fall through the cracks.

bowlofeggs (approved):

bowlofeggs:
Sorry for taking so long to reply - things have been on $FIRE lately and I've been overwhelmed.

elyscape:
No worries. Thank you very much!

bowlofeggs:

bowlofeggs:
The patch for this issue was deployed to Fedora's production Bodhi instance yesterday.

PR #2345, opened 2018-05-01
Store greenwave's unsatisfied requirements in its own db column.

### LINKED ISSUE #2334: Bodhi puts too much text into the Greenwave summary string
A recent change (#2273) expanded the amount of text that would be stored in Greenwave's summary string. Unfortunately, updates now hit the limit of the size of the ```greenwave_summary_string``` column in the database now, leading to failures in the ```bodhi-check-policies``` cron job:

```
[root@bodhi-backend01 ~][STG]# time bodhi-check-policies                                                              
(psycopg2.DataError) value too long for type character varying(255)                                                   
 [SQL: 'UPDATE updates SET greenwave_summary_string=%(greenwave_summary_string)s WHERE updates.id = %(updates_id)s'] [parameters: {'updates_id': 111436, 'greenwave_summary_string': u'18 of 35 required test results missing\n Missing: dist.rpmdeplint, dist.rpmdeplint, dist.rpmdeplint, dist.rpmdeplint, dist.rpmdeplint, dist.rpmdepli ... (57 characters truncated) ... st.rpmdeplint, dist.rpmdeplint, dist.rpmdeplint, dist.rpmdeplint, dist.rpmdeplint, dist.rpmdeplint, dist.rpmdeplint, dist.rpmdeplint, dist.rpmdeplint'}]
(psycopg2.DataError) value too long for type character varying(255)                                                   
 [SQL: 'UPDATE updates SET greenwave_summary_string=%(greenwave_summary_string)s WHERE updates.id = %(updates_id)s'] [parameters: {'updates_id': 111529, 'greenwave_summary_string': u'18 of 70 required test results missing\n Missing: dist.rpmdeplint, dist.rpmdeplint, dist.rpmdeplint, dist.rpmdeplint, dist.rpmdeplint, dist.rpmdepli ... (57 characters truncated) ... st.rpmdeplint, dist.rpmdeplint, dist.rpmdeplint, dist.rpmdeplint, dist.rpmdeplint, dist.rpmdeplint, dist.rpmdeplint, dist.rpmdeplint, dist.rpmdeplint'}]
```

The data in this case is redundant. I think we should make the code that creates this string a little smarter in a couple ways:

* It should not repeat the same test name again and again, but should instead print the name once and perhaps the number of occurrences.
* It should verify that the final string it produces does not exceed 255 characters. It could finish with elipses if there were more failures than fit in 255 characters.

In any case, this is a 3.7.0 blocker as it is a regression from 3.6.1.

### LINKED ISSUE #2335: Newlines in the greenwave_summary_string break the JavaScript that renders test results
While testing #2273 in staging today, I noticed that updates that having missing tests do not render anything in the test results tab. I was able to reproduce this locally and I found that the problem is that the newline gets included in the JavaScript in a way that makes the JS invalid (and the browser refuses to process the rest of the JS as a result). The issue is [here](https://github.com/fedora-infra/bodhi/blob/3.6.1/bodhi/server/templates/update.html#L86):

```
var summary = '${update.greenwave_summary_string}';
```

It may be best to either use a different string to split the missing tests from the summary (perhaps something more unique too, like :::), or to put it into a new field of its own (which might be cleaner but will require a database migration).

This is a 3.7.0 release blocker.

bowlofeggs:
fixes #2334
fixes #2335

### DIFF
--- a/bodhi/server/migrations/versions/59c0f5fbc1b2_add_a_greenwave_unsatisfied_.py
+++ b/bodhi/server/migrations/versions/59c0f5fbc1b2_add_a_greenwave_unsatisfied_.py
@@ -1,0 +1,42 @@
+# Copyright (c) 2018 Red Hat, Inc.
+#
+# This file is part of Bodhi.
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+"""
+Add a greenwave_unsatisfied_requirements column to the updates table.
+
+Revision ID: 59c0f5fbc1b2
+Revises: be25565a1211
+Create Date: 2018-05-01 15:37:07.346034
+"""
+from alembic import op
+import sqlalchemy as sa
+
+
+# revision identifiers, used by Alembic.
+revision = '59c0f5fbc1b2'
+down_revision = 'be25565a1211'
+
+
+def upgrade():
+    """Add a greenwave_unsatisfied_requirements to the updates table."""
+    op.add_column('updates',
+                  sa.Column('greenwave_unsatisfied_requirements', sa.UnicodeText(), nullable=True))
+
+
+def downgrade():
+    """Drop the greenwave_unsatisfied_requirements from the updates table."""
+    op.drop_column('updates', 'greenwave_unsatisfied_requirements')
--- a/bodhi/server/models.py
+++ b/bodhi/server/models.py
@@ -1681,6 +1681,9 @@ class Update(Base):
             Greenwave integration was not enabled when the update was created.
         greenwave_summary_string (unicode): A short summary of the outcome from Greenwave
             (e.g. 2 of 32 required tests failed).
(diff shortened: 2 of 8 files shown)


bowlofeggs:
The UI changes in this PR look like this:

screenshot from 2018-05-01 14-56-01

pypingou (approved):

bowlofeggs:
I'm gonna add this patch to this PR to give a little more explanation of the weird util function:

```
index 79f3a89e..d204576e 100644
--- a/bodhi/server/util.py
+++ b/bodhi/server/util.py
@@ -573,10 +573,20 @@ def greenwave_unsatisfied_requirements_html(context, update):
     if not update.greenwave_unsatisfied_requirements:
         return ''

+    # The Greenwave data is in a list of dictionaries that describe unsatisfied requirements. Let's
+    # collate that data into type of requirement (such as missing test, failed test, etc.) --> test
+    # name --> list of nvrs that fail that test name. To do this, we'll build a dictionary of
+    # dictionaries of lists, like this:
+    #
+    # {'test-result-missing': {'dist.rpmdeplint': ['bodhi-3.6.1-fc28']}}
     reqs = defaultdict(lambda: defaultdict(list))
     for r in json.loads(update.greenwave_unsatisfied_requirements):
+        # The Greenwave response contains a lot of duplicate information. For now, let's ignore
+        # anything that isn't a koji build. https://pagure.io/greenwave/issue/169
         if r['item'].get('type') == 'koji_build':
             reqs[r['type']][r['testcase']].append(r['item']['item'])
+    # Now that we've built our data structure, let's render it into an HTML snippet that can be
+    # included in the update page.
     html = '<div>Unsatisfied requirements:'
     for t in reqs.keys():
         html = html + '<div>{}: {}</div>'.format(
```

pypingou:
:+1: helps the readability quite a bit :)

ralphbean (approved):
Lovely.

In the case where you have `test-result-failed`, you have a `result_id` in the greenwave response that could be used to construct a URL to taskotron so the user can look at the logs...

... but you do already show the resultsdb results further down the page making it still easy for the user to scroll down and get there.

PR #2284, opened 2018-04-16
Prepare tests of CSRF for Python 3

frenzymadness:
Hello.

I've ported tests of CSRF to Python 3 compatible form. Changes contain:

- Be explicit when requesting JSON
- Compare decoded results

Have a nice day. Lumír

PR #2346, opened 2018-05-02
Do not allow updates to be pushed if they fail the gating check.

### LINKED ISSUE #2342: Bodhi does not gate updates based on Greenwave when updates have non-truthy requirements
[```Update.check_requirements()```](https://github.com/fedora-infra/bodhi/blob/3.6.1/bodhi/server/models.py#L3165-L3248) returns True if ```Update.requirements``` is non-truthy. In Fedora's production Bodhi, updates do not have requirements by default, which means that the [greenwave decision check](https://github.com/fedora-infra/bodhi/blob/3.6.1/bodhi/server/models.py#L3243-L3244) is never reached unless developers type something into the update's requirements field. Since most developers do not do this, Fedora's Bodhi instance is effectively not gating any updates based on test results.

bowlofeggs:
fixes #2342

### DIFF
--- a/bodhi/server/models.py
+++ b/bodhi/server/models.py
@@ -3199,6 +3199,9 @@ class Update(Base):
             tuple: A tuple containing (result, reason) where result is a bool
                 and reason is a str.
         """
+        if config.get('test_gating.required') and not self.test_gating_passed:
+            return (False, "Required tests did not pass on this update")
+
         requirements = tokenize(self.requirements or '')
         requirements = list(requirements)
 
@@ -3265,9 +3268,6 @@ class Update(Base):
                 if latest['outcome'] not in ['PASSED', 'INFO']:
                     return False, "Required task %s returned %s" % (
                         latest['testcase']['name'], latest['outcome'])
-
-        if config.get('test_gating.required') and not self.test_gating_passed:
-            return (False, "Required tests did not pass on this update")
 
         # TODO - check require_bugs and require_testcases also?
 
--- a/bodhi/server/templates/update.html
+++ b/bodhi/server/templates/update.html
@@ -587,8 +587,6 @@ $(document).ready(function(){
               ${self.util.push_to_batched_or_stable_button(update) | n}
             % elif update.meets_testing_requirements:
               ${self.util.push_to_batched_or_stable_button(update) | n}
-            % elif update.stable_karma not in (0, None) and update.karma >= update.stable_karma and not update.autokarma:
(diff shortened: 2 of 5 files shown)


puiterwijk (approved):
This must be among the most humorous code vs test proportions I've seen for a bugfix :) :+1:

puiterwijk:
The failed tests are only in py3, and at least two of them are in a file which has most other functions skip on py3.

bowlofeggs:
Yeah, strangely the py3 tests passed in the Vagrant box. @frenzymadness noted that some Python 3 tests pass and fail seemingly randomly when the requester did not explicitly ask for JSON, because the response is randomly chosen between html and JSON when no request header is present. He's been fixing such tests by having them explicitly request JSON.

bowlofeggs:
I believe https://github.com/fedora-infra/bodhi/pull/2284 might help some of these tests to pass.

pypingou (approved):
lgtm, thanks for looking into this! :)

bowlofeggs:
I discovered that this patch is unstable in Python 3 as is, as two of the new tests will randomly fail/pass. @frenzymadness fixed this problem in the base class's ```get_csrf_token()``` so I'll just adjust the tests to use that instead:

```
index 7f425928..5c60c4c7 100644
@@ -708,8 +708,7 @@ class TestSetRequest(base.BaseTestCase):
         up.test_gating_status = TestGatingStatus.failed
         up.date_testing = datetime.utcnow() - timedelta(days=8)
         up.request = None
-        post_data = dict(update=nvr, request='stable',
-                         csrf_token=self.app.get('/csrf').json_body['csrf_token'])
+        post_data = dict(update=nvr, request='stable', csrf_token=self.get_csrf_token())

         res = self.app.post_json('/updates/%s/request' % str(nvr), post_data, status=400)

@@ -728,8 +727,7 @@ class TestSetRequest(base.BaseTestCase):
         up.requirements = ''
         up.test_gating_status = TestGatingStatus.passed
         up.date_testing = datetime.utcnow() - timedelta(days=8)
-        post_data = dict(update=nvr, request='stable',
-                         csrf_token=self.app.get('/csrf').json_body['csrf_token'])
+        post_data = dict(update=nvr, request='stable', csrf_token=self.get_csrf_token())

         res = self.app.post_json('/updates/%s/request' % str(nvr), post_data, status=200)
```

pypingou:
:+1: for that patch

Issue #2361, opened 2018-05-08
Waiving failed gating tests pops up an error in the UI, even though the waiver was successful


bowlofeggs:
I discovered today while testing WaiverDB integration in staging that waiving failed gating tests pops up an error in the UI, even though the waiver was successful:

![screenshot from 2018-05-08 14-11-55](https://user-images.githubusercontent.com/354506/39774800-f11c384a-52c9-11e8-9825-73c8281ef558.png)

This may be due to the JS interpreting the [HTTP 201](https://http.cat/201) response as an error.

bowlofeggs:
This patch is planned to be included in the upcoming [3.8.0 release](https://bodhi.stg.fedoraproject.org/docs/user/release_notes.html) and is currently deployed to [Fedora's staging Bodhi instance](https://bodhi.stg.fedoraproject.org).

PR #2367, opened 2018-05-09
Accept a range of status codes when calling other APIs.

ralphbean:
Notably, [201](https://http.cat/201) which is not the same as [200](https://http.cat/200).

### DIFF
--- a/bodhi/server/util.py
+++ b/bodhi/server/util.py
@@ -1297,7 +1297,8 @@ def call_api(api_url, service_name, error_key=None, method='GET', data=None, hea
             'Bodhi failed to get a resource from {0} at the following URL '
             '"{1}". The status code was "{2}".')
         rv = http_session.get(api_url, timeout=60)
-    if rv.status_code == 200:
+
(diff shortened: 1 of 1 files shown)


ralphbean:
Hm, I see.  There are tests which explicitly mock the status_code.  Ugh.  :/  Will revise.

ralphbean:
Intended to fix #2361.

mikebonnet:
:+1: Looks good!

puiterwijk (approved):

bodhi/server/util.py
```diff
@@ -1297,7 +1297,8 @@ def call_api(api_url, service_name, error_key=None, method='GET', data=None, hea
             'Bodhi failed to get a resource from {0} at the following URL '
             '"{1}". The status code was "{2}".')
         rv = http_session.get(api_url, timeout=60)
-    if rv.status_code == 200:
+
+    if rv.status_code >= 200 and rv.status_code < 300:
```
pypingou, line 1301: Stupid question: would rv.ok work here? (assuming the http_session is a python requests session)

PR #2373, opened 2018-05-11
Make the repo sanity checks much more strict

puiterwijk:
This makes us use librepo/hawkey to verify that a repo can be correctly read and used
by DNF, which should prevent us from accepting a repo if DNF will then crash on using
it.

Note: per the hawkey docs, it's obsoleted, and one is supposed to use libhif,
but the libhif (redirected to libdnf) repo says it is being "reworked and is unstable".
From my tests, it seems that hawkey calls libdnf underwater, so I think that this
is reasonable to do for now.

### DIFF
--- a/bodhi/server/util.py
+++ b/bodhi/server/util.py
@@ -25,6 +25,7 @@ import hashlib
 import json
 import os
 import pkg_resources
+import shutil
 import socket
 import subprocess
 import tempfile
@@ -35,6 +36,9 @@ from pyramid.i18n import TranslationStringFactory
 import arrow
 import bleach
 import colander
+import createrepo_c as cr
+import hawkey
+import libcomps
 import libravatar
 import librepo
 import markdown
@@ -95,7 +99,41 @@ def get_rpm_header(nvr, tries=0):
     raise ValueError("No rpm headers found in koji for %r" % nvr)
 
 
-def mkmetadatadir(path):
+def insert_in_repo(comp_type, repodata, filetype, extension, source):
+    """
+    Inject a file into the repodata with the help of createrepo_c.
+
+    Args:
+        comp_type (int): createrepo_c compression type indication.
+        repodata (basestring): The path to the repo where the metadata will be inserted.
+        filetype (basestring): What type of metadata will be inserted by createrepo_c.
+            This does allow any string to be inserted (custom types). There are some
+            types which are used with dnf repos as primary, updateinfo, comps, filelist etc.
+        extension (basestring): The file extension (xml, sqlite).
+        source (basestring): A file path. File holds the dump of metadata until
+            copied to the repodata folder.
+    """
+    log.info('Inserting %s.%s into %s', filetype, extension, repodata)
+    target_fname = os.path.join(repodata, '%s.%s' % (filetype, extension))
+    shutil.copyfile(source, target_fname)
+    repomd_xml = os.path.join(repodata, 'repomd.xml')
+    repomd = cr.Repomd(repomd_xml)
+    # create a new record for our repomd.xml
+    rec = cr.RepomdRecord(filetype, target_fname)
+    # compress our metadata file with the comp_type
+    rec_comp = rec.compress_and_fill(cr.SHA256, comp_type)
+    # add hash to the compresed metadata file
+    rec_comp.rename_file()
+    # set type of metadata
+    rec_comp.type = filetype
+    # insert metadata about our metadata in repomd.xml
+    repomd.set_record(rec_comp)
+    with open(repomd_xml, 'w') as repomd_file:
+        repomd_file.write(repomd.xml_dump())
+    os.unlink(target_fname)
+
+
+def mkmetadatadir(path, updateinfo=None, comps=None):
     """
     Generate package metadata for a given directory.
 
@@ -103,10 +141,45 @@ def mkmetadatadir(path):
 
     Args:
         path (basestring): The directory to generate metadata for.
+        updateinfo (basestring or None or bool): The updateinfo to insert instead of example.
+            No updateinfo is inserted if False is passed. Passing True provides undefined
+            behavior.
+        comps (basestring or None): The comps to insert instead of example.
     """
+    compsfile = '''<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE comps PUBLIC "-//Red Hat, Inc.//DTD Comps info//EN" "comps.dtd">
+<comps>
+  <group>
+    <id>testable</id>
+    <_name>Testable</_name>
+    <_description>comps group for testing</_description>
+    <packagelist>
+      <packagereq>testpkg</packagereq>
+    </packagelist>
+  </group>
+</comps>'''
+    updateinfofile = ''
     if not os.path.isdir(path):
         os.makedirs(path)
-    subprocess.check_call(['createrepo_c', '--xz', '--database', '--quiet', path])
+    if not comps:
+        comps = os.path.join(path, 'comps.xml')
+        with open(comps, 'w') as f:
+            f.write(compsfile)
+    if updateinfo is None:
+        updateinfo = os.path.join(path, 'updateinfo.xml')
+        with open(updateinfo, 'w') as f:
+            f.write(updateinfofile)
+
+    subprocess.check_call(['createrepo_c',
+                           '--groupfile', 'comps.xml',
+                           '--deltas',
+                           '--xz',
+                           '--database',
+                           '--quiet',
+                           path])
+    if updateinfo is not False:
+        insert_in_repo(cr.XZ, os.path.join(path, 'repodata'), 'updateinfo', 'xml',
+                       os.path.join(path, 'updateinfo.xml'))
 
 
 def flash_log(msg):
@@ -264,7 +337,7 @@ def sanity_check_repodata(myurl):
     Args:
         myurl (basestring): A path to a repodata directory.
     Raises:
-        bodhi.server.exceptions.RepodataException: If the repodata is not valid or does not exist.
+        Exception: If the repodata is not valid or does not exist.
     """
     h = librepo.Handle()
     h.setopt(librepo.LRO_REPOTYPE, librepo.LR_YUMREPO)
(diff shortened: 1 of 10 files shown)


puiterwijk:
@centos-ci please test

puiterwijk:
@centos-ci retest this please

bodhi/server/util.py
```diff
@@ -35,11 +35,15 @@
...
 import colander
+import createrepo_c as cr
+import hawkey
+import libcomps
 import libravatar
 import librepo
 import markdown
 import requests
 import rpm
+import shutil
```
bowlofeggs: For PEP-8, this import should be included with the first group of imports, since it's part of the Python stdlib.

bodhi/server/util.py
```diff
@@ -95,18 +99,85 @@ def get_rpm_header(nvr, tries=0):
     raise ValueError("No rpm headers found in koji for %r" % nvr)
 
 
-def mkmetadatadir(path):
+def insert_in_repo(comp_type, repodata, filetype, extension, source):
+    """
+    Inject a file into the repodata with the help of createrepo_c.
+
+    Args:
```
bowlofeggs, line 106: Let's also document the ```comp_type``` argument here.

bodhi/server/util.py
```diff
@@ -95,18 +99,85 @@ def get_rpm_header(nvr, tries=0):
...
     """
     Generate package metadata for a given directory.
 
     If the metadata doesn't exist, then create it.
 
     Args:
         path (basestring): The directory to generate metadata for.
-    """
+        updateinfo (basestring or None or False): The updateinfo to insert instead of example.
+            No updateinfo is inserted if False is passed.
```
bowlofeggs: It would be good to document the subtle difference between ```None``` and ```False``` here. Also, technically ```False``` isn't a type - the ()'s should document types that are accepted, so we should probably put ```bool``` there instead of ```False```. It would also be good to document that passing ```True``` causes undefined behavior.

bodhi/server/util.py
```diff
@@ -95,18 +99,85 @@ def get_rpm_header(nvr, tries=0):
...
+        comps (basestring or None): The comps to insert instead of example.
+    """
+    compsfile = '''<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE comps PUBLIC "-//Red Hat, Inc.//DTD Comps info//EN" "comps.dtd">
```
bowlofeggs: I'm betting the answer is no... but is there a URL to this dtd? Do we have a copy of it anywhere to reference?

bodhi/server/util.py
```diff
@@ -95,18 +99,85 @@ def get_rpm_header(nvr, tries=0):
...
+  </group>
+</comps>'''
+    updateinfofile = ''
     if not os.path.isdir(path):
         os.makedirs(path)
-    subprocess.check_call(['createrepo_c', '--xz', '--database', '--quiet', path])
+    if not comps:
+        comps = os.path.join(path, 'comps.xml')
+        with open(comps, 'w') as f:
+            f.write(compsfile)
```
bowlofeggs, line 167: Is this for unit tests? If so, I would prefer we use mock to insert it rather than include this in the production code.

bodhi/server/util.py
```diff
@@ -264,7 +335,7 @@ def sanity_check_repodata(myurl):
     Args:
         myurl (basestring): A path to a repodata directory.
     Raises:
-        bodhi.server.exceptions.RepodataException: If the repodata is not valid or does not exist.
+        Exception: If the repodata is not valid or does not exist.
```
bowlofeggs, line 340: It looks like the code still raises ```RepodataException``` - should we change this doc back?

bowlofeggs (changes requested):
We should document the new dependencies on hawkey and libcomps in the release notes and requirements.txt.

bodhi/server/util.py
puiterwijk: The entire function `mkmetadatadir` is only ever used in the tests.....

bodhi/server/util.py
puiterwijk: Yes, but not an official one that the libraries will probably accept: https://github.com/rpm-software-management/libcomps/blob/master/libcomps/tests/comps.dtd

bodhi/server/util.py
bowlofeggs: Ah, in that case can we move it into the tests folder somewhere? Perhaps ```bodhi/tests/server/test_util.py```?

bodhi/server/util.py
puiterwijk: Well, I've seen all kinds of methods of hawkey and librepo fail due to... weird reasons unrelated to the actual repodata in the other functions.

bodhi/server/util.py
bowlofeggs: We talked about this on IRC - for now let's just leave it here and I'll move it in another PR separately.

devel/ci/rawhide-packages
```diff
@@ -10,4 +10,6 @@
     python3-cornice \
     python3-cornice-sphinx \
     python2-webtest \
-    python3-pyramid-fas-openid
+    python3-pyramid-fas-openid \
+    python2-hawkey \
+    python2-libcomps
```
bowlofeggs: I have a feeling the Python 3 versions of these packages are getting pulled in via ```requirements.txt``` instead of via RPM packages. We should probably add those to ```devel/bodhi/devel/ci/rpm-packages```.

bowlofeggs (changes requested):

bowlofeggs:
jenkies test

devel/ci/rawhide-packages
puiterwijk: They are in the base system as they're used by DNF...

bowlofeggs: Ah cool

requirements.txt
```diff
@@ -24,6 +24,8 @@ pyramid_mako
 pyramid_tm
 python-bugzilla
 python-fedora
+python-hawkey
```
bowlofeggs: It looks like hawkey is not available under PyPI (or at least, not under this name). This is OK, as there are other Bodhi dependencies that are also not on PyPI (like Koji), but it does cause the pip container to fail to build:

```
pip	Collecting python-hawkey (from -r /bodhi/requirements.txt (line 27))
pip	[91m  Could not find a version that satisfies the requirement python-hawkey (from -r /bodhi/requirements.txt (line 27)) (from versions: )
pip	[0m[91mNo matching distribution found for python-hawkey (from -r /bodhi/requirements.txt (line 27))
```

To fix this, you can add it to ```devel/ci/pip-packages``` so that it gets installed via RPM for that container.

bowlofeggs (changes requested):

requirements.txt
bowlofeggs: There is something called hawkey, but it has no description so I have no idea if it's the same one:

https://pypi.org/project/hawkey/

Anyways, CI is still failing due to hawkey:

```
pip	Collecting python-hawkey (from -r /bodhi/requirements.txt (line 27))
pip	[91m  Could not find a version that satisfies the requirement python-hawkey (from -r /bodhi/requirements.txt (line 27)) (from versions: )
pip	[0m[91mNo matching distribution found for python-hawkey (from -r /bodhi/requirements.txt (line 27))
```

bowlofeggs (changes requested):

bowlofeggs:
I might be able to fix this next week if you don't have time to get to it.

requirements.txt
bowlofeggs: I'm going to just drop hawkey from requirements.txt and document it in the release notes. Then we can get this PR moving along.

bowlofeggs (approved):

bowlofeggs:
Hmm, the rawhide container is [busted](https://pagure.io/releng/issue/7693) due to GPG sigs :/

bodhi/server/util.py
```diff
@@ -278,17 +351,44 @@ def sanity_check_repodata(myurl):
...
+    primary_sack = hawkey.Sack()
+    hk_repo = hawkey.Repo(myurl)
+    try:
+        hk_repo.filelists_fn = repo_info['filelists']
+        hk_repo.presto_fn = repo_info['prestodelta']
+        hk_repo.primary_fn = repo_info['primary']
+        hk_repo.repomd_fn = repo_info['repomd']
+        hk_repo.updateinfo_fn = repo_info['updateinfo']
+    except KeyError:
+        raise RepodataException('Required part not in repomd.xml')
```
bowlofeggs, line 372: I manually ran the CI tests on my laptop, and they failed due to missing coverage on this exception handler.

bowlofeggs (changes requested):

bowlofeggs:
Also, in order to get the tests to run, I applied this patch:

```
diff --git a/devel/ci/Dockerfile-header b/devel/ci/Dockerfile-header
index 25417d93..9e2f2151 100644
--- a/devel/ci/Dockerfile-header
+++ b/devel/ci/Dockerfile-header
@@ -14,9 +14,11 @@ RUN dnf install --disablerepo rawhide-modular -y \
     liberation-mono-fonts \
     packagedb-cli \
     python2-createrepo_c \
+    python2-hawkey \
     python2-jinja2 \
     python2-koji \
     python2-librepo \
     python2-yaml \
     python3-createrepo_c \
+    python3-hawkey \
     python3-yaml \
diff --git a/devel/ci/f27-packages b/devel/ci/f27-packages
index d95e404f..8f35cf1d 100644
--- a/devel/ci/f27-packages
+++ b/devel/ci/f27-packages
@@ -7,7 +7,6 @@
     python-pytest \
     python-simplemediawiki \
     python-webtest \
-    python2-hawkey \
     python2-libcomps
 
 RUN pip-2 install cornice
diff --git a/devel/ci/pip-packages b/devel/ci/pip-packages
index ef3cf715..0086abb0 100644
--- a/devel/ci/pip-packages
+++ b/devel/ci/pip-packages
@@ -6,9 +6,7 @@
     python3-devel \
     python3-simplemediawiki \
     redhat-rpm-config \
-    python2-hawkey \
     python2-libcomps \
-    python3-hawkey \
     python3-libcomps
 
 COPY requirements.txt /bodhi/requirements.txt
diff --git a/devel/ci/rawhide-packages b/devel/ci/rawhide-packages
index 7f90a0e4..5a1a13d9 100644
--- a/devel/ci/rawhide-packages
+++ b/devel/ci/rawhide-packages
@@ -11,5 +11,4 @@
     python3-cornice-sphinx \
     python2-webtest \
     python3-pyramid-fas-openid \
-    python2-hawkey \
     python2-libcomps
diff --git a/devel/ci/rpm-packages b/devel/ci/rpm-packages
index a1e39fab..dc1abaa0 100644
--- a/devel/ci/rpm-packages
+++ b/devel/ci/rpm-packages
@@ -26,7 +26,6 @@
     python3-fedora \
     python3-feedgen \
     python3-flake8 \
-    python3-hawkey \
     python3-kitchen \
     python3-libcomps \
     python3-markdown \
diff --git a/requirements.txt b/requirements.txt
index 8619c603..3f78b9c3 100644
--- a/requirements.txt
+++ b/requirements.txt
@@ -8,7 +8,6 @@ cryptography
 dogpile.cache
 fedmsg[consumers]
 feedgen
-hawkey
 iniparse
 jinja2
 kitchen
@@ -25,7 +24,6 @@ pyramid_mako
 pyramid_tm
 python-bugzilla
 python-fedora
-python-libcomps
 simplemediawiki
 six
 sqlalchemy
```

I will push that up to your branch now.

bodhi/server/util.py
bowlofeggs: I have a test to address this now. I'll push it up shortly:

```
diff --git a/bodhi/tests/server/test_util.py b/bodhi/tests/server/test_util.py
index 206ec88f..d916816b 100644
--- a/bodhi/tests/server/test_util.py
+++ b/bodhi/tests/server/test_util.py
@@ -16,6 +16,7 @@
+from xml.etree import ElementTree
 import json
 import os
 import shutil
@@ -577,6 +578,24 @@ class TestSanityCheckRepodata(unittest.TestCase):

         self.assertEqual(str(exc.exception), 'Comps file empty')

+    def test_repomd_missing_updateinfo(self):
+        """If the updateinfo data tag is missing in repomd.xml, an Exception should be raised."""
+        util.mkmetadatadir(self.tempdir)
+        repomd_path = os.path.join(self.tempdir, 'repodata', 'repomd.xml')
+        repomd = ElementTree.parse(repomd_path)
+        ElementTree.register_namespace('', 'http://linux.duke.edu/metadata/repo')
+        root = repomd.getroot()
+        # Find the <data type="updateinfo"> tag and delete it
+        for data in root.findall('{http://linux.duke.edu/metadata/repo}data'):
+            if data.attrib['type'] == 'updateinfo':
+                root.remove(data)
+        repomd.write(repomd_path, encoding='UTF-8', xml_declaration=True)
+
+        with self.assertRaises(util.RepodataException) as exc:
+            util.sanity_check_repodata(self.tempdir)
+
+        self.assertEqual(str(exc.exception), 'Required part not in repomd.xml')
+

 class TestType2Icon(unittest.TestCase):
     """Test the type2icon() function."""
```

bowlofeggs (approved):

bowlofeggs:

bowlofeggs:

PR #2379, opened 2018-05-18
Use the full path when cd'ing to the Bodhi folder in Vagrant.

bowlofeggs:
Without this patch, starting tmux inside the Vagrant guest landed
you in /home/vagrant/bodhi/bodhi instead of /home/vagrant/bodhi,
which was annoying. Now it always cd's by full path.


puiterwijk (approved):
Simple enough. :rocket:

PR #2380, opened 2018-05-20
Add a css rule to style blockquotes.

### LINKED ISSUE #2333: Quote markdown does not work
According to the Fedora-Flavored Markdown cheatsheet, using the ``>`` prefix should produce block quotes.

> Like this!

But it doesn't actually. https://bodhi.fedoraproject.org/updates/kernel-4.16.4-200.fc27#comment-774421 was supposed to read:

> Missing the patch for the KVM problem as discussed in: https://bodhi.fedoraproject.org/updates/FEDORA-2018-b6be90818d

Yes... etc.

This makes reading comments containing quotes confusing :(

mattiaverga:
This is intended to be a temporary fix until the rule is applied upstream in fedora-bootstrap. See https://pagure.io/fedora-bootstrap/issue/13

Fixes #2333 

### DIFF
--- a/bodhi/server/static/css/site.css
+++ b/bodhi/server/static/css/site.css
@@ -216,6 +216,11 @@ ul.updateslist li {
     overflow-wrap:  break-word; /* new-school */
     padding:        5px;
 }
+.markdown blockquote {
+    margin: 0 1rem;
+    color: #6a737d;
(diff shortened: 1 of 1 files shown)


bowlofeggs (approved):
Thanks so much!

bowlofeggs:

bowlofeggs:

PR #2383, opened 2018-05-22
Add 100% Python 2 test coverage for bodhi.server.consumers.masher.

bowlofeggs:
This commit also brings coverage for Python 3 to 91% by getting one
more test to run there.


bowlofeggs:
jenkies test

bowlofeggs:
Looks like Centos CI is having trouble getting a worker - reported at https://bugs.centos.org/view.php?id=14852

pypingou (approved):

PR #2386, opened 2018-05-23
Let's use a f28 vagrant box for the development environment

cverna:

### DIFF
--- a/devel/Vagrantfile.example
+++ b/devel/Vagrantfile.example
@@ -10,9 +10,9 @@
 VAGRANTFILE_API_VERSION = "2"
 
 Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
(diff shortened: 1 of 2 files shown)


puiterwijk:
@centos-ci add to whitelist

bowlofeggs:
jenkies test

bowlofeggs (changes requested):
It seems that ```make``` no longer gets installed with F28 as it did with F27, which causes the ```btest``` alias not to work anymore:

```
[vagrant@bodhi-dev bodhi]$ btest
-bash: make: command not found
```

You can edit the packages that get installed in ```devel/ansible/roles/dev/tasks/main.yml```, so we can just have it install ```make``` there.

bowlofeggs:
I did verify that ```d60f52c9cb04bfd4c5e950410611c746641f0b5f76830a53d44a0f1a43ab3fac``` is the correct hash, via F28 GPG signature.

bowlofeggs:
jenkies test

bowlofeggs (approved):
Excellent, everything seems to work. Thanks!

PR #2392, opened 2018-05-28
Add new API endpoint to retrieve greenwave's decision about an update and refactor code a little

pypingou:
This is a first PR on the path to adding waiving support to bodhi-cli.

This refactors a little bit of code and adds a new API endpoint where can be retrieved greenwave's decision about an update.

This is currently missing unit-tests, I'll work on them tomorrow, but if someone has a little time to go through it and give a basic ack/nack on the idea itself it would be appreciated.

Thanks

### DIFF
--- a/bodhi/server/models.py
+++ b/bodhi/server/models.py
@@ -1958,7 +1958,14 @@ class Update(Base):
         return json.dumps(self.greenwave_subject)
 
     def get_test_gating_info(self):
-        """Query Greenwave about this update and returned the information retrieved."""
+        """
+        Query Greenwave about this update and return the information retrieved.
+
+        Returns:
+            dict: The response from Greenwave for this update.
+        Raises:
+            BodhiException: When the ``greenwave_api_url`` is undefined in configuration.
+        """
         if not config.get('greenwave_api_url'):
             raise BodhiException('No greenwave_api_url specified')
 
--- a/bodhi/server/models.py
+++ b/bodhi/server/models.py
@@ -48,7 +48,7 @@ from bodhi.server.exceptions import BodhiException, LockedUpdateException
 from bodhi.server.util import (
     avatar as get_avatar, build_evr, flash_log, get_critpath_components,
     get_rpm_header, header, tokenize, pagure_api_get,
-    greenwave_api_post, waiverdb_api_post)
+    waiverdb_api_post)
 import bodhi.server.util
 
 if six.PY2:
@@ -1957,8 +1957,11 @@ class Update(Base):
         """
         return json.dumps(self.greenwave_subject)
 
-    def update_test_gating_status(self):
-        """Query Greenwave about this update and set the test_gating_status as appropriate."""
+    def get_test_gating_info(self):
+        """Query Greenwave about this update and returned the information retrieved."""
+        if not config.get('greenwave_api_url'):
+            raise BodhiException('No greenwave_api_url specified')
+
         # We retrieve updates going to testing (status=pending) and updates
         # (status=testing) going to stable.
         # If the update is pending, we want to know if it can go to testing
@@ -1974,7 +1977,11 @@ class Update(Base):
         }
         api_url = '{}/decision'.format(config.get('greenwave_api_url'))
 
-        decision = util.greenwave_api_post(api_url, data)
+        return util.greenwave_api_post(api_url, data)
+
+    def update_test_gating_status(self):
+        """Query Greenwave about this update and set the test_gating_status as appropriate."""
+        decision = self.get_test_gating_info()
         if decision['policies_satisfied']:
             # If an unrestricted policy is applied and no tests are required
             # on this update, let's set the test gating as ignored in Bodhi.
@@ -2647,15 +2654,7 @@ class Update(Base):
         if self.test_gating_passed:
             raise BodhiException("Can't waive test resuts on an update that passes test gating")
 
-        decision_context = u'bodhi_update_push_testing'
-        if self.status == UpdateStatus.testing:
-            decision_context = u'bodhi_update_push_stable'
-        data = {
-            'product_version': self.product_version,
-            'decision_context': decision_context,
-            'subject': self.greenwave_subject
-        }
-        decision = greenwave_api_post('{}/decision'.format(config.get('greenwave_api_url')), data)
+        decision = self.get_test_gating_info()
         for requirement in decision['unsatisfied_requirements']:
             data = {
                 'subject': requirement['item'],
--- a/bodhi/server/services/updates.py
+++ b/bodhi/server/services/updates.py
@@ -25,6 +25,7 @@ from cornice import Service
 from cornice.validators import colander_body_validator, colander_querystring_validator
 from sqlalchemy import func, distinct
 from sqlalchemy.sql import or_
+from requests import RequestException, Timeout as RequestsTimeout
 
 from bodhi.server import log, security
 from bodhi.server.exceptions import BodhiException, LockedUpdateException
@@ -91,6 +92,13 @@ update_waive_test_results = Service(
     description='Waive test results that block transitioning the update to next state',
     factory=security.PackagerACLFactory,
     cors_origins=bodhi.server.security.cors_origins_rw
+)
+
+update_get_test_results = Service(
+    name='update_get_test_results',
+    path='/updates/{id}/get-test-results',
+    description='Get test results for a specified update',
+    cors_origins=bodhi.server.security.cors_origins_ro,
 )
 
 
@@ -583,3 +591,41 @@ def waive_test_results(request):
         request.errors.add('body', 'request', str(e))
 
     return dict(update=update)
+
+
+@update_get_test_results.get(schema=bodhi.server.schemas.GetTestResultsSchema,
+                             validators=(validate_update_id),
+                             renderer='json',
+                             error_handler=bodhi.server.services.errors.json_handler)
+def get_test_results(request):
+    """
+    Get the test results on a given update when gating is on.
+
+    Args:
+        request (pyramid.request): The current request.
+    Returns:
+        dict: A dictionary mapping the key "update" to the update.
+    """
+    update = request.validated['update']
+
+    decision = None
+    try:
+        decision = update.get_test_gating_info()
+    except RequestsTimeout as e:
+        log.exception("Error querying greenwave for test results - timed out")
+        request.errors.add('body', 'request', str(e))
+        request.errors.status = 504
+    except (RequestException, RuntimeError) as e:
+        log.exception("Error querying greenwave for test results")
+        request.errors.add('body', 'request', str(e))
+        request.errors.status = 502
+    except BodhiException as e:
+        log.exception("Failed to query greenwave for test results")
+        request.errors.add('body', 'request', str(e))
+        request.errors.status = 501
+    except Exception as e:
+        log.exception("Unhandled exception in get_test_results")
+        request.errors.add('body', 'request', str(e))
+        request.errors.status = 500
+
+    return dict(decision=decision)
--- a/bodhi/tests/server/services/test_updates.py
+++ b/bodhi/tests/server/services/test_updates.py
@@ -26,6 +26,7 @@ import unittest
 from mock import ANY
 import koji
 import mock
+import requests
 import six
 from six.moves.urllib import parse as urlparse
 
@@ -35,6 +36,7 @@ from bodhi.server.models import (
     BuildrootOverride, Compose, Group, RpmPackage, ModulePackage, Release,
     ReleaseState, RpmBuild, Update, UpdateRequest, UpdateStatus, UpdateType,
     UpdateSeverity, User, TestGatingStatus)
+from bodhi.server.util import call_api
 from bodhi.tests.server.base import BaseTestCase, BodhiTestApp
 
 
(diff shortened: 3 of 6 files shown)


pypingou:
jenkies test

pypingou:
Jenkins trips over: ````rawhide	pkg_resources.DistributionNotFound: The 'pyflakes<1.7.0,>=1.5.0' distribution was not found and is required by flake8```` but the tests are passing otherwise

pypingou:
An added value to this approach is that it becomes way easier to mock the output from greenwave for local testing :)

pypingou:
Another benefit: it allows us to go around the CORS headers

bodhi/server/services/updates.py
```diff
@@ -94,6 +94,14 @@
     cors_origins=bodhi.server.security.cors_origins_rw
 )
 
+update_get_test_results = Service(
+    name='update_get_test_results',
+    path='/updates/{id}/get-test-results',
+    description='Get test results for a specified update',
+    factory=security.PackagerACLFactory,
+    cors_origins=bodhi.server.security.cors_origins_rw
```
bowlofeggs: I think we can make this a GET request instead of a POST, which would allow us to use ```cors_origins_ro``` here. What do you think?

bodhi/server/services/updates.py
```diff
@@ -584,3 +592,33 @@ def waive_test_results(request):
...
 
     return dict(update=update)
+
+
+@update_get_test_results.post(schema=bodhi.server.schemas.GetTestResultsSchema,
+                              validators=(colander_body_validator,
+                                          validate_update_id,
+                                          validate_acls),
+                              permission='edit', renderer='json',
+                              error_handler=bodhi.server.services.errors.json_handler)
```
bowlofeggs: Similar here - I think we can turn this into a ```GET``` and use a read permissions instead of edit, since it will not change any data in Bodhi or Greenwave.

bodhi/tests/server/services/test_updates.py
```diff
@@ -5167,3 +5167,82 @@ def test_unexpected_exception(self, log_exception, waive_test_results, *args):
         self.assertEquals(res.json_body['errors'][0]['description'],
                           u'IOError. oops!')
         log_exception.assert_called_once_with("Unhandled exception in waive_test_results")
+
+
+class TestGetTestResults(base.BaseTestCase):
+    """
+    This class contains tests for the get_test_results() function.
+    """
+    @unittest.skipIf(six.PY3, 'Not working with Python 3 yet')
```
bowlofeggs: Let's see if we can get this test to pass with Python 3 so we don't raise @frenzymadness's eyebrows :)

bodhi/tests/server/services/test_updates.py
```diff
@@ -5167,3 +5167,82 @@ def test_unexpected_exception(self, log_exception, waive_test_results, *args):
...
+        Ensure that we get an error if trying to get test results of an update
+        when there is no greenwave_api_url defined in the configuration.
+        """
+        nvr = u'bodhi-2.0-1.fc17'
+
+        up = self.db.query(Update).filter_by(title=nvr).one()
+        up.locked = False
+
+        post_data = dict(update=nvr,
+                         csrf_token=self.app.get('/csrf').json_body['csrf_token'])
```
bowlofeggs: There is a ```self.get_csrf_token()``` that you can use here that will work for Python 3 - the above call might not work with Python 3.

bodhi/tests/server/services/test_updates.py
```diff
@@ -5167,3 +5167,82 @@ def test_unexpected_exception(self, log_exception, waive_test_results, *args):
...
+    @mock.patch('bodhi.server.services.updates.log.exception')
+    def test_unexpected_exception(self, log_exception, get_test_gating_info, *args):
+        """Ensure that an unexpected Exception is handled by get_test_results()."""
+        nvr = u'bodhi-2.0-1.fc17'
+
+        up = self.db.query(Update).filter_by(title=nvr).one()
+        up.locked = False
+
+        post_data = dict(update=nvr, request='stable',
+                         csrf_token=self.app.get('/csrf').json_body['csrf_token'])
```
bowlofeggs: Similar here - I think we can get this working with Python 3.

bodhi/tests/server/services/test_updates.py
```diff
@@ -5167,3 +5167,82 @@ def test_unexpected_exception(self, log_exception, waive_test_results, *args):
...
+        res = self.app.post_json('/updates/%s/get-test-results' % str(nvr), post_data, status=400)
```
bowlofeggs: Shouldn't an unhandled Exception return a 500?

bodhi/tests/server/services/test_updates.py
```diff
@@ -5167,3 +5167,82 @@ def test_unexpected_exception(self, log_exception, waive_test_results, *args):
...
+                'subject': [
+                    {'item': u'bodhi-2.0-1.fc17', 'type': 'koji_build'},
+                    {'original_spec_nvr': u'bodhi-2.0-1.fc17'},
+                    {'item': u'FEDORA-2018-a3bbe1a8f2', 'type': 'bodhi_update'}
+                ]
+            },
+            method='POST',
+            retries=3,
+            service_name='Greenwave'
+        )
```
bowlofeggs, line 5234: Let's also assert that the return value is correct.

bodhi/server/services/updates.py
```diff
@@ -584,3 +592,33 @@ def waive_test_results(request):
...
+
+    decision = None
+    try:
+        decision = update.get_test_gating_info()
+    except BodhiException as e:
+        log.exception("Failed to query greenwave for test results")
+        request.errors.add('body', 'request', str(e))
+    except Exception as e:
+        log.exception("Unhandled exception in get_test_results")
+        request.errors.add('body', 'request', str(e))
```
bowlofeggs: We should set a 500 code here.

bodhi/server/services/updates.py
```diff
@@ -584,3 +592,33 @@ def waive_test_results(request):
...
+        dict: A dictionary mapping the key "update" to the update.
+    """
+    update = request.validated['update']
+
+    decision = None
+    try:
+        decision = update.get_test_gating_info()
+    except BodhiException as e:
+        log.exception("Failed to query greenwave for test results")
+        request.errors.add('body', 'request', str(e))
```
bowlofeggs: It would be good to catch the various ```requests``` ```Exceptions``` here so we can set our error code accordingly, such as if Greenwave is unreachable (HTTP 504) or our call to it times out (HTTP 502).

docs/server_api/updates.rst
```diff
@@ -3,4 +3,4 @@ Updates
 
 .. cornice-autodoc::
    :modules: bodhi.server.services.updates, bodhi.server.services.zz_redirects
-   :services: bodhi1_update_redirect, update, update_edit, update_request, updates, updates_rss
+   :services: bodhi1_update_redirect, update, update_edit, update_get_test_results, update_request, updates, updates_rss
```
bowlofeggs: Thanks for remembering this - I'd have forgotten!

bowlofeggs (changes requested):
This looks like a good direction I think. It'll make it so the CLI and web UI can both use the same code to get the data, which I think is kinda nice.

@ralphbean Do you think this would be good to use with your JS patch?

bodhi/tests/server/services/test_updates.py
pypingou: That would mean querying greenwave during the tests no?

bodhi/server/services/updates.py
pypingou: Adjusted

pypingou:
I think I got them all but one, let's see if Jenkins is happy with this now :)

ralphbean:
> Do you think this would be good to use with your JS patch?

Your call.  :)

bowlofeggs:
On 06/05/2018 08:55 AM, Ralph Bean wrote:
>     Do you think this would be good to use with your JS patch?
> 
> Your call. :)

I don't see a strong incentive to decide one way vs. the other at the
moment. Since you have a patch that works, let's just go with it as-is
for now. We can always refactor later if it becomes clear that there
would be advantage in doing so.

bodhi/tests/server/services/test_updates.py
bowlofeggs: We can use the mock to set the return value to some particular dictionary, like {'return': 'value'} and then just assert that the right data is returned. Otherwise there isn't a test that proves that the API actually does what it says it does.

pypingou:
On Tue, Jun 05, 2018 at 05:27:03PM +0000, Randy Barlow wrote:
>    We can use the mock to set the return value to some particular dictionary,
>    like {'return': 'value'} and then just assert that the right data is
>    returned. Otherwise there isn't a test that proves that the API actually
>    does what it says it does.

Ok, I'll add that :)

pypingou:
Tests are passing for me locally, I don't quite understand why they don't pass on jenkins anymore :(

The error I see:
````
rawhide	bodhi/tests/server/services/test_updates.py:5112: in <module>
rawhide	    class TestGetTestResults(base.BaseTestCase):
rawhide	E   NameError: name 'base' is not defined
````
is most cryptic, all classes in that file inherit from base and that class definition is not line 5112 but line 5286, so I don't understand what's going on there...

**Comment by bowlofeggs** [comment]
@pypingou The CI system rebases your PR on develop before testing it to make sure your patch works with the latest HEAD.

**Comment by pypingou** [comment]
@bowlofeggs alright, I see the change now, somehow I thought I had rebased maybe not recently enough :)

File: bodhi/server/models.py [hunk 11]
```diff
@@ -1951,8 +1951,11 @@ def greenwave_subject_json(self):
         """
         return json.dumps(self.greenwave_subject)
 
-    def update_test_gating_status(self):
-        """Query Greenwave about this update and set the test_gating_status as appropriate."""
+    def get_test_gating_info(self):
+        """Query Greenwave about this update and returned the information retrieved."""
```
bowlofeggs: This method now has a return value and can raise a ```BodhiException```, so we should document both of these.

```
"""
Query Greenwave about this update.

Returns:
    dict: The response from Greenwave for this update.
Raises:
    BodhiException: When the ```greenwave_api_url``` is undefined in configuration.
"""
```

bodhi/server/services/updates.py
```diff
@@ -583,3 +592,37 @@ def waive_test_results(request):
...
+    Args:
+        request (pyramid.request): The current request.
+    Returns:
+        dict: A dictionary mapping the key "update" to the update.
+    """
+    update = request.validated['update']
+
+    decision = None
+    try:
+        decision = update.get_test_gating_info()
```
bowlofeggs, line 613: It would be good to also explicitly catch ```requests.exceptions.Timeout``` and set a 504 when that happens.

bodhi/tests/server/services/test_updates.py
```diff
@@ -5106,3 +5107,122 @@ def test_unexpected_exception(self, log_exception, waive_test_results, *args):
...
+
+        self.assertEquals(res.json_body, {u'decision': {u'foo': u'bar'}})
```
bowlofeggs: Let's add one more test for a 504 gateway timeout. This will help us differentiate between problems if Bodhi can't communicate with Greenwave.

bowlofeggs (changes requested):

bodhi/server/models.py

bodhi/server/services/updates.py
pypingou: The reason I caught the high level exception is that if we set a 504 when querying greenwave, we won't be able to distinguish between a 504 coming from bodhi vs a 504 coming from greenwave.
So I went with: if the error comes from greenwave, always return a 502 and include in the JSON returned the exact error (500, 504, 502...).

The test at: https://github.com/fedora-infra/bodhi/pull/2392/files#diff-1ff79f6cd36ea2ba9a09899a2799ae1dR5154 shows this behavior.

I can definitely adjust if you prefer though

bowlofeggs:
On 06/07/2018 04:21 AM, Pierre-Yves Chibon wrote:
> The reason I caught the high level exception is that if we set a 504
> when querying greenwave, we won't be able to distinguish between a 504
> coming from bodhi vs a 504 coming from greenwave.

Bodhi returning a 504 means that Bodhi was unable to perform a requested
action due to being unable to get a response from a remote service in a
reasonable amount of time. What would it mean for there to be "a 504
coming from Bodhi"? I can't think of a reason for Bodhi to return a 504,
other than cases like this specific one.

pypingou:
Adjusted

bodhi/server/models.py
```diff
@@ -1951,8 +1951,18 @@ def greenwave_subject_json(self):
         """
         return json.dumps(self.greenwave_subject)
 
-    def update_test_gating_status(self):
-        """Query Greenwave about this update and set the test_gating_status as appropriate."""
+    def get_test_gating_info(self):
+        """
+        Query Greenwave about this update and returned the information retrieved.
```
bowlofeggs: s/returned/return/

bodhi/server/models.py
```diff
@@ -1951,8 +1951,18 @@ def greenwave_subject_json(self):
...
+
+        Returns:
+            dict: The response from Greenwave for this update.
+        Raises:
+            BodhiException: When the ```greenwave_api_url``` is undefined in configuration.
```
bowlofeggs: The docs are rendered in RST, which uses 2 `'s instead of 3 for monospace fonts.

bodhi/server/services/updates.py
```diff
@@ -93,6 +94,14 @@
     cors_origins=bodhi.server.security.cors_origins_rw
 )
 
+update_get_test_results = Service(
+    name='update_get_test_results',
+    path='/updates/{id}/get-test-results',
+    description='Get test results for a specified update',
+    factory=security.PackagerACLFactory,
```
bowlofeggs: Why do we want to limit this query to packagers?

bodhi/tests/server/services/test_updates.py
```diff
@@ -5106,3 +5107,167 @@ def test_unexpected_exception(self, log_exception, waive_test_results, *args):
...
+        """
+        Ensure that we get an error if trying to get test results of an update
+        when there is no greenwave_api_url defined in the configuration.
+        """
+        nvr = u'bodhi-2.0-1.fc17'
+
+        up = self.db.query(Update).filter_by(title=nvr).one()
+        up.locked = False
+
+        res = self.app.get('/updates/%s/get-test-results' % str(nvr), status=400)
```
bowlofeggs: 4xx codes indicate that the client made an error, and specifically 400 means the request is bad. In this case, the error is that the server isn't configured to talk to Greenwave. I suggest a 501 for not implemented.

bodhi/server/services/updates.py
```diff
@@ -583,3 +592,41 @@ def waive_test_results(request):
...
+        log.exception("Error querying greenwave for test results - timed out")
+        request.errors.add('body', 'request', str(e))
+        request.errors.status = 504
+    except RequestException as e:
+        log.exception("Error querying greenwave for test results")
+        request.errors.add('body', 'request', str(e))
+        request.errors.status = 502
+    except BodhiException as e:
+        log.exception("Failed to query greenwave for test results")
+        request.errors.add('body', 'request', str(e))
```
bowlofeggs, line 624: What raises this Exception in the stack?

bodhi/server/services/updates.py
```diff
@@ -583,3 +592,41 @@ def waive_test_results(request):
...
+    try:
+        decision = update.get_test_gating_info()
+    except RequestsTimeout as e:
+        log.exception("Error querying greenwave for test results - timed out")
+        request.errors.add('body', 'request', str(e))
+        request.errors.status = 504
+    except RequestException as e:
+        log.exception("Error querying greenwave for test results")
+        request.errors.add('body', 'request', str(e))
+        request.errors.status = 502
```
bowlofeggs: ```call_api()``` will raise a ```RuntimeError``` if Greenwave returns a non-200 code. We should catch that and turn it into a ```502```.

bodhi/tests/server/services/test_updates.py
```diff
@@ -5106,3 +5107,167 @@ def test_unexpected_exception(self, log_exception, waive_test_results, *args):
...
+                    {'original_spec_nvr': u'bodhi-2.0-1.fc17'},
+                    {'item': u'FEDORA-2018-a3bbe1a8f2', 'type': 'bodhi_update'}
+                ]
+            },
+            method='POST',
+            retries=3,
+            service_name='Greenwave'
+        )
+
+        self.assertEquals(res.json_body, {u'decision': {u'foo': u'bar'}})
```
bowlofeggs: We should also add a test for when Greenwave returns an HTTP 500 code.

bowlofeggs (changes requested):

bodhi/server/services/updates.py
pypingou: done

bodhi/server/services/updates.py
pypingou:
````
        if not config.get('greenwave_api_url'):
            raise BodhiException('No greenwave_api_url specified')
````

File: bodhi/tests/server/services/test_updates.py (hunk 17 above)
Comment (pypingou) [reply bodhi/tests/server/services/test_updates.py]: sure if you want...

File: bodhi/server/services/updates.py (hunk 16 above)
Comment (pypingou) [reply bodhi/server/services/updates.py]: removed

File: bodhi/server/models.py (hunk 15 above)
Comment (pypingou) [reply bodhi/server/models.py]: fixed

File: bodhi/server/models.py (hunk 14 above)
Comment (pypingou) [reply bodhi/server/models.py]: fixed

File: bodhi/tests/server/services/test_updates.py (hunk 20 above)
Comment (pypingou) [reply bodhi/tests/server/services/test_updates.py]: Done

**Comment by pypingou** [comment]
Rebased

**Review by bowlofeggs — CHANGES_REQUESTED** [review]
This service doesn't appear to work for nonauthenticated users:

```
[vagrant@bodhi-dev bodhi]$ curl -I http://localhost:6543/updates/FEDORA-2018-fe64338e71/get-test-results
HTTP/1.1 403 Forbidden
Content-Length: 10348
Content-Type: text/html; charset=UTF-8
Date: Thu, 14 Jun 2018 17:41:16 GMT
Server: waitress
X-Content-Type-Options: nosniff
```

I think it is because you are requiring 'read' permission on the view, but non-authenticated users don't have permissions.

**Comment by pypingou** [comment]
@bowlofeggs anything else on this PR?

**Comment by pypingou** [comment]
Oh sorry, just saw your comment

**Comment by bowlofeggs** [comment]
This PR does not work:

```
[vagrant@bodhi-dev bodhi]$ curl -I http://localhost:6543/updates/FEDORA-2018-64b48748a1/get-test-results
HTTP/1.1 403 Forbidden
Content-Length: 10348
Content-Type: text/html; charset=UTF-8
Date: Mon, 18 Jun 2018 14:11:43 GMT
Server: waitress
X-Content-Type-Options: nosniff
```

**Comment by pypingou** [comment]
I have the fix for this making the curl command work for me as well, but I'd very much like to have a test checking this, would you have any pointer?

**Comment by pypingou** [comment]
jenkies test

**Comment by pypingou** [comment]
jenkies test

File: bodhi/server/services/updates.py [hunk 21]
```diff
@@ -93,6 +94,13 @@
     cors_origins=bodhi.server.security.cors_origins_rw
 )
 
+update_get_test_results = Service(
+    name='update_get_test_results',
+    path='/updates/{id}/get-test-results',
```
cverna, line 99: just a nitpick here but the endpoint could just be `/updates/{id}/test-results`, I think removing the `get-`here makes it nicer :)

pypingou: I went with this because the waive endpoint is : ``path='/updates/{id}/waive-test-results',``, I can adjust it if we want.

cverna: As I said it was a nitpick,I think it is fine to leave it as it is. 

But it would have been nice to have only one endpoint `test-results` using the `POST` method to waive and the `GET` method to get the results.

cverna:
I just successfully tested this PR in the vagrant dev environment using the following commands 
```
[vagrant@bodhi-dev bodhi]$ curl -I http://localhost:6543/updates/FEDORA-2018-64b48748a1/get-test-results
HTTP/1.1 200 OK
Content-Length: 213
Content-Type: application/json
Date: Wed, 27 Jun 2018 10:37:09 GMT
Server: waitress
X-Content-Type-Options: nosniff
```
and 
```
[vagrant@bodhi-dev bodhi]$ curl  http://localhost:6543/updates/FEDORA-2018-64b48748a1/get-test-results
{"decision": {"applicable_policies": ["taskotron_release_critical_tasks_for_testing", "atomic_ci_pipeline_results"], "policies_satisfied": true, "summary": "no tests are required", "unsatisfied_requirements": []}}
```
@pypingou if you are done with this PR, I ll be happy to merge it \o/

bodhi/server/services/updates.py
pypingou: I am unopiniated on this, I tried following the current style but I'm ok with changing if that's what we want.

pypingou:
> @pypingou if you are done with this PR, I ll be happy to merge it \o/

I'm very much looking forward for this :)

bowlofeggs:

bowlofeggs:

Issue #2370, opened 2018-05-09
Bodhi should live query greenwave


ralphbean:
Talking with @bowlofeggs today, we think that the database table cache, and the cronjob and the fedmsg updater are not worth it for the greenwave integration.

Instead, let's do a few things:

- [x] Replace the frontend display of resultsdb results and greenwave decision with a live ajax query to greenwave to get both the results and the decision.
- [ ] Replace the backend query of the cache at compose time with a live http query to greenwave to get the decision.
- [ ] The cornice service that accepts requests to push to stable or batched needs to also live query greenwave to know if the packager's request is allowed.
- [ ] Delete the cronjob.  Delete the database model and add a migration to drop that biz.

Tests, of course.

bowlofeggs:
If we do this refactor as proposed, I believe #2364 should be automatically fixed.

bowlofeggs:
Thanks to @puiterwijk for placing the thought of this design into my head ☺

bowlofeggs:
Technically removing the test_gating_status from the DB will also remove it from the API, unless we make the API also live query Greenwave. I'm OK with removing it, but we should note that if we do that it will be a backwards incompatible change. I've been piling up quite a few of those for a while, so maybe it's just time to make Bodhi 4 now.

ralphbean:
More like Bodhi 4000 :rocket:.

bowlofeggs:
Question: would it be interesting to have the Bodhi backend live-query Greenwave instead of doing it in JS? Pros/cons:

Pro: We can share more code between the CLI and the web UI for test gating - 1 implementation instead of 2.
Pro: IMO, the mako templating language is nicer to work with than creating HTML with JS
Con: Bodhi's response will be slower than it is now.

@puiterwijk suggested a hybrid of giving Bodhi and endpoint that queries greenwave and returns. Similar pro/con as above. At first I thougth you would lose the second pro, but the JS could ask for HTML from the new endpoint so we could still have it render a template if we want.

bowlofeggs:
No matter how we do this, we will need to make sure Bodhi does sane things when Greenwave doesn't give back a response in time, or when it gives back an error.

bowlofeggs:
[...]

PR #2393, opened 2018-05-29
Query greenwave with JS.

ralphbean:
This changes the frontend only for #2370.

Signed-off-by: Ralph Bean <rbean@redhat.com>


ralphbean:
Doh, when I try to `btest` I get:

```
Running Sphinx v1.6.6
making output directory...

Extension error:
Could not import extension cornice_sphinx (exception: No module named cornice_sphinx)
make: *** [Makefile:50: html] Error 1
make: Leaving directory '/home/vagrant/bodhi/docs'
Shared connection to 192.168.121.101 closed.
```

ralphbean:
Huh.. `flake8-2` in Jenkins flipped out with:

```
rawhide	pkg_resources.DistributionNotFound: The 'pyflakes<1.7.0,>=1.5.0' distribution was not found and is required by flake8
```

pypingou:
In https://github.com/fedora-infra/bodhi/pull/2392 I'm adding a new API endpoint to query greenwave so that the CLI can also get this information without having to specify which greenwave instance/endpoint to query and without duplicating the logic in the server and the CLI on how to create the subjects and all.

Should we take advantage of this new endpoint here as well?

ralphbean:
> Should we take advantage of this new endpoint here as well?

Hm.  That would reduce the number of places we have to specify that greenwave URL at the cost of increased network latency.

I leave it to you guys to decide.  I can go either way.  :)

ralphbean:
Here's a video showing it in action:

http://threebean.org/img/live-query.webm

bowlofeggs:
jenkies test

bodhi/server/templates/update.html
```diff
@@ -37,62 +35,62 @@
       ABSENT: 'question-circle',
     }
 
-    // XXX - For development, we typically have updates in our development
-    // bodhi DB from a snapshot from over a year ago -- before the deployment
-    // of taskotron.  As such, taskotron doesn't know about the updates that
-    // we ask about.  Therefore, it is useful sometimes to mess around here
-    // and hardcode something like this:
-    //var update = 'ugene-1.14.2-1.fc21';
```
bowlofeggs, line 47: Getting out the old dusty broom - +1

bodhi/server/templates/update.html
```diff
@@ -365,26 +334,17 @@
...
     $("#tab-automatedtests").append(" <span class='fa fa-spinner fa-spin'></span>");
 
-    for (index = 0, len = builds.length; index < len; ++index) {
-      request_results(base_url+"results/latest?original_spec_nvr="
-        +builds[index]+"&testcases=org.centos.prod.ci.pipeline.complete");
-      request_results(base_url+"results/latest?item="+builds[index]+"&type=koji_build&testcases:like=dist.*");
-    }
+    get_greenwave_information({
+        'product_version': '${update.product_version}',
+        'decision_context': 'bodhi_update_push_stable',
```
bowlofeggs: In the Bodhi backend code, there is also a query to Greenwave to assert that it's OK to push an update to testing when the update is pending. Should we do that here as well?

bowlofeggs (commented):
I'm fine either way with having this patch use the new API from @pypingou's patch, or having it query Greenwave directly like this. If we go with a JS implementation, I may later refactor it to use the API, but I'm kind of on the fence. The pros and cons don't strongly sway me one way or the other.

bodhi/server/templates/update.html
ralphbean: Hm, yes.

ralphbean: See f2a4919b82c685ce9038e1bd536bec214b928e32.

bowlofeggs:
The test failures seem to be mostly (if not all) because there were tests for the removed Python function: https://ci.centos.org/job/bodhi-bodhi/1539/ so we can probably just drop those tests.

I tested the PR against production Greenwave and it seems to work, though of course production Greenwave is currently configured to pass all updates. Have you been able to test this against updates that fail Greenwave gating too?

bowlofeggs:
Let's add a note to ```docs/user/release_notes.rst``` that mentions that Greenwave is now a required service for Bodhi deployments that wish to continue displaying test results in the UI.

bowlofeggs:
There seems to be one failing test that is asserting that the template tells the developer that the update cannot be pushed.

ralphbean:
> There seems to be one failing test that is asserting that the template tells the developer that the update cannot be pushed.

Ack - haven't had a chance to resolve that one yet.

ralphbean:
> There seems to be one failing test that is asserting that the template tells the developer that the update cannot be pushed.

Should be happy now.  :)

ralphbean:
> Have you been able to test this against updates that fail Greenwave gating too?

Yes.  :)

> I tested the PR against production Greenwave and it seems to work, though of course production Greenwave is currently configured to pass all updates.

Try staging.  There are actual requirements in the staging greenwave policies.

ralphbean:
Rebased against `develop`.

ralphbean:
Rebased against develop again.  :)

bowlofeggs:
Would you care to update your commit message to include this?

```
fixes #2425
```

Seems like you saved us from another issue!

ralphbean:
@bowlofeggs, done in e7cb406.

ralphbean:
@bowlofeggs, @pypingou - anything else to change here?

bowlofeggs (approved):
I spot checked this PR against a handful of updates and I didn't notice any problems. However, I didn't find any updates that had missing or wavied tests in staging Greenwave. I will take your word that you tested those scenarios and merge, since Bodhi doesn't currently handle those scenarios well today anyway.

bowlofeggs:
Thanks for the patch!

bowlofeggs:

bowlofeggs:
These patches were deployed to Fedora's production Bodhi instance yesterday.

AdamWill:
I think this broke display of openQA results. At least, it's the best suspect I have. openQA is still happily testing updates and forwarding the results to resultsdb, but Bodhi's Automated Tests tab is not showing them. e.g. [here](https://bodhi.fedoraproject.org/updates/FEDORA-2018-87ba0312c2) - there certainly are [results for that update in ResultsDB](https://taskotron.fedoraproject.org/resultsdb/results?item:like=FEDORA-2018-87ba0312c2*).

At a quick glance it looks like this changed Bodhi from querying resultsdb directly to querying Greenwave and relying on Greenwave to give it all the relevant results from resultsdb. I guess Greenwave is not picking up openQA results at present. :/

bowlofeggs:
Yeah looks like Greenwave isn't returning that data. No match here:

```
curl 'https://greenwave-web-greenwave.app.os.fedoraproject.org/api/v1.0/decision' -H 'User-Agent: Mozilla/5.0 (X11; Fedora; Linux x86_64; rv:63.0) Gecko/20100101 Firefox/63.0' -H 'Accept: */*' -H 'Accept-Language: en-US,en;q=0.5' --compressed -H 'Content-Type: application/json' -H 'Origin: https://bodhi.fedoraproject.org' -H 'DNT: 1' -H 'Connection: keep-alive' -H 'Cache-Control: max-age=0' --data '{"product_version":"fedora-29","decision_context":"bodhi_update_push_stable","subject":[{"item":"kernel-4.19.5-300.fc29","type":"koji_build"},{"item":"kernel-headers-4.19.5-300.fc29","type":"koji_build"},{"item":"kernel-tools-4.19.5-300.fc29","type":"koji_build"},{"original_spec_nvr":"kernel-4.19.5-300.fc29"},{"original_spec_nvr":"kernel-headers-4.19.5-300.fc29"},{"original_spec_nvr":"kernel-tools-4.19.5-300.fc29"},{"item":"FEDORA-2018-87ba0312c2","type":"bodhi_update"}],"verbose":true}' | grep -i openqa
```

AdamWill:
Yeah, I think greenwave has simply never considered openqa results because we never tried to gate on them. It was set up to pull in taskotron and pipeline results as we tried to do gating based on tests from both of those, but not openqa. So when bodhi stopped querying resultsdb directly and started just querying greenwave, we lost the openqa results.

The fix should be a change to the fedora greenwave policy in infra ansible git, I think. I'll try and figure that out this afternoon.

AdamWill:
[Done](https://infrastructure.fedoraproject.org/cgit/ansible.git/commit/?id=0bc53e1122fdfaea3b9e670a64a01489613ccfea), and it works. yay.

PR #2407, opened 2018-06-01
Replaced hardcoded link for package build with config variable

### LINKED ISSUE #2182: The Koji URL is hardcoded in the update.html template
I noticed today that the staging Bodhi instance was linking to the production Koji instance, and it turned out that the Koji URL is hardcoded in the [template](https://github.com/fedora-infra/bodhi/blob/3.4.0/bodhi/server/templates/update.html#L1013). We should use the value from settings instead.

ghost:
fixes #2182

### DIFF
--- a/bodhi/server/config.py
+++ b/bodhi/server/config.py
@@ -443,6 +443,9 @@ class BodhiConfig(dict):
         'waiverdb.access_token': {
             'value': None,
             'validator': _validate_none_or(six.text_type)},
+        'koji_web_url': {
+            'value': 'https://koji.fedoraproject.org/koji/',
+            'validator': _validate_tls_url},
         'koji_hub': {
             'value': 'https://koji.stg.fedoraproject.org/kojihub',
             'validator': str},
--- a/bodhi/server/templates/update.html
+++ b/bodhi/server/templates/update.html
@@ -1,6 +1,12 @@
 <%inherit file="master.html"/>
 <%namespace name="captcha" module="bodhi.server.captcha"/>
 <%namespace name="json" module="json"/>
+<%
+  try:
+    from urllib.parse import urljoin
+  except ImportError:
+    from urlparse import urljoin
+%>
 
 <%block name="pagetitle">
 % if update.alias:
@@ -1033,7 +1039,11 @@ $(document).ready(function(){
         % for build in update.builds:
         <tr class="media">
           <td>
-            <a href="https://koji.fedoraproject.org/koji/search?terms=${build.nvr}&type=build&match=glob" target="_blank">
+            <%
+              koji_web_url = request.registry.settings.get('koji_web_url').strip('/') + '/'
+              build_url = urljoin(koji_web_url, 'search?terms='+build.nvr+'&type=build&match=glob')
+            %>
+            <a href="${build_url}" target="_blank">
                 ${build.nvr}</a>
           </td>
           <td class="pull-right">
--- a/bodhi/tests/server/services/test_updates.py
+++ b/bodhi/tests/server/services/test_updates.py
@@ -291,6 +291,52 @@ class TestNewUpdate(BaseTestCase):
         self.assertEquals(up['errors'][0]['description'],
                           "Koji error getting build: bodhi-2.0.0-2.fc17")
 
+    @mock.patch(**mock_valid_requirements)
+    @mock.patch('bodhi.server.notifications.publish')
+    def test_koji_config_url(self, publish, *args):
+        """
+        Test html rendering of default build link
+        """
+        self.app.app.registry.settings['koji_web_url'] = u'https://koji.fedoraproject.org/koji/'
+        nvr = u'bodhi-2.0.0-2.fc17'
+        resp = self.app.post_json('/updates/', self.get_update(nvr))
+
+        resp = self.app.get('/updates/%s' % nvr, headers={'Accept': 'text/html'})
+
+        self.assertRegexpMatches(str(resp), ('https://koji.fedoraproject.org/koji'
+                                             '/search\?terms=.*\&amp;type=build\&amp;match=glob'))
+
+    @mock.patch(**mock_valid_requirements)
+    @mock.patch('bodhi.server.notifications.publish')
+    def test_koji_config_url_without_trailing_slash(self, publish, *args):
+        """
+        Test html rendering of default build link without trailing slash
+        """
+        self.app.app.registry.settings['koji_web_url'] = u'https://koji.fedoraproject.org/koji'
+        nvr = u'bodhi-2.0.0-2.fc17'
+        resp = self.app.post_json('/updates/', self.get_update(nvr))
+
+        resp = self.app.get('/updates/%s' % nvr, headers={'Accept': 'text/html'})
+
+        self.assertRegexpMatches(str(resp), ('https://koji.fedoraproject.org/koji'
+                                             '/search\?terms=.*\&amp;type=build\&amp;match=glob'))
+
+    @mock.patch(**mock_valid_requirements)
+    @mock.patch('bodhi.server.notifications.publish')
+    def test_koji_config_mock_url_without_trailing_slash(self, publish, *args):
+        """
+        Test html rendering of build link using a mock config variable 'koji_web_url'
+        without a trailing slash in it
+        """
+        self.app.app.registry.settings['koji_web_url'] = u'https://host.org'
+        nvr = u'bodhi-2.0.0-2.fc17'
+        resp = self.app.post_json('/updates/', self.get_update(nvr))
+
+        resp = self.app.get('/updates/%s' % nvr, headers={'Accept': 'text/html'})
+
+        self.assertRegexpMatches(str(resp), ('https://host.org'
+                                             '/search\?terms=.*\&amp;type=build\&amp;match=glob'))
+
     @mock.patch.dict('bodhi.server.validators.config', {'acl_system': u'dummy'})
     @mock.patch(**mock_uuid4_version1)
     @mock.patch(**mock_valid_requirements)
--- a/production.ini
+++ b/production.ini
@@ -303,6 +303,9 @@ use = egg:bodhi-server
 # want to use 'koji'.
 # buildsystem = dev
 
+# The base URL to Koji, used to construct HTML links to Koji builds in the web UI
+# koji_web_url = https://koji.fedoraproject.org/koji/
+
 # Koji's XML-RPC hub
 # koji_hub = https://koji.stg.fedoraproject.org/kojihub
 


bodhi/server/config.py
```diff
@@ -443,6 +443,9 @@ class BodhiConfig(dict):
         'waiverdb.access_token': {
             'value': None,
             'validator': _validate_none_or(six.text_type)},
+        'koji_build_web_url': {
```
bowlofeggs: Let's name this more simply as ```koji_web_url``` - it's redundant to include "build" since Koji is a build system.

bodhi/server/config.py
```diff
@@ -443,6 +443,9 @@ class BodhiConfig(dict):
...
+            'value': 'https://koji.fedoraproject.org/koji/search',
```
bowlofeggs: Let's make this default to ```https://koji.fedoraproject.org```. That way we can use it for other Koji things if needed later, and also the name doesn't imply that it's just a URL to searching.

bodhi/server/templates/update.html
```diff
@@ -1045,7 +1045,7 @@ <h3>Related Bugs <span class="badge">${len(update.bugs)}</span></h3>
         % for build in update.builds:
         <tr class="media">
           <td>
-            <a href="https://koji.fedoraproject.org/koji/search?terms=${build.nvr}&type=build&match=glob" target="_blank">
+            <a href="${request.registry.settings.get('koji_build_web_url')}?terms=${build.nvr}&type=build&match=glob" target="_blank">
```
bowlofeggs: We need to do a ```urljoin``` here, so that we are forgiving if the admin does or does not set ```koji_web_url``` with a trailing slash. We will also need to include the ```/koji/search``` here since I suggested removing that from the setting above.

bodhi/tests/server/services/test_updates.py
```diff
@@ -309,6 +309,20 @@ def test_new_rpm_update_koji_error(self, publish, *args):
...
+        """
+        Test html rendering of koji build link using a mock config variable 'koji_build_web_url'
+        """
+        self.app.app.registry.settings['koji_build_web_url'] = u'build_host/param'
+
+        nvr = u'bodhi-2.0.0-2.fc17'
+        resp = self.app.post_json('/updates/', self.get_update(nvr))
+        resp = self.app.get('/updates/%s' % nvr, headers={'Accept': 'text/html'})
+
+        self.assertIn(u'build_host/param', resp)
```
bowlofeggs: Let's assert that the full URL is in there, including the /koji/search and the GET parameters that come after it.

Let's also add one more test to assert that the setting works with or without a trailing slash.

bowlofeggs:
jenkies test

bowlofeggs (changes requested):

bowlofeggs:
One more thing - we should also add this setting to ```production.ini```, with a comment over it describing what it is.

bowlofeggs:
This PR is failing pydocstyle tests.

production.ini
```diff
@@ -303,6 +303,9 @@ use = egg:bodhi-server
 # want to use 'koji'.
 # buildsystem = dev
 
+# Koji's web URL
```
bowlofeggs: Let's give this more of a description than this - this comment is the exact same text as the name of the variable.

bodhi/server/util.py
```diff
@@ -1473,3 +1474,19 @@ def _container_image_url(build, registry, tag=None):
...
+    https://koji.fedoraproject.org/koji/search?terms=nginx-1.12-20180523111257.9c690d0e&type=build&match=glob
+
+    Args:
+        base (str): The base URL of the host.
+        href (str): The parameters component of the URL.
+
+    Returns:
+        str: A URL obtained by joining the two components.
+    """
+    return urlparse.urljoin(base, href)
```
bowlofeggs: I thought we had talked about using ```urljoin()``` directly in the template? It doesn't make much sense to create a wrapper function around it.

bodhi/server/util.py
```diff
@@ -1473,3 +1474,19 @@ def _container_image_url(build, registry, tag=None):
         tag = '{}-{}'.format(build.nvr_version, build.nvr_release)
 
     return 'docker://{}/{}:{}'.format(registry, image_name, tag)
+
+def join_url(self, base, href):
```
bowlofeggs: This is a function, not a method, so it should not accept a ```self``` parameter.

bowlofeggs (changes requested):

bodhi/server/templates/update.html
```diff
@@ -1033,7 +1033,13 @@ <h3>Related Bugs <span class="badge">${len(update.bugs)}</span></h3>
         % for build in update.builds:
         <tr class="media">
           <td>
-            <a href="https://koji.fedoraproject.org/koji/search?terms=${build.nvr}&type=build&match=glob" target="_blank">
+            <%
+              import urlparse
```
bowlofeggs: I'm not too familiar with mako templates, but I think we can probably move this import statement into the top of the template in another ```<% %>``` block, to more closely resemble PEP-8 Python import style.

bodhi/server/templates/update.html
```diff
@@ -1033,7 +1033,13 @@ <h3>Related Bugs <span class="badge">${len(update.bugs)}</span></h3>
...
+
+              build_url = urlparse.urljoin(request.registry.settings.get('koji_web_url'),
+                                           'koji/search?terms='+build.nvr+'&type=build&match=glob')
```
bowlofeggs: It looks like Fedora's Koji is configured to be served at ```https://koji.fedoraproject.org/koji/```, which would mean that the ```koji_web_url``` setting should be set to that value, which I think means that we should drop ```koji/``` from this join.

bodhi/tests/server/services/test_updates.py
```diff
@@ -291,6 +291,51 @@ def test_new_rpm_update_koji_error(self, publish, *args):
...
+    @mock.patch('bodhi.server.notifications.publish')
+    def test_koji_config_url(self, publish, *args):
+        """
+        Test html rendering of default build link
+        """
+        self.app.app.registry.settings['koji_web_url'] = u'https://koji.fedoraproject.org'
+
+        nvr = u'bodhi-2.0.0-2.fc17'
+        resp = self.app.post_json('/updates/', self.get_update(nvr))
+        resp = self.app.get('/updates/%s' % nvr, headers={'Accept': 'text/html'})
```
bowlofeggs: Let's add a space above this line, since we are testing the response from the GET call here.

bodhi/tests/server/services/test_updates.py
```diff
@@ -291,6 +291,51 @@ def test_new_rpm_update_koji_error(self, publish, *args):
...
                           "Koji error getting build: bodhi-2.0.0-2.fc17")
 
+    @mock.patch(**mock_valid_requirements)
+    @mock.patch('bodhi.server.notifications.publish')
+    def test_koji_config_url(self, publish, *args):
+        """
+        Test html rendering of default build link
+        """
+        self.app.app.registry.settings['koji_web_url'] = u'https://koji.fedoraproject.org'
+
```
bowlofeggs: Let's drop this line, since the lines below it are really set up for the test.

bodhi/tests/server/services/test_updates.py
```diff
@@ -291,6 +291,51 @@ def test_new_rpm_update_koji_error(self, publish, *args):
...
+    def test_koji_config_mock_url_without_trailing_slash(self, publish, *args):
+        """
+        Test html rendering of build link using a mock config variable 'koji_web_url'
+        without a trailing slash in it
+        """
+        self.app.app.registry.settings['koji_web_url'] = u'https://host.org'
+
+        nvr = u'bodhi-2.0.0-2.fc17'
+        resp = self.app.post_json('/updates/', self.get_update(nvr))
+        resp = self.app.get('/updates/%s' % nvr, headers={'Accept': 'text/html'})
```
bowlofeggs: Similar comments here regarding the spaces.

bodhi/tests/server/services/test_updates.py
```diff
@@ -291,6 +291,51 @@ def test_new_rpm_update_koji_error(self, publish, *args):
...
+    def test_koji_config_mock_url_with_trailing_slash(self, publish, *args):
+        """
+        Test html rendering of build link using a mock config variable 'koji_web_url' and
+        with a trailing slash in it
+        """
+        self.app.app.registry.settings['koji_web_url'] = u'https://host.org/'
+
+        nvr = u'bodhi-2.0.0-2.fc17'
+        resp = self.app.post_json('/updates/', self.get_update(nvr))
+        resp = self.app.get('/updates/%s' % nvr, headers={'Accept': 'text/html'})
```
bowlofeggs: And here.

bodhi/server/config.py
```diff
@@ -443,6 +443,9 @@ class BodhiConfig(dict):
         'waiverdb.access_token': {
             'value': None,
             'validator': _validate_none_or(six.text_type)},
+        'koji_web_url': {
+            'value': 'https://koji.fedoraproject.org',
```
bowlofeggs: Let's set the default to ```https://koji.fedoraproject.org/koji/```, since that is where it is hosted.

bodhi/server/config.py
```diff
@@ -443,6 +443,9 @@ class BodhiConfig(dict):
...
+            'validator': str},
```
bowlofeggs: ```urljoin()``` will behave differently if this setting ends in a ```/``` than if it does not:

```
>>> u('https://example.com/one', 'thing/other_thing')
'https://example.com/thing/other_thing'
>>> u('https://example.com/one/', 'thing/other_thing')
'https://example.com/one/thing/other_thing'
```

To avoid exposing Bodhi admins to this, let's add a new validator in this module that ensures that the setting ends in a ```/```. This way, Bodhi will allow the admin to set ```https://koji.fedoraproject.org/koji``` here and it should still work.

production.ini
```diff
@@ -303,6 +303,9 @@ use = egg:bodhi-server
 # want to use 'koji'.
 # buildsystem = dev
 
+# The base URL to Koji, used to construct HTML links to Koji builds in the web UI
+# koji_web_url = https://koji.fedoraproject.org
```
bowlofeggs: Let's change this to ```https://koji.fedoraproject.org/koji/``` as well.

bodhi/tests/server/services/test_updates.py
```diff
@@ -291,6 +291,51 @@ def test_new_rpm_update_koji_error(self, publish, *args):
...
+
+        self.assertRegexpMatches(str(resp), ('https://koji.fedoraproject.org/koji'
+                                             '/search\?terms=.*\&amp;type=build\&amp;match=glob'))
+
+    @mock.patch(**mock_valid_requirements)
+    @mock.patch('bodhi.server.notifications.publish')
+    def test_koji_config_mock_url_without_trailing_slash(self, publish, *args):
+        """
+        Test html rendering of build link using a mock config variable 'koji_web_url'
+        without a trailing slash in it
```
bowlofeggs, line 329: The test above does not have a trailing slash - do we need another test without a trailing slash?

bodhi/tests/server/services/test_updates.py
```diff
@@ -291,6 +291,51 @@ def test_new_rpm_update_koji_error(self, publish, *args):
...
+
+        self.assertIn(u'https://host.org/', resp)
+
+    @mock.patch(**mock_valid_requirements)
+    @mock.patch('bodhi.server.notifications.publish')
+    def test_koji_config_mock_url_with_trailing_slash(self, publish, *args):
+        """
+        Test html rendering of build link using a mock config variable 'koji_web_url' and
+        with a trailing slash in it
+        """
```
bowlofeggs: If we add a validator that ensures that the setting always has a trailing slash, we don't need a test here for this condition since it would not be able to happen. Instead, let's make a config validator and let's add a test that ensures that it works, and we can just trust it to do its job for the template rendering.

bowlofeggs (changes requested):

bowlofeggs:
There seem to be quite a few failing tests.

bodhi/tests/server/services/test_updates.py
```diff
@@ -291,6 +291,36 @@ def test_new_rpm_update_koji_error(self, publish, *args):
...
+        Test html rendering of build link using a mock config variable 'koji_web_url'
+        without a trailing slash in it
+        """
+        self.app.app.registry.settings['koji_web_url'] = u'https://host.org'
+        nvr = u'bodhi-2.0.0-2.fc17'
+        resp = self.app.post_json('/updates/', self.get_update(nvr))
+
+        resp = self.app.get('/updates/%s' % nvr, headers={'Accept': 'text/html'})
+
+        self.assertIn(u'https://host.org/', resp)
```
bowlofeggs: Let's make this assertion a little more opinionated so that it looks for the ```search?terms='+build.nvr+'&type=build&match=glob``` bits too.

bodhi/tests/server/services/test_updates.py
```diff
@@ -291,6 +291,36 @@ def test_new_rpm_update_koji_error(self, publish, *args):
...
+                                             '/search\?terms=.*\&amp;type=build\&amp;match=glob'))
+
+    @mock.patch(**mock_valid_requirements)
+    @mock.patch('bodhi.server.notifications.publish')
+    def test_koji_config_mock_url_without_trailing_slash(self, publish, *args):
+        """
+        Test html rendering of build link using a mock config variable 'koji_web_url'
+        without a trailing slash in it
+        """
+        self.app.app.registry.settings['koji_web_url'] = u'https://host.org'
```
bowlofeggs, line 331: Let's add one more test with the setting set like this:

```
https://koji.fedoraproject.org/koji
```

If the user types something like the above, make sure the right thing happens.

bowlofeggs (changes requested):

bowlofeggs (approved):

PR #2414, opened 2018-06-04
Do not run flake8 on Fedora 29.

### LINKED ISSUE #2412: The CI tests fail because flake8 is broken on Rawhide
There's an open BZ about this:

https://bugzilla.redhat.com/show_bug.cgi?id=1582075

It sounds like they are waiting for upstream to make a release and it's not clear how long that will take. We have a few options:

0) Send the flake8 spec file a PR to backport the cited patch from upstream so the Fedora package can work again.
1) Detect that we are testing Rawhide and disable the flake8 check (remembering to revisit it in a few weeks to see if it's been fixed - we can immediately open a PR to re-enable it and just wait until that PR can pass CI).

I'd prefer to pursue the first option no matter what, but we can also do the second option as a workaround since this is blocking CI from passing.

bowlofeggs:
fixes #2412

### DIFF
--- a/devel/test_container.sh
+++ b/devel/test_container.sh
@@ -45,7 +45,10 @@ mkdir -p /usr/local/lib/python$py3_version/site-packages/
 if ! rpm -q python2-flake8; then
     flake8 || fail
 else
-    flake8-2 || fail
+    # flake8 fails to run on Fedora 29 https://github.com/fedora-infra/bodhi/issues/2412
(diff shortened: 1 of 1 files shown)


bowlofeggs:
Once this is merged, I plan to immediately open another PR to revert it, and keep that PR open until flake8 is fixed.

pypingou (approved):

PR #2417, opened 2018-06-04
Added pagination flags for overrides and updates (CLI and Py bindings).

### LINKED ISSUE #2405: Python binding API list_overrides cannot return all overrides
`list_overrides` just returns overrides in the first page of `/overrides` by the default page size. Meanwhile, it does not accept argument, such as `page`, to return overrides from the rest pages.

ghost:
fixes #2405

### DIFF
--- a/bodhi/client/__init__.py
+++ b/bodhi/client/__init__.py
@@ -95,6 +95,15 @@ save_edit_options = [
     staging_option,
     url_option]
 
+# Basic options for pagination of query result
+pagination_options = [
+    click.option('--rows', default=None,
+                 type=click.IntRange(1, 100, clamp=False),
+                 help='Limits number of results shown per page'),
+    click.option('--page', default=None,
+                 type=click.IntRange(1, clamp=False),
+                 help='Go to page number')]
+
 
 def add_options(options):
     """
@@ -376,8 +385,9 @@ def edit(user, password, url, **kwargs):
 @click.option('--mine', is_flag=True, help='Show only your updates')
 @staging_option
 @url_option
+@add_options(pagination_options)
 @handle_errors
-def query(url, mine=False, **kwargs):
+def query(url, mine=False, rows=None, **kwargs):
     # User Docs that show in the --help
     """Query updates on Bodhi.
 
@@ -400,7 +410,7 @@ def query(url, mine=False, **kwargs):
     if mine:
         client.init_username()
         kwargs['user'] = client.username
-    resp = client.query(**kwargs)
+    resp = client.query(rows_per_page=rows, **kwargs)
     print_resp(resp, client)
 
 
@@ -606,9 +616,11 @@ def overrides():
 @click.option('--builds', default=None,
               help='Query by comma-separated build id(s)')
 @url_option
+@add_options(pagination_options)
 @handle_errors
 def query_buildroot_overrides(url, user=None, mine=False, packages=None,
-                              expired=None, releases=None, builds=None, **kwargs):
+                              expired=None, releases=None, builds=None,
+                              rows=None, page=None, **kwargs):
     # Docs that show in the --help
     """Query the buildroot overrides."""
     # Developer Docs
@@ -625,6 +637,8 @@ def query_buildroot_overrides(url, user=None, mine=False, packages=None,
         expired (bool): If supplied, True returns only expired overrides, False only active.
         releases (unicode): If supplied, the overrides for these releases are queried.
         builds (unicode): If supplied, the overrides for these builds are queried.
+        rows (unicode): The limit of rows displayed per page for query result.
+        page (unicode): If supplied, returns the results for a specific page number.
         kwargs (dict): Other keyword arguments passed to us by click.
     """
     client = bindings.BodhiClient(base_url=url, staging=kwargs['staging'])
@@ -632,7 +646,8 @@ def query_buildroot_overrides(url, user=None, mine=False, packages=None,
         client.init_username()
         user = client.username
     resp = client.list_overrides(user=user, packages=packages,
-                                 expired=expired, releases=releases, builds=builds)
+                                 expired=expired, releases=releases, builds=builds,
+                                 rows_per_page=rows, page=page)
     print_resp(resp, client)
 
 
--- a/bodhi/client/bindings.py
+++ b/bodhi/client/bindings.py
@@ -413,7 +413,8 @@ class BodhiClient(OpenIdBaseClient):
 
     @errorhandled
     def list_overrides(self, user=None, packages=None,
-                       expired=None, releases=None, builds=None):
+                       expired=None, releases=None, builds=None,
+                       rows_per_page=None, page=None):
         """
         List buildroot overrides.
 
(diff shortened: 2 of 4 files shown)


cverna:
jenkies test

bodhi/client/__init__.py
```diff
@@ -374,10 +374,15 @@ def edit(user, password, url, **kwargs):
               type=click.Choice(['newpackage', 'security', 'bugfix', 'enhancement']))
 @click.option('--user', help='Updates submitted by a specific user')
 @click.option('--mine', is_flag=True, help='Show only your updates')
+@click.option('--rows', default=None,
```
cverna: We could avoid duplicating these options ( in query and query_buildroot_overrides). You can check how the `@url_option` or `@new_edit_options` decorator works.

bodhi/client/__init__.py
```diff
@@ -374,10 +374,15 @@ def edit(user, password, url, **kwargs):
...
+              type=click.IntRange(1, 100, clamp=True),
```
cverna: Are we sure that we can't get over 100 pages ?

cverna (changes requested):

bodhi/client/__init__.py
ghost: I think that is the limit on number of records per page.

bodhi/client/__init__.py
```diff
@@ -376,8 +384,9 @@ def edit(user, password, url, **kwargs):
 @click.option('--mine', is_flag=True, help='Show only your updates')
 @staging_option
 @url_option
+@pagination_options
```
bowlofeggs: ```pagination_options``` is a list, and lists cannot be used as decorators:

```
[vagrant@bodhi-dev bodhi]$ bodhi
/usr/lib64/python3.6/importlib/_bootstrap_external.py:426: ImportWarning: Not importing directory /usr/lib/python3.6/site-packages/zope: missing __init__
  _warnings.warn(msg.format(portions[0]), ImportWarning)
/usr/lib64/python3.6/importlib/_bootstrap_external.py:426: ImportWarning: Not importing directory /usr/lib/python3.6/site-packages/repoze: missing __init__
  _warnings.warn(msg.format(portions[0]), ImportWarning)
/usr/lib64/python3.6/importlib/_bootstrap_external.py:426: ImportWarning: Not importing directory /usr/lib/python3.6/site-packages/paste: missing __init__
  _warnings.warn(msg.format(portions[0]), ImportWarning)
/usr/lib/python3.6/site-packages/koji/__init__.py:46: DeprecationWarning: the imp module is deprecated in favour of importlib; see the module's documentation for alternative uses
  import imp
/usr/lib/python3.6/site-packages/fedora/client/bodhi.py:48: DeprecationWarning: fedora.client.bodhi has been deprecated. Please use bodhi.client.bindings instead.
  DeprecationWarning)
  File "/usr/local/bin/bodhi", line 11, in <module>
  File "/usr/lib/python3.6/site-packages/pkg_resources/__init__.py", line 476, in load_entry_point
    return get_distribution(dist).load_entry_point(group, name)
  File "/usr/lib/python3.6/site-packages/pkg_resources/__init__.py", line 2700, in load_entry_point
    return ep.load()
  File "/usr/lib/python3.6/site-packages/pkg_resources/__init__.py", line 2318, in load
    return self.resolve()
  File "/usr/lib/python3.6/site-packages/pkg_resources/__init__.py", line 2324, in resolve
    module = __import__(self.module_name, fromlist=['__name__'], level=0)
  File "/home/vagrant/bodhi/bodhi/client/__init__.py", line 389, in <module>
    def query(url, mine=False, rows=None, **kwargs):
TypeError: 'list' object is not callable
```

There is an ```add_options``` function that is designed to accept a list of ```click``` options, which can be used like this:

```
@add_options(pagination_options)
```

bodhi/client/bindings.py
```diff
@@ -425,6 +426,10 @@ def list_overrides(self, user=None, packages=None,
             releases (basestring): Comma separated Release shortnames to filter buildroot overrides
                 by.
             builds (basestring): Comma separated build NVRs to filter overrides by.
+            rows_per_page (int): Limit the results to a certain number of rows per page.
+                (default:20)
+            page (int): Return a specific page of results.
+                (default:1)
```
bowlofeggs: I would not document these defaults here, because they are API defaults and not bindings defaults. This means that if the API defaults ever change it is unlikely that we will remember to change this here and it'll then be incorrect. Instead, I would just document that the value None is the default, which will cause it to use the API default.

bowlofeggs (changes requested):

bodhi/client/__init__.py
```diff
@@ -95,6 +95,14 @@ def _warn_if_url_and_staging_set(ctx, param, value):
     staging_option,
     url_option]
 
+# Basic options for pagination of query result
+pagination_options = [
+    click.option('--rows', default=None,
+                 type=click.IntRange(1, 100, clamp=True),
```
bowlofeggs: I think it could be confusing if we silently clamp larger values to ```100``` here. Users may try to do ```--rows 200``` and may not notice they are getting 100 back, and then may page to page 2 skipping 100 items. Let's use ```clamp=False``` so that users are aware when they exceed the page size.

bodhi/client/__init__.py
```diff
@@ -95,6 +95,14 @@ def _warn_if_url_and_staging_set(ctx, param, value):
...
+                 help='Limits number of results shown per page'),
+    click.option('--page', default=None,
+                 help='Go to page number')]
```
bowlofeggs: We should also enforce this one to be an ```IntRange``` with a min of ```1``` and no max.

bodhi/client/__init__.py
```diff
@@ -625,14 +636,19 @@ def query_buildroot_overrides(url, user=None, mine=False, packages=None,
         expired (bool): If supplied, True returns only expired overrides, False only active.
         releases (unicode): If supplied, the overrides for these releases are queried.
         builds (unicode): If supplied, the overrides for these builds are queried.
+        rows (unicode): The limit of rows displayed per page for query result.
+                        Default: 20, Min: 1, Max: 100
+        page (unicode): If supplied, returns the results for a specific page number.
+                        Default: 1
```
bowlofeggs: Let's not document these defaults here - they are API defaults and are not particular to the CLI.

bodhi/tests/client/test_bindings.py
```diff
@@ -535,6 +535,30 @@ def test_with_builds(self):
...
                                                     params={'builds': 'python-1.5.6-3.fc26'})
 
+    def test_with_rows_per_page(self):
+        """
+        Test with the rows_per_page parameter.
+        """
+        client = bindings.BodhiClient()
+        client.send_request = mock.MagicMock(return_value='response')
+
+        response = client.list_overrides(rows_per_page=10)
```
bowlofeggs, line 545: Let's add a space under this line, since this is the line we are testing.

bodhi/tests/client/test_bindings.py
```diff
@@ -535,6 +535,30 @@ def test_with_builds(self):
...
+                                                    params={'rows_per_page': 10})
+
+    def test_with_page(self):
+        """
+        Test with the page parameter.
+        """
+        client = bindings.BodhiClient()
+        client.send_request = mock.MagicMock(return_value='response')
+
+        response = client.list_overrides(page=5)
```
bowlofeggs, line 558: Let's add a space under this line, since this is the line we are testing.

bowlofeggs (changes requested):

bowlofeggs (changes requested):
We need to rebase to resolve a conflict, but LGTM otherwise.

bowlofeggs (approved):
Looks good, and most importantly it seems to work. Nice!

bowlofeggs:
I assume that @cverna is also good with this, since we did a video review together on Monday, so I will merge.

PR #2418, opened 2018-06-04
test_bindings: Initialize BodhiClient with a username

owtaylor:
Previously the username (which the test required to be bowlofeggs)
was entered by mocking the user inputing it in the case where no
cached username is found. This fails in a non-isolated environment
if there is a cached Fedora username, so just pass it to the
constructor.


bowlofeggs (approved):

bowlofeggs:
jenkies test

bowlofeggs:

Issue #2350, opened 2018-05-03
Subclass TestApp to automaticaly add header `accept` to get requests


frenzymadness:
As you may spot in my previous pull request, a lot of tests can be fixed (in Python 3) just by specifying `accept` header to explicitly request JSON response.

Because now I know that this is the issue for a huge amount of tests, I have an idea. What about subclass `webtest.TestApp` and modify `get()` method to add this header automatically?

Actually, I tried it and it works well and looks simple.

```python
class BodhiTestApp(TestApp):
    """
    Subclass of webtest.TestApp which autimaticaly adds {'Accept': 'application/json'}
    header to each GET request where this header is omitted.
    """
    def get(self, *args, **kwargs):
        if 'headers' not in kwargs:
            kwargs['headers'] = {'Accept': 'application/json'}
        else:
            headers_set = set([key.lower() for key in kwargs['headers']])
            if 'accept' not in headers_set:
                kwargs['headers']['accept'] = 'application/json'
        return super(BodhiTestApp, self).get(*args, **kwargs)
```

This subclass overwrites `get()` method just for adding `accept` header to headers if no other `accept` is specified and calls the original method at the end. It also means that it would be necessary to specify `accept` header in all tests where it's omitted now and where a response has to be something else than JSON.

As the result of the mentioned subclass and few minor changes, I have a patch which will unblock more than 240 tests in Python 3 (and I didn't check all files).

I am not submitting PR because I am working on top of the PR #2286 which also unblocks some tests.

Currently I have only 84 tests skipped in Python 3.

What do you thing about it?

bowlofeggs:
Hi @frenzymadness!

That plan sounds very reasonable to me. +1

bowlofeggs:
Whoops, mean to hit comment not close.

frenzymadness:
Great! I went further and I fixed almost everything and now I have only 19 skipped tests in Python 3 which is more than 300 unblocked tests.

I'll wait until the merge of PR #2286  and then I'll create a new PR.

bowlofeggs:
On 05/04/2018 03:37 AM, frenzymadness wrote:
> Great! I went further and I fixed almost everything and now I have only
> 19 skipped tests in Python 3 which is more than 300 unblocked tests.

Wow, that's fantastic!

bowlofeggs:
Hopefully I'll be able to push that cornice update to stable on Monday so we can unblock #2286 next week.

frenzymadness:
[...]

PR #2421, opened 2018-06-05
Python 3 ready tests

frenzymadness:
Hello.

As discussed in https://github.com/fedora-infra/bodhi/issues/2350 I've prepared a change which allows me to unblock approx 300 tests in Python 3.

With that applied, there are only 20 tests skipped in Python 3! We can track the work on the rest of failing tests in the https://github.com/fedora-infra/bodhi/issues/2059. Currently, I am unsure when I'll have free cycles to take a look but I think it's still a huge progress.

What do you think?

### DIFF
--- a/bodhi/server/views/metrics.py
+++ b/bodhi/server/views/metrics.py
@@ -39,8 +39,7 @@ def compute_ticks_and_data(db, releases, update_types):
     """
     data, ticks = [], []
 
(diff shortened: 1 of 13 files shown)


bowlofeggs (approved):
Beautiful!

bowlofeggs:

PR #2422, opened 2018-06-05
With the recent switch to Fedora 28, we can use python2- packages.

bowlofeggs:


bowlofeggs:
Note that this PR will be slightly disruptive to developers who use Vagrant. Inspired by #2262, I've decided to rename ```devel/Vagrantfile.example``` to ```Vagrantfile```, which will conflict with developers who have this file. They can simply remove the file if they haven't made changes, or rename it if they have (and set up a suitable environment variable).

pypingou (approved):
I think I'll keep the Vagrantfile in the .gitignore on my side so I can tweak it as I want w/o having to pay attention to not commit it to the tree.

PR #2424, opened 2018-06-06
Merged the separate CLI used to manage releases under bodhi CLI

### LINKED ISSUE #2419: Fold the functionality of bodhi-manage-releases into the Bodhi CLI
It is silly that Bodhi has a separate CLI to manage releases. This duplicates boilerplate code and also requires users to install Bodhi's server on their machine in order to perform administrative actions with its CLI. We should move all functionality from ```bodhi-manage-releases``` into the ```bodhi``` CLI under a new ```releases``` command.

In doing this, we should mark ```bodhi-manage-releases``` as deprecated in the release notes, in its man page, in its ```--help``` text, and have it print a deprecation warning when it is used.

ghost:
fixes #2419

### DIFF
--- a/bodhi/client/__init__.py
+++ b/bodhi/client/__init__.py
@@ -28,6 +28,7 @@ import functools
 
 import click
 import six
+import munch
 
 from bodhi.client import bindings
 from fedora.client import AuthError
@@ -95,6 +96,7 @@ save_edit_options = [
     staging_option,
     url_option]
 
+
 # Basic options for pagination of query result
 pagination_options = [
     click.option('--rows', default=None,
@@ -103,6 +105,33 @@ pagination_options = [
     click.option('--page', default=None,
                  type=click.IntRange(1, clamp=False),
                  help='Go to page number')]
+
+
+# Common releases options
+release_options = [
+    click.option('--username'),
+    click.option('--password', prompt=True, hide_input=True),
+    click.option('--name', help='Release name (eg: F20)'),
+    click.option('--long-name', help='Long release name (eg: "Fedora 20")'),
+    click.option('--id-prefix', help='Release prefix (eg: FEDORA)'),
+    click.option('--version', help='Release version number (eg: 20)'),
+    click.option('--branch', help='Git branch name (eg: f20)'),
+    click.option('--dist-tag', help='Koji dist tag (eg: f20)'),
+    click.option('--stable-tag', help='Koji stable tag (eg: f20-updates)'),
+    click.option('--testing-tag',
+                 help='Koji testing tag (eg: f20-updates-testing)'),
+    click.option('--candidate-tag',
+                 help='Koji candidate tag (eg: f20-updates-candidate)'),
+    click.option('--pending-stable-tag',
+                 help='Koji pending tag (eg: f20-updates-pending)'),
+    click.option('--pending-testing-tag',
+                 help='Koji pending testing tag (eg: f20-updates-testing-testing)'),
+    click.option('--override-tag', help='Koji override tag (eg: f20-override)'),
+    click.option('--state', type=click.Choice(['disabled', 'pending', 'current',
+                                               'archived']),
+                 help='The state of the release'),
+    staging_option,
+    url_option]
 
 
 def add_options(options):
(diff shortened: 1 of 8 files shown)


bowlofeggs:
This one is failing several tests too - there are tests that assert the output of bodhi-manage-releases, and they are not expecting the new deprecation warnings. We will need to adjust them to pass with the new output.

bodhi/client/__init__.py
```diff
@@ -25,6 +25,7 @@
 import traceback
 import re
 import functools
+import munch
```
bowlofeggs: So this is a third party library, and PEP-8 says third party libraries should be grouped together, so let's move this to be with click and six below.

bodhi/server/scripts/manage_releases.py
```diff
@@ -37,7 +37,11 @@
 
 @click.group()
 def main():
-    """Create and manage releases in Bodhi."""
+    """
+    Create and manage releases in Bodhi.
+
+    (This utility has been deprecated. Please use 'bodhi releases' instead.)
+    """
     pass  # pragma: no cover
```
bowlofeggs: Let's do the deprecation warning here instead of in each command:

```
click.echo("This utility has been deprecated. Please use 'bodhi releases' instead.)
```

bodhi/server/scripts/manage_releases.py
```diff
@@ -37,7 +37,11 @@
```
bowlofeggs, line 34: Let's find a way to share code between this old module and the new one. Ideally we leave almost all the code in the new one above and have this module only print the deprecation warning and otherwise use the code from the other module.

bowlofeggs (changes requested):
There is a merge conflict to resolve.

bodhi/server/scripts/manage_releases.py
```diff
@@ -27,17 +27,29 @@
...
-    """Create and manage releases in Bodhi."""
+@click.pass_context
+def main(ctx):
+    """
+    Create and manage releases in Bodhi.
+
+    (This utility has been deprecated. Please use 'bodhi releases' instead.)
+    """
+    click.echo("This utility has been deprecated. Please use 'bodhi releases' instead.")
+    ctx.forward(cli)
```
bowlofeggs: It looks like the tests fail due to missing coverage on these two lines.

bodhi/server/scripts/manage_releases.py
```diff
@@ -27,17 +27,29 @@
...
 
 from bodhi.client import bindings
+from bodhi.client import (
+    cli,
+    add_options,
+    release_options,
+    create_release,
+    edit_release,
+    info_release
+)
```
bowlofeggs: There may be a simpler way to do this that won't involve importing and wrapping all of these commands. It looks like you can access the ```bodhi releases``` ```Command``` like this:

```
cli.commands['releases']
```

I think something like this might work nicely, and require less code in here:

```
def main():
    """Print a deprecation warning and call the bodhi releases command."""
    click.echo("This utility has been deprecated. Please use 'bodhi releases' instead.")
    cli.commands['releases']()
```

bowlofeggs (changes requested):
The failing tests are due to missing coverage.

bowlofeggs:
There is also a merge conflict to resolve here.

bodhi/server/scripts/manage_releases.py
ghost: Is there a way to test the main() function here? I believe that invoking it using CliRunner() will display its '--help' text and will not execute the code after it. (Reason maybe because it is a @click.group and not @click.command)

bodhi/client/__init__.py
```diff
@@ -105,6 +107,31 @@ def _warn_if_url_and_staging_set(ctx, param, value):
...
+    click.option('--candidate-tag',
+                 help='Koji candidate tag (eg: f20-updates-candidate)'),
+    click.option('--pending-stable-tag',
+                 help='Koji pending tag (eg: f20-updates-pending)'),
+    click.option('--pending-testing-tag',
+                 help='Koji pending testing tag (eg: f20-updates-testing-testing)'),
+    click.option('--override-tag', help='Koji override tag (eg: f20-override)'),
+    click.option('--state', type=click.Choice(['disabled', 'pending', 'current',
+                                               'archived']),
+                 help='The state of the release')]
```
cverna: you could add the `url_option` here instead of using the decorator on each release command. 

See as an example  https://github.com/fedora-infra/bodhi/blob/develop/bodhi/client/__init__.py#L95

cverna: I think it would not hurt to add the `staging_option` too

bodhi/client/__init__.py
```diff
@@ -787,5 +814,137 @@ def print_resp(resp, client, verbose=False):
...
+    """Interact with releases."""
+    # Developer Docs
+    """Manage the releases."""
+    pass  # pragma: no cover
+
+
+@releases.command(name='create')
+@handle_errors
+@add_options(release_options)
+@url_option
```
cverna: That would not be needed if you add it to the `release_options` list

bodhi/client/__init__.py
```diff
@@ -787,5 +814,137 @@ def print_resp(resp, client, verbose=False):
...
+def create_release(username, password, url, **kwargs):
+    """Create a release."""
+    client = bindings.BodhiClient(base_url=url, username=username, password=password)
```
cverna: If you add the `staging_option` that would need to be given as an argument here. you can check https://github.com/fedora-infra/bodhi/blob/develop/bodhi/client/__init__.py#L269

bodhi/client/__init__.py
```diff
@@ -787,5 +814,137 @@ def print_resp(resp, client, verbose=False):
...
+    kwargs['csrf_token'] = client.csrf()
+
+    save(client, **kwargs)
+
+
+@releases.command(name='edit')
+@handle_errors
+@add_options(release_options)
+@click.option('--new-name', help='New release name (eg: F20)')
+@url_option
```
cverna: Same as `create` command

bodhi/client/__init__.py
```diff
@@ -787,5 +814,137 @@ def print_resp(resp, client, verbose=False):
...
+        if v is not None:
+            data[k] = v
+
+    save(client, **data)
+
+
+@releases.command(name='info')
+@handle_errors
+@click.argument('name')
+@url_option
```
cverna, line 881: Maybe we can add the `staging_option` here too and add it to the BodhiClient args.

bodhi/server/scripts/manage_releases.py
```diff
@@ -27,183 +27,19 @@
...
-
-    Args:
-        errors (munch.Munch): The errors to be formatted and printed.
-    """
-    for error in data['errors']:
-        print("ERROR: %s" % error['description'])
-
-    sys.exit(1)
+    click.echo("This utility has been deprecated. Please use 'bodhi releases' instead.")
+    cli.commands['releases']()
```
cverna, line 42: Nice :+1:

ghost:
jenkies test

bodhi/client/__init__.py
cverna: @vismay-golwala let's update this then we can merge this PR :monkey_face:

cverna (approved):

PR #2426, opened 2018-06-07
Measure branch coverage.

bowlofeggs:
Though Bodhi currently has 99% line test coverage, there are some
branch paths that are not currently covered, so this commit also
reduces the required coverage percentage to 98%.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,4 +1,5 @@
 [run]
+branch = True
 source = bodhi
 omit =
     bodhi/scripts/*
@@ -8,7 +9,7 @@ omit =
 
 [report]
 precision = 2
-fail_under = 100
+fail_under = 97
 exclude_lines =
     pragma: no cover
     class DevBuildsys


cverna (dismissed):

pypingou:
The jenkins error is... interesting:
````f28	FAIL Required test coverage of 98% not reached. Total coverage: 98.00%````

bowlofeggs:
Yeah that is a strange error. It seemed to pass locally when I created this patch. I'll have a look later to see if I can figure out what's going on.

bowlofeggs:
I wonder if it's a matter of significant digits or something. When I run the tests locally, I see a slightly different output about it having 98.01% instead of 98.00%, and it passes. I wonder if I just set it to 97% if that'll get it working…

bowlofeggs:
Note that the goal is that I'll slowly work to raise it to 100%, but this PR is just about switching the measurement of branch coverage on, so IMO if it passes with it set to 97% so be it.

bowlofeggs:
Yeah it's happy with 97%, so I bet it's like 97.999% or something and so setting it to 98 fails.

I don't intend to merge this PR until after I submit one more to get 100% line coverage. I probably won't have time to do that for a month or so, so this PR might just sit for a while.

bowlofeggs:
Once https://github.com/fedora-infra/bodhi/pull/2493 is merged I will rebase this one and get it passing.

bowlofeggs:
This PR passes CI now and is ready for review.

bowlofeggs:
I believe @puiterwijk put the no-mergify label on this due to my comment about how I didn't intend to merge it until my other PR is merged. Since that PR is now merged, I will remove the no-mergify label.

pypingou:
Still :+1: for me

puiterwijk (dismissed):

puiterwijk (approved):

PR #2427, opened 2018-06-07
Configure coverage to print the missing lines in the text report.

bowlofeggs:

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -15,3 +15,4 @@ exclude_lines =
     def __repr__
     def multicall_enabled
     if __name__ == .__main__.:
+show_missing = True


pypingou (approved):

pypingou:
For such PRs, @bowlofeggs would you be ok with me merging them directly?

bowlofeggs:
On 06/08/2018 04:04 AM, Pierre-Yves Chibon wrote:
> For such PRs, @bowlofeggs <https://github.com/bowlofeggs> would you be
> ok with me merging them directly?

I usually rebase all my PRs to keep the git history clean from merge
commits, and I also gpg sign my commits, so if someone else does a
rebase the signature will be removed. So I'd prefer to just merge them
myself, but thanks for the offer!

<acknowledgement that signing commits doesn't do much>

PR #2429, opened 2018-06-08
Add unit-tests checking the behavior of the waive result endpoint and allow specifying which tests to waive

pypingou:
Somehow we only had test checking the failures on the endpoint... This is
now testing the success of the endpoint, when querying greenwave and
waiverdb worked.

### DIFF
--- a/bodhi/server/models.py
+++ b/bodhi/server/models.py
@@ -47,8 +47,7 @@ from bodhi.server.config import config
 from bodhi.server.exceptions import BodhiException, LockedUpdateException
 from bodhi.server.util import (
     avatar as get_avatar, build_evr, flash_log, get_critpath_components,
-    get_rpm_header, header, tokenize, pagure_api_get,
-    waiverdb_api_post)
+    get_rpm_header, header, tokenize, pagure_api_get)
 import bodhi.server.util
 
 if six.PY2:
@@ -1984,7 +1983,7 @@ class Update(Base):
         }
         api_url = '{}/decision'.format(config.get('greenwave_api_url'))
 
-        return util.greenwave_api_post(api_url, data)
+        return bodhi.server.util.greenwave_api_post(api_url, data)
 
     def update_test_gating_status(self):
         """Query Greenwave about this update and set the test_gating_status as appropriate."""
@@ -2637,13 +2636,16 @@ class Update(Base):
         topic = u'update.request.%s' % action
         notifications.publish(topic=topic, msg=dict(update=self, agent=username))
 
-    def waive_test_results(self, username, comment=None):
+    def waive_test_results(self, username, comment=None, tests=None):
         """
         Attempt to waive test results for this update.
 
         Args:
             username (basestring): The name of the user who is waiving the test results.
             comment (basestring): A comment from the user describing their decision.
+            tests (list of basestring): A list of testcases to be waived. Defaults to ``None``
+                If left as ``None``, all ``unsatisfied_requirements`` returned by greenwave
+                will be waived, otherwise only the testcase found in both list will be waived.
         Raises:
             LockedUpdateException: If the Update is locked.
             BodhiException: If test gating is not enabled in this Bodhi instance,
@@ -2661,8 +2663,15 @@ class Update(Base):
         if self.test_gating_passed:
             raise BodhiException("Can't waive test resuts on an update that passes test gating")
 
+        # Ensure we can always iterate over tests
+        tests = tests or []
+
         decision = self.get_test_gating_info()
         for requirement in decision['unsatisfied_requirements']:
+
+            if tests and requirement['testcase'] not in tests:
+                continue
+
             data = {
                 'subject': requirement['item'],
                 'testcase': requirement['testcase'],
@@ -2672,7 +2681,8 @@ class Update(Base):
                 'comment': comment
             }
             log.debug('Waiving test results: %s' % data)
-            waiverdb_api_post('{}/waivers/'.format(config.get('waiverdb_api_url')), data)
+            bodhi.server.util.waiverdb_api_post(
+                '{}/waivers/'.format(config.get('waiverdb_api_url')), data)
 
     def add_tag(self, tag):
         """
--- a/bodhi/server/schemas.py
+++ b/bodhi/server/schemas.py
@@ -108,6 +108,12 @@ class Updates(colander.SequenceSchema):
     update = colander.SchemaNode(colander.String())
 
 
+class Tests(colander.SequenceSchema):
+    """A SequenceSchema to validate a list of Test objects."""
+
+    test = colander.SchemaNode(colander.String())
+
+
 class BugFeedback(colander.MappingSchema):
     """A schema for BugFeedback to be provided via API parameters."""
 
@@ -807,6 +813,7 @@ class WaiveTestResultsSchema(CSRFProtectedSchema, colander.MappingSchema):
         colander.String(),
         missing=None,
     )
+    tests = Tests(colander.Sequence(accept_scalar=True), missing=None, preparer=[util.splitter])
 
 
 class GetTestResultsSchema(CSRFProtectedSchema, colander.MappingSchema):
--- a/bodhi/tests/server/services/test_updates.py
+++ b/bodhi/tests/server/services/test_updates.py
@@ -5077,6 +5077,151 @@ class TestWaiveTestResults(BaseTestCase):
                           u'IOError. oops!')
         log_exception.assert_called_once_with("Unhandled exception in waive_test_results")
 
+    @mock.patch.dict(config, [('test_gating.required', True)])
+    @mock.patch('bodhi.server.models.waiverdb_api_post')
+    @mock.patch('bodhi.server.models.greenwave_api_post')
+    @mock.patch('bodhi.server.models.User.openid', mock.MagicMock(return_value=None))
+    @mock.patch('bodhi.server.models.User.avatar', mock.MagicMock(return_value=None))
+    def test_waive_test_results_1_unsatisfied_requirement(
+            self, greenwave_api_post, waiverdb_api_post, *args):
+        """Ensure that waiverdb and greenwaved are properly called when greenwave returns only one
+        unsatisfied requirements."""
+        nvr = u'bodhi-2.0-1.fc17'
+        greenwave_api_post.return_value = {
+            u'unsatisfied_requirements': [
+                {
+                    u'item': {
+                        u'item': u'bodhi-2.0-1.fc17',
+                        u'type': u'koji_build'
+                    },
+                    u'scenario': None,
+                    u'testcase': u'dist.rpmdeplint',
+                    u'type': u'test-result-missing'
+                }
+            ],
+        }
+
+        up = self.db.query(Update).filter_by(title=nvr).one()
+        up.test_gating_status = TestGatingStatus.failed
+
+        post_data = dict(update=nvr, comment="This is expected", csrf_token=self.get_csrf_token())
+        res = self.app.post_json('/updates/%s/waive-test-results' % str(nvr), post_data, status=200)
+
+        greenwave_api_post.assert_called_once_with(
+            'https://greenwave-web-greenwave.app.os.fedoraproject.org/api/v1.0/decision',
+            {
+                'product_version': u'fedora-17',
+                'decision_context': u'bodhi_update_push_testing',
+                'subject': [
+                    {'item': u'bodhi-2.0-1.fc17', 'type': 'koji_build'},
+                    {'original_spec_nvr': u'bodhi-2.0-1.fc17'},
+                    {'item': u'FEDORA-2018-a3bbe1a8f2', 'type': 'bodhi_update'}
+                ]
+            }
+        )
+
+        waiverdb_api_post.assert_called_once_with(
+            'https://waiverdb-web-waiverdb.app.os.fedoraproject.org/api/v1.0/waivers/',
+            {
+                'username': u'guest',
+                'comment': u'This is expected',
+                'waived': True,
+                'product_version': u'fedora-17',
+                'testcase': u'dist.rpmdeplint',
+                'subject': {
+                    u'item': u'bodhi-2.0-1.fc17', u'type': u'koji_build'
+                }
+            }
+        )
+
+        self.assertEqual(list(res.json_body.keys()), ['update'])
+        self.assertEqual(res.json_body['update'], up.__json__())
+
+    @mock.patch.dict(config, [('test_gating.required', True)])
+    @mock.patch('bodhi.server.models.waiverdb_api_post')
+    @mock.patch('bodhi.server.models.greenwave_api_post')
+    @mock.patch('bodhi.server.models.User.openid', mock.MagicMock(return_value=None))
+    @mock.patch('bodhi.server.models.User.avatar', mock.MagicMock(return_value=None))
+    def test_waive_test_results_2_unsatisfied_requirements(
+            self, greenwave_api_post, waiverdb_api_post, *args):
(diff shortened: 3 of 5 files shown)


pypingou:
Rebased on the top of develop

pypingou:
Rebased

pypingou:
@bowlofeggs anything to change on this PR?

bodhi/tests/server/services/test_updates.py
```diff
@@ -5028,3 +5028,169 @@ def test_unexpected_exception(self, log_exception, waive_test_results, *args):
         self.assertEquals(res.json_body['errors'][0]['description'],
                           u'IOError. oops!')
         log_exception.assert_called_once_with("Unhandled exception in waive_test_results")
+
+    @mock.patch.dict(config, [('test_gating.required', True)])
+    @mock.patch('bodhi.server.models.waiverdb_api_post')
+    @mock.patch('bodhi.server.models.greenwave_api_post')
+    def test_waive_test_results_1_unsatisfied_requirement(
+            self, greenwave_api_post, waiverdb_api_post, *args):
+        """Ensure that we get an error if trying to waive test results of a locked update"""
```
bowlofeggs: The comment says the update is locked, but the name and body seem to be about there being one unsatisfied requirement.

bodhi/tests/server/services/test_updates.py
```diff
@@ -5028,3 +5028,169 @@ def test_unexpected_exception(self, log_exception, waive_test_results, *args):
...
+                u'content_type', u'critpath', u'date_approved', u'date_modified', u'date_pushed',
+                u'date_stable', u'date_submitted', u'date_testing', u'display_name',
+                u'greenwave_summary_string', u'greenwave_unsatisfied_requirements', u'karma',
+                u'locked', u'meets_testing_requirements', u'notes', u'old_updateid', u'pushed',
+                u'release', u'request', u'require_bugs', u'require_testcases', u'requirements',
+                u'severity', u'stable_karma', u'status', u'submitter', u'suggest', u'test_cases',
+                u'test_gating_status', u'title', u'type', u'unstable_karma', u'updateid', u'url',
+                u'user'
+            ]
+        )
```
bowlofeggs: It might be easier and more thorough to do this as an assertion in place of the above two:

```
self.assertEqual(res.json_body, up.__json__())
```

bodhi/tests/server/services/test_updates.py
```diff
@@ -5028,3 +5028,169 @@ def test_unexpected_exception(self, log_exception, waive_test_results, *args):
...
+                'product_version': u'fedora-17',
+                'testcase': u'dist.rpmdeplint',
+                'subject': {
+                    u'item': u'bodhi-2.0-1.fc17', u'type': u'koji_build'
+                }
+            }
+        )
+
+        self.assertEquals(list(res.json_body.keys()), ['update'])
+        self.assertEquals(
```
bowlofeggs: Note here and elsewhere that ```assertEquals``` is deprecated:

```
DeprecationWarning: Please use assertEqual instead.
```

bodhi/tests/server/services/test_updates.py
```diff
@@ -5028,3 +5028,169 @@ def test_unexpected_exception(self, log_exception, waive_test_results, *args):
...
+                    'waived': True,
+                    'product_version': u'fedora-17',
+                    'testcase': u'atomic_ci_pipeline_results',
+                    'subject': {
+                        u'item': u'bodhi-2.0-1.fc17', u'type': u'koji_build'
+                    }
+                }
+            )
+        ]
+        waiverdb_api_post.assert_has_calls(calls)
```
bowlofeggs: You can make a stronger assertion here, and then you don't need to assert the len() is 2 above:

```
self.assertEqual(waiverdb_api_post.mock_calls, calls)
```

bodhi/tests/server/services/test_updates.py
```diff
@@ -5028,3 +5028,169 @@ def test_unexpected_exception(self, log_exception, waive_test_results, *args):
...
+                u'content_type', u'critpath', u'date_approved', u'date_modified', u'date_pushed',
+                u'date_stable', u'date_submitted', u'date_testing', u'display_name',
+                u'greenwave_summary_string', u'greenwave_unsatisfied_requirements', u'karma',
+                u'locked', u'meets_testing_requirements', u'notes', u'old_updateid', u'pushed',
+                u'release', u'request', u'require_bugs', u'require_testcases', u'requirements',
+                u'severity', u'stable_karma', u'status', u'submitter', u'suggest', u'test_cases',
+                u'test_gating_status', u'title', u'type', u'unstable_karma', u'updateid', u'url',
+                u'user'
+            ]
+        )
```
bowlofeggs: Similarly here, you could do the ```up.__json__()``` comparison like I mentioned above.

bodhi/server/models.py
```diff
@@ -2617,13 +2617,16 @@ def set_request(self, db, action, username):
         topic = u'update.request.%s' % action
         notifications.publish(topic=topic, msg=dict(update=self, agent=username))
 
-    def waive_test_results(self, username, comment=None):
+    def waive_test_results(self, username, comment=None, waive_tests=None):
```
bowlofeggs: You could just name the parameter ```tests```, which would read more naturally since the method name already implies that we are waiving tests.

bodhi/server/models.py
```diff
@@ -2617,13 +2617,16 @@ def set_request(self, db, action, username):
...
         """
         Attempt to waive test results for this update.
 
         Args:
             username (basestring): The name of the user who is waiving the test results.
             comment (basestring): A comment from the user describing their decision.
+            waive_tests (list of basestring): A list of testcases to be waived. Defaults to ``None``
```
bowlofeggs: s/$/./

bodhi/tests/server/services/test_updates.py
```diff
@@ -5194,3 +5194,87 @@ def test_waive_test_results_2_unsatisfied_requirements(
...
```
bowlofeggs: Similar suggestion here about this assertion.

bodhi/tests/server/services/test_updates.py
```diff
@@ -5194,3 +5194,87 @@ def test_waive_test_results_2_unsatisfied_requirements(
...
+                'username': u'guest',
+                'comment': None,
+                'waived': True,
+                'product_version': u'fedora-17',
+                'testcase': u'atomic_ci_pipeline_results',
+                'subject': {
+                    u'item': u'bodhi-2.0-1.fc17', u'type': u'koji_build'
+                }
+            }
+        )
```
bowlofeggs, line 5293: I also suggest the stronger assertion here, so you don't need the first one.

bodhi/tests/server/services/test_updates.py
```diff
@@ -5194,3 +5194,87 @@ def test_waive_test_results_2_unsatisfied_requirements(
                 u'user'
             ]
         )
+
+    @mock.patch.dict(config, [('test_gating.required', True)])
+    @mock.patch('bodhi.server.models.waiverdb_api_post')
+    @mock.patch('bodhi.server.models.greenwave_api_post')
+    def test_waive_test_results_1_of_2_unsatisfied_requirements(
```
bowlofeggs: I suggest one more test that tries to waive 2 test results, to make sure the schema can handle lists.

bodhi/server/schemas.py
```diff
@@ -807,3 +807,8 @@ class WaiveTestResultsSchema(CSRFProtectedSchema, colander.MappingSchema):
         colander.String(),
         missing=None,
     )
+
+    waive_tests = colander.SchemaNode(
+        colander.String(),
+        missing=None,
+    )
```
bowlofeggs: I recommend just naming this ```tests``` as well, since the API already implies that waiving is happening.

Also, I don't think the tests will come in as a list of strings with this schema, but just as a string.

bowlofeggs (changes requested):

bodhi/tests/server/services/test_updates.py
```diff
@@ -5028,3 +5028,253 @@ def test_unexpected_exception(self, log_exception, waive_test_results, *args):
...
+                u'user'
+            ]
+        )
+
+    @mock.patch.dict(config, [('test_gating.required', True)])
+    @mock.patch('bodhi.server.models.waiverdb_api_post')
+    @mock.patch('bodhi.server.models.greenwave_api_post')
+    def test_waive_test_results_2_unsatisfied_requirements(
+            self, greenwave_api_post, waiverdb_api_post, *args):
+        """Ensure that we get an error if trying to waive test results of a locked update"""
```
bowlofeggs: This one also has a misleading comment.

bodhi/tests/server/services/test_updates.py
```diff
@@ -5028,3 +5028,253 @@ def test_unexpected_exception(self, log_exception, waive_test_results, *args):
...
+                u'user'
+            ]
+        )
+
+    @mock.patch.dict(config, [('test_gating.required', True)])
+    @mock.patch('bodhi.server.models.waiverdb_api_post')
+    @mock.patch('bodhi.server.models.greenwave_api_post')
+    def test_waive_test_results_1_of_2_unsatisfied_requirements(
+            self, greenwave_api_post, waiverdb_api_post, *args):
+        """Ensure that we get an error if trying to waive test results of a locked update"""
```
bowlofeggs: And this comment.

bodhi/server/models.py

bodhi/server/schemas.py
pypingou: Name adjusted.

> Also, I don't think the tests will come in as a list of strings with this schema, but just as a string.

I was a little afraid of this, but I didn't find how to ask for a list, any clues for me?

bodhi/tests/server/services/test_updates.py
pypingou: I'll adjust all the docstrings

bodhi/tests/server/services/test_updates.py
pypingou: +1

bodhi/tests/server/services/test_updates.py
pypingou: Will adjust

bodhi/tests/server/services/test_updates.py
pypingou: The first one is not needed at all actually, but I think this assertion here is strong enough since it's ``assert_called_once``

bodhi/tests/server/services/test_updates.py
pypingou: So this unfortunately doesn't work as it seems one of them include openid and avatar information while the other doesn't :(
````
E       -                 u'user': {u'avatar': u'https://seccdn.libravatar.org/avatar/eb48e08cc23bcd5961de9541ba5156c385cd39799e1dbf511477aa4d4d3a37e7?s=24&d=retro',
E       +                'user': {'avatar': None,
````

````
E       -                           u'openid': u'guest.id.fedoraproject.org',
E       +                         'openid': None,
````

I'll look if I can fix this but I'm not seeing anywhere else where you used this assertion before so seems to be a new territory

Comment (pypingou) [reply bodhi/tests/server/services/test_updates.py]: Some mock magic fixed it :)

````
@mock.patch('bodhi.server.models.User.openid', mock.MagicMock(return_value=None))
@mock.patch('bodhi.server.models.User.avatar', mock.MagicMock(return_value=None))
````

**Comment by pypingou** [comment]
I think I addressed all your comments :)

File: bodhi/tests/server/services/test_updates.py [hunk 14]
```diff
@@ -5074,3 +5074,148 @@ def test_unexpected_exception(self, log_exception, waive_test_results, *args):
...
+
+    @mock.patch.dict(config, [('test_gating.required', True)])
+    @mock.patch('bodhi.server.models.waiverdb_api_post')
+    @mock.patch('bodhi.server.models.greenwave_api_post')
+    @mock.patch('bodhi.server.models.User.openid', mock.MagicMock(return_value=None))
+    @mock.patch('bodhi.server.models.User.avatar', mock.MagicMock(return_value=None))
+    def test_waive_test_results_2_unsatisfied_requirements(
+            self, greenwave_api_post, waiverdb_api_post, *args):
+        """Ensure that waiverdb and greenwaved are properly called when greenwave returns two
+        unsatisfied requirement."""
```
bowlofeggs: s/requirement/requirements/

bodhi/tests/server/services/test_updates.py
```diff
@@ -5219,3 +5219,165 @@ def test_waive_test_results_2_unsatisfied_requirements(
...
+
+    @mock.patch.dict(config, [('test_gating.required', True)])
+    @mock.patch('bodhi.server.models.waiverdb_api_post')
+    @mock.patch('bodhi.server.models.greenwave_api_post')
+    @mock.patch('bodhi.server.models.User.openid', mock.MagicMock(return_value=None))
+    @mock.patch('bodhi.server.models.User.avatar', mock.MagicMock(return_value=None))
+    def test_waive_test_results_1_of_2_unsatisfied_requirements(
+            self, greenwave_api_post, waiverdb_api_post, *args):
+        """Ensure that waiverdb and greenwaved are properly called when greenwave returns only one
+        unsatisfied requirement but only one of them is waived."""
```
bowlofeggs: In the test, Greenwave returns two unsatisfied requirements, not one.

bodhi/tests/server/services/test_updates.py
```diff
@@ -5219,3 +5219,165 @@ def test_waive_test_results_2_unsatisfied_requirements(
...
+
+    @mock.patch.dict(config, [('test_gating.required', True)])
+    @mock.patch('bodhi.server.models.waiverdb_api_post')
+    @mock.patch('bodhi.server.models.greenwave_api_post')
+    @mock.patch('bodhi.server.models.User.openid', mock.MagicMock(return_value=None))
+    @mock.patch('bodhi.server.models.User.avatar', mock.MagicMock(return_value=None))
+    def test_waive_test_results_2_of_2_unsatisfied_requirements(
+            self, greenwave_api_post, waiverdb_api_post, *args):
+        """Ensure that waiverdb and greenwaved are properly called when greenwave returns only one
+        unsatisfied requirement and both of them is waived."""
```
bowlofeggs: In the test, Greenwave returns two unsatisfied requirements.

Also, s/requirement and both of them is/requirements and both of them are/

bodhi/tests/server/services/test_updates.py
```diff
@@ -5219,3 +5219,165 @@ def test_waive_test_results_2_unsatisfied_requirements(
...
+                    'subject': {
+                        u'item': u'bodhi-2.0-1.fc17', u'type': u'koji_build'
+                    }
+                }
+            )
+        ]
+        self.assertEqual(waiverdb_api_post.mock_calls, calls)
+
+        self.assertEqual(list(res.json_body.keys()), ['update'])
+        self.assertEqual(res.json_body['update'], up.__json__())
```
bowlofeggs, line 5385: We should add a test asserting that Bodhi returns a bad request error if the user requests waiving a test that Greenwave wasn't failing.

bowlofeggs (changes requested):

bodhi/tests/server/services/test_updates.py
pypingou: Still todo

pypingou: Do we really need one? waiverdb will accept any waiver just fine, regardless of whether the test is failing or is required. This is going to add logic to bodhi that waiverdb itself doesn't enforce.

cverna (approved):
LGTM

pypingou:
I'll rebase :)

bowlofeggs:

bowlofeggs:

PR #2439, opened 2018-06-12
Fix tests when the timezone isn't UTC

owtaylor:
The masher tests were failing when the local date is was different from
the UTC date because the fake compose directory was generated from the
local date but checked against the UTC date.


owtaylor:
jenkies test

bowlofeggs (approved):
Thanks!

Issue #2214, opened 2018-03-13
The CLI gives unhelpful error messages if --notes are not given


bowlofeggs:
I've not identified the cause of the issue yet, but I noticed that the CLI cannot create updates:

```
$ bodhi updates new s2i-core-0-13.f27container --url http://localhost:6543
Required
You may not specify an empty list of builds.
ACL validation mechanism was unable to determine ACLs.
```

bowlofeggs:
Surprisingly, it seems that the error is due to missing notes. If I include a ```--notes``` flag, we do not get the above error:

```
$ bodhi updates new s2i-core-0-13.f27container --notes "a note"
Invalid tag: s2i-core-0-13.f27container not tagged with any of the following tags [u'f22-updates-candidate', u'dist-6E-epel-testing-candidate', u'f21-updates-candidate', u'f25-updates-candidate', u'f24-updates-candidate', u'epel7-testing-candidate', u'f27-modular-updates-candidate', u'dist-5E-epel-testing-candidate', u'f23-updates-candidate', u'f26-updates-candidate', u'f27-updates-candidate', u'f28-updates-candidate', u'f28-modular-updates-candidate', u'f27container-updates-candidate']
```

Thus it seems that we simply have an error message propagation issue, which is important to fix but not critical.

bowlofeggs:
Note: This "smells" like an API issue to me, but @cverna pointed out that it would be good for the CLI to also require the notes field. Thus, I suggest ensuring that the API hands back a reasonable error message in this case (for non CLI users) and that the CLI just errors immediately when notes aren't given (for CLI users).

bowlofeggs:
A fix for this issue is planned to be included in the upcoming 3.9.0 release: #2477.

bowlofeggs:
3.9.0 was released.

PR #2440, opened 2018-06-12
avoid spurious error messages after schema validation fails

owtaylor:
If validation against the colander schema for the request fails, then
request.validated is left empty - this will cause subsequent validation
to fail for reasons that have nothing to do with what was submitted and
returning the error messages to the user can be quite confusing.

For example: if a new update is submitted with empty notes, not only does
the user see 'Notes : Required', they also see 'Builds : You may not specify
an empty list of builds.' and 'Builds : ACL validation mechanism was unable
to determine ACLs.' - which are entirely spurious.

In all validators that are run after the colander schema validator, check
to see if a) request.validated is empty b) request.errors is non-empty,
and in this case skip further validation. This is done as a decorator to
have to avoid having dozens of separate code paths for validation-after-schema-
failure that would need to be tested/covered.

### DIFF
--- a/bodhi/server/validators.py
+++ b/bodhi/server/validators.py
@@ -19,6 +19,7 @@
 """A collection of validators for Bodhi requests."""
 
 from datetime import datetime, timedelta
+from functools import wraps
 
 from pyramid.exceptions import HTTPNotFound, HTTPBadRequest
 from pyramid.httpexceptions import HTTPFound, HTTPNotImplemented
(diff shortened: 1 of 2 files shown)


bowlofeggs (approved):
Nice!

This seems to help a bit with #2214 though it still doesn't fully fix it (but a step in a better direction!)

```
$ bodhi updates new bodhi-2.8.1-1.fc28 
Required
```

Still weird and mysterious, but less weird and mysterious than what was happening before.

Thanks!

bowlofeggs:

bowlofeggs:

PR #2454, opened 2018-06-15
Auto clean composes after successfully completing a compose.

### LINKED ISSUE #1304: The masher should clean up old mashes at the end of every run

ghost:
fixes #1304

### DIFF
--- a/bodhi/server/consumers/masher.py
+++ b/bodhi/server/consumers/masher.py
@@ -48,6 +48,7 @@ from bodhi.server.exceptions import BodhiException
 from bodhi.server.metadata import UpdateInfoMetadata
 from bodhi.server.models import (Compose, ComposeState, Update, UpdateRequest, UpdateType, Release,
                                  UpdateStatus, ReleaseState, ContentType)
+from bodhi.server.scripts import clean_old_mashes
 from bodhi.server.util import (copy_container, sorted_updates, sanity_check_repodata,
                                transactional_session_maker)
 
@@ -429,8 +430,13 @@ class ComposerThread(threading.Thread):
             self.check_all_karma_thresholds()
             self.obsolete_older_updates()
 
+            # Clean old composes
+            self.save_state(ComposeState.cleaning)
+            clean_old_mashes.remove_old_composes()
+
             self.save_state(ComposeState.success)
             self.success = True
+
             self.remove_state()
 
         except Exception:
--- a/bodhi/server/migrations/versions/2c5d45f0c932_add_a_new_state_for_cleaning_composes.py
+++ b/bodhi/server/migrations/versions/2c5d45f0c932_add_a_new_state_for_cleaning_composes.py
@@ -1,0 +1,50 @@
+# Copyright (c) 2018 Red Hat, Inc.
+#
+# This file is part of Bodhi.
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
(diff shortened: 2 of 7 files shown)


bodhi/server/migrations/versions/2c5d45f0c932_add_a_new_state_for_cleaning_composes.py
```diff
@@ -0,0 +1,50 @@
...
+        op.execute('SET LOCAL bdr.permit_ddl_locking = true')
+    except exc.ProgrammingError:
+        # This server doesn't use BDR, so no problem.
+        pass
+    op.execute("ALTER TYPE ck_compose_state ADD VALUE 'cleaning' AFTER 'signing_repo'")
+
+
+def downgrade():
+    """Raise an exception explaining that this migration cannot be reversed."""
+    raise NotImplemented('This migration cannot be reversed.')
```
bowlofeggs: NotImplemented has been deprecated in favor of NotImplementedError.

bodhi/server/scripts/clean_old_mashes.py
```diff
@@ -33,6 +33,12 @@
 @click.command()
 @click.version_option(message='%(version)s')
 def clean_up():
+    """Delete any repo mashes that are older than the newest 10 from each repo series."""
+    _clean_up_()
+
+
+# Helper function used in auto clean composes (masher.py)
+def _clean_up_():
```
bowlofeggs: Let's give this a public name (no leading underscore). To avoid clashing with the other function, let's call this ```remove_old_composes()```.

bodhi/server/consumers/masher.py
```diff
@@ -431,6 +432,11 @@ def work(self):
 
             self.save_state(ComposeState.success)
             self.success = True
+
+            # Clean old composes after successful completion
+            self.save_state(ComposeState.cleaning)
+            clean_old_mashes._clean_up_()
```
bowlofeggs: Let's move this to be before the success state.

bodhi/server/util.py
```diff
@@ -491,6 +491,7 @@ def composestate2html(context, state):
         'punging': 'warning',
         'notifying': 'warning',
         'success': 'success',
+        'cleaning': 'success',
```
bowlofeggs: Let's use the warning class here. The name warning is weird, but we're really using it to get things to be orange/yellow in the UI and the cleaning is running so we want to indicate that things aren't done yet.

bodhi/tests/server/test_util.py
```diff
@@ -446,6 +446,11 @@ def test_failed(self):
...
                          "<span class='label label-danger'>Failed</span>")
 
+    def test_cleaning(self):
+        """Assert correct return value with the cleaning state."""
+        self.assertEqual(util.composestate2html(None, ComposeState.cleaning),
+                         "<span class='label label-success'>Cleaning old composes</span>")
+
 
 class TestCanWaiveTestResults(base.BaseTestCase):
     """Test the can_waive_test_results() function."""
```
bowlofeggs: We need a test to assert that the composer calls the clean up code.

bowlofeggs (changes requested):

cverna:
jenkies test

bodhi/server/consumers/masher.py
```diff
@@ -429,8 +430,13 @@ def work(self):
             self.check_all_karma_thresholds()
             self.obsolete_older_updates()
 
+            # Clean old composes after successful completion
```
cverna: I think you can remove the `after successful completion` from the comment since this is not true any more.

cverna (commented):
one small nitpick, but it looks good :)

cverna (approved):
:+1:

PR #2457, opened 2018-06-18
Dropped uniqueness constraint on user email

### LINKED ISSUE #2387: Drop Bodhi's uniqueness constraint on e-mails
Bodhi currently has a uniqueness constraint on e-mails. This causes problems when users have more than one account and then change an account's e-mail address to be the same as the other accounts. When they log into Bodhi, they will get a crash because Bodhi hits the e-mail field's uniqueness constraint check. Bodhi does not look up users by e-mail address, and only uses the field to send e-mails, so it's ok if two accounts have the same address. We don't even need to index the field because we don't look up or search users by e-mail address.

ghost:
fixes #2387

### DIFF
--- a/bodhi/server/migrations/versions/39132c56a47e_drop_uniqueness_constraint_on_user_email.py
+++ b/bodhi/server/migrations/versions/39132c56a47e_drop_uniqueness_constraint_on_user_email.py
@@ -1,0 +1,40 @@
+# Copyright (c) 2018 Red Hat, Inc.
+#
+# This file is part of Bodhi.
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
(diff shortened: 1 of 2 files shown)


bodhi/server/migrations/versions/39132c56a47e_drop_uniqueness_constraint_on_user_email.py
```diff
@@ -0,0 +1,40 @@
+# Copyright (c) 2017 Red Hat, Inc.
```
bowlofeggs: Let's update the year to 2018.

bowlofeggs (changes requested):

vismay-golwala:
jenkies test

vismay-golwala:
Jenkies test

bowlofeggs:
jenkies test

bowlofeggs (approved):
This LGTM, but should not be merged until #2454 is merged because it references a ```down_revision``` that is created there.

PR #2458, opened 2018-06-18
Only use comments after last reset

### LINKED ISSUE #1996: bodhi.server.models.Update.num_admin_approvals includes invalid comments
[```bodhi.server.models.Update.num_admin_approvals```](https://github.com/fedora-infra/bodhi/blob/3.0.0/bodhi/server/models.py#L2530-L2542) uses ```self.comments``` when it should use [```self.comments_since_karma_reset()```](https://github.com/fedora-infra/bodhi/blob/3.0.0/bodhi/server/models.py#L1210-L1230) instead.

This means that admin approvals are not cleared when developers change the Builds associated with an Update.

ghost:
fixes #1996

### DIFF
--- a/bodhi/server/models.py
+++ b/bodhi/server/models.py
@@ -1882,11 +1882,13 @@ class Update(Base):
         Karma is reset when :class:`Builds <Build>` are added or removed from an update.
 
         Returns:
-            generator: :class:`Comments <Comment>` since the karma reset.
+            comments (list): class:`Comments <Comment>` since the karma reset.
         """
         # We want to traverse the comments in reverse order so we only consider
         # the most recent comments from any given user and only the comments
         # since the most recent karma reset event.
+        comments_since_karma_reset = []
+
         for comment in reversed(self.comments):
             if (comment.user.name == u'bodhi' and
                     ('New build' in comment.text or 'Removed build' in comment.text)):
@@ -1896,7 +1898,9 @@ class Update(Base):
                 # reverse order, once we find one of these comments we can
                 # simply exit this loop.
                 break
-            yield comment
+            comments_since_karma_reset.append(comment)
+
+        return comments_since_karma_reset
 
     @staticmethod
     def contains_critpath_component(builds, release_name):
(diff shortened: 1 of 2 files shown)


vismay-golwala:
jenkies test

bodhi/server/models.py
```diff
@@ -2396,7 +2397,8 @@ def get_testcase_karma(self, testcase):
             positive karma.
         """
         good, bad, seen = 0, 0, set()
-        for comment in reversed(self.comments):
+        # Reversing a generator object {self.comments_since_karma_reset}
+        for comment in reversed([_ for _ in self.comments_since_karma_reset]):
```
cverna: see comment above

bodhi/server/models.py
```diff
@@ -2846,7 +2848,8 @@ def __str__(self):
         val += u"""
   Submitter: %s
   Submitted: %s\n""" % (username, self.date_submitted)
-        if len(self.comments):
+        # Counting length of a generator object {self.comments_since_karma_reset}
+        if sum(1 for _ in self.comments_since_karma_reset):
```
cverna: PEP8 recommends the following 
```
Yes: if not seq:
        if seq:

No: if len(seq):
       if not len(seq):
```
So we should drop the `if len()`/`if sum()` here. 

I would recommend converting `self.comments_since_karma_reset` into a list and store it in a variable (for example `comments_since_karma_reset`) then use this variable in the if statement.

bodhi/server/models.py
```diff
@@ -2846,7 +2848,8 @@ def __str__(self):
...
             val += u"   Comments: "
             comments = []
             for comment in self.comments:
```
cverna: I think we would want to use the `comments_since_karma_reset` list here

cverna (changes requested):

bodhi/server/models.py
```diff
@@ -2373,7 +2373,8 @@ def get_bug_karma(self, bug):
             positive karma.
         """
         good, bad, seen = 0, 0, set()
-        for comment in reversed(self.comments):
+        # Reversing a generator object {self.comments_since_karma_reset}
+        for comment in reversed([_ for _ in self.comments_since_karma_reset]):
```
cverna: If you check `self.comments_since_karma_reset` is already reversing the order of the comments. So it should not be needed here.

https://github.com/fedora-infra/bodhi/blob/develop/bodhi/server/models.py#L1890

cverna (approved):
LGTM

PR #2460, opened 2018-06-20
Single notification for BZ ticket update

### LINKED ISSUE #336: Do not send multiple notifications due to BZ ticket updates
When my update was pushed into stable, I received two notifications from BZ [1], one with comment and second with remaining flags modified. Is there chance to do this in one action, preventing BZ to send multiple emails?

[1] https://bugzilla.redhat.com/show_bug.cgi?id=1253872

ghost:
fixes #336

### DIFF
--- a/bodhi/server/consumers/updates.py
+++ b/bodhi/server/consumers/updates.py
@@ -194,9 +194,8 @@ class UpdatesHandler(fedmsg.consumers.FedmsgConsumer):
                 log.info("Commenting on %r" % bug.bug_id)
                 comment = config['initial_bug_msg'] % (
                     update.title, update.release.long_name, update.abs_url())
-                bug.add_comment(update, comment)
 
                 log.info("Modifying %r" % bug.bug_id)
-                bug.modified(update)
+                bug.modified(update, comment)
             except Exception:
                 log.warning('Error occurred during updating single bug', exc_info=True)
--- a/bodhi/server/models.py
+++ b/bodhi/server/models.py
@@ -4164,9 +4164,11 @@ class Bug(Base):
         ])
         bugs.bugtracker.close(self.bug_id, versions=versions, comment=self.default_message(update))
 
-    def modified(self, update):
+    def modified(self, update, comment):
         """
         Change the status of this bug to MODIFIED unless it is a parent security bug.
+
+        Also, comment on the bug stating that an update has been submitted.
 
         Args:
             update (Update): The update that is associated with this bug.
@@ -4174,7 +4176,7 @@ class Bug(Base):
         if update.type is UpdateType.security and self.parent:
             log.debug('Not modifying on parent security bug %s', self.bug_id)
         else:
-            bugs.bugtracker.modified(self.bug_id)
+            bugs.bugtracker.modified(self.bug_id, comment)
 
 
 user_group_table = Table('user_group_table', Base.metadata,
--- a/bodhi/tests/server/test_bugs.py
+++ b/bodhi/tests/server/test_bugs.py
@@ -268,17 +268,31 @@ class TestBugzilla(unittest.TestCase):
     @mock.patch('bodhi.server.bugs.log.info')
     @mock.patch.dict('bodhi.server.bugs.config', {'bz_products': 'aproduct'})
     def test_modified(self, info):
-        """Test the modified() method"""
+        """Ensure correct execution of the modified() method."""
         bz = bugs.Bugzilla()
         bz._bz = mock.MagicMock()
         bz._bz.getbug.return_value.product = 'aproduct'
         bz._bz.getbug.return_value.bug_status = 'NEW'
 
-        bz.modified(1411188)
+        bz.modified(1411188, 'A mean message.')
 
         bz._bz.getbug.assert_called_once_with(1411188)
         info.assert_called_once_with("Setting bug #1411188 status to MODIFIED")
-        self.assertEqual(bz._bz.getbug.return_value.setstatus.call_count, 1)
+        bz._bz.getbug.return_value.setstatus.assert_called_once_with('MODIFIED',
+                                                                     comment='A mean message.')
+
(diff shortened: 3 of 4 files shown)


bodhi/server/models.py
```diff
@@ -4164,17 +4164,19 @@ def close_bug(self, update):
...
+        Also, comment on the bug stating that an update has been submitted.
+
         Args:
             update (Update): The update that is associated with this bug.
         """
         if update.type is UpdateType.security and self.parent:
             log.debug('Not modifying on parent security bug %s', self.bug_id)
         else:
-            bugs.bugtracker.modified(self.bug_id)
+            bugs.bugtracker.modified(self.bug_id, comment)
```
cverna, line 4179: I think it would be nice to follow what is done in the `testing` method and set the comment value here. 
That way the `modified` method in models.py does not require an extra `comment` argument

See https://github.com/fedora-infra/bodhi/blob/develop/bodhi/server/models.py#L4156

bodhi/server/consumers/updates.py
```diff
@@ -194,9 +194,8 @@ def work_on_bugs(self, session, update, bugs):
                 log.info("Commenting on %r" % bug.bug_id)
                 comment = config['initial_bug_msg'] % (
```
cverna: if we choose to define the comment in models.py `modified` method this is not needed anymore.

bodhi/server/consumers/updates.py
```diff
@@ -194,9 +194,8 @@ def work_on_bugs(self, session, update, bugs):
...
                     update.title, update.release.long_name, update.abs_url())
-                bug.add_comment(update, comment)
 
                 log.info("Modifying %r" % bug.bug_id)
-                bug.modified(update)
+                bug.modified(update, comment)
```
cverna: same here you will not need to pass the comment.

bodhi/tests/server/test_bugs.py
```diff
@@ -274,20 +274,33 @@ def test_modified(self, info):
...
-        bz.modified(1411188)
+        bz.modified(1411188, 'A mean message.')
 
         bz._bz.getbug.assert_called_once_with(1411188)
         info.assert_called_once_with("Setting bug #1411188 status to MODIFIED")
         self.assertEqual(bz._bz.getbug.return_value.setstatus.call_count, 1)
 
+    @mock.patch.dict('bodhi.server.bugs.config', {'bz_products': 'aproduct'})
+    def test_modified_after_verified(self):
+        """Test the modified() method"""
```
cverna: the docstring needs to be updated here to describe the test case

bodhi/tests/server/test_bugs.py
```diff
@@ -274,20 +274,33 @@ def test_modified(self, info):
         bz._bz.getbug.return_value.product = 'aproduct'
         bz._bz.getbug.return_value.bug_status = 'NEW'
 
-        bz.modified(1411188)
+        bz.modified(1411188, 'A mean message.')
 
         bz._bz.getbug.assert_called_once_with(1411188)
         info.assert_called_once_with("Setting bug #1411188 status to MODIFIED")
```
cverna, line 280: I guess we could add an assert to check the comment here

bodhi/tests/server/test_bugs.py
```diff
@@ -300,7 +313,7 @@ def test_modified_exception(self, exception_log):
         bz._bz = mock.MagicMock()
         bz._bz.getbug.side_effect = Exception()
 
-        bz.modified(1411188)
+        bz.modified(1411188, 'A mean message.')
 
         bz._bz.getbug.assert_called_once_with(1411188)
```
cverna, line 319: I guess we could add an assert to check the comment here

bodhi/tests/server/test_bugs.py
```diff
@@ -274,20 +274,33 @@ def test_modified(self, info):
...
         """Test the modified() method when the bug's product is not in the bz_products config."""
         bz = bugs.Bugzilla()
         bz._bz = mock.MagicMock()
         bz._bz.getbug.return_value.product = 'not fedora!'
 
-        bz.modified(1411188)
+        bz.modified(1411188, 'A mean message.')
 
         bz._bz.getbug.assert_called_once_with(1411188)
         info.assert_called_once_with("Skipping set modified on 'not fedora!' bug #1411188")
```
cverna, line 307: I guess we could add an assert to check the comment here

cverna (changes requested):

bodhi/server/models.py
ghost: I see the use of a config template ```config['testing_bug_msg']``` in the ```default_message``` method called by ```testing```. Is that reusable? Or we need to define a new template for ```modified```?

cverna: hum yeah good point I thought that the bug.default_message was using the `config['testing_bug_msg']` template. So let's keep this the way you have done it :+1:

bodhi/tests/server/test_bugs.py
ghost: Here, it is testing the case when bug's product is not in bz_products and so it skips the functionality and returns. Hence, I believe, the comment assertion will not be possible because it is not getting executed.

bodhi/tests/server/test_bugs.py
ghost: We are adding an ```Exception``` side effect to the ```getbug``` method, and hence the remaining part will not be executed. Control will go to catch block and assertion of comment will not be possible.

bodhi/tests/server/test_bugs.py
cverna: Sure let's keep this test as it is then :+1:

bodhi/tests/server/test_bugs.py
cverna: Make sense, I copied paste these comments a bit too fast :speak_no_evil:

cverna:
LGTM :+1:

Issue #2189, opened 2018-03-05
Bodhi does not recognize/process karma on kernel regression test


heis2201:
Leaving karma on kernel regression test cases is not shown on the corresponding site (example: https://bodhi.fedoraproject.org/updates/FEDORA-2018-817647a89b)

dhgutteridge:
Presumably this is related to the regression I reported in #2130.

bowlofeggs:
Might also be related to #2456.

cverna:
Solved by #2461

bowlofeggs:
A fix for this issue is planned to be included in the upcoming 3.9.0 release: #2477.

bowlofeggs:
The patch for this issue was deployed to Fedora's production Bodhi instance yesterday.

Issue #2130, opened 2018-01-20
Bodhi has stopped registering feedback that references bug IDs the submitter has associated


dhgutteridge:
Until some time a couple of days ago, when I included feedback on a specific bug ID the submitter associated with a Bodhi update, it would register visibly in the content posted. Now it seems to have stopped doing this.

For example, I ticked off a bug in this submission a few days ago, and it's registered and visible as one would expect: [https://bodhi.fedoraproject.org/updates/FEDORA-2018-462f982e20](https://bodhi.fedoraproject.org/updates/FEDORA-2018-462f982e20)

However, I just did the same for another submission, and it didn't register. It doesn't show up in the content posted. Selecting the same bug in a second feedback submission just ends up with entirely empty content. (The same is true for the email generated.) This example is: [https://bodhi.fedoraproject.org/updates/FEDORA-2018-6f90feaeee](https://bodhi.fedoraproject.org/updates/FEDORA-2018-6f90feaeee)

cverna:
Solved by #2461

bowlofeggs:
A fix for this issue is planned to be included in the upcoming 3.9.0 release: #2477.

bowlofeggs:
The patch for this issue was deployed to Fedora's production Bodhi instance yesterday.

PR #2461, opened 2018-06-21
Resolved bug/testcase specific karma error

### LINKED ISSUE #2456: bugs can't be marked as fixed
In https://bodhi.fedoraproject.org/updates/FEDORA-2018-1a6e6196b9 I tried to mark #1580530 as fixed by giving it +1 feedback (directly for that bug). But only generic karma was shown in my [comment](https://bodhi.fedoraproject.org/updates/ansible-2.5.5-2.fc28#comment-800652), not the specific bug report karma. Also in Bugs tab the feedback was not shown. I tried again (this time without giving generic karma), and [again](https://bodhi.fedoraproject.org/updates/ansible-2.5.5-2.fc28#comment-800653) it was ignored.

It seems that giving karma for particular bugs is broken atm.

ghost:
fixes #2456, #2189, #2130

### DIFF
--- a/bodhi/server/schemas.py
+++ b/bodhi/server/schemas.py
@@ -145,6 +145,11 @@ class TestcaseFeedbacks(colander.SequenceSchema):
 class SaveCommentSchema(CSRFProtectedSchema, colander.MappingSchema):
     """An API schema for bodhi.server.services.comments.new_comment()."""
 
+    def deserialize(self, cstruct):
+        """Unflatten comment before parsing into Schema."""
+        appstruct = SaveCommentSchema().unflatten(cstruct)
+        return super(SaveCommentSchema, self).deserialize(appstruct)
+
     update = colander.SchemaNode(colander.String())
     text = colander.SchemaNode(
         colander.String(),
--- a/bodhi/tests/server/services/test_comments.py
+++ b/bodhi/tests/server/services/test_comments.py
@@ -104,7 +104,8 @@ class TestCommentsService(base.BaseTestCase):
     @mock.patch('bodhi.server.notifications.publish')
     def test_commenting_with_bug_feedback(self, publish):
         comment = self.make_comment()
-        comment['bug_feedback'] = [{'bug_id': 12345, 'karma': 1}]
+        comment['bug_feedback.0.bug_id'] = 12345
+        comment['bug_feedback.0.karma'] = 1
         res = self.app.post_json('/comments/', comment)
         self.assertNotIn('errors', res.json_body)
         self.assertIn('comment', res.json_body)
@@ -119,7 +120,8 @@ class TestCommentsService(base.BaseTestCase):
     @mock.patch('bodhi.server.notifications.publish')
     def test_commenting_with_testcase_feedback(self, publish):
         comment = self.make_comment()
-        comment['testcase_feedback'] = [{'testcase_name': "Wat", 'karma': -1}]
+        comment['testcase_feedback.0.testcase_name'] = "Wat"
+        comment['testcase_feedback.0.karma'] = -1
         res = self.app.post_json('/comments/', comment)
         self.assertNotIn('errors', res.json_body)
         self.assertIn('comment', res.json_body)


bodhi/server/validators.py
```diff
@@ -989,6 +989,39 @@ def validate_comment_id(request, **kwargs):
...
+
+    Args:
+        request (pyramid.util.Request): The current request.
+        kwargs (dict): The kwargs of the related service definition. Unused.
+    """
+    # In-method import used because there is mutual referencing between
+    # 'bodhi.server.schemas' and 'bodhi.server.validators'
+    from bodhi.server.schemas import SaveCommentSchema
+
+    content = str(request.body)
```
cverna: I think here you could use the [request.json ](https://docs.pylonsproject.org/projects/pyramid/en/latest/api/request.html#pyramid.request.Request.json)attribute that will give you directly a dictionary to work with.

bodhi/server/validators.py
```diff
@@ -989,6 +989,39 @@ def validate_comment_id(request, **kwargs):
...
+
+    if content:
+        content = eval(content)
+
+        if type(content) is dict:
+            if 'bug_feedback' not in content and 'testcase_feedback' not in content:
+                content = SaveCommentSchema().unflatten(content)
+            content = str(content)
+
+    if type(content) is str:
```
cverna: You should check the pyramid escape function for this. See https://docs.pylonsproject.org/projects/pyramid/en/latest/_modules/html.html#escape

bodhi/server/validators.py
```diff
@@ -989,6 +989,39 @@ def validate_comment_id(request, **kwargs):
...
+        request.body = content.replace("'", "\"")
+    if type(content) is bytes:
+        request.body = content.replace(b"'", b"\"")
```
cverna: What is the use case of receiving bytes in the body ?

cverna (changes requested):

bodhi/server/validators.py
ghost: But that won't be set in test environment. So should we set that also in testing or make a decision based logic of whether parameters are set?

cverna: in the test instead of setting `request.body` you should be able to set `request.json` instead.

ghost: ```request.json``` stores every key-value pair in ```unicode``` format. However, the ```colander``` parser requires it to be ```str```. So I can either convert each key-value of ```request.json``` into ```str``` or I can use the method above. Which one would be more efficient according to you?

bodhi/server/validators.py
ghost: Testing using python 3, it sets the ```request.body``` as ```bytes``` and in that case, force converting into ```str``` causes the following error:
```TypeError: You can only set Request.body to bytes (not <class 'str'>)```

That's why, I kept separate logics for python 2 and 3.

File: bodhi/server/validators.py (hunk 2 above)
Comment (ghost) [reply bodhi/server/validators.py]: In the line you are suggesting, I am not escaping quotes - but instead converting single quotes to double quotes because that is the format colander parses string in.

However, it might be good to actually escape the ```text``` field of request to filter HTML characters.

Comment (cverna) [reply bodhi/server/validators.py]: ok I need to look at colander in more details :small_airplane:

**Comment by cverna** [comment]
@vismay-golwala, I gave a quick look at [colander](https://docs.pylonsproject.org/projects/colander/en/latest/basics.html)  and I don't think we should use a validator to format the request body the way colander is validating it.

I think we should look at the colander deserialize or preparer to format the data before they get validated.

Also I tried running this code using Python 3 and it is not working. I suspect the `str()` calls are not doing what you expect in Python 3.

**Comment by ghost** [comment]
@cverna , so the process of execution is somewhat like this:
```submit_feedback (request object) -> ColanderSchema (deserialize) -> Preparer -> Validator -> rest_of_code```


```deserializing``` needs the data to be 'unflattened'. So I think the way to achieve it is we filter the request object before it goes to ```ColanderSchema```. This is not data validation, but instead, request validation which just converts it into an appropriate format for deserialization. 


If we go for ```Preparer```, it gets executed after deserialization. So, if the data was not formatted before, it would cause ```bug_feedback``` or ```testcase_feedback``` to be empty when it reaches ```Preparer```.


Regarding Python 3 issue, can you please tell me how to run the server instance using Py3 instead of 2?

File: .pytest_cache/v/cache/lastfailed [hunk 4]
```diff
@@ -0,0 +1 @@
+{}
```
cverna: this file should not be part of the commit. You could add `.pytest_cache` to the `.gitignore` file

bodhi/server/schemas.py
```diff
@@ -145,6 +145,11 @@ class TestcaseFeedbacks(colander.SequenceSchema):
 class SaveCommentSchema(CSRFProtectedSchema, colander.MappingSchema):
     """An API schema for bodhi.server.services.comments.new_comment()."""
 
+    def deserialize(self, cstruct):
+        """Unflatten comment before parsing into Schema."""
+        appstruct = SaveCommentSchema().unflatten(cstruct)
+        return super(SaveCommentSchema, self).deserialize(appstruct)
+
```
cverna, line 152: Nice :+1:

cverna (changes requested):
Code looks good, there is just one file that should not be included in the commit

cverna (approved):
+1

PR #2464, opened 2018-06-28
Resolved update edit bug while using CLI

### LINKED ISSUE #2208: Bodhi's REST API should separate the create and edit endpoints
While working on #1814 today I discovered that the REST API will set an update's severity to "unspecified" if the POST data does not include the severity field. This means that users who edit an update with the CLI and don't specify a ```--severity``` will end up altering the update to reset its severity to "unspecified".

ghost:
fixes #2208

### DIFF
--- a/bodhi/client/__init__.py
+++ b/bodhi/client/__init__.py
@@ -370,10 +370,22 @@ def edit(user, password, url, **kwargs):
             resp = client.query(**query_param)
             title = resp['updates'][0]['title']
         elif re.search(bindings.UPDATE_TITLE_RE, kwargs['update']):
+            query_param = {'like': kwargs['update']}
+            resp = client.query(**query_param)
             title = kwargs['update']
         del(kwargs['update'])
         kwargs['builds'] = title
         kwargs['edited'] = title
+
+        # Convert list of 'Bug' instances in DB to comma separated bug_ids for parsing.
+        former_update = resp['updates'][0]
+        if not kwargs['bugs']:
+            kwargs['bugs'] = ",".join([str(bug['bug_id']) for bug in former_update['bugs']])
+
+        # Replace empty fields with former values from database.
+        for field in kwargs:
+            if kwargs[field] in (None, '') and field in former_update:
+                kwargs[field] = former_update[field]
 
         resp = client.save(**kwargs)
         print_resp(resp, client)
--- a/bodhi/tests/client/test___init__.py
+++ b/bodhi/tests/client/test___init__.py
@@ -1121,12 +1121,12 @@ class TestEdit(unittest.TestCase):
             mock.call(
                 bindings_client, 'updates/', auth=True, verb='POST',
                 data={
-                    'close_bugs': False, 'stable_karma': None, 'csrf_token': 'a_csrf_token',
+                    'close_bugs': False, 'stable_karma': 3, 'csrf_token': 'a_csrf_token',
                     'staging': False, 'builds': u'nodejs-grunt-wrap-0.3.0-2.fc25',
                     'autokarma': False, 'edited': u'nodejs-grunt-wrap-0.3.0-2.fc25',
-                    'suggest': None, 'notes': None, 'notes_file': None, 'requirements': None,
-                    'request': None, 'bugs': u'', 'unstable_karma': None, 'type': 'bugfix',
-                    'severity': 'low'
+                    'suggest': u'unspecified', 'notes': u'New package.',
+                    'notes_file': None, 'request': None, 'unstable_karma': -3,
+                    'bugs': '1420605', 'requirements': u'', 'type': 'bugfix', 'severity': u'low'
                 }
             ),
             mock.call(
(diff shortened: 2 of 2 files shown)


bodhi/client/__init__.py
```diff
@@ -370,11 +370,25 @@ def edit(user, password, url, **kwargs):
...
             title = resp['updates'][0]['title']
         elif re.search(bindings.UPDATE_TITLE_RE, kwargs['update']):
+            query_param = {'like': kwargs['update']}
+            resp = client.query(**query_param)
             title = kwargs['update']
         del(kwargs['update'])
         kwargs['builds'] = title
         kwargs['edited'] = title
 
+        data = munch.unmunchify(resp)
```
cverna: you should be able to use the Munch object just like a dictionary, so I am not sure this is needed

bodhi/client/__init__.py
```diff
@@ -370,11 +370,25 @@ def edit(user, password, url, **kwargs):
...
+        if 'updates' in data:
+            former_update = data['updates'][0]
+        else:
+            former_update = data
+
+        # Replace None params (not provided) with previous database values
+        for k in kwargs:
+            if not kwargs[k] and k in former_update:
```
cverna: What happens here when `kwargs[k] is False` ? it will use the value of `former_update` which is not what we expect.
I think here checking for `kwargs[k] is None and ...` would work.

bodhi/client/__init__.py
```diff
@@ -370,11 +370,25 @@ def edit(user, password, url, **kwargs):
...
+                if former_update[k]:
```
cverna: This will not work when `former_update[k] is Flase` . I don't think we need a test here.

bodhi/server/config.py
```diff
@@ -1,3 +1,4 @@
+
```
cverna: Looks like this change should not have been committed.

cverna (changes requested):

bodhi/client/__init__.py
ghost: Yes, that is a good point. Will make the changes.

bodhi/client/__init__.py
ghost: Yes, will modify to just include None. Although, I think we might need this test because the line above only checks for key k in the dictionary. So if there is a None value by default and we try to update that to something else, it will again be replaced by None.

cverna:
jenkies test

bodhi/client/__init__.py
```diff
@@ -370,11 +371,23 @@ def edit(user, password, url, **kwargs):
...
+            resp = client.query(**query_param)
             title = kwargs['update']
         del(kwargs['update'])
         kwargs['builds'] = title
         kwargs['edited'] = title
 
+        # Convert list of 'Bug' objects in DB to comma separated bug_ids for parsing.
+        # Creating deepcopy prevents original object to be modified (useful in test environment).
+        former_update = copy.deepcopy(resp['updates'][0])
+        former_update['bugs'] = ",".join([str(bug['bug_id']) for bug in former_update['bugs']])
```
cverna: you might be able to avoid the deepcopy, buy doing this in the for loop. That way you would populate ```kwargs['bugs']``` instead of ```former_update['bugs']```

bodhi/tests/client/test_bindings.py
```diff
@@ -858,6 +858,25 @@ def test_without_type_(self):
         kwargs['csrf_token'] = 'a token'
         client.send_request.assert_called_once_with('updates/', verb='POST', auth=True, data=kwargs)
 
+    def test_with_edited(self):
```
cverna: This test seems to do the same assertions as  the `test_without_type_`. So I am not sure this is needed.

ghost:
jenkies test

cverna:
@vismay-golwala you will need to rebase this PR since there is a conflicting file. I think that's the reason why the tests are failing.

bodhi/client/__init__.py
```diff
@@ -370,11 +370,22 @@ def edit(user, password, url, **kwargs):
...
+            query_param = {'like': kwargs['update']}
+            resp = client.query(**query_param)
             title = kwargs['update']
         del(kwargs['update'])
         kwargs['builds'] = title
         kwargs['edited'] = title
 
+        # Convert list of 'Bug' instances in DB to comma separated bug_ids for parsing.
+        former_update = resp['updates'][0]
+        kwargs['bugs'] = ",".join([str(bug['bug_id']) for bug in former_update['bugs']])
```
cverna: You need to do this in the `for field in kwargs` loop after checking that `kwargs['bugs'] is None` otherwise this will delete bugs that the user gave using the CLI.

cverna (changes requested):
Just one comment, but otherwise this is looking good :+1:

bodhi/client/__init__.py
ghost: Keeping check inside the for loop will add an extra if and that will repeat multiple times. Can we do a check separately outside like this?

```
if not kwargs['bugs']:
       kwargs['bugs'] = ",".join([str(bug['bug_id']) for bug in former_update['bugs']])
```

cverna: Yes I think that would work :+1:

cverna (approved):
LGTM thanks :small_airplane:

PR #2465, opened 2018-07-02
Include the CI status in the text representation of an update in the CLI

pypingou:

### DIFF
--- a/bodhi/client/bindings.py
+++ b/bodhi/client/bindings.py
@@ -44,9 +44,10 @@ except ImportError:  # pragma: no cover
     # dnf is not available on EL 7.
     dnf = None  # pragma: no cover
 import koji
+import requests.exceptions
 import six
 
-from fedora.client import AuthError, OpenIdBaseClient, FedoraClientError
+from fedora.client import AuthError, OpenIdBaseClient, FedoraClientError, ServerError
 import fedora.client.openidproxyclient
 
 
@@ -349,6 +350,18 @@ class BodhiClient(OpenIdBaseClient):
         if 'bugs' in kwargs and kwargs['bugs'] == '':
             kwargs['bugs'] = None
         return self.send_request('updates/', verb='GET', params=kwargs)
+
+    def get_test_status(self, update):
+        """
+        Query bodhi for the test status of the specified update..
+
+        Args:
+            update (basestring): The title or identifier of the update to
+                retrieve the test status of.
+        Returns:
+            munch.Munch: The response from Bodhi describing the query results.
+        """
+        return self.send_request('updates/%s/get-test-results' % update, verb='GET')
 
     @errorhandled
     def comment(self, update, comment, karma=0, email=None):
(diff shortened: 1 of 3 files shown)


pypingou:
I love when tests pass locally and not on jenkins...

bodhi/client/bindings.py
```diff
@@ -725,6 +738,20 @@ def update_str(self, update, minimal=False):
...
+        try:
+            test_status = self.get_test_status(update['alias'])
+        except (ServerError, requests.exceptions.RequestException):
+            test_status = None
+
+        if test_status:
+            info = None
+            if 'errors' in test_status:
+                info = '\n'.join([el.description for el in test_status.errors])
+            elif 'decision' in test_status:
```
cverna, line 750: Can we have any other case here ? Or could we do something like :  
``` 
if 'decision' in test_status: 
    info = .... (we have the decision)
else:
    info = .... (this is an error)
```

bodhi/client/bindings.py
```diff
@@ -725,6 +738,20 @@ def update_str(self, update, minimal=False):
                 update['autokarma'], update['unstable_karma'], update['stable_karma']))
         ]
 
+        try:
+            test_status = self.get_test_status(update['alias'])
+        except (ServerError, requests.exceptions.RequestException):
+            test_status = None
+
+        if test_status:
```
cverna, line 746: `if test_status is None` should we display that we couldn't retrieve the status ? something like `CI Status: error while requesting the tests status` ?

pypingou: Sure could do. My original thought was that I didn't want to show the ``CI Status`` line for bodhi instances not using CI and/or greenwave.
What do you think?

bodhi/client/bindings.py
pypingou: I had this construct first but that lead to errors in the unit-tests. A bunch of the test define what ``send_request`` returns so that it returns the JSON representation of an update (for example) and that would make this code fail since it couldn't find neither ``errors`` nor ``decision``, so I went with this construct albeit a little more odd but leading to less changes overall.

bodhi/client/bindings.py
cverna: I can see that we don't want to show the `CI Status` for updates that are not using CI.  
But that will lead to the case where an update using CI fails to get the tests status, this will just silently fail. This is maybe not a use case that is worth considering.

bodhi/client/bindings.py
cverna: Did that impact many tests ? I just feel that it would a shame to make the logic more complex because of the tests.

bodhi/client/bindings.py
pypingou: The cases that we are hiding are the ones where the URL entered in bodhi's config is invalid (ConnectionError: https://ci.centos.org/job/bodhi-bodhi/1738/console) and the case where the data returned isn't JSON: https://github.com/fedora-infra/python-fedora/blob/develop/fedora/client/openidbaseclient.py#L252), the other error cases are handled by bodhi itself and should return both JSON and an informative message that we will display :)

bodhi/client/bindings.py
pypingou: Getting the number for you, but basically:
````
bodhi/tests/client/test___init__.py ............F..FF..FF.FFF.F...F.F.......FFFFF..............F.
````

bodhi/client/bindings.py
cverna: Ok that's sounds good :+1:

bodhi/client/bindings.py
cverna: Ok let's keep the current logic then :+1:

cverna (approved):
:ship: it

bowlofeggs:

bowlofeggs:

Issue #2433, opened 2018-06-11
Bodhi should know which packages are critical path without outside help


bowlofeggs:
PDC is going away, and Bodhi uses it to know which packages are critical path. It seems that Bodhi is the only application in Fedora Infrastructure that cares about whether a package is critical path, so it is sensible to move this information into Bodhi's database. This will reduce one of our external service dependencies, which should increase reliability.

Packages are critical path *per release*, which means that a package, for example, might be critical path in Fedora 27 but not Fedora 28, so we will need to be sure to model it in such a way that allows that information to be encoded.

Release engineering has a script they use to update PDC with information about which packages are critical path. We will need to work with them to ensure that their script is able to update Bodhi's critical path packages.

cverna:
@vismay-golwala 
I think we should add to the `Release` db model a new attribute that will list the critical path packages for this update. To do that you should look at the [Many-to-Many](https://docs.sqlalchemy.org/en/latest/orm/basic_relationships.html#many-to-many) relationship in sqlalchemy. 
Since we want a relationship between `Release` and `Packages` and one `Release` can have many critpath `Packages` and one `Packages` can be critpath to many `Releases`

ghost:
Should the critical path packages for a release be displayed in the server view (https://bodhi.fedoraproject.org/releases/F27)?

If yes, what details should be included for the package?

ghost:
Also, is it a good idea to rename the ```--username``` option to ```--user``` for ```bodhi releases``` command? As all other commands (```bodhi updates```, ```bodhi overrides```) use ```--user```.

https://github.com/fedora-infra/bodhi/blob/develop/bodhi/client/__init__.py#L112

bowlofeggs:
On 07/18/2018 12:15 PM, Vismay Golwala wrote:
> Should the critical path packages for a release be displayed in the
> server view (https://bodhi.fedoraproject.org/releases/F27)?

For the purposes of this issue I don't think we need to make the web UI
show critical path packages. If someone really wants that they can add
it later in a separate patch, but I expect that nobody will want it.

bowlofeggs:
On 07/18/2018 12:17 PM, Vismay Golwala wrote:
> Also, is it a good idea to rename the |--username| option to |--user|
> for |bodhi releases| command? As all other commands (|bodhi updates|,
> |bodhi overrides|) use |--user|.

That would be a good thing to do, but I have two thoughts:

[...]

PR #2466, opened 2018-07-02
Add a critpath field to packages and allow filtering packages using it

pypingou:


pypingou:
We'll need to have a way to update this boolean, but I figured we could do this later :)

cverna:
@pypingou, have a look at https://github.com/fedora-infra/bodhi/issues/2433. I don't think that having the boolean on the Packages model will work since a package can be critpath = True for F27 but critpath = Flase for F28.

cc'ing @vismay-golwala since he started to look at this too :)

pypingou:
Hm fair point @cverna and reading that ticket, I'm pondering a little. The many-to-many relation is tempting but both infos (package_id and release_id) are also on the builds table. So I'm wondering if we just want to add a new table or if we should refactor the builds table as well to relate to that new table (tbh, I'm not convinced by this idea)

Anyway, if @vismay-golwala is working on it, I'll back out :)

pypingou:
(waiting for the tests to report, before I close this PR :))

ghost:
@pypingou ```builds``` table has foreign keys (```release_id``` and ```package_id```). What might be the potential issue if we refactor only ```releases``` and ```packages```?

pypingou:
My ponder wasn't about refactorring `releases` and `packages` but potentially refactorring `builds` since it contains foreign_keys to `releases` and `packages` which the new table would also.

PR #2467, opened 2018-07-03
Add a --debug argument and some logging when retrieving CI status

pypingou:


bodhi/client/bindings.py
```diff
@@ -172,6 +174,12 @@ def __init__(self, base_url=BASE_URL, username=None, password=None, staging=Fals
         if base_url[-1] != '/':
             base_url = base_url + '/'
 
+        if debug:
+            ch = logging.StreamHandler()
+            ch.setLevel(logging.DEBUG)
+            log.addHandler(ch)
+            log.setLevel(logging.DEBUG)
```
bowlofeggs: libraries should not set up logging handlers or set log levels. This library is not just used by the bodhi cli and so it should be unopinionated about logging. It's OK for the lib to log things at debug level, but the logger should not be configured by it.

bowlofeggs (changes requested):

bodhi/client/bindings.py
pypingou: Any preference where we could configure it?
I thought this was a good place since it was common with all the actions performed by the CLI, but I'm fine w/ changing this :)

bowlofeggs:
On 07/25/2018 11:03 AM, Pierre-Yves Chibon wrote:
> Any preference where we could configure it?

We should do it in bodhi/client/__init__.py. One way I can think of to
do it as a common action so you don't have to repeat code is you could
set up a validator on the --debug flag that does this if --debug is set.
There's already a validator in there on at least one flag as an example.

pypingou:
> One way I can think of to do it as a common action so you don't have to repeat code is you could set up a validator on the --debug flag that does this if --debug is set. There's already a validator in there on at least one flag as an example.

Is the validator you have in mind the ``callback`` used for the ``--url`` argument? If not then I need to look further for this validator as I think I didn't find it then :s

I took a little bit of a different approach in the mean while, I've added the ``--debug`` argument to the ``cli()`` method, so all actions have it, but it needs to be specified at the beginning of the argument list, for example:
``bodhi --debug updates waive  python-arrow-0.8.0-5.fc28 --show --staging``
works while:
``bodhi updates waive  python-arrow-0.8.0-5.fc28 --show --staging --debug`` does not.

Again, if you prefer the validator approach I'll look into it some more :)

bowlofeggs:
On 07/26/2018 07:23 AM, Pierre-Yves Chibon wrote:
> Is the validator you have in mind the |callback| used for the |--url|
> argument? If not then I need to look further for this validator as I
> think I didn't find it then :s

Yeah it's the callback system.

> I took a little bit of a different approach in the mean while, I've
> added the |--debug| argument to the |cli()| method, so all actions have
> it, but it needs to be specified at the beginning of the argument list,
> for example:
> |bodhi --debug updates waive python-arrow-0.8.0-5.fc28 --show --staging|
> works while:
> |bodhi updates waive python-arrow-0.8.0-5.fc28 --show --staging --debug|
> does not.
> 
> Again, if you prefer the validator approach I'll look into it some more :)

I think I prefer the validator approach since it will be more consistent
with how the rest of Bodhi's flags work, unless you can find a way to
make it work both ways (I don't know a way to do it off the top of my
head, but it might be possible.)

pypingou:
I've pushed a commit porting ``--debug`` to use a callback mechanism as ``--url`` does. You will see it requires more changes than the ``cli()`` approach. If you're fine with that approach, I'll merge the third commit into the first one to clean up the history :)

pypingou:
Rebased to fix the conflict

pypingou:
--hm these tests are definitively passing locally :(--

Nevermind the last comment from jenkins confused me, they are also passing on jenkins now it seems :)

bodhi/client/__init__.py
```diff
@@ -453,7 +488,9 @@ def query(url, mine=False, rows=None, **kwargs):
         mine (Boolean): If the --mine flag was set
         kwargs (dict): Other keyword arguments passed to us by click.
     """
+    kwargs.pop('debug')
     client = bindings.BodhiClient(base_url=url, staging=kwargs['staging'])
+
```
bowlofeggs: This is an unrelated style change, so let's leave it as it is.

bodhi/client/__init__.py
```diff
@@ -295,7 +327,7 @@ def new(user, password, url, **kwargs):
                        True.
         kwargs (dict): Other keyword arguments passed to us by click.
     """
-
+    kwargs.pop('debug')
```
bowlofeggs: Is this what you meant by being a larger change? If so, a cleaner way to do this is to add ```debug``` as a parameter to each of these functions so it doesn't end up in ```kwargs```.

bowlofeggs (changes requested):

bodhi/client/__init__.py
pypingou: Good idea, will fix

bodhi/client/__init__.py
pypingou: ok

bodhi/client/__init__.py
```diff
@@ -244,7 +274,8 @@ def composes():
 @staging_option
 @click.option('-v', '--verbose', is_flag=True, default=False, help='Display more information.')
 @url_option
-def list_composes(url, staging, verbose):
+@debug_option
+def list_composes(url, staging, verbose, debug):
```
bowlofeggs, line 279: Note that we still document parameters in the "developer docs" section below, here and elsewhere in the file.

bodhi/client/__init__.py
```diff
@@ -295,7 +327,6 @@ def new(user, password, url, **kwargs):
                        True.
         kwargs (dict): Other keyword arguments passed to us by click.
     """
-
```
bowlofeggs: This is an unrelated style change.

bodhi/client/__init__.py
```diff
@@ -432,9 +464,10 @@ def edit(user, password, url, **kwargs):
 @click.option('--mine', is_flag=True, help='Show only your updates')
 @staging_option
 @url_option
+@debug_option
 @add_options(pagination_options)
 @handle_errors
-def query(url, mine=False, rows=None, **kwargs):
+def query(url, mine=False, rows=None, debug=False, **kwargs):
```
bowlofeggs: Let's not define the default for ```debug``` in two places. If you move this to be after ```url```, you don't need to give it a default.

bodhi/tests/client/test___init__.py
```diff
@@ -324,6 +324,46 @@ def test_severity_flag(self, send_request):
         ]
         self.assertEqual(send_request.mock_calls, calls)
 
+    @mock.patch.dict(client_test_data.EXAMPLE_UPDATE_MUNCH, {'severity': 'urgent'})
+    @mock.patch.dict(os.environ, {'BODHI_URL': 'http://example.com/tests/'})
+    @mock.patch('bodhi.client.bindings.BodhiClient.csrf',
+                mock.MagicMock(return_value='a_csrf_token'))
+    @mock.patch('bodhi.client.bindings.BodhiClient.send_request',
+                return_value=client_test_data.EXAMPLE_UPDATE_MUNCH, autospec=True)
+    def test_severity_and_debug_flags(self, send_request):
```
bowlofeggs: This test is really about testing the debug flag, so let's name it ```test_debug_flag``` for simplicity.

bowlofeggs (changes requested):
Just a few more minor changes.

bodhi/client/__init__.py
pypingou: Will add

bodhi/client/__init__.py
pypingou: will remove

bodhi/client/__init__.py
pypingou: Will fix

bodhi/tests/client/test___init__.py
pypingou: Will adjust

pypingou (commented):
thanks for the review

pypingou:
I think I got all of your comments, I also merged the new approach into the old one making the history cleaner.

bowlofeggs:
I forgot to mention when reviewing this one - can you also squash your
commits into 1? Thanks!

bodhi/client/__init__.py
```diff
@@ -53,13 +57,37 @@ def _warn_if_url_and_staging_set(ctx, param, value):
...
+    """
+    Set up the logging level to "debug".
+
+    This allows us to print more information on the user's screen and thus helps following what is
+    going on.
+
+    Args:
+        ctx (click.core.Context): The Click context. Unused.
+        param (click.core.Option): The option being handled. Unused.
+        value (unicode): The value of the --debug flag.
```
bowlofeggs: This is a ```bool```, not a ```unicode```.

bodhi/client/__init__.py
```diff
@@ -53,13 +57,37 @@ def _warn_if_url_and_staging_set(ctx, param, value):
...
+    Returns:
+        unicode: The value of the --debug flag.
```
bowlofeggs: This is also a ```bool```.

bodhi/client/__init__.py
```diff
@@ -53,13 +57,37 @@ def _warn_if_url_and_staging_set(ctx, param, value):
...
+    return value
+
+
 url_option = click.option('--url', envvar='BODHI_URL', default=bindings.BASE_URL,
                           help=('URL of a Bodhi server. Ignored if --staging is set. Can be set '
                                 'with BODHI_URL environment variable'),
                           callback=_warn_if_url_and_staging_set)
 staging_option = click.option('--staging', help='Use the staging bodhi instance',
                               is_flag=True, default=False)
-
```
bowlofeggs: Unrelated style change.

bowlofeggs (changes requested):
Just 3 small things I'd like to fix, then we'll be good to merge.

bowlofeggs:
I will fix these three things myself.

bowlofeggs (approved):

pypingou:
Thanks @bowlofeggs :)

bowlofeggs:

bowlofeggs:
These patches are available in Copr:

https://copr.fedorainfracloud.org/coprs/bowlofeggs/bodhi-pre-release/

PR #2468, opened 2018-07-04
Implement the possibility to waive missing requirements via bodhi-cli

pypingou:
This commit basically implements the command: bodhi update waive, see
--help for the arguments.
It queries greenwave via bodhi to know the list of unsatisfied
requirements and uses the waiving endpoint in bodhi's API to waive either
all or just some of them.

### DIFF
--- a/bodhi/client/__init__.py
+++ b/bodhi/client/__init__.py
@@ -635,6 +635,79 @@ def _get_notes(**kwargs):
         return kwargs['notes']
 
 
+@updates.command()
+@click.argument('update')
+@click.argument('comment', required=False)
+@click.option(
+    '--show', is_flag=True, default=None,
+    help="List all the required unsatisfied requirements")
+@click.option(
+    '--test', multiple=True,
+    help="Waive the specifiy test(s), to automatically waive all unsatisfied "
+    "requirements, specify --test=all")
+@staging_option
+@url_option
+@handle_errors
+def waive(update, show, test, comment, url, **kwargs):
+    # User Docs that show in the --help
+    """
+    Show or waive unsatified requirements (ie: missing or failing tests) on an existing update.
+
+    UPDATE: The title of the update (e.g. FEDORA-2017-f8e0ef2850)
+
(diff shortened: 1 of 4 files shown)


bodhi/client/__init__.py
```diff
@@ -621,6 +621,79 @@ def _get_notes(**kwargs):
...
+                       True.
+        kwargs (dict): Other keyword arguments passed to us by click.
+    """
+    client = bindings.BodhiClient(base_url=url, staging=kwargs['staging'])
+
+    if show and test:
+        click.echo(
+            'You can not list the unsatisfied requirements and waive them at the same '
+            'time, please use either --show or --test=... but not both.')
+        return
```
cverna: other command are using `sys.exit(1)` in that case, we could try to stay consistent

bodhi/client/__init__.py
```diff
@@ -621,6 +621,79 @@ def _get_notes(**kwargs):
...
+            if test_status.decision.unsatisfied_requirements:
+                click.echo('Missing tests:')
+                for req in test_status.decision.unsatisfied_requirements:
+                    click.echo('  - %s' % req)
+            else:
+                click.echo('Missing tests: None')
+    else:
+        if not comment:
+            click.echo('Comment are mandatory when waiving unsatisfied requirements')
+            return
```
cverna: same here, we could use `sys.exit(1)` and add the `ERROR:` in front of the message

bodhi/client/__init__.py
cverna: also it seems that `ERROR:` is added at the beginning of the message just like here https://github.com/fedora-infra/bodhi/blob/develop/bodhi/client/__init__.py#L576

pypingou: I'll adjust

bodhi/client/__init__.py
pypingou: Will do :)

cverna:
:+1:

bowlofeggs:

bowlofeggs:
This patch was released with 3.9.0 last week.

PR #2469, opened 2018-07-05
Moved hardcoded templates in mail.py to filesystem.

### LINKED ISSUE #2396: Bodhi crashes while composing a modular repo if it has a mailing list defined
Bodhi was [recently configured to have a mailing list configured for ```fedora_modular```](https://pagure.io/fedora-infrastructure/issue/6872). The next time a modular repo was composed, it failed with this traceback:

```
  File "/usr/lib/python2.7/site-packages/bodhi/server/consumers/masher.py", line 337, in run
    self.work()
  File "/usr/lib/python2.7/site-packages/bodhi/server/consumers/masher.py", line 422, in work
    self.send_stable_announcements()
  File "/usr/lib/python2.7/site-packages/bodhi/server/consumers/masher.py", line 70, in wrapper
    retval = method(self, *args, **kwargs)
  File "/usr/lib/python2.7/site-packages/bodhi/server/consumers/masher.py", line 732, in send_stable_announcements
    update.send_update_notice()
  File "/usr/lib/python2.7/site-packages/bodhi/server/models.py", line 2777, in send_update_notice
    for subject, body in mail.get_template(self, templatetype):
  File "/usr/lib/python2.7/site-packages/bodhi/server/mail.py", line 354, in get_template
    use_template = globals()[use_template]
KeyError: u'fedora_modular_errata_template'
```

This is because Bodhi's e-mail system hard codes releases into the Python code rather than using settings, and the modular repo isn't hard coded :/

ghost:
fixes #2396


bodhi/server/models.py
```diff
@@ -839,6 +839,7 @@ class Release(Base):
     pending_testing_tag = Column(UnicodeText, nullable=False)
     pending_stable_tag = Column(UnicodeText, nullable=False)
     override_tag = Column(UnicodeText, nullable=False)
+    template = Column(UnicodeText, default=u'default_template', nullable=True)
```
cverna: Since the `default_template.tpl` is the same as `fedora_errata_template.tpl`, we could use directly `fedora_errata_template.tpl` that way we don't need to have the `default_template.tpl` file

bodhi/server/mail.py
```diff
@@ -339,6 +340,34 @@
...
+    """
+    Read template text from file.
+
+    Args:
+        name (basestring): The name of the email template stored in 'release' table in database.
+    Returns:
+        basestring: The text read from the file.
+    """
+    location = config.get('mail.templates_basepath')
+    module, final = location.split(':')
```
cverna: Since the following code is reused from errors.py, we could make it a function and call this function in errors.py, here and in the tests.

The function could be defined in util.py for example

bodhi/server/migrations/versions/abc52bce138c_added_field_to_store_email_template_.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+# revision identifiers, used by Alembic.
+revision = 'abc52bce138c'
+down_revision = '39132c56a47e'
+
+
+def upgrade():
+    """Add a template field to the releases table."""
+    op.add_column('releases', sa.Column('template', sa.UnicodeText(), nullable=True))
```
cverna: You are creating the template column here but you do not set any value to it. It is also not used in the code in order to select the correct template for a release.

You should probably change the logic around [here](https://github.com/fedora-infra/bodhi/blob/develop/bodhi/server/models.py#L2793) to use the value stored in `template`.

Also I think `template` is a little bit too generic, let call it  `mail_template` so we know the purpose of this template :)

cverna (changes requested):

bodhi/server/migrations/versions/abc52bce138c_added_field_to_store_email_template_.py
ghost: Do we have to write SQLAlchemy ```UPDATE``` statements inside migration ```upgrade()```? Or is there another way we can write statements to update ```mail_template``` value of ```releases``` table?

cverna: Yes you should use the `upgrade` function to add these statements. So when the new column is added it will be populated with the correct information too

bodhi/server/mail.py
```diff
@@ -241,102 +242,31 @@
...
+
+    if os.path.exists(template_path):
+        try:
+            with open(template_path) as template_file:
+                return to_unicode(template_file.read())
+        except IOError as e:
+            log.error("Unable to read template file: %s" % (template_path))
+            log.error("IO Error[%s]: %s" % (e.errno, e.strerror))
+    else:
+        log.error("Invalid template name: %s." % (name))
```
cverna: I wondering if we really need to log this error , I think the `Path does not exist: ...` is enough

cverna (commented):
Just one little comment here, I think we will also need to make it possible to set the `mail_template` when creating a new release using the cli.
You can do that in this PR or do it in another PR once this one is merged.

ghost:
Let's do it in another PR. That way commits remain logically distinguishable.

cverna (approved):

PR #2472, opened 2018-07-09
Return appropriate error when 'notes' are not provided.

### LINKED ISSUE #2214: The CLI gives unhelpful error messages if --notes are not given

ghost:
fixes #2214

### DIFF
--- a/bodhi/client/__init__.py
+++ b/bodhi/client/__init__.py
@@ -305,6 +305,10 @@ def new(user, password, url, **kwargs):
 
     kwargs['notes'] = _get_notes(**kwargs)
 
+    if not kwargs['notes']:
+        click.echo("ERROR: must specify at least one of --notes, --notes-file")
+        sys.exit(1)
+
     for update in updates:
         try:
             resp = client.save(**update)
@@ -363,6 +367,10 @@ def edit(user, password, url, **kwargs):
                                   staging=kwargs['staging'])
 
     kwargs['notes'] = _get_notes(**kwargs)
+
+    if not kwargs['notes']:
+        click.echo("ERROR: must specify at least one of --notes, --notes-file")
+        sys.exit(1)
 
     try:
         if re.search(bindings.UPDATE_ID_RE, kwargs['update']):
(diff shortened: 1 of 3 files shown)


bodhi/client/__init__.py
```diff
@@ -305,6 +305,10 @@ def new(user, password, url, **kwargs):
 
     kwargs['notes'] = _get_notes(**kwargs)
 
+    if not kwargs['notes']:
```
cverna, line 308: we will need to check that also in the `edit` command

bodhi/tests/client/test___init__.py
```diff
@@ -299,7 +299,7 @@ def test_severity_flag(self, send_request):
         result = runner.invoke(
             client.new,
             ['--user', 'bowlofeggs', '--password', 's3kr3t', '--autokarma', 'bodhi-2.2.4-1.el7',
-             '--severity', 'urgent'])
+             '--severity', 'urgent', '--notes', 'No description.'])
```
cverna, line 302: I think we should test the command without --notes or --notes-file to make sure that we are getting the error message

cverna (changes requested):

bodhi/server/schemas.py
```diff
@@ -213,6 +213,7 @@ class SaveUpdateSchema(CSRFProtectedSchema, colander.MappingSchema):
     notes = colander.SchemaNode(
         colander.String(),
         validator=colander.Length(min=2),
+        missing_msg='Required description for the update.'
```
cverna: Maybe we can change this to `A description is required for the update`. I find it easier to understand as an error message

bodhi/tests/client/test___init__.py
ghost: Thanks for pointing it out. Added a separate test for that!

cverna (approved):
LGTM

PR #2473, opened 2018-07-10
Modified RSS feed titles.

### LINKED ISSUE #1119: rss-feeds should have more speaking titles
eg.: https://bodhi.fedoraproject.org/rss/updates/?releases=F24
the feed is named "updates"
maybe it should be renamed to "Fedora XX Released Updates" where XX would be 24,...

ghost:
fixes #1119

### DIFF
--- a/bodhi/tests/server/services/test_updates.py
+++ b/bodhi/tests/server/services/test_updates.py
@@ -1465,6 +1465,23 @@ class TestUpdatesService(BaseTestCase):
                            headers={'Accept': 'application/atom+xml'})
         self.assertIn('application/rss+xml', res.headers['Content-Type'])
         self.assertIn('bodhi-2.0-1.fc17', res)
+        self.assertIn('Released updates', res)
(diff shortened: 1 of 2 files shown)


bodhi/tests/server/services/test_updates.py
```diff
@@ -1466,6 +1466,12 @@ def test_list_updates_rss(self):
         self.assertIn('application/rss+xml', res.headers['Content-Type'])
         self.assertIn('bodhi-2.0-1.fc17', res)
 
+    def test_list_updates_rss_with_filter(self):
+        res = self.app.get('/rss/updates/', {'severity': 'low'},
+                           headers={'Accept': 'application/atom+xml'})
+        self.assertIn('application/rss+xml', res.headers['Content-Type'])
+        self.assertIn('Filtered on: severity(low)', res)
+
```
cverna: Maybe we should have a test where we do not filter, and one where we have more than one key

cverna (approved):

PR #2476, opened 2018-07-12
Added mapping for critical packages for a release.

### LINKED ISSUE #2433: Bodhi should know which packages are critical path without outside help

ghost:
fixes #2433


bodhi/server/models.py
```diff
@@ -786,6 +786,12 @@ class ComposeState(DeclEnum):
     Column('user_id', Integer, ForeignKey('users.id')),
     Column('package_id', Integer, ForeignKey('packages.id')))
 
+package_release_table = Table(
+    'package_release_table', Base.metadata,
+    Column('package_id', Integer, ForeignKey('packages.id')),
+    Column('release_id', Integer, ForeignKey('releases.id'))
```
bowlofeggs: Let's set ```nullable=False``` on these.

ghost:
jenkies test

bowlofeggs:
The second commit in this pull request is not related to the first commit. Please remove it from this pull request and open a new one so we can evaluate them independently.

bodhi/client/bindings.py
```diff
@@ -855,6 +856,66 @@ def get_releases(self, **kwargs):
...
+            kwargs (dict): The parameters to send with the request.
+        """
+        res = self.send_request('releases/', verb='POST', auth=True,
+                                data=kwargs)
+
+        if 'errors' in res:
+            self.print_errors(res)
+
+        else:
+            print("Saved release:")
```
bowlofeggs: The bindings should not assume they have stdout/stderr because they are intended for general use (and Bodhi server uses them too).

bodhi/client/bindings.py
```diff
@@ -855,6 +856,66 @@ def get_releases(self, **kwargs):
...
+            self.print_release(res)
+
+        return res
+
+    def print_release(self, release, verbose=False):
```
bowlofeggs: Why remove this function from the client and move it into the bindings? The bindings are intended to be a general purpose Python client for programmatic use, not human use, so they shouldn't be printing things to the terminal (because there might not always be a terminal).

bodhi/client/bindings.py
```diff
@@ -855,6 +856,66 @@ def get_releases(self, **kwargs):
...
+        """
+        Print errors to the terminal and exit with code 1.
+
+        Args:
+            errors (munch.Munch): The errors to be formatted and printed.
+        """
+        for error in data['errors']:
+            print("ERROR: %s" % error['description'])
+
+        sys.exit(1)
```
bowlofeggs: The bindings should never call ```exit()``` because they can be used by long-running services and we don't want to exit on their behalf.

Also, it is surprising for a function called "print" to exit.

bodhi/server/models.py
```diff
@@ -786,6 +786,12 @@ class ComposeState(DeclEnum):
     Column('user_id', Integer, ForeignKey('users.id')),
     Column('package_id', Integer, ForeignKey('packages.id')))
 
+package_release_table = Table(
+    'package_release_table', Base.metadata,
+    Column('package_id', Integer, ForeignKey('packages.id'), nullable=False),
+    Column('release_id', Integer, ForeignKey('releases.id'), nullable=False)
+)
```
bowlofeggs: Let's give this table a more descriptive name, because it doesn't sound like it is related to critical path packages. How about ```critical_path_packages_table```?

bodhi/server/models.py
```diff
@@ -951,6 +957,35 @@ def from_tags(cls, tags, session):
             if release:
                 return release
 
+    def add_critpath_pkgs(self, packages, session):
+        """
+        Add the critcal path packages for a Release given a list of packages name.
+
+        Args:
+            packages (list): A list of packages that are critical path for an associated release.
+            session (sqlalchemy.orm.session.Session): A database session.
```
bowlofeggs: We don't need to pass the session around anymore so let's drop this argument.

bodhi/server/models.py
```diff
@@ -951,6 +957,35 @@ def from_tags(cls, tags, session):
...
+        """
+        if packages:
+            for pkg in packages:
+                pkg_object = session.query(Package).filter_by(name=pkg).first()
```
bowlofeggs: You can do this same query without session like this:

```
Package.query.filter_by(…
```

bodhi/server/models.py
```diff
@@ -951,6 +957,35 @@ def from_tags(cls, tags, session):
...
+                    self.critpath_pkgs.append(pkg_object)
+
+    def remove_critpath_pkgs(self, packages, session):
+        """
+        Remove the critcal path packages for a Release given a list of packages name.
+
+        Args:
+            packages (list): A list of packages that are not critical path anymore
+                for an associated release.
+            session (sqlalchemy.orm.session.Session): A database session.
```
bowlofeggs: Let's drop the session here too.

bodhi/server/models.py
```diff
@@ -951,6 +957,35 @@ def from_tags(cls, tags, session):
             if release:
                 return release
 
+    def add_critpath_pkgs(self, packages, session):
+        """
+        Add the critcal path packages for a Release given a list of packages name.
```
bowlofeggs: s/packages name/package names/

bodhi/server/models.py
```diff
@@ -951,6 +957,35 @@ def from_tags(cls, tags, session):
...
+        """
+        if packages:
+            for pkg in packages:
+                pkg_object = session.query(Package).filter_by(name=pkg).first()
+                if pkg_object is not None and pkg_object not in self.critpath_pkgs:
+                    self.critpath_pkgs.append(pkg_object)
+
+    def remove_critpath_pkgs(self, packages, session):
+        """
+        Remove the critcal path packages for a Release given a list of packages name.
```
bowlofeggs: s/packages name/package names/

bodhi/server/services/releases.py
```diff
@@ -56,6 +57,9 @@
                    # Note, this 'rw' is not a typo.  the @comments service has
                    # a ``post`` section at the bottom.
                    cors_origins=bodhi.server.security.cors_origins_rw)
+critical_packages = Service(name='critical_packages', path='/releases/{name}/critical_packages',
```
bowlofeggs: s/critical_packages/critpath_packages/ (or critical_path_packages).

bodhi/server/services/releases.py
```diff
@@ -327,3 +346,55 @@ def save_release(request):
     request.db.flush()
 
     return r
+
+
+@critical_packages.put(schema=bodhi.server.schemas.UpdateCriticalPackages,
+                       renderer='json',
+                       error_handler=bodhi.server.services.errors.json_handler,
+                       validators=(colander_body_validator, validate_tags, validate_enums)
+                       )
```
bowlofeggs: We need to be sure to use the admin ACL factory here so that only admins can edit the critpath packages.

bodhi/server/services/releases.py
```diff
@@ -327,3 +346,55 @@ def save_release(request):
...
+    """
+    Add a list of critical path packages to a specific release.
+
+    Args:
+        request (pyramid.request): The current request.
+    Returns:
+        dict: The dictionary {'status': 'success'}.
+    """
+    data = request.validated
+    pkg_list = data.pop('packages', None)
```
bowlofeggs: We should use a validator to ensure that the pkg_list is already Package objects and not package names.

bodhi/server/services/releases.py
```diff
@@ -327,3 +346,55 @@ def save_release(request):
...
+
+    id = request.matchdict.get('name')
+    release = Release.get(id)
```
bowlofeggs: We should use a validator to turn this into a Release object before we enter this function.

bodhi/server/services/releases.py
```diff
@@ -327,3 +346,55 @@ def save_release(request):
...
+    release.add_critpath_pkgs(pkg_list, session)
+
+    return dict(status=u'success', packages=release.critpath_pkgs)
+
+
+@critical_packages.delete(schema=bodhi.server.schemas.UpdateCriticalPackages,
+                          permission='edit', renderer='json',
+                          error_handler=bodhi.server.services.errors.json_handler,
+                          validators=(colander_body_validator, validate_tags, validate_enums)
+                          )
```
bowlofeggs: This one needs to use the admin ACLs too so ordinary users cannot edit the list.

bodhi/server/services/releases.py
```diff
@@ -327,3 +346,55 @@ def save_release(request):
...
 
     return r
+
+
+@critical_packages.put(schema=bodhi.server.schemas.UpdateCriticalPackages,
+                       renderer='json',
+                       error_handler=bodhi.server.services.errors.json_handler,
+                       validators=(colander_body_validator, validate_tags, validate_enums)
+                       )
+def add_packages(request):
```
bowlofeggs: Let's call this ```add_critpath_packages``` or similar so it's clearer what it does.

bodhi/server/services/releases.py
```diff
@@ -327,3 +346,55 @@ def save_release(request):
...
+
+    return dict(status=u'success', packages=release.critpath_pkgs)
+
+
+@critical_packages.delete(schema=bodhi.server.schemas.UpdateCriticalPackages,
+                          permission='edit', renderer='json',
+                          error_handler=bodhi.server.services.errors.json_handler,
+                          validators=(colander_body_validator, validate_tags, validate_enums)
+                          )
+def remove_packages(request):
```
bowlofeggs: Let's call this ```remove_critpath_packages``` or similar.

bodhi/server/services/releases.py
```diff
@@ -327,3 +346,55 @@ def save_release(request):
...
+    Args:
+        request (pyramid.request): The current request.
+    Returns:
+        dict: The dictionary {'status': 'success'}.
+    """
+    data = request.validated
+    pkg_list = data.pop('packages', None)
+
+    id = request.matchdict.get('name')
+    release = Release.get(id)
```
bowlofeggs: Similarly here, we should use validators to ensure that pkg_list is already Package objects and release should already be validated as an object as well.

bodhi/tests/client/test___init__.py
```diff
@@ -1475,7 +1475,7 @@ def test_url_flag(self, send_request):
         result = runner.invoke(
             client.create_release,
             ['--name', 'F27', '--url', 'http://localhost:6543', '--username', 'bowlofeggs',
-             '--password', 's3kr3t'])
+             '--password', 's3kr3t', '--critpath-pkgs', 'python,perl,php'])
```
bowlofeggs: This test is here to test proper operation of the ```--url``` flag, so let's not complicate it by inserting this flag.

bodhi/tests/client/test___init__.py
```diff
@@ -1618,6 +1618,41 @@ def test_edit_with_errors(self, send_request):
...
+    def test_edit_critpath_pkgs(self, send_request):
+        """
+        Assert correct behavior while editing critpath_pkgs.
+        """
+        runner = testing.CliRunner()
+
+        result = runner.invoke(
+            client.edit_release,
+            ['--name', 'F27', '--critpath-pkgs', 'python,perl,php', '--url',
+             'http://localhost:6543', '--username', 'bowlofeggs', '--password', 's3kr3t'])
```
bowlofeggs: You shouldn't need the ```--url```, ```--username```, or ```--password``` flags here, and removing them makes the test simpler and more targeted.

bodhi/tests/client/test___init__.py
```diff
@@ -1659,6 +1694,26 @@ def test_info_with_errors(self, send_request):
...
+        Assert correct behavior with the --verbose flag.
+        """
+        runner = testing.CliRunner()
+
+        result = runner.invoke(client.info_release, ['--url', 'http://localhost:6543', 'F27',
+                                                     '--verbose'])
+
+        self.assertEqual(result.exit_code, 0)
+        self.assertEqual(result.output,
+                         client_test_data.EXPECTED_INFO_RELEASE_OUTPUT)
```
bowlofeggs: Since ```EXAMPLE_INFO_RELEASE_MUNCH``` does not have critpath packages in it, this test does not assert that they are rendered in the output.

bodhi/tests/server/services/test_releases.py
```diff
@@ -251,6 +261,43 @@ def test_edit_release(self):
...
+        r = res.json_body
+
+        r["edited"] = name
+        r["csrf_token"] = self.get_csrf_token()
+        r["critpath_pkgs"] = "bodhi"
+
+        res = self.app.post("/releases/", r, status=200)
+
+        r = self.db.query(Release).filter(Release.name == name).one()
+        self.assertEquals(r.critpath_pkgs[0].name, u'bodhi')
```
bowlofeggs: Let's assert that Bodhi is the only critpath package too:

```
self.assertEqual([p.name for p in r.critpath_pkgs], ['bodhi'])
```

Note also that ```assertEquals``` is deprecated in favor of ```assertEqual```.

bodhi/tests/server/services/test_releases.py
```diff
@@ -251,6 +261,43 @@ def test_edit_release(self):
...
+        self.assertEquals(r.critpath_pkgs[0].critpath_releases[0].name, u'F22')
```
bowlofeggs: Similarly here let's assert that the list is only F22.

bodhi/tests/server/services/test_releases.py
```diff
@@ -251,6 +261,43 @@ def test_edit_release(self):
...
+        r = res.json_body
+
+        r["edited"] = name
+        r["csrf_token"] = self.get_csrf_token()
+        r["critpath_pkgs"] = "bodhi,bodhi,bodhi"
+
+        res = self.app.post("/releases/", r, status=200)
+
+        r = self.db.query(Release).filter(Release.name == name).one()
+        self.assertEquals(len(r.critpath_pkgs), 1)
```
bowlofeggs: We can use the stronger assertion I wrote above here to assert it is 1 package and that it is the correct one.

bodhi/tests/server/services/test_releases.py
```diff
@@ -299,3 +346,32 @@ def test_query_releases_html_two_releases_same_state(self):
...
+                     headers={'Accept': 'application/json'})
+
+        release = self.db.query(Release).filter(Release.name == 'F22').one()
+        pkg_list = [pkg.name for pkg in release.critpath_pkgs]
+
+        self.assertEquals(len(release['critpath_pkgs']), 4)
+        self.assertIn('python', pkg_list)
+        self.assertIn('json', pkg_list)
+        self.assertIn('nodejs', pkg_list)
+        self.assertIn('git', pkg_list)
```
bowlofeggs: We also need a test that ```PUT``` works with releases that already had packages.

bowlofeggs: You could just assert that the set of packages is equal rather than 4 separate assertions.

bodhi/tests/server/services/test_releases.py
```diff
@@ -299,3 +346,32 @@ def test_query_releases_html_two_releases_same_state(self):
...
+                        headers={'Accept': 'application/json', 'Content-Type': 'application/json'})
+
+        release = self.db.query(Release).filter(Release.name == 'F22').one()
+        pkg_list = [pkg.name for pkg in release.critpath_pkgs]
+
+        self.assertEquals(len(release['critpath_pkgs']), 2)
+        self.assertIn('python', pkg_list)
+        self.assertIn('nodejs', pkg_list)
+        self.assertNotIn('json', pkg_list)
+        self.assertNotIn('git', pkg_list)
```
bowlofeggs: Similarly here, a set assertion would be simpler.

bowlofeggs (changes requested):

bodhi/client/bindings.py
bowlofeggs: Note also that we should use ```click.echo()``` instead of ```print()```.

bowlofeggs:
One more thought - this PR provides two ways for API users to set critical path packages. I believe in the [Zen of Python](https://www.python.org/dev/peps/pep-0020/), which states "There should be one-- and preferably only one --obvious way to do it."

bodhi/tests/client/test___init__.py
ghost: I think it does as I added a new munch [here](https://github.com/fedora-infra/bodhi/pull/2476/files#diff-5df7c10498c4e9e4b99b37891f121ad4R902).

bodhi/tests/client/test___init__.py
ghost: Wouldn't we need ```--username``` and ```--password``` for authentication?
Test goes into forever wait if I don't provide those. (waiting on authentication maybe)

ghost:
@bowlofeggs - do you suggest calling the server API while editing through CLI?

bodhi/tests/client/test___init__.py
bowlofeggs: Ah yes, I had missed that. Nevermind then.

bowlofeggs:
On 07/24/2018 04:09 PM, Vismay Golwala wrote:
> Wouldn't we need |--username| and |--password| for authentication?
> Test goes into forever wait if I don't provide those. (waiting on
> authentication maybe)

Ah yes, I forgot about that. Probably because it's prompting you and you
can't get prompts in the tests. OK, let's leave username/password but
drop url.

bowlofeggs:
On 07/24/2018 04:16 PM, Vismay Golwala wrote:
> @bowlofeggs <https://github.com/bowlofeggs> - do you suggest calling the
> server API while editing through CLI?

We do want the CLI to use the server API because the CLI isn't
guaranteed to be on the same computer as the server.

ghost:
@bowlofeggs : Regarding two ways to set critical path packages

The server API is to add_packages and remove_packages. 

However, while creating a new release or editing an existing one in CLI, user specifies the list to be set as critical path packages for that particular release. Do we want to add two separate fields as in ```--add-critpath-pkgs``` and ```--remove-critpath-pkgs```? If so, we can set the default to ```None``` while creating a ```Release```.

bowlofeggs:
On 07/30/2018 01:59 PM, Vismay Golwala wrote:
> However, while creating a new release or editing an existing one in CLI,
> user specifies the list to be set as critical path packages for that
> particular release. Do we want to add two separate fields as in
> |--add-critpath-pkgs| and |--remove-critpath-pkgs|? If so, we can set
> the default to |None| while creating a |Release|.

I don't think we want the user to have to specify the whole list every
time they want to edit it. There can be hundreds of critical path
packages, so it would be quite burdensome to require the user to specify
the whole list when they want to add or remove a package. So yeah, I
think we probably want it to default to the empty list, and we may even
want a dedicated CLI command to add/remove critical path packages.

Note also that adding this feature to the CLI is not a strict
requirement at this time - it's more of a nice to have. There is no
current CLI to do this as is, so all we are required to provide is a
Python API (via bindings.py).

bodhi/client/__init__.py
```diff
@@ -132,6 +132,8 @@ def _warn_if_url_and_staging_set(ctx, param, value):
     click.option('--state', type=click.Choice(['disabled', 'pending', 'current',
                                                'archived']),
                  help='The state of the release'),
+    click.option('--critpath-pkgs',
+                 help='Comma-separated list of critical packages for this release.'),
```
bowlofeggs: We don't want to make the user tell us the complete list of packages. There can be hundreds or even a few thousand of them, so if a user wants to add or remove one we don't want to make them pass the entire list to us or it will be burdensome to use. Instead, we should have dedicated commands to manage critical path packages with simple ```add``` and ```remove``` subcommands. Something like this:

```
$ bodhi releases critpath add F28 firefox kernel gcc…
$ bodhi releases critpath remove F28 tuxracer bzflag…
```

It would also be sensible to list the critpath packages per release like this:

```
$ bodhi releases critpath list F28
```

These should correspond to a dedicated API for these actions:

```
PUT /releases/<id>/critpath-packages
DELETE /releases/<id>/critpath-packages
GET /releases/<id>/critpath-packages
```

bodhi/client/__init__.py
```diff
@@ -941,7 +943,7 @@ def edit_release(user, password, url, **kwargs):
     edited = kwargs.pop('name')
 
     if edited is None:
-        print("ERROR: Please specify the name of the release to edit")
+        click.echo("ERROR: Please specify the name of the release to edit")
```
bowlofeggs: While this change is fine and even good, it's not related to the purpose of the rest of the pull request and so should not be made along with it. Commits should be atomic, meaning that they have a single focus in the change they make. This change should be its own commit in its own pull request so it can be considered on its own merit. The same is true for the other places like this below.

bodhi/client/__init__.py
```diff
@@ -1000,31 +1003,35 @@ def save(client, **kwargs):
...
+    click.echo("  Testing Tag:         %s" % release['testing_tag'])
+    click.echo("  Candidate Tag:       %s" % release['candidate_tag'])
+    click.echo("  Pending Signing Tag: %s" % release['pending_signing_tag'])
+    click.echo("  Pending Testing Tag: %s" % release['pending_testing_tag'])
+    click.echo("  Pending Stable Tag:  %s" % release['pending_stable_tag'])
+    click.echo("  Override Tag:        %s" % release['override_tag'])
+    click.echo("  State:               %s" % release['state'])
+
+    if verbose and 'critpath_pkgs' in release:
+        critpath_pkgs = ",".join([pkg['name'] for pkg in release['critpath_pkgs']])
```
bowlofeggs: It would be more readable if you added a space after the comma.

bodhi/server/models.py
```diff
@@ -951,6 +957,35 @@ def from_tags(cls, tags, session):
...
+        Add the critcal path packages for a Release given a list of package names.
+
+        Args:
+            packages (list): A list of packages that are critical path for an associated release.
+        """
+        if packages:
+            for pkg in packages:
+                if isinstance(pkg, six.string_types):
+                    pkg = Package.query.filter_by(name=pkg).first()
+                if pkg is not None and pkg not in self.critpath_pkgs:
```
bowlofeggs: Let's be more opinionated about the object type we will accept in the docblock above and demand that it must be a list of Package objects. Then we don't need to check if it is ```None``` and don't need to worry about whether we got a package name or not, and this method gets much simpler. Since we control the code using this function, we can ensure that it is used correctly.

bodhi/server/models.py
```diff
@@ -951,6 +957,35 @@ def from_tags(cls, tags, session):
...
+
+        Args:
+            packages (list): A list of packages that are not critical path anymore
+                for an associated release.
+        """
+        if packages:
+            for pkg in packages:
+                if isinstance(pkg, six.string_types):
+                    pkg = Package.query.filter_by(name=pkg).first()
+                if pkg is not None and pkg in self.critpath_pkgs:
```
bowlofeggs: Same here, let's just assume we get a list of Package objects.

bodhi/server/models.py
```diff
@@ -1004,6 +1039,10 @@ class to create when loading rows from the database.
     committers = relationship('User', secondary=user_package_table,
                               backref='packages')
 
+    # Many-to-many relationship
+    critpath_releases = relationship('Release', secondary=critical_path_packages_table,
+                                     backref='critpath_pkgs')
```
bowlofeggs, line 1044: Let's make sure this does not get serialized with the release, otherwise the critpath packages table will have to be queried every time a release is queried. You can do that by adding this to ```__exclude_columns__```.

bodhi/server/services/releases.py
```diff
@@ -57,6 +57,10 @@
                    # Note, this 'rw' is not a typo.  the @comments service has
                    # a ``post`` section at the bottom.
                    cors_origins=bodhi.server.security.cors_origins_rw)
+critpath_packages = Service(name='critpath_packages', path='/releases/critpath_packages',
```
bowlofeggs: Let's add the release to the path rather than getting it from PUT data. This way we can also GET critpath packages.

bodhi/server/services/releases.py
```diff
@@ -57,6 +57,10 @@
...
+                            description='Update critical path packages service',
```
bowlofeggs: The service should also be able to get them, so let's say "Retrieve and edit critical path packages."

bodhi/server/services/releases.py
```diff
@@ -322,6 +329,17 @@ def save_release(request):
             for k, v in data.items():
                 setattr(r, k, v)
 
+        # Update the critical packages of the release, if provided.
+        if new_critpath_pkgs:
```
bowlofeggs: Let's not allow two ways to edit critical path packages. The critpath_packages service gives users a way to do it, so let's drop this block of code.

bodhi/server/services/releases.py
```diff
@@ -332,3 +350,47 @@ def save_release(request):
...
+    """
+    Add a list of critical path packages to a specific release.
+
+    Args:
+        request (pyramid.request): The current request.
+    Returns:
+        dict: The dictionary {'status': 'success'}.
+    """
+    data = request.validated
+    release = data['release']
```
bowlofeggs: Let's move the release into the URL instead of the PUT data. This will allow us to add a GET method.

bodhi/server/services/releases.py
```diff
@@ -332,3 +350,47 @@ def save_release(request):
...
+
+    release.add_critpath_pkgs(data['packages'])
```
bowlofeggs: We should add a validator for packages that ensures they exist, creates them if they don't, and turns them into Package objects instead of strings.

bodhi/server/services/releases.py
```diff
@@ -332,3 +350,47 @@ def save_release(request):
...
+
+    Args:
+        request (pyramid.request): The current request.
+    Returns:
+        dict: The dictionary {'status': 'success'}.
+    """
+    data = request.validated
+    release = data['release']
+
+    release.remove_critpath_pkgs(data['packages'])
```
bowlofeggs: Similar here as above.

bodhi/server/services/releases.py
```diff
@@ -332,3 +350,47 @@ def save_release(request):
...
+    """
+    Remove a list of critical path packages from a specific release.
+
+    Args:
+        request (pyramid.request): The current request.
+    Returns:
+        dict: The dictionary {'status': 'success'}.
+    """
+    data = request.validated
+    release = data['release']
```
bowlofeggs: Similar here as above.

bodhi/tests/client/__init__.py
```diff
@@ -892,4 +892,34 @@
...
+  Dist Tag:            f27
+  Stable Tag:          f27-updates
+  Testing Tag:         f27-updates-testing
+  Candidate Tag:       f27-updates-candidate
+  Pending Signing Tag: f27-signing-pending
+  Pending Testing Tag: f27-updates-testing-pending
+  Pending Stable Tag:  f27-updates-pending
+  Override Tag:        f27-override
+  State:               pending
+  Critpath Packages:   python,kernel
```
bowlofeggs: Let's keep this as its own separate command so that the server doesn't have to fetch and serialize this data on every GET request for a release.

bodhi/tests/server/services/test_releases.py
```diff
@@ -41,6 +41,16 @@ def setUp(self):
...
         self.db.add(release)
+
+        packages = [
+            Package(name=u'python', type=ContentType.rpm),
+            Package(name=u'json', type=ContentType.rpm),
+            Package(name=u'nodejs', type=ContentType.rpm),
+            Package(name=u'git', type=ContentType.module)
+        ]
+        for pkg in packages:
+            self.db.add(pkg)
```
bowlofeggs, line 57: Let's make sure there's a test for packages that already exist, and another test for packages that do not already exist. The API should create packages that do not exist automatically.

bodhi/tests/server/services/test_releases.py
```diff
@@ -279,6 +289,52 @@ def test_edit_release(self):
...
+        """
+        name = u"F22"
+
+        res = self.app.get('/releases/%s' % name, status=200)
+        r = res.json_body
+
+        # Test adding a package.
+        r["edited"] = name
+        r["csrf_token"] = self.get_csrf_token()
+        r["critpath_pkgs"] = "bodhi"
```
bowlofeggs: We don't want there to be two ways to change this field - let's drop this method in favor of the PUT/DELETE calls.

bowlofeggs (changes requested):

bodhi/client/__init__.py
cverna: Looking at the current releng script to update the critpath packages, it is taking a txt file as input containing the list of packages. See [here](https://pagure.io/releng/blob/master/f/scripts/update-critpath#_64)

So it might be a good idea to add support for this in the bodhi cli.

bodhi/tests/server/services/test_releases.py
ghost: That will ignore invalid package names due to typing error. 

How about a flag such as ```--create```, which if passed, creates packages that don't exist. Otherwise, prevents invalid package names.

ghost:
This will not work until [this pull request](https://github.com/fedora-infra/python-fedora/pull/213) is merged and the ```python-fedora``` package is installed on Bodhi.

ghost:
jenkies test

bodhi/client/__init__.py
```diff
@@ -985,6 +985,90 @@ def info_release(name, url, **kwargs):
         print_release(res)
 
 
+@releases.group(name='critpath')
+def critpath_pkgs():
+    # Docs that show in the --help
+    """Interact with critical path packages of a release."""
+    # Developer Docs
+    """Manage the critical path packages of a release."""
```
bowlofeggs: Since these are essentially the same messages, let's just document it once:

```
def critpath_pkgs():
    """Manage critical path packages."""
```

bodhi/client/__init__.py
```diff
@@ -985,6 +985,90 @@ def info_release(name, url, **kwargs):
...
+@handle_errors
+@click.argument('name')
+@url_option
+@staging_option
+def list_critpath_packages(name, url, **kwargs):
+    """
+    List critical path packages of a release.
+
+    NAME: Release name (e.g. F27)
+    """
```
bowlofeggs, line 1060: For this one we do want to show both developer docs (argument types) and user docs (what you have here), similar to what we've done on other functions in this file.

bodhi/client/__init__.py
```diff
@@ -985,6 +985,90 @@ def info_release(name, url, **kwargs):
...
+    client = bindings.BodhiClient(base_url=url, staging=kwargs['staging'])
+
+    res = client.send_request('releases/{}/critpath-packages'.format(name), verb='GET', auth=False)
+
+    if 'errors' in res:
```
bowlofeggs: If you make a binding for getting critical path packages, you can wrap it with ```bodhi.client.bindings.errorhandled()``` and then you don't need to do this because it will do it for you. This will also help releng who will want to refactor their script to use the bindings (not the CLI).

bodhi/client/__init__.py
```diff
@@ -985,6 +985,90 @@ def info_release(name, url, **kwargs):
...
+@click.argument('packages', nargs=-1)
+@url_option
+@staging_option
+def add_critpath_packages(name, packages, url, **kwargs):
+    """
+    Add critical path packages to a release.
+
+    NAME: Release name (e.g. F27)
+
+    [PACKAGES]: List of packages to add (e.g. firefox vim python)
```
bowlofeggs: s/List/Space-separated list/

bodhi/client/__init__.py
```diff
@@ -985,6 +985,90 @@ def info_release(name, url, **kwargs):
...
+    """
+    client = bindings.BodhiClient(base_url=url, staging=kwargs['staging'])
+
+    packages_str = ','.join(packages)
+    res = client.send_request('releases/{}/critpath-packages'.format(name), verb='PUT',
+                              auth=True, data={"packages": packages_str})
```
bowlofeggs: Let's make this into a binding so that the releng script can use the bindings and not the CLI executable.

bodhi/client/__init__.py
```diff
@@ -985,6 +985,90 @@ def info_release(name, url, **kwargs):
...
+
+    if 'errors' in res:
```
bowlofeggs: If we use ```errorhandled()``` we don't need this here and can let the bindings do this for us.

bodhi/client/__init__.py
```diff
@@ -985,6 +985,90 @@ def info_release(name, url, **kwargs):
...
+@click.argument('packages', nargs=-1)
+@url_option
+@staging_option
+def remove_critpath_packages(name, packages, url, **kwargs):
+    """
+    Remove critical path packages from a release.
+
+    NAME: Release name (e.g. F27)
+
+    [PACKAGES]: List of packages to remove (e.g. firefox vim python)
```
bowlofeggs: s/List/Space-separated list/

bodhi/client/__init__.py
```diff
@@ -985,6 +985,90 @@ def info_release(name, url, **kwargs):
...
+    """
+    client = bindings.BodhiClient(base_url=url, staging=kwargs['staging'])
+
+    # DELETE request method ignores the body sometimes, so passing dictionary doesn't work
+    # However, passing it's string representation works somehow.
+    packages_str = ','.join(packages)
+    res = client.send_request('releases/{}/critpath-packages'.format(name), verb='DELETE',
+                              auth=True, data='{"packages": "%s" }' % packages_str)
```
bowlofeggs: Let's turn this into a method in the bindings too, and then we don't need to handle errors below.

bodhi/server/models.py
```diff
@@ -951,6 +957,37 @@ def from_tags(cls, tags, session):
...
                 return release
 
+    def get_critpath_pkgs(self):
+        """
+        Get a list of package names that are critcal path for this Release.
+
+        Returns:
+            packages (list): A list of packages that are critical path for this Release.
+        """
+        return self.critpath_pkgs
```
bowlofeggs: We don't really need a getter for this - it's conventional in Python to just directly use the attribute rather than a getter.

bodhi/server/services/releases.py
```diff
@@ -57,6 +58,10 @@
                    # Note, this 'rw' is not a typo.  the @comments service has
                    # a ``post`` section at the bottom.
                    cors_origins=bodhi.server.security.cors_origins_rw)
+critpath_packages = Service(name='critpath_packages', path='/releases/{name}/critpath-packages',
+                            description='Retrieve and edit critical path packages.',
+                            factory=security.AdminACLFactory,
```
bowlofeggs, line 63: This makes it so that only authorized users can ```GET``` the critpath packages:

```
[vagrant@bodhi-dev bodhi]$ curl -I http://localhost:6543/releases/F28/critpath-packages
HTTP/1.1 403 Forbidden
Content-Length: 10512
Content-Type: text/html; charset=UTF-8
Date: Wed, 15 Aug 2018 16:26:58 GMT
Server: waitress
X-Content-Type-Options: nosniff
```

I *think* it is possible to move the ```factory``` argument down to the service handlers below to require admin for write calls but allow read calls to anyone.

Another way to do it would be to make a new factory that inspects the request to determine whether admin ACLs are required. Either way is fine with me, but we do want to allow open reading of the critpath packages while requiring admin permissions to write them. As it is, ```bodhi releases critpath list F28``` does not work.

bodhi/server/schemas.py
```diff
@@ -403,6 +403,22 @@ class SaveReleaseSchema(CSRFProtectedSchema, colander.MappingSchema):
...
+    """
+    An API schema for critical path packages.
+
+    This schema is used by bodhi.server.services.releases.add_critpath_packages()
+    and bodhi.server.services.releases.remove_critpath_packages().
+    """
+
+    packages = Packages(
+        colander.Sequence(accept_scalar=True),
+        location='querystring',
```
bowlofeggs: Shouldn't the location be the body? We probably don't want to use the querystring in POST and PUT calls to send the data.

bodhi/server/schemas.py
```diff
@@ -403,6 +403,22 @@ class SaveReleaseSchema(CSRFProtectedSchema, colander.MappingSchema):
...
+        missing=None,
```
bowlofeggs: It might be possible to use ```missing=list``` here to generate an empty list. I'm not 100% sure you can use a callable there, so try it out and see what happens.

bodhi/server/services/releases.py
```diff
@@ -332,3 +337,75 @@ def save_release(request):
...
+    Get a list of critical path packages for a specific release.
+
+    Args:
+        request (pyramid.request): The current request.
+    Returns:
+        dict: The dictionary {'packages': critpath_packages}.
+    """
+    data = request.validated
+
+    release = data.pop('release', None)
```
bowlofeggs: You don't need to ```pop``` this - you can just use ```request.validated['release']``` directly below and save two lines of code.

bodhi/server/services/releases.py
```diff
@@ -332,3 +337,75 @@ def save_release(request):
     request.db.flush()
 
     return r
+
+
+@critpath_packages.get(permission='edit', renderer='json',
```
bowlofeggs: This actually might be the issue - the ```edit``` permission might be why the 403 is happening that I wrote about earlier. Unauthenticated users don't get any permissions, so you should be able to just drop that argument here.

bodhi/server/services/releases.py
```diff
@@ -332,3 +337,75 @@ def save_release(request):
...
+                       error_handler=bodhi.server.services.errors.json_handler,
+                       validators=(colander_body_validator, validate_release_name)
+                       )
```
bowlofeggs: My personal preference would be to include this on the line above it to save precious monitor real estate, but this is fine if you prefer it as long as flake8 is happy.

bodhi/server/services/releases.py
```diff
@@ -332,3 +337,75 @@ def save_release(request):
...
+
+    Args:
+        request (pyramid.request): The current request.
+    Returns:
+        dict: The dictionary {'packages': updated_critpath_packages}.
+    """
+    data = request.validated
+
+    release = data.pop('release', None)
+    packages = data.pop('packages', None)
```
bowlofeggs: Similar here, you can drop all three of the above lines and simplify things a bit.

bodhi/server/services/releases.py
```diff
@@ -332,3 +337,75 @@ def save_release(request):
...
```
bowlofeggs: Similar simplification here.

bodhi/server/validators.py
```diff
@@ -710,6 +710,22 @@ def validate_release(request, **kwargs):
...
 
+def validate_release_name(request, **kwargs):
+    """
+    Make sure the referenced release exists.
+
+    Args:
+        request (pyramid.util.Request): The current request.
+        kwargs (dict): The kwargs of the related service definition. Unused.
+    """
+    release = Release.get(request.matchdict.get('name'))
```
bowlofeggs, line 723: If you change the parameter name from ```name``` to ```release```, I think you might be able to reuse ```validate_release()``` defined below instead of creating another validator.

bodhi/tests/client/test___init__.py
```diff
@@ -1696,6 +1696,158 @@ def test_info_with_errors(self, send_request):
...
+        """
+        Assert errors are printed if returned back in the request
+        """
+        runner = testing.CliRunner()
+
+        result = runner.invoke(client.list_critpath_packages,
+                               ['--url', 'http://localhost:6543', 'F27'])
+
+        self.assertEqual(result.exit_code, 1)
+        self.assertEqual(result.output, ("ERROR: an error was encountered... :(\n"))
```
bowlofeggs: It would still be good to assert that the correct API call was made here.

bodhi/tests/client/test___init__.py
```diff
@@ -1696,6 +1696,158 @@ def test_info_with_errors(self, send_request):
...
+
+class TestAddCritpathPackages(unittest.TestCase):
+    """
+    Test the add_critpath_packages() function.
+    """
+    @mock.patch('bodhi.client.bindings.BodhiClient.csrf',
+                mock.MagicMock(return_value='a_csrf_token'))
+    @mock.patch('bodhi.client.bindings.BodhiClient.send_request',
+                return_value=client_test_data.EXAMPLE_CRITPATH_PACKAGES_MUNCH, autospec=True)
+    def test_url_flag(self, send_request):
```
bowlofeggs: It seems that the other tests set the ```--url``` flag anyway, so this test may be redundant? Same for the similarly named other tests.

bodhi/tests/client/test___init__.py
```diff
@@ -1696,6 +1696,158 @@ def test_info_with_errors(self, send_request):
...
+        """
+        Assert errors are printed if returned back in the request
+        """
+        runner = testing.CliRunner()
+
+        result = runner.invoke(client.add_critpath_packages,
+                               ['--url', 'http://localhost:6543', 'F27', 'python', 'kernel'])
+
+        self.assertEqual(result.exit_code, 1)
+        self.assertEqual(result.output, ("ERROR: an error was encountered... :(\n"))
```
bowlofeggs: Similarly here, it would still be good to assert that the right call was made.

bodhi/tests/client/test___init__.py
```diff
@@ -1696,6 +1696,158 @@ def test_info_with_errors(self, send_request):
...
+        """
+        Assert errors are printed if returned back in the request
+        """
+        runner = testing.CliRunner()
+
+        result = runner.invoke(client.remove_critpath_packages,
+                               ['--url', 'http://localhost:6543', 'F27', 'python', 'kernel'])
+
+        self.assertEqual(result.exit_code, 1)
+        self.assertEqual(result.output, ("ERROR: an error was encountered... :(\n"))
```
bowlofeggs: Here too.

bodhi/tests/server/services/test_releases.py
bowlofeggs: To solve this problem, we can use the koji client to verify that packages we haven't seen before exist in Koji. This way we can handle new packages while also verifying that they really exist.

bodhi/tests/server/services/test_releases.py
```diff
@@ -327,3 +339,49 @@ def test_query_releases_html_two_releases_same_state(self):
...
+        self.assertEqual(packages[0]['name'], u'bodhi')
+
+    def test_add_critpath_packages(self):
+        self.app.put('/releases/F22/critpath-packages',
+                     {'packages': 'python,json,nodejs,git'})
+
+        release = self.db.query(Release).filter(Release.name == 'F22').one()
+        pkg_list = [pkg.name for pkg in release.critpath_pkgs]
+
+        self.assertEqual(set(pkg_list), {'bodhi', 'python', 'json', 'nodejs', 'git'})
```
bowlofeggs: Rather than creating a list and then a set, you can use a set comprehension on line 355 above to get it as a set in the first place. Same for other places that do this below.

bowlofeggs (changes requested):

bodhi/server/validators.py
ghost: ```validate_release``` accepts parameter from ```request.validated``` and not the URL. That's why created a new validator

bodhi/tests/client/test___init__.py
ghost: Is it a good idea to test normal execution? That test just checks whether, if normally executed, function returns expected output. It's name is not appropriate though - maybe ```test_list_critpath_packages```

bowlofeggs: Yeah if that's the goal, let's change the name and keep it.

ghost:
jenkies test

bowlofeggs:
The test failure is due to [GPG issues](https://pagure.io/releng/issue/7693), but I can confirm that the CI tests pass on my laptop for f27, f28, and pip.

bodhi/server/validators.py
```diff
@@ -600,12 +600,21 @@ def validate_packages(request, **kwargs):
...
 
-    if bad_packages:
-        request.errors.add('querystring', 'packages',
-                           "Invalid packages specified: {}".format(
-                               ", ".join(bad_packages)))
-    else:
-        request.validated["packages"] = validated_packages
+    # Check if bad_packages are in koji
+    # If yes, create new package in Bodhi
+    koji_session = koji.ClientSession(config.get('koji_hub'))
```
bowlofeggs: Let's use ```bodhi.server.buildsys.get_session()``` here instead of instantiating it directly.

bodhi/tests/client/test___init__.py
bowlofeggs: Don't forget to change the name here and the description below.

bodhi/tests/client/test_bindings.py
```diff
@@ -1483,6 +1483,64 @@ def test_get_releases(self):
...
+class TestBodhiClient_manage_critpath_packages(unittest.TestCase):
+    """
+    Test the BodhiClient.manage_critpath_packages() method.
+    """
+    @mock.patch('bodhi.client.bindings.BodhiClient.critpath_package_request')
+    def test_get_critpath_packages(self, request_method):
+        """Assert correct behavior from the get_critpath_packages() method."""
+        client = bindings.BodhiClient()
+
+        client.get_critpath_packages(release='F27')
```
bowlofeggs: We should assert that the return value is correct.

bodhi/tests/client/test_bindings.py
```diff
@@ -1483,6 +1483,64 @@ def test_get_releases(self):
...
+
+        request_method.assert_called_once_with(release='F27', verb='GET', auth=False, data=None)
+
+    @mock.patch('bodhi.client.bindings.BodhiClient.critpath_package_request')
+    def test_add_critpath_packages(self, request_method):
+        """Assert correct behavior from the add_critpath_packages() method."""
+        client = bindings.BodhiClient()
+
+        client.add_critpath_packages(release='F27', packages=['python', 'firefox'])
```
bowlofeggs: We should assert the return value here too.

bodhi/tests/client/test_bindings.py
```diff
@@ -1483,6 +1483,64 @@ def test_get_releases(self):
...
+
+        request_method.assert_called_once_with(release='F27', verb='PUT', auth=True,
+                                               data={'packages': 'python,firefox'})
+
+    @mock.patch('bodhi.client.bindings.BodhiClient.critpath_package_request')
+    def test_remove_critpath_packages(self, request_method):
+        """Assert correct behavior from the remove_critpath_packages() method."""
+        client = bindings.BodhiClient()
+
+        client.remove_critpath_packages(release='F27', packages=['python', 'firefox'])
```
bowlofeggs: And here.

bodhi/tests/server/services/test_releases.py
```diff
@@ -383,3 +395,63 @@ def test_query_releases_html_two_releases_same_state(self):
         self.assertEqual(res.content_type, 'text/html')
         self.assertIn('Fedora 22', res)
         self.assertIn('Fedora 42', res)
+
+    def test_get_critpath_packages(self):
+        res = self.app.get('/releases/F22/critpath-packages')
+        packages = res.json_body['packages']
+
+        self.assertEqual(len(packages), 1)
+        self.assertEqual(packages[0]['name'], u'bodhi')
```
bowlofeggs: A simpler way to assert the length and the name at the same time:

```
self.assertEqual([p['name'] for p in packages], ['bodhi'])
```

bowlofeggs:
For future reference, the new python-fedora can be installed in the Vagrant box like this until it is released upstream:

```
$ sudo pip-3 install git+https://github.com/vismay-golwala/python-fedora.git@support_put_delete --upgrade
```

bowlofeggs (changes requested):
I am not able to use the CLI to add critpath packages, as it does not give me a way to authenticate:

```
[vagrant@bodhi-dev bodhi]$ bodhi releases critpath add F28 bodhi firefox kernel
: Check your FAS username & password
```

This should work like the other CLI commands that require authentication - I should be able to pass my username via environment variables or flags, and if I do not and it is needed I should be prompted to enter my username and password.

bodhi/client/bindings.py
```diff
@@ -858,6 +858,81 @@ def get_releases(self, **kwargs):
...
+
+        Args:
+            release (unicode): The release name to remove packages from.
+            packages (list): List of critpath package names to remove.
+        Returns:
+            dict: A dictionary returned by critpath_package_requests() method.
+        """
+        package_str = ','.join(packages)
+        # DELETE request method ignores the body sometimes, so passing dictionary doesn't work
+        # However, passing it's string representation works somehow.
```
bowlofeggs, line 901: We should identify why this is the case. Is it due to a bug in python-fedora? Is it the REST API?

bodhi/tests/client/test___init__.py
ghost: @bowlofeggs - I removed ```--url``` flag from the tests below it. So this one is just for testing that flag.

bowlofeggs:
Progress on this PR should wait until there is a python-fedora release with this patch included:

https://github.com/fedora-infra/python-fedora/pull/213

bowlofeggs:
I have requested a new release at https://github.com/fedora-infra/python-fedora/issues/219

cverna:
Closing this in favour of #3049

PR #2482, opened 2018-07-17
Don't require notes when --file is used in the CLI.

### LINKED ISSUE #2480: CLI forces user to pass --notes or --notes-file when --file is used
```fedpkg``` passes update notes via the update ini file, but the development version of Bodhi now requires ```--notes``` or ```--notes-file``` to be passed to the client. This breaks ```fedpkg update```:

```
$ fedpkg update
ERROR: must specify at least one of --notes, --notes-file
Could not execute update: Could not generate update request: Command 'bodhi updates new --file bodhi.template --user bowlofeggs js-jquery-file-upload-9.22.0-1.fc27' returned non-zero exit status 1.
```

This is a 3.9.0 release blocker.

### PR DESCRIPTION
**Pull request by bowlofeggs**
fixes #2480

### DIFF
--- a/bodhi/client/__init__.py
+++ b/bodhi/client/__init__.py
@@ -307,8 +307,8 @@ def new(user, password, url, **kwargs):
 
     kwargs['notes'] = _get_notes(**kwargs)
 
-    if not kwargs['notes']:
-        click.echo("ERROR: must specify at least one of --notes, --notes-file")
+    if not kwargs['notes'] and not kwargs['file']:
+        click.echo("ERROR: must specify at least one of --file, --notes, or --notes-file")
         sys.exit(1)
 
     for update in updates:
--- a/bodhi/tests/client/test___init__.py
+++ b/bodhi/tests/client/test___init__.py
@@ -20,6 +20,7 @@
 import datetime
 import os
 import platform
+import tempfile
 import unittest
 import copy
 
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Review by cverna — DISMISSED** [review]
LGTM

**Comment by pypingou** [comment]
Looks like one of the test doesn't pass (output doesn't fit expectations)

**Comment by bowlofeggs** [comment]
Strange, the ```bdiff-cover``` command in Vagrant passes all the tests, even when I rebase it onto the latest develop. I'll push up that rebase to see if that helps.

**Comment by bowlofeggs** [comment]
This run failed CI due to https://bugzilla.redhat.com/show_bug.cgi?id=1607916

**Comment by bowlofeggs** [comment]
Hmm, I'm still not able to reproduce this failure locally. I've pushed up a patch with some debugging print statement so I can see what isn't matching in that output.

**Comment by bowlofeggs** [comment]
Ah, I figured out what the failure was *and* why it passes locally. In Vagrant, the ```BODHI_URL``` env var is set which influences the URLs printed by the CLI on update reports. It happened to be set to what the test data expected to see. In CI that is not set and so the URL was different in the output. I've got a fix incoming.

**Comment by bowlofeggs** [comment]
Here is the fix:

```
index b837510d..984e25a6 100644
@@ -410,8 +410,8 @@ class TestNew(unittest.TestCase):
                  '--file', update_file.name])

-        expected_output = client_test_data.EXPECTED_UPDATE_OUTPUT.replace('example.com/tests',
-                                                                          'localhost:6543')
+        expected_output = client_test_data.EXPECTED_UPDATE_OUTPUT.replace(
+            'http://example.com/tests', 'https://bodhi.fedoraproject.org')
         self.assertTrue(compare_output(result.output, expected_output))
         bindings_client = send_request.mock_calls[0][1][0]
         calls = [
```

puiterwijk (approved):
LGTM

PR #2492, opened 2018-07-23
Only logged in users can add karma to updates.

### LINKED ISSUE #2474: karma voting section for anonymous users should indicate their karma will be ignored
https://bodhi.fedoraproject.org/updates/FEDORA-EPEL-2018-88c0c447a4
has what appears to be three karma: 👍items.
But, afaict, the third, from anonymous is ignored by the calculator.
We believe that's to protect against bots. That's fine, but I'm fairly confident a human wrote that message.

It would be nice if the interface told the human that they should use an account if they want their vote to be tallied.

mattiaverga:
Anonymous users will no more be able to post main karma to updates. This change will also remind users to log in if they want they're votes to be taken into account by the "autopush to stable" feature.

Users not logged in will see a lock icon and the description greyed out, plus the radio buttons will be disabled:
immagine

Fixes #2474 

### DIFF
--- a/bodhi/server/templates/update.html
+++ b/bodhi/server/templates/update.html
@@ -766,7 +766,22 @@ $(document).ready(function(){
                         <td data-class="danger">  <input type="radio" name="karma" value="-1"> </td>
                         <td>                      <input type="radio" name="karma" value="0" checked> </td>
                         <td data-class="success"> <input type="radio" name="karma" value="1"> </td>
-                        <td>Is the update generally functional?</td>
+                        % if not request.user:
+                        <td>
+                          <p class="text-muted">Is the update generally functional? (karma)</p>
+                          <p class="text-danger" style="font-size: 0.8rem;">You need to be logged in to add karma!</p>
+                          <script type="text/javascript">
+                            $(document).ready(function(){
+                              var karmaRadio = document.getElementsByName("karma");
+                              for(var i=0, iLen=karmaRadio.length; i<iLen; i++) {
+                                karmaRadio[i].disabled = true;
+                              }
+                            });
+                          </script>
+                        </td>
+                        % else:
+                        <td>Is the update generally functional? (karma)</td>
+                        % endif
                       </tr>
 
                     </table>
--- a/bodhi/server/templates/update.html
+++ b/bodhi/server/templates/update.html
@@ -733,35 +733,6 @@ $(document).ready(function(){
                         </tr>
                       </thead>
 
-                      % for bug in update.bugs:
-                      <tr>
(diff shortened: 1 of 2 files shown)


bowlofeggs (commented):
I personally like this change because I think it makes it more obvious that you need to be logged in for your vote to count, but since it's a bit of a departure from what we've been doing I'd like to see if someone from QA would like to chime in too.

@AdamWill @kparal Do either of you have an opinion on this?

kparal:
I think requiring login to submit karma is fine, better than allowing it but not counting it which might be confusing for maintainers. Any issues should be written out in comment anyway, so we're not missing out.

I don't like the UI, though. The mockup in #2474 looks clearer to me. The red lock doesn't tell me anything. I assume a popup is implemented with more description on mouse over? Anyway, popups are fine for non-important stuff, but important stuff should be written out clearly, as in the mockup. It's more visible and more understandable.

I'd also make the karma vote appear as the first one in the list, and the critpath vote the second one (then bugs and test cases). It always confused me why the most important option is listed last.

Those are just my nitpicks, nothing serious.

mattiaverga:
> I assume a popup is implemented with more description on mouse over?

Yes, I forgot to say that.
Anyway, I can easily revert to the mockup if you prefer, or make a combo of the two, with or without strikethrough text:
immagine
We can choose the preferred one!


> I'd also make the karma vote appear as the first one in the list, and the critpath vote the second one (then bugs and test cases).

+1

bowlofeggs:
Cool I like the new mockup too. I don't have a strong feeling, but I have a slight preference to the one without strikethrough. Either one is cool with me though if others feel strongly.

kparal:
> Cool I like the new mockup too. I don't have a strong feeling, but I have a slight preference to the one without strikethrough.

That's the second one or the third one? :-) I'd personally pick the third one and remove the lock icon (to avoid visual clutter with too many icons, bodhi page is colorful enough already). So IOW the first one without the strikethrough. But I do not have strong feelings about it, I just prefer important messages to be immediately visible instead of hidden behind popup, that's all.

bowlofeggs:
@kparal I don't feel strongly about the lock icon, but I agree that the presence of the red text is helpful and is clearer than a hoverover.

@ryanlerch do you have an opinion here?

mattiaverga:
I've changed the style as in the original mockup and I also moved karma and critpath votes on top of the list:
immagine

bowlofeggs (approved):
Looks great, thanks for adjusting to the feedback!

bowlofeggs:

bowlofeggs:
These patches have been deployed to Fedora's staging Bodhi instance:

PR #2495, opened 2018-07-24
Change the '--username' option to '--user' for 'bodhi release'.

ghost:

### DIFF
--- a/bodhi/tests/client/test___init__.py
+++ b/bodhi/tests/client/test___init__.py
@@ -1458,7 +1458,7 @@ class TestCreate(unittest.TestCase):
 
         result = runner.invoke(
             client.create_release,
-            ['--name', 'F27', '--url', 'http://localhost:6543', '--username', 'bowlofeggs',
+            ['--name', 'F27', '--url', 'http://localhost:6543', '--user', 'bowlofeggs',
              '--password', 's3kr3t'])
 
         self.assertEqual(result.exit_code, 0)
@@ -1486,7 +1486,7 @@ class TestCreate(unittest.TestCase):
 
         result = runner.invoke(
             client.create_release,
-            ['--name', 'F27', '--url', 'http://localhost:6543', '--username', 'bowlofeggs',
+            ['--name', 'F27', '--url', 'http://localhost:6543', '--user', 'bowlofeggs',
              '--password', 's3kr3t'])
 
         self.assertEqual(result.exit_code, 1)
@@ -1510,7 +1510,7 @@ class TestEditRelease(unittest.TestCase):
         result = runner.invoke(
             client.edit_release,
             ['--name', 'F27', '--long-name', 'Fedora 27, the Greatest Fedora!', '--url',
-             'http://localhost:6543', '--username', 'bowlofeggs', '--password', 's3kr3t'])
+             'http://localhost:6543', '--user', 'bowlofeggs', '--password', 's3kr3t'])
 
         self.assertEqual(result.exit_code, 0)
         self.assertEqual(result.output, client_test_data.EXPECTED_RELEASE_OUTPUT)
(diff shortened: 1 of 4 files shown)


vismay-golwala:
jenkies test

ghost:
jenkies test

bowlofeggs:
The CI failure is due to https://bugzilla.redhat.com/show_bug.cgi?id=1607916.

bodhi/tests/client/test___init__.py
```diff
@@ -1597,7 +1597,7 @@ def test_edit_with_errors(self, send_request):
         result = runner.invoke(
             client.edit_release,
             ['--name', 'F27', '--long-name', 'Fedora 27, the Greatest Fedora!', '--url',
-             'http://localhost:6543', '--username', 'bowlofeggs', '--password', 's3kr3t'])
+             'http://localhost:6543', '--user', 'bowlofeggs', '--password', 's3kr3t'])
 
         self.assertEqual(result.exit_code, 1)
         self.assertEqual(result.output, ("ERROR: an error was encountered... :(\n"))
```
bowlofeggs, line 1603: We should add a test to make sure the old ```bodhi-manage-releases``` CLI still uses ```--username``` for edit and create.

bowlofeggs (changes requested):

bodhi/tests/client/test___init__.py
ghost: What could be a good way to test this use case? I am thinking of two ways currently:

1. Modify the ```cli``` object used in ```bodhi releases``` and call that using ```--username```.

2. Call ```bodhi-manage-releases``` by accessing terminal from test code.

bowlofeggs: I think the click runner might still work, similarly to how it is used in the existing tests. If this doesn't work, we can just use mock to ensure the expected changes are made to the cli object by ```main()```. I don't think you want to run it in a shell because it will not be possible to mock and it will talk to a live server.

ghost: The click runner requires a method to invoke. However, we don't have a method for ```bodhi-manage-releases create``` as we just forward it to ```bodhi releases create```. So I will probably go with using mock to ensure expected changes.

bowlofeggs:
On 07/25/2018 03:52 PM, Vismay Golwala wrote:
> The click runner requires a method to invoke. However, we don't have a
> method for |bodhi-manage-releases create| as we just forward it to
> |bodhi releases create|.

I don't know if it would work, but my suggestion was to give click the
main() function as the function to run, and then "create…" as the
argument to it to see if that would do the trick. It might not work, but
is worth a shot.

ghost:
Yes definitely, I tried that first thing. But the invoke method of testing has following format:

```runner.invoke(group.method, [param1, param2,...])```

As we are just forwarding the request coming in to ```bodhi-manage-releases```, the main() function is not a ```click``` group and we dont have a dedicated method for ```bodhi-manage-releases create```.

**Comment by ghost** [comment]
Also I noticed that the only [test](https://github.com/fedora-infra/bodhi/blob/develop/bodhi/tests/server/scripts/test_manage_releases.py#L65) in ```test_manage_releases.py``` is wrong and not actually testing anything. Because after trying to execute the main() method, it causes Exception and jumps to the catch clause. I think Exception might be caused because main method is not called normally when we enter ```bodhi-manage-releases``` on CLI. We will need to simulate that w.r.t. how it is mapped from [here](https://github.com/fedora-infra/bodhi/blob/develop/setup.py#L163)

Might need help with that.

File: bodhi/tests/server/scripts/test_manage_releases.py [hunk 2]
```diff
@@ -68,3 +41,19 @@ def test_main(self, releases):
...
+        self.assertEqual(cli.commands['releases'].commands['create'].params[0].opts, ['--user'])
+        self.assertEqual(cli.commands['releases'].commands['edit'].params[0].opts, ['--user'])
+
+        try:
+            manage_releases.main()
+        except SystemExit:
+            pass
+
+        self.assertEqual(cli.commands['releases'].commands['create'].params[0].opts, ['--username'])
+        self.assertEqual(cli.commands['releases'].commands['edit'].params[0].opts, ['--username'])
```
bowlofeggs, line 59: Cool, this looks good enough to me.

bowlofeggs (approved):

PR #2500, opened 2018-07-25
Allow mergify PRs to run CI.

bowlofeggs:


puiterwijk (changes requested):
I added it to the whitelist in comments in #2499.
This would give it more access than needed.

PR #2503, opened 2018-07-25
Add a bci alias for Vagrant development environment.

bowlofeggs:
This commit adds a new bash alias, bci, for the Vagrant environment
that runs the CI test suite so that developers can locally
reproduce CI test results.

### DIFF
--- a/devel/ansible/roles/dev/files/.bashrc
+++ b/devel/ansible/roles/dev/files/.bashrc
@@ -9,6 +9,7 @@ fi
 # export SYSTEMD_PAGER=
 
 shopt -s expand_aliases
+alias bci="sudo dnf install -y docker parallel && sudo systemctl enable docker && sudo systemctl start docker && sudo devel/run_tests.sh $@"
 alias bdiff-cover="btest; diff-cover /home/vagrant/bodhi/coverage.xml --compare-branch=develop --fail-under=100"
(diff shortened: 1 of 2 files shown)


devel/ansible/roles/dev/files/.bashrc
```diff
@@ -9,6 +9,7 @@ fi
 # export SYSTEMD_PAGER=
 
 shopt -s expand_aliases
+alias bci="sudo dnf install -y docker parallel && sudo systemctl enable docker && sudo systemctl start docker && sudo devel/run_tests.sh $@"
```
cverna, line 12: do we really want to run dnf install docker and systemctl everytime this command is used ? Why not just add these to the ansible role and make `btest` be the same as the CI ?

bowlofeggs: I thought about that, but most installs probably won't use this command and the SELinux bits of the install take an annoyingly long time. That said, I'm happy to change it if you prefer because I do feel "on the fence" about it.

Also, weirdly, I will never use this command because it causes one of my computers to freeze, and the command causes the Vagrant VM to irreversibly pause on my other computer. I've not been able to figure out why this happens on either computer, but @jeremycline says that doing this on his system works so maybe it'll work for some people.

bowlofeggs:
jenkies test

devel/ansible/roles/dev/files/.bashrc
bowlofeggs: I've thought about it some more and I think I am swayed. I kinda felt bad about doing it this way in the first place.

I'll close for now and reopen later.

PR #2513, opened 2018-07-27
Use podman instead of docker for running CI jobs.

bowlofeggs:

### DIFF
--- a/devel/ci/run_tests.sh
+++ b/devel/ci/run_tests.sh
@@ -22,8 +22,6 @@
 
 
 sudo yum install -y epel-release
-sudo yum install -y bzip2 docker parallel
-
-sudo systemctl start docker
+sudo yum install -y bzip2 podman parallel
 
 BUILD_PARALLEL="-j 4" ./devel/run_tests.sh -a
--- a/devel/run_tests.sh
+++ b/devel/run_tests.sh
@@ -21,7 +21,7 @@
 # containers to test Bodhi across supported Fedora releases (See the RELEASES variable below). You
 # can set the RELEASES environment variable to override the RELEASES for a given test run if you
 # like. You can pass a -x flag to it to get it to exit early if a build or test run fails.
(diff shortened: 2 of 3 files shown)


bowlofeggs:
Hmm, it seems there is some sort of networking issue with this patch:

```
f28	STEP 3: RUN dnf upgrade -y libsolv || echo "We are not trying to test dnf upgrade, so ignoring dnf failure."
f28	error running container: error configuring network list if0 for [/bin/sh -c dnf upgrade -y libsolv || echo "We are not trying to test dnf upgrade, so ignoring dnf failure."]: failed to set bridge addr: could not add IP address to "cni0": file exists
f28	error building at step {Env:[DISTTAG=f28container FGC=f28  PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin] Command:run Args:[dnf upgrade -y libsolv || echo "We are not trying to test dnf upgrade, so ignoring dnf failure."] Flags:[] Attrs:map[] Message:RUN dnf upgrade -y libsolv || echo "We are not trying to test dnf upgrade, so ignoring dnf failure." Original:RUN dnf upgrade -y libsolv || echo "We are not trying to test dnf upgrade, so ignoring dnf failure."}: exit status 1
```

pypingou:
Change in itself looks fine to me, but I'm curious if there is a specific reason to switch? Is one faster than the other?

bowlofeggs:
On 07/30/2018 03:10 AM, Pierre-Yves Chibon wrote:
> I'm curious if there is a specific reason to switch? Is one faster than
> the other?

There was a call to test out podman as a drop-in replacement for docker.
Since docker is a hostile upstream, I'm inclined to aid the effort to
replace it.

Clearly, it's not quite a drop-in replacement on CentOS 7 (though it
does work on Fedora).

bowlofeggs:
Eh, I now feel unmotivated to pursue this at this time.

PR #2514, opened 2018-07-27
Execute skipped tests for Python 3

ghost:

### DIFF
--- a/bodhi/server/consumers/masher.py
+++ b/bodhi/server/consumers/masher.py
@@ -1309,7 +1309,7 @@ class PungiComposerThread(ComposerThread):
         master_repomd_url = self._get_master_repomd_url(arch)
 
         with open(repomd) as repomdf:
-            checksum = hashlib.sha1(repomdf.read()).hexdigest()
+            checksum = hashlib.sha1(repomdf.read().encode('utf-8')).hexdigest()
         while True:
             try:
                 self.log.info('Polling %s' % master_repomd_url)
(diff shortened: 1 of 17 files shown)


bodhi/server/consumers/masher.py
```diff
@@ -1302,7 +1302,7 @@ def _wait_for_sync(self):
         master_repomd_url = self._get_master_repomd_url(arch)
 
         with open(repomd) as repomdf:
-            checksum = hashlib.sha1(repomdf.read()).hexdigest()
+            checksum = hashlib.sha1(str(repomdf.read()).encode('utf-8')).hexdigest()
```
bowlofeggs: You should not need the call to ```str()``` here:

```
$ python3
>>> import hashlib
>>> with open('bodhi/server/__init__.py') as tf:
...     hashlib.sha1(tf.read().encode('utf-8')).hexdigest()
...
'56ee127f54c7fcf7453817d257106747cfa9b461'
```

bodhi/server/consumers/masher.py
```diff
@@ -1311,7 +1311,7 @@ def _wait_for_sync(self):
                 self.log.exception('Error fetching repomd.xml')
                 time.sleep(200)
                 continue
-            newsum = hashlib.sha1(masterrepomd.read()).hexdigest()
+            newsum = hashlib.sha1(str(masterrepomd.read()).encode('utf-8')).hexdigest()
```
bowlofeggs: Same here.

bowlofeggs (changes requested):

bowlofeggs (approved):
The tests pass everywhere except Rawhide, and that failure is a failure in Rawhide itself, so I'll just wait for that to get fixed and then we can merge this. It'll probably be at least a day or two before that gets fixed.

bowlofeggs:
jenkies test

bowlofeggs:

PR #2516, opened 2018-07-27
Update the docs for style checks.

bowlofeggs:
The docs referred to two unit tests that used to be used to enforce
style but no longer exist. This commit removes those references.


pypingou (approved):
Looks fine overall, but I liked the fact that it specified that simply running the tests would take care of the style checks.

bowlofeggs:
On 07/30/2018 03:08 AM, Pierre-Yves Chibon wrote:
> I liked the fact that it specified that simply running the tests would
> take care of the style checks.

It's being removed because that is no longer true. We now rely on the CI
system to run the tests, or the Vagrant shell aliases.

puiterwijk (approved):

PR #2525, opened 2018-08-01
Make Release.from_tags() more predictable

owtaylor:
IIRC, while doing some debugging of a test setup, I noticed that [tag, unknownTag] worked, but [unknownTag, tag] didn't, and found this. It's mostly a cleanup though probably results in better errors in some cases when going through the code paths that didn't catch KeyError.

<hr>

If Release.from_tags() was called with a tag not associated with any
release, it would raise KeyError (unless an earlier tag in the list
was found first.) This was unexpected from the documentation - and was
handled by one caller but not others. Change it to return None in
this case - this was handled by all callers.

### DIFF
--- a/bodhi/server/models.py
+++ b/bodhi/server/models.py
@@ -947,6 +947,8 @@ class Release(Base):
         """
         tag_types, tag_rels = cls.get_tags(session)
         for tag in tags:
+            if tag not in tag_rels:
+                continue
             release = session.query(cls).filter_by(name=tag_rels[tag]).first()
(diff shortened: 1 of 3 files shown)


bowlofeggs (commented):
Looks good to me, but there's a merge conflict. I will see if GitHub will let me rebase your patch and fix it for you.

bowlofeggs (approved):

bowlofeggs:

bowlofeggs:

PR #2526, opened 2018-08-01
Add Flatpak support

owtaylor:
Add a new content type for flatpaks, which is inferred from the Koji metadata.
Flatpaks are mashed exactly the same as containers - the content type is
different mostly to get a non-confusing display of content type in the user
interface and to allow future enhancements like Flatpak-specific testing
instructions.

### DIFF
--- a/bodhi/server/buildsys.py
+++ b/bodhi/server/buildsys.py
@@ -261,6 +261,16 @@ class DevBuildsys(Buildsystem):
                 }
                 break
 
+            if token.endswith("flatpak"):
+                tag = "f%s-updates-testing" % token.replace("fc", "")
+                data['extra'] = {
+                    'container_koji_task_id': 21098765,
+                    'image': {
+                        'flatpak': True
+                    },
+                }
+                break
+
             # Hardcoding for modules in the dev buildsys
             if token.startswith("2017"):
                 tag = "f27M-updates-testing"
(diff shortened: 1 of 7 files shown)


bodhi/server/buildsys.py
```diff
@@ -261,6 +261,16 @@ def getBuild(self, build='TurboGears-1.0.2.2-2.fc17', other=False, testing=False
                 }
                 break
 
+            if token.endswith("flatpak"):
+                tag = "f%s-updates-testing" % token.replace("fc", "").replace("", "")
```
bowlofeggs: Let's drop the ```replace()``` call if it's not needed.

bodhi/server/consumers/masher.py
```diff
@@ -868,6 +869,12 @@ def _compose_updates(self):
                     copy_container(build, destination_tag=dtag)
 
 
+class FlatpakComposerThread(ContainerComposerThread):
+    """Use skopeo to copy and tag flatpak images."""
+
+    ctype = ContentType.flatpak
```
bowlofeggs, line 875: Heh, looks like patching this file was pretty easy ☺

bodhi/server/models.py
```diff
@@ -1075,9 +1080,11 @@ def get_pkg_committers_from_pagure(self):
         """
         pagure_url = config.get('pagure_url')
         # Pagure uses plural names for its namespaces such as "rpms" except for
-        # container
+        # container. Flatpaks build directly from the 'modules' namespace
```
bowlofeggs, line 1085: Ah, so I take this to mean that ACLs on a flatpak are identical to having ACLs to the module of the same name?

bowlofeggs (changes requested):
I will fix the ```replace()``` for you, so no further action needed from you. Thanks for the patch!

bowlofeggs (approved):
Thanks for the patch!

bowlofeggs:

bowlofeggs:
These patches have been deployed to Fedora's staging Bodhi instance:

PR #2529, opened 2018-08-07
Require severity for security update.

### LINKED ISSUE #2206: Security updates should require severity to be set
It recently came to [FESCo's attention](https://pagure.io/fesco/issue/1820) that security updates often don't have severity set by the packager (it's often set to "unspecified", the default). This makes it difficult to tell which updates might be important to push immediately, vs. waiting for the next batch. Outside of that problem, it would also be better for our users if we had indications on the severity of our security updates, now that [severity is included in our updateinfo.xml](https://github.com/fedora-infra/bodhi/pull/2099).

ghost:
fixes #2206

### DIFF
--- a/bodhi/client/__init__.py
+++ b/bodhi/client/__init__.py
@@ -96,7 +96,8 @@ new_edit_options = [
     click.option('--type', default='bugfix', help='Update type', required=True,
                  type=click.Choice(['security', 'bugfix', 'enhancement', 'newpackage'])),
     click.option('--severity', help='Update severity',
-                 type=click.Choice(['unspecified', 'low', 'medium', 'high', 'urgent'])),
+                 type=click.Choice(['unspecified', 'low', 'medium', 'high', 'urgent']),
+                 is_eager=True),
     click.option('--notes', help='Update description'),
     click.option('--notes-file', help='Update description from a file'),
     click.option('--bugs', help='Comma-separated list of bug numbers', default=''),
@@ -306,6 +307,19 @@ def updates():
     pass  # pragma: no cover
 
 
+def require_severity_for_security_update(type, severity):
+    """
+    Print an error message if the user did not provide severity for a security update.
+
+    Args:
+        type (unicode): The value of the update 'type'.
+        severity (unicode): The value of the update 'severity'.
+    """
+    if type == 'security' and severity not in ('low', 'medium', 'high', 'urgent'):
+        raise click.BadParameter('must specify severity for a security update',
+                                 param_hint='severity')
+
+
 @updates.command()
 @add_options(new_edit_options)
 @click.argument('builds')
@@ -348,6 +362,7 @@ def new(user, password, url, debug, **kwargs):
         sys.exit(1)
 
     for update in updates:
+        require_severity_for_security_update(type=update['type'], severity=update['severity'])
         try:
             resp = client.save(**update)
             print_resp(resp, client)
@@ -431,6 +446,8 @@ def edit(user, password, url, debug, **kwargs):
         for field in kwargs:
             if kwargs[field] in (None, '') and field in former_update:
                 kwargs[field] = former_update[field]
+
+        require_severity_for_security_update(type=kwargs['type'], severity=kwargs['severity'])
 
         resp = client.save(**kwargs)
         print_resp(resp, client)
--- a/bodhi/server/services/updates.py
+++ b/bodhi/server/services/updates.py
@@ -57,6 +57,7 @@ from bodhi.server.validators import (
     validate_requirements,
     validate_bugs,
     validate_request,
+    validate_severity,
 )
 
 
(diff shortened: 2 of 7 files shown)


ghost:
jenkies test

bodhi/client/__init__.py
```diff
@@ -313,6 +313,11 @@ def new(user, password, url, **kwargs):
 
     for update in updates:
         try:
+            # Require severity level for a 'security' update
+            if update['type'] == 'security' and update['severity'] not in ('low', 'medium',
+                                                                           'high', 'urgent'):
+                click.echo("ERROR: must specify severity for a security update")
+                sys.exit(1)
```
bowlofeggs: ```click``` has a nice callback system that would be good to use for this. For an example, see ```_warn_if_url_and_staging_set()```.

bodhi/client/__init__.py
```diff
@@ -393,6 +398,12 @@ def edit(user, password, url, **kwargs):
             if kwargs[field] in (None, '') and field in former_update:
                 kwargs[field] = former_update[field]
 
+        # Require severity level for a 'security' update
+        if kwargs['type'] == 'security' and kwargs['severity'] not in ('low', 'medium',
+                                                                       'high', 'urgent'):
+            click.echo("ERROR: must specify severity for a security update")
```
bowlofeggs: By making a callback we won't have to repeat this code.

bodhi/server/services/updates.py
```diff
@@ -448,6 +450,15 @@ def new_update(request):
...
     data.pop('csrf_token')
 
+    # Require severity level for a 'security' update
+    if data['type'] == UpdateType.security and data['severity'] not in (UpdateSeverity.low,
+                                                                        UpdateSeverity.medium,
+                                                                        UpdateSeverity.high,
+                                                                        UpdateSeverity.urgent):
+        log.warn('No severity specified for a security update.')
+        request.errors.add('body', 'severity', 'Must specify severity for a security update')
+        return
```
bowlofeggs: Let's use the validators to require this instead of doing it inline here. The purpose of the validators is to make sure the request is valid, which is the goal here too.

bowlofeggs (changes requested):
It would probably be nice UX-wise if we made the web interface force a choice on the user too.

bodhi/client/__init__.py
ghost: That might not work in case of editing an update. Because values are fetched from server for existing update. Should the user be forced to enter both ```--type``` and ```--severity```?

bodhi/client/__init__.py
```diff
@@ -306,6 +307,19 @@ def updates():
...
+    """
+    Print an error message if the user did not provide severity for a security update.
+
+    Args:
+        type (unicode): The value of the update 'type'.
+        severity (unicode): The value of the update 'severity'.
+    """
+    if type == 'security' and severity not in ('low', 'medium', 'high', 'urgent'):
+        click.echo('ERROR: must specify severity for a security update', err=True)
+        sys.exit(1)
```
bowlofeggs: Instead of calling ```echo()``` and ```exit()```, let's do this:

```
raise click.BadParameter('must specify severity for a security update', param_hint='severity')
```

bodhi/tests/client/test___init__.py
```diff
@@ -1495,6 +1507,22 @@ def test_bodhi_client_exception(self, send_request):
         self.assertEqual(result.exit_code, 0)
         self.assertIn("This is a BodhiClientException message", result.output)
 
+    @mock.patch('bodhi.client.bindings.BodhiClient.csrf',
+                mock.MagicMock(return_value='a_csrf_token'))
+    @mock.patch('bodhi.client.bindings.BodhiClient.query',
+                return_value=client_test_data.EXAMPLE_QUERY_MUNCH, autospec=True)
+    def test_edit_security_update_without_severity(self, query):
```
bowlofeggs: s/without/with_unspecified/

bodhi/server/static/js/update_form.js
```diff
@@ -265,4 +265,20 @@ $(document).ready(function() {
...
     update_markdown_preview($("#notes").val());
 
+    var validate_severity = function() {
+        var type = $("input[name=type]:checked").val();
+        var severity = $("input[name=severity]:checked").val();
+
+        if (type == 'security') {
+            $("input[name=severity][value=unspecified]").attr('disabled', 'disabled');
+            if (severity == 'unspecified')
+                $("input[name=severity][value=medium]").attr('checked', true);
```
bowlofeggs: We don't want to automatically set the severity when the user chooses security, because the user may not notice that we picked medium for them and then submit the form. Incorrect data is worse than unspecified, so this could lead to misinformation. Rather, we should give the user an error if they submit the form without making a choice.

bowlofeggs (changes requested):

ghost:
jenkies test

bowlofeggs (approved):

PR #2532, opened 2018-08-13
Add '--wait' flag to ensure override is active

### LINKED ISSUE #1807: Add a --wait flag to bodhi overrides edit/save
We could add a ```--wait``` (or similar name, name isn't important) flag to ```bodhi overrides edit/save``` which would cause bodhi to run the ```koji wait-repo``` command in a subshell and wait for it to exit. This would be a nice convenience for our users.

Pull request https://github.com/fedora-infra/bodhi/pull/1805 added a feature where these commands can print the ```koji wait-repo``` command that would make sense to run for the given override, so we could piggy back off that work and just have subshell execute the generated command if the user requests it.

I also considered making this the default behavior, having a --no-wait flag (or similar) for users who don't want bodhi to do that for them. I'm kind of on the fence about which default is better, and welcome discussion.

ghost:
fixes #1807

### DIFF
--- a/bodhi/client/__init__.py
+++ b/bodhi/client/__init__.py
@@ -247,23 +247,23 @@ def _save_override(url, user, password, staging, edit=False, **kwargs):
         kwargs (dict): Other keyword arguments passed to us by click.
     """
     client = bindings.BodhiClient(base_url=url, username=user, password=password, staging=staging)
-
     resp = client.save_override(nvr=kwargs['nvr'],
                                 duration=kwargs['duration'],
                                 notes=kwargs['notes'],
                                 edit=edit,
                                 expired=kwargs.get('expire', False))
-    resp['wait'] = kwargs['wait']
-    print_resp(resp, client)
 
-    if resp['wait']:
-        if 'release_id' in resp.build:
-            release = client.get_releases(ids=[resp.build.release_id])['releases'][0]
-            args = ('koji', 'wait-repo', '{}-build'.format(release.dist_tag),
-                    '--build={}'.format(resp.build.nvr))
-            ret = subprocess.call(args, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+    if kwargs['wait']:
+        print_resp(resp, client, override_hint=False)
+        command = _generate_wait_repo_command(resp, client)
+        if command:
+            click.echo("\n\nRunning {}\n".format(' '.join(command)))
+            ret = subprocess.call(command, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
             if ret:
                 click.echo("WARNING: ensuring active override failed for {}".format(resp.build.nvr))
+                sys.exit(ret)
+    else:
+        print_resp(resp, client, override_hint=True)
 
 
 @click.group()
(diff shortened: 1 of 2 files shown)


ghost:
jenkies test

ghost:
jenkies test

bodhi/client/__init__.py
```diff
@@ -215,6 +217,7 @@ def _save_override(url, user, password, staging, edit=False, **kwargs):
     resp = client.save_override(nvr=kwargs['nvr'],
                                 duration=kwargs['duration'],
                                 notes=kwargs['notes'],
+                                wait=kwargs['wait'],
```
bowlofeggs: We don't want the server or the bindings to do the waiting, just the CLI.

If the server does the wait, that will hold the connection open for a long time (Koji regenerates repos every 15 minutes in Fedora's config, which means that this could cause a connection to stay open for up to 15 minutes.)

bodhi/client/__init__.py
```diff
@@ -831,28 +834,6 @@ def edit_buildroot_overrides(user, password, url, staging, **kwargs):
...
-        override (munch.Munch): A Munch of the Override we want to print a hint about.
-        client (bodhi.client.bindings.BodhiClient): A BodhiClient that we can use to query the
-            server for Releases.
-    """
-    if 'release_id' in override.build:
-        release = client.get_releases(ids=[override.build.release_id])['releases'][0]
-        click.echo(
-            '\n\nUse the following to ensure the override is active:\n\n'
-            '\t$ koji wait-repo {}-build --build={}\n'.format(
-                release.dist_tag, override.build.nvr))
```
bowlofeggs, line 941: We still want to print this when the user does not use the wait flag.

bodhi/server/schemas.py
```diff
@@ -817,6 +817,11 @@ class SaveOverrideSchema(CSRFProtectedSchema, colander.MappingSchema):
         missing=None,
     )
 
+    wait = colander.SchemaNode(
+        colander.Boolean(),
+        missing=False,
+    )
```
bowlofeggs: We should not have to alter the server for this ticket.

bowlofeggs (changes requested):

bodhi/client/__init__.py
```diff
@@ -244,13 +246,24 @@ def _save_override(url, user, password, staging, edit=False, **kwargs):
         kwargs (dict): Other keyword arguments passed to us by click.
     """
     client = bindings.BodhiClient(base_url=url, username=user, password=password, staging=staging)
+
```
bowlofeggs: This is an unrelated style change.

bodhi/client/__init__.py
```diff
@@ -244,13 +246,24 @@ def _save_override(url, user, password, staging, edit=False, **kwargs):
...
     resp = client.save_override(nvr=kwargs['nvr'],
                                 duration=kwargs['duration'],
                                 notes=kwargs['notes'],
                                 edit=edit,
                                 expired=kwargs.get('expire', False))
+    resp['wait'] = kwargs['wait']
```
bowlofeggs: Since ```wait``` is not an attribute of an update, this is confusing. Why not just access wait directly below?

bodhi/client/__init__.py
```diff
@@ -244,13 +246,24 @@ def _save_override(url, user, password, staging, edit=False, **kwargs):
...
     print_resp(resp, client)
 
+    if resp['wait']:
+        if 'release_id' in resp.build:
+            release = client.get_releases(ids=[resp.build.release_id])['releases'][0]
+            args = ('koji', 'wait-repo', '{}-build'.format(release.dist_tag),
+                    '--build={}'.format(resp.build.nvr))
```
bowlofeggs: This is duplicating effort with the other code that also forms this statement. We should share code so there's only one implementation.

bowlofeggs (changes requested):
I will fix these issues.

bowlofeggs (approved):

bowlofeggs:

bowlofeggs:

PR #2533, opened 2018-08-15
[WIP] Enforce the ContentSecurityPolicy header

### LINKED ISSUE #2523: The search bar does not work in staging
I have so far been unable to reproduce the problem in the development environment, but I am unable to get the search bar to work in staging. It seems that the code in ```bodhi/server/static/js/search.js``` does not get executed on page load, but I do not know why yet.

bowlofeggs:


bowlofeggs:
This patch is not ready for review yet - I was working on it on the train to Flock but only really got it halfish done. I will finish it later since it's now not a release blocker to fix this.

bowlofeggs:
I put no-mergify on this because I would like @puiterwijk to review it when it is ready. It is not ready now.

mattiaverga:
Can I suggest to also change `http://0.0.0.0:6543 http://localhost:6543` to `'self'` in [development.ini.example#L20-L21](https://github.com/fedora-infra/bodhi/blob/3faf603cfd97d36b1ccffb38560c781d67ff6049/devel/development.ini.example#L20-L21)?

I've spent a lot of time figuring out why the comment button wasn't working in my Vagrant box and it turned out that using address `127.0.0.1:6543` instead of `localhost:6543` was the problem...

bodhi/server/templates/master.html
```diff
@@ -14,6 +14,27 @@
...
 
+    <script type="text/javascript" nonce="${request.js_nonce}">
+        var settings = {};
+        var settings_loaded = false;
+        var to_load_after_settings = [];
+        function run_after_ready(func) {
+            if (settings_loaded) {
+                func();
+            } else {
+                to_load_after_settings.push(func);
```
mattiaverga: I think this can be better handled by creating a custom event in document when the settings are loaded.
Something like: 
on document ready try to load settings by an ajax call;
if settings are correctly loaded, trigger a custom event "settingsLoaded";
change scripts which depends on those settings to run when $(document).on('settingsLoaded') instead of on $(document).ready()

This way, if settings cannot be retrieved for whatever reason, we don't try to run scripts with broken settings and we can also display an error message.

mattiaverga: I also think that would be nice to save settings to sessionStorage to avoid requesting them on every page load.

bowlofeggs: I'm considering dropping the new settings endpoint and just using a nonce to inject the settings into the master template with a script tag. I think that'll be simple and simple is good ☺

bowlofeggs:
@mattiaverga I'm not familiar with the use of ```self``` for CORS origins. I couldn't find documentation on that value - do you know a good resource about it?

bowlofeggs:
Ah, I see self referenced here: https://content-security-policy.com/

@mattiaverga if you want to get that working, feel free to send a pull request.

bowlofeggs:
Alright, I was able to make a lot of progress on this today. It turned out that we need a newer version of ```fedora-bootstrap``` and ```jQuery``` to be able to enforce the new ```Content-Security-Policy``` in this patch.

```fedora-bootstrap``` < 1.0.2 [used some code that is disallowed](https://pagure.io/fedora-bootstrap/issue/12). I would prefer not to bundle it anymore (and that goes for all of our JavaScript and CSS), but this pull request is already getting a bit out of scope so that's a problem I think I may solve another day. For now I bundled it, but put it in its own commit so it's at least isolated. I also filed a [ticket](https://pagure.io/fedora-bootstrap/issue/15) to request it to be packaged for Fedora. There does seem to be a problem, however, as Bodhi does not render properly with this version of ```fedora-bootstrap```:

screenshot from 2019-01-09 15-08-08

I don't know what is going on there. @ryanlerch, can you tell me what I might need to do to get Bodhi to look right with the new version of ```fedora-bootstrap```?

For ```jQuery```, I was able to use the distro-provided package to get a much newer version so that was nice. We no longer bundle it, so we'll get updates as they come out (and someone else will maintain it for us now!)

This pull request is still a work in progress until I can figure out what is wrong with the styling, and I also need to go through more pages to find problems. So far I've gotten the home page and the new update page to be error free with the policy.

Lastly, it occurred to me that the last commit (which is the one that @mattiaverga might be the most interested in) is not really dependent on the others. It just adds a formal way to have settings in our JS code. If desired, I could split that out into its own pull request so we can get it in sooner, if figuring out the style problems ends up taking a while.

mattiaverga:
I only have one doubt about the script which creates the `settings` var: by putting it at the end of the master.html file it means it will be executed after all the other scripts are loaded and executed.
Maybe would be safer to put that piece of script in the `<head></head>` section to be sure the `settings` var will be available to the other scripts (and to scripts loaded by sub-templates).

bowlofeggs:
@mattiaverga yeah good suggestion, I'll move it back to the top like it was before.

mattiaverga:
It seems that migrating to the new fedora-bootstrap needs a huge amount of work since would mean to migrate all the templates to Bootstrap 4: https://getbootstrap.com/docs/4.0/migration/
Can we have help from a skilled web-designer, maybe from the design team?

Apart from that, fedora-bootstrap completely breaks the default Bootstrap navbar behavior, by setting `flex-direction:column` instead of the default `flex-direction:row`.

bowlofeggs:
I think we should probably separate this PR into two, because one of the goals is easy to do and benefits @mattiaverga (injecting settings into the JS) and the other is very difficult due to the bootstrap stuff (getting the CSP header in place).

mattiaverga:
> I think we should probably separate this PR into two, because one of the goals is easy to do and benefits @mattiaverga (injecting settings into the JS) and the other is very difficult due to the bootstrap stuff (getting the CSP header in place).

I agree. About the bootstrap upgrade, I've already did a PR against bowlofeggs:2523 branch which should make bodhi templates compatible (at 90% at least...) with Bootstrap 4. @bowlofeggs if you can create a branch here to work on bootstrap upgrade and put the work you already did, I can rebase my work here also.

bowlofeggs:
I have pulled the bit that separates the settings out into #2990, leaving this pull request to focus on enforcing the ContentSecurityPolicy header (and bootstrap problems).

mattiaverga:
I've put the fedora-bootstrap upgrade stuff in a separate PR #3001 .

About getting CSP headers to work, I found some issues here:
- missing nonce in several places, see 7398c1b67e46841df364e99338a0c27f6e9c84af
- upgrade submission form not working, see the fix in 9dde776f663335c490b6c95c42ff8523472ea36c (please ignore the styling changes)
- in #3001 I've used custom checkboxes and radios from Bootstrap4, this will require to add `data:` to img-src CSP Header

PR #2535, opened 2018-08-15
Set a git user and e-mail for the CI rebase.

bowlofeggs:
CI began to fail recently at the git rebase step with this message:

+ git rebase --preserve-merges origin/develop

*** Please tell me who you are.

Run

  git config --global user.email "you@example.com"
  git config --global user.name "Your Name"

to set your account's default identity.
Omit --global to set the identity only in this repository.

fatal: unable to auto-detect email address (got 'bodhi@slave03.(none)')
You need to set your committer info first

This commit sets the e-mail and name so that git will be happy with
the rebase.

### DIFF
--- a/devel/ci/githubprb-project.yml
+++ b/devel/ci/githubprb-project.yml
@@ -88,6 +88,8 @@
             env > jenkins-env
             $ssh_cmd yum -y install rsync
             # Checkout the pull request that we received from the githubPRB plugin
+            git config --global user.email "bodhi-ci@fedoraproject.org"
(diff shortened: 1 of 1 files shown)


bowlofeggs:
This PR will not pass CI before it is merged, because it edits the script that CI runs. The CI system runs the merged version of the script, not the proposed version, so there is no way to get this PR to pass CI even though it should fix the CI failures.

Once it is merged, I should be able to rebase all the open PRs to get them passing.

puiterwijk (approved):

PR #2538, opened 2018-08-15
Edit existing override when trying to create new with same build.

### LINKED ISSUE #2030: BROs are double plus unfun when users try to create them and they already exist
@AdamWill reported today a frustrating user experience: if you try to create a buildroot override for an nvr, and an override already exists for that nvr, it will fail to create the BRO but will not redirect you or give you a link to the existing one.

We could have it just force the existing BRO to be to the new parameters, but that might be bad since the existing notes might be lost. A couple ideas:

1. It could redirect you to the edit form of the existing BRO. This way you can see and edit the existing notes field without blindly replacing it.
2. It could append your notes to the existing notes, perhaps with some autoformatting so it's clear which user wrote what. It could also pick the latter of the two dates if the BRO dates differ.

I think the second option wouldn't be too difficult. If we do it that way, the same fix might also work for the CLI.

ghost:
fixes #2030


vismay-golwala:
jenkies test

bowlofeggs (approved):
The CI tests are broken, but the tests seem to pass in Vagrant, so I'll just merge this once CI is fixed and formally passes.

bowlofeggs:
jenkies test

PR #2547, opened 2018-08-24
Use assertRegex() instead of assertRegexpMatches().

bowlofeggs:
This fixes a few DeprecationWarnings.


bowlofeggs:
Hmmm. It looks like Python 2 does not have ```assertRegex()```. I may just keep this PR around until Bodhi drops Python 2 support since that shouldn't be too long from now.

bowlofeggs:
Eh, I'll just make myself a personal ToDo to come back to this later.

pypingou (approved):
Simple enough

Zlopez (approved):
LGTM

PR #2549, opened 2018-08-24
Move util.mkmetadatadir() to bodhi.tests.server.base.

bowlofeggs:
I noticed that mkmetadatadir() was only used by the tests, and that
it was included in bodhi.server.util(). This commit moves the
function into the tests.


cverna (dismissed):
Looks good

bowlofeggs:
This is needed in 3.10 in order for https://github.com/fedora-infra/bodhi/pull/2566 to apply.

PR #2562, opened 2018-08-29
Add a clean_old_composes setting to production.ini.

### LINKED ISSUE #2561: Add a config option to turn off automatic cleaning of old composes
A recent change in Fedora has started to create lots of root owned files in the composes, which causes the [compose to fail](https://pagure.io/fedora-infrastructure/issue/7194) when it tries to clean up old composes. Bodhi recently gained the ability to automatically clean these files, so it is sad, but we must disable the feature in Fedora.

In the interest of semantic versioning, we should add a new setting that allows deployments to disable this feature. The setting should default to cleaning up the old composes.

bowlofeggs:
fixes #2561

### DIFF
--- a/bodhi/server/config.py
+++ b/bodhi/server/config.py
@@ -359,6 +359,9 @@ class BodhiConfig(dict):
         'captcha.ttl': {
             'value': 300,
             'validator': int},
+        'clean_old_composes': {
+            'value': True,
+            'validator': _validate_bool},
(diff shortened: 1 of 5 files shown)


bowlofeggs:
This feature has been deployed to Fedora's staging Bodhi instance:

puiterwijk (dismissed):
⚓

PR #2563, opened 2018-08-30
Correctly handle unicode chars in update notes.

### LINKED ISSUE #2546: Encoding error in bodhi.client.bindings.BodhiClient.update_str()
Bastien Nocera [reported](https://bugzilla.redhat.com/show_bug.cgi?id=1585660) a traceback when creating an update that used the [‘](https://www.fileformat.info/info/unicode/char/2018/index.htm) character in the update text:

```
  File "/usr/lib/python2.7/site-packages/bodhi/client/__init__.py", line 273, in new
    print_resp(resp, client)
  File "/usr/lib/python2.7/site-packages/bodhi/client/__init__.py", line 736, in print_resp
    click.echo(client.update_str(resp))
  File "/usr/lib/python2.7/site-packages/bodhi/client/bindings.py", line 712, in update_str
    buf = six.moves.cStringIO(update['notes'])
UnicodeEncodeError: 'ascii' codec can't encode character u'\u2018' in position 98: ordinal not in range(128)
```

mattiaverga:
Fixes #2546 

### DIFF
--- a/bodhi/client/bindings.py
+++ b/bodhi/client/bindings.py
@@ -738,7 +738,7 @@ class BodhiClient(OpenIdBaseClient):
         #  |-->          80 chars in total         <--|
         wrap_width = 66
         wrap_line = functools.partial(textwrap.wrap, width=wrap_width)
-        line_formatter = '{0:>12}: {1}\n'
+        line_formatter = u'{0:>12}: {1}\n'
 
         update_lines = ['{:=^80}\n'.format('=')]
         update_lines += [
@@ -797,16 +797,12 @@ class BodhiClient(OpenIdBaseClient):
             ]
 
         if update['notes']:
-            buf = six.moves.cStringIO(update['notes'])
             notes_lines = list(itertools.chain(
-                *[wrap_line(line) for line in buf]
+                *[wrap_line(update['notes'])]
             ))
-            buf.close()
             indent_lines = ['Notes'] + [' '] * (len(notes_lines) - 1)
-            update_lines += [
-                line_formatter.format(indent, line)
-                for indent, line in six.moves.zip(indent_lines, notes_lines)
-            ]
+            for indent, line in six.moves.zip(indent_lines, notes_lines):
+                update_lines.append(line_formatter.format(indent, line))
 
         update_lines += [
             line_formatter.format('Submitter', update['user']['name']),
--- a/bodhi/tests/client/test_bindings.py
+++ b/bodhi/tests/client/test_bindings.py
@@ -1183,6 +1183,18 @@ class TestBodhiClient_update_str(unittest.TestCase):
                 "Comments: bodhi (unauthenticated) ")))
 
     @mock.patch.dict(
+        client_test_data.EXAMPLE_UPDATE_MUNCH,
+        {u'notes': u'This note contains a unicode char ☺'})
+    def test_update_with_unicode_note(self):
+        """Ensure unicode content in update notes is correctly handled"""
+        client = bindings.BodhiClient()
+        client.base_url = 'http://example.com/tests/'
+
+        text = client.update_str(client_test_data.EXAMPLE_UPDATE_MUNCH)
+
+        self.assertIn(u'Notes: This note contains a unicode char ☺', text)
+
+    @mock.patch.dict(
         client_test_data.EXAMPLE_UPDATE_MUNCH, {u'alias': None})
     def test_update_with_no_alias(self):
         """Ensure we render an update with no alias set properly"""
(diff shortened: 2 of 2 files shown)


mattiaverga:
Sorry, I've tested this locally with pytest-2 and I thought it was Python 3 compatible, but obviously it's not.

It doesn't even need to encode() if plans are to switch to Python 3, maybe we can just add the test and let Python 2 problems die with it?
Are we running tests in Jenkies with `pytest-3`, but bodhi in production is still using Python 2, am I right?

bodhi/tests/client/test_bindings.py
```diff
@@ -1182,6 +1182,17 @@ def test_update_with_anon_comment(self):
...
+    @mock.patch.dict(
+        client_test_data.EXAMPLE_UPDATE_MUNCH, {u'notes': u'This note contains `unicode` char'})
+    def test_update_with_unicode_note(self):
+        """Ensure unicode content in update notes is correctly handled"""
+        client = bindings.BodhiClient()
+        client.base_url = 'http://example.com/tests/'
+
+        text = client.update_str(client_test_data.EXAMPLE_UPDATE_MUNCH)
+
+        self.assertIn(u'Notes: This note contains `unicode` char', text)
```
bowlofeggs: I modified this test a bit to make sure it generates the same traceback as reported in the original ticket:

```
diff --git a/bodhi/tests/client/test_bindings.py b/bodhi/tests/client/test_bindings.py
index 6cd8e2df..900c82a5 100644
--- a/bodhi/tests/client/test_bindings.py
+++ b/bodhi/tests/client/test_bindings.py
@@ -1183,7 +1183,8 @@ class TestBodhiClient_update_str(unittest.TestCase):
                 "Comments: bodhi (unauthenticated) ")))

     @mock.patch.dict(
-        client_test_data.EXAMPLE_UPDATE_MUNCH, {u'notes': u'This note contains `unicode` char'})
+        client_test_data.EXAMPLE_UPDATE_MUNCH,
+        {u'notes': u'This note contains a unicode char ☺'})
     def test_update_with_unicode_note(self):
         """Ensure unicode content in update notes is correctly handled"""
         client = bindings.BodhiClient()
@@ -1191,7 +1192,7 @@ class TestBodhiClient_update_str(unittest.TestCase):

         text = client.update_str(client_test_data.EXAMPLE_UPDATE_MUNCH)

-        self.assertIn(u'Notes: This note contains `unicode` char', text)
+        self.assertIn(u'Notes: This note contains a unicode char ☺', text)

     @mock.patch.dict(
         client_test_data.EXAMPLE_UPDATE_MUNCH, {u'alias': None})
```

bodhi/client/bindings.py
```diff
@@ -797,7 +797,7 @@ def update_str(self, update, minimal=False):
             ]
 
         if update['notes']:
-            buf = six.moves.cStringIO(update['notes'])
+            buf = six.moves.cStringIO(update['notes'].encode('utf-8'))
```
bowlofeggs: So I think there are a few tricky things going on in here, and after playing around a bit with the modified test I mentioned in my other comment I haven't arrived at a good solution that works in Python 2 and Python 3 at the same time.

One tricky thing is that ```cStringIO``` has different behavior here than ```StringIO``` has. Note also that Python 3 doesn't have a ```cStringIO``` and so this ```six.moves``` thing is just an alias for ```io.StringIO``` in Python 3.

We may need to do some "If Python 2... else..." type things to solve this one.

bodhi/tests/client/test_bindings.py
bowlofeggs: This test causes this error message if you run it against the unmodified bindings:

```
        if update['notes']:
>           buf = six.moves.cStringIO(update['notes'])
E           UnicodeEncodeError: 'ascii' codec can't encode character u'\u263a' in position 34: ordinal not in range(128)                                                                                                                     

bodhi/client/bindings.py:800: UnicodeEncodeError
```

bowlofeggs (changes requested):
Yeah this one is kinda tricky. I spent a little time on it tonight but it's time to stop working and I didn't arrive at a good solution either, though I do have a modification to the test that generates the same traceback.

mattiaverga:
It seems that now tests are successful, but build is failing because of the reduction in line coverage. Not sure how to handle this. I don't think using "If Python 2... else..." would solve the problem either, would it?

I think update comments should be fixed as well to work with unicode characters.

bodhi/client/bindings.py
mattiaverga: I think we don't need to use cStringIO at all... looking how updates['comments'] are managed I removed those six stuff and it seems to work just fine in Python 2 and Python 3. Correct me if there was some reason to use cStringIO that I didn't see.

bowlofeggs: Yeah I think that makes a lot of sense, thanks!

bowlofeggs (dismissed):

bowlofeggs (approved):
Thanks!

PR #2566, opened 2018-08-30
Move insert_in_repo() to metadata.py.

### LINKED ISSUE #2565: The web server recently gained a dependency on createrepo_c
@nirik noticed that the Bodhi server did not start because it didn't have ```createrepo_c``` available to it. One of the patches that went into Bodhi 3.10.0 added a function to ```bodhi.server.util``` that uses ```createrepo_c```. That function is actually only used by the composer via ```bodhi.server.metadata```, and thus isn't really needed by the web server.

Rather than adding a new dependency to the web server, I think we should just move this code from ```bodhi.server.util``` to ```bodhi.server.metadata```.

bowlofeggs:
insert_in_repo() is only used by bodhi.server.metadata, but it had
been in bodhi.server.util. This added an unnecessary dependency on
createrepo_c to the web server. By moving it into metadata.py, this
dependency is avoided.

fixes #2565

### DIFF
--- a/bodhi/server/metadata.py
+++ b/bodhi/server/metadata.py
@@ -21,6 +21,7 @@ from datetime import datetime
 import logging
 import os
 import shelve
+import shutil
 import tempfile
 
 from kitchen.text.converters import to_bytes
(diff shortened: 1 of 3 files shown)


bowlofeggs:

Zlopez (dismissed):

cverna (dismissed):

PR #2570, opened 2018-09-04
Fix a typo in the update template JavaScript.

### LINKED ISSUE #2569: ReferenceError: req is not defined
There is a regression introduced by #2394, where ```req``` is used instead of ```requirement``` [here](https://github.com/fedora-infra/bodhi/blob/4d1c284df55cc74fdc3d1ee4ff7ba6eabe9ea5a8/bodhi/server/templates/update.html#L69).

bowlofeggs:
fixes #2569

### DIFF
--- a/bodhi/server/templates/update.html
+++ b/bodhi/server/templates/update.html
@@ -66,7 +66,7 @@ ${update.type} update for ${update.beautify_title(amp=True) | n}
             if ($.inArray(requirement.testcase, requirements) == -1) {
                 requirements.push(requirement.testcase);
             }
-            $('#failed_requirements').append('<li class="list-group-item">' + req.item + '</li>');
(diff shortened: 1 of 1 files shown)


cverna (dismissed):

bowlofeggs:
I am amending my commit to add the Signed-Off-By line, but am not altering the contents.

bowlofeggs:
Looks like the CI job is not getting run because it was deleted:

https://lists.centos.org/pipermail/ci-users/2018-September/000860.html

mattiaverga:
I suppose this will fix #2571 also (?)

bowlofeggs:
@mattiaverga I don't believe so - I think that bug happened in staging when this patch was applied.

PR #2578, opened 2018-09-11
[WIP] Rely on Greenwave to resolve test results from update id.

bowlofeggs:
Greenwave's API docs[0] state that we can hand Greenwave only the
update ID and it will find test results for the Koji builds. Prior
to this commit we had been handing Greenwave the update ID, Koji
builds, and original_nvr fields, which were redundant.

[0] https://docs.pagure.org/greenwave/policies.html#subject-types

### DIFF
--- a/bodhi/server/models.py
+++ b/bodhi/server/models.py
@@ -1967,13 +1967,7 @@ class Update(Base):
             list: A list of dictionaries that are appropriate to be passed to the Greenwave API
                 subject field for a decision about this Update.
         """
-        # See discussion on https://pagure.io/greenwave/issue/34 for why we use these subjects.
-        subject = [{'item': build.nvr, 'type': 'koji_build'} for build in self.builds]
-        # Additionally add some subject entries for "CI"
-        # https://pagure.io/greenwave/issue/61
-        subject.extend([{'original_spec_nvr': build.nvr} for build in self.builds])
-        subject.append({'item': self.alias, 'type': 'bodhi_update'})
-        return subject
+        return [{'item': self.alias, 'type': 'bodhi_update'}]
 
     @property
     def greenwave_subject_json(self):


bodhi/server/models.py
```diff
@@ -1967,13 +1967,7 @@ def greenwave_subject(self):
...
                 subject field for a decision about this Update.
         """
-        # See discussion on https://pagure.io/greenwave/issue/34 for why we use these subjects.
-        subject = [{'item': build.nvr, 'type': 'koji_build'} for build in self.builds]
-        # Additionally add some subject entries for "CI"
-        # https://pagure.io/greenwave/issue/61
-        subject.extend([{'original_spec_nvr': build.nvr} for build in self.builds])
-        subject.append({'item': self.alias, 'type': 'bodhi_update'})
-        return subject
+        return [{'item': self.alias, 'type': 'bodhi_update'}]
```
bowlofeggs, line 1970: @ralphbean We had talked on Friday about adjusting Bodhi to use just the bodhi_update API. Is this what you had in mind? If so, can you confirm that the ```original_spec_nvr``` bit is not needed? The issue that had been cited via comment is not closed, so I'm curious if removing that is OK at this time.

bowlofeggs:
I am a bit hesitant to do this until we figure out https://pagure.io/greenwave/issue/298.

ralphbean:
Ack.  Added some comments over on [the greenwave issue](https://pagure.io/greenwave/issue/298).

bodhi/server/models.py
ralphbean: FWIW, I just checked the greenwave code, and it looks like `original_spec_nvr` is already handled and (assuming the broad gist of this PR is what we want) it shouldn't need to be specified by bodhi.

bowlofeggs:
It sounds like Greenwave has removed the code that resolves update IDs to builds, so I will close this PR.

PR #2581, opened 2018-09-12
s/autodoc_default_flags/autodoc_default_options/

bowlofeggs:
This fixes a docs build failure that emitted this message:

Warning, treated as error:
autodoc_default_flags is now deprecated. Please use
autodoc_default_options instead.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -37,7 +37,7 @@ extensions = [
     'sphinx.ext.todo',
     'sphinx.ext.viewcode']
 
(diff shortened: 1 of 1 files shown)


puiterwijk (dismissed):

cverna (dismissed):

puiterwijk (approved):
`Setting None is equivalent to giving the option name in the list format (i.e. it means “yes/true/on”).`.... Well then.

PR #2584, opened 2018-09-13
Report each CI target separately.

bowlofeggs:


puiterwijk (dismissed):
Let's give it a try

puiterwijk (changes requested):
Let's keep with `job-template` (dash)

puiterwijk (approved):

PR #2600, opened 2018-09-14
Use jenkins-jobs to test the JJB template.

bowlofeggs:


cverna (approved):
LGTM

bowlofeggs:
I got the Jenkies Pipeline working much sooner than I expected, and so the JJB template is deleted now. I thought it would be a lot more work than it ended up being. Sorry for the wasted review effort!

PR #2603, opened 2018-09-16
Allow multiple users filtering for comments, updates and overrides.

### LINKED ISSUE #2489: The API endpoints should be able to take multiple values.
It seems that the API endpoints do not accept multiple values. Thus, the API does not allow for refinement of searches.

For example : 
I would like to receive all comments, but I would like not to receive comment from the user *bodhi*. I can use `https://bodhi.fedoraproject.org/comments/?ignore_user=bodhi` and it works as expected. However, should I require all comments, but skip user *bodhi* and *fedpeg*, I cannot use this query `https://bodhi.fedoraproject.org/comments/?ignore_user=bodhi&ignore_user=fedpeg` nor this one `https://bodhi.fedoraproject.org/comments/?ignore_user=bodhi,fedpeg`.

There should be a way to do it.

mattiaverga:
This will allow multiple users filtering when searching for comments, updates or overrides:
```
comments/?ignore_user=bolowfeggs,mattia
comments/?user=bolowfeggs,mattia
comments/?update_owner=bolowfeggs,mattia
updates/?user=bolowfeggs,mattia
overrides/?user=bolowfeggs,mattia
```
I did not changed the input parameter name to plural (`user` --> `users`) because I think it's better to keep backward compatibility. Is it ok?

Fixes #2489 

### DIFF
--- a/bodhi/server/schemas.py
+++ b/bodhi/server/schemas.py
@@ -90,6 +90,12 @@ class Packages(colander.SequenceSchema):
     package = colander.SchemaNode(colander.String())
 
 
+class Users(colander.SequenceSchema):
+    """A SequenceSchema to validate a list of User objects."""
+
+    user = colander.SchemaNode(colander.String())
+
+
 class Releases(colander.SequenceSchema):
     """A SequenceSchema to validate a list of Release objects."""
 
@@ -653,10 +659,11 @@ class ListUpdateSchema(PaginatedSchema, SearchableSchema, Cosmetics):
         validator=colander.OneOf(list(ContentType.values())),
     )
 
-    user = colander.SchemaNode(
-        colander.String(),
+    user = Users(
+        colander.Sequence(accept_scalar=True),
         location="querystring",
         missing=None,
+        preparer=[util.splitter],
     )
 
     updateid = Builds(
@@ -734,22 +741,25 @@ class ListCommentSchema(PaginatedSchema, SearchableSchema):
         preparer=[util.splitter],
     )
 
-    user = colander.SchemaNode(
-        colander.String(),
+    user = Users(
+        colander.Sequence(accept_scalar=True),
         location="querystring",
         missing=None,
+        preparer=[util.splitter],
     )
 
-    update_owner = colander.SchemaNode(
-        colander.String(),
+    update_owner = Users(
+        colander.Sequence(accept_scalar=True),
         location="querystring",
         missing=None,
+        preparer=[util.splitter],
     )
 
-    ignore_user = colander.SchemaNode(
-        colander.String(),
+    ignore_user = Users(
+        colander.Sequence(accept_scalar=True),
         location="querystring",
         missing=None,
+        preparer=[util.splitter],
     )
 
     anonymous = colander.SchemaNode(
@@ -795,10 +805,11 @@ class ListOverrideSchema(PaginatedSchema, SearchableSchema, Cosmetics):
         preparer=[util.splitter],
     )
 
-    user = colander.SchemaNode(
-        colander.String(),
+    user = Users(
+        colander.Sequence(accept_scalar=True),
         location="querystring",
         missing=None,
+        preparer=[util.splitter],
     )
 
 
--- a/bodhi/server/services/comments.py
+++ b/bodhi/server/services/comments.py
@@ -24,7 +24,7 @@ from cornice import Service
 from cornice.validators import colander_body_validator, colander_querystring_validator
 from pyramid.httpexceptions import HTTPBadRequest
 from sqlalchemy import func, distinct
-from sqlalchemy.sql import or_
+from sqlalchemy.sql import or_, and_
 
 from bodhi.server import log
 from bodhi.server.models import Comment, Build, Update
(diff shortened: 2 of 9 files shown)


bodhi/server/services/comments.py
```diff
@@ -155,15 +155,15 @@ def query_comments(request):
     update_owner = data.get('update_owner')
     if update_owner is not None:
```
Zlopez, line 156: What will happen if more conditions are true like `update_owner` and `ignore_user`? I'm not sure if this is even possible. If not, then fine, but otherwise, there should be test for this.

mattiaverga: In that case the result will display all comments in updates owned by `update_owner` list, made by users that are not `ignore_user` list. It will act in the same manner as now:
https://bodhi.stg.fedoraproject.org/comments/?ignore_user=bodhi&update_owner=mattia
The only thing that will change is that now `ignore_user` and `update_owner` can be made of multiple users.
And I don't see the reason for a test with multiple conditions, since each condition acts like the single one.

The odd thing is that (as now) you can pass both `user` and `ignore_user` conditions to the query, but since this is stupid and doesn't trigger any error (it will return an empty list) I don't think we need to explicitly forbid that:
https://bodhi.stg.fedoraproject.org/comments/?ignore_user=mattia&user=mattia

Zlopez (dismissed):

bodhi/tests/server/services/test_comments.py
```diff
@@ -521,7 +521,7 @@ def test_list_comments_with_unexisting_ignore_user(self):
         self.assertEqual(len(body.get('comments', [])), 0)
         self.assertEqual(res.json_body['errors'][0]['name'], 'ignore_user')
         self.assertEqual(res.json_body['errors'][0]['description'],
-                         "Invalid user specified: santa")
+                         "Invalid users specified: santa")
 
     def test_put_json_comment(self):
         """ We only want to POST comments, not PUT. """
```
bowlofeggs, line 620: Let's add some tests that assert that the ```update_owner```, ```ignore_user```, and ```user``` parameters can accept multiple values.

bodhi/server/schemas.py
```diff
@@ -653,10 +659,11 @@ class ListUpdateSchema(PaginatedSchema, SearchableSchema, Cosmetics):
         validator=colander.OneOf(list(ContentType.values())),
     )
 
-    user = colander.SchemaNode(
-        colander.String(),
+    user = Users(
+        colander.Sequence(accept_scalar=True),
         location="querystring",
         missing=None,
+        preparer=[util.splitter],
```
bowlofeggs, line 666: It is not necessary to support splitting the users by commas, as you can just specify the ```user``` parameter multiple times in a single GET request. For example, here's a query for comments that exclude a bunch of users:

http://localhost:6543/comments/?ignore_user=labbott&ignore_user=elyscape&ignore_user=kevin&ignore_user=bowlofeggs&ignore_user=bodhi

bodhi/tests/server/services/test_overrides.py
```diff
@@ -203,7 +203,7 @@ def test_list_overrides_by_unexisting_username(self):
         self.assertEqual(len(errors), 1)
         self.assertEqual(errors[0]['name'], 'user')
         self.assertEqual(errors[0]['description'],
-                         "Invalid user specified: santa")
+                         "Invalid users specified: santa")
 
     def test_list_overrides_by_like(self):
         """
```
bowlofeggs, line 241: Let's add a test to ensure the multi-user search works here too.

bodhi/tests/server/services/test_updates.py
```diff
@@ -2580,7 +2580,7 @@ def test_list_updates_by_unexisting_username(self):
         self.assertEqual(len(body.get('updates', [])), 0)
         self.assertEqual(res.json_body['errors'][0]['name'], 'user')
         self.assertEqual(res.json_body['errors'][0]['description'],
-                         "Invalid user specified: santa")
+                         "Invalid users specified: santa")
 
     @mock.patch(**mock_uuid4_version1)
     @mock.patch(**mock_valid_requirements)
```
bowlofeggs, line 2614: Let's add a test to ensure the multi-user search works here too.

bowlofeggs (changes requested):
Cool, looks pretty nice! My comment about the commas is optional - I'm not opposed to keeping the split on commas if you want to keep that. I would like to see some tests that exercise this functionality though.

Also, it'd be good to document this in the bodhi CLI man page (see docs/user/man_pages/bodhi.rst).

bodhi/server/schemas.py
mattiaverga: I would prefer a comma separated values list of users for few reasons:
- it's the same behaviour of other filters, like `packages` and `updates`
- it has a better readability (in my opinion): the provided example URL would become http://localhost:6543/comments/?ignore_user=labbott,elyscape,kevin,bowlofeggs,bodhi
- URL queries support both methods of passing a csv list or many `ignore_user` inputs
- command line queries would not support multiple users filtering without further modifications

However, I can comprehend that for native English users it might be not very comfortable specifying a csv list in a `user` (singular) input...

bowlofeggs: Fair enough.

bowlofeggs:
jenkies test

bowlofeggs:
Looks good, thanks for the patch!

A recent commit I made should have fixed up the CI.

bowlofeggs (dismissed):

bowlofeggs:
jenkies test

bowlofeggs (approved):

bowlofeggs:
Mergify seems stuck, so I'm going to manually rebase and merge this one.

PR #2604, opened 2018-09-16
Avoid empty comments without text or feedback

### LINKED ISSUE #2009: empty comment when accidentally hit "add comment & feedback" button
Hi,

The other day, I accidentally hit the "add comment & feedback" button, and it resulted in an empty comment with all karma 0s: https://bodhi.fedoraproject.org/updates/wget-1.19.2-1.fc25#comment-692806

I believe this should be disallowed since it just adds noise. Maybe check if the comments are empty and all karma are 0s, and show a popup dialog to notify when submitting?

mattiaverga:
Fixes #2009 

### DIFF
--- a/bodhi/server/models.py
+++ b/bodhi/server/models.py
@@ -3012,11 +3012,20 @@ class Update(Base):
         if not author:
             raise ValueError('You must provide a comment author')
 
-        caveats = []
-
         # Listify these
         bug_feedback = bug_feedback or []
         testcase_feedback = testcase_feedback or []
+
+        got_feedback = False
+        for feedback_dict in (bug_feedback + testcase_feedback):
+            if feedback_dict['karma'] != 0:
+                got_feedback = True
+                break
+
+        if (not text and not karma and not karma_critpath and not got_feedback):
+            raise ValueError('You must provide either some text or feedback')
+
+        caveats = []
 
         if self.user.name == author:
             if karma != 0:
--- a/bodhi/server/services/comments.py
+++ b/bodhi/server/services/comments.py
@@ -228,6 +228,9 @@ def new_comment(request):
     try:
         comment, caveats = update.comment(
             session=request.db, author=author, anonymous=anonymous, **data)
+    except ValueError as e:
+        request.errors.add('body', 'comment', str(e))
+        return
     except Exception as e:
         log.exception(e)
         request.errors.add('body', 'comment', 'Unable to create comment')
--- a/bodhi/tests/server/services/test_comments.py
+++ b/bodhi/tests/server/services/test_comments.py
@@ -260,6 +260,19 @@ class TestCommentsService(base.BaseTestCase):
         self.assertEqual(res.json_body['errors'][0]['description'],
                          "You must provide an author")
 
+    def test_empty_comment(self):
+        """Ensure that a comment without text or feedback is not permitted."""
+        comment = self.make_comment(text='')
+        comment['bug_feedback.0.bug_id'] = 12345
+        comment['bug_feedback.0.karma'] = 0
+        comment['testcase_feedback.0.testcase_name'] = "Wat"
+        comment['testcase_feedback.0.karma'] = 0
+        res = self.app.post_json('/comments/', comment, status=400)
+
+        self.assertEqual(res.json_body['status'], 'error')
+        self.assertEqual(res.json_body['errors'][0]['description'],
+                         u'You must provide either some text or feedback')
+
     @mock.patch('bodhi.server.services.comments.Update.comment',
                 side_effect=IOError('IOError. oops!'))
     def test_unexpected_exception(self, exception):
--- a/bodhi/tests/server/test_models.py
+++ b/bodhi/tests/server/test_models.py
@@ -3276,6 +3276,13 @@ class TestUpdate(ModelTest):
 
         self.assertEqual(str(exc.exception), 'You must provide a comment author')
 
+    def test_comment_empty(self):
+        """A comment with no text or feedback should raise a ValueError."""
+        with self.assertRaises(ValueError) as exc:
+            self.obj.comment(self.db, '', author=u'bowlofeggs')
+
+        self.assertEqual(str(exc.exception), 'You must provide either some text or feedback')
+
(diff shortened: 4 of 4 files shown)


Zlopez (dismissed):

bowlofeggs (changes requested):
If you use the web interface to add an empty comment to an update, you will see a pop up error like this:

"Comment : Unable to create comment"

The CLI gives a similar experience:

```
[vagrant@bodhi-dev bodhi]$ bodhi updates comment FEDORA-2018-14a1773017 ""
Unable to create comment
```

That's being generated in [```new_comment()```](https://github.com/fedora-infra/bodhi/blob/3.10.0/bodhi/server/services/comments.py#L192-L236):

```
    try:
        comment, caveats = update.comment(
            session=request.db, author=author, anonymous=anonymous, **data)
    except Exception as e:
        log.exception(e)
        request.errors.add('body', 'comment', 'Unable to create comment')
        return
```

Perhaps we should catch the ValueError there and show the user its contents so they can see this "You must provide" message. Let's add a new block to explicitly catch it, because the generic error handler also logs the Exception and we don't want to log server tracebacks every time a user sends an empty comment.

We should also add a test that the REST API gives back the appropriate error message to the end user.

bodhi/server/models.py
```diff
@@ -3008,6 +3008,10 @@ def comment(self, session, text, karma=0, author=None, anonymous=False,
         if not author:
             raise ValueError('You must provide a comment author')
 
+        if (not text and not karma and not karma_critpath and not
+                bug_feedback and not testcase_feedback):
+            raise ValueError('You must provide either some text or feedback')
```
bowlofeggs: Hmm, strangely I found that if you try to enter an empty comment (with no bug feedback) on an Update that is associated with a Bug (for example, http://localhost:6543/updates/FEDORA-2018-576bea0c62) the ```bug_feedback``` here is actually not "falsey" like I would expect, so this exception is not raised.

The update in question already has bug feedback from another user (anonymous) and that appears to be the bug feedback being provided here.

However, this oddity is not due to the change you are making here so I think it does not block this pull request.

I've filed an issue about this at https://github.com/fedora-infra/bodhi/issues/2623.

bowlofeggs (dismissed):

bowlofeggs:
jenkies test

bodhi/server/models.py
mattiaverga: My error, I have been misleaded by the fact bug_feedback and testcase_feedback had None as default value, but in fact they're lists of dictionaries. Moreover, the web UI always posts bug ids and bug karmas, so the object was never None (but only in CLI, so the tests passed).

The error should be solved by the last commit I made, I also set default values to `[]` since both the variables were set like that anyway few lines later.

bodhi/server/models.py
```diff
@@ -2998,7 +2998,7 @@ def obsolete_if_unstable(self, db):
         return
 
     def comment(self, session, text, karma=0, author=None, anonymous=False,
-                karma_critpath=0, bug_feedback=None, testcase_feedback=None,
+                karma_critpath=0, bug_feedback=[], testcase_feedback=[],
```
bowlofeggs: In Python it's problematic to set function parameter defaults to mutable objects, because what you might expect isn't what happens. This bug_feedback list object will be instantiated once in the lifetime of the process and will persist between method calls. This means that if anything in the method ever alters the list, the altered list will be used as the default during the next method call. The technique of using ```None``` is a common way to avoid this problem.

bodhi/server/models.py
```diff
@@ -3008,11 +3008,16 @@ def comment(self, session, text, karma=0, author=None, anonymous=False,
         if not author:
             raise ValueError('You must provide a comment author')
 
-        caveats = []
+        got_feedback = False
+        for feedback_dict in (bug_feedback + testcase_feedback):
+            if feedback_dict['karma'] != 0:
+                got_feedback = True
+                break
```
bowlofeggs, line 3023: IIRC, when I discovered #2623 last week I found that ```bug_feedback``` contains comments that are not part of the current request. I don't believe the code above will solve that particular problem.

bowlofeggs (changes requested):
I suggest reverting back to the commit I had approved last week. Solving #2623 doesn't depend on that commit, and that commit doesn't depend on #2623 so they are better off being separate pull requests.

bodhi/server/models.py
mattiaverga: I have reverted back to the initial commit. 
About the second commit, I've tested it with the example update you suggested and it was working as expected. I will have a second look at that.

mattiaverga: For example, the bug_feedback passed for that update is
`u'bug_feedback': [{u'bug_id': 1618561, u'comment_id': 841956, u'bug': {u'bug_id': 1618561, u'security': False, u'parent': False, u'title': u'libssh_threads.so.4: cannot open shared object file'}, u'karma': 0}]`
It does not contain previous feedbacks

bowlofeggs (approved):

mattiaverga:
I'm still convinced that this PR, without the second commit I've added, is formally wrong, both in the code and in tests.
When posting a comment on an update which has a bug or a testcase attached, `bug_feedback` or `testcase_feedback` are never `None`, so this commit would will not prevent an empty comment to be added. And the test does not reflect correctly this behavior, which is why it passes.
However, if you prefer, I'll do another PR after this commit is merged.

bowlofeggs (changes requested):
If you'd prefer to fix both issues together, that's fine with me. I'm marking it as unapproved so that mergify doesn't merge it.

Please keep the two issues as separate commits.

bowlofeggs:
Cool, this does seem to work. Sorry for all the back and forth :/

I'm going to rebase your patch so it picks up the new CI stuff and approve it. If it passes CI, Mergify should merge it (and CI should be MUCH faster now!)

bowlofeggs (dismissed):

bowlofeggs (dismissed):

bowlofeggs:
Alright, Mergify seems stuck again so I will manually rebase and merge this.

PR #2605, opened 2018-09-17
Add a ComposeState for _wait_for_sync()

### LINKED ISSUE #2550: Add a ComposeState for waiting on the sync to mirrors
We should add a new ComposeState() to get set when ```_wait_for_sync()``` is called in the composer.

mattiaverga:
Fix #2550 

### DIFF
--- a/bodhi/server/migrations/versions/68f9ccb1f388_add_a_syncing_repo_composestate_enum_.py
+++ b/bodhi/server/migrations/versions/68f9ccb1f388_add_a_syncing_repo_composestate_enum_.py
@@ -1,0 +1,54 @@
+# Copyright (c) 2018 Mattia Verga
+#
+# This file is part of Bodhi.
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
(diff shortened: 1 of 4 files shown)


bowlofeggs (changes requested):
Nice! This looks pretty good. The only thing we still need that I can think of is a database migration to extend the enum. There are a couple migrations that do something like this already that you can base your migration off of, for example:

https://github.com/fedora-infra/bodhi/blob/3.10.0/bodhi/server/migrations/versions/2c5d45f0c932_add_a_new_state_for_cleaning_composes.py

You can use alembic to start the template for you like this:

```
$ alembic revision --autogenerate -m "Add a syncing_repo ComposeState enum value."
```

mattiaverga:
I've added the `syncing_repo` value after `punging` instead at the end of the list to maintain the order specified in the docstring (where I added it in the right point of the sequence).
I'm totally new to alembic, but running the upgrade locally returned no errors, so I think it's ok...

mattiaverga:
jenkies test

bodhi/server/migrations/versions/68f9ccb1f388_add_a_syncing_repo_composestate_enum_.py
```diff
@@ -0,0 +1,54 @@
+# Copyright (c) YEAR AUTHOR
```
bowlofeggs: Let's fill in the year and author:

```
# Copyright (c) 2018 Mattia Verga
```

bowlofeggs (changes requested):
Looks great - just need you to claim copyright on your migration and we should be good to merge.

bodhi/server/migrations/versions/68f9ccb1f388_add_a_syncing_repo_composestate_enum_.py
bowlofeggs: This is a minor enough of a change that I'll just force push your branch and fix it for you. Fix incoming!

bowlofeggs (dismissed):

bowlofeggs:
jenkies test

bowlofeggs (approved):

bowlofeggs:
jenkies test

bowlofeggs:
Mergify seems stuck, so I'm going to manually rebase and merge this one.

PR #2616, opened 2018-09-26
Add bodhi-ci, a new CI test running script.

bowlofeggs:
This commit is part of a larger effort to add a true CI test suite
to Bodhi. This new bodhi-ci CLI tool is intended to eventually
replace the devel/run_tests.sh test running script. As so often
happens, what started off as a simple shell script has grown
enough requirements to be too complex to sanely implement as a
shell script. In particular, we are going to need to be able to
specify dependent container URLs (such as Koji or Pagure), with
defaults. These sorts of things are much easier to do with click
than they are with bash.

The new script has a few new features that the old script lacked
as well, many of which would be tricky to add to the old script:

  * It prints out a summary of all the test jobs at the end.
  * It buffers failed jobs' console output and prints those after
    the output from the successful jobs, which means the useful
    error messages will be near the end of the job output instead
    of buried in the middle.
  * It runs the py.test-2 and py.test-3 jobs in parallel instead
    of in series. Users can also select just one of those if
    desired.
  * It allows the caller to select between podman and docker
    (the default) as the container runtime. podman doesn't quite
    work fully yet, but this should help test it going forward.
  * SIGINT is handled and all jobs are stopped before exiting.
  * Full --help text is provided on all commands and flags.
  * Users can opt to skip the build portion of the job.
  * All jobs can be run in parallel. The old script could run
    release tests in parallel, but individual tests on a given
    release were run serially. Now all test are run in parallel.

It is not 100% feature complete with the old script yet, which is
why this commit does not remove the old script. In particular, the
new script does not yet fully support the -x (failfast) flag, or
the -j flag to control how many parallel jobs run. These are
planned for later improvement, and for now I intend to keep the
devel/run_tests.sh script in place since it supports those
features.


bowlofeggs:
CI will certainly fail on this job due to a chicken-and-egg situation - the JJB needs to be edited for CI to pass, but the JJB template isn't used until it is merged.

devel/ci/Dockerfile-pip
```diff
@@ -0,0 +1,74 @@
+FROM registry.fedoraproject.org/fedora:27
```
Zlopez, line 1: Why pip dockerfile is running on F27?

bowlofeggs:
On Wed, 2018-09-26 at 00:55 -0700, Zlopez wrote:
> Why pip dockerfile is running on F27?

This is not a change in this pull request - the pip Dockerfile uses
Fedora 27 in the develop branch as well. That said, production Bodhi
runs on Fedora 27, and Fedora 27 is still officially supported, so I
think it's sensible to keep testing the pip container there. It's
arbitrary mostly because most of the dependencies are pip installed
except for the ones that aren't available via pip.

Zlopez (dismissed):

bowlofeggs:
I'm rebasing and manually merging this.

PR #2617, opened 2018-09-26
Do not expose the Duffy key in CI logs.

bowlofeggs:

### DIFF
--- a/devel/ci/githubprb-project.yml
+++ b/devel/ci/githubprb-project.yml
@@ -60,8 +60,8 @@
     builders:
         - shell: |
             set +e
-            set -x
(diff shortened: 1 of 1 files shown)


bowlofeggs:
This bug was introduced in 49e664057c09c07350fb002fdcf2ec68d6dc7f70.

bowlofeggs:
The pip test failures are due to https://github.com/fedora-infra/bodhi/issues/2618 - we can safely ignore them.

puiterwijk (approved):
This is unfortunate, but it happens.
Nice catching it.

bowlofeggs:
Thanks for the review and kind words!

PR #2625, opened 2018-09-28
Add the {pyver} variable to the JJB unit test job template.

bowlofeggs:
A recent change to the JJB template split the unit test jobs into
two, one for testing Python 2 and one for Python 3. Unfortunately,
I forgot to include the {pyver} variable in the JJB unit test
template and the jobs ended up getting names like "unit tests
(f27-python[2, 3])" instead of "unit tests (f27-python2)", which
is what Mergify expects to find. This commit should fix that.

### DIFF
--- a/devel/ci/githubprb-project.yml
+++ b/devel/ci/githubprb-project.yml
@@ -122,7 +122,7 @@
     <<: *job_template_defaults
 
 - job-template:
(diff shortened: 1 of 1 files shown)


puiterwijk (approved):
Heh, mistakes happen. Let's hope this fixes it! 🚀

Zlopez (approved):

PR #2627, opened 2018-10-01
Do not marked Duffy nodes as failed.

bowlofeggs:
In recent weeks, Bodhi started using far more test jobs per pull
request in order to give more granular test reporting. When jobs
failed, Bodhi's CI system would mark the nodes used as failed,
which gave use 12 hours to log into those systems to poke around.
This was a nice feature, but now that we run 20 tests per pull
requests, I've been informed that we are now causing large numbers
of systems in CentOS CI to wait around for 12 hours in case we want
to debug on them. I rarely use that feature (though it is nice), so
this pull requests adjusts the script such that we always return
the node as successful so we don't exhaust the pool of workers.

### DIFF
--- a/devel/ci/githubprb-project.yml
+++ b/devel/ci/githubprb-project.yml
@@ -98,17 +98,15 @@
             rtn_code=$?
             # Copy the archive results back to the master
             rsync -e "ssh $sshopts" -Ha $CICO_hostname:payload/test_results $(pwd)/
+            cico node done $CICO_ssid;
             if [ $rtn_code -eq 0 ]; then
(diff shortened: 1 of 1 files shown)


bstinsonmhk:
:+1:

puiterwijk (approved):
:+1:

bowlofeggs:
jenkies test

bowlofeggs:
I am going to manually merge this. Mergify won't merge pull requests that have pending results, and there will be pending results in Bodhi until [some JJB jobs I accidentally created are deleted](https://bugs.centos.org/view.php?id=15341).

PR #2634, opened 2018-10-02
Increase the size of the Vagrant guest.

bowlofeggs:
With the recent addition of the bci alias, the Vagrant guest would
benefit from having more RAM and virtual CPUs. bci all will fail
with the current allocation of RAM, and due to the large number of
parallel processes it is beneficial to give the guest more virtual
cores as well.


Zlopez (dismissed):

bowlofeggs:
I'm going to rebase this so the new CI runs on it.

puiterwijk (approved):
ack

PR #2636, opened 2018-10-03
Add a --no-tty flag to bodhi-ci and configure Jenkies to use it.

bowlofeggs:
The console output in Jenkins was "weird" looking because we were
running the Docker commands with its -t flag. This flag is nice
for local development because it gives you colorized output, but
in Jenkins it makes the output hard to read.


Zlopez (approved):

puiterwijk (approved):
pre-acked

PR #2609, opened 2018-09-18
Add a Jenkinsfile to use the CICO pipeline.

bowlofeggs:
This commit adds a Jenkinsfile so that we can use the Jenkins
Pipeline plugin. The intention is to switch to this instead of
Jenkins Job Builder once we can reach feature parity with our JJB
configuration.

Signed-off-by: Randy Barlow <randy@electronsweatshop.com>

PR #2637, opened 2018-10-03
Temporarily disable JJB CI.

bowlofeggs:
I am working on adding a Jenkinsfile[0] to test Bodhi with the
Jenkins Pipeline plugin, and it is difficult to test the new script
when the existing JJB tasks are taking up all the available
workers. This commit temporarily disables the JJB jobs so I can
iterate more rapidly on replacing them.

[0] https://github.com/fedora-infra/bodhi/pull/2609


puiterwijk (approved):
LEt's try it this way.

PR #2638, opened 2018-10-04
Rename the bodhi-ci stages to match their GitHub contexts.

bowlofeggs:
The primary goal here is to get the Python 2 unit and Python 3
unit stages to have different names, but it was easy to just use
the same name we use for the GitHub context.


Zlopez (approved):

puiterwijk (approved):
🖃

PR #2646, opened 2018-10-06
Drop some stray debug print statements from bodhi-ci.

bowlofeggs:


bowlofeggs:
The Rawhide failure is due to https://bugzilla.redhat.com/show_bug.cgi?id=1636611 and can be ignored.

Zlopez (approved):
LGTM

bowlofeggs:
This was actually achieved in https://github.com/fedora-infra/bodhi/pull/2661 because that PR just dropped this entire method.

PR #2648, opened 2018-10-06
Refine javascript code in New Update form

mattiaverga:
I've reworked the javascript code that runs the form in the New Update submission page. This is the list of what I want to achieve with this PR:

- Builds and bugs lists are refreshed every time a new package is selected in the input field. Before, the bugs and builds retrieved were appended to the existing list, which caused duplicated entries if the user selected the same package twice.
- Manual added bugs and builds are not added to the lists if they are already present after having been retrieved from package selection.
- When an error in ajax query occurs it is now displayed as an error message, while before it was displayed as a info message.
- Added a timeout to ajax queries (10 sec)
- Avoid form submit when pressing Enter while entering text in the package text input field

While testing this PR be advised that I noticed a couple of bugs unrelated to these changes (they're already hitting production now): latest_candidates list from bodhi seems to not be working and apps.fedoraproject.org API seems to report packages twice.

### DIFF
--- a/bodhi/server/static/js/update_form.js
+++ b/bodhi/server/static/js/update_form.js
@@ -145,24 +145,24 @@ $(document).ready(function() {
     // fire off two async js calls to get bugs and builds.  Those are then
     // added to their respective checkbox lists once they are retrieved.
     $('#packages-search input.typeahead').on('typeahead:selected', function (e, datum) {
-        $("#candidate-checkboxes").prepend("<img class='spinner' src='static/img/spinner.gif'>")
-        $("#bugs-checkboxes").prepend("<img class='spinner' src='static/img/spinner.gif'>")
         // Get a list of currently checked items
-        var checked_bug_ids = [];
+        var checked_bugs = [];
         $("#bugs-checkboxes input:checkbox:checked").each(function(){
-            checked_bug_ids.push(parseInt($(this).val()));
+            var bug = {id: parseInt($(this).val()), title: $(this).parent().text().replace(/^#\d+\s/m, '')};
+            checked_bugs.push(bug);
         });
         var checked_candidate_ids = [];
         $("#candidate-checkboxes input:checkbox:checked").each(function(){
             checked_candidate_ids.push(parseInt($(this).attr('data-build-id')));
         });
+        // Empty lists
+        document.getElementById("candidate-checkboxes").innerHTML = "<img class='spinner' src='static/img/spinner.gif'>";
+        document.getElementById("bugs-checkboxes").innerHTML = "<img class='spinner' src='static/img/spinner.gif'>";
         // Get the candidate builds
         $.ajax({
             url: 'latest_candidates',
             data: $.param({package: datum.name}),
             success: function(builds) {
-                $("#candidate-checkboxes .spinner").remove();
-                $("#candidate-checkboxes input:checkbox:not(:checked)").parents("div.checkbox").remove();
                 if (builds.length == 0) {
                     return messenger.post({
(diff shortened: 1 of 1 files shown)


bodhi/server/static/js/update_form.js
```diff
@@ -200,9 +205,14 @@ $(document).ready(function() {
...
                 $.each(data.rows, function(i, bug) {
                     // Insert the checkbox only if this ID is not already listed
-                    if ($.inArray(bug.id, checked_bug_ids) == -1) {
-                        add_bug_checkbox(bug.id, bug.description, false);
-                    }
+                    var listed = false;
+                    $.each(checked_bugs, function(i, checked_bug) {
+                        if (bug.id == checked_bug.id) {
+                            listed = true;
+                            return;
```
bowlofeggs: Since you have a ```return``` here, I wonder if we can get away without having the ```listed``` boolean.

bodhi/server/static/js/update_form.js
```diff
@@ -70,22 +70,14 @@ $(document).ready(function() {
...
-            message: 'No bugs found for ' + package,
-            type: 'error',
-        });
-    }
+    // Do not submit form when pressing enter while typing packages
+    // Instead select the first typeahead suggestion
+    $('#packages-search .typeahead').keydown(function(event) {
+        if(event.keyCode == 13) {
+            event.preventDefault();
+            $(".tt-selectable").first().click();
```
bowlofeggs: This leads to a bit of an unexpected behavior if the user uses the arrow keys to scroll down to a different entry and then pressed enter. For example, if I search for "kernel" and then arrow down to something that isn't the first entry and press enter, it searches for the first entry.

bowlofeggs (changes requested):
Nice, just a couple thoughts.

bodhi/server/static/js/update_form.js
mattiaverga: The problem here is that it seems there's no way to break/continue the outer $.each loop, so I need to use a boolean var. The `return` here is not real, no value is returned by the $.each function.
Anyway I realized the `return` statement is wrong, since it translates in `continue`, while I should have used `return false` to break the inner loop here.

mattiaverga:
@bowlofeggs I have some more enhancements for the New Update form, specifically:
- Retrieve bug information for manually entered bugs
- Show a label with product and version (like "Fedora 28" or "Fedora EPEL epel6" for each bug
immagine

This however will require to add also new CORS authorization to headers for bugzilla.redhat.com.

Can I add these enhancements to this PR or do you prefer that I will open a new PR after this one is merged, since you already started a review?

bowlofeggs:
@mattiaverga I'm not sure whether we can get CORS headers added to Bugzilla, or even who to talk to about that. I recommend getting such headers in place before submitting a PR to Bodhi so we don't have to keep a pull request open for a long time.

bowlofeggs:
I'm going to force push to your branch to get it to re-run CI.

bowlofeggs (approved):
Seems to work well for me, thanks!

bowlofeggs:
This fix is planned for inclusion in the upcoming 3.11.0 release: https://github.com/fedora-infra/bodhi/pull/2709

bowlofeggs:
Bodhi 3.11.0b5 has been deployed to staging:

https://bodhi.stg.fedoraproject.org

It is also available via Copr:

mattiaverga:
Most of the javascript stuff is still untestable in staging for the same reason of #2523

bowlofeggs:
This fix is included in Bodhi 3.11.0 which is deployed to Fedora's production instance.

PR #2650, opened 2018-10-06
Add bodhi sar script to gather data about user

### LINKED ISSUE #2553: Add bodhi-sar
I created a script for Fedora that gathers all the data Bodhi knows about a user account and prints it to stdout in JSON format:

https://infrastructure.fedoraproject.org/cgit/ansible.git/tree/roles/bodhi2/backend/files/bodhi_sar.py

It would be best to add this script to Bodhi upstream so that we can test and maintain it together with Bodhi, and so that other Bodhi deployments can respond to GDPR requests.

sebwoj:
Fix #2553

@bowlofeggs, unittests will be next step, but first let me know about your opinion.

### DIFF
--- a/bodhi/server/scripts/sar.py
+++ b/bodhi/server/scripts/sar.py
@@ -1,0 +1,111 @@
+# Copyright (c) 2018 Red Hat, Inc. and Sebastian Wojciechowski
+#
+# This file is part of Bodhi.
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+"""This script will print out SAR data for a FAS account."""
+
+import json
+import sys
+
+import click
+import sqlalchemy
+
+from bodhi.server import config, initialize_db, models
+
+
+def print_human_readable_format(sar_data):
+    """
+    Print user data in human readable format.
+
+    Args:
+        sar_data (dict): User data to be printed.
+    """
+    header_start = "==========>"
+    header_stop = "<=========="
+    chapter_start = "---->"
+    chapter_stop = "<----"
+
+    for user in sar_data:
+        click.echo("{} User account data for: {} {}\n".format(
+            header_start, sar_data[user]['name'], header_stop
+        ))
+        click.echo("email: {}".format(sar_data[user]['email']))
+        click.echo("groups: {}".format(sar_data[user]['groups']))
+        click.echo("popups: {}".format(sar_data[user]['show_popups']))
+
+        click.echo("\n{} Comments: {}".format(chapter_start, chapter_stop))
+        for idx, comment in enumerate(sar_data[user]['comments'], 1):
+            click.echo("\nComment no {}:".format(idx))
+            for item in sorted(comment):
+                click.echo("{}: {}".format(item, comment[item]))
+
+        click.echo("\n{} Updates: {}".format(chapter_start, chapter_stop))
+        for idx, update in enumerate(sar_data[user]['updates'], 1):
+            click.echo("\nUpdate no {}:".format(idx))
+            for item in sorted(update):
+                click.echo("{}: {}".format(item, update[item]))
+
+
+@click.command()
+@click.option('--username', envvar='SAR_USERNAME', required=True,
+              help='The username that sar data should be gathered.')
+@click.option('--human-readable', default=False, is_flag=True,
+              help='Print user data in human readable format.'
+              ' The script defaults to JSON output.'
+              )
+@click.version_option(message='%(version)s')
+def get_user_data(username, human_readable):
+    """Get username SAR data."""
+    initialize_db(config.config)
+
+    sar_data = {}
+
+    try:
+        user = models.User.query.filter_by(name=username).one()
+    except sqlalchemy.orm.exc.NoResultFound:
+        if human_readable:
+            click.echo("User not found.")
+        sys.exit(0)
+
+    sar_data[user.name] = {}
+    sar_data[user.name]['comments'] = [
+        {'karma': c.karma, 'karma_critpath': c.karma_critpath, 'text': c.text,
+         'anonymous': c.anonymous, 'timestamp': c.timestamp.strftime('%Y-%m-%d %H:%M:%S'),
+         'update_alias': c.update.alias, 'username': c.user.name}
+        for c in user.comments]
+    sar_data[user.name]['email'] = user.email
+    sar_data[user.name]['groups'] = [g.name for g in user.groups]
+    sar_data[user.name]['name'] = user.name
+    sar_data[user.name]['show_popups'] = user.show_popups
+    sar_data[user.name]['updates'] = [
+        {'autokarma': u.autokarma, 'stable_karma': u.stable_karma,
+         'unstable_karma': u.unstable_karma, 'requirements': u.requirements,
+         'require_bugs': u.require_bugs, 'require_testcases': u.require_testcases,
+         'notes': u.notes, 'type': str(u.type), 'severity': str(u.severity),
+         'suggest': str(u.suggest), 'close_bugs': u.close_bugs, 'alias': u.alias,
+         'builds': [b.nvr for b in u.builds], 'release_name': u.release.name,
+         'bugs': [b.bug_id for b in u.bugs], 'user': u.user.name,
+         'date_submitted': u.date_submitted.strftime('%Y-%m-%d %H:%M:%S')}
+        for u in user.updates]
+
+    if human_readable:
+        print_human_readable_format(sar_data)
+    else:
+        click.echo(json.dumps(sar_data, sort_keys=True))
+
+
+if __name__ == '__main__':
+    get_user_data()
--- a/bodhi/tests/server/scripts/test_sar.py
+++ b/bodhi/tests/server/scripts/test_sar.py
@@ -1,0 +1,152 @@
+# -*- coding: utf-8 -*-
+# Copyright © 2018 Sebastian Wojciechowski
+#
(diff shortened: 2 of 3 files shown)


bodhi/server/scripts/sar.py
```diff
@@ -0,0 +1,108 @@
+#!/usr/bin/python2
```
bowlofeggs: Since we'll be using setuptools to autogenerate an executable for us, let's not put a shebang here (plus, this will automatically switch to Python 3 with the other scripts when we switch to Python 3, which should happen soon.

bodhi/server/scripts/sar.py
```diff
@@ -0,0 +1,108 @@
...
+# Copyright (c) 2018 Red Hat, Inc.
```
bowlofeggs: Feel free to change this to "Copyright © 2018 Red Hat, Inc. and Sebastian Wojciechowski" - you've made a significant change to it and should get a claim on your change ☺

bodhi/server/scripts/sar.py
```diff
@@ -0,0 +1,108 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+"""This script will print out SAR data for a FAS account."""
+
+import json
+import sys
+import click
```
bowlofeggs: For [PEP-8](https://www.python.org/dev/peps/pep-0008/#imports), ```click``` should be grouped with ```sqlalchemy``` since it's a third party import.

bodhi/server/scripts/sar.py
```diff
@@ -0,0 +1,108 @@
...
+    Args:
+        sar_data (dict): User data to be printed.
+    """
+    header_start = "==========>"
+    header_stop = "<=========="
+    chapter_start = "---->"
+    chapter_stop = "<----"
+
+    for user in sar_data:
+        print("{} User account data for: {} {}\n".format(
```
bowlofeggs: Suggestion: might be nice to use ```click.echo()``` instead of ```print()```, here and elsewhere in the file.

bodhi/server/scripts/sar.py
```diff
@@ -0,0 +1,108 @@
...
+    """Get username SAR data."""
+    initialize_db(config.config)
+
+    sar_data = {}
+
+    try:
+        user = models.User.query.filter_by(name=username).one()
+    except sqlalchemy.orm.exc.NoResultFound:
+        click.echo("User not found.")
+        sys.exit(1)
```
bowlofeggs: This script gets run as part of an all-of-fedora's-apps playbook:

https://infrastructure.fedoraproject.org/cgit/ansible.git/tree/playbooks/manual/gdpr/sar.yml#n22

Due to this, we probably don't want to exit with a non-0 exit code or it will cause that playbook to halt. Also, we should not print the "User not found." message unless we are in ```human_readable``` mode, or the Fedora SAR script will collect that as the SAR data.

bodhi/server/scripts/sar.py
```diff
@@ -0,0 +1,108 @@
...
+            print("\nUpdate no {}:".format(idx))
+            for item in update:
+                print("{}: {}".format(item, update[item]))
+
+
+@click.command()
+@click.option('--username', envvar='SAR_USERNAME', required=True,
+              help='The username that sar data should be gathered.')
+@click.option('--human-readable', default=False, is_flag=True,
+              help='Print user data in human readable format.')
```
bowlofeggs: Maybe add "The script defaults to JSON output."

bowlofeggs (changes requested):
Looks great so far, thanks for taking this on!

bodhi/tests/server/scripts/test_sar.py
```diff
@@ -0,0 +1,150 @@
...
+class TestSar(BaseTestCase):
+    """This class contains tests for the get_user_data() function."""
+
+    def test_invalid_user(self):
+        """Ensure nothing is pirinted when user is not found and human readable is off."""
+        runner = testing.CliRunner()
+        r = runner.invoke(sar.get_user_data, ["--username=" + "invalid_user"])
+
+        self.assertEqual(r.exit_code, 0)
+        self.assertEqual(r.output, "")
```
mattiaverga, line 96: I think this is not standard compliant: it should, at least, return an empty set `{}`.
In my opinion the best option would be a structure like
```
{
  "error": null,
  "result": {<USER_DATA>}
}
```
for successful retrieval, and
```
{
  "error": {
    "message": "User not found."
  },
  "result": null
}
```
for the "user not found" error. In this case the output could also handle database connection error messages.

bowlofeggs: One note: If there is a failure (like a database connection problem), then we should exit non-0. We don't want to exit non-0 for a user not existing, because that will happen during normal operation and we don't want to block the Fedora script, but if there's a database error we do want to cause the calling script to see an error.

bowlofeggs:
I'm going to rebase this on develop so that I can squash the commits and get CI to run again (it should pass on Rawhide again now).

bodhi/tests/server/scripts/test_sar.py
```diff
@@ -0,0 +1,152 @@
+# -*- coding: utf-8 -*-
+# Copyright © 2016-2018 Red Hat, Inc. and others.
```
bowlofeggs:
```suggestion
# Copyright © 2018 Sebastian Wojciechowski
```

bodhi/tests/server/scripts/test_sar.py
```diff
@@ -0,0 +1,152 @@
...
+    '"stable_karma": 3, "suggest": "unspecified", "type": "bugfix", '
+    '"unstable_karma": -3, "user": "guest"}]}}\n'
+)
+
+
+class TestSar(BaseTestCase):
+    """This class contains tests for the get_user_data() function."""
+
+    def test_invalid_user(self):
+        """Ensure nothing is pirinted when user is not found and human readable is off."""
```
bowlofeggs:
```suggestion
        """Ensure nothing is printed when user is not found and human readable is off."""
```

bodhi/tests/server/scripts/test_sar.py
bowlofeggs: For the record, the script that this is intended to replace also prints nothing, so I think this is fine.

bodhi/tests/server/scripts/test_sar.py
```diff
@@ -0,0 +1,152 @@
...
+        runner = testing.CliRunner()
+        r = runner.invoke(sar.get_user_data, ["--username=" + "invalid_user"])
+
+        self.assertEqual(r.exit_code, 0)
+        self.assertEqual(r.output, "")
+
+    def test_invalid_user_human_readable(self):
+        """Ensure proper info is pirinted when user is not found and human readable is on."""
```
bowlofeggs:
```suggestion
        """Ensure proper info is printed when user is not found and human readable is on."""
```

bodhi/tests/server/scripts/test_sar.py
```diff
@@ -0,0 +1,152 @@
...
+        runner = testing.CliRunner()
+        r = runner.invoke(sar.get_user_data, ["--username=" + "invalid_user", "--human-readable"])
+
+        self.assertEqual(r.exit_code, 0)
+        self.assertEqual(r.output, u"User not found.\n")
+
+    def test_valid_user(self):
+        """Ensure json with user data is pirinted when human readable is off."""
```
bowlofeggs:
```suggestion
        """Ensure json with user data is printed when human readable is off."""
```

bowlofeggs (changes requested):

bowlofeggs:
All of the changes I'm suggesting are minor, so I am going to make them for you and force push to your branch. If the tests pass after that, it'll get merged. Thanks!

bowlofeggs:
I'm applying this patch:

```
diff --git a/bodhi/tests/server/scripts/test_sar.py b/bodhi/tests/server/scripts/test_sar.py
index 76a1dd71..4442be08 100644
--- a/bodhi/tests/server/scripts/test_sar.py
+++ b/bodhi/tests/server/scripts/test_sar.py
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-# Copyright © 2016-2018 Red Hat, Inc. and others.
+# Copyright © 2018 Sebastian Wojciechowski
 #
 # This file is part of Bodhi.
 #
@@ -88,7 +88,7 @@ class TestSar(BaseTestCase):
     """This class contains tests for the get_user_data() function."""
 
     def test_invalid_user(self):
-        """Ensure nothing is pirinted when user is not found and human readable is off."""
+        """Ensure nothing is printed when user is not found and human readable is off."""
         runner = testing.CliRunner()
         r = runner.invoke(sar.get_user_data, ["--username=" + "invalid_user"])
 
@@ -96,7 +96,7 @@ class TestSar(BaseTestCase):
         self.assertEqual(r.output, "")
 
     def test_invalid_user_human_readable(self):
-        """Ensure proper info is pirinted when user is not found and human readable is on."""
+        """Ensure proper info is printed when user is not found and human readable is on."""
         runner = testing.CliRunner()
         r = runner.invoke(sar.get_user_data, ["--username=" + "invalid_user", "--human-readable"])
 
@@ -104,7 +104,7 @@ class TestSar(BaseTestCase):
         self.assertEqual(r.output, u"User not found.\n")
 
     def test_valid_user(self):
-        """Ensure json with user data is pirinted when human readable is off."""
+        """Ensure json with user data is printed when human readable is off."""
         now = datetime.utcnow()
         now_str = str(now.strftime("%Y-%m-%d %H:%M:%S"))
         comment = self.db.query(models.Comment).all()[0]
@@ -120,7 +120,7 @@ class TestSar(BaseTestCase):
 
     @mock.patch.dict(os.environ, {"SAR_USERNAME": "guest"})
     def test_valid_user_envvar(self):
-        """Ensure json with user data is pirinted when human readable is off."""
+        """Ensure json with user data is printed when human readable is off."""
         now = datetime.utcnow()
         now_str = str(now.strftime("%Y-%m-%d %H:%M:%S"))
         comment = self.db.query(models.Comment).all()[0]
@@ -135,7 +135,7 @@ class TestSar(BaseTestCase):
         self.assertEqual(r.output, expected_output)
 
     def test_valid_user_human_readable(self):
-        """Ensure user data are pirinted when human readable is on."""
+        """Ensure user data are printed when human readable is on."""
         now = datetime.utcnow()
         now_str = str(now.strftime("%Y-%m-%d %H:%M:%S"))
         comment = self.db.query(models.Comment).all()[0]
```

bowlofeggs (approved):

PR #2651, opened 2018-10-07
Set up a log handler for bodhi-check-policies

### LINKED ISSUE #2156: bodhi-check-policies should set up a log handler to silence an error message
From Freenode:

```
<nirik> bowlofeggs: FYI, No handlers could be found for logger "bodhi.server" is being spewed by bodhi-check-policies. 
<puiterwijk> for that, you should just setup handlers...
<puiterwijk> logging.basicConfig(level=logging.INFO)
<nirik> I sent it to dev/null for now 
<puiterwijk> I guess that works
```

sebwoj:
Fix #2156

### DIFF
--- a/bodhi/server/scripts/check_policies.py
+++ b/bodhi/server/scripts/check_policies.py
@@ -23,9 +23,15 @@ Check the enforced policies by Greenwave for each open update.
 Ideally, this should be done in a fedmsg consumer but we currently do not have any
 messages in the message bus yet.
 """
+import logging
+
 import click
 
 from bodhi.server import config, initialize_db, models, Session
+
+
+logger = logging.getLogger('check-policies')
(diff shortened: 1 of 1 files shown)


bodhi/server/scripts/check_policies.py
```diff
@@ -24,10 +24,15 @@
 messages in the message bus yet.
 """
 import click
+import logging
```
bowlofeggs, line 26: For [PEP-8](https://www.python.org/dev/peps/pep-0008/#imports), ```logging``` should be imported in its own group since it is part of the standard library, and that group should come first.

bowlofeggs (changes requested):
I'll fix up the PEP-8 thing for you and rebase. Thanks for the patch!

bowlofeggs (approved):

PR #2653, opened 2018-10-08
Relax the strictness of two time based assertions.

bowlofeggs:
Sometimes the Duffy nodes we get don't perform well enough to
handle the large number of parallel tests we have without context
switching, and the context switching occasionally fails on a couple
assertions that check timestamps. Technically these aren't great
assertions since they do depend on process scheduling to pass, but
for now let's just try to make them a little less strict by
doubling the range of time they will accept.

### DIFF
--- a/bodhi/tests/server/test_metadata.py
+++ b/bodhi/tests/server/test_metadata.py
@@ -17,7 +17,7 @@
 # with this program; if not, write to the Free Software Foundation, Inc.,
 # 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 
-from datetime import datetime
+from datetime import datetime, timedelta
 from hashlib import sha256
 from os.path import join, exists, basename
 import glob
@@ -132,6 +132,10 @@ class TestAddUpdate(UpdateInfoMetadataTestCase):
 
     def test_date_modified_none(self):
         """The metadata should use utcnow() if an update's date_modified is None."""
+        test_start_time = datetime.utcnow()
+        # The UpdateRecord's updated_date attribute strips microseconds, so let's force them to 0
+        # so our assertions at the end of this test will pass.
+        test_start_time = test_start_time - timedelta(microseconds=test_start_time.microsecond)
         update = self.db.query(Update).one()
         update.date_modified = None
         md = UpdateInfoMetadata(update.release, update.request, self.db, self.temprepo,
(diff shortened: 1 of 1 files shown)


puiterwijk (commented):
Maybe it's an idea to just take a `utcnow()` just before we generate the Metadata, and one after, and make sure the timestamps inserted are in between those values?

````
        self.assertTrue(before < md.uinfo.updates[0].updated_date < after)
````

**Comment by bowlofeggs** [comment]
@puiterwijk

I'm amending my patch with this diff to take your suggestion into account. However, ```createrepo``` trims microseconds from the date, so I had to use a ```<=``` instead of a ```<```:

```
diff --git a/bodhi/tests/server/test_metadata.py b/bodhi/tests/server/test_metadata.py
index 7c901c03..747bb4f5 100644
--- a/bodhi/tests/server/test_metadata.py
+++ b/bodhi/tests/server/test_metadata.py
@@ -17,7 +17,7 @@
 # with this program; if not, write to the Free Software Foundation, Inc.,

-from datetime import datetime
+from datetime import datetime, timedelta
 from hashlib import sha256
 from os.path import join, exists, basename
 import glob
@@ -132,6 +132,10 @@ class TestAddUpdate(UpdateInfoMetadataTestCase):

     def test_date_modified_none(self):
         """The metadata should use utcnow() if an update's date_modified is None."""
+        test_start_time = datetime.utcnow()
+        # The UpdateRecord's updated_date attribute strips microseconds, so let's force them to 0
+        # so our assertions at the end of this test will pass.
+        test_start_time = test_start_time - timedelta(microseconds=test_start_time.microsecond)
         update = self.db.query(Update).one()
         update.date_modified = None
         md = UpdateInfoMetadata(update.release, update.request, self.db, self.temprepo,
@@ -141,10 +145,14 @@ class TestAddUpdate(UpdateInfoMetadataTestCase):

         md.shelf.close()
         self.assertEqual(len(md.uinfo.updates), 1)
-        self.assertTrue(abs((datetime.utcnow() - md.uinfo.updates[0].updated_date).seconds) < 4)
+        self.assertTrue(test_start_time <= md.uinfo.updates[0].updated_date <= datetime.utcnow())

     def test_date_pushed_none(self):
         """The metadata should use utcnow() if an update's date_pushed is None."""
+        test_start_time = datetime.utcnow()
+        # The UpdateRecord's updated_date attribute strips microseconds, so let's force them to 0
+        # so our assertions at the end of this test will pass.
+        test_start_time = test_start_time - timedelta(microseconds=test_start_time.microsecond)
         update = self.db.query(Update).one()
         update.date_pushed = None
         md = UpdateInfoMetadata(update.release, update.request, self.db, self.temprepo,
@@ -154,7 +162,7 @@ class TestAddUpdate(UpdateInfoMetadataTestCase):

         md.shelf.close()
         self.assertEqual(len(md.uinfo.updates), 1)
-        self.assertTrue(abs((datetime.utcnow() - md.uinfo.updates[0].issued_date).seconds) < 4)
+        self.assertTrue(test_start_time <= md.uinfo.updates[0].issued_date <= datetime.utcnow())

     def test_rpm_with_arch(self):
         """Ensure that an RPM with a non 386 arch gets handled correctly."""
```

puiterwijk (dismissed):
:shipit:

bowlofeggs:
I'm going to manually rebase and merge this since Mergify is stuck.

PR #2655, opened 2018-10-08
Collect junit results even when the tests failed.

bowlofeggs:


bodhi/tests/client/test___init__.py
```diff
@@ -79,6 +79,7 @@ def test_url_flag(self, send_request):
         """
         Assert correct behavior with the --url flag.
         """
+        self.fail()
```
bowlofeggs: This is here to test the change; I will remove it before merging.

puiterwijk (approved):
Looks sane to me. 🇫🇰

PR #2661, opened 2018-10-12
Add --failfast/-x and --concurrency/-j flags to bodhi-ci.

bowlofeggs:
It is now possible for bodhi-ci to exit immediately upon failure,
or upon receiving the keyboard interrupt signal.

Additionally, it now supports a ```-j``` flag to specify how many jobs to run in parallel.

These features are done in individual commits.

### DIFF
--- a/devel/ci/bodhi-ci
+++ b/devel/ci/bodhi-ci
@@ -17,18 +17,20 @@
 # along with this program; if not, write to the Free Software
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
 """Bodhi's CI command tool."""
+import asyncio
 import os
 import signal
 import subprocess
 import sys
-import time
+import uuid
 
 import click
 
 
 CONTAINER_NAME = 'bodhi-ci'
-# We label the containers we run so it's easy to find them if the user uses ctrl-c to stop the job.
-CONTAINER_LABEL = 'purpose=bodhi-ci'
+# We label the containers we run so it's easy to find them when we run _stop_all_jobs() at the end.
+# UUID is used so that one bodhi-ci process does not stop jobs started by a different one.
+CONTAINER_LABEL = 'purpose=bodhi-ci-{}'.format(uuid.uuid4())
 PROJECT_PATH = os.path.abspath(os.path.join(os.path.dirname(__file__), '..', '..'))
 RELEASES = ('f27', 'f28', 'f29', 'rawhide', 'pip')
 
@@ -53,19 +55,18 @@ def _set_container_runtime(ctx, param, value):
     return value
 
 
-def _set_tty(ctx, param, value):
+def _set_global(ctx, param, value):
     """
-    Set up the tty variable.
+    Set up a global variable based on click input.
 
     Args:
         ctx (click.core.Context): The Click context, unused.
-        param (click.core.Option): The option being handled. Unused.
-        value (str): The value of the --tty/--no-tty flag.
+        param (click.core.Option): The option being handled. Used to find the global we are setting.
+        value (str): The value of the flag.
     Returns:
-        bool: The value of the --tty/--no-tty flag.
+        bool: The value of the flag.
     """
-    global tty
-    tty = value
+    globals()[param.name] = value
     return value
 
 
@@ -77,7 +78,9 @@ container_runtime_option = click.option(
     help='Select the container runtime to use. Defaults to docker.',
     callback=_set_container_runtime)
 failfast_option = click.option('--failfast', '-x', is_flag=True,
-                               help='Exit immediately upon error.')
+                               help='Exit immediately upon error.', callback=_set_global)
+init_option = click.option('--init/--no-init', default=True,
+                           help="Use the container runtime's --init flag.", callback=_set_global)
 no_build_option = click.option(
     '--no-build', is_flag=True,
     help='Do not run docker build if the image already exists.')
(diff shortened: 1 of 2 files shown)


devel/ci/bodhi-ci
```diff
@@ -302,13 +320,100 @@ def _ensure_builds_exist(releases):
...
+                    # This was canceled, so let's print it now so the error output can go
+                    # last.
+                    click.echo(result['stdout'])
+                else:
+                    error_output = '{}\n{}'.format(error_output, result['stdout'])
+            if not returncode:
+                returncode = result['returncode']
+        summary = summary + result['summary']
+
+    # Let's sort the summary lexigraphiaclly so releases show near each other.
```
Zlopez: lexigraphiaclly -> lexicographically

devel/ci/bodhi-ci
```diff
@@ -346,6 +451,64 @@ def _generate_unit_command_maps(failfast, pyvers):
     return command_maps
 
 
+def _handle_sigint():
+    """
+    Catch SIGINT, but don't do anything.
+
+    We just don't want the signal to cause the Exception to bubble up out of the main loop.
+    """
+
```
Zlopez: Maybe add `pass` here to made it more readable.

Zlopez (commented):
I only found these two small things

bowlofeggs:
Hi @Zlopez,

I will apply this patch to address your comments:

```
diff --git a/devel/ci/bodhi-ci b/devel/ci/bodhi-ci
index 1018929e..68c78989 100755
--- a/devel/ci/bodhi-ci
+++ b/devel/ci/bodhi-ci
@@ -383,7 +383,7 @@ def _process_results(loop, done, pending):
                 returncode = result['returncode']
         summary = summary + result['summary']

-    # Let's sort the summary lexigraphiaclly so releases show near each other.
+    # Let's sort the summary lexicographically so releases show near each other.
     summary = '\n'.join(sorted([l for l in summary.split('\n')]))

     return {'summary': summary, 'output': error_output, 'returncode': returncode}
@@ -451,14 +451,6 @@ def _generate_unit_command_maps(pyvers):
     return command_maps


-def _handle_sigint():
-    """
-    Catch SIGINT, but don't do anything.
-
-    We just don't want the signal to cause the Exception to bubble up out of the main loop.
-    """
-
-
 async def _run_process(job, *args, **kwargs):
     """
     Run a subprocess, returning its label, summary, return code, and stdout.
@@ -539,7 +531,9 @@ def _run_processes(jobs):
     if failfast:
         return_when = asyncio.FIRST_EXCEPTION
     future = asyncio.wait(processes, return_when=return_when)
-    loop.add_signal_handler(signal.SIGINT, _handle_sigint)
+    # Catch SIGINT, but don't do anything. We just don't want the signal to cause the Exception to
+    # bubble up out of the main loop.
+    loop.add_signal_handler(signal.SIGINT, lambda: None)

     try:
         done, pending = loop.run_until_complete(future)
```

bowlofeggs:
The Rawhide failure is due to https://bugzilla.redhat.com/show_bug.cgi?id=1637711, and the fix has not yet propagated to the mirrors. The F29 failure is fixed in https://github.com/fedora-infra/bodhi/pull/2668 and is not due to this pull request.

devel/ci/bodhi-ci
```diff
@@ -17,18 +17,20 @@
...
+import uuid
 
 import click
 
 
 CONTAINER_NAME = 'bodhi-ci'
-# We label the containers we run so it's easy to find them if the user uses ctrl-c to stop the job.
-CONTAINER_LABEL = 'purpose=bodhi-ci'
+# We label the containers we run so it's easy to find them when we run _stop_all_jobs() at the end.
+# UUID is used so that one bodhi-ci process does not stop jobs started by a different one.
```
puiterwijk, line 33: Good idea!

devel/ci/bodhi-ci
```diff
@@ -346,6 +451,56 @@ def _generate_unit_command_maps(failfast, pyvers):
...
+
+    try:
+        stdout, stderr = await process.communicate()
+    except asyncio.CancelledError as e:
+        try:
+            process.terminate()
+        except ProcessLookupError:
+            # The process is already stopped, nothing to see here.
+            pass
+        stdout, stderr = await process.communicate()
```
puiterwijk: This would toss any output captured before it's cancelled?

puiterwijk (dismissed):
One small comment, but 👍  further.

devel/ci/bodhi-ci
bowlofeggs: heh

devel/ci/bodhi-ci
bowlofeggs: Ugh, I actually had that fixed in an earlier version of this but this went through like 10 revisions last week and my brain lost track of everything. Easy fix.

bowlofeggs: I'm adding this patch:

```
diff --git a/devel/ci/bodhi-ci b/devel/ci/bodhi-ci
index 68c78989..6d6ef6b0 100755
--- a/devel/ci/bodhi-ci
+++ b/devel/ci/bodhi-ci
@@ -470,6 +470,7 @@ async def _run_process(job, *args, **kwargs):
             stdout (str): The process' stdout.
     """
     cancelled = False
+    stdout = ''

     process = await asyncio.create_subprocess_exec(*args, **kwargs)

@@ -481,7 +482,8 @@ async def _run_process(job, *args, **kwargs):
         except ProcessLookupError:
             # The process is already stopped, nothing to see here.
             pass
-        stdout, stderr = await process.communicate()
+        cancelled_stdout, stderr = await process.communicate()
+        stdout = stdout + cancelled_stdout
         cancelled = True

     returncode = -2 if cancelled else process.returncode
```

bowlofeggs: Note: The stderr is going to be empty always, because we redirect stderr to stdout when we setup the process.

bowlofeggs: Aand one more patch where I changed ```stdout = ''``` to ```stdout = b''``` so that it doesn't fail later when it gets concatenated with ```cancelled_stdout```.

bowlofeggs:
I've added a second commit that gives it a ```-j``` flag to control concurrency.

puiterwijk (approved):
+1

Zlopez (approved):
+1

Issue #1957, opened 2017-10-30
The CLI should support paging


bowlofeggs:
I noticed today that the CLI does not support paging. If I want to see more of the updates that match this query, there isn't a way to do that:

```
$ bodhi updates query --request batched
httpd-2.4.29-1.fc26                   rpm    enhancement  testing   2017-10-27
libreoffice-5.3.7.2-1.fc26            rpm    bugfix       testing   2017-10-27
gdm-3.26.1-2.fc27                     rpm    bugfix       pending   2017-10-26
systemd-233-7.fc26                    rpm    security     testing   2017-10-26
java-1.8.0-openjdk-1.8.0.151-1.b12.fc25 rpm    security     testing   2017-10-26
java-1.8.0-openjdk-1.8.0.151-1.b12.fc26 rpm    security     testing   2017-10-26
java-1.8.0-openjdk-1.8.0.151-1.b12.fc27 rpm    security     testing   2017-10-27
libinput-1.9.0-1.fc26                 rpm    enhancement  testing   2017-10-26
python-mccabe-0.6.1-6.fc27            rpm    enhancement  pending   2017-10-25
gtk3-3.22.21-2.fc26                   rpm    bugfix       testing   2017-10-25
samba-4.6.9-0.fc26                    rpm    bugfix       testing   2017-10-25
rsyslog-8.30.0-3.fc26                 rpm    bugfix       testing   2017-10-25
python-mccabe-0.6.1-5.fc27            rpm    enhancement  testing   2017-10-24
js-responsive-elements-1.0.2-1.fc27   rpm    newpackage   testing   2017-10-25
grub2-2.02-19.fc27                    rpm    enhancement  testing   2017-10-24
gnome-software-3.22.7-4.fc25          rpm    bugfix       testing   2017-10-26
gnome-software-3.24.3-3.fc26          rpm    bugfix       testing   2017-10-24
SDL2-2.0.7-1.fc27                     rpm    enhancement  testing   2017-10-24
SDL2-2.0.7-1.fc26                     rpm    security     testing   2017-10-24
gnome-boxes-3.26.1-4.fc27             rpm    bugfix       testing   2017-10-24
263 updates found (20 shown)
```

mcepl:
+1 … either should bodhi behave as a good Unix tool and just spit out results and not care about consequences (and user would have to use ``less(1)`` or something), or if it does care it should make it possible to display all results.

sebwoj:
[...]

PR #2665, opened 2018-10-14
Add pagination to bodhi releases list command

sebwoj:


Zlopez (dismissed):
LGTM

sebwoj:
Fix #1957

bodhi/client/__init__.py
```diff
@@ -1060,12 +1060,17 @@ def info_release(name, url, **kwargs):
...
 @url_option
 @debug_option
+@add_options(pagination_options)
 @staging_option
-def list_releases(display_archived, url, **kwargs):
+def list_releases(display_archived, url, rows=None, page=None, **kwargs):
     """Retrieve and print list of releases."""
+    if not display_archived and rows is not None:
+        click.echo("ERROR: must set --display_archived to use pagination")
+        sys.exit(1)
```
bowlofeggs: I'm not sure we should block the user from doing this. Sure, it's unnecessary to page when there are so few options, but I don't think we need to stop them from doing it.

bowlofeggs (changes requested):

bodhi/client/__init__.py
sebwoj: bodhi cli omits archived releases if flag is display_archived is disabled, so when we type for instance  rows=2 and page=1, nothing is printed. The reason is that filtering releases state is done after request is  received(in cli side). So should I remove mentioned guard?

bowlofeggs: Ah, I see now. I hadn't realized the archived filtering happens client side. That is pretty strange. It seems like this whole thing would work better if that flag just queried the server for releases based on whether they are archived or not, rather than filtering them client side.

bodhi/server/services/releases.py
```diff
@@ -260,6 +260,10 @@ def query_releases_json(request):
         query = query.join(Release.builds).join(Build.package)
         query = query.filter(or_(*[Package.id == p.id for p in packages]))
 
+    exclude_archived = data.get('exclude_archived')
+    if exclude_archived is not None and exclude_archived:
```
bowlofeggs: You could simplify this:

```suggestion
    if exclude_archived:
```

bodhi/tests/server/services/test_releases.py
```diff
@@ -226,6 +226,32 @@ def test_list_releases_with_a_wrong_state(self):
...
+    def test_list_releases_exclude_archived_on(self):
+        """ Test that we can filter releases using the exclude_archived flag"""
+        r = Release.query.filter_by(name='F17').first()
+        r.state = ReleaseState.archived
+        r = Release.query.filter_by(name='F22').first()
+        r.state = ReleaseState.current
+        self.db.flush()
+        res = self.app.get('/releases/', {"exclude_archived": True})
+        body = res.json_body
+        self.assertEqual(len(body['releases']), 1)
```
bowlofeggs: We can make this assertion a little stronger:

```suggestion
        self.assertEqual([r['name'] for r in body['releases']], ['F22'])
```

bodhi/tests/server/services/test_releases.py
```diff
@@ -226,6 +226,32 @@ def test_list_releases_with_a_wrong_state(self):
...
+        Test that "archived" releases are included in response when the exclude_archived is off
+        """
+        r = Release.query.filter_by(name='F17').first()
+        r.state = ReleaseState.archived
+        r = Release.query.filter_by(name='F22').first()
+        r.state = ReleaseState.current
+        self.db.flush()
+        res = self.app.get('/releases/')
+        body = res.json_body
+        self.assertEqual(len(body['releases']), 2)
```
bowlofeggs, line 257: Similarly here:

```suggestion
        self.assertEqual({r['name'] for r in body['releases']}, {'F17', 'F22'})
```

bowlofeggs (changes requested):
Cool looks great. I just have a couple minor tweaks, so I will make them myself and force push to your branch.

bowlofeggs (approved):

Issue #1542, opened 2017-05-22
Bodhi cannot create updates with ~300 builds in them


bowlofeggs:
Today a user tried to create an update with about 300ish builds and PostgreSQL was not happy about the size of the index for the title field:

```
2017-05-22 19:18:40,485 ERROR [bodhi.server][Dummy-1] Failed to create update
Traceback (most recent call last):
  File "/usr/lib/python2.7/site-packages/bodhi/server/services/updates.py", line 412, in new_update
    result, _caveats = Update.new(request, _data)
  File "/usr/lib/python2.7/site-packages/bodhi/server/models.py", line 915, in new 
    up.set_request(db, req, request.user.name)
  File "/usr/lib/python2.7/site-packages/bodhi/server/models.py", line 1340, in set_request
    action.description, username, notes), author=u'bodhi')
  File "/usr/lib/python2.7/site-packages/bodhi/server/models.py", line 1640, in comment
    user = session.query(User).filter_by(name=author).one()
  File "/usr/lib64/python2.7/site-packages/sqlalchemy/orm/query.py", line 2395, in one 
    ret = list(self)
  File "/usr/lib64/python2.7/site-packages/sqlalchemy/orm/query.py", line 2437, in __iter__
    self.session._autoflush()
  File "/usr/lib64/python2.7/site-packages/sqlalchemy/orm/session.py", line 1208, in _autoflush
    util.raise_from_cause(e)
  File "/usr/lib64/python2.7/site-packages/sqlalchemy/util/compat.py", line 199, in raise_from_cause
    reraise(type(exception), exception, tb=exc_tb)
  File "/usr/lib64/python2.7/site-packages/sqlalchemy/orm/session.py", line 1198, in _autoflush
    self.flush()
  File "/usr/lib64/python2.7/site-packages/sqlalchemy/orm/session.py", line 1919, in flush
    self._flush(objects)
  File "/usr/lib64/python2.7/site-packages/sqlalchemy/orm/session.py", line 2037, in _flush
    transaction.rollback(_capture_exception=True)
  File "/usr/lib64/python2.7/site-packages/sqlalchemy/util/langhelpers.py", line 60, in __exit__
    compat.reraise(exc_type, exc_value, exc_tb)
  File "/usr/lib64/python2.7/site-packages/sqlalchemy/orm/session.py", line 2001, in _flush
    flush_context.execute()
[...]

PR #2666, opened 2018-10-14
Add option to query according to update title

### LINKED ISSUE #251: add option to query according to update title
If I know the update title, there seems to be no way to retrieve the corresponding update object. There is no `query(title=)` or similar. Can you please implement this?

From IRC:

```
<threebean> kparal: can you file an RFE to add a 'title' to the ListUpdatesSchema ?
```

And a follow-up:

```
<lmacken> kparal: like=title should work?
<lmacken> but we should have a proper title field
<kparal> lmacken: that works!
```

python-fedora 0.5.5

sebwoj:
Fix #251

### DIFF
--- a/bodhi/tests/client/test_bindings.py
+++ b/bodhi/tests/client/test_bindings.py
@@ -817,6 +817,19 @@ class TestBodhiClient_query(unittest.TestCase):
         client.send_request.assert_called_once_with(
             'updates/', verb='GET', params={'packages': 'bodhi', 'page': 5})
 
+    def test_with_title(self):
+        """
+        Test with the 'title' kwarg.
+        """
+        client = bindings.BodhiClient()
+        client.send_request = mock.MagicMock(return_value='return_value')
+
+        result = client.query(title='nest-2.16.0-1.fc28')
+
+        self.assertEqual(result, 'return_value')
+        client.send_request.assert_called_once_with(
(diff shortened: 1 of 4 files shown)


bodhi/tests/client/test_bindings.py
```diff
@@ -817,6 +817,19 @@ def test_with_page(self):
         client.send_request.assert_called_once_with(
             'updates/', verb='GET', params={'packages': 'bodhi', 'page': 5})
 
+    def test_with_titile(self):
```
Zlopez: typo titile -> title

Zlopez (commented):
Only found a small typo

bowlofeggs:
Two notes:

0. This actually does not fix #251 because that issue is about adding a title search to the REST API, not just to the Python bindings.
1. We are planning to drop the title field anyway in order to fix #1542, so we don't really need to fix #251.

We will likely delete the title field soon (in the next few months) so this patch won't be in Bodhi for very long, however, it does work so I see no reason not to accept it. I am going to rebase,d squash your commits, and then merge. Thanks for the patch!

bowlofeggs (approved):

bowlofeggs:
Mergify seems stuck again, so I will manually merge this.

bowlofeggs:

bowlofeggs:

PR #2667, opened 2018-10-15
Be technically correct when logging about pungi calls

puiterwijk:
We actually call pungi as a subprocess rather than a thread, so this
will make the logs technically correct: the Best Kind of Correct(R).

### DIFF
--- a/bodhi/server/consumers/masher.py
+++ b/bodhi/server/consumers/masher.py
@@ -1218,9 +1218,9 @@ class PungiComposerThread(ComposerThread):
         """
         self.save_state(ComposeState.punging)
         if pungi_process is None:
-            self.log.info('Not waiting for pungi thread, as there was no pungi')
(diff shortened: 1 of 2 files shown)


bowlofeggs (dismissed):
Thanks!

bowlofeggs:
I'm going to rebase your patch so the new required tests run.

bowlofeggs (approved):

bowlofeggs:
Mergify seems stuck, so I will manually merge this.

bowlofeggs:
This fix is planned for inclusion in the upcoming 3.11.0 release: https://github.com/fedora-infra/bodhi/pull/2709

bowlofeggs:
Bodhi 3.11.0b5 has been deployed to staging:

https://bodhi.stg.fedoraproject.org

It is also available via Copr:

bowlofeggs:
This fix is included in Bodhi 3.11.0 which is deployed to Fedora's production instance.

PR #2668, opened 2018-10-15
Explicitly install python3-librepo in CI tests.

bowlofeggs:


puiterwijk (approved):
LGTM

bowlofeggs:
The CI failures are all due to an I/O failure on ```cdn.registry.fedoraproject.org```. However, I ran them locally and they pass, and this is urgent since it is causing tests to fail, so I am going to manually merge.

PR #2669, opened 2018-10-15
Install click from EPEL in CI instead of from PyPI.

bowlofeggs:


bowlofeggs:
I'm marking this WIP until https://bodhi.fedoraproject.org/updates/FEDORA-EPEL-2018-6ee6690872 goes stable. However, I wanted to test it out with ```bodhi-ci``` before it went stable, which is why I opened this pull request.

puiterwijk (approved):

PR #2670, opened 2018-10-16
Update bodhi-push to only resume failed composes

### LINKED ISSUE #2579: bodhi-push will happily resume Composes that are not in a stopped state
```bodhi-push``` does not prevent or warn users when they ask it to resume a compose that isn't in a stopped state (which I think would only include the ```failed``` state). This is bad because the backend will happily run two simultaneous composes if asked to do so.

We should make it error if the compose isn't in a stopped state, or at least warn. I lean towards erroring and refusing, but there are situations (such as fedmsg-hub stopping) where composes could get left in what appears to be a running state even when they are not running, and it would be nice if there were a way to force it to start over.

### PR DESCRIPTION
**Pull request by ryanlerch**
This updates bodhi-push so it only resumes failed composes. This, however does not allow the user to restart other composes that might be failed.

Any ideas on how to re-structure this to allow both use-cases?

Fixes #2579

### DIFF
--- a/bodhi/server/push.py
+++ b/bodhi/server/push.py
@@ -71,15 +71,15 @@ def push(username, cert_prefix, **kwargs):
     db_factory = transactional_session_maker()
     composes = []
     with db_factory() as session:
-        if not resume and session.query(Compose).count():
-            click.confirm('Existing composes detected: {}. Do you wish to resume them all?'.format(
-                ', '.join([str(c) for c in session.query(Compose).all()])), abort=True)
+        if not resume and session.query(Compose).filter(Compose.state == ComposeState.failed).count():
+            click.confirm('Failed composes detected: {}. Do you wish to resume them all?'.format(
+                ', '.join([str(c) for c in session.query(Compose).filter(Compose.state == ComposeState.failed).all()])), abort=True)
             resume = True
             resume_all = True
 
         # If we're resuming a push
         if resume:
-            for compose in session.query(Compose).all():
+            for compose in session.query(Compose).filter(Compose.state == ComposeState.failed).all():
                 if len(compose.updates) == 0:
                     # Compose objects can end up with 0 updates in them if the masher ejects all the
                     # updates in a compose for some reason. Composes with no updates cannot be

### CODE REVIEW

File: bodhi/server/push.py [hunk 1]
```diff
@@ -71,15 +71,15 @@ def push(username, cert_prefix, **kwargs):
     db_factory = transactional_session_maker()
     composes = []
     with db_factory() as session:
-        if not resume and session.query(Compose).count():
-            click.confirm('Existing composes detected: {}. Do you wish to resume them all?'.format(
-                ', '.join([str(c) for c in session.query(Compose).all()])), abort=True)
+        if not resume and session.query(Compose).filter(Compose.state == ComposeState.failed).count():
+            click.confirm('Failed composes detected: {}. Do you wish to resume them all?'.format(
+                ', '.join([str(c) for c in session.query(Compose).filter(Compose.state == ComposeState.failed).all()])), abort=True)
```
bowlofeggs, line 76: I would probably make the query match non-failed composes instead of detecting failed ones. We could then print a warning and exit, something like "WARNING: running composes detected, refusing to resume as that is a dangerous operation. If you are sure they are not really running, you can re-run with the --force flag." Of course, this would require a ```--force``` flag to be added as well.

bodhi/server/push.py
```diff
@@ -71,15 +71,15 @@ def push(username, cert_prefix, **kwargs):
...
             resume = True
             resume_all = True
 
         # If we're resuming a push
         if resume:
-            for compose in session.query(Compose).all():
+            for compose in session.query(Compose).filter(Compose.state == ComposeState.failed).all():
```
bowlofeggs, line 82: If we add a ```--force``` flag, we would want to respect that here as well.

bowlofeggs (changes requested):
Can you amend your commit message to sign the Developer Certificate of Origin as described [here](https://bodhi.fedoraproject.org/docs/developer/index.html#contribution-guidelines)?

bowlofeggs:
Hey @ryanlerch! Do you still want to work on this?

bowlofeggs:
Closing due to inactivity.

PR #2671, opened 2018-10-17
Allow bodhi-ci tests to wait on builds.

bowlofeggs:
Prior to this commit, bodhi-ci would run all the builds in
parallel, but would wait for all the builds to finish before
starting any of the tests that depended on them. This was
inefficient, and could also be annoying when certain builds fail
for reasons beyond our control (as often happens with Rawhide).

Now it is able to mark test jobs as depending on other test jobs,
so that they can wait for their dependent job to finish before
starting.

### DIFF
--- a/devel/ci/bodhi-ci
+++ b/devel/ci/bodhi-ci
@@ -18,6 +18,7 @@
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
 """Bodhi's CI command tool."""
 import asyncio
+import functools
 import multiprocessing
 import os
 import signal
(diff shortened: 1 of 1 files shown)


bowlofeggs:
Hmm, looks like maybe I broke the ```-a``` flag. I'll fix that tomorrow.

bowlofeggs:
Heh, well it would help if I had actually wired it up to something… easy fix, but I'm tired now.

pypingou:
I'm not very familiar with this code but I didn't see anything that stood out for me. So this is a :+1: for me if you accept these premises.

It did make me wonder how much of this is very bodhi-specific (ie: how much of this could be re-used by other apps in our portfolio :))

PR #2672, opened 2018-10-17
Remove devel/run_tests.sh.

bowlofeggs:
The new bodhi-ci tool has gained all of the abilities of
run_tests.sh, and so it is no longer needed.


puiterwijk (approved):
Cleanup!

cverna (approved):

PR #2673, opened 2018-10-17
podman does not have --init.

bowlofeggs:

### DIFF
--- a/devel/ci/bodhi-ci
+++ b/devel/ci/bodhi-ci
@@ -107,8 +107,10 @@ container_runtime_option = click.option(
     callback=_set_container_runtime)
 failfast_option = click.option('--failfast', '-x', is_flag=True,
                                help='Exit immediately upon error.', callback=_set_global)
-init_option = click.option('--init/--no-init', default=True,
-                           help="Use the container runtime's --init flag.", callback=_set_global)
+init_option = click.option(
+    '--init/--no-init', default=True, callback=_set_global,
+    help=("Use docker's --init flag. This option only has an effect when docker is "
+          "the container runtime being used."))
 no_build_option = click.option(
     '--no-build', is_flag=True, callback=_set_global,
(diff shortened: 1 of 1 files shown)


puiterwijk (changes requested):
Maybe, instead of silently ignoring it, error out if the init and podman combo is used?

bowlofeggs:
```--init``` is the default, so if we did that then the user would have to type ```--no-init``` when using podman, which is a bit burdensome.  The only reason the ```--init``` flag exists as a flag at all is that it fails on CentOS because the flag doesn't exist in that version of Docker, but I want to use it when developing locally because it helps the jobs cancel more quickly when using the ```-x``` flag or ```ctrl-c```, and I want that to be the default when humans are using it. There really is no reason a human should want to use the ```--init``` or ```--no-init``` flag - it really just exists so the ```Jenkiesfile``` can set ```--no-init```.

Anyways, all of that to say, I think I'd prefer not to make the human type out ```--no-init``` when using ```podman```. Would one of these alternatives be amenable to you:

* Add a note to the ```--init``` helptext that explicitly states that it only does anything when ```docker``` is used. I prefer this option over the other two.
* Print a warning when ```--init``` is set and ```podman``` is being used that says that ```--init``` is being ignored. I don't love this option because it is noisy, but it would be easy.
* I guess I could try to detect the version of the container runtime and selectively use ```--init``` when I can. This could be cool, though probably a bit more work and seems overkill since the ```--init``` flag isn't intended to be used by humans anyway.

bowlofeggs:
I chatted with @puiterwijk and he was willing to approve if I implement the first option I mentioned above, so I will make it so.

bowlofeggs:
I believe that https://github.com/fedora-infra/bodhi/pull/2684 will fix the CI failure, but it is def something we can ignore here.

puiterwijk (dismissed):
Looks good to me

PR #2674, opened 2018-10-18
Bring diff-cover testing back.

bowlofeggs:
During all the recent CI refactoring, I somehow forgot to test for
diff coverage and we recently had a patch get through that dropped
coverage. This commit brings it back to bodhi-ci and to our
Jenkiesfile.


pypingou:
Looks fine to me

cverna (approved):
+1

Issue #2317, opened 2018-04-26
bodhi-push should not select Rawhide builds


bowlofeggs:
When we place Bodhi over Rawhide, we do not intend to have Bodhi compose Rawhide, at least not initially. As a result, we need to give the ```Release``` model a boolean to indicate whether the release is composed by Bodhi or not so that the Rawhide release can be marked as not composed by Bodhi. Then the ```bodhi-push``` tool should exclude releases that indicate that they are not composed by Bodhi.

bowlofeggs:
This feature is planned for inclusion in Bodhi 3.13: #2917.

bowlofeggs:
This feature is now deployed to staging: https://bodhi.stg.fedoraproject.org/docs/user/release_notes.html

It is also available via Copr: https://copr.fedorainfracloud.org/coprs/bowlofeggs/bodhi-pre-release/

bowlofeggs:
Bodhi 3.13.0 has been released:

https://github.com/fedora-infra/bodhi/releases/tag/3.13.0

Fedora Infrastructure plans to deploy it to production on Monday.

PR #2681, opened 2018-10-21
Exclude releases not composed by bodhi in bodhi push

sebwoj:


bowlofeggs:
I'm not sure I follow the motivation here. When would Bodhi know about Composes that weren't created by Bodhi?

sebwoj:
I forgot to mention related issue #2317

bowlofeggs:
@sebwoj Ah, I see! So that is part of a much larger project: https://github.com/fedora-infra/bodhi/projects/3

That card is about half way down the todo list, which is roughly sorted in order of implementation, so it's a bit early to add this. However, it doesn't really hurt anything to add it now so +1.

In order to make this usable, we should make it so that ```bodhi releases edit/new``` can alter this boolean, otherwise it's a value that can't be set or used.

bodhi/client/__init__.py
```diff
@@ -164,6 +164,8 @@ def _set_logging_debug(ctx, param, value):
                                                'archived']),
                  help='The state of the release'),
     click.option('--mail-template', help='Name of the email template for this release'),
+    click.option('--not_composed_by_bodhi', is_flag=True, default=False,
```
bowlofeggs:
```suggestion
    click.option('--not-composed-by-bodhi', is_flag=True, default=False,
```

bowlofeggs: Also, it'ls possible to get click to name this ```composed_by_bodhi``` and to allow the caller to pass an explicit flag to enable it, like this:

```suggestion
    click.option('--composed-by-bodhi/--not-composed-by-bodhi', is_flag=True, default=True,
```

bodhi/client/__init__.py
```diff
@@ -1009,11 +1011,12 @@ def releases():
 @releases.command(name='create')
 @handle_errors
 @add_options(release_options)
-def create_release(user, password, url, debug, **kwargs):
+def create_release(user, password, url, debug, not_composed_by_bodhi, **kwargs):
     """Create a release."""
     client = bindings.BodhiClient(base_url=url, username=user, password=password,
                                   staging=kwargs['staging'])
     kwargs['csrf_token'] = client.csrf()
+    kwargs['composed_by_bodhi'] = not not_composed_by_bodhi
```
bowlofeggs: If you like my suggestion above, you can change this line to be a little more readable:

```suggestion
    kwargs['composed_by_bodhi'] = composed_by_bodhi
```

bodhi/client/__init__.py
```diff
@@ -1009,11 +1011,12 @@ def releases():
 @releases.command(name='create')
 @handle_errors
 @add_options(release_options)
-def create_release(user, password, url, debug, **kwargs):
+def create_release(user, password, url, debug, not_composed_by_bodhi, **kwargs):
```
bowlofeggs: If you like my suggestion above, this becomes:

```suggestion
def create_release(user, password, url, debug, composed_by_bodhi, **kwargs):
```

bodhi/server/migrations/versions/cf7257bfa910_add_composed_by_bodhi_flag_to_releases_.py
```diff
@@ -0,0 +1,43 @@
+# Copyright (c) 2018 Red Hat, Inc.
```
bowlofeggs:
```suggestion
# Copyright (c) 2018 Sebastian Wojciechowski
```

bodhi/server/migrations/versions/cf7257bfa910_add_composed_by_bodhi_flag_to_releases_.py
```diff
@@ -0,0 +1,43 @@
...
+
+# revision identifiers, used by Alembic.
+revision = 'cf7257bfa910'
+down_revision = '68f9ccb1f388'
+
+
+def upgrade():
+    """Add a new composed_by_bodhi column to releases."""
+    op.add_column(
+        'releases', sa.Column('composed_by_bodhi', sa.Boolean(), server_default='t', default=True)
```
bowlofeggs: Hmm, I'd think you could do this to be clearer:

```suggestion
        'releases', sa.Column('composed_by_bodhi', sa.Boolean(), server_default=True, default=True)
```

bodhi/client/__init__.py
```diff
@@ -164,6 +164,8 @@ def _set_logging_debug(ctx, param, value):
                                                'archived']),
                  help='The state of the release'),
     click.option('--mail-template', help='Name of the email template for this release'),
+    click.option('--not_composed_by_bodhi', is_flag=True, default=False,
+                 help='The flag that indicates whether the release is composed by Bodhi or not'),
```
bowlofeggs, line 168: We should document this flag in the man page too (as you've noticed, I often forget about that when reviewing patches, so this time I remembered!)

bodhi/server/push.py
```diff
@@ -73,21 +73,24 @@ def push(username, cert_prefix, yes, **kwargs):
     db_factory = transactional_session_maker()
     composes = []
     with db_factory() as session:
-        if not resume and session.query(Compose).count():
+        bodhi_composes = session.query(Compose).join(Release).filter(
+            Release.composed_by_bodhi == True
+        ).all()
```
bowlofeggs: I don't think we need to filter ```Compose```s here like this, since line 228 below ensures that we don't create ```Compose```s for non-managed releases.

bodhi/tests/server/test_push.py
```diff
@@ -554,6 +554,84 @@ def test_locked_updates(self, publish, mock_init):
             self.assertFalse(u.locked)
             self.assertIsNone(u.date_locked)
 
+    @mock.patch('bodhi.server.push.bodhi.server.notifications.init')
+    @mock.patch('bodhi.server.push.bodhi.server.notifications.publish')
+    def test_composed_by_bodhi_release(self, publish, mock_init):
+        """
+        Test correct operation when there are some locked updates for releases not composed
+        by Bodhi.
```
bowlofeggs: Updates get locked when they are being composed, so updates that are part of a release that are not composed by Bodhi should not get locked.

bodhi/tests/server/test_push.py
```diff
@@ -723,6 +801,91 @@ def test_releases_flag(self, publish, mock_init):
         self.assertEqual(f26_python_paste_deploy.compose.release.id, f26.id)
         self.assertEqual(f26_python_paste_deploy.compose.request, models.UpdateRequest.testing)
 
+    @mock.patch('bodhi.server.push.bodhi.server.notifications.init')
+    @mock.patch('bodhi.server.push.bodhi.server.notifications.publish')
+    def test_create_composes_for_releases_marked_as_composed_by_bodhi(self, publish, mock_init):
+        """
+        Assert that composes are crate only for releases marked as 'composed_by_bodhi'.
```
bowlofeggs:
```suggestion
        Assert that composes are created only for releases marked as 'composed_by_bodhi'.
```

bowlofeggs (changes requested):

bodhi/server/push.py
sebwoj: I understand. But lets consider situation, where user runs bodhi-push.py without "-y" and without "--resume", and there are 2 composes in failed state, one for Rawhide(composed_by_bodhi=F) and one for normal release(composed_by_bodhi=T):
- At line 81, script will ask if user want to resume 2 composes(for both releases)
- At lines 160 and 169, script will print/ask about compose/updates only for the second release(one with composed_by_bodhi=T), and silently skip the compose/updates for first release
What do you thing about this situation?

sebwoj: I understand. But lets consider situation, where user runs bodhi-push.py without "-y" and without "--resume", and there are 2 composes in failed state, one for Rawhide(composed_by_bodhi=F) and one for normal release(composed_by_bodhi=T):
- At line 81, script will ask if user want to resume 2 composes(for both releases)
- At lines 160 and 169, script will print/ask about compose/updates only for the second release(one with composed_by_bodhi=T), and silently skip the compose/updates for first release
What do you thing about this situation?

bodhi/server/migrations/versions/cf7257bfa910_add_composed_by_bodhi_flag_to_releases_.py
sebwoj: The reason why I choose 't'
https://docs.sqlalchemy.org/en/latest/core/metadata.html#sqlalchemy.schema.Column.params.server_default

bodhi/tests/server/test_push.py
sebwoj: So the situation in
https://github.com/fedora-infra/bodhi/pull/2681#discussion_r231678175
will never take place?
Composes for releases "not composed by Bodhi" will never be created?

bodhi/server/migrations/versions/cf7257bfa910_add_composed_by_bodhi_flag_to_releases_.py
bowlofeggs: @sebwoj I'm not sure I follow. Since this is a Boolean field and not a text field, I think it's sensible to use a ```bool``` value.

bodhi/server/push.py
bowlofeggs: There should not ever be a ```Compose``` object for Rawhide, since Bodhi won't manage Rawhide. Composes are models that are used to track compose tasks that Bodhi is running, and Bodhi won't compose Rawhide.

bodhi/tests/server/test_push.py
bowlofeggs: @sebwoj correct, composes for releases not composed by Bodhi will not create Compose objects in Bodhi's database.

bodhi/server/migrations/versions/cf7257bfa910_add_composed_by_bodhi_flag_to_releases_.py
sebwoj: "server_default" docs mentioned earlier points that it should be a "str" type. At first I used "True", but   after running alembic to update my DB, an error was rised. I found solution on StackOverflow("t").

bodhi/server/push.py
```diff
@@ -221,6 +221,9 @@ def _filter_releases(session, query, releases=None):
         _releases = session.query(Release).filter(or_(Release.state == ReleaseState.current,
                                                       Release.state == ReleaseState.pending))
 
+    # Filter only releases composed by Bodhi.
+    _releases = [r for r in _releases if r.composed_by_bodhi]
```
bowlofeggs: Why not start the queries off with a query for releases that are composed by Bodhi, so we can let the database do the filtering? We could start a query before the ```if/else``` and then have the ```if/else``` add an additional ```.filter()``` call as appropriate.

bowlofeggs (changes requested):

bodhi/server/push.py
sebwoj: Good idea! I will make this change.

sebwoj:
It looks like builds: pip-build and rawhide-build, failed (some timeout in pip-build).
What can be the root cause?

bowlofeggs:
On Tue, 2018-12-11 at 22:24 +0000, Sebastian Wojciechowski wrote:
> I looks like builds: pip-build and rawhide-build, failed (some
> timeout in pip-build).
> What can be the root cause?

Yeah I've been seeing similar problems over in
https://github.com/fedora-infra/bodhi/pull/2532. It seems like there
might be some trouble with contacting the Fedora mirrors to get the
packages needed for the builds. I advise waiting a bit to see if it
gets sorted out.

bowlofeggs (approved):
Thanks!

bowlofeggs:
The test failures will be fixed by https://github.com/fedora-infra/bodhi/pull/2891

bowlofeggs:
I'm going to rebase this to pick up the fixes for the failed tests. :crossed_fingers:

bowlofeggs:
The CI system is [offline](https://bugs.centos.org/view.php?id=15665) so I am going to manually merge this. It passes all tests when run locally with ```devel/ci/bodhi-ci all```.

bowlofeggs:

bowlofeggs:

PR #2687, opened 2018-10-26
Mock calls to sleep() in the tests.

bowlofeggs:
Bodhi uses three calls to time.sleep():

0. The Composer sleeps for 3 seconds after launching Pungi to see
   if it exited quickly before continuing.
1. The Update consumer uses it to work around some bug (I did not
   investigate whether this is still necessary, as it was easy
   enough to just mock it.
2. utils.call_api() uses it to wait and retry if remote API calls
   fail.

These collectively caused the tests to spend a lot of time
sleeping. By mocking them, we are about to save 2 minutes and 15
seconds on the Python 3 tests in the Vagrant environment on my
laptop. I measured a time of 5m16.013s before, and 3m0.519s after
this commit.


puiterwijk (dismissed):

cverna (dismissed):
LGTM

bowlofeggs:
Strange - Mergify is saying that some branch protection setting is blocking it, but it merged other PRs just fine and all the checks are passing. I'm going to manually rebase and merge.

puiterwijk (approved):

bowlofeggs:
Hmm, the tests all got killed.

PR #2696, opened 2018-10-31
Add bodhi-push --yes flag to docs

sebwoj:

### DIFF
--- a/docs/user/man_pages/bodhi-push.rst
+++ b/docs/user/man_pages/bodhi-push.rst
@@ -44,6 +44,10 @@ Options
 
     Resume one or more previously failed pushes.
 
+``-y, --yes``
+
+    Answers yes to the various questions.
+
 ``--staging``
 
     Use the staging bodhi instance instead of the production instance.


docs/user/man_pages/bodhi-push.rst
```diff
@@ -65,4 +69,4 @@ request::
 
     https://github.com/fedora-infra/bodhi
 
-Bodhi's documentation is available online: https://bodhi.fedoraproject.org/docs
\ No newline at end of file
+Bodhi's documentation is available online: https://bodhi.fedoraproject.org/docs
```
bowlofeggs: I'm going to force push to your branch to remove this change.

bowlofeggs (changes requested):

bowlofeggs (approved):

docs/user/man_pages/bodhi-push.rst
sebwoj: It looks like vim is adding new line by default(hedump -> 0x0a) I see that all files in man_pages have 0a on the end. Do you think we should we left this file without newline at the end of file?

PR #2701, opened 2018-11-02
Set up a log handler for bodhi-approve-testing

### LINKED ISSUE #2698: bodhi-approve-testing also needs logging set up
Similar to #2156, ```bodhi-approve-testing``` also prints a warning about not having a logger set up.

sebwoj:
fix #2698

### DIFF
--- a/bodhi/server/scripts/approve_testing.py
+++ b/bodhi/server/scripts/approve_testing.py
@@ -23,10 +23,11 @@ The script is responsible for commenting on updates after they reach the mandato
 spent in the testing repository.
 """
 
+import logging
 import os
 import sys
 
-from pyramid.paster import get_appsettings
+from pyramid.paster import get_appsettings, setup_logging
 import six
 
 from ..models import Update, UpdateStatus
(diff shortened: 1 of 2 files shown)


bodhi/server/scripts/approve_testing.py
```diff
@@ -64,7 +65,12 @@ def main(argv=sys.argv):
     if len(argv) != 2:
         usage(argv)
 
-    settings = get_appsettings(argv[1])
+    config_uri = argv[1]
+
+    setup_logging(config_uri)
+    log = logging.getLogger(__name__)
+
+    settings = get_appsettings(config_uri)
```
bowlofeggs, line 73: I don't think we should change where the logs go with this patch - the messages should continue to go to the console. 

Something along these lines would work:

https://github.com/fedora-infra/bodhi/pull/2651/files#diff-70dadc71359f6d15c0061f0626e61917R34

In general, I don't think we need to stop using the ```print``` statement in here, we just want to set up a logger so the warning goes away.

bowlofeggs (changes requested):

bowlofeggs:
Closing as #2698 was fixed by https://github.com/fedora-infra/bodhi/pull/2804.

PR #2704, opened 2018-11-03
Updates should get obsoleted when their release becomes archived

### LINKED ISSUE #2204: Updates should get obsoleted when their release becomes archived
We should mark all non-stable updates in a release as obsolete when the release gets archived.

We should also write a migration to obsolete existing updates that are associated with archived releases.

Once we've done this, we won't need the feature from #1845 anymore, so we can consider removing that as well.

sebwoj:
fix #2204

I'm not sure about final solution for migration script.
@bowlofeggs could you give me some hints?
Should it be added to Bodhi setup.py as a general script?
Should script prints any information?

### DIFF
--- a/bodhi/server/services/releases.py
+++ b/bodhi/server/services/releases.py
@@ -324,6 +324,23 @@ def save_release(request):
             log.info("Editing release: %s" % edited)
             r = request.db.query(Release).filter(Release.name == edited).one()
             for k, v in data.items():
+                # We have to change updates status to obsolete
+                # if state of release changes to archived
+                if k == "state" and v == ReleaseState.archived and \
+                        r.state != ReleaseState.archived:
+                    updates = request.db.query(Update).filter(Update.release_id == r.id).filter(
+                        Update.status.notin_(
+                            [UpdateStatus.obsolete, UpdateStatus.stable, UpdateStatus.unpushed]
+                        )
+                    ).all()
+                    for u in updates:
+                        u.status = UpdateStatus.obsolete
+                        u.comment(
+                            request.db,
+                            'This update is marked obsolete because '
+                            'the {} release is archived.'.format(u.release.name),
+                            author='bodhi',
+                        )
                 setattr(r, k, v)
 
     except Exception as e:
--- a/bodhi/server/migrations/versions/8e9dc57e082d_obsolete_updates_for_archived_release.py
+++ b/bodhi/server/migrations/versions/8e9dc57e082d_obsolete_updates_for_archived_release.py
@@ -1,0 +1,43 @@
+# -*- coding: utf-8 -*-
+# Copyright (c) 2019 Sebastian Wojciechowski
+#
+# This file is part of Bodhi.
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+"""
+Obsolete updates for archived release.
+
+Revision ID: 8e9dc57e082d
+Revises: d986618207bc
+Create Date: 2019-01-06 13:04:35.158562
+"""
+from alembic import op
+
+
+# revision identifiers, used by Alembic.
+revision = '8e9dc57e082d'
+down_revision = 'd986618207bc'
+
+
+def upgrade():
+    """Set archived releases updates status to 'obsolete'."""
+    op.execute("UPDATE updates SET status='obsolete' WHERE release_id in \
+                (SELECT id FROM releases WHERE state='archived') AND status NOT IN \
+                ('unpushed', 'obsolete', 'stable')")
+
+
+def downgrade():
+    """Raise an exception explaining that this migration cannot be reversed."""
+    raise NotImplementedError('This migration cannot be reversed.')
--- a/bodhi/tests/server/services/test_releases.py
+++ b/bodhi/tests/server/services/test_releases.py
@@ -23,7 +23,7 @@ import webtest
 
 from bodhi import server
 from bodhi.server.config import config
-from bodhi.server.models import Release, ReleaseState, Update
+from bodhi.server.models import Release, ReleaseState, Update, UpdateStatus
 from bodhi.server.util import get_absolute_path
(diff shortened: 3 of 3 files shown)


bodhi/server/scripts/obsolete_updates_for_archived_releases.py
```diff
@@ -0,0 +1,51 @@
...
+        else:
+            click.echo("No updates need change.")
+        """
+        for update in updates:
+            update.status = UpdateStatus.obsolete
+            db.commit()
+        """
+
+if __name__ == '__main__':
+    make_updates_obsolete()
```
bowlofeggs: We only need to do this once, when we first deploy the Bodhi that has the auto-obsoletion behavior, so let's do it with a database migration with alembic, instead of a console script.

bodhi/server/services/releases.py
```diff
@@ -325,6 +325,15 @@ def save_release(request):
             r = request.db.query(Release).filter(Release.name == edited).one()
             for k, v in data.items():
                 setattr(r, k, v)
+                # We have to change status for release updates to obsolete
+                if k == "state" and v == ReleaseState.archived:
```
bowlofeggs: It might be good to make sure we only do this if the value is changing to archived, for efficiency.

bodhi/server/services/releases.py
```diff
@@ -325,6 +325,15 @@ def save_release(request):
...
+                    updates = request.db.query(Update).filter(Update.release_id == r.id).filter(
+                            and_(*[Update.status != s for s in (
+                                UpdateStatus.obsolete, UpdateStatus.stable
```
bowlofeggs: We should also include unpushed.

You can express the above in a simpler fashion:

```
updates = request.db.query(Update).filter(Update.release_id == r.id).filter(
    Update.status.notin_([UpdateStatus.obsolete, UpdateStatus.stable, UpdateStatus.unpushed])
```

bowlofeggs (changes requested):

bowlofeggs:
The ```rawhide-python3-diff-cover``` miiiiight get fixed by https://src.fedoraproject.org/rpms/python-diff-cover/c/354c48310713d859431e7733ac1a37a609b664b4?branch=master so we'll ignore that one.

bodhi/server/services/releases.py
```diff
@@ -324,6 +324,17 @@ def save_release(request):
             log.info("Editing release: %s" % edited)
             r = request.db.query(Release).filter(Release.name == edited).one()
             for k, v in data.items():
+                # We have to change updates status to obsolete
+                # if state of release changes to archived
+                if (k == "state" and v == ReleaseState.archived and
+                        r.state != ReleaseState.archived):
```
bowlofeggs: ```flake8``` doesn't like this line:

```
00:21:24 ./bodhi/server/services/releases.py:329:25: W504 line break after binary operator
00:21:24                 if (k == "state" and v == ReleaseState.archived and
```

bowlofeggs (changes requested):

bodhi/server/services/releases.py
```diff
@@ -324,6 +324,17 @@ def save_release(request):
...
+                # if state of release changes to archived
+                if k == "state" and v == ReleaseState.archived and \
+                        r.state != ReleaseState.archived:
+                    updates = request.db.query(Update).filter(Update.release_id == r.id).filter(
+                        Update.status.notin_(
+                            [UpdateStatus.obsolete, UpdateStatus.stable, UpdateStatus.unpushed]
+                        )
+                    ).all()
+                    for u in updates:
+                        u.status = UpdateStatus.obsolete
```
bowlofeggs, line 337: The ```Update``` model has a dedicated helper method for setting this state:

```suggestion
                        u.obsolete(request.db)
```

There are two problems with this suggestion, however:

* It causes the request to take a long time. I tried it on the F28 release on my laptop and it took 6 minutes to finish the request. Two solutions I can think of are to add a fedmsg consumer to watch for changes to releases and do this there (in the background), or to add a task worker process to Bodhi that does this with something like Celery. The latter solution is something I've been wanting to do for a while because I don't love the fedmsg solution since it loses messages, but it's also kinda unrelated to this change.
* It untags the builds in Koji. I'm actually not sure if this is a problem or not - I think we should ask infra and releng what they think. @nirik or @puiterwijk or @mohanboddu: is it OK if we untag builds that didn't make it to stable in koji from releases when they become archived at EOL?

bowlofeggs (commented):
I am on the fence about my proposed change, so let's not apply it just yet - I'd like to get feedback from others before proceeding here.

bodhi/server/services/releases.py
bowlofeggs: CC @mohanboddu

bowlofeggs: To clarify a bit, by untag, I mean this function:

https://github.com/fedora-infra/bodhi/blob/3.11.2/bodhi/server/models.py#L3198-L3215

It should only remove tags that Bodhi knows about, which are tags returned from this function:

https://github.com/fedora-infra/bodhi/blob/3.11.2/bodhi/server/models.py#L929-L953

mohanboddu: The only thing is, if we untag a build, there is a chance that it might get untagged from all tags, if and when it happens the build might get garbage collected. Even then I dont think there would be problem as we garbage collect lot of other builds.

So, I think it should be fine. But, I would like to get other's opinions as well.

bowlofeggs:
On Mon, 2018-12-03 at 15:15 -0800, mohanboddu wrote:
> The only thing is, if we untag a build, there is a chance that it
> might get untagged from all tags, if and when it happens the build
> might get garbage collected. Even then I dont think there would be
> problem as we garbage collect lot of other builds.

Hmm. I'd think we probably don't want to garbage collect builds that
were in the testing repositories, right?

mohanboddu:
> On Mon, 2018-12-03 at 15:15 -0800, mohanboddu wrote: The only thing is, if we untag a build, there is a chance that it might get untagged from all tags, if and when it happens the build might get garbage collected. Even then I dont think there would be problem as we garbage collect lot of other builds.
> Hmm. I'd think we probably don't want to garbage collect builds that were in the testing repositories, right?

Yeah, right, I forgot they will be in testing repos.

bodhi/server/services/releases.py
bowlofeggs: OK, now that I've had time to think about this some more, I'm convinced that we should not use the obsoletion function and the way you've done it here is good.

One thing I think we should do, however, is add a comment on each update when we change it to obsoleted state. Something like this:

```
u.comment('This update is marked obsolete because the {} release is archived.'.format(u.release.name), agent='bodhi')
```

bowlofeggs (changes requested):

sebwoj:
PR updated.
BTW. Can we change releases status to archived when its updates are locked during composing?

bowlofeggs (approved):
Thanks!

bowlofeggs:
@sebwoj we don't want to change the release state unless the official state of that Fedora release is changed.

bowlofeggs:

bowlofeggs:

PR #2708, opened 2018-11-06
Simplify the Vagrant playbook by combining the roles.

bowlofeggs:
I've often regretted creating the separated roles for the Vagrant
environment since there is only one guest and will likely only ever
be one guest. This commit simplifies the Vagrant playbook by
distilling it down to just one role.

This refactor also resolves an Ansible deprecation warning about
using the dnf module with with_items.


cverna (approved):
LGTM

PR #2711, opened 2018-11-06
Use Python 3 when building the docs.

bowlofeggs:


devel/ci/Dockerfile-pip
```diff
@@ -72,4 +70,7 @@ RUN rm -rf /bodhi/docs/_build
 RUN rm -rf /bodhi/docs/developer/docblocks
 RUN sed -i '/pyramid_debugtoolbar/d' setup.py
 RUN sed -i '/pyramid_debugtoolbar/d' devel/development.ini.example
+# The Makefile is set to use sphinx-build-3, but the pip installed sphinx is just called
+# sphinx-build.
+RUN sed -i 's/sphinx-build-3/sphinx-build/' docs/Makefile
```
cverna, line 75: Couldn't we change this in the Makefile ? since the Makefile is in git.

docs/Makefile
```diff
@@ -3,7 +3,7 @@
 
 # You can set these variables from the command line.
 SPHINXOPTS    =
-SPHINXBUILD   = sphinx-build
+SPHINXBUILD   = sphinx-build-3
```
cverna, line 6: Related to the previous comment 

```suggestion
SPHINXBUILD   = sphinx-build
```

devel/ci/Dockerfile-pip
bowlofeggs: @cverna the problem is that most of the containers do need to call ```sphinx-build-3``` but this one ```pip``` container needs to call ```sphinx-build```. I decided to put ```sphinx-build-3``` in the Makefile since the Fedora 27, 28, 29, and Rawhide containers will all call it that, but that requires this exception in here in the pip container.

docs/Makefile
bowlofeggs: As said above, 4 out of the 5 releases I currently test do call it ```sphinx-build-3```, so I think we do want to explicitly call it that way here.

bowlofeggs:
I will rebase this to fix the conflict.

devel/ci/Dockerfile-pip
cverna: ok that works for me :+1:

cverna (approved):

PR #2712, opened 2018-11-06
Use the Python 3 version of alembic if available.

bowlofeggs:


cverna (approved):
LGTM

PR #2714, opened 2018-11-07
Gracefully handle private bugs

### LINKED ISSUE #344: Gracefully private bugs using the new caveats API.
```
  File "/usr/lib/python2.7/site-packages/bodhi/services/updates.py", line 340, in new_update
    result, _caveats = Update.edit(request, data)
  File "/usr/lib/python2.7/site-packages/bodhi/models/models.py", line 767, in edit
    new_bugs = up.update_bugs(data['bugs'])
  File "/usr/lib/python2.7/site-packages/bodhi/models/models.py", line 1277, in update_bugs
    newbug = bugtracker.getbug(bug_id)
  File "/usr/lib/python2.7/site-packages/bodhi/bugs.py", line 61, in getbug
    return self.bz.getbug(bug_id)
  File "/usr/lib/python2.7/site-packages/bugzilla/base.py", line 1042, in getbug
    exclude_fields=exclude_fields, extra_fields=extra_fields)
  File "/usr/lib/python2.7/site-packages/bugzilla/base.py", line 1035, in _getbug
    extra_fields=extra_fields)[0]
  File "/usr/lib/python2.7/site-packages/bugzilla/base.py", line 1007, in _getbugs
    r = self._proxy.Bug.get(getbugdata)
  File "/usr/lib64/python2.7/xmlrpclib.py", line 1224, in __call__
    return self.__send(self.__name, args)
  File "/usr/lib/python2.7/site-packages/bugzilla/base.py", line 168, in _ServerProxy__request
    ret = ServerProxy._ServerProxy__request(self, methodname, params)
  File "/usr/lib64/python2.7/xmlrpclib.py", line 1578, in __request
    verbose=self.__verbose
  File "/usr/lib/python2.7/site-packages/bugzilla/base.py", line 263, in request
    return self._request_helper(url, request_body)
  File "/usr/lib/python2.7/site-packages/bugzilla/base.py", line 245, in _request_helper
    raise sys.exc_info()[1]
Fault: <Fault 102: 'You are not authorized to access bug #1253789. Most likely the bug has been restricted for internal
 development processes and we cannot grant access. If you are a Red Hat customer with an active subscription, please vi
sit the Red Hat Customer Portal for assistance with your issue If you are a Fedora Project user and require assistance,
 please consider using one of the mailing lists we host for the Fedora Project.'>

```

mattiaverga:
The PR adds a new column to bugs table to flag as private those bugs which are not publicly accessible in RH BZ.
Bodhi will avoid to comment or change status of those bugs, since it doesn't have permission to do so. Those bugs can still be attached to updates as reference for maintainer and to allow the reporter to give feedback. In the GUI they will be displayed with a "Private bug" label, and the same string will be used for title (as Bodhi can't fetch the real title).

This comes with all relevant tests, but I couldn't test in real, because in the vagrant box all users are granted only the 'guest' permissions.

This PR involves change in database schema.
Fixes #344 

### DIFF
--- a/bodhi/server/consumers/updates.py
+++ b/bodhi/server/consumers/updates.py
@@ -162,6 +162,9 @@ class UpdatesHandler(fedmsg.consumers.FedmsgConsumer):
         details from Bugzilla, comment on the bug to let watchers know about the update, and mark
         the bug as MODIFIED. If the bug is a security issue, mark the update as a security update.
 
+        If the bug is private, Bodhi can't retrieve any information, comment on it, or modify
+        it, so we just associate the bug id with the update and mark it to be private.
+
         If handle_bugs is not True, return and do nothing.
 
         Args:
@@ -181,6 +184,10 @@ class UpdatesHandler(fedmsg.consumers.FedmsgConsumer):
 
                 log.info("Updating our details for %r" % bug.bug_id)
                 bug.update_details(rhbz_bug)
+                if bug.private:
+                    # Bodhi can't retrieve any information so just continue with the next bug
+                    log.info("  Skipping bug %r because it is private" % (bug.bug_id))
+                    continue
                 log.info("  Got title %r for %r" % (bug.title, bug.bug_id))
 
                 # If you set the type of your update to 'enhancement' but you
--- a/bodhi/server/mail.py
+++ b/bodhi/server/mail.py
@@ -325,9 +325,9 @@ def get_template(update, use_template='fedora_errata_template'):
                     if parent and not bug.parent:
                         log.debug("Skipping tracker bug %s" % bug)
                         continue
-                title = (
-                    bug.title != 'Unable to fetch title' and bug.title != 'Invalid bug number') \
-                    and ' - %s' % bug.title or ''
+                title = (bug.title != 'Unable to fetch title'
+                         and bug.title != 'Invalid bug number'
+                         and not bug.private) and ' - %s' % bug.title or ''
                 info['references'] += u"  [ %d ] Bug #%d%s\n        %s\n" % \
                                       (i, bug.bug_id, title, bug.url)
                 i += 1
--- a/bodhi/server/migrations/versions/3a14c47250fb_add_a_private_flag_to_bugs.py
+++ b/bodhi/server/migrations/versions/3a14c47250fb_add_a_private_flag_to_bugs.py
@@ -1,0 +1,43 @@
+# Copyright (c) 2018 Mattia Verga
+#
+# This file is part of Bodhi.
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+"""
+Add a private flag to bugs.
+
+Revision ID: 3a14c47250fb
+Revises: 68f9ccb1f388
+Create Date: 2018-11-04 14:55:57.939008
+"""
+from alembic import op
+import sqlalchemy as sa
+
+
+# revision identifiers, used by Alembic.
+revision = '3a14c47250fb'
+down_revision = '68f9ccb1f388'
+
+
+def upgrade():
+    """Add a new Bool column to bugs table, set default to False."""
+    op.add_column('bugs', sa.Column('private', sa.Boolean(), nullable=True))
+    op.execute("""UPDATE bugs SET private = FALSE""")
+    op.alter_column('bugs', 'private', nullable=False)
+
+
+def downgrade():
+    """Just remove the new column."""
+    op.drop_column('bugs', 'private')
--- a/bodhi/server/bugs.py
+++ b/bodhi/server/bugs.py
@@ -147,6 +147,9 @@ class Bugzilla(BugTracker):
             if len(comment) > 65535:
                 raise InvalidComment("Comment is too long: %s" % comment)
             bug = self.bz.getbug(bug_id)
+            if bug.private:
+                log.info('Unable to comment on private bug #%d' % bug_id)
+                return
             attempts = 0
             while attempts < 5:
                 try:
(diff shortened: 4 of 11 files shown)


mattiaverga:
JENKIES TEST

mattiaverga:
JENKINS TEST

bodhi/server/consumers/updates.py
```diff
@@ -162,6 +162,9 @@ def work_on_bugs(self, session, update, bugs):
         details from Bugzilla, comment on the bug to let watchers know about the update, and mark
         the bug as MODIFIED. If the bug is a security issue, mark the update as a security update.
 
+        If the bug is private, Bodhi can't retrieve any information, neither comment it or modify
```
bowlofeggs:
```suggestion
        If the bug is private, Bodhi can't retrieve any information, comment on it, or modify
```

bodhi/server/consumers/updates.py
```diff
@@ -162,6 +162,9 @@ def work_on_bugs(self, session, update, bugs):
...
+        it, so we just associate the bug id to the update and mark it to be private.
```
bowlofeggs:
```suggestion
        it, so we just associate the bug id with the update and mark it to be private.
```

bodhi/server/mail.py
```diff
@@ -325,8 +325,9 @@ def get_template(update, use_template='fedora_errata_template'):
                     if parent and not bug.parent:
                         log.debug("Skipping tracker bug %s" % bug)
                         continue
-                title = (
-                    bug.title != 'Unable to fetch title' and bug.title != 'Invalid bug number') \
+                title = (bug.title != 'Unable to fetch title'
+                    and bug.title != 'Invalid bug number'
+                    and bug.title != 'Private bug') \
```
bowlofeggs:
```suggestion
                    and not bug.private) \
```

Also, ```flake8``` is not happy with these lines:

```
04:25:11 ./bodhi/server/mail.py:329:21: W503 line break before binary operator
04:25:11                     and bug.title != 'Invalid bug number'
04:25:11                     ^
04:25:11 ./bodhi/server/mail.py:329:21: E128 continuation line under-indented for visual indent
04:25:11                     and bug.title != 'Invalid bug number'
04:25:11                     ^
04:25:11 ./bodhi/server/mail.py:330:21: W503 line break before binary operator
04:25:11                     and bug.title != 'Private bug') \
04:25:11                     ^
04:25:11 ./bodhi/server/mail.py:330:21: E128 continuation line under-indented for visual indent
04:25:11                     and bug.title != 'Private bug') \
```

bodhi/server/migrations/versions/3a14c47250fb_add_a_private_flag_to_bugs.py
```diff
@@ -0,0 +1,43 @@
...
+# revision identifiers, used by Alembic.
+revision = '3a14c47250fb'
+down_revision = '68f9ccb1f388'
+
+
+def upgrade():
+    """Add a new Bool column to bugs table, set default to False."""
+    op.add_column('bugs', sa.Column('private', sa.Boolean(), nullable=True))
+    op.execute("""UPDATE bugs SET private = FALSE""")
+    op.alter_column('bugs', 'private', nullable=False, server_default='false')
```
bowlofeggs: Let's let the Python models deal with defaults instead of the server.

```suggestion
    op.alter_column('bugs', 'private', nullable=False)
```

bowlofeggs (changes requested):

mattiaverga:
mmm it seems flake8 doesn't like line breaks before binary operator nor after:
```
./bodhi/server/mail.py:328:26: W504 line break after binary operator
                title = (bug.title != 'Unable to fetch title' and
                         ^
```
```
bodhi/server/mail.py:329:26: W503 line break before binary operator
                         and bug.title != 'Invalid bug number'
                         ^
```
Searching around it seems [PEP-8](https://www.python.org/dev/peps/pep-0008/#should-a-line-break-before-or-after-a-binary-operator) recently changed the best practice to be line break before binary operator, so I will change the code to do so. Probably we need to disable W503 error check in flake8.

bowlofeggs:
Yeah let's break before as recommended by PEP-8. If we need to disable that warning to get it to accept that, so be it, though it would be good to file an issue with ```flake8``` about this.

mattiaverga:
>  it would be good to file an issue with `flake8` about this.

It seems that someone already did it and the reply was "add w503 to --ignore"... 
https://gitlab.com/pycqa/flake8/issues/463

May I add that setting myself in this same PR, but in a separate commit?

mattiaverga:
More on this: the best solution would be to change `ignore` to `extend-ignore`, which would not overwrite flake8 defaults. Unfortunately, `extend-ignore` was introduced in version 3.6, which is not yet available in Fedora repositories (devel environment seems to use 3.4.1).

BTW is there any compelling reason to ignore E712?
```
./bodhi/server/models.py:2296:36: E712 comparison to False should be 'if cond is False:' or 'if not cond:'
                     Update.locked == False,
                                   ^
./bodhi/server/views/generic.py:84:36: E712 comparison to True should be 'if cond is True:' or 'if cond:'
            models.Update.critpath == True)
                                   ^
```

bowlofeggs:
On Fri, 2018-12-07 at 08:35 -0800, Mattia Verga wrote:
> It seems that someone already did it and the reply was "add w503 to
> --ignore"...
> https://gitlab.com/pycqa/flake8/issues/463

Ha, that's so strange.

> May I add that setting myself in this same PR, but in a separate
> commit?

Sure, and thanks!

bowlofeggs:
On Fri, 2018-12-07 at 09:01 -0800, Mattia Verga wrote:
> More on this: the best solution would be to change ignore to extend-
> ignore, which would not overwrite flake8 defaults. Unfortunately,
> extend-ignore was introduced in version 3.6, which is not yet
> available in Fedora repositories (devel environment seems to use
> 3.4.1).

Since flake8 is only used by CI and Vagrant, I think it's OK for us to
pip install a newer version if you want, instead of using the RPM. The
downstream spec file does not run flake8, so it really is just used by
us here.

I'll make myself a todo to send a PR to the python-flake8 package to
get it updated and switch our Rawhide back to the RPM when that is
done.

> BTW is there any compelling reason to ignore E712?
> ./bodhi/server/models.py:2296:36: E712 comparison to False should be
> 'if cond is False:' or 'if not cond:'
>                      Update.locked == False,
>                                    ^
> ./bodhi/server/views/generic.py:84:36: E712 comparison to True should
> be 'if cond is True:' or 'if cond:'
>             models.Update.critpath == True)
>                                    ^

Not really - I think I just was lazy. I did not know if sqlalchemy
works with the is operator and just hadn't taken the time to
investigate.

mattiaverga:
For now I've simply added W503 to the `exclude` list.
When flake8 3.6 will be available we can simply correct those two E712 and get rid of the entire `exclude` line, since W503 will be ignored by default.

bowlofeggs:
The ```f29-python3-unit``` test actually passed but CI failed while notifying GitHub about it, seemingly. I suspect the integration test failure is due to https://github.com/fedora-infra/bodhi/issues/2835

bowlofeggs (approved):
Thanks!

bowlofeggs:
Ah, looks like there's a conflict - I'll rebase and fix that for you.

bowlofeggs:

bowlofeggs:

bowlofeggs:

PR #2719, opened 2018-11-09
Add integration testing to bodhi-ci

abompard:
It currently only works for F27 and Python 2 (which is what we currently
have in prod) but it should be possible to expand that in the future if
there's a need.

The created container is different from the currently created bodhi
containers because it tries to be as close as possible to the production
setup: bodhi is running behind apache, with the production.ini
configuration file, etc.

The integration test suite executes on the host running bodhi-ci via
pytest, but will start and test containers running the components
(currently only Bodhi and PostgreSQL). It only tests a successful GET on
/ for now, it should be easy to add more tests.

Signed-off-by: Aurélien Bompard <aurelien@bompard.org>

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -29,6 +29,7 @@ pull_request_rules:
   - status-success=pip-python2-diff-cover
   - status-success=pip-python3-unit
   - status-success=pip-python3-diff-cover
+  - status-success=f29-python2-integration
   name: default
 - actions:
     merge:
--- a/devel/ci/bodhi-ci
+++ b/devel/ci/bodhi-ci
@@ -260,6 +260,24 @@ def unit(archive, concurrency, container_runtime, no_build, failfast, init, pyve
                                archive_path=archive_path))
 
 
+@cli.command()
+@archive_option
+@concurrency_option
+@container_runtime_option
+@failfast_option
+@init_option
+@no_build_option
+@pyver_option
+@releases_option
+@archive_path_option
+@tty_option
+def integration(archive, concurrency, container_runtime, no_build, failfast, init, pyver, release,
+                archive_path, tty):
+    """Run the integration tests."""
+    _run_jobs(_build_jobs_list('integration', concurrency, release, pyvers=pyver, archive=archive,
+                               archive_path=archive_path))
+
+
 class Job(object):
     """
     Represent a CI job.
@@ -388,7 +406,7 @@ class Job(object):
                 error.result = self
                 raise error
 
-        except asyncio.CancelledError as e:
+        except asyncio.CancelledError:
             self.cancelled = True
         finally:
             # If the job's been cancelled or successful, let's go ahead and print its output now.
@@ -705,9 +723,9 @@ class UnitJob(Job):
 
         self.pyver = pyver
         self._label = '{}-py{}'.format(self._label, pyver)
-        pytest_flags = ''
+        pytest_flags = '--junit-xml=nosetests.xml bodhi/tests'
         if failfast:
-            pytest_flags = '-x'
+            pytest_flags += ' -x'
 
         test_command = ('({} setup.py develop && {} {} || (cp *.xml /results && exit 1)) '
                         '&& cp *.xml /results')
@@ -755,6 +773,70 @@ class UnitJob(Job):
         return self
 
 
+class IntegrationBuildJob(BuildJob):
+    """
+    Define a Job for building container images for integration testing.
+
+    See the Job superclass's docblock for details about its attributes.
+    """
+
+    def __init__(self, app_name, *args, **kwargs):
+        """
+        Initialize the IntegrationBuildJob.
+
+        See the superclass's docblock for details about accepted parameters.
+        """
+        super(IntegrationBuildJob, self).__init__(*args, **kwargs)
+
+        self._label = 'integration-build-{}'.format(app_name)
+        self._container_image = '{}-integration-{}'.format(CONTAINER_NAME, app_name)
+        dockerfile = os.path.join(PROJECT_PATH, 'devel', 'ci', 'integration', app_name, 'Dockerfile')
+        self._command = [container_runtime, 'build', '--pull', '-t', self._container_image,
+                         '-f', dockerfile, '.']
+
+
+class IntegrationJob(Job):
+    """
+    Define a Job for running the integration tests.
+
+    See the Job superclass's docblock for details about its attributes.
+
+    Attributes:
+        pyver (int): The Python version we are testing.
+    """
+
+    _label = 'integration'
+
+    def __init__(self, archive, archive_path, pyver, *args, **kwargs):
+        """
+        Initialize the IntegrationJob.
+
+        See the superclass's docblock for details about additional accepted parameters.
+
+        Args:
+            archive (bool): If True, set up the volume mount so we can retrieve the test results
+                from the container.
+            archive_path (str): A path on the host to share as a volume into the container for
+                its /results path.
+            pyver (int): Which major version of Python we are testing with.
+        """
+        super(IntegrationJob, self).__init__(*args, **kwargs)
+
+        self.pyver = pyver
+        self._command = [sys.executable, '-m', 'pytest', '--no-cov', 'devel/ci/integration/tests/']
+        bodhi_container_image = '{}-integration-bodhi'.format(CONTAINER_NAME)
+        self._popen_kwargs["env"] = os.environ.copy()
+        self._popen_kwargs["env"]["BODHI_INTEGRATION_IMAGE"] = bodhi_container_image
+        if failfast:
+            self._command.append('-x')
+        if archive:
+            self._command.append(
+                '--junit-xml={}/{}/nosetests.xml'.format(
+                    archive_path, self._label
+                )
+            )
+
+
 def _build_jobs_list(
         command: str, concurrency: int, releases: typing.Iterable[str],
         pyvers: typing.Iterable[int] = None, archive: bool = False,
@@ -780,34 +862,46 @@ def _build_jobs_list(
         buffer_output = concurrency != 1 and len(releases) != 1
     jobs = []
     for release in releases:
-        build_job = BuildJob(release, buffer_output=buffer_output)
-        jobs.append(build_job)
-        if command != 'build':
-            if command in ('all', 'docs'):
-                docs_job = DocsJob(release, depends_on=build_job, buffer_output=buffer_output)
-                jobs.append(docs_job)
-            if command in ('all', 'flake8'):
-                flake8_job = Flake8Job(release, depends_on=build_job, buffer_output=buffer_output)
-                jobs.append(flake8_job)
-            if command in ('all', 'pydocstyle'):
-                pydocstyle_job = PydocstyleJob(release, depends_on=build_job,
-                                               buffer_output=buffer_output)
-                jobs.append(pydocstyle_job)
-            if command in ('all', 'diff_cover', 'unit'):
-                for pyver in pyvers:
-                    unit_job = UnitJob(
-                        archive=archive, archive_path=archive_path, pyver=pyver, release=release,
-                        depends_on=build_job, buffer_output=buffer_output)
-                    jobs.append(unit_job)
-                    if command in ('all', 'diff_cover'):
-                        if pyver == 2 and release not in ('f27', 'f28', 'f29', 'pip'):
-                            # We don't support Python 2 on the server for Fedora >= 30, so don't
-                            # enforce coverage there.
-                            continue
-                        diff_cover_job = DiffCoverJob(
-                            archive=archive, archive_path=archive_path, pyver=pyver,
-                            release=release, depends_on=unit_job, buffer_output=buffer_output)
-                        jobs.append(diff_cover_job)
+        if command in ('all', 'build', 'docs', 'flake8', 'pydocstyle', 'diff_cover', 'unit'):
+            build_job = BuildJob(release, buffer_output=buffer_output)
+            jobs.append(build_job)
+        if command in ('all', 'docs'):
+            docs_job = DocsJob(release, depends_on=build_job, buffer_output=buffer_output)
+            jobs.append(docs_job)
+        if command in ('all', 'flake8'):
+            flake8_job = Flake8Job(release, depends_on=build_job, buffer_output=buffer_output)
+            jobs.append(flake8_job)
+        if command in ('all', 'pydocstyle'):
+            pydocstyle_job = PydocstyleJob(release, depends_on=build_job,
+                                           buffer_output=buffer_output)
+            jobs.append(pydocstyle_job)
+        if command in ('all', 'diff_cover', 'unit'):
+            for pyver in pyvers:
+                unit_job = UnitJob(
+                    archive=archive, archive_path=archive_path, pyver=pyver, release=release,
+                    depends_on=build_job, buffer_output=buffer_output)
+                jobs.append(unit_job)
+                if command in ('all', 'diff_cover'):
+                    if pyver == 2 and release not in ('f27', 'f28', 'f29', 'pip'):
+                        # We don't support Python 2 on the server for Fedora >= 30, so don't
+                        # enforce coverage there.
+                        continue
+                    diff_cover_job = DiffCoverJob(
+                        archive=archive, archive_path=archive_path, pyver=pyver,
+                        release=release, depends_on=unit_job, buffer_output=buffer_output)
+                    jobs.append(diff_cover_job)
+    if command in ('all', 'integration') and 3 in pyvers and "f29" in releases:
+        build_jobs = [
+            IntegrationBuildJob(app_name="resultsdb", release="f29", buffer_output=buffer_output),
+            IntegrationBuildJob(app_name="waiverdb", release="f29", buffer_output=buffer_output),
+            IntegrationBuildJob(app_name="greenwave", release="f29", buffer_output=buffer_output),
+            IntegrationBuildJob(app_name="bodhi", release="f29", buffer_output=buffer_output),
+        ]
+        jobs.extend(build_jobs)
+        integration_job = IntegrationJob(
+            archive=archive, archive_path=archive_path, pyver=3, release="f29",
+            depends_on=build_jobs, buffer_output=buffer_output)
+        jobs.append(integration_job)
 
     return jobs
 
@@ -889,6 +983,10 @@ def _run_jobs(jobs):
     Args:
         jobs (list): A list of Jobs to run.
     """
+    if not jobs:
+        click.echo("No jobs!")
+        sys.exit(3)
+
     loop = asyncio.get_event_loop()
 
     processes = [j.run() for j in jobs]
--- a/devel/ci/bodhi-ci
+++ b/devel/ci/bodhi-ci
@@ -271,8 +271,8 @@ class Job(object):
         archive_dir (str): A path where this job's test results is on the host. Only set if an
             archive is used (not used on all jobs).
         release (str): The release this job is testing.
-        depends_on (Job or None): If not None, this Job will wait for the depends_on job's complete
-            Event to start.
+        depends_on (list of Job): This Job will wait for all the jobs' complete Event in the list
+            to start.
         cancelled (bool): True if this Job has been cancelled, False otherwise.
         complete (asyncio.Event): An Event that allows other Jobs to wait for this Job to complete.
             complete.set() is called at the end of run().
@@ -287,13 +287,17 @@ class Job(object):
 
         Args:
             release (str): The release this Job pertains to.
-            depends_on (Job): Another Job that this Job should wait to complete before starting.
+            depends_on (list of Job): A list of Job that this Job should wait to complete before
+                starting. It is also possible to provide a Job instance if there is a single
+                dependency.
             buffer_output (bool): If True, this Job will buffer its stdout and stderr into its
                 output property. If False, child processes will send their output straign to stdout
                 and stderr.
         """
         self.release = release
-        self.depends_on = depends_on
+        self.depends_on = depends_on or []
+        if not isinstance(self.depends_on, list):
+            self.depends_on = [self.depends_on]
 
         self.cancelled = False
         # Used to block dependent processes until this Job is done.
@@ -342,9 +346,9 @@ class Job(object):
             Job: Returns self.
         """
         try:
-            if self.depends_on:
-                await self.depends_on.complete.wait()
-                if self.depends_on.returncode != 0 and not self.depends_on.skipped:
+            for job in self.depends_on:
+                await job.complete.wait()
+                if job.returncode != 0 and not job.skipped:
                     # If the Job we depend on failed, we should cancel.
                     raise asyncio.CancelledError()
 
@@ -569,7 +573,7 @@ class DiffCoverJob(Job):
         """
         if not os.path.exists(self.archive_dir):
             os.makedirs(self.archive_dir)
-        shutil.copy(os.path.join(self.depends_on.archive_dir, 'coverage.xml'),
+        shutil.copy(os.path.join(self.depends_on[0].archive_dir, 'coverage.xml'),
                     os.path.join(self.archive_dir, 'coverage.xml'))
 
         super(DiffCoverJob, self)._pre_start_hook()
--- a/devel/ci/cico.pipeline
+++ b/devel/ci/cico.pipeline
@@ -43,10 +43,13 @@ def synctoduffynode(rsyncpath) {
  */
 def configure_node = {
     onmyduffynode 'yum -y install epel-release'
-    onmyduffynode 'yum install -y docker python36-click rsync'
+    onmyduffynode 'yum install -y docker python36-click python36-requests rsync'
     // Workaround https://bugzilla.redhat.com/show_bug.cgi?id=1655214
     onmyduffynode 'yum downgrade -y docker-1.13.1-75.git8633870.el7.centos.x86_64 docker-client-1.13.1-75.git8633870.el7.centos.x86_64 docker-common-1.13.1-75.git8633870.el7.centos.x86_64'
     onmyduffynode 'systemctl start docker'
+    // To run the integration testsuite
+    onmyduffynode 'python3.6 -m ensurepip'
+    onmyduffynode 'python3.6 -m pip install \'pytest<4.0\' pytest-cov conu munch psycopg2'
 }
 
 
@@ -150,7 +153,10 @@ node('bodhi') {
             f28: {test_release('f28')},
             f29: {test_release('f29')},
             pip: {test_release('pip')},
-            rawhide: {test_release('rawhide')}
+            rawhide: {test_release('rawhide')},
+            integration: {
+                bodhi_ci('f29', 'integration', 'python3-integration', '--no-build --no-init -p 3')
+            }
         )
     } catch (e) {
         currentBuild.result = "FAILURE"
--- a/devel/ci/cico.pipeline
+++ b/devel/ci/cico.pipeline
@@ -44,6 +44,8 @@ def synctoduffynode(rsyncpath) {
 def configure_node = {
     onmyduffynode 'yum -y install epel-release'
     onmyduffynode 'yum install -y docker python36-click rsync'
+    // Workaround https://bugzilla.redhat.com/show_bug.cgi?id=1655214
+    onmyduffynode 'yum downgrade -y docker-1.13.1-75.git8633870.el7.centos.x86_64 docker-client-1.13.1-75.git8633870.el7.centos.x86_64 docker-common-1.13.1-75.git8633870.el7.centos.x86_64'
     onmyduffynode 'systemctl start docker'
 }
 
--- a/devel/ci/integration/bodhi/Dockerfile
+++ b/devel/ci/integration/bodhi/Dockerfile
@@ -1,0 +1,108 @@
+FROM registry.fedoraproject.org/fedora:29
+LABEL \
+  name="bodhi-web" \
+  vendor="Fedora Infrastructure" \
+  maintainer="Aurelien Bompard <abompard@fedoraproject.org>" \
+  license="MIT"
+
+RUN curl -o /etc/yum.repos.d/infra-tags.repo https://infrastructure.fedoraproject.org/cgit/ansible.git/plain/files/common/fedora-infra-tags.repo
+
+# While dnf has a --nodocs, it doesen't have a --docs...
+RUN sed -i '/nodocs/d' /etc/dnf/dnf.conf
+
+# Install Bodhi deps
+# 1. bodhi-server
+# 2. bodhi-composer
+# 3. python3-bodhi-client
+RUN dnf install -y \
+    \
+    httpd \
+    intltool \
+    liberation-mono-fonts \
+    python3-dogpile-cache \
+    python3-koji \
+    python3-librepo \
+    python3-mod_wsgi \
+    python3-pillow \
+    "python3dist(alembic)" \
+    "python3dist(arrow)" \
+    "python3dist(bleach)" \
+    "python3dist(click)" \
+    "python3dist(colander)" \
+    "python3dist(cornice)" \
+    "python3dist(cryptography)" \
+    "python3dist(fedmsg)" \
+    "python3dist(feedgen)" \
+    "python3dist(jinja2)" \
+    "python3dist(kitchen)" \
+    "python3dist(markdown)" \
+    "python3dist(psycopg2)" \
+    "python3dist(pylibravatar)" \
+    "python3dist(pyramid)" \
+    "python3dist(pyramid-fas-openid)" \
+    "python3dist(pyramid-mako)" \
+    "python3dist(python-bugzilla)" \
+    "python3dist(python-fedora)" \
+    "python3dist(python3-openid)" \
+    "python3dist(simplemediawiki)" \
+    "python3dist(six)" \
+    "python3dist(sqlalchemy)" \
+    "python3dist(waitress)" \
+    \
+    python3-createrepo_c \
+    python3-hawkey \
+    skopeo \
+    \
+    /usr/bin/koji \
+    python3-dnf \
+    python3-koji \
+    "python3dist(click)" \
+    "python3dist(iniparse)" \
+    "python3dist(kitchen)" \
+    "python3dist(python-fedora)" \
+    "python3dist(six)" \
+    \
+    fedmsg \
+    fedmsg-base \
+    git
+
+# Create bodhi user
+RUN groupadd -r bodhi && \
+    useradd  -r -s /sbin/nologin -d /home/bodhi/ -m -c 'Bodhi Server' -g bodhi bodhi
+
+# Copy source
+WORKDIR /bodhi
+COPY . /bodhi
+
+# Install it
+RUN sed -i -e '/pyramid_debugtoolbar/d' setup.py devel/development.ini.example
+RUN python3 setup.py build && pip3 install .
+
+# Configuration
+RUN mkdir -p /etc/bodhi
+COPY production.ini /etc/bodhi/production.ini
+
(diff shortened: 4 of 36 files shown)


abompard:
Uh ? It looks like the Jenkins tests didn't run.

abompard:
It looks like our Jenkins isn't configured to run the new `integration` sub-command. Is there a file in the source code that I should edit to add that? I haven't found any. @bowlofeggs any idea?

bowlofeggs:
Hey @abompard!

Yeah, have a look at ```devel/ci/cico.pipeline``` - it's Groovyscript - you should be able to find the place where I run the other test commands and just add yours there too. You should also edit ```.mergify.yml``` to add the new test as required to pass for merging.

abompard:
Great, that works. Integration run is now failing because of weird RPMDB corruption errors, but I'll figure that out.

abompard:
OK, tests pass except for the rawhide build, but I think that's because rawhide deps are broken at the moment.

bowlofeggs:
Yeah that Rawhide failure is due to https://bugzilla.redhat.com/show_bug.cgi?id=1649547 and can be ignored. I'll take a look at this PR soon - I'm currently battling the 3.11 deployment in staging. It's the first version of Bodhi to use Python 3 on the server and I keep finding problems.

abompard:
It's currently lacking in documentation but I can add that later if you're OK with the general structure.

abompard:
I don't think the failure on `rawhide-python3-diff-cover` is caused by one of my changes...

So, at the moment the integration tests are running the following separate containers:
- bodhi
- postgresql
- resultsdb
- waiverdb
- greenwave

And all the "read" commands of the Bodhi CLI are tested.

bowlofeggs:
On Wed, 2018-11-28 at 15:49 +0000, Aurélien Bompard wrote:
> I don't think the failure on rawhide-python3-diff-cover is caused by
> one of my changes...

Yeah I've been noticing that fail fairly often too. I think something
is wrong with either the coverage report generation or with diff-cover
on Rawhide. This doesn't happen on the pip container tests either.

bowlofeggs:
This miiiiiight get the rawhide diff-cover test working again: https://src.fedoraproject.org/rpms/python-diff-cover/c/354c48310713d859431e7733ac1a37a609b664b4?branch=master

There was something in the upstream changelog that sounded like it would fix some failures with cobertura results that use whitespace. Plus, we were on a way-too-old version anyhow.

.mergify.yml
```diff
@@ -29,6 +29,7 @@ pull_request_rules:
   - status-success=pip-python2-diff-cover
   - status-success=pip-python3-unit
   - status-success=pip-python3-diff-cover
+  - status-success=f27-python2-integration
```
bowlofeggs: Looks like you have it working with Fedora 29 now, so we can update this.

bodhi/client/bindings.py
```diff
@@ -35,7 +35,6 @@
 import re
 import textwrap
 
-from iniparse.compat import ConfigParser
```
bowlofeggs: Nice! Can you create a ```devel``` version in ```docs/user/release_notes.rst``` and add a note about this so I don't forget to document it?

devel/ci/bodhi-ci
```diff
@@ -751,6 +769,72 @@ class UnitJob(Job):
...
+                from the container.
+            archive_path (str): A path on the host to share as a volume into the container for
+                its /results path.
+            pyver (int): Which major version of Python we are testing with.
+        """
+        super(IntegrationJob, self).__init__(*args, **kwargs)
+
+        self.pyver = pyver
+        self._label = '{}-py{}'.format(self._label, pyver)
+        self._command = [sys.executable, '-m', 'pytest', '--no-cov', 'devel/ci/integration/tests/']
```
bowlofeggs, line 826: Won't this be the Python 3 since this script is executed with Python 3? We probably want to use ```pyver``` here.

devel/ci/bodhi-ci
```diff
@@ -886,6 +977,10 @@ def _run_jobs(jobs):
     Args:
         jobs (list): A list of Jobs to run.
     """
+    if not jobs:
+        click.echo("No jobs!")
+        sys.exit(3)
```
bowlofeggs, line 988: Heh, is there a way to get this to happen?

devel/ci/cico.pipeline
```diff
@@ -45,6 +45,9 @@ def configure_node = {
     onmyduffynode 'yum -y install epel-release'
     onmyduffynode 'yum install -y docker python36-click rsync'
     onmyduffynode 'systemctl start docker'
+    // To run the integration testsuite
+    onmyduffynode 'python3.6 -m ensurepip'
+    onmyduffynode 'python3.6 -m pip install \'pytest<4.0\' pytest-cov conu'
```
bowlofeggs: Do the tests not work with ```pytest 4```? Also, we could run the test suite itself inside a container so we don't need to package all the test suite dependencies for EPEL.

devel/ci/integration/bodhi/Dockerfile
```diff
@@ -0,0 +1,125 @@
...
+
+RUN curl -o /etc/yum.repos.d/infra-tags.repo https://infrastructure.fedoraproject.org/cgit/ansible.git/plain/files/common/fedora-infra-tags.repo
+
+# While dnf has a --nodocs, it doesen't have a --docs...
+RUN sed -i '/nodocs/d' /etc/dnf/dnf.conf
+
+
+# The echo works around https://bugzilla.redhat.com/show_bug.cgi?id=1483553 and any other future dnf
+# upgrade bugs.
+#RUN dnf upgrade -y || echo "We are not trying to test dnf upgrade, so ignoring dnf failure."
```
bowlofeggs: Let's drop this instead of commenting it out.

devel/ci/integration/bodhi/Dockerfile
```diff
@@ -0,0 +1,125 @@
...
+    "python3dist(iniparse)" \
+    "python3dist(kitchen)" \
+    "python3dist(python-fedora)" \
+    "python3dist(six)" \
+    \
+    fedmsg \
+    fedmsg-base \
+    git \
+    \
+    postgresql
```
bowlofeggs: Should the Bodhi container has postgresql?

devel/ci/integration/bodhi/Dockerfile
```diff
@@ -0,0 +1,125 @@
...
+
+# To build the documentation
+#    make \
+#    "python3dist(sphinx)" \
+#    "python3dist(cornice_sphinx)" \
+#    "python3dist(sqlalchemy_schemadisplay)" \
```
bowlofeggs: We can drop this too.

devel/ci/integration/bodhi/Dockerfile
```diff
@@ -0,0 +1,125 @@
...
+    python3-dnf \
+    python3-koji \
+    "python3dist(click)" \
+    "python3dist(iniparse)" \
+    "python3dist(kitchen)" \
+    "python3dist(python-fedora)" \
+    "python3dist(six)" \
+    \
+    fedmsg \
+    fedmsg-base \
```
bowlofeggs, line 66: A lot of the above is in common with the existing Fedora 29 unit test Dockerfile. I wonder if we should make a shared container layer that this Dockerfile and the other one use. Then ```bodhi-ci``` can build and tag that layer first, and then this and the other one can ```FROM``` that intermediate layer?

devel/ci/integration/bodhi/Dockerfile
```diff
@@ -0,0 +1,125 @@
...
+WORKDIR /bodhi
+COPY . /bodhi
+
+# Install it
+RUN sed -i -e '/pyramid_debugtoolbar/d' setup.py devel/development.ini.example
+RUN python3 setup.py build && pip3 install .
+
+## build docs
+#RUN make -C docs html
+#RUN mkdir -p /usr/share/doc/bodhi-docs/ && cp -a docs/_build/html /usr/share/doc/bodhi-docs/
```
bowlofeggs: Let's drop these lines too.

devel/ci/integration/bodhi/fedmsg.py
```diff
@@ -0,0 +1,14 @@
+import socket
+config = {
+    'sign_messages': False,
+    'active': True,
+    # 'cert_prefix': 'bodhi',
+    # 'certnames': {'bodhi.{}'.format(socket.gethostname()): 'bodhi'},
+    # TODO: setup a container for the fedmsg relay
```
bowlofeggs, line 4: There's an open pull request from @jeremycline that switches Bodhi to use fedora messaging - I wonder if we can coordinate the two pull requests? https://github.com/fedora-infra/bodhi/pull/2754

devel/ci/integration/tests/conftest.py
```diff
@@ -0,0 +1,123 @@
+from __future__ import absolute_import, unicode_literals
```
bowlofeggs: Let's put a license at the header of this file:

```
# Copyright © 2018 Red Hat, Inc.
#
# This file is part of Bodhi.
#
# This program is free software; you can redistribute it and/or
# modify it under the terms of the GNU General Public License
# as published by the Free Software Foundation; either version 2
# of the License, or (at your option) any later version.
#
# This program is distributed in the hope that it will be useful,
# but WITHOUT ANY WARRANTY; without even the implied warranty of
# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
# GNU General Public License for more details.
#
```

devel/ci/integration/tests/conftest.py
```diff
@@ -0,0 +1,123 @@
...
+        with open(config_path, "w") as config_file:
+            config.write(config_file)
+
+    container.start()
+    docker_backend.d.connect_container_to_network(container.name, docker_network["Id"])
+    # Update the database schema
+    container.execute(["alembic-3", "-c", "/bodhi/alembic.ini", "upgrade", "head"])
+    # we need to wait for the webserver to start serving
+    container.wait_for_port(8080, timeout=-1)
+    # print(docker_backend.d.inspect_container(container.name))
```
bowlofeggs: Let's drop this print statement.

devel/ci/integration/tests/test_bodhi.py
```diff
@@ -0,0 +1,17 @@
+from tests.utils import read_file
```
bowlofeggs, line 19: Let's put a copyright here too.

devel/ci/integration/tests/test_bodhi.py
```diff
@@ -0,0 +1,17 @@
...
+    # GET on /
+    # this is standard `requests.Response`
+    http_response = bodhi_container.http_request(path="/", port=8080)
+    try:
+        assert http_response.ok
+        assert "Fedora Update System" in http_response.text
+    except AssertionError:
+        print(http_response)
+        print(http_response.text)
+        # print(b"".join(bodhi_container.logs()).decode("utf-8"))
```
bowlofeggs: Let's drop this line.

devel/ci/integration/tests/utils.py
```diff
@@ -0,0 +1,63 @@
+import os
```
bowlofeggs, line 19: And a copyright here.

devel/ci/integration/tests/utils.py
```diff
@@ -0,0 +1,63 @@
...
+    with tempfile.TemporaryDirectory() as tempdir:
+        destfile = os.path.join(tempdir, "file")
+        container.copy_from(path, destfile)
+        mode = "rb" if binary else "r"
+        with open(destfile, mode) as fh:
+            yield fh
+
+
+@contextmanager
+def edit_file(container, path):
```
bowlofeggs, line 106: For test utility functions, let's add docblocks that explain their purpose, arguments, argument types, and return types.

devel/ci/integration/resultsdb/Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+  vendor="Fedora Infrastructure" \
+  license="MIT"
+
+RUN yum install -y https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm
+
+# The echo works around https://bugzilla.redhat.com/show_bug.cgi?id=1483553 and any other future yum
+# upgrade bugs.
+RUN yum upgrade -y || echo "We are not trying to test dnf upgrade, so ignoring dnf failure."
+
+# Install Bodhi deps
```
bowlofeggs: We can probably drop this line.

devel/ci/integration/resultsdb/Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+    mod_wsgi \
+    resultsdb \
+    python-psycopg2
+
+# Configuration
+COPY devel/ci/integration/resultsdb/settings.py /etc/resultsdb/settings.py
+COPY devel/ci/integration/resultsdb/httpd.conf /etc/httpd/conf.d/resultsdb.conf
+
+EXPOSE 80
+#ENTRYPOINT /usr/sbin/httpd -DFOREGROUND
```
bowlofeggs: We can probably drop this line too if it will remain unused.

devel/ci/integration/resultsdb/httpd.conf
```diff
@@ -0,0 +1,37 @@
+WSGIDaemonProcess resultsdb user=apache group=apache threads=5
+WSGIScriptAlias /resultsdb /usr/share/resultsdb/resultsdb.wsgi
+WSGISocketPrefix run/wsgi
+
+# this isn't the best way to force SSL but it works for now
+#RewriteEngine On
+#RewriteCond %{HTTPS} !=on
+#RewriteRule ^/resultsdb/admin/?(.*) https://%{SERVER_NAME}/$1 [R,L]
```
bowlofeggs: Let's drop these and other commented lines in this file.

devel/ci/integration/tests/utils.py
```diff
@@ -61,3 +61,17 @@ def edit_file(container, path):
...
+
+def build_container_image(docker_backend, app_name):
+    """Build the container image if necessary."""
+    container_name = "bodhi-ci-{}".format(app_name)
+    if container_name not in [c.name for c in docker_backend.list_images()]:
+        print("Building container image for {}".format(app_name))
+        subprocess.check_call([
+            "docker", "build", "--pull", "-t", container_name,
+            "-f", "devel/ci/integration/{}/Dockerfile".format(app_name), ".",
+        ])
```
bowlofeggs: Why not have ```bodhi-ci``` do this, so it can be done in parallel with all the other build tasks?

devel/ci/cico.pipeline
```diff
@@ -47,7 +47,7 @@ def configure_node = {
     onmyduffynode 'systemctl start docker'
     // To run the integration testsuite
     onmyduffynode 'python3.6 -m ensurepip'
-    onmyduffynode 'python3.6 -m pip install \'pytest<4.0\' pytest-cov conu'
+    onmyduffynode 'python3.6 -m pip install \'pytest<4.0\' pytest-cov conu munch psycopg2'
```
bowlofeggs, line 52: It might be ideal to run the tests in a container too so we can ```dnf install``` these.

devel/ci/integration/tests/test_bodhi_cli.py
```diff
@@ -0,0 +1,247 @@
+import itertools
```
bowlofeggs, line 19: Let's add a copyright here too.

devel/ci/integration/tests/test_bodhi_cli.py
```diff
@@ -0,0 +1,247 @@
...
+import re
+import textwrap
+
+import psycopg2
+
```
bowlofeggs, line 26: No space here for PEP-8.

devel/ci/integration/tests/test_bodhi_cli.py
```diff
@@ -0,0 +1,247 @@
...
+from conu import ConuException
+from munch import Munch
+
+from .utils import replace_file
+
+
+def _run_cli(bodhi_container, args, **kwargs):
```
bowlofeggs, line 33: Let's give this a docblock since it's a utility function.

devel/ci/integration/tests/test_bodhi_cli.py
```diff
@@ -0,0 +1,247 @@
...
+        output = bodhi_container.execute(
+            ["bodhi"] + args + ["--url", "http://localhost:8080"],
+            **kwargs
+        )
+    except ConuException as e:
+        return Munch(exit_code=1, output=str(e))
+    return Munch(exit_code=0, output="".join(line.decode("utf-8") for line in output))
+
+
+def _db_record_to_munch(cursor, record):
```
bowlofeggs, line 55: Let's give this a docblock since it's a utility function.

devel/ci/integration/tests/test_bodhi_cli.py
```diff
@@ -0,0 +1,247 @@
...
+    ]
+    # The bodhi CLI will execute the koji CLI. Replace that executable with
+    # something we can track.
+    koji_mock = "#!/bin/sh\necho TESTING CALL $0 $@\n"
+    with replace_file(bodhi_container, "/usr/bin/koji", koji_mock):
+        result = _run_cli(bodhi_container, cmd)
+    assert result.exit_code == 0
+    for build_id in builds:
+        assert "Downloading packages from {}".format(build_id) in result.output
+        assert "TESTING CALL /usr/bin/koji download-build {}".format(build_id) in result.output
```
bowlofeggs, line 316: Very nice!

devel/ci/integration/tests/utils.py
```diff
@@ -63,6 +64,22 @@ def edit_file(container, path):
         container.copy_to(destfile, path)
 
 
+@contextmanager
+def replace_file(container, path, contents):
```
bowlofeggs, line 124: Let's give this a docblock since it's a utility function.

devel/ci/integration/bodhi/fedmsg.py
```diff
@@ -1,4 +1,4 @@
-import socket
+# import socket
```
bowlofeggs: Let's drop it instead of commenting it out.

devel/ci/integration/greenwave/fedmsg.py
```diff
@@ -0,0 +1,20 @@
+# import socket
```
bowlofeggs: Let's drop this and other commented lines.

devel/ci/integration/tests/fixtures/backend.py
```diff
@@ -0,0 +1,20 @@
+from __future__ import absolute_import, unicode_literals
```
bowlofeggs, line 19: Let's add a copyright here.

devel/ci/integration/tests/fixtures/bodhi.py
```diff
@@ -0,0 +1,95 @@
+from __future__ import absolute_import, unicode_literals
```
bowlofeggs: Let's add a copyright here.

devel/ci/integration/tests/fixtures/bodhi.py
```diff
@@ -0,0 +1,95 @@
...
+from six.moves.configparser import ConfigParser
+
+from ..utils import make_db_and_user, edit_file
+
+
+@pytest.fixture(scope="session")
+def bodhi_container(
+    docker_backend, docker_network, db_container, resultsdb_container,
+    waiverdb_container, greenwave_container
+):
```
bowlofeggs, line 33: Let's give this a docblock since it's a utility function.

devel/ci/integration/tests/fixtures/db.py
```diff
@@ -0,0 +1,20 @@
+import os
+import pytest
+from conu.backend.docker.container_parameters import DockerContainerParameters
+
+
+@pytest.fixture(scope="session")
+def db_container(docker_backend, docker_network):
```
bowlofeggs, line 25: Let's give this a docblock since it's a utility function.

devel/ci/integration/tests/fixtures/db.py
```diff
@@ -0,0 +1,20 @@
+import os
```
bowlofeggs, line 19: Copyright.

devel/ci/integration/tests/fixtures/greenwave.py
```diff
@@ -0,0 +1,26 @@
+from __future__ import absolute_import, unicode_literals
```
bowlofeggs: I'll stop mentioning copyright and docblocks - let's just make sure all the things that need them have them.

bowlofeggs (changes requested):
Man this is looking nice @abompard! I'm very excited about this.

.mergify.yml
abompard: Oops yeah I'll fix that and the line in `cico.pipeline`

devel/ci/bodhi-ci
abompard: Actually the python bin can come from a virtualenv, so just using `python` or `python-pyver` isn't sufficient. Is there a problem with `sys.executable`?

devel/ci/bodhi-ci
abompard: Yeah, if you ask for integration testing only and for an unsupported fedora release version.

devel/ci/cico.pipeline
abompard: Yeah it failed with pytest 4 but I don't remember why, it may have been because I used to run it with python2. I'll check again.
Running the testsuite inside a container means that this container must be able to create other containers. Is that possible?

devel/ci/integration/bodhi/Dockerfile
abompard: Yeah because I'm using the `psql` client (it's not `postgresql-server`, just the client) to load the database dump. I could probably do differently if necessary.

devel/ci/integration/bodhi/Dockerfile
abompard: That's more Docker knowledge than I currently have, but I agree it would be better, there would be less chances of the specfiles going out of sync.

devel/ci/cico.pipeline
abompard: Same question as above: can containers run other containers?

devel/ci/bodhi-ci
bowlofeggs: Won't ```sys.executable``` reference the Python executable that ```bodhi-ci``` was run with? This means that the ```-p``` flag won't work, right?

devel/ci/cico.pipeline
bowlofeggs: Ah yeah I guess that could be a problem, though I am not sure. Hmm. I guess we can just go this direction for now.

devel/ci/integration/bodhi/Dockerfile
bowlofeggs: Nah, I think that's OK.

devel/ci/integration/bodhi/Dockerfile
bowlofeggs: We can always refactor it later I suppose. I had been thinking about doing something similar for the existing jobs anyway.

devel/ci/cico.pipeline
bowlofeggs: I don't know, but I'm comfortable with the pip installs for now.

bowlofeggs:
There's a conflict on ```bodhi-ci``` - sorry about that!

devel/ci/bodhi-ci
bowlofeggs: Ah, this would be the Python that executes the test suite itself, not the version that executes Bodhi. I think my mind was just in a weird place. I think this is OK.

bowlofeggs:
It might be good to test Python 3 instead of Python 2 if we are only going to do one Python version (which is fine), as I plan to drop Python 2 support soon: https://github.com/fedora-infra/bodhi/issues/2759

Production currently runs on Python 3 on Fedora 29.

abompard:
I just rebased. Yeah the plan is to run with Python 3 now that prod is on Python 3.

devel/ci/cico.pipeline
abompard: I tried with Pytest 4 again, it fails because pytest-cov tries to use a deprecated API (`config.warn`) and there is no pytest-cov update available.

abompard:
OK, the tests pass now and I think I fixed all the issues you raised, besides the factorization of the Dockerfiles. Also, I didn't write the doctests because I first wanted to make sure you're OK with the direction this is taking.
I think it's ready for re-review now, thanks a lot for looking at all that.

abompard:
OK, the tests pass now and I think I fixed all the issues you raised, besides the factorization of the Dockerfiles. Also, I didn't write the doctests because I first wanted to make sure you're OK with the direction this is taking.
I think it's ready for re-review now, thanks a lot for looking at all that.

abompard:
OK, the tests pass now and I think I fixed all the issues you raised, besides the factorization of the Dockerfiles. Also, I didn't write the doctests because I first wanted to make sure you're OK with the direction this is taking.
I think it's ready for re-review now, thanks a lot for looking at all that.

abompard:
OK, the tests pass now and I think I fixed all the issues you raised, besides the factorization of the Dockerfiles. Also, I didn't write the doctests because I first wanted to make sure you're OK with the direction this is taking.
I think it's ready for re-review now, thanks a lot for looking at all that.

bowlofeggs:
I was doing some yard work today and had the thought that it might be nice to separate the running of the dependent containers from the test suite. One of the things that has always been difficult about Bodhi development is that it depends on so many other services. This pull request adds some of those services, but in a way that only can be used by the integration test suite. It would be nice if I could ask ```bodhi-ci``` to launch the dependencies it knows about so that I can have my local development Bodhi use local dependencies.

Anyways, that's just a thought and not a requirement for this PR. IMO, we can merge this PR as is and maybe just consider going that direction as a refactor later, if we want to.

abompard:
OK let's merge this PR and refactor later, that way we'll avoid possible conflicts with the devel branch.

devel/ci/integration/greenwave/Dockerfile
```diff
@@ -0,0 +1,25 @@
+FROM quay.io/factory2/greenwave:prod
+MAINTAINER "Aurelien Bompard" <abompard@fedoraproject.org>
```
cverna: just a small nitpick here, the `MAINTAINER` instruction is depreacted, instead a `maintainer` label should be used

```
LABEL maintainer="Aurelien Bompard <abompard@fedoraproject.org>"
```
That will also save one layer in the image making it a little faster to build :)

devel/ci/integration/resultsdb/Dockerfile
```diff
@@ -0,0 +1,26 @@
+FROM centos:7
+MAINTAINER "Aurelien Bompard" <abompard@fedoraproject.org>
```
cverna: same comment as above

devel/ci/integration/waiverdb/Dockerfile
```diff
@@ -0,0 +1,22 @@
+FROM quay.io/factory2/waiverdb:prod
+MAINTAINER "Aurelien Bompard" <abompard@fedoraproject.org>
```
cverna: same comment as above :)

bowlofeggs:
I'm going to rebase your pull request to fix the mergify config, and hopefully get CI to run.

devel/ci/cico.pipeline
bowlofeggs: Since ```conu``` is a required dependency, can you add it to ```devel/ansible/roles/bodhi/tasks/main.yml``` so the Vagrant environment will be able to run it?

bowlofeggs (changes requested):
Two things:

* Can you squash some of your commits? I don't mind multi-commit PRs, but I'd prefer each commit to be atomic within the set. You can also make it one commit if you prefer.
* I see a failing test when I run this locally:

```
     f29-integration     :  FAILED  (exited with code: 1)


     f29-integration     	============================= test session starts ==============================
     f29-integration     	platform linux -- Python 3.7.1, pytest-3.6.4, py-1.5.4, pluggy-0.6.0
     f29-integration     	rootdir: /home/vagrant/bodhi, inifile: setup.cfg
     f29-integration     	plugins: cov-2.5.1
     f29-integration     	collected 7 items
     f29-integration     	
     f29-integration     	devel/ci/integration/tests/test_bodhi.py E
     f29-integration     	
     f29-integration     	==================================== ERRORS ====================================
     f29-integration     	_______________________ ERROR at setup of test_get_root ________________________
     f29-integration     	
     f29-integration     	self = <docker.api.client.APIClient object at 0x7fda75d95eb8>
     f29-integration     	response = <Response [500]>
     f29-integration     	
     f29-integration     	    def _raise_for_status(self, response):
     f29-integration     	        """Raises stored :class:`APIError`, if one occurred."""
     f29-integration     	        try:
     f29-integration     	>           response.raise_for_status()
     f29-integration     	
     f29-integration     	/usr/lib/python3.7/site-packages/docker/api/client.py:229: 
     f29-integration     	_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
     f29-integration     	
     f29-integration     	self = <Response [500]>
     f29-integration     	
     f29-integration     	    def raise_for_status(self):
     f29-integration     	        """Raises stored :class:`HTTPError`, if one occurred."""
     f29-integration     	    
     f29-integration     	        http_error_msg = ''
     f29-integration     	        if isinstance(self.reason, bytes):
     f29-integration     	            # We attempt to decode utf-8 first because some servers
     f29-integration     	            # choose to localize their reason strings. If the string
     f29-integration     	            # isn't utf-8, we fall back to iso-8859-1 for all other
     f29-integration     	            # encodings. (See PR #3538)
     f29-integration     	            try:
     f29-integration     	                reason = self.reason.decode('utf-8')
     f29-integration     	            except UnicodeDecodeError:
     f29-integration     	                reason = self.reason.decode('iso-8859-1')
     f29-integration     	        else:
     f29-integration     	            reason = self.reason
     f29-integration     	    
     f29-integration     	        if 400 <= self.status_code < 500:
     f29-integration     	            http_error_msg = u'%s Client Error: %s for url: %s' % (self.status_code, reason, self.url)
     f29-integration     	    
     f29-integration     	        elif 500 <= self.status_code < 600:
     f29-integration     	            http_error_msg = u'%s Server Error: %s for url: %s' % (self.status_code, reason, self.url)
     f29-integration     	    
     f29-integration     	        if http_error_msg:
     f29-integration     	>           raise HTTPError(http_error_msg, response=self)
     f29-integration     	E           requests.exceptions.HTTPError: 500 Server Error: Internal Server Error for url: http+docker://localhost/v1.26/containers/d98db33382942419d839daa6cd2315394bc995b9550a12365532ee36961e6905/start
     f29-integration     	
     f29-integration     	/usr/lib/python3.7/site-packages/requests/models.py:940: HTTPError
     f29-integration     	
     f29-integration     	During handling of the above exception, another exception occurred:
     f29-integration     	
     f29-integration     	docker_backend = <conu.backend.docker.backend.DockerBackend object at 0x7fda75d95d68>
     f29-integration     	docker_network = {'Id': '080b46a68cd5e32b3567f29861532243a63679250a8649373ef9a426abf43759', 'Warning': ''}
     f29-integration     	
     f29-integration     	    @pytest.fixture(scope="session")
     f29-integration     	    def db_container(docker_backend, docker_network):
     f29-integration     	        """Fixture preparing and yielding a PostgreSQL container.
     f29-integration     	    
     f29-integration     	        This container can be used by other apps to store data.
     f29-integration     	    
     f29-integration     	        Args:
     f29-integration     	            docker_backend (conu.DockerBackend): The Docker backend (fixture).
     f29-integration     	            docker_network (str): The Docker network ID (fixture).
     f29-integration     	    
     f29-integration     	        Yields:
     f29-integration     	            conu.DockerContainer: The PostgreSQL container.
     f29-integration     	        """
     f29-integration     	        image = docker_backend.ImageClass(
     f29-integration     	            os.environ.get("BODHI_INTEGRATION_POSTGRESQL_IMAGE", "postgres"),
     f29-integration     	            tag="latest"
     f29-integration     	        )
     f29-integration     	        container = image.run_via_api(
     f29-integration     	            DockerContainerParameters(name="db", port_mappings={5432: 5432})
     f29-integration     	        )
     f29-integration     	>       container.start()
     f29-integration     	
     f29-integration     	devel/ci/integration/tests/fixtures/db.py:44: 
     f29-integration     	_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
     f29-integration     	/usr/lib/python3.7/site-packages/conu/backend/docker/container.py:457: in start
     f29-integration     	    self.d.start(self.get_id())
     f29-integration     	/usr/lib/python3.7/site-packages/docker/utils/decorators.py:19: in wrapped
     f29-integration     	    return f(self, resource_id, *args, **kwargs)
     f29-integration     	/usr/lib/python3.7/site-packages/docker/api/container.py:1066: in start
     f29-integration     	    self._raise_for_status(res)
     f29-integration     	/usr/lib/python3.7/site-packages/docker/api/client.py:231: in _raise_for_status
     f29-integration     	    raise create_api_error_from_http_exception(e)
     f29-integration     	_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
     f29-integration     	
     f29-integration     	e = HTTPError('500 Server Error: Internal Server Error for url: http+docker://localhost/v1.26/containers/d98db33382942419d839daa6cd2315394bc995b9550a12365532ee36961e6905/start')
     f29-integration     	
     f29-integration     	    def create_api_error_from_http_exception(e):
     f29-integration     	        """
     f29-integration     	        Create a suitable APIError from requests.exceptions.HTTPError.
     f29-integration     	        """
     f29-integration     	        response = e.response
     f29-integration     	        try:
     f29-integration     	            explanation = response.json()['message']
     f29-integration     	        except ValueError:
     f29-integration     	            explanation = (response.content or '').strip()
     f29-integration     	        cls = APIError
     f29-integration     	        if response.status_code == 404:
     f29-integration     	            if explanation and ('No such image' in str(explanation) or
     f29-integration     	                                'not found: does not exist or no pull access'
     f29-integration     	                                in str(explanation) or
     f29-integration     	                                'repository does not exist' in str(explanation)):
     f29-integration     	                cls = ImageNotFound
     f29-integration     	            else:
     f29-integration     	                cls = NotFound
     f29-integration     	>       raise cls(e, response=response, explanation=explanation)
     f29-integration     	E       docker.errors.APIError: 500 Server Error: Internal Server Error ("driver failed programming external connectivity on endpoint db (4ca44442e9c4db47cc80073a24ae1002fe4d3bd0ff85ef227efa9c268194fdfe): Error starting userland proxy: listen tcp 0.0.0.0:5432: bind: address already in use")
     f29-integration     	
     f29-integration     	/usr/lib/python3.7/site-packages/docker/errors.py:31: APIError
```

abompard:
I've squashed to a single commit, only leaving out the commits that are independent from the integration testing (and can be cherry-picked in Bodhi's mainline).
I've also added conu to Vagrant.
The error you get is because there's already a PostgreSQL server running in your Vagrant env, so Docker can't bind the 5432 port from the PostgreSQL container. I've removed the port mapping because we actually don't need that port to be exposed to the outside, only inside the containers' network. It should work now.

bowlofeggs:
Cool, it looks good to me. I'm going to rebase and force push to get it to run the tests in CI since something seems to have gone weird there.

Once this is merged, we should see if we can get these tests to run on all the releases. I'm about to drop Python 2, so I don't think we need to worry about testing it on Python 2, but I would like to test it on F28 and Rawhide and pip.

bowlofeggs (approved):
Thanks!

Next, let's see if we can get the tests to run across the various releases of Fedora and pip.

PR #2725, opened 2018-11-09
Set the Bodhi team as approvers of pull requests.

bowlofeggs:

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -6,7 +6,7 @@ pull_request_rules:
       strict: true
   conditions:
   - label!=no-mergify
-  - '#approved-reviews-by>=1'
+  - approved-reviews-by=@bodhi
   - status-success=DCO
   - status-success=f27-docs
   - status-success=f27-flake8


puiterwijk:
@bowlofeggs I applied the following diff:

````
diff --git a/.mergify.yml b/.mergify.yml
index c5fd868d..7100459f 100644
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -6,7 +6,7 @@ pull_request_rules:
       strict: true
   conditions:
   - label!=no-mergify
-  - '#approved-reviews-by=@fedora-infra/bodhi'
+  - approved-reviews-by=@bodhi
   - status-success=DCO
   - status-success=f27-docs
   - status-success=f27-flake8
````

The `#` removal is because we are no longer trying to look at the number of items, and the `fedora-infra/` removal is because it seemingly doesn't like the `/` character in there, so I guess it can only look against teams in the same org.

puiterwijk (approved):

bowlofeggs:
This got merged as https://github.com/fedora-infra/bodhi/commit/41c794e4a6f452b92e8146af126d277055fe34b6

PR #2729, opened 2018-11-11
Set a safer delay for triggering queries.

mattiaverga:
Bloodhound's default delay for triggering network requests is 300ms which is quite low if someone types characters slowly.
Raising it to 600 should be safer and avoid unnecessary network bandwidth usage without harm user experience.

Also increase minimum characters from 2 to 3 to starting search in the search bar. This is because I noticed that searching for updates with few characters needs a lot of server resources... increasing minimum character to 3 makes things only a little better.


bodhi/server/static/js/search.js
```diff
@@ -65,7 +69,7 @@ $(document).ready(function() {
     $('#bloodhound .typeahead').typeahead({
         hint: true,
         highlight: true,
-        minLength: 2,
+        minLength: 3,
```
bowlofeggs: There are 55 packages with names that are less than 3 characters (examples: ed and qt), and this change makes it a bit hard to find updates that match them since pressing enter doesn't sort the results that are an exact match to the top like the search box does. I haven't observed performance problems with the setting as is, have you?

mattiaverga: Oh, I haven't considered those packages... I'm going to revert this change.

The performance problem I see is in the time needed to search for updates (and to a lesser extent in the search for overrides). But that's because the query is made looking for `%queryterm%`, which with few characters returns a huge amount of results (for example, with 'ed' typed in the search bar, we get nearly every update in the database, since it matches 'FedORA-' update aliases.

bowlofeggs:
On Wed, 2018-12-05 at 09:32 -0800, Mattia Verga wrote:
> The performance problem I see is in the time needed to search for
> updates (and to a lesser extent in the search for overrides). But
> that's because the query is made looking for %queryterm%, which with
> few characters returns a huge amount of results (for example, with
> 'ed' typed in the search bar, we get nearly every update in the
> database, since it matches 'FedORA-' update aliases.

Ha yeah, that's a funny example.

Bodhi does have a lot of slow queries. One thing I've long wanted to do
is to take the time to analyze some of its common queries to see what
we can do to speed them up. In some cases, I bet a simple index would
be all that is needed.

One other solution I can think of to the problem you wanted to solve
here is to make it so searches that press enter sort exact matches to
the top of the results (the search bar's JS thingy does this, but not
the results you see when pressing enter). That way, "ed" wouldn't send
a query in the JS, but the ed package could still be found by pressing
enter. This might be a lot of work though, and would also mean that we
would need to change the search page that happens when you press enter
(because it only searches updates, where the JS search bar searches
users and packages too).

bowlofeggs (approved):
Thanks!

bowlofeggs:

bowlofeggs:

bowlofeggs:

PR #2732, opened 2018-11-12
Use Python 3 when running Bodhi in Vagrant.

bowlofeggs:


bowlofeggs:
The failing rawhide tests are not due to this pull request, and are fixed in https://github.com/fedora-infra/bodhi/pull/2734.

puiterwijk (approved):

PR #2733, opened 2018-11-13
Default to a JSON response when no accept header is provided.

### LINKED ISSUE #2731: Staging is currently returning HTML to bodhi CLI requests
I am not sure why this is happening, but it seems that staging Bodhi is currently returning HTML to CLI requests. This also happens to requests with ```http``` or ```curl```.

I recall a problem with the unit tests where they would sometimes receive HTML when they didn't explicitly use a request header to ask for a JSON response once we started testing under Python 3. We ended up adjusting the tests to pass that header since this did not seem to happen when serving Bodhi with ```pserve-3```.

It turns out that there really is some problem that seems related to Python 3 since staging Bodhi started doing this same thing.

bowlofeggs:
fixes #2731

### DIFF
--- a/bodhi/server/webapp.py
+++ b/bodhi/server/webapp.py
@@ -50,13 +50,25 @@ def _complete_database_session(request):
 @subscriber(NewRequest)
 def _prepare_request(event):
     """
-    Add callbacks onto every new request.
+    Prepare each incoming request to Bodhi.
 
-    This function adds a callback to clean up the database session when the request is finished.
+    This function does two things:
+        * If requests do not have an Accept header, or if their Accept header is "*/*", it sets the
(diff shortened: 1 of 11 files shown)


bowlofeggs:
The failing rawhide tests are not due to this pull request, and are fixed in #2734.

bowlofeggs:
@frenzymadness you might be interested in this patch - if you recall, there was a strange thing where Bodhi seemed to return a different content type for each request on Python 3, but not for Python 2. It turns out that Pyramid actually documents that requests with no ```Accept``` header or with a non-matching header can be served by any view that otherwise matches. I don't know why this behaviour changes between Python 2 and 3, but we were certainly relying on undefined behaviour before. With this patch, we explicitly default to JSON which I think is a sane approach.

puiterwijk (approved):

PR #2740, opened 2018-11-13
Convert subprocess.communicate()'s output to str in the composer.

### LINKED ISSUE #2738: TypeError: a bytes-like object is required, not 'str'
The composer has a Python 3 compatibility issue:

```
  File "/usr/lib/python3.7/site-packages/bodhi/server/consumers/masher.py", line 339, in run 
    self.work()
  File "/usr/lib/python3.7/site-packages/bodhi/server/consumers/masher.py", line 410, in work
    self._compose_updates()
  File "/usr/lib/python3.7/site-packages/bodhi/server/consumers/masher.py", line 946, in _compose_updates
    self._wait_for_pungi(pungi_process)
  File "/usr/lib/python3.7/site-packages/bodhi/server/consumers/masher.py", line 1235, in _wait_for_pungi
    for line in out.split('\n'):
TypeError: a bytes-like object is required, not 'str'
```

bowlofeggs:
subprocess returns bytes in Python 3, and so we need to convert
the output to a str before using it.

fixes #2738

### DIFF
--- a/bodhi/server/consumers/masher.py
+++ b/bodhi/server/consumers/masher.py
@@ -1222,6 +1222,8 @@ class PungiComposerThread(ComposerThread):
             return
         self.log.info('Waiting for pungi process to finish')
         out, err = pungi_process.communicate()
+        out = out.decode()
+        err = err.decode()
         self.devnull.close()
         if pungi_process.returncode != 0:
             self.log.error('Pungi exited with exit code %d', pungi_process.returncode)
--- a/bodhi/tests/server/consumers/test_masher.py
+++ b/bodhi/tests/server/consumers/test_masher.py
@@ -268,7 +268,7 @@ Compose dir: %s
 That was the actual one''' % mash_dir
 
             fake_popen = mock.MagicMock()
-            fake_popen.communicate = lambda: (fake_stdout, 'hello')
+            fake_popen.communicate = lambda: (fake_stdout.encode(), b'hello')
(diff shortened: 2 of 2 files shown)


bowlofeggs:
The [rawhide-build failure](https://ci.centos.org/job/bodhi-pipeline/job/PR-2740/1/execution/node/86/log/) is due to [rhbz#1649547](https://bugzilla.redhat.com/show_bug.cgi?id=1649547) and can be ignored.

puiterwijk (approved):

owtaylor (approved):
This looks like it should work to me (assuming that you have dropped Python 2 support.) 

My personal preference would be to add to the args of subprocess.Popen encoding="UTF-8" (Python-3 only) or universal_newlines=True (hack that gives you string output for both Python-2 and Python-3)... but it's really 6 one-way half-a-dozen the other.

bowlofeggs:
@owtaylor we actually haven't dropped Python 2 support just yet, but I plan to "soon".

owtaylor:
> @owtaylor we actually haven't dropped Python 2 support just yet, but I plan to "soon".

Ah, forgot that b'' syntax support was backported into Python 2 - looks like it should work there too.

bowlofeggs:
@owtaylor yeah I think it will work fine on Python 2 as well. In any case, Fedora's production will be Python 3 when this patch is deployed (I found this problem while testing Bodhi in staging on Python 3.)

PR #2749, opened 2018-11-15
Remove some cruft that was needed to run on EL 7.

bowlofeggs:
EPEL 7 had an old verion of WebOb and so there were some Python
hacks to get Bodhi to use a newer WebOb from EPEL. Bodhi hasn't
supported EPEL in a long time, so we do not need these hacks
anymore, and they present a problem in Rawhide where we want to
ship the Python 2 version of the client but not the server.


jeremycline (approved):
:shipit:

puiterwijk (approved):
👍

PR #2751, opened 2018-11-16
Mark Bodhi as working with Python 3 in setup.py.

bowlofeggs:


cverna (approved):
:fireworks:

PR #2754, opened 2018-11-19
Convert Bodhi to publish messages via AMQP

jeremycline:
This converts Bodhi to use fedora-messaging to publish to an AMQP message broker. It also sets up a RabbitMQ message broker in the Vagrant environment.

This pull request does the minimal amount necessary to publish via AMQP. It does not define message schema, nor does it attempt to convert the existing fedmsg consumers to the fedora-messaging consumer API. These will be done in follow-up pull requests.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -20,6 +20,10 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
  # if you have other services listening on your host's port 80.
  config.vm.network "forwarded_port", guest: 6543, host: 6543
 
+ # Forward traffic on the host to the RabbitMQ management UI on the guest.
+ # This allows developers to view message queues at http://localhost:15672/
+ config.vm.network "forwarded_port", guest: 15672, host: 15672
+
  # This is an optional plugin that, if installed, updates the host's /etc/hosts
  # file with the hostname of the guest VM. In Fedora it is packaged as
  # ``vagrant-hostmanager``
--- a/bodhi/server/notifications.py
+++ b/bodhi/server/notifications.py
@@ -23,6 +23,7 @@ import logging
 import socket
 
 from sqlalchemy import event
+from fedora_messaging import api, exceptions as fml_exceptions
 import fedmsg
 import fedmsg.config
 import fedmsg.encoding
@@ -97,7 +98,68 @@ def send_fedmsgs_after_commit(session):
             session.info['fedmsg'][topic] = []
 
 
+@event.listens_for(Session, 'after_commit')
+def send_messages_after_commit(session):
+    """
+    Send messages via AMQP after a database commit occurs.
+
+    Args:
+        session (sqlalchemy.orm.session.Session): The session that was committed.
+    """
+    if 'messages' in session.info:
+        for m in session.info['messages']:
+            try:
+                api.publish(m)
+            except fml_exceptions.BaseException:
+                # In the future we should handle errors more gracefully
+                _log.exception("An error occurred publishing %r after a database commit", m)
+        session.info['messages'] = []
+
+
 def publish(topic, msg, force=False):
+    """
+    Send a message via AMQP or fedmsg.
+
+    This is used to send a message to fedmsg if "fedmsg_enabled" is true in the
+    application configuration, otherwise it will use AMQP.
+
+    Args:
+        topic (str): The topic suffix. The "bodhi" prefix is applied, along with
+            the "topic_prefix" and "environment" settings from fedmsg.
+        msg (dict): The message body to send.
+        force (bool): If False (the default), the message is only sent after the
+            currently active database transaction successfully commits. If true,
+            the messages is sent immediately.
+    """
+    if bodhi.server.config.config.get('fedmsg_enabled'):
+        _fedmsg_publish(topic, msg, force)
+        return
+
+    # Dirty, nasty hack that I feel shame for: fedmsg modifies messages quietly
+    # if they have objects with __json__ methods on them. For now, copy that
+    # behavior. In the future, callers should pass fedora_messaging.api.Message
+    # sub-classes or this whole API should go away.
+    body = fedmsg.encoding.loads(fedmsg.encoding.dumps(msg))
+
+    # Dirty, nasty hack #2: fedmsg mangles the topic in various ways. Duplicate
+    # that behavior here for now. In the future callers should pass proper topics.
+    fedmsg_conf = fedmsg.config.load_config()
+    topic = "{prefix}.{env}.bodhi.{t}".format(
+        prefix=fedmsg_conf["topic_prefix"], env=fedmsg_conf["environment"], t=topic)
+
+    message = api.Message(topic=topic, body=body)
+    if force:
+        api.publish(message)
+        return
+
+    session = Session()
+    if 'messages' not in session.info:
+        session.info['messages'] = []
+    session.info['messages'].append(message)
+    _log.debug('Queuing message %r for delivery on session commit', message.id)
+
+
+def _fedmsg_publish(topic, msg, force):
     """Publish a message to fedmsg.
 
     By default, messages are not sent immediately, but are queued in a
@@ -114,10 +176,6 @@ def publish(topic, msg, force=False):
         force (bool): If True, send the message immediately. Else, queue the message to be sent
             after the current database transaction is committed.
     """
-    if not bodhi.server.config.config.get('fedmsg_enabled'):
-        _log.warning("fedmsg disabled.  not sending %r" % topic)
-        return
-
     # Initialize right before we try to publish, but only if we haven't
     # initialized for this thread already.
     if not fedmsg_is_initialized():
--- a/bodhi/tests/server/consumers/test_masher.py
+++ b/bodhi/tests/server/consumers/test_masher.py
@@ -32,6 +32,8 @@ except ImportError:
     IncompleteRead = None
 
 from click import testing
+from fedora_messaging import api
+from fedora_messaging.testing import mock_sends
 import mock
 import six
 import six.moves.urllib.parse as urlparse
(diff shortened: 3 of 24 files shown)


bowlofeggs:
I'm going to rebase this bad boy because the f27 tests are no longer run and don't work.

devel/ansible/roles/bodhi/tasks/main.yml
```diff
@@ -90,6 +91,7 @@
       - python3-pylibravatar
       - python3-pyramid
       - python3-pyramid-fas-openid
+      - python3-pyramid-tm
```
bowlofeggs: I am able to run the tests without pyramid-tm, but I did have to remove it from my ```development.ini``` file. Is it possible you need to refresh your ```development.ini``` file from ```devel/development.ini.example```? You can just delete it and re-vagrant-up it.

Related: https://github.com/fedora-infra/bodhi/pull/2755

Vagrantfile
```diff
@@ -20,6 +20,10 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
  # if you have other services listening on your host's port 80.
  config.vm.network "forwarded_port", guest: 6543, host: 6543
 
+ # Forward traffic on the host to the RabbitMQ management UI on the guest.
+ # This allows developers to view message queues at http://localhost:15672/
+ config.vm.network "forwarded_port", guest: 15672, host: 15672
```
bowlofeggs, line 25: We should document the default username/password in ```docs/developer/vagrant.rst```.

bodhi/server/notifications.py
```diff
@@ -97,7 +98,68 @@ def send_fedmsgs_after_commit(session):
...
+                # In the future we should handle errors more gracefully
+                _log.exception("An error occurred publishing %r after a database commit", m)
+        session.info['messages'] = []
+
+
 def publish(topic, msg, force=False):
+    """
+    Send a message via AMQP or fedmsg.
+
+    This use used to send a message to fedmsg if "fedmsg_enabled" is true in the
```
bowlofeggs:
```suggestion
    This is used to send a message to fedmsg if "fedmsg_enabled" is true in the
```

bodhi/server/notifications.py
```diff
@@ -97,7 +98,68 @@ def send_fedmsgs_after_commit(session):
...
+            the messages is sent immediately.
+    """
+    if bodhi.server.config.config.get('fedmsg_enabled'):
+        _fedmsg_publish(topic, msg, force)
+        return
+
+    # Dirty, nasty hack that I feel shame for: fedmsg modifies messages quietly
+    # if they have objects with __json__ methods on them. For now, copy that
+    # behavior. In the future, callers should pass fedora_messaging.api.Message
+    # sub-classes or this whole API should go away.
```
bowlofeggs, line 141: haha, I guess I should file a ticket to switch to this format too.

bodhi/server/notifications.py
```diff
@@ -97,7 +98,68 @@ def send_fedmsgs_after_commit(session):
...
+    body = fedmsg.encoding.loads(fedmsg.encoding.dumps(msg))
+
+    # Dirty, nasty hack #2: fedmsg mangles the topic in various ways. Duplicate
+    # that behavior here for now. In the future callers should pass proper topics.
+    fedmsg_conf = fedmsg.config.load_config()
+    topic = "{prefix}.{env}.bodhi.{t}".format(
+        prefix=fedmsg_conf["topic_prefix"], env=fedmsg_conf["environment"], t=topic)
```
bowlofeggs, line 148: I wonder if I should just fix this with Bodhi 4 by declaring that Bodhi's message topic prefixes are changing? Would you advise that I just remove the prefix and env and just start the topic with "bodhi"? Or maybe we could make the topic prefix be a Bodhi setting in Bodhi 4 instead of loading the fedmsg config. It'd be nice to fully drop fedmsg in Bodhi 4.

bodhi/tests/server/consumers/test_masher.py
```diff
@@ -2666,7 +2668,8 @@ def test_expires_compose_updates(self):
         t.db = self.db
         t.id = getattr(self.db.query(Release).one(), '{}_tag'.format('stable'))
 
-        t.perform_gating()
+        with mock_sends(api.Message):
```
bowlofeggs, line 2671: Ooooh that seems nice! You guys thought of everything.

bodhi/tests/server/test_notifications.py
```diff
@@ -127,13 +128,28 @@ class TestPublish(base.BaseTestCase):
...
+        notifications.publish('demo.topic', {'such': 'important'})
+
+        self.assertIn('messages', session.info)
+        for expected, actual in zip(expected_msgs, session.info['messages']):
+            self.assertEqual(expected, actual)
         self.assertEqual(0, mock_init.call_count)
 
+    @mock.patch.dict('bodhi.server.config.config', {'fedmsg_enabled': False})
+    def test_fedmsg_publish_off_force(self, mock_init):
+        """Assert when fedmsg publish is off, fedora-messaging messages are queued."""
```
bowlofeggs:
```suggestion
        """Assert that fedora-messaging messages respect the force flag."""
```

bodhi/tests/server/test_notifications.py
```diff
@@ -127,13 +128,28 @@ class TestPublish(base.BaseTestCase):
...
+        expected = api.Message(topic='org.fedoraproject.dev.bodhi.demo.topic',
+                               body={u'such': 'important'})
+
+        with fml_testing.mock_sends(expected):
```
bowlofeggs, line 150: Man that's a nice API.

devel/ansible/roles/bodhi/files/print-messages.service
```diff
@@ -0,0 +1,11 @@
+[Unit]
+Description=Print messages sent to the amq.topic exchange
+After=network-online.target
+Wants=network-online.target
+
+[Service]
+User=vagrant
+ExecStart=/usr/bin/fedora-messaging consume --callback="fedora_messaging.example:printer"
```
bowlofeggs, line 8: Might be cool to add a .bashrc alias like ```bmessaging``` or something that tails these logs. If you opt for that, don't forget to add it to the motd too ☺

bowlofeggs (changes requested):
This looks great!

Does fedora messaging have something like fedmsg-hub, or will we just convert fedmsg-hub to use fedora-messaging instead at some point in the future? I'm curious if we should do something to make the composer and updates handlers work with it.

bowlofeggs:
To get the tests to pass, you can add ```fedora-messaging``` to the Dockerfiles in ```devel/ci/```. For releases that don't have ```fedora-messaging``` (IIRC, it's only in F29+?), feel free to just pip install it.

bowlofeggs:
Oh, and if you want to run the CI tests locally to more quickly test your ```Dockerfile``` changes, check out the Vagrant alias ```bci --help```.

bodhi/server/notifications.py
jeremycline: Yeah, I figured doing it in bits and pieces would be easiest. I don't mind filing tickets for the remaining work items (defining schema, using them, moving consumers over to ``fedora-messaging consume`` from fedmsg-hub)

bodhi/server/notifications.py
jeremycline: It would be very nice to fix the topics, but I think maybe the best migration path is something like:

1. Add new message types with reasonable topics while still publishing old messages with old topics.
2. Publish both old and new messages.
3. Make an announcement about the old messages going away, convert all known consumers of those messages to the new message
4. Drop the old messages.

There's not really anything in fedora-messaging to help you in this, although if we find a common pattern that most apps would use we should definitely add that.

As for what the topic should be, I'd recommend starting with "bodhi" and getting from more general to less general. Something like "bodhi.update.created.<dist>.<pkgname>" could work. Topics can (and in many cases should) be dynamic. There a few restrictions though - they need to be 255 characters (or octets, possibly) or less.

edit: and yeah, I thought about making it be a bodhi setting, if you'd prefer that I can make that happen in this PR. That way dropping fedmsg is simpler for Bodhi 4

bowlofeggs (approved):
Thanks!

bowlofeggs:
Hmm, mergify seems stuck. I'll rebase.

bowlofeggs:
@abompard It seems that the integration tests are failing here, but I am unsure why:

https://ci.centos.org/job/bodhi-pipeline/job/PR-2754/17/execution/node/99/log/

It does appear to be reproducible locally.

bowlofeggs:
Aha! It's buried in the middle of LOTS of output - looks like we just need to add fedora_messaging to the integration test container too. @jeremycline I'll take care of that for you.

bowlofeggs:
Hmm, now there seems to be failing unit tests. This may be due to rebasing on other patches, not sure yet. I'm investigating.

bowlofeggs:
I am applying this patch to fix the unit tests:

```
index 79932ef0..0f20a6a5 100644
@@ -3086,8 +3086,7 @@ class TestUpdatesService(BaseTestCase):
         publish.assert_called_with(topic='update.edit', msg=ANY)

-    @mock.patch('bodhi.server.notifications.publish')
-    def test_pending_update_on_stable_karma_reached_autopush_enabled(self, publish, *args):
+    def test_pending_update_on_stable_karma_reached_autopush_enabled(self, *args):
         """Ensure that a pending update stays in testing if it hits stable karma while pending."""
         nvr = u'bodhi-2.0.0-2.fc17'
         args = self.get_update(nvr)
```

bowlofeggs:
There is an integration test failure in CentOS CI, but I cannot reproduce when I run it locally. I'm going to try to bump your branch to get CI to run again and see if its repeatable.

bowlofeggs:
I have filed a bug about the integration test failure: https://github.com/fedora-infra/bodhi/issues/2835

bowlofeggs:
Ha, and now a conflict. I'll rebase again to resolve that. Who knows, maybe that'll get the tests to pass!

bowlofeggs:

bowlofeggs:

bowlofeggs:

PR #2763, opened 2018-11-20
Build the docs with Python 3 in CI.

bowlofeggs:


bowlofeggs:
The ```continuous-integration/jenkins/branch``` failure is actually left over from an earlier commit on this pull request and can be ignored.

puiterwijk (approved):

PR #2764, opened 2018-11-20
Use strs instead of bytes for smtplib.SMTP.sendmail().

### LINKED ISSUE #2756: Traceback from bodhi-dequeue-stable
@nirik sent me a traceback that seems to be coming from ```bodhi-dequeue-stable```:

```
Unable to send mail
  File "/usr/lib/python3.7/site-packages/bodhi/server/mail.py", line
391, in _send_mail
    smtp.sendmail(from_addr, [to_addr], body)
  File "/usr/lib64/python3.7/smtplib.py", line 861, in sendmail
    (code, resp) = self.mail(from_addr, esmtp_opts)
  File "/usr/lib64/python3.7/smtplib.py", line 534, in mail
    self.putcmd("mail", "FROM:%s%s" % (quoteaddr(sender), optionlist))
  File "/usr/lib64/python3.7/smtplib.py", line 150, in quoteaddr
    displayname, addr = email.utils.parseaddr(addrstring)
  File "/usr/lib64/python3.7/email/utils.py", line 212, in parseaddr
    addrs = _AddressList(addr).addresslist
  File "/usr/lib64/python3.7/email/_parseaddr.py", line 500, in __init__
    self.addresslist = self.getaddrlist()
  File "/usr/lib64/python3.7/email/_parseaddr.py", line 253, in getaddrlist
    ad = self.getaddress()
  File "/usr/lib64/python3.7/email/_parseaddr.py", line 263, in getaddress
    self.gotonext()
  File "/usr/lib64/python3.7/email/_parseaddr.py", line 236, in gotonext
    if self.field[self.pos] in self.LWS + '\n\r':
TypeError: 'in <string>' requires string as left operand, not int
```

bowlofeggs:
Under Python 3, a TypeError is raised if the from address or to
addresses are passed as bytes objects instead of str objects. We
had mistakenly been converting them as part of the Python 3
readiness effort. This commit fixes the code so we pass str
objects instead.

fixes #2756


bowlofeggs:
The ```rawhide-docs``` failure is fixed by https://github.com/fedora-infra/bodhi/pull/2763 and the ```rawhide-python3-diff-cover``` failure should be fixed by https://github.com/fedora-infra/bodhi/pull/2765 so I think we can ignore both of those here.

bowlofeggs:
I don't know why the ```rawhide-python3-diff-cover``` failed, but there must be coverage on that line because all the other Fedora's passed coverage and the Rawhide Python 2 one also passed. Something is weird with coverage on Python 3 on Rawhide, because that test also failed on https://github.com/fedora-infra/bodhi/pull/2766 which didn't even touch any code.

puiterwijk (approved):

PR #2765, opened 2018-11-20
Run diff-cover with just develop instead of origin/develop.

bowlofeggs:
It's better not to assume that all users of bodhi-ci call the
central remote origin.


bowlofeggs:
The Rawhide docs failure is fixed by https://github.com/fedora-infra/bodhi/pull/2763 and can be ignored here.

puiterwijk (approved):
I guess. +1

PR #2766, opened 2018-11-20
Prepare the 3.11.1 release.

bowlofeggs:


bowlofeggs:
The ```rawhide-docs``` failure is fixed by https://github.com/fedora-infra/bodhi/pull/2763 and can be ignored here.

bowlofeggs:
I think we can ignore the rawhide-diff-cover failure as described here: https://github.com/fedora-infra/bodhi/pull/2764#issuecomment-440767648

puiterwijk (approved):

Issue #2759, opened 2018-11-20
Drop Python 2 support for the server


bowlofeggs:
Dropping support for Python 2 will allow us to use some cool new Python 3 stuff. I'm marking it as backwards incompatible mostly because it will prevent deployments from simply doing a ```dnf upgrade```.

bowlofeggs:
This change is planned for inclusion in the upcoming 4.0.0 release: #3221

bowlofeggs:
Bodhi 4.0.0 beta is built and deployed to staging:

https://copr.fedorainfracloud.org/coprs/bowlofeggs/bodhi-pre-release
https://bodhi.stg.fedoraproject.org/

PR #2769, opened 2018-11-21
Handle IncompleteRead in the composer when on Python 3.

### LINKED ISSUE #2758: IncompleteRead while Composer polls for repomd files
It seems that Python's http library can raise a new Exception in Python 3 that we don't catch:

```
[2018-11-20 14:42:34][fedmsg.consumers    INFO]epel7-testing Polling http://download01.phx2.fedoraproject.org/pub/epel/testing/7/aarch64/repodata/repomd.xml
[2018-11-20 14:42:34][fedmsg.consumers   ERROR]epel7-testing Exception in ComposerThread(epel7-testing)
  File "/usr/lib/python3.7/site-packages/bodhi/server/consumers/masher.py", line 410, in work
    self._compose_updates()
  File "/usr/lib/python3.7/site-packages/bodhi/server/consumers/masher.py", line 959, in _compose_updates
    self._wait_for_sync()
  File "/usr/lib/python3.7/site-packages/bodhi/server/consumers/masher.py", line 1327, in _wait_for_sync
    newsum = hashlib.sha1(masterrepomd.read()).hexdigest()
  File "/usr/lib64/python3.7/http/client.py", line 460, in read
    s = self._safe_read(self.length)
  File "/usr/lib64/python3.7/http/client.py", line 612, in _safe_read
    raise IncompleteRead(b''.join(s), amt)
http.client.IncompleteRead: IncompleteRead(3295 bytes read, 1 more expected)
```

This one should be pretty easy to fix - we just need to catch this new Exception and retry, similar to what we do if a URLError is raised.

bowlofeggs:
fixes #2758

### DIFF
--- a/bodhi/tests/server/consumers/test_masher.py
+++ b/bodhi/tests/server/consumers/test_masher.py
@@ -25,6 +25,11 @@ import shutil
 import tempfile
 import time
 import unittest
+try:
+    from http.client import IncompleteRead
+except ImportError:
+    # Python 2 does not have this Exception.
+    IncompleteRead = None
 
 from click import testing
 import mock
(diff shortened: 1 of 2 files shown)


puiterwijk (approved):
LGTM

Zlopez (approved):
LGTM

bodhi/tests/server/consumers/test_masher.py
```diff
@@ -25,6 +25,11 @@
 import tempfile
 import time
 import unittest
+try:
+    from http.client import IncompleteRead
+except ImportError:
+    # Python 2 does not have this Exception.
+    IncompleteRead = None
```
pypingou, line 32: You could import this directly in the test to save you from catching the ImportError

pypingou:
Looks fine to me

bodhi/tests/server/consumers/test_masher.py
bowlofeggs: That would violate PEP-8, so I prefer this style. Bodhi will [drop Python 2](https://github.com/fedora-infra/bodhi/issues/2759) soon anyway, so this code won't be here for long.

PR #2777, opened 2018-11-25
Fix ordering of suggestions to have "unspecified" first

puiterwijk:
Previously there was no ordering, but since 4c7829ff5e741162ba16b5517cd2d9b3bc1f7d90,
that is no longer true.
Let's use the ordering as passed by that.


bowlofeggs:
This patch does not seem to solve the issue. Also, 4c7829f did not cause the issue as it is not deployed to the web servers yet (only to ```backend01```). The issue seems to happen when Bodhi is run under Python 3 and does not occur when run under Python 2.

I'm going to close this PR and open a new one when I figure out a fix.

bowlofeggs:
Ah, so this patch is actually useful because it fixed the ordering for the edit form, but not for the new form. I ended up using this in the patch I'm working on to fix #2778, so thanks!

PR #2781, opened 2018-11-25
Disallow creating BROs if the test gating status is failed

### LINKED ISSUE #2537: BuildrootOverrides should not be allowed if the test gating status is failed
We should add a check that disallows creating BROs if the test gating status is failed.

sebwoj:
fix #2537


bowlofeggs (approved):
Thanks!

bowlofeggs:

bowlofeggs:

PR #2794, opened 2018-11-30
Prevent obsoleting security update by new non-security one

### LINKED ISSUE #1798: If a new build obsoletes a security update it should also be a security update
A maintainer might submit a new build of a package that fixes a bug so will label the update a bugfix, but if it's obsoleteing a security update the new update should also be a security update.

eg: firefox from  firefox-55.0-6.fc26 ->  firefox-55.0.1-1.fc26

https://bodhi.fedoraproject.org/updates/?packages=firefox

sebwoj:
Fixes #1798

### DIFF
--- a/bodhi/server/models.py
+++ b/bodhi/server/models.py
@@ -2339,6 +2339,17 @@ class Update(Base):
                             )
                         })
 
+                # Warn about attempt to obsolete security update by update with
+                # other type and set type of new update to security.
+                if oldBuild.update.type == UpdateType.security and \
+                        self.type is not UpdateType.security:
+                    caveats.append({
+                        'name': 'update',
+                        'description': 'Adjusting type of this update to security,'
+                        'since it obsoletes another security update'
+                    })
+                    self.type = UpdateType.security
+
                 if obsoletable:
                     log.info('%s is obsoletable' % oldBuild.nvr)
 
--- a/bodhi/tests/server/services/test_updates.py
+++ b/bodhi/tests/server/services/test_updates.py
@@ -659,6 +659,120 @@ class TestNewUpdate(BaseTestCase):
 
     @mock.patch(**mock_valid_requirements)
     @mock.patch('bodhi.server.notifications.publish')
+    def test_create_new_nonsecurity_update_when_previous_security_one_exists(self, publish, *args):
+        """
+        Assert that when non-security update obsoletes previous security update, caveat is reported
+        and submitted update type is changed to security.
+        """
+        nvr = u'bodhi-2.0.0-2.fc17'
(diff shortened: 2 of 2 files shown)


bodhi/server/models.py
```diff
@@ -2333,13 +2333,29 @@ def obsolete_older_updates(self, db):
                             'name': 'update',
                             'description': 'Please be aware that there '
                             'is another update in flight owned by %s, '
-                            'containing %s.  Are you coordinating with '
+                            'containing %s. Are you coordinating with '
```
bowlofeggs: This change is OK, but it should be in its own atomic commit (a separate pull request would make sense too) instead of mixed in with the primary purpose of this commit.

bodhi/server/models.py
```diff
@@ -2333,13 +2333,29 @@ def obsolete_older_updates(self, db):
...
+                        'name': 'update',
+                        'description': 'Attempt to obsolete security update'
+                        ' by update with other type: Please be aware that there'
+                        ' is another security update in flight owned by %s,'
+                        ' containing %s. Are you coordinating with '
+                        'them?' % (
+                            oldBuild.update.user.name,
+                            oldBuild.nvr,
+                        )
+                    })
```
bowlofeggs, line 2350: The goal in #1798 is to coerce the new update type into also being a security update, not just to inform the packager that the two types mismatched. I do like the idea of using a caveat to tell them we changed the type, but we should also make sure we change the type.

Perhaps we could change the message on the caveat to "Adjusting type of this update to security, since it obsoletes another security update."

bodhi/tests/server/services/test_updates.py
```diff
@@ -657,6 +657,117 @@ def test_obsoletion(self, publish, *args):
...
+        publish.assert_called_with(
+            topic='update.request.testing', msg=mock.ANY)
+
+        up = self.db.query(Update).filter_by(title=nvr).one()
+        self.assertEqual(up.status, UpdateStatus.obsolete)
+        expected_comment = u'This update has been obsoleted by [bodhi-2.0.0-3.fc17]({}).'
+        expected_comment = expected_comment.format(
+            urlparse.urljoin(config['base_address'],
+                             '/updates/FEDORA-{}-53345602d5'.format(datetime.now().year)))
+        self.assertEqual(up.comments[-1].text, expected_comment)
```
bowlofeggs, line 718: Let's also assert that the type of the new update is security.

bowlofeggs (changes requested):

bowlofeggs (approved):
Thanks!

bowlofeggs:

bowlofeggs:

PR #2798, opened 2018-12-01
Place comment on the update about test_gating_status change

### LINKED ISSUE #2210: [M] Changes to test_gating_status should place comments on the update
We should leave comments on updates whenever their ```test_gating_status``` attribute is changed. This is important so that packagers get notifications when their tests fail.

sebwoj:
Fixes #2210

### DIFF
--- a/bodhi/server/models.py
+++ b/bodhi/server/models.py
@@ -3545,6 +3545,33 @@ class Update(Base):
 
         return result
 
+    @staticmethod
+    def comment_on_test_gating_status_change(target, value, old, initiator):
+        """
+        Place comment on the update when ``test_gating_status`` changes.
+
+        Args:
+            target (Update): The compose that has had a change to its state attribute.
+            value (EnumSymbol): The new value of the test_gating_status.
+            old (EnumSymbol): The old value of the test_gating_status
+            initiator (sqlalchemy.orm.attributes.Event): The event object that is initiating this
+                transition.
+        """
+        if value != old:
+            target.comment(
+                Session(),
+                f"This update test gating status has been changed to '{value}'.",
+                author=u"bodhi",
+            )
+
+
+event.listen(
+    Update.test_gating_status,
+    'set',
+    Update.comment_on_test_gating_status_change,
+    active_history=True,
+)
+
 
 class Compose(Base):
     """
--- a/bodhi/tests/server/__init__.py
+++ b/bodhi/tests/server/__init__.py
@@ -24,7 +24,7 @@ import sqlalchemy
 
 from bodhi.server.models import (
     Bug, BuildrootOverride, Comment, Group, RpmPackage, Release, ReleaseState, RpmBuild,
-    Update, TestGatingStatus, UpdateRequest, UpdateSeverity, UpdateType, User, TestCase)
+    Update, UpdateRequest, UpdateSeverity, UpdateType, User, TestCase)
 
 
 def create_update(session, build_nvrs, release_name=u'F17'):
@@ -75,11 +75,8 @@ def create_update(session, build_nvrs, release_name=u'F17'):
     update = Update(
         title=', '.join(build_nvrs), builds=builds, user=user, request=UpdateRequest.testing,
         notes=u'Useful details!', type=UpdateType.bugfix, date_submitted=datetime(1984, 11, 2),
-        requirements=u'rpmlint', stable_karma=3, unstable_karma=-3,
-        test_gating_status=TestGatingStatus.passed)
+        requirements=u'rpmlint', stable_karma=3, unstable_karma=-3, release=release)
     session.add(update)
-    update.release = release
-
     return update
 
 
--- a/bodhi/tests/server/consumers/test_composer.py
+++ b/bodhi/tests/server/consumers/test_composer.py
@@ -928,12 +928,19 @@ That was the actual one'''
                 builds=[build], user=user,
                 status=UpdateStatus.testing,
                 request=UpdateRequest.stable,
-                notes=u'Useful details!', release=release,
-                test_gating_status=TestGatingStatus.passed,
+                notes=u'Useful details!',
+                release=release,
                 stable_karma=3,
-                unstable_karma=-3)
-            update.type = UpdateType.security
+                unstable_karma=-3,
+                type=UpdateType.security)
+
+            update.test_gating_status = TestGatingStatus.passed
+
             db.add(update)
+
+            # Changing update's test_gating_status calls notifications.publish,
+            # so we have to reset mock calls
+            publish.reset_mock()
 
             # Wipe out the tag cache so it picks up our new release
             Release._tag_cache = None
@@ -1020,11 +1027,18 @@ That was the actual one'''
                 builds=[build], user=user,
                 status=UpdateStatus.testing,
                 request=UpdateRequest.stable,
-                notes=u'Useful details!', release=release,
-                test_gating_status=TestGatingStatus.passed,
                 stable_karma=3,
-                unstable_karma=-3)
-            update.type = UpdateType.enhancement
+                unstable_karma=-3,
+                notes=u'Useful details!',
+                release=release,
+                type=UpdateType.enhancement)
+
+            update.test_gating_status = TestGatingStatus.passed
+
+            # Changing update's test_gating_status calls notifications.publish,
+            # so we have to reset mock calls
+            publish.reset_mock()
+
             db.add(update)
 
             # Wipe out the tag cache so it picks up our new release
@@ -1104,11 +1118,14 @@ That was the actual one'''
                 builds=[build], user=user,
                 status=UpdateStatus.testing,
                 request=UpdateRequest.stable,
-                notes=u'Useful details!', release=release,
-                test_gating_status=TestGatingStatus.passed,
                 stable_karma=3,
-                unstable_karma=-3)
-            update.type = UpdateType.security
+                unstable_karma=-3,
+                notes=u'Useful details!',
+                release=release,
+                type=UpdateType.security)
+
+            update.test_gating_status = TestGatingStatus.passed
+
             db.add(update)
 
             # Wipe out the tag cache so it picks up our new release
@@ -1530,11 +1547,14 @@ That was the actual one'''
                 builds=[build2], user=user,
                 status=UpdateStatus.testing,
                 request=UpdateRequest.stable,
-                notes=u'Useful details!', release=release,
-                test_gating_status=TestGatingStatus.passed,
                 stable_karma=3,
-                unstable_karma=-3)
-            update.type = UpdateType.security
+                unstable_karma=-3,
+                notes=u'Useful details!',
+                release=release,
+                type=UpdateType.security)
+
+            update.test_gating_status = TestGatingStatus.passed
+
             db.add(update)
 
             # Wipe out the tag cache so it picks up our new release
@@ -2327,11 +2347,14 @@ class TestContainerComposerThread__compose_updates(ComposerThreadBaseTestCase):
             builds=[build1, build2], user=user,
             status=UpdateStatus.pending,
             request=UpdateRequest.testing,
-            notes=u'Neat I can compose containers now', release=release,
-            test_gating_status=TestGatingStatus.passed,
             stable_karma=3,
-            unstable_karma=-3)
-        update.type = UpdateType.bugfix
+            unstable_karma=-3,
+            notes=u'Neat I can compose containers now',
+            release=release,
+            type=UpdateType.bugfix)
+
+        update.test_gating_status = TestGatingStatus.passed
+
         self.db.add(update)
         # Wipe out the tag cache so it picks up our new release
         Release._tag_cache = None
@@ -2495,11 +2518,14 @@ class TestFlatpakComposerThread__compose_updates(ComposerThreadBaseTestCase):
             builds=[build1, build2], user=user,
             status=UpdateStatus.pending,
             request=UpdateRequest.testing,
-            notes=u'Neat I can compose flatpaks now', release=release,
-            test_gating_status=TestGatingStatus.passed,
             stable_karma=3,
-            unstable_karma=-3)
-        update.type = UpdateType.bugfix
+            unstable_karma=-3,
+            notes=u'Neat I can compose flatpaks now',
+            release=release,
+            type=UpdateType.bugfix)
+
+        update.test_gating_status = TestGatingStatus.passed
+
         self.db.add(update)
         # Wipe out the tag cache so it picks up our new release
         Release._tag_cache = None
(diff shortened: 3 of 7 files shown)


bodhi/server/models.py
```diff
@@ -2029,8 +2029,9 @@ def get_test_gating_info(self):
 
         return bodhi.server.util.greenwave_api_post(api_url, data)
 
-    def update_test_gating_status(self):
+    def update_test_gating_status(self, db):
```
bowlofeggs: Let's not pass the ```Session``` around, it can be accessed via ```Session()``` which is in scope in this module.

In general, I've been meaning to go through the code and find and remove the places that are doing this.

bodhi/server/models.py
```diff
@@ -2046,6 +2047,12 @@ def update_test_gating_status(self):
                 decision.get('unsatisfied_requirements', []))
         self.greenwave_summary_string = decision['summary']
 
+        # Comment test_gating_status change on update
+        if old_test_gating_status != self.test_gating_status:
+            self.comment(db, u"This update test_gating_status has been changed to '%s'." % (
```
bowlofeggs: Let's make it a bit more human friendly, and here's an example of using the ```Session```:

```suggestion
            self.comment(Session(), u"This update test gating status has been changed to '%s'." % (
```

bodhi/server/models.py
```diff
@@ -2681,7 +2688,7 @@ def set_request(self, db, action, username):
         topic = u'update.request.%s' % action
         notifications.publish(topic=topic, msg=dict(update=self, agent=username))
 
-    def waive_test_results(self, username, comment=None, tests=None):
+    def waive_test_results(self, db, username, comment=None, tests=None):
```
bowlofeggs: Similar here.

bodhi/server/models.py
```diff
@@ -2729,8 +2736,15 @@ def waive_test_results(self, username, comment=None, tests=None):
...
                 '{}/waivers/'.format(config.get('waiverdb_api_url')), data)
 
+        old_test_gating_status = self.test_gating_status
         self.test_gating_status = TestGatingStatus.waiting
 
+        # Comment test_gating_status change on update
+        if old_test_gating_status != self.test_gating_status:
+            self.comment(db, u"This update test_gating_status has been changed to '%s'." % (
+                self.test_gating_status
+            ), author=username)
```
bowlofeggs: Since there are two places we are doing this, we could instead take advantage of [```sqlalchemy```'s event listeners](https://docs.sqlalchemy.org/en/latest/core/event.html) to subscribe to changes to the Update's ```test_gating_status``` and then we don't need to remember to do this every place that changes it.

For an example of how to do this, have a look [here](https://github.com/fedora-infra/bodhi/blob/3.11.2/bodhi/server/models.py#L3917).

bodhi/server/scripts/check_policies.py
```diff
@@ -57,7 +57,7 @@ def check():
 
     for update in updates:
         try:
-            update.update_test_gating_status()
+            update.update_test_gating_status(session)
```
bowlofeggs: Similar here.

bodhi/server/services/updates.py
```diff
@@ -587,7 +587,7 @@ def waive_test_results(request):
     tests = request.validated.pop('tests', None)
 
     try:
-        update.waive_test_results(request.user.name, comment, tests)
+        update.waive_test_results(request.db, request.user.name, comment, tests)
```
bowlofeggs: And here.

bowlofeggs (changes requested):

bodhi/server/models.py
sebwoj: After applying following code:

```
    @staticmethod
    def comment_on_test_gating_status_change(target, value, old, initiator):
        """
        Place comment on the update when ``test_gating_status`` changes.
        Args:
            target (Update): The compose that has had a change to its state attribute.
            value (EnumSymbol): The new value of the test_gating_status.
            old (EnumSymbol): The old value of the test_gating_status
            initiator (sqlalchemy.orm.attributes.Event): The event object that is initiating this
                transition.
        """
        if value != old:
            target.comment(
                Session(),
                u"This update test_gating_status has been changed to '%s'." % (value),
                author=u"a",
            )


event.listen(
    Update.test_gating_status,
    'set',
    Update.comment_on_test_gating_status_change,
    active_history=True,
)
```

I run the following command:
`python3 -m unittest test_check_policies.TestCheckPolicies.test_policies_unsatisfied`

But received output was:
```======================================================================
ERROR: test_policies_unsatisfied (test_check_policies.TestCheckPolicies)
Assert correct behavior when the policies enforced by Greenwave are unsatisfied
  File "/usr/lib64/python3.7/site-packages/sqlalchemy/engine/base.py", line 1193, in _execute_context
    context)
  File "/usr/lib64/python3.7/site-packages/sqlalchemy/engine/default.py", line 509, in do_execute
    cursor.execute(statement, parameters)
sqlite3.IntegrityError: NOT NULL constraint failed: updates.release_id

The above exception was the direct cause of the following exception:

  File "/home/vagrant/bodhi/bodhi/tests/server/base.py", line 165, in setUp
    populate(self.db)
  File "/home/vagrant/bodhi/bodhi/tests/server/__init__.py", line 116, in populate
    update = create_update(db, [u'bodhi-2.0-1.fc17'])
  File "/home/vagrant/bodhi/bodhi/tests/server/__init__.py", line 80, in create_update
    test_gating_status=TestGatingStatus.passed)
  File "<string>", line 4, in __init__
  File "/usr/lib64/python3.7/site-packages/sqlalchemy/orm/state.py", line 424, in _initialize_instance
    manager.dispatch.init_failure(self, args, kwargs)
  File "/usr/lib64/python3.7/site-packages/sqlalchemy/util/langhelpers.py", line 66, in __exit__
    compat.reraise(exc_type, exc_value, exc_tb)
  File "/usr/lib64/python3.7/site-packages/sqlalchemy/util/compat.py", line 249, in reraise
    raise value
  File "/usr/lib64/python3.7/site-packages/sqlalchemy/orm/state.py", line 421, in _initialize_instance
    return manager.original_init(*mixed[1:], **kwargs)
  File "/usr/lib64/python3.7/site-packages/sqlalchemy/ext/declarative/base.py", line 747, in _declarative_constructor
    setattr(self, k, kwargs[k])
  File "/usr/lib64/python3.7/site-packages/sqlalchemy/orm/attributes.py", line 229, in __set__
    instance_dict(instance), value, None)
  File "/usr/lib64/python3.7/site-packages/sqlalchemy/orm/attributes.py", line 710, in set
    value, old, initiator)
  File "/usr/lib64/python3.7/site-packages/sqlalchemy/orm/attributes.py", line 717, in fire_replace_event
    state, value, previous, initiator or self._replace_token)
  File "/usr/lib64/python3.7/site-packages/sqlalchemy/orm/events.py", line 1893, in wrap
    fn(target, *arg)
  File "/home/vagrant/bodhi/bodhi/server/models.py", line 3709, in comment_on_test_gating_status_change
    author=u"a",
  File "/home/vagrant/bodhi/bodhi/server/models.py", line 3062, in comment
    session.flush()
  File "/usr/lib64/python3.7/site-packages/sqlalchemy/orm/session.py", line 2254, in flush
    self._flush(objects)
  File "/usr/lib64/python3.7/site-packages/sqlalchemy/orm/session.py", line 2380, in _flush
    transaction.rollback(_capture_exception=True)
  File "/usr/lib64/python3.7/site-packages/sqlalchemy/util/langhelpers.py", line 66, in __exit__
    compat.reraise(exc_type, exc_value, exc_tb)
  File "/usr/lib64/python3.7/site-packages/sqlalchemy/util/compat.py", line 249, in reraise
    raise value
  File "/usr/lib64/python3.7/site-packages/sqlalchemy/orm/session.py", line 2344, in _flush
    flush_context.execute()
  File "/usr/lib64/python3.7/site-packages/sqlalchemy/orm/unitofwork.py", line 391, in execute
    rec.execute(self)
  File "/usr/lib64/python3.7/site-packages/sqlalchemy/orm/unitofwork.py", line 556, in execute
    uow
  File "/usr/lib64/python3.7/site-packages/sqlalchemy/orm/persistence.py", line 181, in save_obj
    mapper, table, insert)
  File "/usr/lib64/python3.7/site-packages/sqlalchemy/orm/persistence.py", line 866, in _emit_insert_statements
    execute(statement, params)
  File "/usr/lib64/python3.7/site-packages/sqlalchemy/engine/base.py", line 948, in execute
    return meth(self, multiparams, params)
  File "/usr/lib64/python3.7/site-packages/sqlalchemy/sql/elements.py", line 269, in _execute_on_connection
    return connection._execute_clauseelement(self, multiparams, params)
  File "/usr/lib64/python3.7/site-packages/sqlalchemy/engine/base.py", line 1060, in _execute_clauseelement
    compiled_sql, distilled_params
  File "/usr/lib64/python3.7/site-packages/sqlalchemy/engine/base.py", line 1200, in _execute_context
    context)
  File "/usr/lib64/python3.7/site-packages/sqlalchemy/engine/base.py", line 1413, in _handle_dbapi_exception
    exc_info
  File "/usr/lib64/python3.7/site-packages/sqlalchemy/util/compat.py", line 265, in raise_from_cause
    reraise(type(exception), exception, tb=exc_tb, cause=cause)
  File "/usr/lib64/python3.7/site-packages/sqlalchemy/util/compat.py", line 248, in reraise
    raise value.with_traceback(tb)
  File "/usr/lib64/python3.7/site-packages/sqlalchemy/engine/base.py", line 1193, in _execute_context
    context)
  File "/usr/lib64/python3.7/site-packages/sqlalchemy/engine/default.py", line 509, in do_execute
    cursor.execute(statement, parameters)
sqlalchemy.exc.IntegrityError: (sqlite3.IntegrityError) NOT NULL constraint failed: updates.release_id [SQL: 'INSERT INTO updates (title, autokarma, stable_karma, unstable_karma, requirements, require_bugs, require_testcases, display_name, notes, type, status, request, severity, suggest, locked, pushed, critpath, close_bugs, date_submitted, date_modified, date_approved, date_pushed, date_testing, date_stable, alias, old_updateid, release_id, user_id, test_gating_status, greenwave_summary_string, greenwave_unsatisfied_requirements) VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)'] [parameters: ('bodhi-2.0-1.fc17', 1, 3, -3, 'rpmlint', 0, 0, '', 'Useful details!', 'bugfix', 'pending', 'testing', 'unspecified', 'unspecified', 0, 0, 0, 1, '1984-11-02 00:00:00.000000', None, None, None, None, None, None, None, None, 1, None, None, None)] (Background on this error at: http://sqlalche.me/e/gkpj)

Ran 1 test in 0.720s

FAILED (errors=1)
```

It looks like session.flush() ans session.commit() in Update.comment is the problem here.

I found something llittle helpful:
https://stackoverflow.com/questions/36731020/sqlalchemy-orm-event-hook-for-attribute-persisted

But attempt to apply self.comment in before_commit, before_flush and etc didn't help.

Is there anything I can do to use Update.comment in events?
Workarounds:
1. Create function "set_test_gating_status_and_comment_change" to Update
2. Use deffered work using fedmsg/fm/celery
BTW:
@bowlofeggs 
You mentioned that fedmsg is loosing meessages. Do you thing that fedora-messaging has the same issue?

bowlofeggs: Hey @sebwoj!

I wonder if ```comment()``` should be calling a ```commit()``` - that seems possibly wrong. Anyways, a possible fix might be to see if you can set the release on the update in the tests before the ```test_gating_status``` gets set.

Another option might be to use a different event to check the change, like the "before commit" hook. This way you know the release is set.

Deferred work is something I do want to do with Bodhi. We currently do a little bit of it with fedmsg, but I don't love it, and yes fedmsg does lose messages. Since we just added support for fedora-messaging we will have access to a RabbitMQ server that we could use for Celery too, so I may be interested in going that direction.

fedora-messaging uses RabbitMQ, which should not silently drop messages (meaning, you will know if the broker received your message or not, unlike fedmsg).

sebwoj: 1.
> Another option might be to use a different event to check the change, like the "before commit" hook. 

> But attempt to apply self.comment in before_commit, before_flush and etc didn't help.
2. 

> Anyways, a possible fix might be to see if you can set the release on the update in the tests before the test_gating_status gets set.

Could you point me where exactly release should be set to update?
I found that line:
https://github.com/fedora-infra/bodhi/blob/9b02efa8b4efb6a5eff95a60be823037669bdad1/bodhi/server/models.py#L2100

Coulpe lines below we set `test_gating_status`:
https://github.com/fedora-infra/bodhi/blob/9b02efa8b4efb6a5eff95a60be823037669bdad1/bodhi/server/models.py#L2119

bowlofeggs:
@sebwoj do you still want to work on this?

sebwoj:
Yes. And it seems that we are in the middle of a conversation(https://github.com/fedora-infra/bodhi/pull/2798#discussion_r240817358) :)

bodhi/server/models.py
bowlofeggs: Hey @sebwoj!

If you look at the traceback, the problem happens here:

https://github.com/fedora-infra/bodhi/blob/3.12.0/bodhi/tests/server/__init__.py#L76-L82

Note that the release gets set two lines after the test_gating_status gets set. So we need to make sure the release is set first, and I think it will work.

sebwoj: It looks like I focused on the wrong lines.
Now it is working.
Thanks @bowlofeggs!

bodhi/server/models.py
```diff
@@ -3541,6 +3541,33 @@ def __json__(self, request=None, anonymize=False):
...
+            target (Update): The compose that has had a change to its state attribute.
+            value (EnumSymbol): The new value of the test_gating_status.
+            old (EnumSymbol): The old value of the test_gating_status
+            initiator (sqlalchemy.orm.attributes.Event): The event object that is initiating this
+                transition.
+        """
+        if value != old:
+            target.comment(
+                Session(),
+                u"This update test_gating_status has been changed to '%s'." % (value),
```
bowlofeggs: Just a suggestion, but we can use the cool new Python 3 f-strings feature now!

```suggestion
                f"This update test_gating_status has been changed to '{value}'.",
```

Also, since this is meant to be human readable, maybe we want to say "test gating status" or something similar (i.e., no _'s).

bowlofeggs (changes requested):
Hey @sebwoj!

Thanks for working on this, it seems to work great. I used ```bodhi-shell``` to set a new status on an update and the comment made it in.

Let's add one more unit test that sets the same status twice in a row and asserts that only one comment is added (i.e., let's test the if statement ```if value != old```).

My other comment is just a suggestion so we can keep it the way it is if you prefer.

bodhi/server/models.py
sebwoj: Good hint! I forgot that Bodhi is now py3 only.

bowlofeggs (approved):
Thanks!

bowlofeggs:

bowlofeggs:

Issue #1737, opened 2017-08-09
HTTP 400 instead of 403 when users don't have ACLs to edit a package


bowlofeggs:
I noticed today that I get an HTTP 400 when I try to visit https://bodhi.stg.fedoraproject.org/updates/FEDORA-2017-5ce9d89b82/edit when I am logged in to staging Bodhi, but I get a 403 when I visit it and am not logged in. The 400 should be a 403.

amitsin6h:
let me take this as my first bug fix.

bowlofeggs:
@amitsin6h Go 4 it!

linux-modder:
@amitsin6h if you need any help I'm also willing to help you.

egegunes:
After a quick look, it seems fixed? I visited https://bodhi.stg.fedoraproject.org/updates/FEDORA-2017-5ce9d89b82/edit when I'm logged in and not logged in and got 403 both times.

bowlofeggs:
@egegunes hmm, I just tried again and I still get a 400 when I am logged in and visit that page.

steaksauce-:
@bowlofeggs I get a 403 in both scenarios as well.

Might be a bit of a hassle, but what happens if you create a new account and reproduce the steps? Maybe it has something to do with something with your account.

bowlofeggs:
This fix is planned for inclusion in Bodhi 3.13: #2917.

bowlofeggs:
This fix is now deployed to staging: https://bodhi.stg.fedoraproject.org/docs/user/release_notes.html

It is also available via Copr: https://copr.fedorainfracloud.org/coprs/bowlofeggs/bodhi-pre-release/

bowlofeggs:
Bodhi 3.13.0 has been released:

https://github.com/fedora-infra/bodhi/releases/tag/3.13.0

Fedora Infrastructure plans to deploy it to production on Monday.

PR #2803, opened 2018-12-04
Make the update edit page 403 if no acls

ryanlerch:
Previously, the edit update page returned a 400 error
if a user tried to access the edit update page of a
package they don't have the ACLs to edit. This commit
updates Bodhi so it returns a 403 error

Fixes: #1737


bowlofeggs (approved):
Thanks!

bowlofeggs:
This fix is planned for inclusion in Bodhi 3.13: #2917.

bowlofeggs:

PR #2808, opened 2018-12-04
Do not use an array for Mergify since that doesn't seem to work.

bowlofeggs:


puiterwijk (commented):
ack

PR #2814, opened 2018-12-05
Allow the Bodhi team to ack for Mergify for my PRs.

bowlofeggs:

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -38,7 +38,7 @@ pull_request_rules:
   conditions:
   - label!=no-mergify
   - author=bowlofeggs
-  - approved-reviews-by=@fedora-infra
+  - "#approved-reviews-by>=1"
   - status-success=DCO
   - status-success=f28-docs
   - status-success=f28-flake8


bowlofeggs:
I've reported the Mergify syntax error here: https://github.com/Mergifyio/mergify-engine/issues/310

.mergify.yml
```diff
@@ -38,7 +38,7 @@ pull_request_rules:
   conditions:
   - label!=no-mergify
   - author=bowlofeggs
-  - approved-reviews-by=@fedora-infra
+  - "#approved-reviews-by>=1"
```
Zlopez, line 41: Why did you commented this and added >=1 at the same time?

bowlofeggs:
On Thu, 2018-12-06 at 01:09 -0800, Zlopez wrote:
> Why did you commented this and added >=1 at the same time?

It's not a comment, it's the syntax for indicating a number:

https://doc.mergify.io/conditions.html#matching-pull-requests-passing-travis-and-being-approved

There is a bug in mergify that prevents it from matching the Bodhi
team, which is why I am making this change.

Zlopez (approved):

Zlopez:
I looked to the mergify documentation, but probably on the wrong place. In this case LGTM.

PR #2818, opened 2018-12-05
Compare Update.request to UpdateRequest.testing in obsolete_older_upd…

sebwoj:
…ates query

### DIFF
--- a/bodhi/server/models.py
+++ b/bodhi/server/models.py
@@ -2305,7 +2305,7 @@ class Update(Base):
                      Build.package == build.package,
                      Update.locked == False,
                      Update.release == self.release,
-                     or_(Update.request == UpdateStatus.testing,
+                     or_(Update.request == UpdateRequest.testing,
                          Update.request == None),
                      or_(Update.status == UpdateStatus.testing,
                          Update.status == UpdateStatus.pending))


Zlopez (approved):

bodhi/server/models.py
```diff
@@ -2305,7 +2305,7 @@ def obsolete_older_updates(self, db):
                      Build.package == build.package,
                      Update.locked == False,
                      Update.release == self.release,
-                     or_(Update.request == UpdateStatus.testing,
+                     or_(Update.request == UpdateRequest.testing,
                          Update.request == None),
```
bowlofeggs, line 2309: To date, this query has effectively been ```Update.request == None``` since the ```request``` can't be equal to an ```UpdateStatus``` enum. I'm not sure what the goal of comparing the request with testing is, since I think we'd want to obsolete updates that do have a request for testing actually.

My recommendation is to just require the request to be ```None``` and forget the request testing code. I don't see why we'd want to skip those, and no user has complained about this to my knowledge (and this is effectively what the code does today).

bowlofeggs (changes requested):

bodhi/server/models.py
sebwoj: After line "or_(Update.request == UpdateStatus.testing," removal, several(10) unittest failed in services/test_updates.py(related to obsoleting). Are we sure about this line? I can adjust failed UT?

bowlofeggs: Now that I think about it some more, in the SQL that this query becomes, this does effectively ask if the request is "testing" with either form of the query, so what I wrote above it incorrect.

bowlofeggs (approved):

PR #2819, opened 2018-12-06
Do run the unittests on rawhide for python 3

abompard:
Changeset 64ff87102 disabled the unittests on Rawhide for any version of Python, but it's only unsupported on Python 2. As a results, the `diff_cover` check fails because no tests were run on Rawhide & Python 3.


bowlofeggs (approved):
Oh man, that's a terrible mistake on my part. Thanks for catching that!

PR #2820, opened 2018-12-06
Allow CLI users to override the OpenID API.

bowlofeggs:
It has long bothered me that Bodhi development uses Fedora's
production authentication server instead of some other
authentication, because this adds a risk that actions in
development might alter production Bodhi (particular with the CLI).

This commit makes it possible for CLI users to override the OpenID
API used by the CLI, and configures the defaults in Vagrant to use
Fedora's staging identity server instead of production.

I did consider giving the server an API to provide settings like
the OpenID provider, but that would cause another request on every
command and so I opted for allowing the user to specify an OpenID
API.

### DIFF
--- a/bodhi/client/__init__.py
+++ b/bodhi/client/__init__.py
@@ -32,7 +32,7 @@ import six
 import munch
 
 from bodhi.client import bindings
-from fedora.client import AuthError
+from fedora.client import AuthError, openidproxyclient
 
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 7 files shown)


bowlofeggs:
I'm still "near the fence" about whether the client should have the OpenID API setting vs. having a server API that provides this setting, so feel free to argue for a different design if you think there's a better way.

puiterwijk:
Given that for OIDC, the client will need to know where to get the token anyway, let's go with client-side for now.

puiterwijk (approved):
I really don't like this OpenID API anymore, but the patch looks okay..

bowlofeggs:
The test failures will be fixed by #2891.

bowlofeggs:
Looks like I need to rebase this again to resolve another conflict.

bowlofeggs:

bowlofeggs:

bowlofeggs:

bowlofeggs:

PR #2821, opened 2018-12-07
Show more bugs information in the update page

mattiaverga:
This change will change how bugs are added and displayed on the new update page.
Each bug will be displayed on two rows along with a badge with the release-product information. This should reduce the risk to reference by mistake an EPEL bug to a Fedora update or vice versa.

Moreover, information about manually entered bugs will be retrieved from RH Bugzilla and a series of checks will be performed:

- if the bug is private, user will not be able to add it to the update
- if the bug is not related to Fedora, Fedora EPEL or Fedora Modules, user will be prompted to confirm they're sure to reference the bug to the update
- if the bug is already in CLOSED state, again user is prompted to confirm their action

The final result looks like this:
screenshot_20181206_195138

### DIFF
--- a/bodhi/server/static/js/update_form.js
+++ b/bodhi/server/static/js/update_form.js
@@ -304,11 +304,58 @@ $(document).ready(function() {
             $.each(intermediary.trim().split(" "), function(j, item) {
                 item = item.trim()
                 if (item[0] == '#') { item = item.substring(1); }
-                add_bug_checkbox(item, '', true, true);
+                $.ajax({
+                    url: "https://bugzilla.redhat.com/jsonrpc.cgi?method=Bug.get&params=%5B%7B%22ids%22:" + item + "%7D%5D",
+                    timeout: 10000,
+                    dataType: 'jsonp',
+                    success: function(data) {
+                        // Received error response e.g. bug doesn't exist
+                        if (data.error) {
+                            messenger.post({
+                                message: data.error.message,
+                                type: 'error',
+                            });
+                            return;
+                        }
+                        // There should be only one
+                        if (data.result.bugs.length != 1) {
+                            messenger.post({
+                                message: 'Received bad response for bug #' + item,
+                                type: 'error',
+                            });
+                            return;
+                        }
+                        // Check Bug product
+                        if (data.result.bugs[0].product !== "Fedora" && data.result.bugs[0].product !== "Fedora EPEL") {
+                            var r = confirm('Bug #' + item + ' doesn\'t seem to refer to Fedora or Fedora EPEL.\nAre you sure you want to reference it in this update?');
+                            if (r === false) { return; }
+                        }
+                        // Alert user if bug is already closed
+                        if (data.result.bugs[0].status == "CLOSED") {
+                            var r = confirm('Bug #' + item + ' is already in CLOSED state.\nAre you sure you want to reference it in this update?');
+                            if (r === false) { return; }
+                        }
+                        add_bug_checkbox(item, data.result.bugs[0].summary, true, true);
+                    },
+                    error: function(jqXHR, textStatus) {
+                        if (textStatus == 'timeout') {
+                            messenger.post({
+                                message: 'Reached timeout while retrieving details for bug #' + item,
+                                type: 'error',
+                            });
+                        }
+                        else {
+                            messenger.post({
+                                message: 'Unable to retrieve details for bug #' + item,
+                                type: 'error',
+                            });
+                        }
+                    },
+                });
             });
         });
         $("#bugs-adder input").val('');  // Clear the field
-        return false;
+        return;
     }
     var add_builds = function() {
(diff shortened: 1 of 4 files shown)


bodhi/server/static/js/update_form.js
```diff
@@ -298,16 +303,93 @@ $(document).ready(function() {
...
+                        if (data.result.bugs.length != 1) {
+                            messenger.post({
+                                message: 'Received bad response for bug #' + item,
+                                type: 'error',
+                            });
+                            return;
+                        }
+                        // Check Bug product
+                        var product = data.result.bugs[0].product;
+                        if (product !== "Fedora" && product !== "Fedora EPEL" && product !== "Fedora Modules") {
```
bowlofeggs: There are other Bodhi deployments in the wild that do not have products named Fedora or EPEL. We should not hard code product names here.

bowlofeggs (changes requested):
Just one change. Looks reasonable to me, though I find it difficult to test against RHBZ due to the CORS header. How were you able to test this change?

bodhi/server/static/js/update_form.js
mattiaverga: Well, the ajax in the script already fetch the bugs list of the selected package from `apps.fedoraproject.org/packages/fcomm_connector/bugzilla/`, which returns only Fedora, EPEL and (recently applied patch) Fedora Modules. So, in some way, we are already hardcoding allowed products.

Is setting a list of allowed products at the top of the file acceptable to you, or should I remove (or simply comment out) the product check?

mattiaverga:
> Looks reasonable to me, though I find it difficult to test against RHBZ due to the CORS header. How were you able to test this change?

RH Bugzilla guys helped me and made me discover that by using JSONP in place of normal JSON-RPC does not require setting any CORS header.
https://github.com/fedora-infra/bodhi/blob/df40231c1c90d31537502187d78c14eec4d9ab5a/bodhi/server/static/js/update_form.js#L324

So it should now work (it's running fine in my local environment). Do you still have problems running it?

bodhi/server/static/js/update_form.js
bowlofeggs: Yeah Bodhi does have some hardcoded behavior - I've been slowly trying to remove it as I go over time. It's one of the things that makes it hard for the other deployments, or for new deployments (I've talked to a few people who tried to use it and hit these hardcoded things.)

One way to solve this problem is to make the list of product names to check (and the URL to bugs, for that matter) a setting in ```bodhi.server.config```, so that other deployments can configure it as they see fit.

bowlofeggs:
One other thing to consider here is that the JS can't be the authority on data coming in to Bodhi, because data can also come in via the REST API (for example, the CLI). So it's OK to have the JS filter for convenience sake, but if the goal is to stop bugs with the wrong product from getting in we will need to do that server side.

bowlofeggs:
> So it should now work (it's running fine in my local environment). Do you still have problems running it?

It seems to work great now. This is a nice change!

Let me know what you think about turning the product list into a setting.

mattiaverga:
> So it's OK to have the JS filter for convenience sake, but if the goal is to stop bugs with the wrong product from getting in we will need to do that server side.

The goal of that check is only to warn the user that he is adding something "strange" to the update, but this will not prevent the user to add the bug. When adding a bug ID manually and the user maybe types the wrong number, this will make them aware of the error.

About turning the product list into a setting, I have no idea how to make a JS read a setting in `bodhi.server.config`... I agree it would be really nice to have it, but I'm doubtful it's really doable. I will check.

bowlofeggs:
On Tue, 2018-12-11 at 17:14 +0000, Mattia Verga wrote:
> The goal of that check is only to warn the user that he is adding
> something "strange" to the update, but this will not prevent the user
> to add the bug. When adding a bug ID manually and the user maybe
> types the wrong number, this will make them aware of the error.

+1

> About turning the product list into a setting, I have no idea how to
> make a JS read a setting in bodhi.server.config... I agree it would
> be really nice to have it, but I'm doubtful it's really doable. I
> will check.

There are two ways to do it that I can think of. One is to have the
Python template turn the setting into a JS list. The other is to
provide the setting as a JSONP object, which I've been tinkering with
in a (very stale) pull request:

https://github.com/fedora-infra/bodhi/pull/2533

mattiaverga:
So, I think this will have to wait until #2533 is ready.
I'll take a look to help you develop it further.

bodhi/server/static/js/update_form.js
mattiaverga: I've changed the product check to use config['bz_products'] list already defined in settings. However, I added the defaults into the server/config.py file rather than rely on production.ini, mainly because this way it will also work in local development environment without adding the same setting to development.ini.

mattiaverga:
closing, superseded by #3453

PR #2841, opened 2018-12-12
Document how to configure libvirt not to ask for password.

bowlofeggs:
Vagrant is much more pleasant to use if users are not asked for a
password when stopping or starting a guest. This commit documents
how to achieve this for Vagrant users.


puiterwijk (changes requested):
I honestly feel quite uncomfortable suggesting people to do this, given the fact that libvirt can be used for privilege escalation, and this would open that up for anyone getting to run code as the unprivileged user.

bowlofeggs:
OK. I don't feel strongly about it, so I'll just close this. Thanks for looking!

Issue #2401, opened 2018-05-31
Error in setup.py in bodhi-client-3.3.0


tkdchen:
Got this error by pip install.

```
pip install "bodhi-client>=2.0,<4"
Collecting bodhi-client<4,>=2.0
  Downloading https://files.pythonhosted.org/packages/d3/8e/ce082026aca1e337a317c44c44c4834c7d6f7de05948c18f427ea1411026/bodhi-client-3.3.0.tar.gz
    Complete output from command python setup.py egg_info:
    running egg_info
    creating pip-egg-info/bodhi.egg-info
    writing pip-egg-info/bodhi.egg-info/PKG-INFO
    writing top-level names to pip-egg-info/bodhi.egg-info/top_level.txt
    writing dependency_links to pip-egg-info/bodhi.egg-info/dependency_links.txt
    writing manifest file 'pip-egg-info/bodhi.egg-info/SOURCES.txt'
    package init file 'bodhi/__init__.py' not found (or not a regular file)
    reading manifest file 'pip-egg-info/bodhi.egg-info/SOURCES.txt'
    writing manifest file 'pip-egg-info/bodhi.egg-info/SOURCES.txt'
    running egg_info
    creating pip-egg-info/bodhi_client.egg-info
    writing requirements to pip-egg-info/bodhi_client.egg-info/requires.txt
    writing pip-egg-info/bodhi_client.egg-info/PKG-INFO
    writing top-level names to pip-egg-info/bodhi_client.egg-info/top_level.txt
    writing dependency_links to pip-egg-info/bodhi_client.egg-info/dependency_links.txt
    writing entry points to pip-egg-info/bodhi_client.egg-info/entry_points.txt
    writing manifest file 'pip-egg-info/bodhi_client.egg-info/SOURCES.txt'
    reading manifest file 'pip-egg-info/bodhi_client.egg-info/SOURCES.txt'
    reading manifest template 'CLIENT_MANIFEST.in'
    writing manifest file 'pip-egg-info/bodhi_client.egg-info/SOURCES.txt'
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/tmp/pip-build-quDnNB/bodhi-client/setup.py", line 127, in <module>
        server_packages.remove('bodhi')
    ValueError: list.remove(x): x not in list
```

tkdchen:
This also happens in `bodhi-client-2.9.0`

tkdchen:
I tried to add

```python
[...]

Issue #2241, opened 2018-03-23
Drop Bodhi's "stacks" feature


bowlofeggs:
Bodhi has code to deal with "stacks", but it hasn't even been used. We should either figure out what would make it useful and finish it, or delete it.

If we delete it, we should also delete [```bodhi.server.models.BodhiBase.update_relationship()```](https://github.com/fedora-infra/bodhi/blob/3.5.1/bodhi/server/models.py#L470-L504) as it is only used by stacks.

bowlofeggs:
This change is planned for inclusion in the upcoming 4.0.0 release: #3221

bowlofeggs:
Bodhi 4.0.0 beta is built and deployed to staging:

https://copr.fedorainfracloud.org/coprs/bowlofeggs/bodhi-pre-release
https://bodhi.stg.fedoraproject.org/

PR #2842, opened 2018-12-12
Add the bodhi_messages Python package

jeremycline:
fedora-messages uses Python classes to allow authors to define JSON
schema and a Python API to work with their messages. This adds the
basics of a Python package to distribute these schema.

This is similar to fedora_infrastructure_fedmsg_meta, except each
application manages its own schema definitions and how the message is
represented to the user.

Signed-off-by: Jeremy Cline <jcline@redhat.com>

--- 

I had this sitting in a branch and meant to write schema for Bodhi, but I've not found time yet. I thought it best to push what I have (which is just a skeleton package) and have us all chip away and defining schema and changing the code that publishes messages to use them.

I pulled the list of messages from https://fedora-fedmsg.readthedocs.io/en/latest/, but some may not still be sent. Also that list might not actually be complete. I expect as we audit the code we'll alter the list of classes here.

### DIFF
--- a/bodhi/__init__.py
+++ b/bodhi/__init__.py
@@ -17,3 +17,6 @@
 # along with this program; if not, write to the Free Software
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
 """Fedora's update manager."""
+
+# This is a regular expression used to match username mentions in comments.
+MENTION_RE = r'(?<!\S)(@\w+)'
--- a/bodhi/messages/schemas/base.py
+++ b/bodhi/messages/schemas/base.py
@@ -1,0 +1,234 @@
+# Copyright (C) 2018-2019 Red Hat, Inc.
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+"""
+Message schema for Bodhi.
+
+Each message is defined as a Python class. For details, see `fedora-messaging
+<https://fedora-messaging.readthedocs.io/en/stable/>`_ documentation on
+messages.
+"""
+
+import re
+import typing
+
+from fedora_messaging import message
+from fedora_messaging.schema_utils import user_avatar_url
+
+from bodhi import MENTION_RE
+
+
+SCHEMA_URL = 'https://bodhi.fedoraproject.org/message-schemas'
+
+
+class BodhiMessage(message.Message):
+    """A base class for Bodhi messages."""
+
+    @property
+    def app_icon(self) -> str:
+        """
+        Return a URL that points to the application's icon.
+
+        This is used when displaying the message to users.
+
+        Returns:
+            A URL for Bodhi's app icon.
+        """
+        return "https://apps.fedoraproject.org/img/icons/bodhi.png"
+
+    @property
+    def agent(self) -> typing.Union[str, None]:
+        """Return the agent's username for this message.
+
+        Returns:
+            The agent's username, or None if the body has no agent key.
+        """
+        return self.body.get('agent', None)
+
+    @property
+    def agent_avatar(self) -> typing.Union[None, str]:
+        """
+        Return a URL to the avatar of the user who caused the action.
+
+        Returns:
+            The URL to the user's avatar, or None if username is None.
+        """
+        username = self.agent
+        if username is None:
+            return None
+        return user_avatar_url(username)
+
+    @property
+    def usernames(self) -> typing.List[str]:
+        """
+        List of users affected by the action that generated this message.
+
+        Returns:
+            A list of affected usernames.
+        """
+        users = []
+        if self.agent is not None:
+            users.append(self.agent)
+
+        if 'comment' in self.body:
+            text = self.body['comment']['text']
+            mentions = re.findall(MENTION_RE, text)
+            for mention in mentions:
+                users.append(mention[1:])
+
+        users = list(set(users))
+        users.sort()
+        return users
+
+    @property
+    def containers(self) -> typing.Iterable[str]:
+        """
+        List of containers affected by the action that generated this message.
+
+        Returns:
+            A list of affected container names.
+        """
+        return []
+
+    @property
+    def modules(self) -> typing.Iterable[str]:
+        """
+        List of modules affected by the action that generated this message.
+
+        Returns:
+            A list of affected module names.
+        """
+        return []
+
+    @property
+    def flatpaks(self) -> typing.Iterable[str]:
+        """
+        List of flatpaks affected by the action that generated this message.
+
+        Returns:
+            A list of affected flatpaks names.
+        """
+        return []
+
+
+class BuildV1(typing.NamedTuple):
+    """
+    A model for referencing a Build.
+
+    Attributes:
+        nvr: The koji id of the build.
+    """
+
+    nvr: str
+
+    @property
+    def package(self) -> str:
+        """Return the name of the package that this build is associated with."""
+        return self.nvr.rsplit('-', 2)[0]
+
+    @staticmethod
+    def schema() -> dict:
+        """Return a schema snippet for a Build."""
+        return {
+            'type': 'object',
+            'description': 'A build',
+            'properties': {
+                'nvr': {
+                    'type': 'string',
+                    'description': 'The nvr the identifies the build in koji'
+                },
+            },
+            'required': ['nvr']
+        }
+
+
+class UpdateV1(typing.NamedTuple):
+    """
+    A model for referencing an Update object.
+
+    Attributes:
+        alias: The alias of the update.
+        builds: A list of builds associated with the update.
+        submitter: The id of the user who submitted the update.
+    """
+
+    alias: str
+    builds: typing.Iterable[BuildV1]
+    user: 'UserV1'
+    status: str
+    request: typing.Union[None, str]
+
+    @property
+    def packages(self) -> typing.Iterable[str]:
+        """Return a list of package names included in this update."""
+        return [b.package for b in self.builds]
+
+    @staticmethod
+    def schema() -> dict:
+        """Return a schema snippet for an Update."""
+        return {
+            'type': 'object',
+            'description': 'An update',
+            'properties': {
+                'alias': {
+                    'type': 'string',
+                    'description': 'The alias of the update'
+                },
+                'builds': {
+                    'type': 'array',
+                    'description': 'A list of builds included in this update',
+                    'items': {'$ref': '#/definitions/build'}
+                },
+                'request': {
+                    'type': ['null', 'string'],
+                    'description': 'The request of the update, if any',
+                    'enum': [None, 'testing', 'obsolete', 'unpush', 'revoke', 'stable']
+                },
+                'status': {
+                    'type': 'string',
+                    'description': 'The current status of the update',
+                    'enum': [None, 'pending', 'testing', 'stable', 'unpushed', 'obsolete',
+                             'side_tag_active', 'side_tag_expired']
+                },
+                'user': UserV1.schema(),
+            },
+            'required': ['alias', 'builds', 'request', 'status', 'user']
+        }
+
+
+class UserV1(typing.NamedTuple):
+    """
+    A model for referencing a User object.
+
+    Attributes:
+        name: The User's account name
+    """
+
+    name: str
+
+    @staticmethod
+    def schema() -> dict:
+        """Return a schema snippet for a User."""
+        return {
+            'type': 'object',
+            'description': 'The user that submitted the override',
+            'properties': {
+                'name': {
+                    'type': 'string',
+                    'description': "The user's account name"
+                },
+            },
+            'required': ['name']
+        }
--- a/bodhi/messages/__init__.py
+++ b/bodhi/messages/__init__.py
@@ -1,0 +1,16 @@
+# Copyright (C) 2018  Red Hat, Inc.
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+"""This package defines Bodhi's messages."""
--- a/bodhi/messages/schemas/__init__.py
+++ b/bodhi/messages/schemas/__init__.py
@@ -1,0 +1,16 @@
+# Copyright (C) 2019 Red Hat, Inc.
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+"""This package defines Bodhi's message schemas."""
--- a/bodhi/messages/schemas/buildroot_override.py
+++ b/bodhi/messages/schemas/buildroot_override.py
@@ -1,0 +1,144 @@
+# Copyright (C) 2018-2019 Red Hat, Inc.
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+"""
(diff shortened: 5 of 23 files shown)


abompard:
Hey guys, what do you think of this now?

bodhi_messages/bodhi_messages/schemas/stack.py
```diff
@@ -0,0 +1,76 @@
...
+
+    topic = "bodhi.stack.save"
+    _summary_tmpl = "{agent} updated the \"{name}\" stack"
+
+
+class StackDeleteV1(StackMessage):
+    """Sent when a stack is deleted."""
+
+    topic = "bodhi.stack.delete"
+    _summary_tmpl = "{agent} deleted the \"{name}\" stack"
```
bowlofeggs: I'm sad to say that I'm actually planning on [dropping stacks](https://github.com/fedora-infra/bodhi/issues/2241) as part of [bodhi 4.0.0](https://github.com/fedora-infra/bodhi/projects/6). That said, no harm in accepting this change, it's just sad because I'll be removing it really soon.

bodhi_messages/setup.py
```diff
@@ -0,0 +1,81 @@
...
+    version="1.0.0",
+    description="JSON schema for messages sent by Bodhi",
+    long_description=README,
+    url="https://github.com/fedora-infra/bodhi/",
+    # Possible options are at https://pypi.python.org/pypi?%3Aaction=list_classifiers
+    classifiers=[
+        "License :: OSI Approved :: GNU General Public License v2 or later (GPLv2+)",
+        "Operating System :: POSIX :: Linux",
+        "Programming Language :: Python :: 2",
+        "Programming Language :: Python :: 2.7",
```
bowlofeggs: Is there a universe where I can avoid adding a new Python 2 package you think? Do you think it would be fair for me to say that any packages that want to consume Bodhi's cool new messages need to also use Python 3? @hroncok might be happiest if we can be pure Python 3 here.

I've got an open pull request that drops Python 2 support from the server, FWIW: #2759.

bowlofeggs (commented):
I like what I see here. Can we squash some of the commits to be more atomic? The flake8 commit is the main one that stands out as being sensible to squash.

Also I think we should rebase, because the first commit seems to be merged into develop.

@jeremycline are you satisfied with all the other changes here?

bodhi_messages/setup.py
hroncok: I mean, do whatever you want upstream, just don't put the Python 2 part to Fedora ever :)

That said, Python 3 only upstream works best, because nobody is tempted to package it into Fedora as a python2 package :D

jeremycline: You'll only need Python 2 one or more Bodhi message consumers require it. I'm not sure what all consumes Bodhi messages, but it probably wouldn't hurt to start with Python 3 only.

jeremycline:
> @jeremycline are you satisfied with all the other changes here?

It looks like a solid start to me. Once this is merged the work remaining would be:

- Add JSON schema to each class to validate messages

- Go through the Bodhi code and create the various message classes rather than publishing via ``bodhi.notifications`` using a generic message object.

bodhi_messages/setup.py
bowlofeggs: I say let's just force Python 3 for people who want to use this package. That way I don't have to do a big deprecation/removal in the future. People who aren't ready to switch can just keep using the fedmsg relay.

bowlofeggs:
The ```f29-pydocstyle``` test actually passed, it seems that it was just the GitHub notifier itself that failed (weird).

bowlofeggs:
I'm going to rebase this and force push.

bodhi_messages/bodhi_messages/schemas/stack.py
bowlofeggs: Stacks is actually dropped now, so I will drop this code before merging.

bodhi_messages/setup.py
bowlofeggs: I'm going to edit this commit to make it be Python 3 only.

bodhi_messages/setup.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+here = os.path.abspath(os.path.dirname(__file__))
+with open(os.path.join(here, "README.rst")) as fd:
+    README = fd.read()
+
+
+setup(
+    name="bodhi_messages",
+    version="1.0.0",
```
bowlofeggs: This opens the door to me needing to solve a problem I currently have. Until now, the Bodhi server and client packages are versioned together in lock-step. This has advantages and disadvantages. The main advantage is that it's easy and pretty clean. The main disadvantage is that CLI users often have to wait for a new Fedora release just to get new features when I make X releases for the server, even though the CLI itself rarely has backwards incompatible changes.

There are two alternatives I've been considering. One is what you have done here - give each package its own ```setup.py``` and version. The advantage is that it allows me to version each component separately, while also allowing me to make single commits that adjust each of them. The key to the first advantage is that it allows me to be more granular with my semantic versioning. Sometimes I make an X release that doesn't really affect certain classes of users (like CLI or soon to be fedora messaging users). The downside is that I'll have to start tagging them and branching them separately, which might be very strange (especially the branching). (The ```setup.py``` I've got now is suuuuuper incorrect anyway, and just abuses how ```setuptools``` works and uses it in an undocumented manner.)

Another option I've considered is giving each component its own git repo. This way the branching model and tags can remain sensible. The downside is that changes that affect both now have to coordinate across repos (and so of course can't be one commit), and it's more complex to manage three repositories instead of one.

The branching problem does strike me as a tricky one. I currently make X.Y branch names when I want to release, and then cherry pick patches into it as needed. If we have three components with three X.Y's, it's not clear how that would be sensible anymore (unless I just say that the X.Y is for the server version and ignore the weirdness of the other two components. Maybe the weirdness isn't so bad). I don't know how much I love the extra complication of having three repos.

I'm a little inclined to say that all of these concerns should be separate of this pull request since this pull request is about adding schemas to the messages. Would it be OK for now if we just made this do what the CLI does and do even more weirdness in the ```setup.py``` for the sake of keeping this pull request more atomic? If we want to break these components out (which I'm not opposed to), it might be nice to do in a separate PR, especially since I'd like to do it in a consistent fashion for the CLI and base package. An alternative is to make it another commit on this PR if it's important to you to start off with version ```1.0.0```.

What do you guys think?

bowlofeggs: See also: https://github.com/fedora-infra/bodhi/issues/2401

bodhi_messages/setup.py
bowlofeggs: I'm adding this patch on top of the first commit:

```
diff --git a/bodhi_messages/setup.py b/bodhi_messages/setup.py
index 4dcbc521..b5a55bb5 100644
--- a/bodhi_messages/setup.py
+++ b/bodhi_messages/setup.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/python3
 #
 # Copyright (C) 2018  Red Hat, Inc.
 #
@@ -35,11 +35,7 @@ setup(
     classifiers=[
         "License :: OSI Approved :: GNU General Public License v2 or later (GPLv2+)",
         "Operating System :: POSIX :: Linux",
-        "Programming Language :: Python :: 2",
-        "Programming Language :: Python :: 2.7",
         "Programming Language :: Python :: 3",
-        "Programming Language :: Python :: 3.4",
-        "Programming Language :: Python :: 3.5",
         "Programming Language :: Python :: 3.6",
         "Programming Language :: Python :: 3.7",
     ],
```

I made the shebang opinionated about Python 3 and I also opted to only support the versions of Python 3 in current supported Fedora releases, since I currently only "support" Fedora as an installation target.

bowlofeggs (changes requested):

bowlofeggs:
I talked with @jeremycline about this pull request and he seemed OK with me changing it to use the same setup.py we use for the other subpackages. What do you think, @abompard?

abompard:
I don't really have an opinion on that, as long as it produces a separate installable python package, I'm fine with it.

bowlofeggs:
OK. I'm going to move ```bodhi_messages``` into ```bodhi/messages``` and merge its ```setup.py``` with the main one for now, contributing further to the horror that lies within that file. I am interested in finding a better way in the future, but I'm not sure what the best way is at this time.

bowlofeggs:
Alright, I've pushed up a commit that moves things around in the way I proposed earlier. It will not pass CI until https://github.com/fedora-infra/bodhi/pull/2987 is merged, however.

@abompard @jeremycline - if I could get a +1 from one of you, I'll take it from here (I'm going to squash all the commits into 1 if that's OK with you) and I'll merge this after I can rebase it upon #2987.

bowlofeggs:
I just pushed up another change to my commit - I removed the stacks messages because stacks got dropped a while ago. I renamed mash to compose since that change also happened recently. Lastly, I moved the tests into a schema/ folder since the schemas they are testing are in a folder by that name.

bowlofeggs:
I added one more commit to make pydocstyle happy.

abompard:
Yep, +1 from me.

bowlofeggs:
Turns out there are still issues with coverage on this pull request, so I'll add some more commits to bump up the coverage.

bodhi/messages/schemas/base.py
```diff
@@ -0,0 +1,123 @@
...
+
+    def _get_agent_username(self):
+        """Return the agent's username for this message.
+
+        Returns:
+            str or None: The agent's username.
+        """
+        username = self._body.get('agent', None)
+        if isinstance(username, dict):
+            username = username['name']
```
bowlofeggs: @abompard this pattern above is found in several of the message classes and is causing coverage issues because the test messages don't generally have users expressed this way. This code was copied from the old fedmsg code, right? Is it likely that the messages were once this way, and so this code was there for a transition long ago, and Bodhi doesn't encode usernames this way anymore? How complete are the set of messages in the tests? Do they include one of every type of message that Bodhi sends today?

abompard: Yeah it was copied from old fedmsg code, you can very probably drop it, those messages may not be going around on the bus anymore.

bowlofeggs: OK - I'm currently going through the tests and making them be consistent with the data structures I see in datagrepper. I am dropping the tests that check "legacy" data.

bowlofeggs:
I was hoping I'd be able to finish this today but I ran out of time. I got a lot done though!

I took the time to par the tests and message classes down to just cover the current messages that Bodhi 4 sends, and not any legacy messages. This is important to me because this is a contract with our users and I would like to start off clean, especially since I don't know the history here.

Since I needed to learn the message schemas for all the messages we send to do this, it seemed like a sensible time to also add a json schema to our messages so we can have enforcement.

I decided to only add the keys I am committing to maintain to these schemas, rather than documenting the full message that we send today. This way we are free to clean up some of the cruft later on, and it also made it easier to define schemas. We can always add more keys to the schemas later, so I'd rather err by under promising rather than over promising. Thus, the schemas are pretty minimal. I also added properties to allow users to access these documented items, so nobody should ever be using Message.body.

I got all the files done except for ```update.py``` and ```updates.py```. I will try to finish those next week.

bowlofeggs:
I also marked this for Bodhi 4 and for CI gating, because I think it's the only chance we have to stop users from using Message.body. By defining a schema that is minimal with Bodhi 4, we will be free to change our messages to be far less messy without breaking our contract.

If we release without doing this, we will not have a contract and will have to be much more careful when making changes.

bowlofeggs:
Man, I thought I could finish this today but I ended up refactoring a bit more than I expected to reuse code and schema definitions. I am getting close though, and I'm really liking how this is shaping up. Hopefully I will finish tomorrow.

bodhi/messages/schemas/updates.py
```diff
@@ -0,0 +1,90 @@
...
+    """Sent when EPEL updates are synced."""
+
+    topic = "bodhi.updates.epel.sync"
+    _product = "EPEL"
+
+
+class UpdatesFedoraSyncV1(UpdatesSyncMessage):
+    """Sent when Fedora updates are synced."""
+
+    topic = "bodhi.updates.fedora.sync"
```
bowlofeggs: Despite the topic, I don't think the messages in this file are sent by Bodhi. They seem to be sent by the script that does the sync. Note the username in this one, for example:

https://apps.fedoraproject.org/datagrepper/id?id=2019-a5d230e5-e086-4588-ac16-48b14335c4b0&is_raw=true&size=extra-large

When Bodhi sends messages, the username is apache:

https://apps.fedoraproject.org/datagrepper/id?id=2019-cd7f6213-1c0e-465e-9793-5c4cec6a4607&is_raw=true&size=extra-large

I'm conflicted about whether we should have these schemas in Bodhi or not, since Bodhi doesn't actually send them and thus isn't actually in control of their format. What do you think?

bowlofeggs:
Alright, one more day that I didn't get quite done, but there are only either 0 or 2 schemas left to write, depending on what we think about my question on the sync script messages.

bowlofeggs:
After sleeping on it, I'm going to drop these schemas. Bodhi doesn't send them, so Bodhi can't know when/if they change. It's not sensible for Bodhi to own them.

bowlofeggs:
Both of the CI failures are due to hitting postgres startup timeouts. I filed https://github.com/fedora-infra/bodhi/issues/3074 about that problem. The tests all pass locally for me.

bowlofeggs:
This is now ready for review. Could one of @abompard or @jeremycline confirm that this is sane?

I'd prefer to squash this whole thing down to one commit, if that's OK with both of you. When doing so, I will indicate that all three of us authored it. I haven't squashed it yet so it'll be easier for you to see which changes I made.

bodhi/messages/schemas/base.py
```diff
@@ -0,0 +1,232 @@
...
+        Returns:
+            A list of affected usernames.
+        """
+        users = []
+        if self.agent is not None:
+            users.append(self.agent)
+
+        if 'comment' in self.body:
+            text = self.body['comment']['text']
+            mentions = re.findall(r'@\w+', text)
```
abompard: I suppose you'll get domain names if somebody adds an email address to comments, or an URI or something like it. I think there's a way to say that the mention must start with or at a word boundary, but otherwise you could put something like `^|\s+` before the `@` maybe? I'm not sure.

bodhi/messages/schemas/base.py
```diff
@@ -0,0 +1,232 @@
...
+
+    nvr: str
+
+    @property
+    def package(self) -> str:
+        """Return the name of the package that this build is associated with."""
+        return self.nvr.rsplit('-', 2)[0]
+
+    @staticmethod
+    def schema() -> dict:
```
abompard, line 142: What's the rationale for using a `schema()` method instead of the `body_schema` class attribute that the root `Message` class has?

abompard (commented):
Looks very nice! Just a few questions.

bodhi/messages/schemas/base.py
bowlofeggs: Heh, well I had originally made it a classmethod that I was passing some parameters to so the caller could specify the description, but then decided that was overcomplicated and changed it to be this, but it didn't occur to me to make it consistent with the ```Message``` class (note, this is actually not a ```Message```, it's just a ```NamedTuple```). However, I think it would be nice to be consistent, so I'll change it to use the ```body_schema``` as you suggest.

bowlofeggs: @abompard oh you know what? I'm not sure I can easily make it be body_schema. Since this is a ```NamedTuple```, that would mean that body_schema becomes one of the tuple attributes, which I'm not sure we actually want to do. However, I think I could name the method ```body_schema``` if you like. Maybe I could make it be an ```@property``` as a workaround? It wouldn't use ```self```, but might be a nice way to get a similar 'API'?

bowlofeggs: When I think about it though, I tend to dislike making instance methods that don't actually use self. And actually, we also want to be able to use this without instantiating one. I can make it a ```classmethod``` and ```property``` as a better workaround.

bowlofeggs: Technically the classmethod also doesn't use it's ```cls``` parameter, but I guess that's OK.

bowlofeggs: One more thing that might not fit nicely is that this isn't actually a ```body_schema``` anyway - it's not a complete schema, it's just a snippet of a schema. So it wouldn't be the schema for any actual body. A better name might just be ```schema_snippet```?

abompard: Ah you're right, I thought it was a `Message` subclass. Then the staticmethod is fine, both names are OK (`schema` and `schema_snippet`).

bodhi/messages/schemas/base.py
bowlofeggs: So Bodhi actually already has an RE for this that seems to work, but it's part of the server code:

https://github.com/fedora-infra/bodhi/blob/3.13.3/bodhi/server/ffmarkdown.py#L33

However, the server *and* this new message package both use the bodhi base package, which to date has no code. I think I'll just move that RE to live in bodhi/__init__.py and have the other two use it from there.

bowlofeggs: Also, I wrote a test that uses an e-mail address, and you are right - it does match that. Using the other RE seems to work as expected.

bowlofeggs: Also, FTR, I didn't write this code - I think it came from wherever this message code lived before this ☺

abompard: As they say : "... and now you have two problems" ;-)

bodhi/messages/schemas/base.py
bowlofeggs: @abompard After all the above thoughts, I'm kind of inclined to just leave it as is. It really is a staticmethod, but you can't @property a static method, and it also really isn't a body_schema, because it's not a complete schema, so I think the variation on the name is sensible.

What are your thoughts?

bodhi/messages/schemas/base.py
bowlofeggs: @abompard I pushed up a new commit to fix this, and it includes a test to ensure it stays fixed.

bowlofeggs:
This is ready for re-re-re-re-re-review.

bodhi/messages/schemas/base.py
abompard: Yeah it's fine

bodhi/__init__.py
```diff
@@ -17,3 +17,6 @@
 # along with this program; if not, write to the Free Software
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
 """Fedora's update manager."""
+
+# This is a regular expression used to match username mentions in comments.
+MENTION_RE = r'(?<!\S)(@\w+)'
```
abompard, line 22: Well you could have `re.compile`'d the regexp right here and gained execution speed but that's really a detail ;-)

abompard (approved):
Yep, looks good! :-)

bodhi/__init__.py
bowlofeggs: Good idea, I'll make that so.

bowlofeggs: I'm applying this patch before merging:

```
diff --git a/bodhi/__init__.py b/bodhi/__init__.py
index 9d79ab1d..d08279bb 100644
--- a/bodhi/__init__.py
+++ b/bodhi/__init__.py
@@ -18,5 +18,8 @@
 """Fedora's update manager."""
 
+import re
+
+
 # This is a regular expression used to match username mentions in comments.
-MENTION_RE = r'(?<!\S)(@\w+)'
+MENTION_RE = re.compile(r'(?<!\S)(@\w+)')
```

bowlofeggs (approved):

bodhi/__init__.py
bowlofeggs: Actually I'm going to unapply that patch - ffmarkdown seems to really want text. Also, Python does cache recently used regular expressions, so I think we effectively get the same performance here anyway since we don't use any other regular expressions in the package.

bowlofeggs:

bowlofeggs:

PR #2844, opened 2018-12-13
Use --force-rm on bodhi-ci build commands.

bowlofeggs:


puiterwijk (changes requested):
This would mean the entire container needs to be rebuilt from scratch every time you run it though?
What is the reason for doing this, since this means that DNF must be run every time someone wanted to rerun the container tests?

puiterwijk (approved):
Never mind, I misread the docs for `--force-rm`: it remotes intermediate *containers*, not images.

PR #2845, opened 2018-12-13
Improve readability of release notes

abitrolly:
https://screenshots.firefox.com/u9k8X2UfLGsxaPoZ/bodhi.fedoraproject.org

I am not sure if Sphinx will pick up this stylesheet, but there is an empty custom.css in web site sources, so it may work.


bowlofeggs (approved):
It seems to work OK with sphinx-build, so I expect it'll work in production the same way. Thanks!

bowlofeggs:
@abitrolly Can you edit your commit message to sign the Developer Certificate of Origin, as described here?

bowlofeggs (changes requested):
@abitrolly Can you edit your commit message to sign the Developer Certificate of Origin, as described here?

abitrolly:
I thought GitHub signs it automatically with Verified mark. Adjusted manually.

bowlofeggs:
@abitrolly the verified mark means that the commit was gpg signed, but that's different than the DCO agreement. Thanks, looks good now!

bowlofeggs (approved):

abitrolly:
How does that improve the workflow if developers will still have to go to the web interface to approve the PR? Instead of approving and waiting for merge and fail, it would be faster to do rebase and merge directly, no?

bowlofeggs:
@abitrolly I'm not sure I understand your question. Can you rephrase?

abitrolly:
The comment is about **mergify**.

bowlofeggs:
@abitrolly well Mergify doesn't just do a rebase and merge, it also waits for CI to pass. There are times that a rebased patch fails CI when it passed previously (because another patch was merged), so this is intended to catch those sorts of issues. If I were to run the tests on every patch after rebasing and then merge, that would take a lot of my time.

abitrolly:
Waiting for CI makes sense.

PR #2847, opened 2018-12-14
Print bodhi-ci's status report after every job exits.

bowlofeggs:
One thing that has bugged me about bodhi-ci is that it was not
always clear what it was doing at the moment. This commit prints
out a summary report of all the jobs after each one finishes, and
adds some new states as well (waiting, running, and skipped).

Signed-off-by: Randy Barlow <randy@electronsweatshop.com>

PR #2852, opened 2018-12-14
Add a timer feature to bodhi-ci.

bowlofeggs:
bodhi-ci will now report the time that each of its tests took to
run.


devel/ci/bodhi-ci
```diff
@@ -434,16 +437,19 @@ class Job(object):
         if self.cancelled:
             color_start = '\033[0;33m' if tty else ''
             color_end = '\033[0m' if tty else ''
-            return '{}:  {}CANCELED{}\n'.format(self.label, color_start, color_end)
+            return '{}:  {}CANCELED{}  [{}]\n'.format(self.label, color_start, color_end,
```
abompard, line 440: Just printing the substraction will display the microseconds, won't it? Do you want to format it differently or is it fine as is?

abompard (approved):
Looks good to me, just a minor comment on formatting.

abompard:
Actually I think there may be a bug, I'm getting the following traceback:
```
File "./devel/ci/bodhi-ci", line 427, in run
    click.echo(self.summary_line)
  File "./devel/ci/bodhi-ci", line 474, in summary_line
    self.label, color_start, color_end, self._finish_time - self._start_time,
AttributeError: 'IntegrationJob' object has no attribute '_finish_time'
```
There's a display of `self.summary_line` on line 427 inside the try/except, while `self._finish_time` has not yet been set. I think that the `_finish_time` should be set there as well.

bowlofeggs:
On Thu, 2018-12-20 at 12:46 +0000, Aurélien Bompard wrote:
> Just printing the substraction will display the microseconds, won't
> it? Do you want to format it differently or is it fine as is?

Yeah it does print the microseconds. I spent like 5 minutes trying to
figure out the format string for timedelta objects, but it turns out
that strftime() is only on datetime objects and not timedelta objects.
Additionally, timedelta objects' microseconds attribute is readonly, so
you can't just set its microseconds to 0. I guess you could set the
start and end time's microseconds to 0. Anyways, it was started to feel
more complicated than it was worth, so I decided to just let it print
microseconds.

I'm a bit surprised that timedelta doesn't support the str format
language.

bowlofeggs:
> Actually I think there may be a bug, I'm getting the following traceback:
> 
> ```
> File "./devel/ci/bodhi-ci", line 427, in run
>     click.echo(self.summary_line)
>   File "./devel/ci/bodhi-ci", line 474, in summary_line
>     self.label, color_start, color_end, self._finish_time - self._start_time,
> AttributeError: 'IntegrationJob' object has no attribute '_finish_time'
> ```
> 
> There's a display of `self.summary_line` on line 427 inside the try/except, while `self._finish_time` has not yet been set. I think that the `_finish_time` should be set there as well.

Oh no!

That line is actually removed in https://github.com/fedora-infra/bodhi/pull/2847. We could either get that reviewed/merged, or I can submit a PR that just deletes that one line.

PR #2860, opened 2018-12-17
Drop the 'sent-at' header from messages so tests pass reliably

### LINKED ISSUE #2858: Intermittent test failure in one of Bodhi's fedora-messaging tests
There's a [test assertion](https://github.com/fedora-infra/bodhi/blob/e30c5f2/bodhi/tests/server/test_notifications.py#L141) in Bodhi that occasionally fails like this:

```
bodhi/tests/server/test_notifications.py .......F

=================================== FAILURES ===================================
_____________________ TestPublish.test_fedmsg_publish_off ______________________

self = <bodhi.tests.server.test_notifications.TestPublish testMethod=test_fedmsg_publish_off>
mock_init = <MagicMock name='init' id='140416901561296'>

    @mock.patch.dict('bodhi.server.config.config', {'fedmsg_enabled': False})
    def test_fedmsg_publish_off(self, mock_init):
        """Assert when fedmsg publish is off, fedora-messaging messages are queued."""
        expected_msgs = [api.Message(topic='org.fedoraproject.dev.bodhi.demo.topic',
                                     body={u'such': 'important'})]
        session = Session()
    
        notifications.publish('demo.topic', {'such': 'important'})
    
        self.assertIn('messages', session.info)
        for expected, actual in zip(expected_msgs, session.info['messages']):
>           self.assertEqual(expected, actual)
E           AssertionError: Message(id='7e76630d-3e13-4478-9e82-f4f4a5975c0e', topic='org.fedoraproject.dev.bodhi.demo.topic', body={u'such': 'important'}) != Message(id='74855ca5-6638-472b-9018-6e5bde011267', topic='org.fedoraproject.dev.bodhi.demo.topic', body={u'such': u'important'})

bodhi/tests/server/test_notifications.py:141: AssertionError
```

The problem seems to be that the ```Message.id``` attribute is not the same between the two messages. We could fix it by either stopping to compare the ```id```, or maybe by contributing an assertion method to ```fedora-messaging``` that allows users to compare messages without having to compare the ```id```s.

@jeremycline what would you recommend here?

jeremycline:
This header is the time when the message was sent. If the test runs
quickly enough the two time stamps could be equal, which is why this
doesn't always fail.

fixes #2858

Signed-off-by: Jeremy Cline <jcline@redhat.com>


abompard:
Could you rebase on develop to get CI to pass please?

bowlofeggs (approved):

bowlofeggs:
I'm going to re-rebase this so it picks up the new ```mypy``` test.

PR #2862, opened 2018-12-17
Add a bugzilla_api_key setting and make a 3.12.0 release

bowlofeggs:
This pull request has two commits. The first adds a ```bugzilla_api_key``` setting and the second makes a new release.


bodhi/server/bugs.py
```diff
@@ -93,7 +93,10 @@ def _connect(self):
         password = config.get('bodhi_password')
         url = config.get("bz_server")
         log.info("Using BZ URL %s" % url)
-        if user and password:
+        if config.get('bugzilla_api_key'):
+            self._bz = bugzilla.Bugzilla(url=url, api_key=config.get('bugzilla_api_key'),
```
jeremycline, line 97: Not that it's wrong, but you know the key is present and has a value so you could just do ``config['bugzilla_api_key']``, right?

docs/user/release_notes.rst
```diff
@@ -2,6 +2,36 @@
 Release notes
 =============
 
+v3.12.0
+-------
+
+This is a small feature release.
```
jeremycline, line 8: Other features have been merged, right? I assume you're going to do some cherry-pick dance so these release notes are true?

jeremycline (approved):
I had one incredibly tiny nitpick you should feel free to ignore and just spelled out what I assume your plans are for the release notes.

docs/user/release_notes.rst
bowlofeggs: Yeah it's merged here: https://github.com/fedora-infra/bodhi/pull/2861

bodhi/server/bugs.py
bowlofeggs: No problem, I changed it to that style. I prefer that too anyway - I was just using the style from elsewhere here, which predates configs having defined values in Bodhi.

bowlofeggs:
On Mon, 2018-12-17 at 13:27 -0800, Jeremy Cline wrote:
> Other features have been merged, right? I assume you're going to do
> some cherry-pick dance so these release notes are true?

Oh also to clarify, I only plan to include the two mentioned patches in
this release. I would have made this a .z release, but teeeechnically
adding a new setting is a feature (even though it fixes what seems like
a bug) so it's a featurebug release? Hard to say…

bowlofeggs:
The integration test is an intermittent problem that is not due to this patch. I've documented that problem here: https://github.com/fedora-infra/bodhi/issues/2863

I'm going to manually merge this.

Issue #2856, opened 2018-12-17
Drop Python 2 support for bodhi.client


bowlofeggs:
We need to wait until no packages depend on it, but when we are able to we should drop Python 2 support from ```bodhi.client```.

bowlofeggs:
No packages depend on the Python 2 version of Bodhi in Fedora 30 now.

bowlofeggs:
This change is planned for inclusion in the upcoming 4.0.0 release: #3221

bowlofeggs:
Bodhi 4.0.0 beta is built and deployed to staging:

https://copr.fedorainfracloud.org/coprs/bowlofeggs/bodhi-pre-release
https://bodhi.stg.fedoraproject.org/

PR #2871, opened 2018-12-19
Drop support for Python 2 in bodhi.server.

### LINKED ISSUE #1970: Drop support for pkgdb
Hello.

We had a short discussion with @bowlofeggs about packagedb-cli dependency and how to deal with it during making Bodhi Python 3 compatible.

The fact is that PackageDB is read-only from August and its CLI won't be ported to Python 3. Another fact is that there are still some installations of Bodhi using it and we should release all backward-incompatible changes in one release (if possible).

I think that I can postpone removal of PackageDB-cli from Bodhi until the Python 3 compatibility will be ready for merge. In the meantime, I'll use decorators/conditions to skip related code and tests. Then we can decide whether we want to remove PackageDB-cli from Bodhi or let it there until its EOL.

### LINKED ISSUE #2759: Drop Python 2 support for the server

bowlofeggs:
Python 2 support goes away upstream in two months, so it is time
to embrace the glorious future.

This change also drops support for pkgdb integration, since the
pkgdb client library does not support Python 3.

Mock imports were moved to be grouped with the stdlib imports as
per PEP-8, since mock is part of the stdlib in Python 3.

fixes #1970
fixes #2759

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -12,14 +12,12 @@ pull_request_rules:
   - status-success=f28-flake8
   - status-success=f28-pydocstyle
   - status-success=f28-python2-unit
-  - status-success=f28-python2-diff-cover
   - status-success=f28-python3-unit
   - status-success=f28-python3-diff-cover
   - status-success=f29-docs
   - status-success=f29-flake8
   - status-success=f29-pydocstyle
   - status-success=f29-python2-unit
-  - status-success=f29-python2-diff-cover
   - status-success=f29-python3-unit
   - status-success=f29-python3-diff-cover
   - status-success=pip-docs
@@ -27,7 +25,6 @@ pull_request_rules:
   - status-success=pip-mypy
   - status-success=pip-pydocstyle
   - status-success=pip-python2-unit
-  - status-success=pip-python2-diff-cover
   - status-success=pip-python3-unit
   - status-success=pip-python3-diff-cover
   - status-success=f29-python3-integration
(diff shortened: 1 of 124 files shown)


.mergify.yml
```diff
@@ -12,21 +12,18 @@ pull_request_rules:
   - status-success=f28-flake8
   - status-success=f28-pydocstyle
   - status-success=f28-python2-unit
```
abompard, line 14: You can drop the unit tests too, no?

bowlofeggs: We can't yet, because the client still supports Python 2.

bowlofeggs: Python 2 support is planned to be dropped from the client in https://github.com/fedora-infra/bodhi/issues/2856. I am waiting just a bit to see if my dependencies will stop using it before making that change.

sebwoj:
https://github.com/fedora-infra/bodhi/blob/df7d153b1344c78160688054cdf64c28a2798810/devel/ansible/roles/bodhi/tasks/main.yml#L120

What is the pourpose of installing bodhi for py2 and py3?
I assume that one of the reasons is bodhi-client for py2.
Are there any others?

bowlofeggs:
On Thu, 2018-12-20 at 12:23 -0800, Sebastian Wojciechowski wrote:
> What is the pourpose of installing bodhi for py2 and py3?
> I assume that one of the reasons is bodhi-client for py2.
> Are there any others?

Yes, it's because the client still supports Python 2 for now. See
#2856.

Zlopez (approved):
LGTM

bowlofeggs:

bowlofeggs:

PR #2873, opened 2018-12-19
Force the commit_we_are_testing branch in CI.

bowlofeggs:
I forgot that CI jobs persist the git repository between runs. This
means that my last commit causes failures in CI because you can't
create a branch that already exists.

This commit forces the commit_we_are_testing branch to be the
current HEAD every run, which should work for the persisted branch
name.


cverna (approved):
+1

PR #2837, opened 2018-12-11
Make it possible to run the integration tests on different releases

abompard:
Also, the integration Dockerfiles are now based on the unittest Dockerfiles to avoid duplication.

PR #2880, opened 2018-12-21
Build the integration test containers with the build command.

bowlofeggs:
The integration test containers had been getting built when
bodhi-ci integration was run. This commit moves them to be built
when the bodhi-ci build command is used instead.


bowlofeggs:
It turns out that this PR is trying to do something similar to what @abompard did over in https://github.com/fedora-infra/bodhi/pull/2837, and so we might want to just go with his implementation.

What do you think Aurélien?

abompard:
Yeah I'd rather use my implementation because it also handles multiple Fedora releases.

PR #2881, opened 2018-12-21
Use containers from Fedora and CentOS registries for testing.

bowlofeggs:
I noticed that we were puling a couple containers from Dockerhub.
This commit converts us to using the Fedora and CentOS registries
instead.

### DIFF
--- a/devel/ci/integration/resultsdb/Dockerfile
+++ b/devel/ci/integration/resultsdb/Dockerfile
@@ -1,4 +1,4 @@
-FROM centos:7
+FROM registry.centos.org/centos:7
 LABEL \
   name="resultsdb" \
   vendor="Fedora Infrastructure" \
--- a/devel/ci/integration/tests/fixtures/backend.py
+++ b/devel/ci/integration/tests/fixtures/backend.py
@@ -49,4 +49,7 @@ def docker_network(docker_backend):
     """
     network = docker_backend.d.create_network(f"bodhi_test-{uuid.uuid4()}", driver="bridge")
     yield network
-    docker_backend.d.remove_network(network["Id"])
+    try:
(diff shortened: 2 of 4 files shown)


bowlofeggs:
Hmm, it seems that my PostgreSQL container change causes the integration test suite to hang. I'll see if I can figure that out in January.

bowlofeggs:
I looked into this some more today. It turns out that the Fedora container requires some environment variables to be defined for the postgresql user and password:

https://github.com/sclorg/postgresql-container/tree/generated/10

I've adjusted the patch to start up by defining the admin user's password, but it might be a lot of annoying work to get each container we have that connects to postgres to use this setting. I'm putting this down again for now.

cverna:
> I looked into this some more today. It turns out that the Fedora container requires some environment variables to be defined for the postgresql user and password:
> 
> https://github.com/sclorg/postgresql-container/tree/generated/10
> 
> I've adjusted the patch to start up by defining the admin user's password, but it might be a lot of annoying work to get each container we have that connects to postgres to use this setting. I'm putting this down again for now.

The images we have in Fedora's registry are the s2i images that are meant to be run in openshift that's why you need all the extra variables etc ...

bowlofeggs:
On Wed, 2019-03-06 at 23:18 -0800, Clement Verna wrote:
> The images we have in Fedora's registry are the s2i images that are
> meant to be run in openshift that's why you need all the extra
> variables etc ...

Yeah, makes sense. It also would be better for our CI tests to ensure
we can authenticate to PG since that's a more realistic deployment
model. I will hopefully get back to this patch one day.

bowlofeggs:
I don't think I'll be working on this any time soon, so I'm going to close this.

PR #2886, opened 2018-12-29
Commit the transaction with each update in approve_testing script

### LINKED ISSUE #1016: Bodhi's approve testing script should commit the transaction with each update
While working on #1009, I noticed that the [code that generates Bodhi nag e-mails](https://github.com/fedora-infra/bodhi/blob/2.2.4/bodhi/server/scripts/approve_testing.py#L52-L73) loops over all testing updates in a single transaction:

```
    with transaction.manager:
        testing = db.query(Update).filter_by(status=UpdateStatus.testing,
                                             request=None)
        for update in testing:
            # If this release does not have any testing requirements, skip it
            if not update.release.mandatory_days_in_testing:
                print('%s doesn\'t have mandatory days in testing' % update.release.name)
                continue
            # If this has already met testing requirements, skip it
            if update.met_testing_requirements:
                continue
            if update.meets_testing_requirements:
                print('%s now meets testing requirements' % update.title)
                text = config.get('testing_approval_msg') % update.days_in_testing
                update.comment(db, text, author='bodhi')

            # Approval message when testing based on karma threshold
            if update.stable_karma not in (0, None) and update.karma >= update.stable_karma \
                    and not update.autokarma:
                print('%s now reaches stable karma threshold' % update.title)
                text = config.get('testing_approval_msg_based_on_karma')
                update.comment(db, text, author='bodhi')
```

Due to this design, any error that happens causes the transaction to be rolled back for all changes made so far. This means that users might receive e-mails from Bodhi with comments that do not appear on the update.

Instead, we should commit our changes on each iteration of the loop.

sebwoj:
Fixes #1016

### DIFF
--- a/bodhi/server/scripts/approve_testing.py
+++ b/bodhi/server/scripts/approve_testing.py
@@ -90,6 +90,7 @@ def main(argv=sys.argv):
                 print('%s now reaches stable karma threshold' % update.title)
                 text = config.get('testing_approval_msg_based_on_karma')
                 update.comment(db, text, author=u'bodhi')
+                db.commit()
                 continue
 
             # If autokarma updates have reached the testing threshold, say something! Keep in mind
@@ -106,8 +107,8 @@ def main(argv=sys.argv):
                 notifications.publish(
                     topic='update.requirements_met.stable',
                     msg=dict(update=update))
+                db.commit()
 
-        db.commit()
     except Exception as e:
         print(str(e))
         db.rollback()
--- a/bodhi/tests/server/scripts/test_approve_testing.py
+++ b/bodhi/tests/server/scripts/test_approve_testing.py
@@ -20,7 +20,7 @@ This module contains tests for the bodhi.server.scripts.approve_testing module.
 """
 from datetime import datetime, timedelta
 from io import StringIO
-from mock import patch
+from mock import call, patch
 
 from fedora_messaging import api, testing as fml_testing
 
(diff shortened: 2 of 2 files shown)


bodhi/server/scripts/approve_testing.py
```diff
@@ -108,8 +109,8 @@ def main(argv=sys.argv):
                 notifications.publish(
                     topic='update.requirements_met.stable',
                     msg=dict(update=update))
+                db.commit()
```
bowlofeggs, line 110: Why not just slide this ```commit()``` left a notch and get rid of the first ```commit()```?

bowlofeggs (changes requested):

bodhi/server/scripts/approve_testing.py
sebwoj: My intention was to prevent unnecessary db.commit() calls(performance). For example, when query returns 100 updates, and only 2 will need call to db.commit(), 98 other will be unnecessary. But if you recommend db.comment() after each "for" iteration, I can change it.

bowlofeggs: Cool, that sounds like a fine justification to me.

bowlofeggs (approved):

bowlofeggs:
This fix is planned for inclusion in Bodhi 3.13: #2917.

bowlofeggs:

PR #2887, opened 2018-12-30
Remove_an_assert_statement_in_server.consumers.updates.UpdatesHandler…

### LINKED ISSUE #1200: bodhi.server.consumers.updates.UpdatesHandler.consumer uses an assert statement
```assert``` statements get removed from optimized code, which is what gets built in Koji (which is where production Bodhi builds come from). Thus, the assertion will not be present in production.

If the assertion is important, we should either reconcile the database discrepancy, or raise an Exception. It might make the most sense to reconcile the database discrepancy.

sebwoj:
….consumer

Fixed #1200


bowlofeggs (approved):
Thanks!

bowlofeggs:

bowlofeggs:

PR #2899, opened 2019-01-04
Remove the /admin/ API.

### LINKED ISSUE #1985: Delete bodhi.server.views.admin
Bodhi has a strange view that tells admin users what their username and principals are, but does not allow non-admin users to use it:

https://github.com/fedora-infra/bodhi/blob/3.0.0/bodhi/server/views/admin.py

When I visit https://bodhi.fedoraproject.org/admin/ I see:

```
{"principals": ["system.Everyone", "system.Authenticated", "bowlofeggs", "group:packager", "group:infra-sig", "group:bodhiadmin"], "user": "bowlofeggs"}
```

I don't know what the purpose of this view was, but I'm pretty sure we can delete it.

bowlofeggs:
The /admin/ API was not useful; it simply told you your username
and Pyramid principals, but only if you were in the list of Bodhi
admins. This commit removes it.

fixes #1985


sayanchowdhury (approved):

bowlofeggs:

bowlofeggs:

PR #2906, opened 2019-01-07
zxDrop bodhi.server.services.zz_redirects

### LINKED ISSUE #2869: Drop bodhi.server.services.zz_redirects
This module exists to redirect legacy Bodhi 1 URLs to the Bodhi 2 counterparts, but I don't think we need it anymore. Bodhi 2 is not backwards compatible with Bodhi 1, and Bodhi 4 will also be further incompatible.

sebwoj:
Fixes #2869


bowlofeggs (changes requested):
Cool, looks good. Let's add an entry to the release notes to the backwards incompatible changes section and then we should be good to merge.

bowlofeggs (changes requested):
There also seems to be a conflict to resolve.

bowlofeggs (approved):
Thanks!

bowlofeggs:

bowlofeggs:

PR #2909, opened 2019-01-07
Remove deprecated update.old_updateid

sebwoj:


bodhi/server/migrations/versions/6ac5b2dedb36_remove_legacy_old_updateid.py
```diff
@@ -0,0 +1,41 @@
...
+Revises: d986618207bc
+Create Date: 2019-01-06 17:11:13.421147
+"""
+from alembic import op
+import sqlalchemy as sa
+
+
+# revision identifiers, used by Alembic.
+revision = '6ac5b2dedb36'
+down_revision = 'd986618207bc'
```
bowlofeggs: It looks like another PR got merged that used this ```down_revision```, which is why some tests are failing. We need to rebase this migration on the latest one to get those tests to pass.

bowlofeggs (changes requested):
Let's add a release note since this is also backwards incompatible.

bowlofeggs (approved):
Thanks!

bowlofeggs:

bowlofeggs:

PR #2914, opened 2019-01-09
Compare strings to strings to in Greenwave decision in web UI.

bowlofeggs:
There was a bug discovered by Adam Williamson where the web
interface always used the stable decision context for updates,
even when they were pending testing. The bug was due to an if
statement that compared Enums to strings in the HTML template,
rather than comparing the string values of the Enums to string.

### DIFF
--- a/bodhi/server/templates/update.html
+++ b/bodhi/server/templates/update.html
@@ -3,6 +3,8 @@
 <%namespace name="json" module="json"/>
 <%
   from urllib.parse import urljoin
+
+  from bodhi.server import models
 %>
 
 <%block name="pagetitle">
@@ -362,7 +364,8 @@ ${update.type} update for ${update.beautify_title(amp=True) | n}
 
     get_greenwave_information({
         'product_version': '${update.product_version}',
-% if update.request == 'testing' and update.status == 'pending':
+% if update.request == models.UpdateRequest.testing and \
+    update.status == models.UpdateStatus.pending:
         'decision_context': 'bodhi_update_push_testing',
 % else:
         'decision_context': 'bodhi_update_push_stable',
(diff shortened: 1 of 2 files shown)


bowlofeggs:
I don't have time to write tests at the moment, so I applied the no-mergify label.

AdamWill:
maybe credit dmoluguw with discovering the bug - I only investigated it :)

AdamWill:
otherwise, looks :+1:

bodhi/server/templates/update.html
```diff
@@ -362,7 +362,7 @@
 
     get_greenwave_information({
         'product_version': '${update.product_version}',
-% if update.request == 'testing' and update.status == 'pending':
+% if update.request.value == 'testing' and update.status.value == 'pending':
```
pypingou: should we ``.lower()`` the values just in case?

bowlofeggs:
On Thu, 2019-01-10 at 03:37 -0800, Pierre-Yves Chibon wrote:
> should we .lower() the values just in case?

I don't think that's necessary for two reasons. First, the values are
PostgreSQL enums and so they cannot change without some very tricky
database migrations (PostgreSQL does not support removing or renaming
Enum values, so you have to do some gymnastics of creating a whole new
column, with a new temporary enum, copy all data from old column to
temporary column, then delete the old column and enum, then recreate
the old column and enum [but this time with the altered enum], and then
copy all data back from the old enum.) All this to say, nobody is going
to accidentally change the value to be capitalized.

Secondly, I will be adding tests to this patch as stated earlier, so
even if someone did do enum gymnastics my test will catch it.

cverna (approved):

bowlofeggs:
We can ignore the ```rawhide-build``` failure - that's a GPG problem present in Rawhide right now and is not related to this patch.

PR #2915, opened 2019-01-10
Drop support for CVEs.

### LINKED ISSUE #1998: The CVE model seems to be unused
While documenting ```models.py``` today, I noticed that the [```CVE``` model](https://github.com/fedora-infra/bodhi/blob/3.0.0/bodhi/server/models.py#L2704-L2714) is unused:

```
>>> m.CVE.query.count()
0L
```

bowlofeggs:
Bodhi has code to deal with CVEs, but it has never worked and was
not used. Since nobody has ever asked for it to work during my
tenure, I decided to trim it out of the code for now so we don't
have to maintain it.

fixes #1998

### DIFF
--- a/bodhi/client/__init__.py
+++ b/bodhi/client/__init__.py
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-# Copyright © 2014-2018 Red Hat, Inc. and others.
+# Copyright © 2014-2019 Red Hat, Inc. and others.
 #
 # This file is part of Bodhi.
 #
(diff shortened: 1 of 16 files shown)


pypingou (approved):
Less-code++

bowlofeggs:
I'm going to rebase this to fix the conflicts and hopefully get the tests to pass too.

bowlofeggs:
I am also applying this patch since another migration has since been added:

```
diff --git a/bodhi/server/migrations/versions/5c86a3f9dc03_drop_support_for_cve_tracking.py b/bodhi/server/migrations/versions/5c86a3f9dc03_drop_support_for_cve_tracking.py                                      
index 140c169a..66c670d4 100644
--- a/bodhi/server/migrations/versions/5c86a3f9dc03_drop_support_for_cve_tracking.py
+++ b/bodhi/server/migrations/versions/5c86a3f9dc03_drop_support_for_cve_tracking.py
@@ -28,7 +28,7 @@ import sqlalchemy as sa

 revision = '5c86a3f9dc03'
-down_revision = '8e9dc57e082d'
+down_revision = 'a418acf470f8'


 def upgrade():
```

bowlofeggs:

bowlofeggs:

PR #2924, opened 2019-01-11
get_pkg_committers_from_pagure: use flatpaks/ namespace for Flatpaks

owtaylor:
Flatpaks will be moved from sharing the modules/ namespace to their
own namespace in Fedora dist-git, so change the committer lookup
accordingly.

(I considered making this more complicated and trying the flatpaks/ namespace and falling back to modules/, but that would have required considerable added complexity, and I didn't want to leave that in the code just to make transition a bit easier. I'm thinking that we can just coordinate landing this with reconfiguring the other pieces needed for the flatpaks/ namespace, and if flatpak updates can't be filed for a few days, that's OK.)

### DIFF
--- a/bodhi/server/models.py
+++ b/bodhi/server/models.py
@@ -1058,8 +1058,6 @@ class Package(Base):
         # container. Flatpaks build directly from the 'modules' namespace
         if self.type.name == 'container':
             namespace = self.type.name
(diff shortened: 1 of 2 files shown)


bowlofeggs (approved):
Thanks!

PR #2929, opened 2019-01-13
Remove unnecessary assignment statement from models.update.set_request

sebwoj:

### DIFF
--- a/bodhi/server/models.py
+++ b/bodhi/server/models.py
@@ -2546,7 +2546,6 @@ class Update(Base):
         # keep the status at testing
         elif self.request in (UpdateRequest.stable, UpdateRequest.batched) and \
                 self.status is UpdateStatus.testing and action is UpdateRequest.revoke:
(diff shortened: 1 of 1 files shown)


bodhi/server/models.py
```diff
@@ -2553,7 +2553,6 @@ def set_request(self, db, action, username):
         # keep the status at testing
         elif self.request in (UpdateRequest.stable, UpdateRequest.batched) and \
                 self.status is UpdateStatus.testing and action is UpdateRequest.revoke:
-            self.status = UpdateStatus.testing
```
bowlofeggs, line 2549: hahah that's pretty funny. Nice find.

bowlofeggs (approved):
Thanks!

bodhi/server/models.py
sebwoj: At first I assumed some magic here, but turns out just a redundant assignment :)

PR #2930, opened 2019-01-13
Move mail templates to /etc/bodhi/email_templates

### LINKED ISSUE #2753: Mail templates should be in /etc/bodhi instead of in bodhi.server.templates
Bodhi currently looks for its e-mail templates in the code (i.e., in ```/usr/lib/python3.7/site-packages/bodhi/server```) instead of in its configuration (i.e., ```/etc/bodhi```), which makes it hard for deployments to customize. These are really configuration and so should be in ```/etc/bodhi``` instead.

I am marking this as backwards incompatible in case we just switch the location. Technically, we could add a location without it being backwards incompatible too.

sebwoj:
Fixes #2753

I'm not sure about final solution. How to copy email templates during deployment? Should it be done by some scripts or manually?

### DIFF
--- a/bodhi/server/config.py
+++ b/bodhi/server/config.py
@@ -475,7 +475,7 @@ class BodhiConfig(dict):
             'value': True,
             'validator': _validate_bool},
         'mail.templates_basepath': {
-            'value': 'bodhi:server/email/templates/',
+            'value': '/etc/bodhi/email_templates/',
             'validator': str},
         'mako.directories': {
             'value': 'bodhi:server/templates',
--- a/bodhi/server/mail.py
+++ b/bodhi/server/mail.py
@@ -26,7 +26,7 @@ from kitchen.text.converters import to_unicode, to_bytes
 
 from bodhi.server import log
 from bodhi.server.config import config
-from bodhi.server.util import get_rpm_header, get_absolute_path
+from bodhi.server.util import get_rpm_header
 
 
 #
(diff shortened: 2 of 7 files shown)


bodhi/server/config.py
```diff
@@ -475,7 +475,7 @@ class BodhiConfig(dict):
             'value': True,
             'validator': _validate_bool},
         'mail.templates_basepath': {
-            'value': 'bodhi:server/email/templates/',
```
bowlofeggs, line 478: I either didn't know or forgot that this was a setting (strange). I noticed that it's not documented in ```production.ini```, so I documented that here: https://github.com/fedora-infra/bodhi/pull/2931

Let's wait to merge this PR until that one is merged, so we can rebase this PR on that one and change that doc to show this new path.

I also want to merge the other one first so I can release it in a 3.y release series (this patch will go to 4.0.0).

bowlofeggs (changes requested):
As for the templates and getting them into ```/etc/bodhi```, I'm of the opinion that it would be better for us not to ship release specific templates as part of Bodhi's upstream code and leave that to downstream (i.e., the Fedora Infrastructure Ansible repository). However, I do think it would be good for Bodhi to have a generic e-mail template that gets used when a release doesn't have a template defined. So why don't we:

* Document how users can make e-mail templates, including documenting the template language used (an external link to the library we use is fine, no need to describe the syntax), the variables available for substitution, and how the templates are located.
* Create a default generic template (we can hardcode this in ```bodhi.server.mail```) to be used for releases that don't define a custom template. For this we don't use words like "Fedora" and just try to be very plain. Doesn't need to be fancy.
* Delete all the templates we have - we can just add them to the Fedora Ansible repository.

bowlofeggs:
Hey @sebwoj, do you still want to work on this patch?

bowlofeggs:
Closing due to inactivity.

PR #2931, opened 2019-01-15
Document the mail.templates_basepath setting.

bowlofeggs:

### DIFF
--- a/production.ini
+++ b/production.ini
@@ -126,6 +126,12 @@ use = egg:bodhi-server
 # domain for generated message IDs
 # message_id_email_domain = admin.fedoraproject.org
 
+# Where to search for e-mail templates. Each release has a mail_template attribute that defines the
+# name of a mail template, and Bodhi will look in the path defined by this setting to find the named
+# template.
+# mail.templates_basepath = bodhi:server/email/templates/
+
+
 ##
 ## Masher settings
 ##


production.ini
```diff
@@ -126,6 +126,11 @@ use = egg:bodhi-server
 # domain for generated message IDs
 # message_id_email_domain = admin.fedoraproject.org
 
+# Where to search for e-mail templates. Eash release has an attribute that defines the name of a
```
sebwoj: Just small typo: Eash

sebwoj (changes requested):

production.ini
bowlofeggs: Good catch, fixed!

production.ini
```diff
@@ -126,6 +126,11 @@ use = egg:bodhi-server
 # domain for generated message IDs
 # message_id_email_domain = admin.fedoraproject.org
 
+# Where to search for e-mail templates. Each release has an attribute that defines the name of a
```
pypingou: Would it be worth documenting which attribute that is?

pypingou (commented):
One question, otherwise looks fine :)

production.ini
bowlofeggs: @pypingou I'm applying this patch to document that attribute:

```
index b2e57d9e..891bd1f8 100644
--- a/production.ini
+++ b/production.ini
@@ -126,8 +126,9 @@ use = egg:bodhi-server
 # domain for generated message IDs
 # message_id_email_domain = admin.fedoraproject.org

-# Where to search for e-mail templates. Each release has an attribute that defines the name of a
-# mail template, and Bodhi will look in the path defined by this setting to find the named template.
+# Where to search for e-mail templates. Each release has a mail_template attribute that defines the
+# name of a mail template, and Bodhi will look in the path defined by this setting to find the named
+# template.
 # mail.templates_basepath = bodhi:server/email/templates/
```

pypingou: Looks good to me, thanks :)

bowlofeggs:

bowlofeggs:

PR #2932, opened 2019-01-16
Remove Bodhi's stacks feature.

### LINKED ISSUE #2241: Drop Bodhi's "stacks" feature

bowlofeggs:
The stacks feature was never used so this patch removes it. Its
removal revealed a test isolatio problem where a test on the
comments service began to fail due to the removal of the stacks
tests. It turned out that bodhi/tests/server/test_renderers.py had
been making global configuration changes that the stacks tests
were presumably reverting. Thus, this commit also fixes the test
isolation issue.

fixes #2241


pypingou (approved):
Less-code++

bowlofeggs:

bowlofeggs:

PR #2935, opened 2019-01-16
Add type annotations and enforcement for bodhi.server.bugs.

bowlofeggs:
This commit adds Python type annotations to bodhi.server.bugs, and
configured CI to test those annotations with mypy.

mypy complained about the shared superclass BugTracker not having
the same method signatures on the child class implementations.
Since there is no reason to have a shared superclass in this file,
I opted to remove the superclass rather than try to solve the
signature discrepancy.

I also had to configure mypy to ignore missing imports for a few
external libraries, since they don't provide stub files for type
annotations.

Lastly, I configured mypy to be silent about errors found while
following imports for now, since there are quite a few problems
to fix before that could pass. We're starting off small here.


bodhi/server/bugs.py
```diff
@@ -103,46 +92,46 @@ def _connect(self):
...
         Args:
-            bug_id (basestring or int): The id of the bug you want a URL for.
+            bug_id: The id of the bug you want a URL for.
         Returns:
             basestring: The requested URL.
         """
         return "%s/show_bug.cgi?id=%s" % (config['bz_baseurl'], bug_id)
 
-    def getbug(self, bug_id):
+    def getbug(self, bug_id: int) -> typing.Any:
```
pypingou, line 117: the bz method above returns a ``bugzilla.Bugzilla``, this one a ``bugzilla.bug.Bug`` according to the doc, any reason to specify ``bugzilla.bug.Bug`` here?

pypingou: Below we're specifying a type using the string (cf the ``models.Bug`` below), can it also be used in the returned type?

pypingou (commented):
Couple of questions, I'd +1 but I'm curious about the answers first :)

bowlofeggs:
On Thu, 2019-01-17 at 00:58 -0800, Pierre-Yves Chibon wrote:
> the bz method above returns a bugzilla.Bugzilla, this one a
> bugzilla.bug.Bug according to the doc, any reason to specify
> bugzilla.bug.Bug here?

The problem is that python-bugzilla tries to enforce the public API,
which does not include the Bug class, so unfortunately it is not
possible to import it to list it as the return type. I wish they had
not done that, because Python does not have a formal concept of
public/private APIs, but due to that there is not a way to reference
the object.

bowlofeggs:
On Thu, 2019-01-17 at 00:58 -0800, Pierre-Yves Chibon wrote:
> Below we're specifying a type using the string (cf the models.Bug
> below), can it also be used in the returned type?

No - for mypy you have to import the object (even when referencing it
as a string) and since I cannot do that there is no way to use it as a
return type (see my other comment).

pypingou (approved):
Thank you for your explanations :)

PR #2940, opened 2019-01-17
Trim all database migrations prior to the head of 3.13.0.

bowlofeggs:
Rather than maintaining database migrations in perpituity, I opt
to trim them when making major Bodhi releases. This commit removes
database migrations prior to 4.0.0, with the exception of keeping
the head migration from 3.13.0 for alemic's history.


bowlofeggs:
We should wait to merge this pull request until 3.13.0 is deployed to production so that the Vagrant database download will be able to migrate from the production data to the head of ```develop```.

bowlofeggs:
quay.io seems to be down right now, which is why the integration tests failed.

bowlofeggs:
The integration test failures are due to quay.io being down.

Zlopez (approved):
I only see one typo in commit message - "alemic's" instead of "alembic's". Otherwise looks good.

PR #2942, opened 2019-01-17
Introduce a new warm_cache_on_start setting.

bowlofeggs:
If warm_cache_on_start is True (the default), Bodhi will build two
caches when the server process starts as it has done for many
months. If False, it will not. The goal of this patch is to allow
the development environment to turn this caching off because it
takes a long time (about 33 seconds on my laptop) and when
developing Bodhi it is typical to restart the server repeatedly.


pypingou (approved):

bowlofeggs:

bowlofeggs:

PR #2943, opened 2019-01-17
Use bodhi-shell command instead of bshell in vagrant developer docs

sebwoj:


bowlofeggs (approved):
Thanks!

PR #2945, opened 2019-01-17
Set default sqlalchemy.url setting to PostgreSQL

### LINKED ISSUE #2040: The sqlalchemy.url setting should default to PostgreSQL
Since Bodhi only works with PostgreSQL, we should default the ```sqlalchemy.url``` setting to it. Here's what might be a same default:

```
postgresql://localhost/bodhi
```

Of course, any sane sysadmin should adjust this setting anyway, but it would be better if we defaulted to postgres instead of sqlite.

sebwoj:
Fixes #2040

### DIFF
--- a/alembic.ini
+++ b/alembic.ini
@@ -12,7 +12,7 @@ script_location = bodhi:server/migrations
 # revision_environment = false
 
 # Don't bother, this is obtained from the Bodhi config file
-sqlalchemy.url = sqlite://bodhi.db
+sqlalchemy.url = postgresql://localhost/bodhi
 
 # Set to true to acquire the global DDL lock for BDR for offline SQL scripts.
 # BDR is automatically detected during online migrations and this setting is
--- a/bodhi/server/config.py
+++ b/bodhi/server/config.py
@@ -585,7 +585,7 @@ class BodhiConfig(dict):
             'value': None,
             'validator': _validate_none_or(str)},
         'sqlalchemy.url': {
-            'value': 'sqlite:////var/cache/bodhi.db',
+            'value': 'postgresql://localhost/bodhi',
             'validator': str},
         'stable_bug_msg': {
             'value': ('%s has been pushed to the %s repository. If problems still persist, please '
--- a/devel/ansible/roles/bodhi/tasks/main.yml
+++ b/devel/ansible/roles/bodhi/tasks/main.yml
@@ -147,12 +147,6 @@
   args:
       creates: /home/vagrant/bodhi/development.ini
 
-- name: Switch the database connection to postgres
-  replace:
-    dest: /home/vagrant/bodhi/development.ini
-    regexp: "^sqlalchemy.url = sqlite.*$"
-    replace: "sqlalchemy.url = postgresql://postgres:anypasswordworkslocally@localhost/bodhi2"
-
 # Bug in zope.interface filled here https://github.com/zopefoundation/zope.interface/issues/114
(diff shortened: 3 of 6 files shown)


bowlofeggs (changes requested):
It looks like some tests are failing.

devel/ci/Dockerfile-f28
```diff
@@ -43,6 +43,7 @@ RUN dnf install --disablerepo rawhide-modular -y \
     python3-munch \
     python3-openid \
     python3-pillow \
+    python3-psycopg2 \
```
bowlofeggs: Was this needed to get the tests to pass?

If so, we should probably mock the setting in the unit tests so they they keep using sqlite, otherwise we won't be able to run the unit tests downstream very easily. Also, I expect that the unit tests are probably slower when run with PG.

What do you think?

sebwoj: As far as I remember I added `python3-psycopg2` because of failing integration tests.
I thing that unit tests are still using sqllite:
https://github.com/fedora-infra/bodhi/blob/27ea7fb92c37beb45da66f5a7b0b50d5e8a0e75d/bodhi/tests/server/base.py#L38

I'm think that I changed DB engine to PG in development.ini. The difference between vagrant and Integration tests are DB engine. Ansible changes DB engine to PG in vagrant, but in integration test development.ini is not changed. This could be the root cause of IT fail.
Do you think that we should leave sqllite in development.ini as it is now?

I will do some benchmark between sqllite and PG during the weekend.

bowlofeggs:
I'm going to rebase this, and add a teensy patch:

```
diff --git a/devel/ci/Dockerfile-pip b/devel/ci/Dockerfile-pip
index dd0adf1c..b11bc770 100644
--- a/devel/ci/Dockerfile-pip
+++ b/devel/ci/Dockerfile-pip
@@ -30,7 +30,6 @@ RUN dnf install -y \
 COPY requirements.txt /bodhi/requirements.txt

 RUN pip-3 install -r /bodhi/requirements.txt
-RUN pip-3 install psycopg2
 RUN pip-2 install \
     click \
     diff-cover \
@@ -44,6 +43,7 @@ RUN pip-3 install \
     diff-cover \
     flake8 \
     mock \
+    psycopg2 \
     responses \
     pydocstyle \
     pytest \
```

devel/ci/Dockerfile-f28
bowlofeggs: I realized that there was a patch that does this already on develop. I'd forgotten that the tests started failing on develop a while ago. My guess is that you based your patch on a commit that needed psycopg2, and therefore that it's actually not related to the change you are making, which is a coincidence.

After rebasing your patch, I see that these changes aren't even part of it anymore since they are already in develop.

bowlofeggs:
Actually, one more patch too, because I realized that the pip tests install the requires file, which has this package already:

```
diff --git a/devel/ci/Dockerfile-pip b/devel/ci/Dockerfile-pip
index b11bc770..0c6d6136 100644
--- a/devel/ci/Dockerfile-pip
+++ b/devel/ci/Dockerfile-pip
@@ -43,7 +43,6 @@ RUN pip-3 install \
     diff-cover \
     flake8 \
     mock \
-    psycopg2 \
     responses \
     pydocstyle \
     pytest \
```

bowlofeggs (approved):
Thanks!

bowlofeggs:

bowlofeggs:

PR #2946, opened 2019-01-17
Drop assert statement in server.services.updates.new_update

### LINKED ISSUE #2848: bodhi.server.services.updates.new_update() has an assert statement
There is an [```assert``` statement](https://github.com/fedora-infra/bodhi/blob/3.11.3/bodhi/server/services/updates.py#L507) in ```bodhi.server.services.updates.new_update()```. This assert looks like it might be important for correct operation, but the problem is that assert statements get stripped out of optimized Python builds (so the statement won't be present in production).

sebwoj:
Fixes #2848

### DIFF
--- a/bodhi/server/services/updates.py
+++ b/bodhi/server/services/updates.py
@@ -485,7 +485,6 @@ def new_update(request):
 
             log.info('Editing update: %s' % data['edited'])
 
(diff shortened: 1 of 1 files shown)


bowlofeggs (approved):
Thanks!

bowlofeggs:

bowlofeggs:

PR #2953, opened 2019-01-20
Block creating anonymous comments

### LINKED ISSUE #2700: We should drop anonymous comments
SSIA

sebwoj:
Fixes #2700

PR is not finished.
1. I'm not sure what to do with 'anonymous' field in models.Comment.
When system creates anonymous comment, new comment is related to existing DB user
with name "anonymous"(id 207). So it turns out that info about 'anonymous' is somehow redundant
('anonymous' field is True, and user is 'anonymous'). But when we call notifications.publish,
we are using 'anonymous' to create json from comment:
https://github.com/fedora-infra/bodhi/blob/b64d82ce0fff8284164348c6f1d280191f908e7f/bodhi/server/models.py#L3036
How it will affect external systems?

2. Should we force user to log in before comment(bodhi-client)?
For now comment form is just disabled for unauthenticated users.

BTW. unauthenticated user have to provide email address.
Next, provided email addr is droped in service.comment.


bodhi/server/captcha.py
```diff
@@ -1,281 +0,0 @@
...
-
-import base64
-import math
-import random
-import binascii
-
-from PIL import Image, ImageDraw, ImageFont
-from pyramid.httpexceptions import HTTPGone, HTTPNotFound
-from pyramid.view import view_config
-import cryptography.fernet
```
bowlofeggs, line 31: We should remove PIL and cryptography from the dependencies in ```requirements.txt```, and we should also add a note to the release notes about the dropped dependencies.

bodhi/server/config.py
```diff
@@ -336,33 +336,6 @@ class BodhiConfig(dict):
...
-            'validator': int},
-        'captcha.image_width': {
-            'value': 300,
-            'validator': int},
-        'captcha.padding': {
-            'value': 5,
-            'validator': int},
-        'captcha.secret': {
-            'value': None,
-            'validator': _validate_none_or(_validate_fernet_key)},
```
bowlofeggs, line 362: I think we can also drop the ```_validate_fernet_key()``` function.

bodhi/server/config.py
```diff
@@ -336,33 +336,6 @@ class BodhiConfig(dict):
...
             'value': None,
             'validator': _validate_none_or(validate_path)},
-        'captcha.background_color': {
-            'value': '#ffffff',
-            'validator': _validate_color},
-        'captcha.font_color': {
-            'value': '#000000',
-            'validator': _validate_color},
-        'captcha.font_path': {
-            'value': '/usr/share/fonts/liberation/LiberationMono-Regular.ttf',
```
bowlofeggs, line 346: I think we have the Ansible project installing this, and if so we can drop that too.

bodhi/server/models.py
```diff
@@ -2965,12 +2964,10 @@ def comment(self, session, text, karma=0, author=None, anonymous=False,
                 caveats.append({'name': 'karma', 'description': notice})
 
         comment = Comment(
-            text=text, anonymous=anonymous,
+            text=text, anonymous=False,
```
bowlofeggs: To answer one of your questions, I think we can drop this field from the class. We do want to keep the historical data about which comments were anonymous, but as you pointed out that data is also present in the username which I think is good enough.

bodhi/server/models.py
```diff
@@ -1817,7 +1817,7 @@ def _composite_karma(self):
         negative_karma = 0
         users_counted = set()
         for comment in self.comments_since_karma_reset:
-            if comment.karma and not comment.anonymous and comment.user.name not in users_counted:
+            if comment.karma and comment.user.name not in users_counted:
```
bowlofeggs, line 1798: We are going to need to take some action to ensure that historical anonymous karma does not get counted as karma once this patch is deployed. One way I can think to do that is to make a data migration that finds all anonymous comments, sets their karma field to 0, and modifies the comment text to indicate whether they had put a +1 or -1.

What do you think?

bowlofeggs (changes requested):

bowlofeggs:
>     1. I'm not sure what to do with 'anonymous' field in models.Comment.
>        When system creates anonymous comment, new comment is related to existing DB user
>        with name "anonymous"(id 207). So it turns out that info about 'anonymous' is somehow redundant
>        ('anonymous' field is True, and user is 'anonymous'). But when we call notifications.publish,
>        we are using 'anonymous' to create json from comment:

>        How it will affect external systems?

I answered a bit of this in a line comment. I think we should drop the anonymous column on that table. It will change the message schema, so we should also document that change in our release notes. This is OK since we are making a major version release for it. Lastly, the other thing we should do is make sure that historical anonymous comments don't get their karma counted.

>     2. Should we force user to log in before comment(bodhi-client)?
>        For now comment form is just disabled for unauthenticated users.

Yeah - I think we can hide the comment form if the user isn't logged in. Or maybe for bonus points have a "log in to comment" message if you think that's a good thing to do.

> BTW. unauthenticated user have to provide email address.
> Next, provided email addr is droped in service.comment.

Hahah yeah I noticed that about a year ago. It raised both of my eyebrows.

bodhi/server/models.py
sebwoj: I didn't realize that in the past(before 3.12, so 17 Dec 2018) anno users where allowed to leave karma(and critpathkarma, but I assume this field will be removed in 4.0). So removing karma from anno comments is mandatory hiere. But I'm not sure about leaving +1/-1 note about karma in text. For instance: [FEDORA-EPEL-2018-65bdd9ec8c](https://bodhi.fedoraproject.org/updates/FEDORA-EPEL-2018-65bdd9ec8c) shows that karma is +2(right panel), but actualy there are a few more votes for this update, and user can be suprised that there is a special algorithm for counting karma in backend. Maybe 4.0 will be a good opportunity to prepare some clarification in this topic, something like "Anonymous user karma:1 (anonymous votes are not take in to account when counting overall update karma)" or we can do some support for legacy "anno" comments leaving this kind of note below all anno comments in frontend(and maybe in CLI).

sebwoj:
> Yeah - I think we can hide the comment form if the user isn't logged in. Or maybe for bonus points have a "log in to comment" message if you think that's a good thing to do.

Good idea. I forgot about that while removing  anno comment form.

bowlofeggs:
@sebwoj do you still want to work on this?

sebwoj:
Yes :)
Could you send me your opinion about:
https://github.com/fedora-infra/bodhi/pull/2953#discussion_r255338317

bodhi/server/models.py
bowlofeggs: Since we are removing the feature to add anonymous comments, I don't think we need to worry too much about how we store the historical information. I think it'd be OK to just modify the comment, and add ```"\n\nkarma: +1"``` to it to record what the karma had been.

Alternatively, we could treat the "anonymous" username as special since it historically was used to do this, and modify our karma counting algorithm to ignore karma from that particular username, instead of looking for the anonymous=True boolean.

Alternatively, we could just keep the boolean for now, and solve this problem later. We would just remove the ability for new updates to get the boolean set to True.

I'm OK with any of the above solutions.

I think our users are probably mostly aware that anonymous karma isn't counted. In any case, since we are removing anonymous comments, I think it's OK if we don't try to make it clear that anonymous comments are ignored for karma purposes, since only historic data will have them anymore.

bodhi/client/bindings.py
```diff
@@ -880,10 +877,8 @@ def update_str(self, update, minimal=False):
         if len(update['comments']):
             comments_lines = []
             for comment in update['comments']:
-                anonymous = " (unauthenticated)" if comment['anonymous'] else ''
-                comments_lines.append('{0}{1} - {2} (karma {3})'.format(
-                    comment['user']['name'], anonymous,
-                    comment['timestamp'], comment['karma']))
+                comments_lines.append('{0} - {1} (karma {2})'.format(
+                    comment['user']['name'], comment['timestamp'], comment['karma']))
```
bowlofeggs, line 881: We could use f-strings here too, but this is also fine.

bodhi/server/migrations/versions/c98beb4940b5_remove_the_comments_anonymous_column.py
```diff
@@ -0,0 +1,65 @@
...
+    )
+    # Remove information about comment karma from comment's text
+    op.execute(
+        "UPDATE comments SET text=REPLACE(text, '\n\nkarma +1', ''), "
+        "karma = 1 where anonymous = true and text like '%\n\nkarma +1'"
+    )
+    op.execute(
+        "UPDATE comments SET text=REPLACE(text, '\n\nkarma -1', ''), "
+        "karma = -1 where anonymous = true and text like '%\n\nkarma -1'"
+    )
```
bowlofeggs, line 65: Whoah that's some fancy querying there ☺

docs/user/release_notes.rst
```diff
@@ -34,6 +34,8 @@ Backwards incompatible changes
 * The ``bodhi-monitor-composes`` script has been removed (:issue:`2171`).
 * The stacks feature has been removed (:issue:`2241`).
 * The ``bodhi-manage-releases`` script has been removed (:issue:`2420`).
+* Support for anonymous comments was dropped. As a result, the ``anonymous`` field on the Comment
+  object was removed and comments query API parameter ``anonymous`` was droped (:issue:`2700`).
```
bowlofeggs: We should also add a note about pillow, liberation-mono-fonts, and cryptography not being required dependencies anymore.

bowlofeggs (changes requested):
Very nice!

I will add the suggested release notes change myself and merge. Thanks so much!

bowlofeggs:
I'm applying this patch:

```
diff --git a/docs/user/release_notes.rst b/docs/user/release_notes.rst
index 443f39c7..68edabd1 100644
--- a/docs/user/release_notes.rst
+++ b/docs/user/release_notes.rst
@@ -35,7 +35,8 @@ Backwards incompatible changes
 * The stacks feature has been removed (:issue:`2241`).
 * The ``bodhi-manage-releases`` script has been removed (:issue:`2420`).
 * Support for anonymous comments was dropped. As a result, the ``anonymous`` field on the Comment
-  object was removed and comments query API parameter ``anonymous`` was droped (:issue:`2700`).
+  object was removed and comments query API parameter ``anonymous`` was droped. All ``captcha.*``
+  settings were removed (:issue:`2700`).
 * Bodhi client and server no longer support Python 2. Python 3.6+ are the only supported Python
   releases (:issue:`2759`).
 * Support for the ``ci_url`` on the ``Build`` object was dropped (:issue:`2782`).
@@ -56,6 +57,11 @@ Dependency changes
 ^^^^^^^^^^^^^^^^^^
 
 * pkgdb is no longer required (:issue:`1970`).
+* cryptography is no longer required (:issue:`2700`).
+* Fonts are no longer required for the captcha (Bodhi previously defaulted to using
+  liberation-mono-fonts, but this wasn't a strict requirement since the font was configurable)
+  (:issue:`2700`).
+* pillow is no longer required (:issue:`2700`).
 * six is no longer required for the client or server (:issue:`2759`).
```

bowlofeggs:
I'm adding one more small change, to make the karma comment use a colon, like "karma: -1":

```
diff --git a/bodhi/server/migrations/versions/c98beb4940b5_remove_the_comments_anonymous_column.py b/bodhi/server/migrations/versions/c98beb4940b5_remove_the_comments_anonymous_column.py
index 10f6ec47..fe88126a 100644
--- a/bodhi/server/migrations/versions/c98beb4940b5_remove_the_comments_anonymous_column.py
+++ b/bodhi/server/migrations/versions/c98beb4940b5_remove_the_comments_anonymous_column.py
@@ -35,11 +35,11 @@ def upgrade():
     """Remove the comment's anonymous column and clear anonymous comments karma."""
     # Insert information about comment karma in comment's text
     op.execute(
-        "UPDATE comments SET text=text || '\n\nkarma +1', "
+        "UPDATE comments SET text=text || '\n\nkarma: +1', "
         "karma = 0 where anonymous = true and karma = 1"
     )
     op.execute(
-        "UPDATE comments SET text=text || '\n\nkarma -1', "
+        "UPDATE comments SET text=text || '\n\nkarma: -1', "
         "karma = 0 where anonymous = true and karma = -1"
     )
     op.drop_column('comments', 'anonymous')
@@ -56,10 +56,10 @@ def downgrade():
     )
     # Remove information about comment karma from comment's text
     op.execute(
-        "UPDATE comments SET text=REPLACE(text, '\n\nkarma +1', ''), "
-        "karma = 1 where anonymous = true and text like '%\n\nkarma +1'"
+        "UPDATE comments SET text=REPLACE(text, '\n\nkarma: +1', ''), "
+        "karma = 1 where anonymous = true and text like '%\n\nkarma: +1'"
     )
     op.execute(
-        "UPDATE comments SET text=REPLACE(text, '\n\nkarma -1', ''), "
-        "karma = -1 where anonymous = true and text like '%\n\nkarma -1'"
+        "UPDATE comments SET text=REPLACE(text, '\n\nkarma: -1', ''), "
+        "karma = -1 where anonymous = true and text like '%\n\nkarma: -1'"
     )
```

bowlofeggs (approved):
Thanks!

bowlofeggs:
The integration tests seem to have all failed due to being killed. They were likely killed by the groovyscript, which has a maximum time per task. I will rebase and run the tests locally and merge if they pass.

bodhi/server/migrations/versions/c98beb4940b5_remove_the_comments_anonymous_column.py
sebwoj: Thanks!

bowlofeggs:

bowlofeggs:

PR #2954, opened 2019-01-20
Remove_unused_UpdateStatus.processing

### LINKED ISSUE #1999: UpdateStatus.processing seems to be unused
While documenting ```models.py``` I noticed that ```UpdateStatus``` has a [```processing```](https://github.com/fedora-infra/bodhi/blob/3.0.0/bodhi/server/models.py#L377) state, which does not seem to be used. We should remove it.

sebwoj:
Fixes #1999

If this patch will be included to bodhi 4.0 rel, I will add note in release notes.

### DIFF
--- a/bodhi/client/__init__.py
+++ b/bodhi/client/__init__.py
@@ -539,7 +539,7 @@ def edit(user, password, url, debug, openid_api, **kwargs):
               help='Updates that have been submitted before a certain time')
 @click.option('--status', help='Filter by update status',
               type=click.Choice(['pending', 'testing', 'stable', 'obsolete',
-                                 'unpushed', 'processing']))
+                                 'unpushed']))
 @click.option('--suggest', help='Filter by post-update user suggestion',
               type=click.Choice(['logout', 'reboot']))
 @click.option('--type', default=None, help='Filter by update type',
--- a/bodhi/client/bindings.py
+++ b/bodhi/client/bindings.py
@@ -337,7 +337,7 @@ class BodhiClient(OpenIdBaseClient):
             releases (basestring): A comma separated list of releases that you wish to query updates
                 for.
             status (basestring): The update status (``pending``, ``testing``, ``stable``,
-                ``obsolete``, ``unpushed``, ``processing``)
+                ``obsolete``, ``unpushed``)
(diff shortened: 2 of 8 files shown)


bowlofeggs (changes requested):
Nice, looks good. Let's add a release note since the CLI change is technically backwards incompatible.

bowlofeggs (approved):
Thanks so much!

bowlofeggs:
BTW, I used the migration pattern you wrote here when I did #2994 ☺

sebwoj:
Found on the net, tunned a llitle bit, and tested with "psql" command line after apply :D

sebwoj:
It looks like alembic head (down revision) is improper. Is it ok to push changes after approval?

bowlofeggs:
Ah, the failures are very likely due to me having merged other migrations - we just need to fix up the down revision. I'll fix that for you.

bowlofeggs:
Heh, I see that you had the same thought - I'll fix it for you, no worries!

bowlofeggs:
I'm applying this patch to your branch:

```
diff --git a/bodhi/server/migrations/versions/6b3eb9ae2b87_remove_unused_updatestatus_processing.py b/bodhi/server/migrations/versions/6b3eb9ae2b87_remove_unused_updatestatus_processing.py
index f615aeac..df37263f 100644
--- a/bodhi/server/migrations/versions/6b3eb9ae2b87_remove_unused_updatestatus_processing.py
+++ b/bodhi/server/migrations/versions/6b3eb9ae2b87_remove_unused_updatestatus_processing.py
@@ -19,7 +19,7 @@
 Remove unused UpdateStatus.processing.
 
 Revision ID: 6b3eb9ae2b87
-Revises: 5c86a3f9dc03
+Revises: aae0d29d49b7
 Create Date: 2019-02-09 12:17:18.444588
 """
 from alembic import op
@@ -28,7 +28,7 @@ from sqlalchemy import exc
 
 revision = '6b3eb9ae2b87'
-down_revision = '5c86a3f9dc03'
+down_revision = 'aae0d29d49b7'
 
 
 def upgrade():
```

bowlofeggs:

bowlofeggs:

Issue #2838, opened 2018-12-11
[M] Drop support for fedmsg


bowlofeggs:
Now that Bodhi supports fedora-messaging, we should drop support for fedmsg.

CC @jeremycline @abompard

AdamWill:
openQA relies on the Bodhi fedmsgs to trigger update tests. I'm happy to migrate to fedora-messaging, but I was sort of waiting for some sort of signal or flag day that we were going to start migrating *everything* in infra - for instance, I can't migrate the *compose* test trigger unless Pungi has also started emitting fedora-messaging messages (I don't know if it has), and so on...

bowlofeggs:
@jeremycline can correct me if I am wrong, but as I understand it, there is a bridge that will translate fedora-messaging messages to fedmsg and vice-versa. If I am correct about that, you should still receive fedmsgs, even if Bodhi isn't sending them.

jeremycline:
Indeed, there is a bridge from AMQP to fedmsg and fedmsg to AMQP (and they're currently running in production) so there will be no flag day. You can migrate at your leisure.

bowlofeggs:
Fantastic.

On Wed, 2018-12-12 at 13:36 -0800, Jeremy Cline wrote:
> You can migrate at your leisure.

Did you pronounce that in your head like "leezure" or "lehzure"?

bowlofeggs:
@abompard @jeremycline is there anything like fedmsg-hub for fedora messaging? I've got a few message consumers in Bodhi. Most of them would actually be better to replace with Celery, but there is one listener that looks for messages from robosignatory so Bodhi knows when builds are signed.

If there isn't a fedmsg-hub equivalent, I guess I could just make my own process that listens on the robosignatory topic.

abompard:
@bowlofeggs You can either write a small script to call the listening API, or use the `fedora-messaging` command line and point it at a python callback. Docs and examples are here: https://fedora-messaging.readthedocs.io/en/stable/tutorial/usage.html#listening

jeremycline:
Yes, the command is ``fedora-messaging consume`` (https://fedora-messaging.readthedocs.io/en/stable/fedora-messaging.html), although it looks like https://fedora-messaging.readthedocs.io/en/stable/consuming.html isn't linked to it which it probably should

bowlofeggs:
This change is planned for inclusion in the upcoming 4.0.0 release: #3221

bowlofeggs:
[...]

Issue #130, opened 2015-03-04
Allow provenpackagers to add additional builds to an existing update


lmacken:
```
14:48  sgallagh> lmacken: RFE for bodhi: if an update for Package X exists, and I as a provenpackager want to add another package to that update (because they need to go stable together), I should be either able to edit the existing 
                 update or create a new one that consumes the existing one. Neither of these is currently the case :(
```
/cc @sgallagher

PR #2957, opened 2019-01-23
Configure the consumers to use fedora_messaging.

bowlofeggs:
This is needed as part of #2838.

### DIFF
--- a/bodhi/server/consumers/__init__.py
+++ b/bodhi/server/consumers/__init__.py
@@ -1,4 +1,4 @@
-# Copyright 2017 Red Hat, Inc.
+# Copyright 2019 Red Hat, Inc.
 #
 # This file is part of Bodhi.
 #
@@ -15,4 +15,48 @@
 # You should have received a copy of the GNU General Public License
 # along with this program; if not, write to the Free Software
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
-"""Contains Bodhi's fedmsg consumers."""
+"""
+fedora-messaging consumer.
+
+This module is responsible for consuming the messaging from the fedora-messaging bus.
+It has the role to inspect the topics of the message and call the correct handler.
+"""
+import logging
+
+import fedora_messaging
+
+from bodhi.server.consumers.composer import ComposerHandler
+from bodhi.server.consumers.signed import SignedHandler
+from bodhi.server.consumers.updates import UpdatesHandler
+
+
+log = logging.getLogger('bodhi')
+
+
+def messaging_callback(msg: fedora_messaging.api.Message):  # noqa: D401
+    """
+    Callback method called by fedora-messaging consume.
+
+    Redirect messages to the correct handler using the
+    message topic.
+
+    Args:
+        msg: The message received from the broker.
+    """
+    log.info(f'Received message from fedora-messaging with topic: {msg.topic}')
+
+    if msg.topic.endswith('.bodhi.composer.start'):
+        handler = ComposerHandler()
+        log.debug('Passing message to the Composer handler')
+        handler(msg)
+
+    if msg.topic.endswith('.buildsys.tag'):
+        handler = SignedHandler()
+        log.debug('Passing message to the Signed handler')
+        handler(msg)
+
+    if msg.topic.endswith('.bodhi.update.request.testing') \
+       or msg.topic.endswith('.bodhi.update.edit'):
+        handler = UpdatesHandler()
+        log.debug('Passing message to the Updates handler')
+        handler(msg)
--- a/bodhi/server/consumers/composer.py
+++ b/bodhi/server/consumers/composer.py
@@ -19,28 +19,31 @@
 The Bodhi "Composer".
 
 This module is responsible for the process of "pushing" updates out. It's
-comprised of a fedmsg consumer that launches threads for each repository being
+comprised of a fedore messaging consumer that launches threads for each repository being
 composed.
 """
 
 import functools
 import hashlib
 import json
+import logging
 import os
 import shutil
 import subprocess
 import tempfile
 import threading
 import time
+import typing
 from datetime import datetime
 from http.client import IncompleteRead
 from urllib.error import HTTPError, URLError
 from urllib.request import urlopen
 
-import fedmsg.consumers
 import jinja2
+import fedora_messaging
 
-from bodhi.server import bugs, initialize_db, log, buildsys, notifications, mail
+
+from bodhi.server import bugs, initialize_db, buildsys, notifications, mail
 from bodhi.server.config import config, validate_path
 from bodhi.server.exceptions import BodhiException
 from bodhi.server.metadata import UpdateInfoMetadata
@@ -49,6 +52,9 @@ from bodhi.server.models import (Compose, ComposeState, Update, UpdateRequest, U
 from bodhi.server.scripts import clean_old_composes
 from bodhi.server.util import (copy_container, sorted_updates, sanity_check_repodata,
                                transactional_session_maker)
+
+
+log = logging.getLogger('bodhi')
 
 
 def checkpoint(method):
@@ -80,7 +86,7 @@ def checkpoint(method):
     return wrapper
 
 
-class Composer(fedmsg.consumers.FedmsgConsumer):
+class ComposerHandler(object):
     """
     The Bodhi Composer.
 
@@ -128,16 +134,13 @@ class Composer(fedmsg.consumers.FedmsgConsumer):
         - see if any updates now meet the stable criteria, and set the request
     """
 
-    config_key = 'composer'
-
-    def __init__(self, hub, db_factory=None, compose_dir=config.get('compose_dir'),
-                 *args, **kw):
+    def __init__(
+            self, db_factory: typing.Union[transactional_session_maker, None] = None,
+            compose_dir: str = config.get('compose_dir')):
         """
         Initialize the Composer.
 
         Args:
-            hub (moksha.hub.hub.CentralMokshaHub): The hub this handler is consuming messages from.
-                It is used to look up the hub config values.
             db_factory (bodhi.server.util.TransactionalSessionMaker or None): If given, used as the
                 db_factory for this Composer. If None (the default), a new TransactionalSessionMaker
                 is created and used.
@@ -154,13 +157,7 @@ class Composer(fedmsg.consumers.FedmsgConsumer):
         buildsys.setup_buildsystem(config)
         bugs.set_bugtracker()
         self.compose_dir = compose_dir
-        prefix = hub.config.get('topic_prefix')
-        env = hub.config.get('environment')
-        self.topic = prefix + '.' + env + '.' + hub.config.get('composer_topic')
-        self.valid_signer = hub.config.get('releng_fedmsg_certname')
-        if not self.valid_signer:
-            log.warning('No releng_fedmsg_certname defined'
-                        'Cert validation disabled')
+
         self.max_composes_sem = threading.BoundedSemaphore(config.get('max_concurrent_composes'))
 
         # This will ensure that the configured paths exist, and will raise ValueError if any does
@@ -171,28 +168,53 @@ class Composer(fedmsg.consumers.FedmsgConsumer):
             except ValueError as e:
                 raise ValueError('{} Check the {} setting.'.format(str(e), setting))
 
-        super(Composer, self).__init__(hub, *args, **kw)
-        log.info('Bodhi composer listening on topic: %s' % self.topic)
-
-    def consume(self, msg):
+    def __call__(self, message: fedora_messaging.api.Message):
         """
-        Receive a fedmsg and call work() with it.
+        Begin the push process.
+
+        Here we organize & prioritize the updates, and fire off separate
+        threads for each repo tag being composed.
+
+        If there are any security updates in the push, then those repositories
+        will be executed before all others.
 
         Args:
-            msg (munch.Munch): The fedmsg that was received.
+            message: The message we are processing. This is how we know what compose jobs to run.
         """
-        self.log.info(msg)
-        if self.valid_signer:
-            if not fedmsg.crypto.validate_signed_by(msg['body'], self.valid_signer,
-                                                    **self.hub.config):
-                self.log.error('Received message with invalid signature!'
-                               'Ignoring.')
-                # TODO: send email notifications
-                return
+        message = message.body["msg"]
+        resume = message.get('resume', False)
+        agent = message.get('agent')
+        notifications.publish(topic="composer.start", msg=dict(agent=agent), force=True)
 
-        self.work(msg)
+        results = []
+        threads = []
+        for compose in self._get_composes(message):
+            log.info('Now starting composes')
 
-    def _get_composes(self, msg):
+            composer = get_composer(ContentType.from_string(compose['content_type']))
+            if not composer:
+                log.error(
+                    'Unsupported content type %s submitted for composing. SKIPPING',
+                    compose['content_type']
+                )
+                continue
+
+            thread = composer(self.max_composes_sem, compose, agent, self.db_factory,
+                              self.compose_dir, resume)
+            threads.append(thread)
+            thread.start()
+
+        log.info('All of the batches are running. Now waiting for the final results')
+        for thread in threads:
+            thread.join()
+            for result in thread.results():
+                results.append(result)
+
+        log.info('Push complete!  Summary follows:')
+        for result in results:
+            log.info(result)
+
+    def _get_composes(self, msg: dict):
         """
         Return a list of dictionaries that represent the :class:`Composes <Compose>` we should run.
 
@@ -204,7 +226,7 @@ class Composer(fedmsg.consumers.FedmsgConsumer):
         message.
 
         Args:
-            msg (munch.Munch): The body of the received fedmsg.
+            msg: The body of the received message.
         Returns:
             list: A list of dictionaries, as returned from :meth:`Compose.__json__`.
         """
@@ -219,46 +241,6 @@ class Composer(fedmsg.consumers.FedmsgConsumer):
                 c.state = ComposeState.pending
 
             return [c.__json__(composer=True) for c in composes]
-
-    def work(self, msg):
-        """Begin the push process.
-
-        Here we organize & prioritize the updates, and fire off separate
-        threads for each repo tag being composed.
-
-        If there are any security updates in the push, then those repositories
-        will be executed before all others.
-        """
-        body = msg['body']['msg']
-        resume = body.get('resume', False)
-        agent = body.get('agent')
-        notifications.publish(topic="compose.start", msg=dict(agent=agent), force=True)
-
-        results = []
-        threads = []
-        for compose in self._get_composes(body):
-            self.log.info('Now starting composes')
-
-            composer = get_composer(ContentType.from_string(compose['content_type']))
-            if not composer:
-                self.log.error('Unsupported content type %s submitted for composing. SKIPPING',
-                               compose['content_type'])
-                continue
-
-            thread = composer(self.max_composes_sem, compose, agent, self.log, self.db_factory,
-                              self.compose_dir, resume)
-            threads.append(thread)
-            thread.start()
-
-        self.log.info('All of the batches are running. Now waiting for the final results')
-        for thread in threads:
-            thread.join()
-            for result in thread.results():
-                results.append(result)
-
-        self.log.info('Push complete!  Summary follows:')
-        for result in results:
-            self.log.info(result)
 
 
 def get_composer(content_type):
@@ -283,7 +265,7 @@ class ComposerThread(threading.Thread):
 
     ctype = None
 
-    def __init__(self, max_concur_sem, compose, agent, log, db_factory, compose_dir, resume=False):
+    def __init__(self, max_concur_sem, compose, agent, db_factory, compose_dir, resume=False):
         """
         Initialize the ComposerThread.
 
@@ -302,7 +284,6 @@ class ComposerThread(threading.Thread):
         """
         super(ComposerThread, self).__init__()
         self.db_factory = db_factory
-        self.log = log
         self.agent = agent
         self.max_concur_sem = max_concur_sem
         self._compose = compose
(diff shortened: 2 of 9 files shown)


bodhi/tests/server/consumers/test_masher.py
```diff
@@ -2232,11 +2232,123 @@ def test_expire_buildroot_overrides_exception(self, expire, exception_log, publi
...
+
+    @mock.patch.dict('bodhi.server.config.config', {'pungi.cmd': '/does/not/exist'})
+    def test___init___missing_paths(self):
+        """__init__() should raise a ValueError if configured paths do not exist."""
+        for s in ('pungi.cmd', 'mash_dir', 'mash_stage_dir'):
+            with mock.patch.dict('bodhi.server.config.config', {s: '/does/not/exist'}):
+                with self.assertRaises(ValueError) as exc:
+                    Masher(FakeHub(), db_factory=self.db_factory, mash_dir=self.tempdir)
+
+                    self.assertEqual(
```
sebwoj: Lets move this assertion outside "with" statement. Here, it will be skipped by test.

sebwoj (changes requested):

abompard:
My attempt to add Celery to Bodhi will modify the `tasks/masher.py` and `tasks/updates.py` files too. Should I wait for this PR to land or should we do it the other way around? Once Celery is in, using a fedora-messaging listener for masher and updates will be obsolete, since they're only distributed tasks actually. But you're more advanced than I am, so there's a timing question too.

What do you prefer?

bowlofeggs:
On Tue, 2019-02-05 at 10:14 +0000, Aurélien Bompard wrote:
> My attempt to add Celery to Bodhi will modify the tasks/masher.py and
> tasks/updates.py files too. Should I wait for this PR to land or
> should we do it the other way around? Once Celery is in, using a
> fedora-messaging listener for masher and updates will be obsolete,
> since they're only distributed tasks actually. But you're more
> advanced than I am, so there's a timing question too.
> What do you prefer?

Hey Aurélien!

Good thoughts. There is a possible advantage to doing this pull request
first, because we can cherry pick this patch back to the 3.x series.
That would only be beneficial if we want to make more 3.x releases
before 4.0 comes out. I could go either way on that decision. If we
make more 3.x series, we can keep patches flowing through the release
pipeline while we work on 4.0, but the downside is that it will create
more work for us too (patches will be increasingly difficult to
backport as the branches diverge).

So I guess the question is whether we want to keep making 3.x releases
as we go. I have been working on a 3.13 lately. So far it's not too bad
to backport patches, though I do get conflicts on many of them already.
So far they have been easy conflicts.

If we don't want to make 3.x fully work with fedora_messaging, then I
agree it might be more efficient to convert many of these things to be
Celery workers (though some of them do still need to be
fedora_messaging consumers, like the signing handler).

abompard:
Oh, you want to bring fedora-messaging support to the 3.x branch? In that case then yeah let's finish that PR first, and add Celery later.

bowlofeggs:
@abompard well, we *may* want to do that. 3.13.0 does have the ability to send, but not to receive, AMQP. I think it's up to us to decide whether we want to make a 3.14 or just wait for 4.0.

bodhi/server/consumers/masher.py
```diff
@@ -228,45 +250,52 @@ def _get_composes(self, msg):
...
         Here we organize & prioritize the updates, and fire off separate
         threads for each repo tag being mashed.
 
         If there are any security updates in the push, then those repositories
         will be executed before all others.
+
+        Note: This should be folded into __call__() when support for fedmsg is removed. It is
+        separate only because Masher.consume() receives a Munch object and fedora_messaging
+        receives a Message object and so this gives us a nice way to hand a Mapping from both.
+
```
abompard: One thing to remember with fedora-messaging, is that duplicate messages may happen. Therefore I would recommend consuming classes to write explicitely in their documentation how they plan to handle duplicates. It may be as simple as "in case of duplicates, the task will be run again", but for more expensive tasks we could be looking at timestamps, use a locking system (and avoid deadlocks), etc.
It's one thing to keep in mind, and I think each consumers should document how they plan to handle it so we're sure it's not overlooked.

bodhi/server/consumers/signed.py
```diff
@@ -100,10 +103,60 @@ def consume(self, message):
...
+                        u'user': u'sharkcz',
+                        u'version': u'1.3.4',
+                        u'owner': u'sharkcz',
+                        u'release': u'1.fc26'
+                    },
+                },
+            }
+
+        The message can contain additional keys.
+
```
abompard, line 79: Same thing here, we should document how duplicate messages are handled.

abompard:
A couple of other PRs are adding FM consumers so I think we should finish this one to standardize on the proper way to write FM consumers in Bodhi before adding new ones.

abompard:
@bowlofeggs do you need help with that PR?

bowlofeggs:
@abompard You can take it if you really want to - I've just been swamped dealing with Greenwave and other things. I will eventually get to it if I can find some time…

cverna:
@abompard, @bowlofeggs  if you want to give a look at my commit, I think what is left to do is 

- Add fedora-messaging config file and systemd service file that run fedora-messaging consume

- Decide how we handle possible duplicated message ?

- Most likely something else ?

Also I am not sure why the CI does not trigger

cverna:
jenkies test

bodhi/server/consumers/__init__.py
```diff
@@ -15,4 +15,47 @@
...
+    Redirect to messages to the correct handler in function of the
+    message topic.
+    """
+    log.info(
+        'Received message from fedora-messaging with topic: %s', msg.topic
+    )
+
+    if msg.topic.endswith('.bodhi.masher.start'):
+        handler = ComposerHandler()
+        log.info('Passing message to the Masher handler')
```
abompard: These 3 logger calls should probably be debug level, no?

bodhi/server/consumers/masher.py
```diff
@@ -171,28 +156,53 @@ def __init__(self, hub, db_factory=None, mash_dir=config.get('mash_dir'),
...
+    def __call__(self, message: fedora_messaging.api.Message):
         """
-        Receive a fedmsg and call work() with it.
+        Begin the push process.
+
+        Here we organize & prioritize the updates, and fire off separate
+        threads for each repo tag being mashed.
+
+        If there are any security updates in the push, then those repositories
+        will be executed before all others.
```
abompard: We should document how duplicate messages are handled.

bodhi/server/consumers/updates.py
```diff
@@ -66,43 +66,28 @@ class UpdatesHandler(fedmsg.consumers.FedmsgConsumer):
...
             bug_module.set_bugtracker()
 
-        super(UpdatesHandler, self).__init__(hub, *args, **kwargs)
-        log.info('Bodhi updates handler listening on:\n'
-                 '%s' % pprint.pformat(self.topic))
-
-    def consume(self, message):
+    def __call__(self, message: fedora_messaging.api.Message):
         """
         Process the given message, updating relevant bugs and test cases.
```
abompard, line 81: Same here, we should document how duplicate messages are handled.

bodhi/tests/server/consumers/test_consumers.py
```diff
@@ -0,0 +1,45 @@
...
+    """Test class for the messaging_callback function """
+
+    @mock.patch('bodhi.server.consumers.ComposerHandler')
+    def test_messaging_callback_composer(self, handler):
+        msg = Message(
+            topic="org.fedoraproject.prod.bodhi.masher.start",
+            body={}
+        )
+        messaging_callback(msg)
+        handler.assert_called_once_with()
```
abompard: You're only checking that the correct class has been instantiated, not that it has been called with the message as argument. We should probably test that but it's not extremely important since they are all called the same way.

bodhi/server/consumers/masher.py
```diff
@@ -127,24 +133,9 @@ class Masher(fedmsg.consumers.FedmsgConsumer):
...
-
-        Args:
-            hub (moksha.hub.hub.CentralMokshaHub): The hub this handler is consuming messages from.
-                It is used to look up the hub config values.
-            db_factory (bodhi.server.util.TransactionalSessionMaker or None): If given, used as the
-                db_factory for this Masher. If None (the default), a new TransactionalSessionMaker
-                is created and used.
-            mash_dir (basestring): The directory in which to place mashes.
-        Raises:
-            ValueError: If pungi.cmd is set to a path that does not exist.
```
abompard: It may be worthwhile to keep this docstring (minus the hub).

bodhi/server/consumers/signed.py
```diff
@@ -23,50 +23,30 @@
...
-    def __init__(self, hub, *args, **kwargs):
-        """
-        Initialize the SignedHandler, configuring its topic and database.
-
-        Args:
-            hub (moksha.hub.hub.CentralMokshaHub): The hub this handler is consuming messages from.
-                It is used to look up the hub config.
-        """
+    def __init__(self):
+        """Initialize the database."""
```
abompard: This docstring seems wrong.

abompard (changes requested):

abompard:
> Add fedora-messaging config file and systemd service file that run fedora-messaging consume

For the systemd service file, fedora-messaging already provides one as a service template. You just need to drop the configuration file in `/etc/fedora-messaging/bodhi.toml` and enable the `fm-consumer@bodhi.service` service and it should work.

> Decide how we handle possible duplicated message ?

I don't know the inner workings of Bodhi well enough to answer that, I suspect that we want to avoid running two long-running tasks at the same time if we receive duplicate messages, but we may want to do something else. @bowlofeggs do you have an opinion on that?

bowlofeggs:
@cverna There has been a problem for the past few weeks where CentOS CI is running out of GitHub API quota quickly, so a lot of Bodhi pull requests don't get CI lately ☹

I've been running it manually on my own system. You can run it with ```bci all``` in the Vagrant environment (see it's ```--help``` for usage info). Outside of Vagrant, you can run it but it needs sudo: ```sudo devel/ci/bodhi-ci all```.

bowlofeggs:
@abompard @cverna The only long running task Bodhi has is the composer. The ```bodhi-push``` tool generates messages for it. We don't currently do a good job of handling multiple messages received in ```fedmsg```, so I honestly think it's not a regression if we continue to not do a good job of that in this pull request. I.e., I'm happy to merge this while continuing to ignore that problem. The other consumers should be idempotent, so it's OK if they receive a message more than once.

My plan to fix that problem is to use Celery, and have a dedicated worker to handle composes that only has a concurrency of 1. That way we know that only 1 can happen at a time. IMO, that's outside the scope of this PR, just wanted to let you know that I have an idea of how to solve it.

bodhi/server/consumers/__init__.py
```diff
@@ -15,4 +15,48 @@
...
+"""
+fedora-messaging consumer.
+
+This module is responsible for consuming the messaging from the fedora-messaging bus.
+It has the role to inspect the topics of the message and call the correct handler.
+"""
+import logging
+
+import fedora_messaging
+
```
bowlofeggs, line 27:
```suggestion
```

bodhi/server/consumers/__init__.py
```diff
@@ -15,4 +15,48 @@
...
+
+log = logging.getLogger('bodhi')
+
+
+def messaging_callback(msg: fedora_messaging.api.Message):  # noqa: D401
+    """
+    Callback method called by fedora-messaging consume.
+
+    Redirect to messages to the correct handler in function of the
```
bowlofeggs:
```suggestion
    Redirect messages to the correct handler using the
```

bodhi/server/consumers/__init__.py
```diff
@@ -15,4 +15,48 @@
...
+    message topic.
+    """
+    log.info(
+        'Received message from fedora-messaging with topic: %s', msg.topic
```
bowlofeggs: f-strings are faster and easier to read, and we can finally use them ☺

```suggestion
        f'Received message from fedora-messaging with topic: {msg.topic}'
```

bodhi/server/consumers/composer.py
```diff
@@ -19,28 +19,31 @@
 The Bodhi "Composer".
 
 This module is responsible for the process of "pushing" updates out. It's
-comprised of a fedmsg consumer that launches threads for each repository being
+comprised of a fedore messaging consumer that launches threads for each repository being
```
bowlofeggs, line 22:
```suggestion
comprised of a fedora messaging consumer that launches threads for each repository being
```

bodhi/server/consumers/composer.py
```diff
@@ -19,28 +19,31 @@
...
 import time
+import typing
 from datetime import datetime
 from http.client import IncompleteRead
 from urllib.error import HTTPError, URLError
 from urllib.request import urlopen
 
-import fedmsg.consumers
 import jinja2
+import fedora_messaging
```
bowlofeggs, line 44:
```suggestion
```

bodhi/server/consumers/composer.py
```diff
@@ -51,6 +54,9 @@
                                transactional_session_maker)
 
 
+log = logging.getLogger('bodhi')
```
bowlofeggs, line 57: This will make the logger show which module emitted the message, which might be a helpful hint:

```suggestion
log = logging.getLogger(__name__)
```

bodhi/server/consumers/__init__.py
```diff
@@ -15,4 +15,48 @@
...
+
+log = logging.getLogger('bodhi')
+
+
+def messaging_callback(msg: fedora_messaging.api.Message):  # noqa: D401
+    """
+    Callback method called by fedora-messaging consume.
+
+    Redirect to messages to the correct handler in function of the
+    message topic.
```
bowlofeggs, line 41:
```suggestion
    message topic.
    
    Args:
        msg: The message received from the broker.
```

bodhi/server/consumers/composer.py
```diff
@@ -642,7 +622,7 @@ def expire_buildroot_overrides(self):
 
     def remove_pending_tags(self):
         """Remove all pending tags from the updates."""
-        self.log.debug("Removing pending tags from builds")
+        log.debug("Removing pending tags from builds")
```
bowlofeggs, line 617: This pull request isn't really about changing the way we reference our logger. I'm not opposed to the change in general, but it would be better to put it in a separate pull request, since it is a separate concern.

bodhi/server/consumers/composer.py
```diff
@@ -1084,27 +1063,26 @@ def _punge(self):
...
+                                        # Nope. No shell for you
+                                        shell=False,
+                                        # Should be useless, but just to set something predictable
+                                        cwd=self.compose_dir,
+                                        # Pungi will log the output compose dir to stdout
+                                        stdout=subprocess.PIPE,
+                                        # Stderr should also go to pungi.global.log if it starts
+                                        stderr=subprocess.PIPE,
+                                        # We will never have additional input
+                                        stdin=self.devnull)
```
bowlofeggs: Why this change? I think ```flake8``` will likely dislike it…

bodhi/server/consumers/signed.py
```diff
@@ -23,50 +23,29 @@
...
-import pprint
-
-import fedmsg.consumers
 
 from bodhi.server import initialize_db
 from bodhi.server.config import config
 from bodhi.server.models import Build
 from bodhi.server.util import transactional_session_maker
 
+import fedora_messaging
```
bowlofeggs: This is a third party import, so it should be placed above the Bodhi imports for PEP-8.

bodhi/tests/server/consumers/test_composer.py
```diff
@@ -35,9 +35,12 @@
 
 from bodhi.server import buildsys, exceptions, log, push
 from bodhi.server.config import config
+
```
bowlofeggs, line 38: For PEP-8:

```suggestion
```

bodhi/tests/server/consumers/test_composer.py
```diff
@@ -35,9 +35,12 @@
...
 from bodhi.server.consumers.composer import (
-    checkpoint, Composer, ComposerThread, ContainerComposerThread, FlatpakComposerThread,
-    RPMComposerThread, ModuleComposerThread, PungiComposerThread)
+    checkpoint, ComposerHandler, ComposerThread, ContainerComposerThread,
+    FlatpakComposerThread, RPMComposerThread, ModuleComposerThread,
+    PungiComposerThread)
+
```
bowlofeggs, line 43:
```suggestion
```

bodhi/tests/server/consumers/test_consumers.py
```diff
@@ -0,0 +1,53 @@
+import mock
```
bowlofeggs:
```suggestion
# Copyright © 2019 Red Hat, Inc.
#
# This file is part of Bodhi.
# This program is free software; you can redistribute it and/or
# modify it under the terms of the GNU General Public License
# as published by the Free Software Foundation; either version 2
# of the License, or (at your option) any later version.
#
# This program is distributed in the hope that it will be useful,
# but WITHOUT ANY WARRANTY; without even the implied warranty of
# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
# GNU General Public License for more details.
#
# You should have received a copy of the GNU General Public License
# along with this program; if not, write to the Free Software
# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
"""Test the bodhi.server.consumers package."""

from unittest import mock
```

bodhi/tests/server/consumers/test_consumers.py
```diff
@@ -0,0 +1,53 @@
...
+
+from bodhi.tests.server import base
+from fedora_messaging.api import Message
```
bowlofeggs, line 20: For PEP-8, this should be in the third party block above the Bodhi imports and below the suggested mock import above.

bodhi/tests/server/consumers/test_consumers.py
bowlofeggs: (Note the use of mock from the standard library - my pull request that drops Python 2 also drops the third party mock.)

bodhi/tests/server/consumers/test_signed.py
```diff
@@ -21,37 +21,16 @@
...
-    def test_calls_super(self, __init__):
-        """Assert that __init__() calls the superclass __init__()."""
-        hub = mock.MagicMock()
-        hub.config = {'environment': 'environment', 'topic_prefix': 'topic_prefix'}
-
-        handler = signed.SignedHandler(hub)
-
-        self.assertEqual(handler.topic, ['topic_prefix.environment.buildsys.tag'])
-        __init__.assert_called_once_with(hub)
+from fedora_messaging.api import Message
```
bowlofeggs, line 23: Same here - we should move the import for PEP-8.

bowlofeggs:
It would be nice if we had the Vagrant environment set up to use these consumers too. Not a requirement for this pull request, but I wouldn't oppose it being part of it. If we decide not to include it here, I'll add it later.

bodhi/server/consumers/__init__.py
cverna: Cool I kept consistency with the old stuff. But more than happy to use f-strings \o/

cverna:
> It would be nice if we had the Vagrant environment set up to use these consumers too. Not a requirement for this pull request, but I wouldn't oppose it being part of it. If we decide not to include it here, I'll add it later.

I ll be more incline merge this without it, to avoid keeping this PR open for too long and having to resolve merge conflicts :crying_cat_face: .

abompard:
Now that Python2 support has been dropped, there are a few places where we could make the code a bit simpler (like classes don't have to inherit from `object` anymore, `super()` can be called without arguments, etc...) but I can make a PR for that later, I understand Clément's pain of dealing with merge conflicts ;-)

abompard:
It looks OK to me, but there are two issues I'd like to address, and I need info from @bowlofeggs for that:

* are the actions triggered by these messages idempotent? With AMQP, it's possible to receive duplicate messages.
* can the actions triggered by these messages last more than 50 seconds? The message callback is run in the main thread, so the client can't heartbeat to the server while the action is running (see [FM issue #130](https://github.com/fedora-infra/fedora-messaging/issues/130). If there are no long-lasting actions it's not a problem, but if there are we must think of a way to handle that. In our case, I think that the best solution will be to pass the action to a task scheduling system like celery, but that's the subject of another PR, so in the meantime we can leave it as-is: the client will get disconnected, will reconnect and get the missing messages, no big deal.

bodhi/tests/server/consumers/test_updates.py
```diff
@@ -283,87 +269,75 @@ def test_update_not_found(self, work_on_bugs, fetch_test_cases, sleep):
...
+        """
+        h = updates.UpdatesHandler()
+        h.db_factory = base.TransactionalSessionMaker(self.Session)
+        message = Message(
+            topic='bodhi.update.edit',
+            body={'msg': {'update': {}, 'new_bugs': ['12345']}}
+        )
+        h(message)
+
+>>>>>>> Replace fedmsg by fedora-messaging in consumers.
```
abompard: Uh, this looks like a merge conflict leftover...

abompard (changes requested):

abompard:
Those two issues have been logged as #3069 and #3070, so they're not blocking this one.

bodhi/tests/server/consumers/test_updates.py
cverna: Fixed

cverna:
Ok I have most of the tests passing now, not sure why the 3 others tests are failing.

abompard:
One is pydocstyle being unhappy that `bodhi/server/consumers/signed.py:44`'s `__init__` method has no docstring, the other is a timeout waiting for PostgreSQL to start up (you can ignore that one).

bowlofeggs:
On Mon, 2019-03-18 at 13:56 -0700, Aurélien Bompard wrote:
> the other is a timeout waiting for PostgreSQL to start up (you can
> ignore that one).

Does your commit tree have
https://github.com/fedora-infra/bodhi/commit/db5780d99851e8801916c24c2dd6a0bc242936aa
in it? If not, you might be able to fix this by rebasing. If so, maybe
we need to raise the timeouts even further?

abompard:
I don't think it has it, the timeout was 30s IIRC.

cverna:
jenkies test

abompard:
The integration tests were "Killed by signal 15", I don't know what happened...

cverna:
Ok so locally all the integration tests pass except for rawhide where starting the container seems to fail.

See https://paste.fedoraproject.org/paste/y3uIUqe2Dh-zpztvO-Berw

abompard (approved):
OK, looks good, we'll address the issues of duplicate messages in following PRs

bowlofeggs:
On Tue, 2019-03-19 at 08:00 +0000, Clement Verna wrote:
> jenkies test

Unfortunately, I don't think there's any way to re-trigger the tests
with our current setup, other than pushing commits.

bowlofeggs:
On Tue, 2019-03-19 at 08:59 +0000, Aurélien Bompard wrote:
> The integration tests were "Killed by signal 15", I don't know what
> happened...

The Groovyscript has a timeout - could it be that the tests took too
long and hit that timeout?

We recently started testing on Fedora 30, which increases the load on
the testing box. The testing Box just runs all tests in parallel (i.e.,
it doesn't limit them like bodhi-ci does). I can think of three
approaches that might help:

0. Increase the Groovyscript timeout.
1. Figure out how to get Groovy to limit how many jobs it runs at once.
   I think this is possible.
2. Refactor the Groovyscript to run bodhi-ci once and let it manage
   this for us, since it already has the feature. For this to work, we
   will need to give bodhi-ci a feature to emit machine readable text
   about the state of all the jobs so the Groovyscript can report the
   test results to GitHub.

bowlofeggs:

bowlofeggs:

PR #2961, opened 2019-01-24
[WIP] Automatically create updates.

### LINKED ISSUE #3007: [L] Bodhi should be able to automatically create updates when builds are tagged into Koji tags
When Bodhi sees Koji builds getting tagged for a Release, it should automatically create updates for those builds.

In order to achieve this, we will need the following:

* A fedora-messaging listener that looks for builds getting tagged into particular tags, creating updates for those tags automatically.
* The update creator should be set to whichever FAS user created the build.
* Looks at the changelog to see if there are referenced RHBZ #'s that should be associated with the update.
* Uses the changelog entries to fill out the update description.

Note: This ticket will need some clearer requirements defined. For example, should the ```Release``` model have a boolean attribute that indicates whether or not automatic updates should be created for that release, or should Bodhi just create updates for any builds it sees into the specified Koji tags? i.e., would we want Bodhi to automatically create updates for stable Fedora releases too, and not just Rawhide?

bowlofeggs:
When Bodhi sees Koji builds getting tagged for a Release that is
configured to have automatic updates built (such as Fedora
Rawhide), it can now automatically create Updates for those builds.

fixes #3007 

### DIFF
--- a/bodhi/server/consumers/create_update.py
+++ b/bodhi/server/consumers/create_update.py
@@ -1,0 +1,66 @@
+# Copyright © 2019 Red Hat Inc., and others.
+#
+# This file is part of Bodhi.
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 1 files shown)


abompard:
This depends on #2957 being finished, to align with the way fedora-messaging consumers should be written.

bowlofeggs:
@nphilipp is working on this so I will close this.

PR #2962, opened 2019-01-24
[WIP] Provide a way for users to create updates from Koji side tags.

### LINKED ISSUE #3009: [S] - API users should be able to create updates based on Koji side tags
Bodhi currently allows users to create updates by specifying a list of builds via the API. In order to support Rawhide gating without addition additional steps to packagers for Rawhide packaging*, we will need to provide a way for packagers to create multi-build updates. In #3007 we are planning to automatically create updates when we see builds tagged into particular Koji tags. That will work for single build updates, but packagers often need to update multiple packages atomically due to build dependencies.

The plan here is to allow packagers to work in Koji side tags to do their builds. When they are ready to publish their work, they will ask Bodhi to create an update based on that side tag. Bodhi will then query Koji to get the needed information and create an update with the builds found there.

The API needs to allow packagers to create updates based on NVRs, as they do today, or based on koji side tag names. When packagers use a side tag, Koji should be queried to find the list of NVRs, and then Bodhi can proceed as it would have before.

* Packagers are not used to dealing with Bodhi for Rawhide. We could just start using Bodhi for Rawhide today, but many packagers would be confused and some may dislike the extra step. This plan is about trying to minimize any extra burden placed on packagers as we add gating to Rawhide.

bowlofeggs:
Users can now select a Koji side tag to create an Update. Bodhi
will look at the Koji tag to find the list of Builds and use that
to create the resulting update.

fixes #3009 

### DIFF
--- a/bodhi/client/__init__.py
+++ b/bodhi/client/__init__.py
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-# Copyright © 2014-2018 Red Hat, Inc. and others.
+# Copyright © 2014-2019 Red Hat, Inc. and others.
 #
 # This file is part of Bodhi.
 #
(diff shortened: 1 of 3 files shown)


bowlofeggs:
@nphilipp is working on #3009 so I will close this.

PR #2963, opened 2019-01-24
[WIP] Respond synchronously to Greenwave decisions.

### LINKED ISSUE #3008: [M] Bodhi should respond synchronously to Greenwave decisions
Bodhi currently polls Greenwave to find out its gating decisions, using ```bodhi/server/scripts/check_policies.py```. This isn't great, because the web interface polls Greenwave on every load, which means that there can be a lag between Bodhi's front end and back end on the decision.

We should add a fedora-messaging consumer that processes decisions from Greenwave. It should update the same information that ```check_policies.py``` does.

The lack of this feature causes confusion among packagers, as they see that their update passed checks in the Bodhi UI and so ask infrastructure members why their update hasn't moved to stable yet.

Since Greenwave is still using fedmsg, which is not reliable, we should not remove Bodhi's polling as part of this work. This way our polling can be used to ensure we have eventual consistency when we miss messages.

bowlofeggs:
This commit adds a Greenwave message listener so that Bodhi can
react synchronously to Greenwave decisions.

fixes #3008 

### DIFF
--- a/bodhi/server/consumers/greenwave.py
+++ b/bodhi/server/consumers/greenwave.py
@@ -1,0 +1,43 @@
+# Copyright © 2019 Red Hat Inc., and others.
+#
+# This file is part of Bodhi.
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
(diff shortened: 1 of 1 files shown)


abompard:
This depends on #2957 being finished, to align with it.

bowlofeggs:
@tdawson is working on this, so I'm going to close this WIP patch.

PR #2971, opened 2019-02-02
Disallow null values in stable_karma and unstable_karma.

### LINKED ISSUE #1029: bodhi.server.models.models.Update.stable_karma should disallow NULL values
Many of our columns have this problem, but the stable_karma in particular should disallow NULL values and require the column to always be explicitly True or False. Since it allows NULL, there are all kinds of checks in the code that look like `stable_karma not in (0, None)`. As it is, there is possibility of new code forgetting about the two possible "Falsy" values. It's better for the database to be opinionated.

Of course, we will also need to write a migration to set stable_karma to False where it is None and then convert the column to be non-nullable. Due to the migration, I am not marking this as an easy fix (even though it probably actually is).

sebwoj:
Fixes #1029

### DIFF
--- a/bodhi/server/migrations/versions/bdf0e37ab793_disallow_null_values_in_stable_karma_.py
+++ b/bodhi/server/migrations/versions/bdf0e37ab793_disallow_null_values_in_stable_karma_.py
@@ -1,0 +1,58 @@
+# Copyright (c) 2019 Sebastian Wojciechowski.
+#
+# This file is part of Bodhi.
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+"""
+Disallow null values in stable_karma and unstable_karma.
+
+Revision ID: bdf0e37ab793
+Revises: 6b3eb9ae2b87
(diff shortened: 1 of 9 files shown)


bodhi/client/__init__.py
```diff
@@ -109,8 +109,8 @@ def _set_logging_debug(ctx, param, value):
     click.option('--request', help='Requested repository',
                  type=click.Choice(['testing', 'stable', 'unpush', 'batched'])),
     click.option('--autokarma', is_flag=True, help='Enable karma automatism'),
-    click.option('--stable-karma', type=click.INT, help='Stable karma threshold'),
-    click.option('--unstable-karma', type=click.INT, help='Unstable karma threshold'),
+    click.option('--stable-karma', type=click.INT, help='Stable karma threshold', default=3),
+    click.option('--unstable-karma', type=click.INT, help='Unstable karma threshold', default=-3),
```
bowlofeggs: I think it's better to enforce the defaults server-side like we were doing before.

bowlofeggs (changes requested):
This change does a little bit more than disallowing ```NULL``` values as it also removes the meaning of having the karma values set to ```0```.

That said, I'm not sure what use case there really would be for setting it to ```0```. We do have the ```autokarma``` bool to select whether updates push when they reach a karma threshold, so it seems a little odd to also give ```0``` the meaning of not having Bodhi comment on updates, but that's what the current code seems to do.

I wonder if we should start a discussion on fedora-devel to ask if anyone uses a value of 0 and why. That might give us more confidence in removing it.

sebwoj:
I think that it is impossible to set stable/unstable karma to 0:
https://github.com/fedora-infra/bodhi/blob/ceff3c7a0fc7bf4913b5566d4623bbd43e25e74c/bodhi/server/schemas.py#L213

sebwoj:
Unintentionally closed...

bowlofeggs (approved):
Thanks!

bowlofeggs:
Ah, that conflict is likely due to the other patch that renamed the file - I'll fix it up for you!

bowlofeggs:
Looks like the DB timed out in that CI test. I think https://github.com/fedora-infra/bodhi/pull/3054 will help with that. I'll manually test and merge this one.

bowlofeggs:

bowlofeggs:

PR #2973, opened 2019-02-03
Correct TestCaseKarma comment

sebwoj:


bowlofeggs (approved):
Thanks!

bowlofeggs:
I don't believe this patch could have cased the integration test failure, and especially not only on Fedora 28. Additionally, I've witnessed that same failure a few times before on my own machine, so I think we need to figure out why it happens and mitigate it. I'll manually merge this.

PR #2975, opened 2019-02-04
Limit how many concurrent integration tests run.

bowlofeggs:
I found that on my laptop, the integration tests would often fail
when run with bodhi-ci's default -j flag in a Vagrant guest. I
discovered that the failures were very often due to timeouts in
operations on the postgresql database, and I also discovered that
running bodhi-ci with a small -j flag (such as -j 2) would get them
to pass. Rather than defaulting to a smaller -j flag value, this
commit adds a second concurrency semaphore for IntegrationJobs to
limit how many of them can run at once.

The chosen value is admittedly a bit "magic", but was chosen
because it happens to get the tests to pass on my Lenovo T480s with
a Vagrant guest.


bowlofeggs:
I marked this as WIP because I'm still testing the value to make sure it reliably works on my laptop.

devel/ci/bodhi-ci
```diff
@@ -452,39 +462,47 @@ class Job(object):
...
-                        except ProcessLookupError:
-                            # The process is already stopped, nothing to see here.
-                            pass
-                        cancelled_stdout, stderr = await process.communicate()
-                        if self._stdout:
-                            self._stdout = self._stdout + cancelled_stdout
+            if self.extra_concurrency_semaphore:
+                # This is a job that has stricter limitations on how many can run concurrently, so
+                # we also need to acquire the extra semaphore.
+                await self.extra_concurrency_semaphore.acquire()
```
abompard: Since Python 2.7 and 3.1 you can use [multiple context managers on the same line](https://docs.python.org/3.1/reference/compound_stmts.html#with), so you could do something like: `async with self.concurrency_semaphore, self.extra_concurrency_semaphore:`
You may however need to initialize the extra semaphore to something other than `None`, possibly using a class defining no-op `__aenter__` and `__aexit__` methods (since [contextlib.asynccontextmanager](https://docs.python.org/3.7/library/contextlib.html#contextlib.asynccontextmanager) is only available in Python 3.7).
It would save you the big try/except, but it's also a bit more code so I'm not sure it's worth it.

bowlofeggs: Yeah that's not a bad idea - I'll fiddle with that a bit and see how it works. I agree it'd be nice to not use the giant ```try```/```except```.

bowlofeggs:
Using a value of 2 does seem to run the tests reliably on my laptop. 3 was not reliable.

bowlofeggs:
Hmm, I actually just saw a failure even with 2 running at once. I might need to set concurrency to 1 for it.

bowlofeggs:
Hmm, I've now seen it even fail with 1, possibly due to other regular tests running concurrently. I wonder if I just need to limit the overall job count to a lower number in addition to preventing too many integration tests from running concurrently…

bowlofeggs:
The current failure I think is due to concurrent downloads of the database dump. I've observed that running bodhi-ci locally also concurrently downloads it multiple times instead of once. I suspect that our GroovyScipt is also downloading it more than once. I need to stop work for the day, but I think I can fix that tomorrow.

bowlofeggs:
Logs are [here](https://ci.centos.org/job/bodhi-pipeline/job/PR-2975/5/execution/node/584/log/?consoleFull), and you can ctrl-f for the error:

```
psycopg2.ProgrammingError: constraint "packages_stack_id_fkey" of relation "packages" does not exist
```

abompard:
The Jenkins log file does not show multiple downloads of the dumps, however above the error line it has:
```
22:10:26 22:09:37.581 __init__.py       DEBUG  time="2019-02-05T22:09:37Z" level=error msg="Can't add file /root/payload/devel/ci/integration/dumps/bodhi2.dump to tar: archive/tar: write too long" 
```
Since the Bodhi dump isn't huge, I think this may come from a concurrency issue too. Or because the virtual disk is full.

There may be a workaround: copy the zipped files to the postgres container and unzip them locally while doing the import. I'll add a commit to do that and you can tell me what you think.

bowlofeggs:
@abompard Keep in mind that the log I sent you was from a single run of ```bodhi-ci```. The Groovy script actually starts several ```bodhi-ci``` processes in parallel. You can see the combined logs of all those processes [here](https://ci.centos.org/job/bodhi-pipeline/job/PR-2975/5/consoleFull). If you search there, you will see that it does get downloaded multiple times. I believe it is likely that one download was overwriting a file while it was being read by another process.

I don't think extracting the dumps in the container would necessarily fix it, because the read can still happen at the same time as a write.

The fix I'd been contemplating is to have a new ```bodhi-ci``` command that downloads the database dump and have the Groovy script run that one time before the integration tests. The other thing I had in mind was to ensure that ```bodhi-ci integration``` checks the timestamp on the download to ensure it is within a day (since we do daily backups) and skip downloading if it is (you may already have this in the code - I haven't looked closely at it yet). This second component would be nice because developers probably don't want to have to call ```download``` - it's really just a way to get the Groovy script to do it once since there are multiple processes.

Another fix is just to go back to the way it used to work - have each container download the database dump. Inefficient, but avoids these complexities. However, I think it's not too much work to do the download thing I described above, and you've already done the work to bring the process outside of the container, so I think I'd tend towards solving the single download as described above.

The other thing I've considered is to make it so the Groovy script just runs bodhi-ci all instead of running a bunch of bodhi-ci's in parallel. There are positives and negatives. On the positive, it would allow it to work in a more similar manner to how developers would use it locally. It would also allow us to have one place to manage the somewhat complex dependency chain (since this is an example of duplicate work - we have to fix it in both places). The downsides are that it would be engineering effort in the short term because we'd have to refactor the Groovy and bodhi-ci: we'd need to give bodhi-ci a machine readable output format to describe the jobs and refactor the Groovy to read that format so it can report to GitHub. Another downside is that it might be challenging to get it to "live update" the GitHub statuses as it does today (the Groovy would need to be able to live process as bodhi-ci sends updates to stdout about the progress of the various tasks.)

Anyways, I think refactoring is well outside the scope and purpose of this PR, but I thought I'd throw the idea out there to see what you think. For now, I think we just need to ensure that both bodhi-ci and the Groovy script only download the dump one time.

devel/ci/bodhi-ci
```diff
@@ -1128,6 +1128,7 @@ def _build_jobs_list(
         # Download the DB dumps
         integration_deps_build_jobs.extend([
             IntegrationDumpDownloadJob(app_name="waiverdb", release="prod", buffer_output=buffer_output),
+            IntegrationDumpDownloadJob(app_name="bodhi2", release="prod", buffer_output=buffer_output),
```
bowlofeggs, line 1132: +1 - I had planned to do this today ☺

abompard:
Right, I forgot that Groovy starts multiple bodhi-ci instances. We could also use a lockfile system (I have that almost ready), let me send you the commit and you can tell me if it's worthwhile.

bowlofeggs:
All of the failures were actually in reporting the results to GitHub and not in the tests. You can see that by looking at the red and green balls here:

https://ci.centos.org/job/bodhi-pipeline/job/PR-2975/7/flowGraphTable/

@abompard do you approve my commits for merging? I approve your commit, for the record.

bowlofeggs:
I'm going to merge this without review, because it is blocking me from reviewing other pull requests and from fixing the download issue.

abompard:
Post merge review approval! :-)
Sorry I missed your message.

bowlofeggs:
Thanks @abompard! No worries ☺

PR #2981, opened 2019-02-06
Download database dumps first in CentOS CI

bowlofeggs:
There are two commits in this pull request, with detailed commit messages. The overall theme of this pull request is to make sure that database downloads only happen once, even when multiple bodhi-ci processes are run in CentOS CI.

### DIFF
--- a/devel/ci/bodhi-ci
+++ b/devel/ci/bodhi-ci
@@ -299,6 +299,15 @@ def integration_build(concurrency, container_runtime, failfast, pyver, release, 
     _run_jobs(_build_jobs_list('integration-build', concurrency, release, pyvers=pyver))
 
 
+@cli.command("integration-download")
+@concurrency_option
+@failfast_option
+@tty_option
+def integration_download(concurrency, failfast, tty):
+    """Download the database dumps for integration testing."""
+    _run_jobs(_build_jobs_list('integration-download', concurrency, releases=[], pyvers=[3]))
+
+
 @cli.command()
 @archive_option
 @concurrency_option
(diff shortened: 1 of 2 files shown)


devel/ci/bodhi-ci
```diff
@@ -299,6 +299,15 @@ def integration_build(concurrency, container_runtime, failfast, pyver, release,
     _run_jobs(_build_jobs_list('integration-build', concurrency, release, pyvers=pyver))
 
 
+@cli.command("integration-download")
+@concurrency_option
+@failfast_option
+@tty_option
+def integration_download(concurrency, failfast, tty):
+    """Download the database dumps for integration testing."""
+    _run_jobs(_build_jobs_list('integration-download', concurrency, releases=[], pyvers=[3]))
```
abompard, line 308: You could also have defined the two jobs right here, as with the `clean` command. This way you wouldn't have to add conditionals to the `_build_jobs_list` function.

devel/ci/bodhi-ci
```diff
@@ -1017,14 +1017,38 @@ class IntegrationDumpDownloadJob(BuildJob):
 
         self._app_name = app_name
         self._label = 'integration-dump-download-{}'.format(app_name)
-        filepath = os.path.join("devel", "ci", "integration", "dumps", "{}.dump".format(self._app_name))
+        self.filepath = os.path.join(
+            "devel", "ci", "integration", "dumps", "{}.dump".format(self._app_name))
         url = "https://infrastructure.fedoraproject.org/infra/db-dumps/{app}.dump.xz".format(app=self._app_name)
         self._popen_kwargs['shell'] = True
-        self._command = ["curl -o {filepath}.xz -R -z {filepath}.xz {url} && xz -d --keep --force {filepath}.xz".format(url=url, filepath=filepath)]
```
abompard, line 1023: I would recommend keeping the `-R -z filepath` switches :
- `-R` sets the remote mtime on the downloaded file
- `-z filepath` does not download the file if the remote mtime is older.
I think it's complementary to the "1-day" check you do in `run()`.
Did you remove them for a specific reason? Did they not work?

devel/ci/bodhi-ci
abompard: On the other hand, it means instantiating the classes twice, which is kind of against DRY. Not sure which is the lesser evil.

devel/ci/bodhi-ci
bowlofeggs: @abompard yeah there were actually errors from curl - I think the server didn't always send the date in a format that curl wanted.

That wasn't the only reason though - I also wanted to be able to mark the state as skipped so we had a clear record of what happened.

devel/ci/bodhi-ci
bowlofeggs: Yeah I kinda like having once source of setting up those classes.

devel/ci/bodhi-ci
abompard: OK !

abompard (approved):

PR #2987, opened 2019-02-08
Drop support for Python 2.

### LINKED ISSUE #2856: Drop Python 2 support for bodhi.client

bowlofeggs:
fixes #2856

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -11,25 +11,22 @@ pull_request_rules:
   - status-success=f28-docs
   - status-success=f28-flake8
   - status-success=f28-pydocstyle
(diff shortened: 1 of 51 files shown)


bowlofeggs:
There's still more to do here, like remove the usage of six and get the tests passing. But it's Friday and I'm going afk for the weekend ☺

Zlopez (approved):
It's nice that the Bodhi get rid of the Python 2

bowlofeggs:

bowlofeggs:

PR #2990, opened 2019-02-11
Use a setting to configure the fedmsg-hub WS URL.

### LINKED ISSUE #2523: The search bar does not work in staging

bowlofeggs:
The fedmsg-hub web socket URL was hardcoded in live.js. This commit
adds a setting for that URL and injects it into the master template
via a <script> tag so that all subsequently loaded JavaScript can
assume that setting is available for use.

This change also adds a nice place for other settings to be added
later, if desired.

fixes #2523

### DIFF
--- a/bodhi/server/config.py
+++ b/bodhi/server/config.py
@@ -428,6 +428,9 @@ class BodhiConfig(dict):
         'fedmsg_enabled': {
             'value': False,
             'validator': _validate_bool},
+        'fedmsg_gateway': {
+            'value': 'wss://hub.fedoraproject.org:9939',
+            'validator': str},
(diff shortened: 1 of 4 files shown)


bowlofeggs:
The test failure is due to #2991 and can be ignored.

abompard (approved):

bowlofeggs:

bowlofeggs:

PR #2992, opened 2019-02-12
Wait for PostgreSQL to actually startup

abompard:
See #2991.


bowlofeggs (approved):
Thanks!

bowlofeggs:
I now wonder if this will make it so we don't need the patch I wrote last week to only run fewer integration tests in parallel. Once it merges I'll try reverting my patch and see if it works - I have a feeling this might have been the same problem I was hitting.

PR #2993, opened 2019-02-12
Use Python 3 when provisioning Vagrant with Ansible.

bowlofeggs:

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -57,7 +57,6 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
  config.vm.provision "shell", inline: "sudo dnf upgrade -y"
 
  # bootstrap and run with ansible
- config.vm.provision "shell", inline: "sudo dnf -y install python2-dnf libselinux-python"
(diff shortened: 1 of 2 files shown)


bowlofeggs:
This currently won't work on my Fedora 29 host due to https://github.com/ansible/ansible/issues/49120. I'm not sure how long it will be until the fix for that issue will be available for Fedora hosts, so I'm marking this as no-mergify and we can just hold the PR open until it works.

pypingou (approved):

Zlopez (approved):
:+1:

bowlofeggs:
Three months later, and this error still happens. I guess I'll just try again later. I'll keep this open for a while longer.

cverna:
> Three months later, and this error still happens. I guess I'll just try again later. I'll keep this open for a while longer.

I think you ran into https://github.com/ansible/ansible/pull/50381 with newer version of RabbitMQ is breaking the module.

bowlofeggs:
I tested this again today and it seems to work with ```ansible-2.8.2-1.fc30.noarch```.

PR #2994, opened 2019-02-12
Drop the batching feature.

### LINKED ISSUE #2977: Drop the batching feature
Unfortunately, the batching feature was pretty controversial among the Fedora community and we were unable to garner enough support to keep it. FESCo has [voted](https://pagure.io/fesco/issue/1820) to remove it.

I would personally like to keep the feature, but I would also prefer not to have to maintain it if Fedora is not going to use it.

bowlofeggs:
The batching feature was controversial in the Fedora community.
FESCo opted to disable it several months ago, and more recently
voted to remove the feature.

fixes #2977


bowlofeggs:
Credit to @sebwoj in #2954 for coming up with the migration code to modify an enum.

Zlopez (approved):
LGTM

bowlofeggs:

bowlofeggs:

PR #2995, opened 2019-02-12
Rename prefer_ssl to libravatar_prefer_tls.

### LINKED ISSUE #1921: Rename the prefer_ssl setting
Bodhi has a ```prefer_ssl``` setting, which doesn't do what it sounds like it does at all because it's only used when ```libravatar_dns``` is ```True```, and is only used by ```libravatar``` and not generally all of Bodhi. We should rename it to something that makes that clearer, like ```libravatar_prefer_tls```.

This is a backwards incompatible change and should be done in an X release.

bowlofeggs:
The prefer_ssl setting is renamed to libravatar_prefer_tls by this
commit. The previous name was far too generic of a name, and
implied that it had some global impact when it only affected
libravatar.

The new setting also defaults to True instead of None, a saner
default.

fixes #1921

### DIFF
--- a/bodhi/server/config.py
+++ b/bodhi/server/config.py
@@ -477,6 +477,9 @@ class BodhiConfig(dict):
         'libravatar_enabled': {
             'value': True,
             'validator': _validate_bool},
+        'libravatar_prefer_tls': {
+            'value': True,
+            'validator': bool},
         'mail.templates_basepath': {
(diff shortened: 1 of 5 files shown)


pypingou (approved):

Zlopez (approved):
LGTM

bowlofeggs:

bowlofeggs:

PR #2999, opened 2019-02-13
Better describe Update model

abitrolly:
Correct me if I wrong.

image

### DIFF
--- a/bodhi/server/models.py
+++ b/bodhi/server/models.py
@@ -1679,6 +1679,10 @@ class Update(Base):
     """
     This model represents an update.
 
+    The update contains not just one package, but a collection of packages. Each
+    package can be referenced only once in one Update. Packages are referenced
+    through their Build objects using field `builds` below.
+
     Attributes:
         autokarma (bool): A boolean that indicates whether or not the update will
             be automatically pushed when the stable_karma threshold is reached.


bodhi/server/models.py
```diff
@@ -1555,7 +1555,8 @@ def get_changelog(self, timelimit=0):
 
 class Update(Base):
     """
-    This model represents an update.
+    Update model. Single update contains builds for one or more packages, each
+    package can have more than one build.
```
bowlofeggs: I would word it a bit differently. I'd probably keep the original text and add a new paragraph to expand on it:

```
This model represents an update.

An update contains a set of Build objects that the packager is proposing to update atomically in the release.
```

Though a package can have more than one build in the database, it cannot have more than one build in the same update.

bowlofeggs (changes requested):

bodhi/server/models.py
abitrolly: I don't get it. If package can not have more than one build, then why the documentation says that update contains a set of builds?

bowlofeggs: A package can have more than one build, but an update can only have one build per package.

abitrolly: In other words the update is a just a label or tag on a package build. Is that right?

abitrolly: There is contradiction here. First it is said that "update contains a set of Build objects" but then you say that update can only have one build. Where is the truth?

bowlofeggs: @abitrolly The update has a set of builds. Each build has an associated package. Bodhi enforces a rule that there can only be one build per package associated with an update. As an example of what I mean, it is OK for an update to have httpd-1.0.0-1.fc29 and python3-3.7.0-1.fc29 in it. It is not allowed for a single update to have both httpd-1.0.0-1.fc29 and httpd-1.0.0-2.fc29 in it.

abitrolly: In human world, update contains new package versions. Then it becomes logical. Those builds are also packages, right?

abitrolly: There is a bug in semantic model, which means that `bodhi` interface is confusing. Here is the https://bodhi.fedoraproject.org/updates/?releases=F30&status=pending

image

As you may see `Updates` list actually lists `packages`, not updates. It lists ` oniguruma-6.9.2-1.fc30` package name and not update name, which is `FEDORA-2019-abe79f6459`. Then following https://bodhi.fedoraproject.org/updates/FEDORA-2019-abe79f6459 there is no mention of `oniguruma-6.9.2-1.fc30` unless somebody goes to `Builds` tab.

In the end, the `new package` is called `update` then `build`, and that's why I couldn't understand the code.

How to find an update for more than one build to see how it looks like?

abitrolly: Found an `update pack`. It doesn't stand out from the UI except for the long name, which is not completely evident.

image

There are 5 builds and an instruction how to install it. https://bodhi.fedoraproject.org/updates/FEDORA-2019-7ee81eedb3

These `update packs` could stand out to make it clear that `update` may contain a bunch of packages. A package count on update could do the trick.

nphilipp: > Those builds are also packages, right?

Unfortunately, it's not that simple. In the most common case, the "builds" will be RPM packages, but they could e.g. also be containers or modules. For example, the [FEDORA-MODULAR-2019-088c6693d6](https://bodhi.fedoraproject.org/updates/FEDORA-MODULAR-2019-088c6693d6) update contains the [gimp-2.10-2920190409123811.6c81f848](https://koji.fedoraproject.org/koji/buildinfo?buildID=1247188) "build" for stream 2.10 of the gimp module which in turn contains the [gimp-2.10.10-1.module_f30+3878+d4400ad3](https://koji.fedoraproject.org/koji/buildinfo?buildID=1247146) package as an installable artifact.

abitrolly: There is no way one can figure this out from reading the source code. Without simplifying this part or documenting it, there is a little chance for potential contributors to understand.

bodhi/server/models.py
```diff
@@ -1556,11 +1556,15 @@ def get_changelog(self, timelimit=0):
...
+    
+    The update contains not just one package, but a collection of packages. Each
+    package can be referenced only once in one Update. Packages are referenced
+    through their Build objects using field `builds` below.
 
     Attributes:
         title (unicode): The update's title which uniquely identifies the update.
-            This is generally an ordered list of the build NVRs contained in the
-            update.
+            This can be a single name-version-releases (NVR) of updated package, or
```
```suggestion
            This can be a single name-version-release (NVR) of the updated package, or
```

bodhi/server/models.py
```diff
@@ -1556,11 +1556,15 @@ def get_changelog(self, timelimit=0):
...
+            an ordered list of NVRs if update contains several packages.
```
```suggestion
            an ordered list of NVRs if the update contains several packages.
```

bowlofeggs (dismissed):
Looks good, just a couple typos. I'm going to squash your commits, fix the typos, rebase, and merge.

bowlofeggs:
Actually, the conflict here is that the title attribute was removed (in Bodhi 4.0.0), so all of the typos I mentioned don't need to be fixed since that field is gone now. I'll just drop that and keep the first part.

bowlofeggs:
@abitrolly In order to accept this patch, we need a signature on the DCO as described in the contributor docs:

Can you add that signature to your commit message?

abitrolly:
@bowlofeggs sorry for the delay. Finally got time to revisit it. Added signature and rebased.

The time required to revisit the issue is proportional to the last experience or how complicated the issue was. )

ryanlerch (approved):
@bowlofeggs 's changes appear to be applied now. LGTM

abitrolly:
Rebased.

PR #3002, opened 2019-02-14
Explicitly require psycopg2.

bowlofeggs:
The downstream spec file was recently refactored to use Python
dependency autogeneration, and in the process all Requires: lines
on Python packages were removed, including psycopg2. This caused
Bodhi to fail to run when installed from that spec file,
highlighting that upstream should have listed psycopg2 as a
requirement.

PostgreSQL is Bodhi's only supported database and has long been so,
though it has not been formally listed in the requirements.txt
until this commit.


puiterwijk (approved):
lgtm

PR #3003, opened 2019-02-14
Specify only a minimum version on Pyramid.

bowlofeggs:
Fedora Infrastructure's Application Security Policy dictates that
applications should only specify a minimum version on
dependencies[0]. This commit adjusts the dependency on pyramid to
comply with that policy.

https://fedora-infra-docs.readthedocs.io/en/latest/dev-guide/security_policy.html#dependencies


Zlopez (approved):
@puiterwijk will be happy

PR #3012, opened 2019-02-15
Compare enum values instead of descriptions.

bowlofeggs:
There were multiple places in the template code that were
comparing Enum human readable descriptions instead of the Enum
values. The descriptions are meant to be human readable and are
subject to change, so this code was at risk of breaking if someone
changes those descriptions in the future.


pypingou (approved):

bowlofeggs:

bowlofeggs:

PR #3013, opened 2019-02-16
Drop the relationship between Packages and Users

### LINKED ISSUE #1997: The relationship between Packages and Users is unused
I noticed today while writing docblocks for the models that the Packages.committers relationship is always empty. To confirm this, I examined the database dump, and found that there really are no rows in the ```user_package_table``` through table:

```
--
-- Data for Name: user_package_table; Type: TABLE DATA; Schema: public; Owner: bodhi2
--

COPY user_package_table (user_id, package_id) FROM stdin;
\.
```

We should delete this relationship and table.

sebwoj:
Fixes #1997

### DIFF
--- a/bodhi/tests/server/test_models.py
+++ b/bodhi/tests/server/test_models.py
@@ -92,20 +92,17 @@ class TestBodhiBase(BaseTestCase):
 
     def test__expand_with_m2m_relation(self):
         """Test the _expand() method with a many-to-many relation."""
-        p = model.Package.query.all()[0]
-        new_exclude_columns = list(model.Package.__exclude_columns__)
-        new_exclude_columns.remove('committers')
-        # p.committers is an InstrumentedList, which doesn't have an all() method. We have some
+        u = model.Update.query.all()[0]
+        # u.bugs is an InstrumentedList, which doesn't have an all() method. We have some
         # code to handle m2m relationships with all() methods, but it's not immediately obvious
         # which relationships have that for testing purposes. Thus, we can simulate this for
         # test coverage purposes by setting it to its __iter__() method.
-        p.committers.all = p.committers.__iter__
+        u.bugs.all = u.bugs.__iter__
 
-        with mock.patch.object(model.Package, '__exclude_columns__', new_exclude_columns):
-            committers = p._expand(p, p.committers, [], mock.MagicMock())
+        bugs = u._expand(u, u.bugs, [], mock.MagicMock())
 
-        self.assertEqual(len(committers), 1)
-        self.assertEqual(committers[0]['name'], 'guest')
+        self.assertEqual(len(bugs), 1)
+        self.assertEqual(bugs[0]['bug_id'], 12345)
 
     def test__expand_with_relation_in_seen(self):
         """_expand() should return the relation.id attribute if its type is in seen."""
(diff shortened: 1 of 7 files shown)


bodhi/tests/server/test_models.py
```diff
@@ -3258,22 +3242,6 @@ def test_comment_critpath_unapproved(self):
...
-        bowlofeggs = model.User(name=u'bowlofeggs', email=u'bowlofeggs@fp.o')
-        self.obj.builds[0].package.committers.append(bowlofeggs)
-
-        self.obj.comment(self.db, u'Here is a cool e-mail for you.', author=u'bowlofemail')
-
-        bodies = [c[1][2].decode('utf-8') for c in SMTP.return_value.sendmail.mock_calls]
-        self.assertTrue('lmacken' in bodies[0])
-        # In Python 2 this address is in the middle e-mail and in Python 3 it's in the last e-mail
-        self.assertTrue('bowlofeggs@fp.o' in '\n'.join(bodies))
-        self.assertTrue(all(['Here is a cool e-mail for you.' in b for b in bodies]))
```
bowlofeggs, line 3226: Heh, this test makes me think that we were supposed to do this. The problem is that nothing ever set the committers to anything on packages so this never actually happened. I guess we could fix Bodhi to actually set the committers and keep this table/code instead of deleting it. I'm not sure where would be the best place to do that though. Perhaps it would be one of the backend tasks when we create or edit an update?

What do you think? This is a feature that has obviously never worked, so I am OK with accepting this patch and dropping it. But I would also be OK with fixing it to what seems to be the intention here. Since you did the work, I'll let you decide ☺

sebwoj: Hard to say. Is there any requirement for that in Bodhi?
Do you know about someone who mentioned that he didn't receive any notification?

bowlofeggs: I don't know of any requirement to have this feature, and I've not heard from anyone saying they wanted e-mails that they weren't getting.

sebwoj: If so, I think that we can remove this.

bowlofeggs:
Cool, I am going to rebase this and merge.

bowlofeggs:
I am applying this patch as well:

```
diff --git a/bodhi/server/migrations/versions/7ba286412ad4_drop_the_relationship_between_packages_.py b/bodhi/server/migrations/versions/7ba286412ad4_drop_the_relationship_between_packages_.py
index 633a8af3..cfa62a0b 100644
--- a/bodhi/server/migrations/versions/7ba286412ad4_drop_the_relationship_between_packages_.py
+++ b/bodhi/server/migrations/versions/7ba286412ad4_drop_the_relationship_between_packages_.py
@@ -19,7 +19,7 @@
 Drop the relationship between Packages and Users.
 
 Revision ID: 7ba286412ad4
-Revises: 190ba571c7d2
+Revises: 2fc96aa44a74
 Create Date: 2019-02-16 08:36:48.665825
 """
 from alembic import op
@@ -28,7 +28,7 @@ import sqlalchemy as sa
 
 revision = '7ba286412ad4'
-down_revision = '190ba571c7d2'
+down_revision = '2fc96aa44a74'
 
 
 def upgrade():
```

bowlofeggs (approved):

bowlofeggs:
The CI failures are due to the container registry being offline, which is due to the infrastructure mass reboot that is going on right now. We'll need to try again later.

bowlofeggs:

bowlofeggs:

PR #3019, opened 2019-02-19
Gracefully handle private bugs

bowlofeggs:
There are 4 commits in this pull request, each with detailed descriptions. The overall goal is to improve the error handling in ```bugs.py``` and to fix a regression in 3.13.0.

fixes #344 
fixes #3016

PR #3020, opened 2019-02-19
Backport 3016

bowlofeggs:
Backport the first three commits from #3019.


bowlofeggs:
The tests pass locally with bodhi-ci, so I am merging.

Issue #3025, opened 2019-02-19
The word "rpm" is appearing after update descriptions


bowlofeggs:
FranciscoD on Freenode reported that the word "rpm" is appearing after the update description on https://bodhi.fedoraproject.org/updates/FEDORA-2019-cb3b8e1c2c. Here's a screenshot:

![screenshot from 2019-02-19 09-59-44](https://user-images.githubusercontent.com/354506/53024447-1a3eb280-342d-11e9-92a3-522206b268cc.png)

I also observed this on https://bodhi.fedoraproject.org/updates/dino-0.0-0.3.20190116.git.8e14ac6.fc28.

I expect there's some error in the template that is causing this.

PR #3026, opened 2019-02-19
Remove misplaced variable in the update template

abompard:
Fixes: #3025
Signed-off-by: Aurélien Bompard <aurelien@bompard.org>

### DIFF
--- a/bodhi/server/templates/update.html
+++ b/bodhi/server/templates/update.html
@@ -675,7 +675,6 @@ $(document).ready(function(){
         <div class="col-md-9">
           % if update.notes:
           ${self.util.markup(update.notes) | n}
-          ${update.content_type}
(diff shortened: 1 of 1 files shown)


bowlofeggs:
The CI failure is due to https://bugs.centos.org/view.php?id=15791 - the test did pass.

abompard:
The Jenkins error is just a gigantic Java traceback about not being able to reach `api.github.com`, I think we can ignore it.

bowlofeggs (approved):
Thanks for the fix - I confirmed that "rpm" goes away. I'll rebase and merge.

bowlofeggs:
I'm going to manually run CI on this since CentOS CI seems to be out of GitHub API requests.

Issue #3016, opened 2019-02-18
AttributeError: Bug object has no attribute 'private'


bowlofeggs:
Now that 3.13.0 is in production, I've started to receive tracebacks from bodhi-backend02 related to the patch in #2714. It seems that we either need to ask for the "private" attribute of bugs from Bugzilla, or some other solution if that does not exist:

Message
-------
[2019-02-18 17:04:54][     bodhi   ERROR]
```python
Unable to alter bug #1668946
```

Process Details
---------------
- host:     bodhi-backend02.phx2.fedoraproject.org
- PID:      59293
- name:     fedmsg-hub-3
- command:  /usr/bin/python3 /usr/bin/fedmsg-hub-3
- msg_id:   

Callstack that lead to the logging statement
--------------------------------------------
```python
  File "/usr/lib64/python3.7/threading.py", line 885 in _bootstrap
    
  File "/usr/lib64/python3.7/threading.py", line 917 in _bootstrap_inner
    
  File "/usr/lib64/python3.7/threading.py", line 865 in run
    
  File "/usr/lib64/python3.7/site-packages/twisted/_threads/_threadworker.py", line 46 in work
    
  File "/usr/lib64/python3.7/site-packages/twisted/_threads/_team.py", line 190 in doWork
    
  File "/usr/lib64/python3.7/site-packages/twisted/python/threadpool.py", line 250 in inContext
    
  File "/usr/lib64/python3.7/site-packages/twisted/python/threadpool.py", line 266 in <lambda>
    
  File "/usr/lib64/python3.7/site-packages/twisted/python/context.py", line 122 in callWithContext
    
  File "/usr/lib64/python3.7/site-packages/twisted/python/context.py", line 85 in callWithContext
    
  File "/usr/lib/python3.7/site-packages/moksha/hub/api/consumer.py", line 186 in _work_loop
    
  File "/usr/lib/python3.7/site-packages/moksha/hub/api/consumer.py", line 207 in _do_work
    
  File "/usr/lib/python3.7/site-packages/bodhi/server/consumers/updates.py", line 135 in consume
    
  File "/usr/lib/python3.7/site-packages/bodhi/server/consumers/updates.py", line 206 in work_on_bugs
    
  File "/usr/lib/python3.7/site-packages/bodhi/server/models.py", line 4284 in modified
    
  File "/usr/lib/python3.7/site-packages/bodhi/server/bugs.py", line 313 in modified
    
```
Traceback (most recent call last):
[...]

PR #3028, opened 2019-02-19
Mark the bugs.private field as nullable.

### LINKED ISSUE #3024: 400 Bad Request when associating bugs with updates
When I fixed #3016 yesterday, I ended up reverting the commit https://github.com/fedora-infra/bodhi/commit/be369636529f3d637f9888bf658d77adac2e9b8b. I opted not to remove the new private field, since technically it became part of the API, if even for only a day, and that would be a backwards incompatible change. Instead, I wrote a migration to drop the field for the upcoming 4.0.0 release.

The problem is that the field is required to be non-null. This prevents users from adding new bugs to Bodhi, because there is no longer any code to set the field, but it is required. Requests to add bugs to updates will receive a 400 Bad Request response:

Unable to create update. (psycopg2.IntegrityError) null value in column "private" violates not-null constraint DETAIL: Failing row contains (52898, 1671171, null, f, f, null). [SQL: 'INSERT INTO bugs (bug_id, title, security, parent) VALUES (%(bug_id)s, %(title)s, %(security)s, %(parent)s) RETURNING bugs.id'] [parameters: {'bug_id': 1671171, 'title': None, 'security': False, 'parent': False}] (Background on this error at: http://sqlalche.me/e/gkpj)

The fix is to write a migration that make this field nullable.

bowlofeggs:
In my fix for #3016, I opted to revert the commit that added the
private column to the bugs table, but not to remove the column
itself since that would be a backwards incompatible change.

Unfortunately, it did not occur to me that the private field being
non-nullable would mean that users would not be able to create
updates that have bugs associated with them, because there is no
longer code to set that column to a value.

This commit adjusts the column to be nullable so that we don't
introduce a backwards incompatible change, but also can allow users
to associate updates with bugs. Bodhi 4.0.0 will drop the column
entirely.

fixes #3024

### DIFF
--- a/bodhi/server/migrations/versions/9991cf10ec50_mark_the_bugs_private_field_as_nullable.py
+++ b/bodhi/server/migrations/versions/9991cf10ec50_mark_the_bugs_private_field_as_nullable.py
@@ -1,5 +1,4 @@
-# -*- coding: utf-8 -*-
-# Copyright (c) 2019 Sebastian Wojciechowski
+# Copyright (c) 2019 Red Hat, Inc.
 #
 # This file is part of Bodhi.
 #
@@ -17,27 +16,26 @@
 # along with this program; if not, write to the Free Software
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
 """
-Obsolete updates for archived release.
+Mark the bugs.private field as nullable.
 
-Revision ID: 8e9dc57e082d
-Revises: 9991cf10ec50
-Create Date: 2019-01-06 13:04:35.158562
+Revision ID: 9991cf10ec50
+Revises: d986618207bc
+Create Date: 2019-02-19 15:54:53.058707
 """
 from alembic import op
 
 
 # revision identifiers, used by Alembic.
-revision = '8e9dc57e082d'
-down_revision = '9991cf10ec50'
+revision = '9991cf10ec50'
+down_revision = 'd986618207bc'
 
 
 def upgrade():
-    """Set archived releases updates status to 'obsolete'."""
-    op.execute("UPDATE updates SET status='obsolete' WHERE release_id in \
-                (SELECT id FROM releases WHERE state='archived') AND status NOT IN \
-                ('unpushed', 'obsolete', 'stable')")
+    """Set the bugs.private column to be nullable."""
+    op.alter_column('bugs', 'private', nullable=True)
 
 
 def downgrade():
-    """Raise an exception explaining that this migration cannot be reversed."""
-    raise NotImplementedError('This migration cannot be reversed.')
+    """Set the bugs.private column to be non-nullable."""
+    op.execute("""UPDATE bugs SET private = FALSE WHERE private = NULL""")
+    op.alter_column('bugs', 'private', nullable=False)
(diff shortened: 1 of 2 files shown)


bodhi/server/migrations/versions/9991cf10ec50_mark_the_bugs_private_field_as_nullable.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+# revision identifiers, used by Alembic.
+revision = '9991cf10ec50'
+down_revision = 'd986618207bc'
+
+
+def upgrade():
+    """Set the bugs.private column to be nullable."""
+    op.alter_column('bugs', 'private', nullable=True)
```
pypingou, line 35: If you support mysql you'll likely want to provide the existing_type field and from my experience you also likely want to provide the existing_nullable argument

bodhi/server/migrations/versions/9991cf10ec50_mark_the_bugs_private_field_as_nullable.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+def downgrade():
+    """Set the bugs.private column to be non-nullable."""
+    op.add_column('bugs', sa.Column('private', sa.BOOLEAN(), autoincrement=False, nullable=True))
+    op.execute("""UPDATE bugs SET private = FALSE""")
+    op.alter_column('bugs', 'private', nullable=False)
```
pypingou, line 41: same as above

bodhi/server/migrations/versions/9991cf10ec50_mark_the_bugs_private_field_as_nullable.py
```diff
@@ -0,0 +1,43 @@
...
+
+def upgrade():
+    """Set the bugs.private column to be nullable."""
+    op.alter_column('bugs', 'private', nullable=True)
+
+
+def downgrade():
+    """Set the bugs.private column to be non-nullable."""
+    op.add_column('bugs', sa.Column('private', sa.BOOLEAN(), autoincrement=False, nullable=True))
+    op.execute("""UPDATE bugs SET private = FALSE""")
```
pypingou: You're adding a column and filling it but in the upgrade you're not removing it so I'm not sure this will work.

pypingou (commented):
I'm a little worried about the downgrade() method, I'd recommend the existing_nullable.
The rest is up to you

bodhi/server/migrations/versions/9991cf10ec50_mark_the_bugs_private_field_as_nullable.py
bowlofeggs: Bodhi only supports PostgreSQL.

bodhi/server/migrations/versions/9991cf10ec50_mark_the_bugs_private_field_as_nullable.py
bowlofeggs: @pypingou the data is arbitrary. It just needs to be st to something. There hasn't been a release that was able to use this field in a meaningful way.

bowlofeggs: @pypingou I dropped the add column and made it so it only sets it FALSE if NULL.

pypingou (approved):
I remember having issues w/o the existing_nullable but I trust you've tested this and thus I'm likely mis-remembering.

bowlofeggs:
CentOS CI seems to be stuck due to GitHub API rate limiting. The tests passed ```bodhi-ci``` locally, and this is a critical patch, so I am going to merge.

PR #3033, opened 2019-02-20
Begin integration testing with Fedora 30.

bowlofeggs:


bowlofeggs:
The ```f28-diff-cover``` test actually passed, but for some reason CentOS CI failed to report the status to GitHub. I will manually merge this.

PR #3034, opened 2019-02-20
Add static type checking to bodhi.server.buildsys.

bowlofeggs:


Zlopez (approved):
This is very nice, I should start using it too.

Issue #1714, opened 2017-07-26
bodhi.server.models.Update.alias should be non-nullable


bowlofeggs:
Currently, the ```alias``` field on the ```Update``` model is nullable. None of the existing production updates are null, so this isn't necessary or beneficial. Furthermore, there are places in the code that check to see if Update aliases are defined or not, and that log errors if they are not. For example, see [```metadata.py```](https://github.com/fedora-infra/bodhi/blob/2.8.1/bodhi/server/metadata.py#L85-L89). Instead of having code to handle the alias being non-null, it would be better to make it impossible for the alias to be null.

I started working on this today and I decided to file a ticket instead as I don't have abundant time right now, and it isn't entirely a trivial refactor:

1. We will need a migration. The migration will need to loop over updates to look for any that don't have a "truthy" alias and set one.
2. Lots of test code creates updates without aliases, so we need to adjust all test code to make sure aliases are always defined.
3. We can use SQLAlchemy's [context sensitive default feature](http://docs.sqlalchemy.org/en/rel_1_0/core/defaults.html#context-sensitive-default-functions) to automatically set aliases on updates. With that, we can get rid of [```assign_alias()```](https://github.com/fedora-infra/bodhi/blob/2.8.1/bodhi/server/models.py#L1667-L1678) and all code that uses it. However, when I attempted this, I found that lots of code creates Updates without defining the release so that will take some time to sort out too. Here's some example code we could use for the default:

```
def generate_alias(context):
    prefix = context.current_parameters['release'].id_prefix
    year = time.localtime()[0]
    _id = hashlib.sha1(str(uuid.uuid4())).hexdigest()[:10]
    return u'%s-%s-%s' % (prefix, year, _id)
```

As you can see, the example code assumes that the instantiator of the Update provided a release.

bowlofeggs:
A fix for this issue is planned for inclusion in the upcoming 4.0.0 release: #3221

bowlofeggs:
Bodhi 4.0.0 beta is built and deployed to staging:

https://copr.fedorainfracloud.org/coprs/bowlofeggs/bodhi-pre-release
https://bodhi.stg.fedoraproject.org/

PR #3038, opened 2019-02-22
Drop the update.title field from the database

bowlofeggs:
There are two commits in this pull request. Please see their messages for details. The summary:

* The first one indexes a field that really should have always been indexed. This is needed for the downgrade() on the migration for the second commit, which takes more than 10 minutes without an index (I killed it, so no idea how long it actually takes, just that it's more than 10 minutes). With the index, the downgrade() takes about 7 seconds on my laptop.
* The second commit drops the title field from updates, and disallows referencing updates by title.

### DIFF
--- a/bodhi/server/models.py
+++ b/bodhi/server/models.py
@@ -20,7 +20,6 @@
 from collections import defaultdict
 from datetime import datetime
 from textwrap import wrap
-from urllib.parse import quote
 import hashlib
 import json
 import os
@@ -1555,9 +1554,6 @@ class Update(Base):
     This model represents an update.
 
     Attributes:
-        title (unicode): The update's title which uniquely identifies the update.
-            This is generally an ordered list of the build NVRs contained in the
-            update.
         autokarma (bool): A boolean that indicates whether or not the update will
             be automatically pushed when the stable_karma threshold is reached.
         stable_karma (int): A positive integer that indicates the amount of "good"
@@ -1626,10 +1622,8 @@ class Update(Base):
 
     __tablename__ = 'updates'
     __exclude_columns__ = ('id', 'user_id', 'release_id')
-    __include_extras__ = ('meets_testing_requirements', 'url',)
-    __get_by__ = ('title', 'alias')
-
-    title = Column(UnicodeText, unique=True, default=None, index=True)
+    __include_extras__ = ('meets_testing_requirements', 'url', 'title')
+    __get_by__ = ('alias',)
 
     autokarma = Column(Boolean, default=True, nullable=False)
     stable_karma = Column(Integer, nullable=False)
@@ -1667,7 +1661,7 @@ class Update(Base):
     date_stable = Column(DateTime)
 
     # eg: FEDORA-EPEL-2009-12345
-    alias = Column(Unicode(32), unique=True, nullable=True)
+    alias = Column(Unicode(32), unique=True, nullable=False)
 
     # One-to-one relationships
     release_id = Column(Integer, ForeignKey('releases.id'), nullable=False)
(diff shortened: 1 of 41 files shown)


bowlofeggs:
This pull request will not pass CI until #2987 is merged, as it uses some Python 3-only syntax in the CLI.

Zlopez (approved):
LGTM

bodhi/server/models.py
```diff
@@ -1672,7 +1666,7 @@ class Update(Base):
     date_stable = Column(DateTime)
 
     # eg: FEDORA-EPEL-2009-12345
-    alias = Column(Unicode(32), unique=True, nullable=True)
+    alias = Column(Unicode(32), unique=True, nullable=False)
```
sebwoj, line 1664: With this change, we can mark commit as fixing #1714

bodhi/tests/server/test_models.py
```diff
@@ -3248,7 +3246,7 @@ def test_comment_empty(self):
         self.assertEqual(str(exc.exception), 'You must provide either some text or feedback')
 
     def test_get_url(self):
-        self.assertEqual(self.obj.get_url(), u'updates/TurboGears-1.0.8-3.fc11')
+        self.assertEqual(self.obj.get_url(), f'updates/{self.obj.alias}')
         idx = 'a3bbe1a8f2'
         with mock.patch(target='uuid.uuid4', return_value='wat'):
             self.obj.assign_alias()
```
sebwoj: Method `assign_alias` was removed.

bodhi/tests/server/test_models.py
```diff
@@ -3561,14 +3559,14 @@ def test_validate_release_failure(self):
...
         update = model.Update(
-            title=u'testmodule-master-2',
             builds=[build], user=user,
             status=UpdateStatus.testing,
             request=UpdateRequest.stable,
             type=UpdateType.enhancement,
             notes=u'Useful details!',
             test_gating_status=TestGatingStatus.passed)
         update.release = release
+        update.assign_alias()
```
sebwoj: Method `assign_alias` was removed.

bodhi/tests/server/test_models.py
```diff
@@ -3614,22 +3612,22 @@ def test_validate_release_success(self):
                                    package=package)
         self.db.add(build2)
         update1 = model.Update(
-            title=u'testmodule-master-2',
             builds=[build1], user=user,
             status=UpdateStatus.testing,
             request=UpdateRequest.stable,
             notes=u'Useful details!', release=release,
             test_gating_status=TestGatingStatus.passed)
+        update1.assign_alias()
```
sebwoj: Method `assign_alias` was removed.

bodhi/tests/server/test_models.py
```diff
@@ -3614,22 +3612,22 @@ def test_validate_release_success(self):
...
 
         # This should not raise an Exception.
         update2 = model.Update(
-            title=u'testmodule-master-2',
             builds=[build2], user=user,
             status=UpdateStatus.testing,
             request=UpdateRequest.stable,
             notes=u'Useful details!', release=release,
             test_gating_status=TestGatingStatus.passed)
+        update2.assign_alias()
```
sebwoj: Method `assign_alias` was removed.

bodhi/server/models.py
```diff
@@ -1631,10 +1627,8 @@ class Update(Base):
 
     __tablename__ = 'updates'
     __exclude_columns__ = ('id', 'user_id', 'release_id')
-    __include_extras__ = ('meets_testing_requirements', 'url',)
-    __get_by__ = ('title', 'alias')
```
sebwoj, line 1630: Removing `title` from this container will block access to updates via `/updates/{title}`(like: bodhi.fedoraproject.org/updates/phpdoc-2.9.0-7.fc29)  so maybe we should add note in release notes in B/I section?

sebwoj (changes requested):
It looks like this PR is in WIP state, but I've left just a few comments.

bodhi/server/models.py
bowlofeggs: Fixed, and thanks!

bodhi/tests/server/test_models.py
bowlofeggs: Yeah, I believe I've successfully removed this everywhere.

bodhi/server/models.py
bowlofeggs: Added!

bowlofeggs:
The ```f28-integration``` test failure is a timeout waiting on the DB container to start. I wrote #3054 to help with this, and I think we can safely ignore it here.

bowlofeggs:

bowlofeggs:

PR #3042, opened 2019-02-26
Add json response handler for server internal errors

### LINKED ISSUE #3035: Bodhi's update create form does not gracefully handle errors from the backend
When users were hitting https://github.com/fedora-infra/bodhi/issues/3024, the web form's submit button would spin until it hit the error and then it would stop spinning, but no error was reported to the user. This made it non-obvious that there was a problem, and also made it more difficult to determine what the real problem was.

The web form (and all of Bodhi's forms) should be able to display errors from the backend to the user on non-2xx codes.

sebwoj:
Fixes #3035

### DIFF
--- a/bodhi/server/static/js/forms.js
+++ b/bodhi/server/static/js/forms.js
@@ -70,15 +70,31 @@ Form.prototype.success = function(data) {
 Form.prototype.error = function(data) {
     var self = this;
     self.finish();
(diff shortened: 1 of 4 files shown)


bowlofeggs (approved):
Thanks!

bowlofeggs:

bowlofeggs:

PR #3043, opened 2019-02-26
Link search box results for packages to packages instead of users

### LINKED ISSUE #3040: The search box results for packages links to users
If you click the search box in the upper right hand corner and type "ejabberd", you will get some matches on packages and updates. If you click on the ejabberd package result, it will take you to ```/users/ejabberd```, when it should take you to a view that shows you updates that pertain to that package.

I believe it is due to this code:

https://github.com/fedora-infra/bodhi/blob/09dc4c3808f28eb76f0e3cea02567842f7e47624/bodhi/server/static/js/search.js#L51-L67

Note the ```hasOwnProperty('stack')``` in there. Stacks were removed in #2241, and I must have missed this when I made that change.

To fix it, we'll need the if statement to correctly identify packages so that the packages link is used instead of the user link later in the ```if/else``` block.

sebwoj:
Fixes #3040

### DIFF
--- a/bodhi/server/static/js/search.js
+++ b/bodhi/server/static/js/search.js
@@ -51,10 +51,10 @@ $(document).ready(function() {
     function resultUrl(data){
       if (data.alias != undefined) {
           return '/updates/' + data.alias;
-      } else if (data.hasOwnProperty('stack')) {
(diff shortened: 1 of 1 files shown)


bowlofeggs (approved):
Thanks!

PR #3045, opened 2019-02-26
Mark tests as ignored if Greenwave fails to respond

### LINKED ISSUE #3044: Bodhi should only block updates when Greenwave returns a NOGO response
Currently Bodhi will only allow updates to proceed to repositories if Greenwave gives back a GO response. This means that updates will be blocked from proceeding if Greenwave raises an Exception, which has been [happening often recently](https://pagure.io/greenwave/issue/388). Due to the frequent nature of these errors, it is most sensible to adjust Bodhi such that it will only block an update if Greenwave gives back a positive NOGO answer, and that any other answer (GO, error, etc.) means that the update can proceed.

bowlofeggs:
There are two commits in this pull request with more detail. The first marks Greenwave failures as "tests ignored" in Bodhi, and the second has release notes for a 3.13.3 with that patch.

fixes #3044

### DIFF
--- a/bodhi/server/models.py
+++ b/bodhi/server/models.py
@@ -38,6 +38,7 @@ from sqlalchemy.orm.exc import NoResultFound
 from sqlalchemy.orm.properties import RelationshipProperty
 from sqlalchemy.sql import text
 from sqlalchemy.types import SchemaType, TypeDecorator, Enum
+import requests.exceptions
 
 from bodhi.server import bugs, buildsys, log, mail, notifications, Session, util
 from bodhi.server.config import config
@@ -608,6 +609,7 @@ class TestGatingStatus(DeclEnum):
     running = 'running', 'Running'
     passed = 'passed', 'Passed'
     failed = 'failed', 'Failed'
+    greenwave_failed = 'greenwave_failed', 'Greenwave failed to respond'
 
 
 class UpdateType(DeclEnum):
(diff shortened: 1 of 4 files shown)


bodhi/server/models.py
```diff
@@ -1924,11 +1925,23 @@ def get_test_gating_info(self):
...
+            # policy satisfied. We will use the Exception as the summary so we can mark the status
+            # as ignored for the record.
+            decision = {'policies_satisfied': True, 'summary': str(e), 'greenwave_success': False}
         if decision['policies_satisfied']:
-            # If an unrestricted policy is applied and no tests are required
-            # on this update, let's set the test gating as ignored in Bodhi.
-            if decision['summary'] == 'no tests are required':
+            if not decision['greenwave_success']:
+                # Greenwave failed to respond, so let's mark this as ignored.
+                self.test_gating_status = TestGatingStatus.ignored
```
puiterwijk: Do we maybe want a new status for this, so people can distinguish between ignored due to Greenwave not configured vs `TestGatingStatus.GreenwaveFailed`?

puiterwijk (commented):
Looks fine to me, but one remark. If that gets answered (Either way), I'll ack.

bodhi/server/models.py
bowlofeggs: We could do that, but it would require a database migration. Do note that we are also recording the error message in the ```greenwave_summary_string``` below.

I don't mind adding the migration if you prefer that option. I agree that it would be cleaner. However, I won't be able to add that today (I can do it tomorrow, I just need to stop working due to evening plans.)

bowlofeggs: After thinking about it some more, I do like your suggestion of extending the status to cover this. It will help us be more explicit about what is happening than the summary string will be.

I will add that tomorrow and then we can proceed.

bowlofeggs: @puiterwijk alright, I have a migration in place now that adds a new status to track when Greenwave failed.

bowlofeggs: Oooh, but now I need to make sure that Bodhi does the right thing with this new state…

bowlofeggs: OK, I've got a local patch to make Bodhi care about the new state, but I need tests and I need to leave. I'll get that posted up tomorrow.

bowlofeggs:
This is ready for re-review.

bowlofeggs:
The ```rawhide-build``` failure is due to a GPG key issue in the Rawhide container, which I suspect is a result of the recent branching.

puiterwijk (approved):
LGTM

PR #3047, opened 2019-02-27
Correct the update command for EPEL

### LINKED ISSUE #3017: install instructions for EPEL releases should use yum instead of dnf
The addition of the install instructions (#2799) were deployed with 3.13.0.  They should use `yum` as the command rather than `dnf` for EPEL.  I believe if they use the `--advisory` option they would need to tell users that requires the yum security plugin. I don't use that option, so someone more familiar would need to look at the details.

mattiaverga:
Fixes #3017 

This will correct the command outputted for installing EPEL updates.
However, this fixes only actual cases (EPEL = yum; FEDORA < 22 = yum; FEDORA >= 22 = dnf).
A better and less Fedora-centric solution could be to modify the Release model adding a `pkg_manager` string property and a `testing_repo` string property, so that for every release it can be specified what package manager is in use and what name has the testing repository.
After that, in `update_install_command()` we can build command for the different package managers we want to support.
What do you think?

### DIFF
--- a/bodhi/server/templates/update.html
+++ b/bodhi/server/templates/update.html
@@ -680,10 +680,10 @@ $(document).ready(function(){
           % endif
 
           % if update.content_type:
-          % if update.content_type.value == 'rpm' and ('testing' in update.status or 'stable' in update.status):
+          % if update.status in [models.UpdateStatus.testing, models.UpdateStatus.stable] and not (update.release.package_manager == models.PackageManager.unspecified or update.release.testing_repository is None):
           <div class="p-t-1">
           <h4>How to install</h4>
-            <pre style="position: relative;"><span class="label label-default" id="copybutton" style="position: absolute; top: 0px; right: 0px; cursor: pointer;" title="Copy to clipboard"><i class="fa fa-copy"></i></span><code id="commandtext">${self.util.update_install_command(update)}</code></pre>
+            <pre style="position: relative;"><span class="label label-default" id="copybutton" style="position: absolute; top: 0px; right: 0px; cursor: pointer;" title="Copy to clipboard"><i class="fa fa-copy"></i></span><code id="commandtext">${update.install_command}</code></pre>
           </div>
           % endif
           % endif
--- a/bodhi/server/util.py
+++ b/bodhi/server/util.py
@@ -1487,30 +1487,3 @@ def get_absolute_path(location):
     module, final = location.split(':')
     base = os.path.dirname(__import__(module).__file__)
     return base + "/" + final
-
-
-def update_install_command(context, update):
-    """
-    Return the dnf command for installing the Update.
-
-    Args:
-        context (mako.runtime.Context): Unused.
-        update (bodhi.server.models.Update): The Update you want to install.
-    Returns:
-        basestring: The dnf command to install the Update.
-    Raises:
-        ValueError: When provided update is not in stable or testing state.
-    """
-    status = str(update.status)
-    alias = update.alias
-    update_type = str(update.type)
-
-    if status != 'stable' and status != 'testing':
-        raise ValueError('Only updates in stable or testing can be installed!')
-
-    command = 'sudo dnf {}{} --advisory={}{}'.format(
-        'install' if update_type == 'newpackage' else 'upgrade',
-        ' --enablerepo=updates-testing' if status == 'testing' else '',
-        alias,
-        r' \*' if update_type == 'newpackage' else '')
-    return command
--- a/bodhi/client/__init__.py
+++ b/bodhi/client/__init__.py
@@ -171,6 +171,10 @@ release_options = [
     click.option('--mail-template', help='Name of the email template for this release'),
     click.option('--composed-by-bodhi/--not-composed-by-bodhi', is_flag=True, default=True,
                  help='The flag that indicates whether the release is composed by Bodhi or not'),
+    click.option('--package-manager', type=click.Choice(['unspecified', 'dnf', 'yum']),
+                 help='The package manager used by this release'),
+    click.option('--testing-repository',
+                 help='The name of the testing repository used to test updates'),
     click.option(
         '--create-automatic-updates/--no-create-automatic-updates',
         help=('Configure for this release, whether or not automatic updates are '
@@ -1241,6 +1245,8 @@ def print_release(release):
     click.echo("  Email Template:           %s" % release['mail_template'])
     click.echo("  Composed by Bodhi:        %s" % release['composed_by_bodhi'])
     click.echo("  Create Automatic Updates: %s" % release['create_automatic_updates'])
+    click.echo("  Package Manager:          %s" % release['package_manager'])
+    click.echo("  Testing Repository:       %s" % release['testing_repository'])
 
 
 def print_errors(data):
--- a/bodhi/server/migrations/versions/5703ddfe855d_add_package_manager_and_testing_.py
+++ b/bodhi/server/migrations/versions/5703ddfe855d_add_package_manager_and_testing_.py
@@ -1,0 +1,60 @@
+# Copyright (c) 2019 Mattia Verga
+#
+# This file is part of Bodhi.
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+"""
+Add package_manager and testing_repository columns to the Release model.
+
+Revision ID: 5703ddfe855d
+Revises: e8a059156d38
+Create Date: 2019-05-05 07:05:03.434601
+"""
+from alembic import op
+import sqlalchemy as sa
+from sqlalchemy.dialects import postgresql
+
+
+# revision identifiers, used by Alembic.
+revision = '5703ddfe855d'
+down_revision = 'e8a059156d38'
+
+
+def upgrade():
+    """Add package_manager Enum type and testing_repository string type."""
+    package_manager = postgresql.ENUM('unspecified', 'dnf', 'yum', name='ck_package_manager')
+    package_manager.create(op.get_bind())
+    op.add_column('releases',
+                  sa.Column('package_manager',
+                            postgresql.ENUM('unspecified',
+                                            'dnf',
+                                            'yum',
+                                            name='ck_package_manager'),
+                            nullable=True,
+                            server_default='unspecified')
+                  )
+    op.add_column('releases',
+                  sa.Column('testing_repository',
+                            sa.UnicodeText(),
+                            nullable=True)
+                  )
+    op.alter_column('releases', 'package_manager', server_default=None)
+
+
+def downgrade():
+    """Drop package_manager and testing_repository."""
+    op.drop_column('releases', 'testing_repository')
+    op.drop_column('releases', 'package_manager')
+    op.execute("DROP TYPE ck_package_manager")
--- a/bodhi/server/models.py
+++ b/bodhi/server/models.py
@@ -714,6 +714,21 @@ class ComposeState(DeclEnum):
     cleaning = 'cleaning', 'Cleaning old composes'
 
 
+class PackageManager(DeclEnum):
+    """
+    An enum used to specify what package manager is used by a specific Release.
+
+    Attributes:
+        unspecified (EnumSymbol): for releases where the package manager is not specified.
+        dnf (EnumSymbol): DNF package manager.
+        yum (EnumSymbol): YUM package manager.
+    """
+
+    unspecified = 'unspecified', 'package manager not specified'
+    dnf = 'dnf', 'dnf'
+    yum = 'yum', 'yum'
+
+
 ##
 #  Association tables
 ##
@@ -761,6 +776,10 @@ class Release(Base):
         create_automatic_updates (bool): A flag indicating that updates should
             be created automatically for Koji builds tagged into the
             `candidate_tag`. Defaults to False.
+        package_manager (EnumSymbol): The package manager this release uses. This must be one of
+            the values defined in :class:`PackageManager`.
+        testing_repository (unicode): The name of repository where updates are placed for
+            testing before being pushed to the main repository.
     """
 
     __tablename__ = 'releases'
@@ -788,6 +807,9 @@ class Release(Base):
     create_automatic_updates = Column(Boolean, default=False)
 
     _version_int_regex = re.compile(r'\D+(\d+)[CMF]?$')
+
+    package_manager = Column(PackageManager.db_type(), default=PackageManager.unspecified)
+    testing_repository = Column(UnicodeText, nullable=True)
 
     @property
     def version_int(self):
@@ -1916,6 +1938,32 @@ class Update(Base):
         api_url = '{}/decision'.format(config.get('greenwave_api_url'))
 
         return util.greenwave_api_post(api_url, data)
+
+    def install_command(self) -> str:
+        """
+        Return the appropriate command for installing the Update.
+
+        Returns:
+            The dnf command to install the Update.
+        Raises:
+            ValueError: When the update is not in stable or testing state, or when
+                the package manager or the testing repository are not known.
+        """
+        if self.status != UpdateStatus.stable and self.status != UpdateStatus.testing:
+            raise ValueError('Only updates in stable or testing can be installed!')
+
+        if self.release.package_manager == PackageManager.unspecified \
+                or self.release.testing_repository is None:
+            raise ValueError('We don\'t know the package manager or the testing repository!')
+
+        command = 'sudo {} {}{} --advisory={}{}'.format(
+            self.release.package_manager.value,
+            'install' if self.type == UpdateType.newpackage else 'upgrade',
+            (' --enablerepo=' + self.release.testing_repository)
+            if self.status == UpdateStatus.testing else '',
+            self.alias,
+            r' \*' if self.type == UpdateType.newpackage else '')
+        return command
 
     def update_test_gating_status(self):
         """Query Greenwave about this update and set the test_gating_status as appropriate."""
--- a/bodhi/server/schemas.py
+++ b/bodhi/server/schemas.py
@@ -24,6 +24,7 @@ from bodhi.server import util
 from bodhi.server.config import config
 from bodhi.server.models import (
     ContentType,
+    PackageManager,
     ReleaseState,
     UpdateRequest,
     UpdateSeverity,
@@ -399,6 +400,15 @@ class SaveReleaseSchema(CSRFProtectedSchema, colander.MappingSchema):
     create_automatic_updates = colander.SchemaNode(
         colander.Boolean(true_choices=('true', '1')),
         missing=False,
+    )
+    package_manager = colander.SchemaNode(
+        colander.String(),
+        validator=colander.OneOf(list(PackageManager.values())),
+        missing="unspecified",
+    )
+    testing_repository = colander.SchemaNode(
+        colander.String(),
+        missing=None,
     )
 
 
--- a/bodhi/server/validators.py
+++ b/bodhi/server/validators.py
@@ -36,6 +36,7 @@ from .models import (
     ContentType,
     Group,
     Package,
+    PackageManager,
     Release,
     RpmBuild,
     ReleaseState,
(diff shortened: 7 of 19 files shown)


bodhi/server/util.py
```diff
@@ -1453,15 +1453,18 @@ def update_install_command(context, update):
...
 
-    command = 'sudo dnf {}{} --advisory={}{}'.format(
+    command = 'sudo {} {}{} --advisory={}{}'.format(
+        'dnf' if (update.release.id_prefix == 'FEDORA'
+                  and update.release.version_int >= 22) else 'yum',
         'install' if update_type == 'newpackage' else 'upgrade',
-        ' --enablerepo=updates-testing' if status == 'testing' else '',
-        alias,
+        (' --enablerepo=' + testing_repo) if status == 'testing' else '',
+        update.alias,
```
bowlofeggs: I think a more general command would be useful, and I like your suggestion of adding a string to the Release model. Let's try that route instead. Sound good to you?

bowlofeggs (changes requested):

bodhi/server/util.py
mattiaverga: I can think two way of doing this.
In the first one I can add a string and an Enum to the Release model. The string will carry the testing repository name, while the Enum will permit the choice of the package manager used by that release (currently 'yum' or 'dnf').

The second way is to create a new PackageManager model to handle all package managers stuff and reference it in the Release model.
For example:
```
class PackageManager(Base):
    """
    This model represent a package manager supported by Bodhi.

    In command and options strings the following placeholders are available:
    - {update}: the update alias
    - {options}: command options, like option_enable_testing
    - {testing_repository}: the testing repository name associated to the release

    Attributes:
        name (unicode): The name of the Package Manager.
        install_command (unicode): The string used to install a 'newpackage' update.
        upgrade_command (unicode): The string used to install the update.
        option_enable_testing (unicode): The string used to enable the testing repository.
    """

    __tablename__ = 'package_managers'
    __get_by__ = ('name',)
    __exclude_columns__ = ('id',)

    name = Column(UnicodeText, unique=True, nullable=False)
    install_command = Column(UnicodeText, nullable=False,
                             default='sudo dnf install {options} --advisory={update} \*')
    upgrade_command = Column(UnicodeText, nullable=False,
                             default='sudo dnf upgrade {options} --advisory={update}')
    option_enable_testing = Column(UnicodeText, nullable=False,
                                   default='--enablerepo={testing_repository}')

    release = relationship('Release', backref='package_managers', lazy=False)
```
and add in the Release model
```
package_manager_id = Column(Integer, ForeignKey('package_managers.id'))
testing_repository = Column(UnicodeText, nullable=False)
```

What would you prefer?

bowlofeggs:
On Mon, 2019-03-18 at 09:44 -0700, Mattia Verga wrote:
> What would you prefer?

Hey Mattia!

I think the Enum would be preferable over the extra model, since it
avoids a join. However, I think it would also be OK to just let it be a
text field since it's really just a setting. I don't mind it being an
Enum if you like the strict error checking that an Enum provides.

mattiaverga:
Is there a way to test Release creation from the client in Vagrant? It always returns 'invalid credentials'.

bowlofeggs:
On Sat, 2019-03-23 at 10:19 -0700, Mattia Verga wrote:
> Is there a way to test Release creation from the client in Vagrant?
> It always returns 'invalid credentials'.

Bodhi is expecting you to be in one of the groups in the admin_groups
setting. You can put yourself in there by adding a group you are in to
that setting in your development.ini.

mattiaverga:
> Bodhi is expecting you to be in one of the groups in the admin_groups setting. You can put yourself in there by adding a group you are in to that setting in your development.ini.

Ah, I thought that adding my username in `acl_dummy_committer` was enough to give me all administrator privileges.

bodhi/server/migrations/versions/e383591ed939_add_package_manager_and_testing_.py
```diff
@@ -0,0 +1,64 @@
...
+    op.add_column('releases', sa.Column('testing_repository',
+                                        sa.UnicodeText(),
+                                        nullable=False,
+                                        server_default='updates-testing'))
+    op.alter_column('releases', 'package_manager', server_default=None)
+    op.alter_column('releases', 'testing_repository', server_default=None)
+    op.execute("UPDATE releases SET package_manager = 'yum' \
+        WHERE id_prefix != 'FEDORA'")
+    op.execute("UPDATE releases SET testing_repository = 'epel-testing' \
+        WHERE id_prefix != 'FEDORA'")
```
bowlofeggs: There are Bodhi installations in the wild other than Fedora's. We should probably not do it this way because it will set all their releases to yum and epel-testing. Might be better to do an exact match. Alternatively, we can leave the existing releases alone and just document it as an upgrade step for installations that want this behavior. I think I'd lean towards that latter option.

bodhi/server/migrations/versions/e383591ed939_add_package_manager_and_testing_.py
```diff
@@ -0,0 +1,64 @@
...
+    """Add package_manager Enum type and testing_repository string type.
+
+    Also set the right values for items already in database.
+    """
+    package_manager = postgresql.ENUM('dnf', 'yum', name='ck_package_manager')
+    package_manager.create(op.get_bind())
+    op.add_column('releases',
+                  sa.Column('package_manager',
+                            postgresql.ENUM('dnf', 'yum', name='ck_package_manager'),
+                            nullable=False,
```
bowlofeggs: There are releases in Bodhi that do not use dnf or yum, such as flatpaks and containers. We might want to allow this to be nullable since we don't have support for update instructions on those releases yet. We could allow null to mean that the release is not supported.

Or another idea is to call this a "yum_package_manager" enum, and make a YumRelease model and have that model be the only one that has this field. That would be a lot more work though…

bodhi/server/migrations/versions/e383591ed939_add_package_manager_and_testing_.py
```diff
@@ -0,0 +1,64 @@
...
+                            server_default='dnf')
+                  )
+    op.add_column('releases', sa.Column('testing_repository',
+                                        sa.UnicodeText(),
+                                        nullable=False,
```
bowlofeggs: Similar issues here - we probably don't want to set the container and flatpak testing_repositories to updates-testing.

bowlofeggs (changes requested):
It would be wise to consider a more flexible design that would allow us to support flatpaks and containers in the future as well. This design seems to mostly assume that you can use the yum/dnf style update command, which would not be true for the other content types.

bodhi/server/migrations/versions/e383591ed939_add_package_manager_and_testing_.py
mattiaverga: I have added an 'unspecified' value as default. I think is more logical to avoid having an Enum default to Null, let me know if it's a problem.

bodhi/server/templates/update.html
```diff
@@ -680,7 +680,7 @@ <h4>Reboot Required</h4>
           % endif
 
           % if update.content_type:
-          % if update.content_type.value == 'rpm' and ('testing' in update.status or 'stable' in update.status):
+          % if update.content_type.value == 'rpm' and ('testing' in update.status or 'stable' in update.status) and not (update.release.package_manager.value == 'unspecified' or update.release.testing_repository is None):
```
mattiaverga: Maybe we can avoid the double test on `content_type`, since we now have the check on `package-manager` and `testing-repository`.

bowlofeggs: Yeah I think we could avoid it, especially since we've ensured that it's dnf, yum, or unspecified. We can leave it up to admins to only set this where it makes sense.

Also, we should compare Enums directly rather than comparing the string values.

bodhi/server/util.py
```diff
@@ -1490,26 +1490,32 @@ def get_absolute_path(location):
...
     """
     status = str(update.status)
-    alias = update.alias
     update_type = str(update.type)
 
     if status != 'stable' and status != 'testing':
         raise ValueError('Only updates in stable or testing can be installed!')
 
-    command = 'sudo dnf {}{} --advisory={}{}'.format(
+    if update.release.package_manager.value == 'unspecified' \
```
bowlofeggs: We should compare the Enum values directly.

bodhi/server/util.py
```diff
@@ -1490,26 +1490,32 @@ def get_absolute_path(location):
...
+
+    command = 'sudo {} {}{} --advisory={}{}'.format(
+        update.release.package_manager.value,
         'install' if update_type == 'newpackage' else 'upgrade',
-        ' --enablerepo=updates-testing' if status == 'testing' else '',
-        alias,
+        (' --enablerepo=' + update.release.testing_repository) if status == 'testing' else '',
+        update.alias,
         r' \*' if update_type == 'newpackage' else '')
+
```
bowlofeggs: This is an unrelated style change.

bowlofeggs (changes requested):
Just a few adjustments and a rebase to fix some conflicts. I'll take care of that for you.

bowlofeggs:
While trying to fix the enum issues I mentioned, I realized that it is not easy to compare enums in util.py, since models.py imports util.py, which means util.py should not import models.py, which it needs to do to compare enums…

Then I realized that the ```update_install_command()``` would make a lot of sense to go on the ```Update``` model, which would put it in models.py, which solves the problem!

So I'm going to make that change as well.

bowlofeggs:
Another refactor we could do is to make the install_command() function just return the empty string if it can't do it instead of raising an Exception. This way the template code doesn't need to reproduce the logic and can just show it if it isn't empty string. I might send that as a follow up PR.

bowlofeggs:
OK, here's a patch I'm applying to move it over to the model and use Enums:

```
index 9afc2a94..a9c0afd0 100644
--- a/bodhi/server/models.py
+++ b/bodhi/server/models.py
@@ -1939,6 +1939,32 @@ class Update(Base):
 
         return util.greenwave_api_post(api_url, data)
 
+    def install_command(self) -> str:
+        """
+        Return the appropriate command for installing the Update.
+
+        Returns:
+            The dnf command to install the Update.
+        Raises:
+            ValueError: When the update is not in stable or testing state, or when
+                the package manager or the testing repository are not known.
+        """
+        if self.status != UpdateStatus.stable and self.status != UpdateStatus.testing:
+            raise ValueError('Only updates in stable or testing can be installed!')
+
+        if self.release.package_manager == PackageManager.unspecified \
+                or self.release.testing_repository is None:
+            raise ValueError('We don\'t know the package manager or the testing repository!')
+
+        command = 'sudo {} {}{} --advisory={}{}'.format(
+            self.release.package_manager.value,
+            'install' if self.type == UpdateType.newpackage else 'upgrade',
+            (' --enablerepo=' + self.release.testing_repository)
+            if self.status == UpdateStatus.testing else '',
+            self.alias,
+            r' \*' if self.type == UpdateType.newpackage else '')
+        return command
+
     def update_test_gating_status(self):
         """Query Greenwave about this update and set the test_gating_status as appropriate."""
         try:
diff --git a/bodhi/server/templates/update.html b/bodhi/server/templates/update.html
index 94ded553..54c69635 100644
--- a/bodhi/server/templates/update.html
+++ b/bodhi/server/templates/update.html
@@ -683,7 +683,7 @@ $(document).ready(function(){
           % if update.status in [models.UpdateStatus.testing, models.UpdateStatus.stable] and not (update.release.package_manager == models.PackageManager.unspecified or update.release.testing_repository is None):
           <div class="p-t-1">
           <h4>How to install</h4>
-            <pre style="position: relative;"><span class="label label-default" id="copybutton" style="position: absolute; top: 0px; right: 0px; cursor: pointer;" title="Copy to clipboard"><i class="fa fa-copy"></i></span><code id="commandtext">${self.util.update_install_command(update)}</code></pre>
+            <pre style="position: relative;"><span class="label label-default" id="copybutton" style="position: absolute; top: 0px; right: 0px; cursor: pointer;" title="Copy to clipboard"><i class="fa fa-copy"></i></span><code id="commandtext">${update.install_command}</code></pre>
           </div>
           % endif
           % endif
index 3fd1da8a..f419b9e3 100644
--- a/bodhi/server/util.py
+++ b/bodhi/server/util.py
@@ -1487,35 +1487,3 @@ def get_absolute_path(location):
     module, final = location.split(':')
     base = os.path.dirname(__import__(module).__file__)
     return base + "/" + final
-
-
-def update_install_command(context, update):
-    """
-    Return the appropriate command for installing the Update.
-
-    Args:
-        context (mako.runtime.Context): Unused.
-        update (bodhi.server.models.Update): The Update you want to install.
-    Returns:
-        basestring: The dnf command to install the Update.
-    Raises:
-        ValueError: When provided update is not in stable or testing state, or when
-        the package manager or the testing repository are not known.
-    """
-    status = str(update.status)
-    update_type = str(update.type)
-
-    if status != 'stable' and status != 'testing':
-        raise ValueError('Only updates in stable or testing can be installed!')
-
-    if update.release.package_manager == models.PackageManager.unspecified \
-            or update.release.testing_repository is None:
-        raise ValueError('We don\'t know the package manager or the testing repository!')
-
-    command = 'sudo {} {}{} --advisory={}{}'.format(
-        update.release.package_manager.value,
-        'install' if update_type == 'newpackage' else 'upgrade',
-        (' --enablerepo=' + update.release.testing_repository) if status == 'testing' else '',
-        update.alias,
-        r' \*' if update_type == 'newpackage' else '')
-    return command
diff --git a/bodhi/tests/server/test_models.py b/bodhi/tests/server/test_models.py
index 97ca4896..182af5f9 100644
--- a/bodhi/tests/server/test_models.py
+++ b/bodhi/tests/server/test_models.py
@@ -1648,6 +1648,96 @@ class TestUpdateGetBugKarma(BaseTestCase):
         self.assertEqual(good, 0)
 
 
+class TestUpdateInstallCommand(BaseTestCase):
+    """Test the update_install_command() function."""
+
+    def test_upgrade_in_testing(self):
+        """Update is an enhancement, a security or a bugfix and is in testing."""
+        update = model.Update.query.first()
+        update.status = UpdateStatus.testing
+        update.type = UpdateType.bugfix
+        update.release.package_manager = PackageManager.dnf
+        update.release.testing_repository = 'updates-testing'
+
+        command = update.install_command()
+
+        self.assertEqual(
+            command,
+            'sudo dnf upgrade --enablerepo=updates-testing --advisory={}'.format(update.alias))
+
+    def test_upgrade_in_stable(self):
+        """Update is an enhancement, a security or a bugfix and is in stable."""
+        update = model.Update.query.first()
+        update.status = UpdateStatus.stable
+        update.type = UpdateType.bugfix
+        update.release.package_manager = PackageManager.dnf
+        update.release.testing_repository = 'updates-testing'
+
+        command = update.install_command()
+
+        self.assertEqual(
+            command,
+            'sudo dnf upgrade --advisory={}'.format(update.alias))
+
+    def test_newpackage_in_testing(self):
+        """Update is a newpackage and is in testing."""
+        update = model.Update.query.first()
+        update.status = UpdateStatus.testing
+        update.type = UpdateType.newpackage
+        update.release.package_manager = PackageManager.dnf
+        update.release.testing_repository = 'updates-testing'
+
+        command = update.install_command()
+
+        self.assertEqual(
+            command,
+            r'sudo dnf install --enablerepo=updates-testing --advisory={} \*'.format(update.alias))
+
+    def test_newpackage_in_stable(self):
+        """Update is a newpackage and is in stable."""
+        update = model.Update.query.first()
+        update.status = UpdateStatus.stable
+        update.type = UpdateType.newpackage
+        update.release.package_manager = PackageManager.dnf
+        update.release.testing_repository = 'updates-testing'
+
+        command = update.install_command()
+
+        self.assertEqual(
+            command,
+            r'sudo dnf install --advisory={} \*'.format(update.alias))
+
+    def test_cannot_install(self):
+        """Update is out of stable or testing repositories."""
+        update = model.Update.query.first()
+        update.status = UpdateStatus.obsolete
+
+        with self.assertRaises(ValueError):
+            update.install_command()
+
+    def test_update_command_not_possible_missing_packagemanager(self):
+        """The Release of the Update misses the package manager definition."""
+        update = model.Update.query.first()
+        update.status = UpdateStatus.stable
+        update.type = UpdateType.newpackage
+        update.release.package_manager = PackageManager.unspecified
+        update.release.testing_repository = 'updates-testing'
+
+        with self.assertRaises(ValueError):
+            update.install_command()
+
+    def test_update_command_not_possible_missing_repo(self):
+        """The Release of the Update misses the testing repository definition."""
+        update = model.Update.query.first()
+        update.status = UpdateStatus.stable
+        update.type = UpdateType.newpackage
+        update.release.package_manager = PackageManager.dnf
+        update.release.testing_repository = None
+
+        with self.assertRaises(ValueError):
+            update.install_command()
+
+
 class TestUpdateGetTestcaseKarma(BaseTestCase):
     """Test the get_testcase_karma() method."""
 

diff --git a/bodhi/tests/server/test_util.py b/bodhi/tests/server/test_util.py
index ea90cca4..000fa7f6 100644
--- a/bodhi/tests/server/test_util.py
+++ b/bodhi/tests/server/test_util.py
@@ -25,9 +25,7 @@ import unittest
 
 from bodhi.server import util, models
-from bodhi.server.models import (
-    ComposeState, PackageManager, TestGatingStatus, Update, UpdateStatus,
-    UpdateType)
+from bodhi.server.models import ComposeState, TestGatingStatus, Update
 from bodhi.tests.server import base
 
 
@@ -1426,100 +1424,3 @@ class TestTransactionalSessionMaker(base.BaseTestCase):
         Session.return_value.commit.assert_called_once_with()
         Session.return_value.close.assert_called_once_with()
         Session.remove.assert_called_once_with()
-
-
-class TestUpdateInstallCommand(base.BaseTestCase):
-    """Test the update_install_command() function."""
-
-    def test_upgrade_in_testing(self):
-        """Update is an enhancement, a security or a bugfix and is in testing."""
-        context = {'request': mock.MagicMock()}
-        update = models.Update.query.first()
-        update.status = UpdateStatus.testing
-        update.type = UpdateType.bugfix
-        update.release.package_manager = PackageManager.dnf
-        update.release.testing_repository = 'updates-testing'
-
-        command = util.update_install_command(context, update)
-
-        self.assertEqual(
-            command,
-            'sudo dnf upgrade --enablerepo=updates-testing --advisory={}'.format(update.alias))
-
-    def test_upgrade_in_stable(self):
-        """Update is an enhancement, a security or a bugfix and is in stable."""
-        context = {'request': mock.MagicMock()}
-        update = models.Update.query.first()
-        update.status = UpdateStatus.stable
-        update.type = UpdateType.bugfix
-        update.release.package_manager = PackageManager.dnf
-        update.release.testing_repository = 'updates-testing'
-
-        command = util.update_install_command(context, update)
-
-        self.assertEqual(
-            command,
-            'sudo dnf upgrade --advisory={}'.format(update.alias))
-
-    def test_newpackage_in_testing(self):
-        """Update is a newpackage and is in testing."""
-        context = {'request': mock.MagicMock()}
-        update = models.Update.query.first()
-        update.status = UpdateStatus.testing
-        update.type = UpdateType.newpackage
-        update.release.package_manager = PackageManager.dnf
-        update.release.testing_repository = 'updates-testing'
-
-        command = util.update_install_command(context, update)
-
-        self.assertEqual(
-            command,
-            r'sudo dnf install --enablerepo=updates-testing --advisory={} \*'.format(update.alias))
-
-    def test_newpackage_in_stable(self):
-        """Update is a newpackage and is in stable."""
-        context = {'request': mock.MagicMock()}
-        update = models.Update.query.first()
-        update.status = UpdateStatus.stable
-        update.type = UpdateType.newpackage
-        update.release.package_manager = PackageManager.dnf
-        update.release.testing_repository = 'updates-testing'
-
-        command = util.update_install_command(context, update)
-
-        self.assertEqual(
-            command,
-            r'sudo dnf install --advisory={} \*'.format(update.alias))
-
-    def test_cannot_install(self):
-        """Update is out of stable or testing repositories."""
-        context = {'request': mock.MagicMock()}
-        update = models.Update.query.first()
-        update.status = UpdateStatus.obsolete
-
-        with self.assertRaises(ValueError):
-            util.update_install_command(context, update)
-
-    def test_update_command_not_possible_missing_packagemanager(self):
-        """The Release of the Update misses the package manager definition."""
-        context = {'request': mock.MagicMock()}
-        update = models.Update.query.first()
-        update.status = UpdateStatus.stable
-        update.type = UpdateType.newpackage
-        update.release.package_manager = PackageManager.unspecified
-        update.release.testing_repository = 'updates-testing'
-
-        with self.assertRaises(ValueError):
-            util.update_install_command(context, update)
-
-    def test_update_command_not_possible_missing_repo(self):
-        """The Release of the Update misses the testing repository definition."""
-        context = {'request': mock.MagicMock()}
-        update = models.Update.query.first()
-        update.status = UpdateStatus.stable
-        update.type = UpdateType.newpackage
-        update.release.package_manager = PackageManager.dnf
-        update.release.testing_repository = None
-
-        with self.assertRaises(ValueError):
-            util.update_install_command(context, update)
```

bowlofeggs (approved):
Thanks!

bowlofeggs:
I opened https://github.com/fedora-infra/bodhi/pull/3308 as a bit of a simplification refactor to implement the idea I mentioned earlier here.

PR #3048, opened 2019-02-27
Reshuffle the greenwave_failed migration.

bowlofeggs:
I accidentally put the migration that adds the greenwave_failed
value to the TestGatingStatus enum at the end of the list of
migrations on the develop branch. This was problematic when I
tried to backport the patch to the 3.13 branch, which does not
have the referenced down_revision, which is a migration that will
be part of the upcoming 4.0 release.

This commit reshuffles the order of migrations so that the new
migrations is inserted at the end of the 3.13 branch head, and
just before the new migrations in 4.0.


bowlofeggs:
I am going to manually cherry pick this into #3046, instead of asking Mergify to do it, because it cannot be cherry picked in its own PR, and that PR won't pass tests without it.

cverna:
+1

PR #3049, opened 2019-02-28
Store critical path packages in bodhi.

cverna:
This PR rebases, updates and finishes #2476. Once this is accepted an merged a new PR will be need to drop bodhi's PDC dependency.


cverna:
jenkies test

bodhi/client/__init__.py
```diff
@@ -1229,6 +1229,90 @@ def print_release(release):
...
+@url_option
+@staging_option
+def set_critpath_packages(user, password, release, packages, url, staging, **kwargs):
+    # User Docs that show in the --help
+    """
+    Set a list of critical path packages to a release.
+
+    RELEASE: Release name (e.g. F27)
+
+    [PACKAGES]: comma separated list of packages to add (e.g. firefox,vim,python)
```
bowlofeggs, line 1287: How many critpath packages do you think we have, approximately? I wonder if this will be a burdensome way to add/remove them. I also wonder if we will hit the max length of shell arguments?

The alternative might also be burdensome, however, which would be to have add/remove commands. Unless maybe the add/remove commands still at least support comma separated lists like is done here?

I guess my main concern is whether we will hit the shell limits or not. I don't have a super convenient way to find out which packages are critical path. Bodhi's database has a bool on whether an *update* is critical path, but that's not a direct way to answer the question. We could probably query PDC to find out, though I think it's data is also not up to date.

bowlofeggs: I also recognize that this is how the releng script works.

One option we could consider is giving the bindings the ability to set critpath packages, like this here, and then just use that in the releng script. We could leave this code for other Bodhi deployments to use, if they want. Then we at least know that Fedora won't hit shell limits.

bodhi/client/bindings.py
```diff
@@ -929,6 +929,64 @@ def get_releases(self, **kwargs):
...
 
+    @errorhandled
+    def get_critpath_packages(self, release):
+        """
+        Get a list of critical path packages for a release.
+
+        Args:
+            release (unicode): The release name to query for.
+        Returns:
+            dict: A dictionary returned by critpath_package_requests() method.
```
bowlofeggs: I haven't documented this yet (it's on my todo list), but I plan to change Bodhi's contributor docs to formally state that we should use type hinting instead of this doc style for documenting types. It's OK to leave this here since I haven't actually done that yet, I'm more letting you know for future patches in Bodhi. Hopefully I'll find the time to write that up soon. (Really, this is fine, and it'll be useful when I come through later and add type hinting ☺)

bodhi/client/bindings.py
```diff
@@ -929,6 +929,64 @@ def get_releases(self, **kwargs):
...
+                            'GET': list packages
+                            'POST': set packages
+            auth (bool): Whether to authenticate for send_request.
+            data (dict or unicode): data to be sent along the request.
+        Returns:
+            dict: A dictionary with a single key, packages, mapping to a list of Package
+                  objects that are in critical path for this Release.
+        """
+        if verb not in ('GET', 'POST'):
+            raise BodhiClientException("Unsupported request type.")
```
bowlofeggs: We could also use a ```ValueError``` here instead.

bodhi/client/bindings.py
```diff
@@ -929,6 +929,64 @@ def get_releases(self, **kwargs):
...
+        return self.send_request(
+            'releases/{}/critpath'.format(release), verb=verb, auth=auth, data=data
```
bowlofeggs: We can use f-strings, now that we are finally Python 3 only!

```suggestion
            f'releases/{release}/critpath', verb=verb, auth=auth, data=data
```

bodhi/server/migrations/versions/a96dadf085c2_add_critcal_path_packages.py
```diff
@@ -0,0 +1,47 @@
...
+Revises: 8c4d6aad9b78
+Create Date: 2019-02-28 10:14:05.792938
+"""
+from alembic import op
+import sqlalchemy as sa
+
+
+# revision identifiers, used by Alembic.
+revision = 'a96dadf085c2'
+down_revision = '8c4d6aad9b78'
```
bowlofeggs: This likely needs to be updated, as I merged some commits today with migrations. As of the time of this writing, the head is ```58b7919b942c```.

```suggestion
down_revision = '58b7919b942c'
```

bodhi/server/migrations/versions/a96dadf085c2_add_critcal_path_packages.py
```diff
@@ -0,0 +1,47 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+"""
+add_critcal_path_packages.
+
+Revision ID: a96dadf085c2
+Revises: 8c4d6aad9b78
```
bowlofeggs:
```suggestion
Revises: 58b7919b942c
```

bodhi/server/migrations/versions/a96dadf085c2_add_critcal_path_packages.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+def upgrade():
+    """Add relationship to store the critical path packages."""
+    op.create_table(
+        'critical_path_packages_table',
+        sa.Column('package_id', sa.Integer(), nullable=False),
+        sa.Column('release_id', sa.Integer(), nullable=False),
+        sa.ForeignKeyConstraint(['package_id'], ['packages.id'], ),
+        sa.ForeignKeyConstraint(['release_id'], ['releases.id'], )
```
bowlofeggs: It would be a good idea to add an index to these columns - PostgreSQL does not automatically index foreign key constraints.

bodhi/server/models.py
```diff
@@ -746,6 +746,12 @@ class ComposeState(DeclEnum):
     Column('user_id', Integer, ForeignKey('users.id')),
     Column('package_id', Integer, ForeignKey('packages.id')))
 
+critical_path_packages_table = Table(
+    'critical_path_packages_table', Base.metadata,
+    Column('package_id', Integer, ForeignKey('packages.id'), nullable=False),
+    Column('release_id', Integer, ForeignKey('releases.id'), nullable=False)
```
bowlofeggs: Also here - good idea to add indices.

In general, I think a lot of Bodhi's slowness issues could be solved if we did an analysis on the queries we run and added indices. It's been on my todo list for like 3 years.

bodhi/client/__init__.py
bowlofeggs: All of the above to say, I'm fine with doing it this way as long as we have the releng script use the bindings instead of shelling out to this, or as long as we think the list is short enough that we don't have to worry. Either way, I'm happy to merge this ☺

bodhi/server/validators.py
```diff
@@ -585,12 +586,24 @@ def validate_packages(request, **kwargs):
...
+    # Check if bad_packages are in koji
+    # If yes, create new package in Bodhi
+    koji_session = buildsys.get_session()
+
+    for p in bad_packages:
+        try:
+            koji_session.listPackages(pkgID=p)
+            package = Package(name=p)
+            validated_packages.append(package)
+        except Exception:
```
bowlofeggs: Let's catch the expected Exception type here instead of all Exceptions.

bodhi/server/validators.py
```diff
@@ -585,12 +586,24 @@ def validate_packages(request, **kwargs):
...
+            request.errors.add(
+                'querystring',
+                'packages',
+                "Invalid packages specified: {}".format(", ".join(bad_packages))
```
bowlofeggs:
```suggestion
                f"Invalid packages specified: {", ".join(bad_packages)}"
```

bodhi/tests/server/services/test_releases.py
```diff
@@ -22,7 +22,9 @@
 
 from bodhi import server
 from bodhi.server.config import config
-from bodhi.server.models import Release, ReleaseState, Update, UpdateStatus
+
+from bodhi.server.models import Release, ReleaseState, Update, UpdateStatus, Package, ContentType
+
```
bowlofeggs: For PEP-8:

```suggestion
```

bodhi/tests/server/services/test_releases.py
```diff
@@ -22,7 +22,9 @@
 
 from bodhi import server
 from bodhi.server.config import config
-from bodhi.server.models import Release, ReleaseState, Update, UpdateStatus
+
```
bowlofeggs: PEP-8:

```suggestion
```

bowlofeggs (changes requested):

bodhi/client/__init__.py
cverna: So looking at pdc there are 756 critical path packages for f29 (https://pdc.fedoraproject.org/rest_api/v1/component-branches/?active=true&critical_path=true&name=f29). So yes passing this into the shell might not be super smart :)

The current releng script takes a .txt file as an entry, what do you think about doing the same for the bodhi client ?

bowlofeggs (changes requested):
There are some pydocstyle and unit test failures.

The pydocstyle failures are these:

```
 bodhi/client/__init__.py:1222 in public function `print_critpath_packages`:
         D202: No blank lines allowed after function docstring (found 1)
 bodhi/server/migrations/versions/a429f263fc3f_add_critical_path_packages.py:1 at module level:
         D400: First line should end with a period (not 's')
 bodhi/server/migrations/versions/a429f263fc3f_add_critical_path_packages.py:34 in public function `upgrade`:
         D103: Missing docstring in public function
 bodhi/server/migrations/versions/a429f263fc3f_add_critical_path_packages.py:56 in public function `downgrade`:
         D103: Missing docstring in public function
      f28-build                      :  SKIPPED
      f28-pydocstyle                 :  FAILED    [0:00:24.480490] (exited with code: 1)
```

The unit test failures seem to be due to https://github.com/fedora-infra/python-fedora/pull/213 not being available for the code to use.

bowlofeggs:
Closing due to inactivity.

Issue #2966, opened 2019-01-27
Bodhi metrics page is broken


mattiaverga:
Both in staging and production, visiting the Bodhi metrics page throws a 500 Error:
https://bodhi.fedoraproject.org/metrics
`'NoneType' object has no attribute 'groups'`

igor-raits:
@bowlofeggs when could one expect this to be deployed in fedora?

bowlofeggs:
@ignatenkobrain I expect to start working on making a Bodhi 4.0.0 beta this week. Once it is deployed to staging, it will probably take 2-4 weeks to make it to prod (estimate).

PR #3050, opened 2019-03-03
Extend integration tests

sebwoj:
This commit adds a few new tests to integration tests. Such tests will prevent
issues like #2966. This will increase overall integration test duration, but will
also increase safety.

### DIFF
--- a/devel/ci/integration/tests/test_bodhi.py
+++ b/devel/ci/integration/tests/test_bodhi.py
@@ -1,4 +1,4 @@
-# Copyright © 2018 Red Hat, Inc.
+# Copyright © 2018-2019 Red Hat, Inc.
 #
 # This file is part of Bodhi.
 #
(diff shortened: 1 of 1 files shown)


sebwoj:
@bowlofeggs I'm not sure about the direction of this PR. My plan is to add new integration tests like:
- GET on every release in DB
- GET on updates query service
- EDIT on update
- CREATE update with bug(test integration with bugzilla if possible)
ans so on.

The list of test cases can be long, so first I would like to know what do you think.

bowlofeggs:
+1, we definitely want to add a bunch of integration tests. To get write working, we need to set up an Ipsilon server in our CI suite. @puiterwijk told me that there's a way to configure Ipsilon to have a test user that sounded really easy and handy. So if we configure it that way, it should be pretty straightforward to start testing things like creating updates and editing updates. That would be very nice to have!

devel/ci/integration/tests/test_bodhi.py
```diff
@@ -16,16 +16,471 @@
...
+            elif key == "locked":
+                if value:
+                    assert "Locked" in http_response.text
+            elif key == "suggest":
+                if value == "reboot":
+                    assert "Reboot Required" in http_response.text
+                elif value == "logout":
+                    assert "Logout Required" in http_response.text
+            else:
+                assert value in http_response.text
```
bowlofeggs: Instead of looping over the keys and values, why not assert the keys you expect directly?

devel/ci/integration/tests/test_bodhi.py
```diff
@@ -16,16 +16,471 @@
...
+        # ffmarkdown...
```
bowlofeggs:
```suggestion
```

devel/ci/integration/tests/test_bodhi.py
```diff
@@ -16,16 +16,471 @@
...
+        # for comment_text in expected_comments:
```
bowlofeggs:
```suggestion
```

devel/ci/integration/tests/test_bodhi.py
```diff
@@ -16,16 +16,471 @@
...
+        #    assert comment_text in http_response.text
```
bowlofeggs:
```suggestion
```

bowlofeggs (changes requested):
Looks good so far!

Rather than making one mega pull request, why don't we focus on finishing out the tests in this one, and then add more tests in subsequent PRs? That was they don't become too huge and are easier to review/edit.

devel/ci/integration/tests/test_bodhi.py
sebwoj: At first I wrote the assertions for comments, but I noticed that we sue "markdown" to create links in the middle of comment text. I decided to drop update's comment tests.

devel/ci/integration/tests/test_bodhi.py
sebwoj: I'm not sure I understand. Do you think that it will be better to check for keys in http_response.text instead of values?

bowlofeggs: I meant more to ask why we don't just expect the keys to be present and just make the above assertions, as opposed to the large if/else block. For example, why not do things like this instead:

```
if update_info['request']:
    assert update_info['request'] in http_response.text
assert f"stable threshold: {update_info['stable_karma']}" in http_response.text
<etc>
````

Comment (sebwoj) [reply devel/ci/integration/tests/test_bodhi.py]: At first, my intention was just to loop over values and check existance(of raw value) in response, but I find out that  most of `update's` fields are rendered different way than just raw value, so I started to adding more and more if/else statements. I will rewrite it.

File: devel/ci/integration/tests/test_bodhi.py [hunk 5]
```diff
@@ -1,4 +1,4 @@
-# Copyright © 2018 Red Hat, Inc.
+# Copyright © 2019 Red Hat, Inc.
```
bowlofeggs:
```suggestion
# Copyright © 2018-2019 Red Hat, Inc.
```

bowlofeggs (changes requested):
Just need to keep the 2018 and rebase - I'll take care of that for you.

bowlofeggs (approved):
Thanks!

PR #3052, opened 2019-03-05
Make the pagure namespace for 'flatpaks' configurable

owtaylor:
The previous change to look up Flatpaks under 'flatpaks/' rather than
'modules/' was backwards incompatible - to restore compatibility, make
this a new config setting: pagure_flatpak_namespace.

### DIFF
--- a/bodhi/server/models.py
+++ b/bodhi/server/models.py
@@ -997,9 +997,12 @@ class Package(Base):
         """
         pagure_url = config.get('pagure_url')
         # Pagure uses plural names for its namespaces such as "rpms" except for
-        # container. Flatpaks build directly from the 'modules' namespace
+        # container. Flatpaks were moved from 'modules' to 'flatpaks' - hence
+        # a config setting.
         if self.type.name == 'container':
             namespace = self.type.name
+        elif self.type.name == 'flatpak':
+            namespace = config.get('pagure_flatpak_namespace')
         else:
             namespace = self.type.name + 's'
         package_pagure_url = '{0}/api/0/{1}/{2}?expand_group=1'.format(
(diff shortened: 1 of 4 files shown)


bodhi/server/models.py
```diff
@@ -1000,9 +1000,12 @@ def get_pkg_committers_from_pagure(self):
         """
         pagure_url = config.get('pagure_url')
         # Pagure uses plural names for its namespaces such as "rpms" except for
-        # container. Flatpaks build directly from the 'modules' namespace
+        # container. Flatpaks were moved from 'modules' to 'flatpaks' - hence
+        # a config setting.
         if self.type.name == 'container':
             namespace = self.type.name
+        elif self.type.name == 'flatpak':
```
bowlofeggs, line 1004: Since this is an Enum, we can compare it directly:

```suggestion
        elif self.type == ContentType.flatpak:
```

Since there's another one above this, I'll just file a new PR to fix this and the other one together.

bowlofeggs (approved):
Thanks!

bowlofeggs:
I'm applying a small patch:

```
diff --git a/bodhi/tests/server/test_models.py b/bodhi/tests/server/test_models.py
index 0ef20007..39d8eeaf 100644
--- a/bodhi/tests/server/test_models.py
+++ b/bodhi/tests/server/test_models.py
@@ -1081,7 +1081,7 @@ class TestFlatpakPackage(ModelTest, unittest.TestCase):
     klass = model.FlatpakPackage
     attrs = dict(name=u"flatpak-runtime")
 
-    def patch_http_session(http_session, namespace):
+    def patch_http_session(self, http_session, namespace):
         """Patch in the correct pagure API result for the particular flatpaks namespace."""
         json_output = {
             "access_groups": {
```

bowlofeggs:
The ```f30-flake8``` test actually passed, but CI failed to report it to GitHub. I will manually merge this.

PR #3055, opened 2019-03-08
Add Fedora 30 to the CI Greenwave policy.

bowlofeggs:
The tests started to fail due to not having Fedora 30 in the
policy.


pypingou (approved):
easy enough :)

PR #3063, opened 2019-03-15
log why buildroot overrides are expired

### LINKED ISSUE #3060: Add logs at every place that expires buildroot overrides
Some users have been having trouble with buildroot overrides getting unexpectedly untagged in Koji, and according to Koji the Bodhi user is the one untagging it. Unfortunately, Bodhi does not currently emit log messages when ```BuildrootOverride.expire()``` is called, so it is not easy to tell which code path is responsible for this.

We should find all the calls to ```BuildrootOverride.expire()``` and add log messages to explain *why* it is being called, for each one. This way we will have more useful debugging information in the future.

nphilipp:
This lets us debug BROs unexpectedly getting untagged in Koji easier.

Fixes #3060

The relevant unit tests seem to run fine here, but I have some trouble with unrelated tests and the dev Vagrant box, so someone with a working env should double-check this to be on the safe side.

### DIFF
--- a/bodhi/server/consumers/composer.py
+++ b/bodhi/server/consumers/composer.py
@@ -608,6 +608,7 @@ class ComposerThread(threading.Thread):
                 for build in update.builds:
                     if build.override:
                         try:
+                            log.debug(f"Expiring BRO for {build.nvr} because it is being pushed.")
                             build.override.expire()
                         except Exception:
                             log.exception('Problem expiring override')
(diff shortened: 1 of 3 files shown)


bowlofeggs (approved):
Thanks!

bowlofeggs:
The test failure is due to https://github.com/fedora-infra/bodhi/issues/3074 - I will run the tests manually and merge.

bowlofeggs:

bowlofeggs:

PR #3071, opened 2019-03-19
Add fedora-messaging consumer to the vagrant environment.

cverna:
This commit adds a fedora-messaging configuration file example and
start the fedora-messaging service.

### DIFF
--- a/devel/ansible/roles/bodhi/files/bodhi.toml
+++ b/devel/ansible/roles/bodhi/files/bodhi.toml
@@ -1,0 +1,63 @@
+# A sample configuration for fedora-messaging. This file is in the TOML format.
+# For complete details on all configuration options, see the documentation.
+
+amqp_url = "amqp://"
+
+publish_exchange = "amq.topic"
+
+callback = "bodhi.server.consumers:messaging_callback"
+
+# The topic_prefix configuration value will add a prefix to the topics of every sent message.
+# This is used for migrating from fedmsg, and should not be used afterwards.
+topic_prefix = "org.fedoraproject.dev"
+
+# Note the double brackets below.
+# To add another binding, add another [[bindings]] section.
+[[bindings]]
+queue = "my_queue"
+exchange = "amq.topic"
+routing_keys = [
+    "org.fedoraproject.*.bodhi.composer.start",
+    "org.fedoraproject.*.buildsys.tag",
+    "org.fedoraproject.*.bodhi.update.request.testing",
+    "org.fedoraproject.*.bodhi.update.edit",
+]
+
+[tls]
+ca_cert = "/etc/pki/tls/certs/ca-bundle.crt"
+keyfile = "/my/client/key.pem"
+certfile = "/my/client/cert.pem"
+
+[client_properties]
+app = "Bodhi"
+
+[queues.my_queue]
+durable = true
+auto_delete = false
+exclusive = false
+arguments = {}
+
+[qos]
+prefetch_size = 0
+prefetch_count = 25
+
+[log_config]
+version = 1
+disable_existing_loggers = true
+
+[log_config.formatters.simple]
+format = "[%(name)s %(levelname)s] %(message)s"
+
+[log_config.handlers.console]
+class = "logging.StreamHandler"
+formatter = "simple"
+stream = "ext://sys.stdout"
+
+[log_config.loggers.fedora_messaging]
+level = "INFO"
+propagate = false
+handlers = ["console"]
+
+[log_config.root]
+level = "WARNING"
+handlers = ["console"]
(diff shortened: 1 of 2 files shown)


abompard:
I think it'd be nice to keep the `print-messages` service, don't you think?

devel/ansible/roles/bodhi/files/bodhi.toml
```diff
@@ -0,0 +1,72 @@
...
+
+amqp_url = "amqp://"
+
+publish_exchange = "amq.topic"
+
+callback = "bodhi.server.consumers:messaging_callback"
+
+# The topic_prefix configuration value will add a prefix to the topics of every sent message.
+# This is used for migrating from fedmsg, and should not be used afterwards.
+topic_prefix = ""
```
abompard: You should set `org.fedoraproject.dev` here if you want to subscribe to the messages with a prefix (see the `[[bindings]]` section below.

devel/ansible/roles/bodhi/files/bodhi.toml
```diff
@@ -0,0 +1,72 @@
...
+
+[tls]
+ca_cert = "/etc/pki/tls/certs/ca-bundle.crt"
+keyfile = "/my/client/key.pem"
+certfile = "/my/client/cert.pem"
+
+[client_properties]
+app = "Bodhi"
+
+[exchanges.custom_exchange]
```
abompard: This section should be removed.

devel/ansible/roles/bodhi/files/bodhi.toml
```diff
@@ -0,0 +1,72 @@
...
+durable = true
+auto_delete = false
+exclusive = false
+arguments = {}
+
+[qos]
+prefetch_size = 0
+prefetch_count = 25
+
+[consumer_config]
```
abompard: If this section is unused, let's remove it.

devel/ansible/roles/bodhi/tasks/main.yml
```diff
@@ -153,7 +153,6 @@
   with_items:
       - bodhi.service
       - fedmsg-tail.service
-      - print-messages.service
```
abompard: I think it would be nice to keep this service, don't you think?

abompard (changes requested):

devel/ansible/roles/bodhi/tasks/main.yml
cverna: yes actually it does not cost much to leave it

devel/ansible/roles/bodhi/files/bodhi.toml
cverna: Done, do I need to change anything from the bindings (remove the org.fedoraproject.*.) ?

bowlofeggs:
On Tue, 2019-03-19 at 08:57 -0700, Clement Verna wrote:
> Done, do I need to change anything from the bindings (remove the
> org.fedoraproject.*.) ?

I would be in favor of killing this leftover from fedmsg. I see the
switch to fedora-messaging as a nice time to reset all expectations
about Bodhi's messages, and we are releasing a major version (4.0.0),
which signals backwards incompatible changes. In 
https://github.com/fedora-infra/bodhi/pull/2842 I am working on
changing the schemas we are publishing. What better way to indicate
that things changed than to stop using this silly prefix? ☺

abompard:
I agree that the topic prefix is silly but topics are actually hardcoded in a lot of places that you don't want to touch (like the java Jenkins plugin), and in other places too because the topic to subscribe to is a class attribute in fedmsg, not configuration. If we change that we have to go over all possible consumers and update them.
My plan was to move to fedora-messaging using the exact same topic, and then change the topics as part of the schema change (since publishing a different schema is basically publishing on a different topic).

devel/ansible/roles/bodhi/files/bodhi.toml
abompard: No, they need to match, so you either add the prefix in `topic_prefix` or remove it from the bindings. It should work as it is now.

bowlofeggs:
On Tue, 2019-03-19 at 09:47 -0700, Aurélien Bompard wrote:
> I agree that the topic prefix is silly but topics are actually
> hardcoded in a lot of places that you don't want to touch (like the
> java Jenkins plugin), and in other places too because the topic to
> subscribe to is a class attribute in fedmsg, not configuration. If we
> change that we have to go over all possible consumers and update
> them.
> My plan was to move to fedora-messaging using the exact same topic,
> and then change the topics as part of the schema change (since
> publishing a different schema is basically publishing on a different
> topic).

Fair enough, +1

cverna:
So are we happy to merge this ? Or does it need something else ?

abompard (approved):
Yep, looks good :-)

PR #3072, opened 2019-03-20
Mitigate one kitchen overuse

hroncok:
This function converted a fixed unicode character to itself

Signed-off-by: Miro Hrončok <miro@hroncok.cz>

Disclaimer: This is untested.

### DIFF
--- a/bodhi/tests/server/test_mail.py
+++ b/bodhi/tests/server/test_mail.py
@@ -88,7 +88,7 @@ class TestGetTemplate(base.BaseTestCase):
 
         # Assemble the template for easier asserting.
         t = '\n'.join([l for l in t[0]])
-        self.assertTrue('\xe7' in t)
+        self.assertIn('\xe7', t)
 
     def test_module_build(self):
         """ModuleBuilds don't have get_latest(), so lets verify that this is OK."""
--- a/bodhi/tests/server/test_mail.py
+++ b/bodhi/tests/server/test_mail.py
@@ -22,8 +22,6 @@ import os
 import smtplib
 import unittest
 
-from kitchen.text import converters
-
 from bodhi.server import config, mail, models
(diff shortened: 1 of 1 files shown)


bodhi/tests/server/test_mail.py
```diff
@@ -90,7 +88,7 @@ def test_encoding(self):
 
         # Assemble the template for easier asserting.
         t = '\n'.join([l for l in t[0]])
-        self.assertTrue(converters.to_unicode('\xe7') in t)
+        self.assertTrue('\xe7' in t)
```
sebwoj: I recommend `assertIn` here.

sebwoj (changes requested):

hroncok:
Should I worry about the CI? This could not have possibly caused breakage in integration tests.

bodhi/tests/server/test_mail.py
hroncok:
```suggestion
        self.assertIn('\xe7', t)
```

hroncok:
Note: If I use the suggestions feature on GitHub, the DCO thing breaks. Off to a terminal.

hroncok:
https://github.com/fedora-infra/bodhi/issues/3074

bodhi/tests/server/test_mail.py
```diff
@@ -90,7 +88,7 @@ def test_encoding(self):
 
         # Assemble the template for easier asserting.
         t = '\n'.join([l for l in t[0]])
-        self.assertTrue(converters.to_unicode('\xe7') in t)
```
bowlofeggs, line 93: Haha, kitchen is so bad.

bowlofeggs (approved):
Thanks for the patch! I'm going to squash it to one commit and merge.

PR #3076, opened 2019-03-20
Use assert methods from unittest instead of raw assert in tests

### LINKED ISSUE #2948: Use assert methods from unittest instead of clean "assert" in tests
SSIA

sebwoj:
Fixes #2948

### DIFF
--- a/bodhi/tests/server/services/test_updates.py
+++ b/bodhi/tests/server/services/test_updates.py
@@ -127,36 +127,36 @@ class TestNewUpdate(BaseTestCase):
     def test_duplicate_build(self, *args):
         res = self.app.post_json(
             '/updates/', self.get_update([u'bodhi-2.0-2.fc17', u'bodhi-2.0-2.fc17']), status=400)
-        assert 'Duplicate builds' in res, res
+        self.assertIn('Duplicate builds', res)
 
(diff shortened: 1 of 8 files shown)


bowlofeggs:
The ```f30-pydocstyle``` test actually passed - the CI system just failed to report it to GitHub.

bodhi/tests/server/services/test_updates.py
```diff
@@ -3624,12 +3623,8 @@ def test_stable_request_when_stable(self, publish, *args):
 
         self.assertEqual(resp.json['update']['status'], 'stable')
         self.assertEqual(resp.json['update']['request'], None)
-        try:
-            publish.assert_called_with(
-                topic='update.request.stable', msg=mock.ANY)
-            assert False, "request.stable fedmsg shouldn't have fired"
-        except AssertionError:
-            pass
```
bowlofeggs, line 3632: Hahahahaha, whaaaaaat was going on there?

bowlofeggs (approved):
Thanks!

bodhi/tests/server/services/test_updates.py
sebwoj: Yeah, there are some places like this one in Bodhi. Take a look at `validate_updateid` :D

PR #3081, opened 2019-03-22
Ignore update passing CI when retrieving the greenwave status information

pypingou:
Basically, this script, ran as a cron job, was asking greenwave for all
updates pending to be move to testing or to stable, regardless of the
CI status that the update could have been assigned to in a previous run.
This means that during freeze periods when updates pile on pending to go
to stable, the cron job will keep on checking all the updates including
the ones that we know already passed CI.

### DIFF
--- a/bodhi/server/scripts/check_policies.py
+++ b/bodhi/server/scripts/check_policies.py
@@ -48,6 +48,11 @@ def check():
     ).filter(
         models.Release.state.in_(
             [models.ReleaseState.current, models.ReleaseState.pending])
+    ).filter(
+        models.Update.test_gating_status.notin_(
+            models.TestGatingStatus.passed,
+            models.TestGatingStatus.ignored
+        )
     ).order_by(
         # Check the older updates first so there is more time for the newer to
         # get their test results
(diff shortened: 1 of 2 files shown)


bodhi/server/scripts/check_policies.py
```diff
@@ -48,6 +48,11 @@ def check():
     ).filter(
         models.Release.state.in_(
             [models.ReleaseState.current, models.ReleaseState.pending])
+    ).filter(
+        models.Update.test_gating_status.notin_(
+            models.TestGatingStatus.passed,
+            models.TestGatingStatus.ignored
+        )
```
bowlofeggs, line 55: We don't want to do this - this is this way on purpose because Greenwave can change its mind later, after having passed. We actually had a meeting about this and chose to go this way. This proposed change takes the "Continuous" out of "Continuous Integration".

bowlofeggs (changes requested):

bodhi/server/scripts/check_policies.py
pypingou: I've discussed this with the folks running the CI pipeline and they agreed that once something has passed it should not be reconsidered, which of course isn't true for failures.

I've pinged them so they can join the discussion here.

Issue #1014, opened 2016-10-08
bodhi.server.models.models.Update.met_testing_requirements assumes a setting value


bowlofeggs:
[`met_testing_requirements()`](https://github.com/fedora-infra/bodhi/blob/develop/bodhi/server/models/models.py#L1791-L1812) assumes that `testing_approval_msg` is set to a particular value, even though it is a setting and should be alterable by the admin:

```
        for comment in self.comments:
            if comment.user.name == 'bodhi' and \
               comment.text.startswith('This update has reached') and \
               'days in testing and can be pushed to stable now if the ' \
               'maintainer wishes' in comment.text:
                return True
        return False
```

We need to devise a better way to tell if Bodhi has made this comment or not.

Issue #2978, opened 2019-02-05
[M] Give Bodhi automation to push based on time in testing


bowlofeggs:
Right now Bodhi has the ability to push updates to stable when they reach a karma threshold, but updates that reach the time in testing need to be manually pushed by a packager.

This RFE is about giving Bodhi the ability to automatically push the update to stable when it reaches the time in testing, so long as it does not have any negative karma (negative karma currently stops autopushing based on karma, so we want something similar for time).

This has been [approved by FESCo](https://pagure.io/fesco/issue/2048), and here is a list of requirements from that ticket:

* The boolean to select push based on time should be independent from the existing boolean to push based on karma.
* The packager should be able to set the number of days they want it to wait, not less than the release's minimum days (i.e., for Fedora they could set it at 7 days on the low end, or 30 if they like 30).

bowlofeggs:
We need this for Rawhide Gating too, so that updates get moved to stable automatically when the CI tests pass on Rawhide.

cverna:
> We need this for Rawhide Gating too, so that updates get moved to stable automatically when the CI tests pass on Rawhide.

If I understand it well what we need for rawhide gating is not exactly the same this ticket is about automatic pushing based on time while rawhide gating is based on the CI test. Are we planning to implement both feature at the same time ? or should we create a specific ticket for automatic push based on the CI ?

cverna:
For rawhide gating we will use time base pushed but time will be set to 0.

cverna:
> The packager should be able to set the number of days they want it to wait, not less than the release's minimum days (i.e., for Fedora they could set it at 7 days on the low end, or 30 if they like 30).

I propose to implement this in a second step, I don't think it is critical for rawhide gating and it will make this change simpler. What do you think ?

bowlofeggs:
On Thu, 2019-03-21 at 05:12 -0700, Clement Verna wrote:
> I propose to implement this in a second step, I don't think it is
> critical for rawhide gating and it will make this change simpler.
> What do you think ?

Since this issue is tied to a FESCo decision, I think we should
implement it as discussed in the FESCo ticket and should not partially
implement it.

cverna:
What should be the default value for the number of days in testing ? As far as I know when creating an update we do not know yet the release associated with this update, so we can't have a per release default value.

[...]

PR #3090, opened 2019-03-25
Add autopush boolean to an update.

cverna:
This commit adds an autopush boolean to an update.
When this boolean is true the approve_testing cronjob
will push the update to stable if it has meet the
testing requirements (time).


bowlofeggs (changes requested):
I think we should also include the date to push in this. If we don't, we have to do it in two stages, the second of which will require a major release of Bodhi, and the extra overhead of tracking that work outweighs the relatively little work it takes to add it.

cverna:
Ok , I have update the PR with the `stable_days` field in the database so that we can store the user input on the number of days in testing required.

If this looks good, I ll work on adding support to the cli.

bodhi/client/__init__.py
```diff
@@ -107,6 +107,9 @@ def _set_logging_debug(ctx, param, value):
     click.option('--close-bugs', is_flag=True, help='Automatically close bugs'),
     click.option('--request', help='Requested repository',
                  type=click.Choice(['testing', 'stable', 'unpush'])),
+    click.option('--autopush', is_flag=True, help='Enable stable push base on time in testing'),
+    click.option('--stable-days', type=click.INT,
+                 help='Day in testing required to push to stable'),
```
bowlofeggs:
```suggestion
                 help='Days in testing required to push to stable'),
```

bodhi/server/migrations/versions/b1fd856efcf6_add_autopush_boolean_and_stable_days_to_update.py
```diff
@@ -0,0 +1,67 @@
+# Copyright (c) 2019
```
bowlofeggs:
```suggestion
# Copyright (c) 2019 Red Hat, Inc.
```

bodhi/server/migrations/versions/b1fd856efcf6_add_autopush_boolean_and_stable_days_to_update.py
```diff
@@ -0,0 +1,67 @@
...
+    # set stable_days to 0 for old updates
+    op.execute(
+        "UPDATE updates SET stable_days=0 "
+        "WHERE updates.status IN ('stable', 'obsolete', 'unpushed')"
+    )
+    # set stable_days to 7 for non critpath FEDORA updates
+    op.execute(
+        "UPDATE updates SET stable_days=7 "
+        "WHERE updates.status NOT IN ('stable', 'obsolete', 'unpushed')"
+        "AND updates.critpath = FALSE AND updates.release_id NOT IN (8, 9, 10)"  # EPEL releases
```
bowlofeggs: There are Bodhi deployments that are not Fedora's. We should not write migrations that depend on particular data in the database.

bodhi/server/models.py
```diff
@@ -1559,8 +1559,12 @@ class Update(Base):
     Attributes:
         autokarma (bool): A boolean that indicates whether or not the update will
             be automatically pushed when the stable_karma threshold is reached.
+        autopush (bool): A boolean that indicates whether or not the update will
+            be automatically pushed when the time threshold is reached.
         stable_karma (int): A positive integer that indicates the amount of "good"
             karma the update must receive before being automatically marked as stable.
+        stable_days (int): A positive integer that indicates the number of days an update
+        needs to spend in testing before being automatically marked as stable.
```
```suggestion
            needs to spend in testing before being automatically marked as stable.
```

bodhi/server/models.py
```diff
@@ -2005,6 +2011,16 @@ def new(cls, request, data):
         release = data.pop('release', None)
         up = Update(**data, release=release)
 
+        # stable_days can be set by the user we want to make sure that the value
```
```suggestion
        # We want to make sure that the value of stable_days
```

bodhi/server/models.py
```diff
@@ -2005,6 +2011,16 @@ def new(cls, request, data):
...
+        # will not be lower than the mandatory_days_in_testing.
+        if up.mandatory_days_in_testing > up.stable_days:
+            up.stable_days = up.mandatory_days_in_testing
+            caveats.append({
+                'name': 'stable days',
+                'description': "The number of stable days required was set to the mandatory "
```
bowlofeggs, line 2049:
```suggestion
                'description': "The number of stable days required was raised to the mandatory "
```

bodhi/server/models.py
```diff
@@ -3330,7 +3356,7 @@ def meets_testing_requirements(self):
         Returns:
             bool: True if the update meets testing requirements, False otherwise.
         """
-        num_days = self.mandatory_days_in_testing
+        num_days = self.stable_days
```
bowlofeggs: I don't think we want this value to be used to determine whether the update has met its testing requirements. We only want this number to be used to determine when Bodhi automatically pushes it - the required days in testing is derived from the release policy.

bodhi/server/models.py
```diff
@@ -3371,7 +3397,7 @@ def met_testing_requirements(self):
         Returns:
             bool: See description above for what the bool might mean.
         """
-        min_num_days = self.mandatory_days_in_testing
+        min_num_days = self.stable_days
```
bowlofeggs: Same here.

bodhi/server/models.py
```diff
@@ -3400,7 +3426,7 @@ def days_to_stable(self):
                 determined.
         """
         if not self.meets_testing_requirements and self.date_testing:
-            num_days = (self.mandatory_days_in_testing - self.days_in_testing)
+            num_days = (self.stable_days - self.days_in_testing)
```
bowlofeggs: Same here.

bodhi/server/scripts/approve_testing.py
```diff
@@ -50,6 +50,23 @@ def usage(argv):
...
+    """
+    Request an update to be pushed to stable.
+
+    Args:
+        update: The update object that will be pushed to stable.
+        db: A database session.
+    """
+    print(f"Automatically marking {update.alias} as stable")
+    update.set_request(db=db, action=UpdateRequest.stable, username="bodhi")
+    update.date_pushed = None
```
bowlofeggs: Why this line? The update push date is usually only edited by the composer.

bodhi/server/scripts/approve_testing.py
```diff
@@ -50,6 +50,23 @@ def usage(argv):
...
+    notifications.publish(
+        topic='update.requirements_met.stable',
+        msg=dict(update=update, status='stable')
+    )
```
bowlofeggs: I think we want to keep sending this in the place we were sending it before - the update can reach stable requirements before it autopushes - we want to keep the concepts of meeting requirements and autopushing separately.

bodhi/server/scripts/approve_testing.py
```diff
@@ -75,10 +92,14 @@ def main(argv=sys.argv):
...
             if not update.release.mandatory_days_in_testing:
-                print('%s doesn\'t have mandatory days in testing' % update.release.name)
-                continue
+                if not update.autopush:
+                    # If this release does not have any testing requirements and is not autopush,
+                    # skip it
+                    print('%s doesn\'t have mandatory days in testing' % update.release.name)
+                    continue
+                # if the release is configured to autopush, push it to stable.
+                update_to_stable(update, db)
```
bowlofeggs: This will push updates that failed gating.

bowlofeggs: Let's make sure we add a test for this case to ensure that updates that failed gating do not get pushed.

bodhi/server/scripts/approve_testing.py
```diff
@@ -100,14 +121,19 @@ def main(argv=sys.argv):
...
-                    config.get('testing_approval_msg') % update.mandatory_days_in_testing)
-                update.comment(db, text, author=u'bodhi')
-
-                notifications.publish(
-                    topic='update.requirements_met.stable',
-                    msg=dict(update=update))
-                db.commit()
+                if update.autopush:
+                    update_to_stable(update, db)
+                else:
```
bowlofeggs: I don't think we want to if/else this - we want both to happen. Note that we also want to ensure that the update actually reached the number of days in testing here before pushing.

bodhi/server/scripts/approve_testing.py
```diff
@@ -100,14 +121,19 @@ def main(argv=sys.argv):
...
+                    text = str(
+                        config.get('testing_approval_msg') % update.stable_days
+                    )
```
bowlofeggs: This is an unrelated change in style (and to a style that I personally dislike since it wastes vertical space).

bodhi/server/templates/new_update.html
```diff
@@ -251,6 +260,16 @@ <h2 class="pull-left">Edit
                       % endif
                     ></dd>
 
+                  <dt data-toggle="tooltip" title="This is required number of days an update needs to spend in testing to automatically upush an update to stable.">
```
bowlofeggs:
```suggestion
                  <dt data-toggle="tooltip" title="This is the number of days an update needs to spend in testing to be automatically pushed to stable.">
```

bowlofeggs (changes requested):

bodhi/server/migrations/versions/b1fd856efcf6_add_autopush_boolean_and_stable_days_to_update.py
cverna: Ha good point, so what should we default stable_days to ?

bodhi/server/models.py
cverna: The reasoning behind this was to keep the number of days required in testing into a single place and not 2 (in one case `stable_days` and in other case `mandatory_days_in_testing`).  With this patch (expect if I missed something) stable_days will always be at least equal to `mandatory_days_in_testing`. 
Now the question is if a packager sets `stable_days` to 20 days for example he will not be able to push it manually after the 7 days if that's the release policy, I guess he could still change the `stable_days` back to the release policy in such scenario.

What do you think ?

bodhi/server/scripts/approve_testing.py
cverna: Honestly I can't remember :smile:

bodhi/server/scripts/approve_testing.py
cverna: Yes good point

bodhi/server/scripts/approve_testing.py
cverna: > Note that we also want to ensure that the update actually reached the number of days in testing here before pushing.

Would not that been check by  `meets_testing_requirements` ?

cverna: Just to make sure I understand this correctly, We want to add a comment to the update to inform the packager this can be pushed to stable, and then if `autopush` is true we push it ? Correct ?

bodhi/server/models.py
cverna: I had a good thought about it today, and I think it is indeed better to keep the 2 value separated since they don't have exactly the same meaning.

bodhi/server/models.py
```diff
@@ -1534,8 +1534,12 @@ class Update(Base):
     Attributes:
         autokarma (bool): A boolean that indicates whether or not the update will
             be automatically pushed when the stable_karma threshold is reached.
+        autopush (bool): A boolean that indicates whether or not the update will
```
bowlofeggs: Maybe we should call this "autotime" since the other one is called "autokarma". "autopush" sounds more generic and the name doesn't make it clear that it's only for time.

bowlofeggs: Alternatively, we could rename autokarma to autopush_karma and this one to autopush_time. That might be cleaner, though it makes the migration annoying. Your option.

bodhi/server/models.py
```diff
@@ -3339,7 +3365,7 @@ def met_testing_requirements(self):
         if min_num_days:
             if not self.meets_testing_requirements:
                 return False
-        else:
+        elif not self.autopush:
```
bowlofeggs: Why this change?

bodhi/server/scripts/approve_testing.py
```diff
@@ -100,13 +102,18 @@ def main(argv=sys.argv):
             # not reached the karma threshold.
             if update.meets_testing_requirements:
                 print(f'{update.alias} now meets testing requirements')
-                text = str(
-                    config.get('testing_approval_msg') % update.mandatory_days_in_testing)
+
+                text = str(config.get('testing_approval_msg') % update.mandatory_days_in_testing)
```
bowlofeggs: This is an unrelated style change.

bodhi/server/scripts/approve_testing.py
```diff
@@ -100,13 +102,18 @@ def main(argv=sys.argv):
...
                 update.comment(db, text, author=u'bodhi')
 
                 notifications.publish(
                     topic='update.requirements_met.stable',
-                    msg=dict(update=update))
+                    msg=dict(update=update)
+                )
```
bowlofeggs: This is an unrelated style change.

bodhi/server/scripts/approve_testing.py
```diff
@@ -100,13 +102,18 @@ def main(argv=sys.argv):
...
+                if update.autopush:
```
bowlofeggs: We should check that the requested number of days was reached.

bowlofeggs: Let's ensure that there are tests that check all of these conditions:

0. The update has not reached mandatory days in testing or stable_days.
1. The update has reached mandatory days in testing but not stable_days.
2. The update has reached mandatory days in testing and stable_days.
3. All of the above but autopush/autotime is False.

bodhi/tests/server/services/test_updates.py
```diff
@@ -501,6 +501,62 @@ def test_new_update_with_multiple_bugs(self, publish, *args):
...
+    @mock.patch(**mock_valid_requirements)
+    @mock.patch('bodhi.server.notifications.publish')
+    def test_new_update_with_invalid_stable_days(self, publish, *args):
+        update = self.get_update('bodhi-2.0.0-2.fc17')
+        update['stable_days'] = 0
+        r = self.app.post_json('/updates/', update, status=400)
+        up = r.json_body
+
+        self.assertEqual(up['status'], 'error')
+        self.assertEqual(up['errors'][0]['description'], "0 is less than minimum value 1")
```
bowlofeggs: Won't we be setting this to 0 for Rawhide? Maybe we should just ensure it is not negative.

bowlofeggs (changes requested):

bodhi/server/models.py
cverna: In case of rawhide min_num_days will be 0 which would make us think that we already met the testing requirements.

bowlofeggs: If the requirement is 0, then we have met the requirements in that case. I think we should keep the concept of autopush and requirement separate. The push date isn't a requirement, it's just an automation.

bodhi/server/models.py
cverna: I did not really want to rename all the `autokarma` so I went for the easy way and renamed it `autotime`

bodhi/server/scripts/approve_testing.py
cverna: Ha yes I used to get this for free in `meets_testing_requirements` but with the last change I need to do the checks.

bowlofeggs:
On Mon, 2019-04-15 at 17:35 +0000, Clement Verna wrote:
> I did not really want to rename all the autokarma so I went for the
> easy way and renamed it autotime

Cool, that works for me. It also avoids yet-another-backwards-
compatible-change (for now). I'll maybe want to rename it later, but
it's fine for now and maybe forever ☺

bodhi/server/models.py
cverna: Ok that makes sense to me, but for some reason in the `approve_testing` script, we do this (see below) if we have met the requirements, which mean we will never push the rawhide updates. 

https://github.com/fedora-infra/bodhi/blob/develop/bodhi/server/scripts/approve_testing.py#L85

The script might be wrong or not, I am not sure :stuck_out_tongue_closed_eyes:

bowlofeggs: @cverna ah yes, I am familiar with that and do not like it!

I think the point of it is to avoid writing a comment on updates that could be pushed every time the script runs, so it's trying to detect if the update has already been commented on. It's not a good way to do that, but I think that's what it's trying to do, if that makes sense.

I think it'd be better if there were:

0. A formal way to mark a Bodhi comment as being the comment that says this. Right now, it checks if the comment was authored by bodhi and has that particular text (ewww). It'd be better if the Comment model had some metadata/enum to mark certain Bodhi comments as being of that type - then we can more cleanly detect whether Bodhi has already added that comment to an update.
1. I don't love that we have ```met_testing_requirements()``` *and* ```meets_testing_requirements()```. The names different in tense, of course, but really I think we just want to make sure we don't comment repeatedly.

Anyways, we don't need to solve these nasty problems in this pull request - I'm just trying to give you some context for what this code seems like it's trying to do. It's confusing…

bowlofeggs: Another hilarious thing about those two functions is that one of them calls the other!

bodhi/server/models.py
```diff
@@ -3339,7 +3365,7 @@ def met_testing_requirements(self):
         if min_num_days:
             if not self.meets_testing_requirements:
                 return False
-        else:
+        elif not self.autotime:
```
bowlofeggs: I still think we should not use the new boolean to determine whether an update has met testing requirements - we need to keep the concept of requirements and automation machinery separate.

bowlofeggs (changes requested):

bodhi/server/models.py
cverna: Yes I think if we get https://github.com/fedora-infra/bodhi/pull/3158 merged before this one then we don't need this change.
Otherwise I can manage this special case in the script instead of here.

cverna:
Fix for #2978

abompard:
The code looks good but you need to rebase your alembic migration on the last migration head.

cverna:
> The code looks good but you need to rebase your alembic migration on the last migration head.

Yes good point, just done it :+1:

cverna:
The pip-docs error seems to be related to sphinx
```
  File "/usr/local/lib/python3.7/site-packages/sphinx/util/docfields.py", line 252, in __init__
     pip-docs                                       self.typemap = directive.get_field_type_map()
     pip-docs                                   AttributeError: 'ServiceDirective' object has no attribute 'get_field_type_map'
```

bowlofeggs:
I'm going to rebase this to resolve the conflict.

bowlofeggs:
I'm also adding this patch:

```
index 078783ea..232225de 100644
@@ -686,7 +686,7 @@ class TestNew(unittest.TestCase):
                     'suggest': None, 'notes': 'No description.', 'request': None,
                     'bugs': '1234567', 'requirements': None, 'unstable_karma': None, 'file': None,
                     'notes_file': None, 'type': 'bugfix', 'severity': None,
-                    'display_name': 'fake display name'
+                    'display_name': 'fake display name', 'autotime': False, 'stable_days': None
                 }
             ),
             mock.call(
```

bodhi/server/models.py
```diff
@@ -1599,7 +1603,9 @@ class Update(Base):
     __get_by__ = ('alias',)
 
     autokarma = Column(Boolean, default=True, nullable=False)
+    autotime = Column(Boolean, default=False, nullable=False)
```
bowlofeggs: It is inconsistent to make this one default to ```False``` when the ```autokarma``` field defaults to ```True```.

```suggestion
    autotime = Column(Boolean, default=True, nullable=False)
```

bodhi/server/models.py
```diff
@@ -1982,6 +1988,16 @@ def new(cls, request, data):
...
         up = Update(**data, release=release)
 
+        # We want to make sure that the value of stable_days
+        # will not be lower than the mandatory_days_in_testing.
+        if up.mandatory_days_in_testing > up.stable_days:
+            up.stable_days = up.mandatory_days_in_testing
+            caveats.append({
+                'name': 'stable days',
+                'description': "The number of stable days required was set to the mandatory "
+                               f"release value {up.mandatory_days_in_testing} days"
```
```suggestion
                               f"release value of {up.mandatory_days_in_testing} days"
```

bodhi/server/models.py
```diff
@@ -2021,6 +2037,16 @@ def edit(cls, request, data):
         caveats = []
         edited_builds = [build.nvr for build in up.builds]
 
+        # stable_days can be set by the user we want to make sure that the value
```
```suggestion
        # stable_days can be set by the user. We want to make sure that the value
```

bodhi/server/models.py
```diff
@@ -2021,6 +2037,16 @@ def edit(cls, request, data):
...
+        # will not be lower than the mandatory_days_in_testing.
+        if up.mandatory_days_in_testing > data.get('stable_days', up.stable_days):
+            data['stable_days'] = up.mandatory_days_in_testing
+            caveats.append({
+                'name': 'stable days',
+                'description': "The number of stable days required was raised to the mandatory "
+                               f"release value {up.mandatory_days_in_testing} days"
```
```suggestion
                               f"release value of {up.mandatory_days_in_testing} days"
```

bodhi/server/scripts/approve_testing.py
```diff
@@ -76,10 +76,12 @@ def main(argv=sys.argv):
         testing = db.query(Update).filter_by(status=UpdateStatus.testing,
                                              request=None)
         for update in testing:
-            # If this release does not have any testing requirements, skip it
             if not update.release.mandatory_days_in_testing:
-                print('%s doesn\'t have mandatory days in testing' % update.release.name)
-                continue
+                if not update.autotime:
```
bowlofeggs: Why not use an ```and``` in the outer if to avoid the nesting?

bodhi/server/scripts/approve_testing.py
```diff
@@ -76,10 +76,12 @@ def main(argv=sys.argv):
...
+                    # If this release does not have any testing requirements and is not autotime,
+                    # skip it
+                    print('%s doesn\'t have mandatory days in testing' % update.release.name)
```
bowlofeggs:
```suggestion
                    print(f"{update.release.name} doesn't have mandatory days in testing")
```

bodhi/server/scripts/approve_testing.py
```diff
@@ -107,6 +109,11 @@ def main(argv=sys.argv):
 
                 notifications.publish(update_schemas.UpdateRequirementsMetStableV1.from_dict(
                     dict(update=update)))
+
+                if update.autotime and update.days_in_testing >= update.stable_days:
+                    print(f"Automatically marking {update.alias} as stable")
+                    update.set_request(db=db, action=UpdateRequest.stable, username="bodhi")
```
bowlofeggs, line 104: If the user sets ```autokarma``` off and the update reaches the karma threshold, this code will not execute because the condition above on line 91 will be ```True```.

Please add a test case for that situation so we can ensure that the right thing happens.

bowlofeggs (dismissed):

abompard:
Please squash the two commits and you should be good to go.

bodhi/server/models.py
```diff
@@ -3391,9 +3417,7 @@ def has_stable_comment(self):
         """
         for comment in self.comments_since_karma_reset:
             if comment.user.name == 'bodhi' and \
-               comment.text.startswith('This update has reached') and \
-               'and can be pushed to stable now if the ' \
-               'maintainer wishes' in comment.text:
+               comment.text == 'This update can be pushed to stable now if the maintainer wishes':
```
sebwoj: 1) We can use value from `testing_approval_msg` config setting instead of hardcoding str. This way we can fix #1014.
2) It looks a little bit backward incompatible, because current comment text is different. On the other hand, it turns out that this change will only cause Bodhi to comment about update readiness twice.

sebwoj (changes requested):

bodhi/server/models.py
cverna: >     1. We can use value from `testing_approval_msg` config setting instead of hardcoding str. This way we can fix #1014.
> 

Yes good idea, I ll change that.

>     2. It looks a little bit backward incompatible, because current comment text is different. On the other hand, it turns out that this change will only cause Bodhi to comment about update readiness twice.

Yes there is a risk that for a some updates Bodhi will comment twice, to me this is acceptable. What do @bowlofeggs and @abompard think about that ?

abompard: What about changing the first test to `.startswith('This update ')` and the second to `'can be pushed to stable now if the maintainer wishes'`? This way it'll find the new version of the comment too.

cverna: @abompard that works for me :-)

bowlofeggs:
On Wed, 2019-06-19 at 00:56 -0700, Clement Verna wrote:
> Yes there is a risk that for a some updates Bodhi will comment twice,
> to me this is acceptable. What do @bowlofeggs and @abompard think
> about that ?

Yeah. I hate the way the detection code works - we really should
refactor it so that the Comment model has a type or tag on it, and then
the detection code should look for Comments with that type to know if
the message had been written rather than looking for an exact match.
It's real ugly how it works today…

That said, I'm fine with two comments here. There's not really an easy
way to avoid it without doing the above refactor.

abompard (approved):
Yep, looks good to me

bodhi/server/scripts/approve_testing.py
```diff
@@ -77,37 +77,32 @@ def main(argv=sys.argv):
...
             # that we don't care about karma here, because autokarma updates get their request set
             # to stable by the Update.comment() workflow when they hit the required threshold. Thus,
             # this function only needs to consider the time requirements because these updates have
             # not reached the karma threshold.
             if update.meets_testing_requirements:
                 print(f'{update.alias} now meets testing requirements')
-                text = str(
-                    config.get('testing_approval_msg') % update.mandatory_days_in_testing)
-                update.comment(db, text, author='bodhi')
+                update.comment(db, str(config.get('testing_approval_msg')), author='bodhi')
```
sebwoj, line 97: Just for curiosity: Do we really need this type cast here? `testing_approval_msg` setting already has: `'validator': str`

sebwoj (commented):
Just one question.

bodhi/tests/server/scripts/test_approve_testing.py
```diff
@@ -488,6 +483,280 @@ def test_usage(self, stdout, exit):
             'usage: nosetests <config_uri>\n(example: "nosetests development.ini")\n')
         exit.assert_called_once_with(1)
 
+def test_autotime_update_meeting_test_requirements_gets_pushed(self):
```
sebwoj: Indentation level here. Unit test are failing.

sebwoj (changes requested):

bowlofeggs:
On Thu, 2019-06-20 at 03:00 -0700, Sebastian Wojciechowski wrote:
> Just for curiosity: Do we really need this type cast here?
> testing_approval_msg setting already has: 'validator': str

I agree - any of Bodhi's non-dynamic settings* should be validated and
cast in bodhi.server.config. You can also always assume they exist and
are valid as long as they are in that file (and we should strive to put
all settings that are possible into that file). Thus, no need for using
.get('something', DEFAULT), and in fact doing that is bad because it
duplicates code and could easily lead to bugs.


* Bodhi has some dynamic settings (which I strongly dislike) where the
  setting is a combination of release names and states and things like
  that. These setting names are not standard and are not predictable,
  so they are not in bodhi.server.config and must be cast when they are
  used. These are the only exception to the above.

PR #3092, opened 2019-03-25
Configure Mergify to backport patches to 3.14.

abompard:


.mergify.yml
```diff
@@ -73,16 +73,16 @@ pull_request_rules:
 - actions:
     backport:
       branches:
-      - '3.12'
```
bowlofeggs: We actually still have 3.12 on F28 and F29, so I think we should keep this around in case we need to backport anything for those releases until they aren't on that version. I don't recall the reason, but I had opted not to bring them to 3.13 when it was released.

bowlofeggs (changes requested):
+1 to the general idea, but let's keep the 3.12 backport for now.

bowlofeggs (approved):

PR #3094, opened 2019-03-25
Drop the dependency on kitchen.

bowlofeggs:
Bodhi had been using kitchen to do a few different things, all of
which were either not useful or silly:

0. Encode strs into strs.
1. Encode strs into bytes. This is silly because we can do that
   just as easily and more correctly with the standard library
   (more correctly because kitchen uses the replace method on
   errors - we would be much better off to know about the errors
   so we can fix them).
2. Iterate on iterables, or turn a non-iterable into a single item
   iterable.

Each of these were replaced by using the standard library.

### DIFF
--- a/bodhi/server/mail.py
+++ b/bodhi/server/mail.py
@@ -21,9 +21,6 @@ from textwrap import wrap
 import os
 import smtplib
 
-from kitchen.iterutils import iterate
-from kitchen.text.converters import to_unicode, to_bytes
-
 from bodhi.server import log
 from bodhi.server.config import config
 from bodhi.server.util import get_rpm_header, get_absolute_path
@@ -258,7 +255,7 @@ def read_template(name):
     if os.path.exists(template_path):
         try:
             with open(template_path) as template_file:
-                return to_unicode(template_file.read())
+                return template_file.read()
         except IOError as e:
             log.error("Unable to read template file: %s" % (template_path))
             log.error("IO Error[%s]: %s" % (e.errno, e.strerror))
@@ -350,17 +347,9 @@ def get_template(update, use_template='fedora_errata_template'):
             elif isinstance(oldtime, list):
                 oldtime = oldtime[0]
             info['changelog'] = u"ChangeLog:\n\n%s%s" % \
-                (to_unicode(build.get_changelog(oldtime)), line)
+                (build.get_changelog(oldtime), line)
 
-        try:
-            templates.append((info['subject'], use_template % info))
-        except UnicodeDecodeError:
-            # We can't trust the strings we get from RPM
-            log.debug("UnicodeDecodeError! Will try again after decoding")
-            for (key, value) in info.items():
-                if value:
-                    info[key] = to_unicode(value)
-            templates.append((info['subject'], use_template % info))
+        templates.append((info['subject'], use_template % info))
 
     return templates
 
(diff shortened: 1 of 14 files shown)


bowlofeggs:
@hroncok I got you an early Christmas present.

bodhi/server/mail.py
```diff
@@ -412,14 +401,14 @@ def send_mail(from_addr, to_addr, subject, body_text, headers=None):
     if to_addr in config.get('exclude_mail'):
         return
 
-    subject = to_bytes(subject)
-    body_text = to_bytes(body_text)
+    subject = subject.encode('utf-8')
+    body_text = body_text.encode('utf-8')
 
-    msg = [b'From: %s' % to_bytes(from_addr), b'To: %s' % to_bytes(to_addr)]
+    msg = [b'From: %s' % from_addr.encode('utf-8'), b'To: %s' % to_addr.encode('utf-8')]
```
hroncok:
```suggestion
    msg = [f'From: {from_addr}'.encode('utf-8'), f'To: {to_addr}'.encode('utf-8')]
```

hroncok: Also, it might be easier to encode the message once it's joined.

hroncok: Something like https://github.com/bowlofeggs/bodhi/pull/2

Zlopez (approved):
LGTM

bowlofeggs:
@hroncok I am going to apply this patch on top of yours to fix a couple of tests that expected the old behavior:

```
diff --git a/bodhi/tests/server/consumers/test_composer.py b/bodhi/tests/server/consumers/test_composer.py
index 6394bb00..dbcb64db 100644
--- a/bodhi/tests/server/consumers/test_composer.py
+++ b/bodhi/tests/server/consumers/test_composer.py
@@ -559,7 +559,7 @@ References:
                                 mock.ANY)
         self.assertEqual(len(mail.mock_calls), 2)
         body = mail.mock_calls[1][1][2]
-        self.assertTrue(body.decode('utf-8').startswith(
+        self.assertTrue(body.startswith(
             ('From: updates@fedoraproject.org\r\nTo: %s\r\nX-Bodhi: fedoraproject.org\r\nSubject: '
              'Fedora 17 updates-testing report\r\n\r\nThe following builds have been pushed to '
              'Fedora 17 updates-testing\n\n    bodhi-2.0-1.fc17\n\nDetails about builds:\n\n\n====='
diff --git a/bodhi/tests/server/test_mail.py b/bodhi/tests/server/test_mail.py
index 9fbd6552..af510456 100644
--- a/bodhi/tests/server/test_mail.py
+++ b/bodhi/tests/server/test_mail.py
@@ -351,7 +351,7 @@ class Test_SendMail(unittest.TestCase):
         mail._send_mail('archer@spies.com', 'lana@spies.com', 'hi')
 
         SMTP.assert_called_once_with('smtp.fp.o')
-        smtp.sendmail.assert_called_once_with('archer@spies.com', ['lana@spies.com'], 'hi')
+        smtp.sendmail.assert_called_once_with('archer@spies.com', ['lana@spies.com'], b'hi')
         warning.assert_called_once_with(
             '"recipient refused" for \'lana@spies.com\', {}'.format(
                 repr(smtp.sendmail.side_effect)))
```

bowlofeggs:
The integration tests failed due to hitting a CI timeout. I've opened https://github.com/fedora-infra/bodhi/pull/3155 to raise that timeout.

I ran the CI tests locally on this patch and they all pass, so I'm going to merge this.

bowlofeggs:

bowlofeggs:

PR #3105, opened 2019-03-27
Configure Vagrant to use unsafe IO.

bowlofeggs:
Since we use Vagrant for development only, and since we don't
store any of our work in Vagrant (the work is stored on the host
and is shared into the guest via sshfs), it is sensible to
configure the guest's IO mode to use unsafe synchronization. This
will give us a significant performance boost, since calls to
fsync() will be ignored by the host.


cverna (approved):
+1

PR #3106, opened 2019-03-27
Drop remaining source code encoding declarations.

bowlofeggs:
Now that we are purely Python 3, UTF-8 is the default encoding for
source code and we no longer need these encoding headers.


cverna (approved):

bowlofeggs:
This one seems to have failed due to a network problem with Fedora's mirror network or metalink server.

PR #3112, opened 2019-03-28
Alphabetize the deps in the Rawhide Dockerfile.

bowlofeggs:


pypingou (approved):

bowlofeggs:
Weird, this one seems to have failed because dnf says that python3-sphinx failed to install. I'll rebase and see if it works.

bowlofeggs:
Ah, the f28-integration test hit the timeout. That should be fixed by #3155.

PR #3115, opened 2019-03-28
Remove all leading u's on strings.

bowlofeggs:
Now that Bodhi is purely Python 3, we no longer need the leading
u's on strings that indicated that they were unicode objects in
Python 2.


pypingou (approved):

bowlofeggs:
Looks like quay.io was down, which caused the builds to fail. I've got an open pull request that should retry on build failures: https://github.com/fedora-infra/bodhi/pull/3156

PR #3117, opened 2019-03-29
Use a dockerignore file to speed up CI.

bowlofeggs:
This greatly reduces the size of the build context that we send to
the Docker daemon and helps the builds to go faster, especially
when they are able to cache the early layers.

I needed to slightly alter bodhi-ci as well to get it to bind mount
the .git folder in for the diff-cover test to work.


pypingou (approved):

bowlofeggs:
Hrm, looks like my bind mount trick didn't work for some reason.

bowlofeggs:
Ah, looks like it's hitting a permission issue when run on CentOS CI that it doesn't hit when run on my laptop. Hopefully I can figure it out.

bowlofeggs:
Aha! Looks like it was SELinux - just needed the Z flag on the docker bind mount.

PR #3118, opened 2019-03-29
Only run flake8 and pydocstyle on the pip release.

bowlofeggs:
The flake8 and pydocstyle tests don't make sense to test across
all releases, since they are tests of code quality and not of
integration with the release. Thus, let's make the tests a little
faster by running them only on the pip release tests.


pypingou (approved):

bowlofeggs:
Since this one alters the mergify.yml file it will need to be manually merged. I've rebased it and will run tests locally with ```bodhi-ci``` to ensure they pass before merging.

PR #3119, opened 2019-03-29
Compare enums directly.

bowlofeggs:
This corrects a function that was comparing enum names rather than
comparing enums directly.


pypingou (approved):
simple enough

bowlofeggs:

bowlofeggs:

PR #3120, opened 2019-03-29
Correct a release note about a changed topic.

bowlofeggs:
The release notes said that mashtask.composing was renamed to
compose.composing, but it was actually called mashtask.mashing
before. This commit corrects it.


pypingou (approved):

bowlofeggs:
This one failed due to getting a 503 from fedora's metalink server. https://github.com/fedora-infra/bodhi/pull/3156 should help with problems like that in the future.

PR #3125, opened 2019-04-03
addbuilds and removebuilds

tdawson:
New edit options for bodhi client
-   bodhi updates edit --addbuilds
-   bodhi updates edit --removebuilds
Given a comma seperated list of build nvr's, it adds, or removes, those builds from the update.

### DIFF
--- a/bodhi/client/__init__.py
+++ b/bodhi/client/__init__.py
@@ -439,6 +439,8 @@ def _validate_edit_update(ctx, param, value):
 @updates.command()
 @click.option('--type', help='Update type',
               type=click.Choice(['security', 'bugfix', 'enhancement', 'newpackage']))
+@click.option('--addbuilds', help='Add Comma-separated list of build nvr')
+@click.option('--removebuilds', help='Remove Comma-separated list of build nvr')
 @add_options(new_edit_options)
 @click.argument('update', callback=_validate_edit_update)
 @openid_option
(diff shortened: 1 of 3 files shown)


cverna:
Change LGTM, but you might want to squash your commit to a single commit.

bodhi/client/__init__.py
```diff
@@ -98,6 +98,8 @@ def _set_logging_debug(ctx, param, value):
 new_edit_options = [
     click.option('--user'),
     click.option('--password', hide_input=True),
+    click.option('--addbuilds', help='Add Comma-separated list of build nvr'),
```
cverna: I just realized that this are the shared options between the update new and the update edit command. I think these options are only really useful for edit. So they should probably be added directly to the edit method. What do you think ?

see https://github.com/fedora-infra/bodhi/blob/develop/bodhi/client/__init__.py#L440 for an example.

tdawson: I agree.

cverna:
I am not 100% sure why but it seems that this change broke some existing tests. You can run `pytest-3 -vvv bodhi/tests/client/test___init__.py ` in the vagrant environment to see the failures.

tdawson:
I needed to remove addbuilds and removebuilds from kwargs each time.
I was removing them whenever they were called, which allowed my tests to pass.  But I didn't realize that was the reason all the other tests were now failing.
Fixed.

cverna:
I am sorry I just thought about one last things :man_facepalming: .We will want to update the man page of the bodhi cli with the new options. https://github.com/fedora-infra/bodhi/blob/develop/docs/user/man_pages/bodhi.rst

Other than that I ran the tests locally and everything looked good :+1:

tdawson:
Good catch.
Manpage for bodhi updated.

cverna:
:+1:

bowlofeggs:
In the future, please don't merge pull requests unless the CI passes (unless the pull request is critical, like a security issue or something along those lines). When CI isn't passing, we should focus on fixing that and hold all pull requests from merging until that issue is resolved.

bowlofeggs:

bowlofeggs:

PR #3126, opened 2019-04-03
Publish BodhiMessages instead of dictionaries.

### LINKED ISSUE #3124: Bodhi needs to publish messages using its schemas instead of dictionaries
SSIA - we need this to release 4.0.0, so that we know we are publishing the schemas we document.

bowlofeggs:
Now that we have bodhi.messages.schemas, it is time to start
publishing messages using them instead of using the unschema'd
dictionaries that we have been using until now. This commit adjusts
our message publishing API to accept these objects, and adjusts
all server code to publish with them instead of with dictionaries.

fixes #3124

### DIFF
--- a/bodhi/messages/schemas/base.py
+++ b/bodhi/messages/schemas/base.py
@@ -21,6 +21,7 @@ Each message is defined as a Python class. For details, see `fedora-messaging
 messages.
 """
 
+import json
 import re
 import typing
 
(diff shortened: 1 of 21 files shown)


bowlofeggs:
The unit tests passed, but CentOS CI failed to notify GitHub due to API limits. Sadly, the current CI system does not run integration tests if non-integration tests fail - I opened #3157 to fix that.

bowlofeggs:
This has gone two weeks without review, and I've been unable to convince anyone to review it for me in IRC. I am going to merge it without review as a result.

abompard:
I had started reviewing it but Github says "This branch is out-of-date", so I was waiting for a rebase. I can review it now if you prefer.

bowlofeggs:
@abompard GitHub will say that for any pull request that doesn't have the latest commit from develop since we are using branch protection. This means that merging a pull request will cause all the other pull requests to display that until they are rebased.

PR #3131, opened 2019-04-05
Spelling

jsoref:
Generated by https://github.com/jsoref/spelling `f`; to maintain your repo, please consider `fchurn`

Most of these changes should be non-controversial. I'll try to call out the ones that have more risk...

PR #3133, opened 2019-04-05
document that response is intentionally truncated

jsoref:
This is because the group names come from the Fedora Account System
which is limited to 16 characters.

https://community.aegirproject.org/developing/architecture/unix-group-limitations/index.html

Signed-off-by: Josh Soref <jsoref@users.noreply.github.com>

split from #3131 because `security_respons` is essentially set in stone.


bowlofeggs (approved):
Thanks!

PR #3134, opened 2019-04-05
spelling: surround [API]

jsoref:
Since I can't determine if this is intentional, I'm splitting it out from the rest of #3131


bowlofeggs:
This seems unintentional to me too and the markdown div is still getting put into place, so LGTM. I'll rebase and approve.

bowlofeggs (approved):
Thanks!

bowlofeggs:
This one actually passed non-integration tests, but CentOS CI failed to report to GitHub. This happens sometimes due to GitHub API limits. I'll run the tests locally and merge if they pass.

PR #3135, opened 2019-04-06
Fixed #3041 Removed updateid from update's query parameters

### LINKED ISSUE #3041: Remove updateid from update's query parameters
Since update's query parameters: `updateid` and `alias` refers to the same update field, we can get rid of `updateid`.

armansabyr:


bowlofeggs (changes requested):
Looks like there's a ```datagrepper_url``` reference in ```production.ini``` that we should drop too.

Can you document the change regarding the updateid in the release notes in the backwards incompatible changes section?

Also, can you sign both of your commits for the DCO as described at https://bodhi.fedoraproject.org/docs/developer/index.html#contribution-guidelines ? Thanks!

bowlofeggs:
@armansabyr do you still want to work on this?

bowlofeggs:
Closing due to inactivity.

Issue #3065, opened 2019-03-17
Remove unused settings from config


sebwoj:
It looks like `datagrepper_url`(like many others) setting is not used in Bodhi, so we could remove it from config.py and other config files.

subhamkrai:
@sebwoj hi I'm new to fedora and I would like to work on this and start contributing.

cverna:
@subhamkrai ticket is now assigned to you :-)

subhamkrai:
@cverna thank you.

subhamkrai:
@sebwoj I think this issue has solved. we are ready to close

PR #3136, opened 2019-04-06
#3065 datagrepper_url removed

armansabyr:


bowlofeggs (changes requested):
Hmm, looks like this commit was also in your other pull request. Can you remove it from there? Also, as I noted there, there's also a reference to this setting in ```production.ini``` that we should remove as well.

bowlofeggs:
@armansabyr Do you still want to work on this?

bowlofeggs:
Closing due to inactivity.

PR #3138, opened 2019-04-09
Do not run the integration tests in parallel.

bowlofeggs:


Zlopez (dismissed):
:+1:

cverna (dismissed):

bowlofeggs:
I've added two more commits to address some missing dependencies for CI in the pip and f28 containers.

cverna:
+1 to the 2 new commits

bowlofeggs:
I've added three more commits to this, in the hopes that the collection of all of these will get CI working again. Two are modified versions of commits I cherry-picked from other PRs, and the third gets pip installed using an RPM.

puiterwijk (approved):

PR #3139, opened 2019-04-09
Do not warm up the home page cache in CI.

bowlofeggs:
The home page cache takes a lot of time to build, and running this
in CI takes a lot of time as a result. This commit configures Bodhi
not to warm the cache for the home page upon process start, and as
a result the integration tests run about 58% faster on my laptop.


abompard (approved):
Smart! :-)

cverna (approved):

PR #3140, opened 2019-04-09
Do not crash CI when killing a non-running container.

bowlofeggs:
Sometimes the DB container isn't running, and when the CI tests go
to kill it that raises an exception of its own. When this happens,
we lose the original traceback and the system is harder to debug.

This commit adjusts it so we don't crash if we try to kill a
container that isn't running.


bowlofeggs:
Strange, it seems that we got an EL6 machine from CentOS CI instead of an EL7 machine - that is why the tests failed:

https://bugs.centos.org/view.php?id=16002

pypingou (approved):

PR #3144, opened 2019-04-11
Drop the submitter field from serialized updates.

bowlofeggs:
Bodhi has been adding a non-existing field, submitter, to
serialized versions of updates for compatibility with Bodhi 1.
Bodhi 1 went EOL a long time ago, and this data is redundant with
the update.user.name field that also appears in all serialized
updates.


pypingou (approved):

bowlofeggs:

bowlofeggs:

PR #3145, opened 2019-04-11
Handle thread safety wrt the messaging API in callbacks

abompard:
The message callbacks are run in a thread by Twisted, and must therefore call the messaging API with the proper Twisted wrapper to ensure thread safety.

See: https://fedora-messaging.readthedocs.io/en/stable/consuming.html#synchronous-and-asynchronous-calls

### DIFF
--- a/bodhi/server/consumers/composer.py
+++ b/bodhi/server/consumers/composer.py
@@ -41,6 +41,8 @@ from urllib.request import urlopen
 
 import jinja2
 import fedora_messaging
+from twisted.internet import reactor
+from twisted.internet.threads import blockingCallFromThread
 
 from bodhi.messages.schemas import compose as compose_schemas, update as update_schemas
 from bodhi.server import bugs, initialize_db, buildsys, notifications, mail
@@ -184,8 +186,13 @@ class ComposerHandler(object):
         message = message.body
         resume = message.get('resume', False)
         agent = message.get('agent')
-        notifications.publish(compose_schemas.ComposeStartV1.from_dict(dict(agent=agent)),
-                              force=True)
+        # This callback is run in a thread by Twisted, and must therefore call
+        # the messaging API with the proper Twisted wrapper to ensure thread
+        # safety.
+        blockingCallFromThread(
+            reactor, notifications.publish,
+            compose_schemas.ComposeStartV1.from_dict(dict(agent=agent)),
+            force=True)
 
         results = []
         threads = []
@@ -355,11 +362,14 @@ class ComposerThread(threading.Thread):
 
         log.info('Running ComposerThread(%s)' % self.id)
 
-        notifications.publish(compose_schemas.ComposeComposingV1.from_dict(
-            dict(repo=self.id,
-                 updates=[' '.join([b.nvr for b in u.builds]) for u in self.compose.updates],
-                 agent=self.agent,
-                 ctype=self.ctype.value)),
+        blockingCallFromThread(
+            reactor,
+            notifications.publish,
+            compose_schemas.ComposeComposingV1.from_dict(
+                dict(repo=self.id,
+                     updates=[' '.join([b.nvr for b in u.builds]) for u in self.compose.updates],
+                     agent=self.agent,
+                     ctype=self.ctype.value)),
             force=True,
         )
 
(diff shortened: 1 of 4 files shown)


requirements.txt
```diff
@@ -6,7 +6,7 @@ colander
 cornice>=3.1.0
 dogpile.cache
 pyasn1-modules  # Due to an unfortunate dash in its name, installs break if pyasn1 is installed first
-fedora_messaging
+fedora_messaging>=1.6.0
 feedgen
 jinja2
 kitchen
```
bowlofeggs: Since we directly use twisted now, we should also add twisted to this list.

bodhi/server/consumers/composer.py
```diff
@@ -184,7 +186,12 @@ def __call__(self, message: fedora_messaging.api.Message):
         message = message.body["msg"]
         resume = message.get('resume', False)
         agent = message.get('agent')
-        notifications.publish(topic="composer.start", msg=dict(agent=agent), force=True)
+        # This callback is run in a thread by Twisted, and must therefore call
+        # the messaging API with the proper Twisted wrapper to ensure thread
+        # safety.
+        blockingCallFromThread(
+            reactor, notifications.publish,
+            topic="composer.start", msg=dict(agent=agent), force=True)
```
bowlofeggs: Would it be cleaner if we made ```notifications.publish()``` do this when passed a new option (perhaps a boolean like ```run_in_reactor``` or some better name)? I don't mind this, but it might be easier to remember to do it if it were done in one place and had an easy parameter to select the behaviour.

What do you think? I'm fine with this style if you prefer it.

bowlofeggs (changes requested):
One suggestion/question and one requested change so far.

bodhi/server/consumers/composer.py
abompard: I tried that at first, but I think it's better that way, because if we do it in `notifications.publish()` we'll have to handle the case where notifications are delayed after the DB transaction, so that one more layer down where we'll have to pass that flag. And since it's an issue that only affects message callbacks and is pretty specific to the way message callbacks are run, I thought it's more logical to handle these particularities there.

requirements.txt
abompard: Right, I've added it.

bodhi/server/consumers/composer.py
bowlofeggs: Cool.

abompard:
Rebased.

bowlofeggs (changes requested):
We should add a release note that twisted is now required - I'll do this for you, push and approve. Thanks!

bowlofeggs:
```
diff --git a/docs/user/release_notes.rst b/docs/user/release_notes.rst
index 681d1d67..c4ce122f 100644
--- a/docs/user/release_notes.rst
+++ b/docs/user/release_notes.rst
@@ -77,6 +77,7 @@ Dependency changes
 * six is no longer required for the client or server (:issue:`2759`).
 * ``bodhi-server`` now depends on ``bodhi-messages``.
 * kitchen is no longer required.
+* Twisted is now required (:issue:`3145`).
 
 
 Server upgrade instructions
```

bowlofeggs (approved):
Thanks!

bowlofeggs:
Strange, the f29 build failed with rpmdb corruption issues:

```
14:10:21  [91merror: rpmdb: BDB0060 PANIC: fatal region error detected; run recovery
14:10:21  [0m[91merror: db5 error(-30973) from db->close: BDB0087 DB_RUNRECOVERY: Fatal error, run database recovery
14:10:21  error: rpmdb: BDB0060 PANIC: fatal region error detected; run recovery
14:10:21  error: db5 error(-30973) from db->close: BDB0087 DB_RUNRECOVERY: Fatal error, run database recovery
```

bowlofeggs:
This failure happened in #3158 as well.

bowlofeggs:
I've filed https://bugzilla.redhat.com/show_bug.cgi?id=1705265 about the rpmdb corruption issue.

bowlofeggs:

bowlofeggs:

bowlofeggs:

PR #3149, opened 2019-04-11
Build the integration test containers in parallel.

bowlofeggs:
I recently altered our Jenkins job such that it would not run the
integration tests in parallel. This was to avoid putting so much
load on our Duffy nodes such as to cause the tests to timeout.
However, I also made it so that it does not build the containers
used by the integration tests in parallel anymore.

This commit adjusts the Jenkins job to again run the builds in
parallel.

### DIFF
--- a/devel/ci/cico.pipeline
+++ b/devel/ci/cico.pipeline
@@ -99,6 +99,7 @@ def bodhi_ci = { String release, String command, String context, String args ->
  */
 def test_release = { String release ->
     retry_with_sleep({ bodhi_ci(release, 'build', 'build', '') })
+    retry_with_sleep({ bodhi_ci(release, 'integration-build', 'integration-build', '') })
 
     parallel(
         docs: {bodhi_ci(release, 'docs', 'docs', '--no-build --no-init')},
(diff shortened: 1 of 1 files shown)


cverna (approved):

bowlofeggs:
I'm making a slight adjustment to make it so it does the integration builds before starting the tests to help keep the load down a bit:

```
diff --git a/devel/ci/cico.pipeline b/devel/ci/cico.pipeline
index be098a0f..fe502633 100644
--- a/devel/ci/cico.pipeline
+++ b/devel/ci/cico.pipeline
@@ -99,6 +99,7 @@ def bodhi_ci = { String release, String command, String context, String args ->
  */
 def test_release = { String release ->
     bodhi_ci(release, 'build', 'build', '')
+    bodhi_ci(release, 'integration-build', 'integration-build', '')
 
     parallel(
         docs: {bodhi_ci(release, 'docs', 'docs', '--no-build --no-init')},
@@ -120,8 +121,7 @@ def test_release = { String release ->
         unit: {
             bodhi_ci(release, 'unit', 'unit', '--no-build --no-init')
             bodhi_ci(release, 'diff_cover', 'diff-cover', '--no-build --no-init')
-        },
-        integration_build: {bodhi_ci(release, 'integration-build', 'integration-build', '')}
+        }
     )
 }
```

PR #3150, opened 2019-04-11
Let bodhi-ci manage parallelism for CI jobs.

bowlofeggs:


bowlofeggs:
We probably don't need this anymore now, CI seems more stable these days.

PR #3156, opened 2019-04-15
CI: Retry on build failures.

bowlofeggs:
CI build failures are often due to network issues and not
necessarily due to problems in Bodhi. This commits adjusts the CI
system to try again on build failures up to 4 times with a 60
second sleep in between. This should make our CI a little more
reliable. The con is that real build failures will take longer to
be reported since they will be tried many times.

### DIFF
--- a/devel/ci/cico.pipeline
+++ b/devel/ci/cico.pipeline
@@ -98,7 +98,7 @@ def bodhi_ci = { String release, String command, String context, String args ->
  * @param release The release to test.
  */
 def test_release = { String release ->
-    bodhi_ci(release, 'build', 'build', '')
+    retry_with_sleep({ bodhi_ci(release, 'build', 'build', '') })
 
     parallel(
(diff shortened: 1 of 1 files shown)


nirik:
Why 4 times? perhaps make it configurable and start at 2 and see if that helps, then go up if needed?

Otherwise, looks good to me. +1

bowlofeggs:
It's arbitrary, but iirc the retry method used to do 3 and bumping it up to 5 helped with some past problem that I no longer recall. My goal here is to only have humans look into it if there's a real failure and not a network issue (CI fails frequently due to failures to reach quay.io [this is a frequent problem], 503's from Fedora's metalink server ["medium" frequency], timeouts on mirrors [not too frequent, but it happens], and things like that.) I think it's a worthwhile trade off to have it try a bunch of times before telling a human that there's a problem.

PR #3157, opened 2019-04-16
Run the integration tests when the unit tests fail.

bowlofeggs:
We still want to run the integration tests even when other tests
fail.


abompard (approved):
:+1: I got hit by that too.

PR #3158, opened 2019-04-16
Refactor and rename the met_testing_requirements function.

cverna:
The met_testing_requirements function is really only used to check
if bodhi has already commented on the update that the requirements
have been met. This commit renames the function and remove duplicated
logic.

### DIFF
--- a/bodhi/server/models.py
+++ b/bodhi/server/models.py
@@ -2508,7 +2508,7 @@ class Update(Base):
                 # If we haven't met the stable karma requirements, check if it
                 # has met the mandatory time-in-testing requirements
                 if self.mandatory_days_in_testing:
-                    if not self.met_testing_requirements and \
+                    if not self.has_stable_comment and \
                        not self.meets_testing_requirements:
                         if self.release.id_prefix == "FEDORA-EPEL":
                             flash_notes = config.get('not_yet_tested_epel_msg')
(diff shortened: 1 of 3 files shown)


bowlofeggs (approved):
Thanks, this is much cleaner!

bowlofeggs:
I will rebase to resolve the conflict.

bodhi/server/models.py
```diff
@@ -3321,28 +3321,17 @@ def meets_testing_requirements(self):
         return self.days_in_testing >= num_days
 
     @property
-    def met_testing_requirements(self):
+    def has_stable_comment(self):
         """
-        Return True if the update has already been found to meet requirements in the past.
+        Return whether or bodhi has commented on the update that the requirements have been met.
```
sebwoj: Let's remove 'or' from this sentence and use `Bodhi` instead of `bodhi`

sebwoj (changes requested):

bodhi/server/models.py
bowlofeggs: +1, I'll apply:

```
index 87d37a28..423849c6 100644
--- a/bodhi/server/models.py
+++ b/bodhi/server/models.py
@@ -3323,7 +3323,7 @@ class Update(Base):
     @property
     def has_stable_comment(self):
         """
-        Return whether or bodhi has commented on the update that the requirements have been met.
+        Return whether Bodhi has commented on the update that the requirements have been met.
 
         This is used to determine whether bodhi should add the comment
         about the Update's eligibility to be pushed, as we only want Bodhi
```

bowlofeggs:
Heh, the test failures are due to my rebase - I'll fix that up too.

bowlofeggs:
This should fix it:

```
index 87d37a28..423849c6 100644
--- a/bodhi/server/models.py
+++ b/bodhi/server/models.py
@@ -3323,7 +3323,7 @@ class Update(Base):
     @property
     def has_stable_comment(self):
         """
-        Return whether or bodhi has commented on the update that the requirements have been met.
+        Return whether Bodhi has commented on the update that the requirements have been met.
 
         This is used to determine whether bodhi should add the comment
         about the Update's eligibility to be pushed, as we only want Bodhi
```

bowlofeggs:
Hmm, we seem to have a repeatable [dnf failure](https://ci.centos.org/job/bodhi-pipeline/job/PR-3158/4/execution/node/194/log/):

```
14:50:45  error: rpmdb: BDB0689 Packages page 3008 is on free list with type 7
14:50:45  error: rpmdb: BDB0061 PANIC: Invalid argument
14:50:45  error: db5 error(-30973) from dbcursor->c_put: BDB0087 DB_RUNRECOVERY: Fatal error, run database recovery
14:50:45  error: error(-30973) adding header #546 record
14:50:45  error: rpmdb: BDB0060 PANIC: fatal region error detected; run recovery
14:50:45  error: db5 error(-30973) from db->sync: BDB0087 DB_RUNRECOVERY: Fatal error, run database recovery
```

See also #3145.

bowlofeggs:
I've filed https://bugzilla.redhat.com/show_bug.cgi?id=1705265 about the rpmdb corruption issue.

bowlofeggs:

bowlofeggs:

bowlofeggs:

PR #3159, opened 2019-04-17
Add WaiverDB and Greenwave to development environment

### LINKED ISSUE #3011: [L] We need a functioning Greenwave and WaiverDB in the development environment
It is currently very difficult to develop against Greenwave and WaiverDB. There isn't test data in staging, and so we can't test them there, and we obviously don't want to write waivers to the production WaiverDB  during development. We need to have the Vagrant environment provision both of these, and provide an easy way for Bodhi developers to simulate failing tests and set policies.

sebwoj:
fix #3011

### DIFF
--- a/devel/ansible/roles/bodhi/files/.bashrc
+++ b/devel/ansible/roles/bodhi/files/.bashrc
@@ -36,6 +36,28 @@ function btest {
     blint && bdocs && py.test-3 $@ /home/vagrant/bodhi/bodhi/tests && diff-cover /home/vagrant/bodhi/coverage.xml --compare-branch=develop --fail-under=100
 }
 
+
+function bstartdeps {
+    pushd /tmp;
+    curl -o waiverdb.dump.xz https://infrastructure.fedoraproject.org/infra/db-dumps/waiverdb.dump.xz
+    xz -d --keep --force waiverdb.dump.xz
+    popd;
+    pushd /home/vagrant/bodhi/devel/docker/settings/policies/;
+    curl -o fedora_tmpl.yaml https://infrastructure.fedoraproject.org/cgit/ansible.git/plain/roles/openshift-apps/greenwave/templates/fedora.yaml;
+    jinja2 --format=yaml -o fedora.yaml fedora_tmpl.yaml
+    rm fedora_tmpl.yaml
+    popd;
+    sudo docker-compose -f /home/vagrant/bodhi/devel/docker/compose-services.yml up -d
+}
+
+alias bstopdeps="sudo docker-compose -f /home/vagrant/bodhi/devel/docker/compose-services.yml stop"
+
+function bremovedeps {
+    sudo docker-compose -f /home/vagrant/bodhi/devel/docker/compose-services.yml down --rmi all -v
+    rm -f /tmp/waiverdb.dump*
+    rm -f /home/vagrant/bodhi/devel/docker/settings/policies/*
+}
+
 export BODHI_URL="http://localhost:6543/"
 export BODHI_OPENID_API="https://id.stg.fedoraproject.org/api/v1/"
 export PYTHONWARNINGS="once"
--- a/devel/development.ini.example
+++ b/devel/development.ini.example
@@ -59,6 +59,10 @@ openid_template = {username}.id.stg.fedoraproject.org
 # The cache warming is annoying during development because it takes so long. Let's disable it by default.
 warm_cache_on_start = false
 
+test_gating.required = True
+# WaiverDB and Greenwave in vagrant are provided via docker local services
+waiverdb_api_url = http://bodhi_user:pass@localhost:6544/api/v1.0
+greenwave_api_url = http://localhost:6545/api/v1.0
 
 [server:main]
 use = egg:waitress#main
--- a/devel/docker/compose-services.yml
+++ b/devel/docker/compose-services.yml
@@ -1,0 +1,54 @@
+version: '3'
+services:
+  wdb:
+    image: postgres:latest
+    container_name: waiverdb_database
+    restart: always
+    expose:
+      - "5432"
+    volumes:
+      - ./settings/restore_waiverdb.sh:/docker-entrypoint-initdb.d/restore_db.sh
+      - /tmp/waiverdb.dump:/docker-entrypoint-initdb.d/wdb_pgdata
+    healthcheck:
+      test: ["CMD-SHELL", "pg_isready -h localhost -U waiverdb -d waiverdb"]
+      interval: 5s
+      timeout: 30s
+      retries: 3
+
+  waiverdb:
+    image: "quay.io/factory2/waiverdb:latest"
+    container_name: waiverdb_service
+    entrypoint: /bin/sh -i -c "waiverdb wait-for-db && waiverdb db upgrade && /usr/bin/gunicorn-3 --bind 0.0.0.0:6544 --access-logfile=- --enable-stdio-inheritance waiverdb.wsgi:app"
+    volumes:
+      - ./settings/waiverdb-settings.py:/etc/waiverdb/settings.py
+    ports:
+     - "6544:6544"
+    user: '0'
+    depends_on:
+      - wdb
+    healthcheck:
+      test: ["CMD", "curl", "-f", "http://localhost:6544/healthcheck"]
+      interval: 5s
+      timeout: 30s
+      retries: 3
+
+  greenwave:
+    image: "quay.io/factory2/greenwave:latest"
+    container_name: greenwave_service
+    entrypoint: ["/bin/sh", "-c", "gunicorn-3 --workers 8 --bind 0.0.0.0:6545 --access-logfile=- --error-logfile=- --enable-stdio-inheritance greenwave.wsgi:app"]
+    volumes:
+      - ./settings/greenwave-settings.py:/etc/greenwave/settings.py
+      - ./settings/policies:/etc/greenwave/policies/
+    ports:
+      - "6545:6545"
+    depends_on:
+      - waiverdb
+    healthcheck:
+      test: ["CMD", "curl", "-f", "http://localhost:6545/api/v1.0/about"]
+      interval: 5s
+      timeout: 30s
+      retries: 3
+
+networks:
+  default:
+    driver: bridge
--- a/devel/docker/settings/greenwave-settings.py
+++ b/devel/docker/settings/greenwave-settings.py
@@ -1,0 +1,15 @@
+HOST = '0.0.0.0'
+PORT = 6545
+DEBUG = False
+POLICIES_DIR = '/etc/greenwave/policies/'
+DIST_GIT_BASE_URL = 'https://src.fedoraproject.org'
+DIST_GIT_URL_TEMPLATE = '{DIST_GIT_BASE_URL}/{pkg_namespace}/{pkg_name}/raw/{rev}/f/gating.yaml'
+KOJI_BASE_URL = 'https://koji.fedoraproject.org/kojihub'
+
+SECRET_KEY = 'this-is-only-for-development'
+WAIVERDB_API_URL = 'http://waiverdb:6544/api/v1.0'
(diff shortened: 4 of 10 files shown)


sebwoj:
It looks like WaiverDB is working. To test wdb api calls I'm using FEDORA-2018-2817c1642a or FEDORA-2018-8586a09c0d. Now I'm working with GW.

devel/ansible/roles/bodhi/files/.bashrc
```diff
@@ -36,6 +36,16 @@ function btest {
...
 
+alias bdestroyservices="sudo docker-compose -f /home/vagrant/bodhi/devel/docker/compose-services.yml down --rmi all -v"
+
+function brunservices {
+    pushd /tmp;
+    curl -o waiverdb.dump.xz https://infrastructure.fedoraproject.org/infra/db-dumps/waiverdb.dump.xz
+    xz -d --keep --force waiverdb.dump.xz
+    popd;
+    sudo docker-compose -f /home/vagrant/bodhi/devel/docker/compose-services.yml up -d
+}
```
bowlofeggs, line 51: We could just make this all be part of bstop and bstart - what do you think? I don't mind it being separate, but maybe we should pick a less generic name if we do keep it separate. Maybe bstopdeps and bstartdeps or similar?

devel/docker/compose-services.yml
```diff
@@ -0,0 +1,54 @@
...
+      - waiverdb
+    healthcheck:
+      test: ["CMD", "curl", "-f", "http://localhost:6545/api/v1.0/about"]
+      interval: 5s
+      timeout: 30s
+      retries: 3
+
+networks:
+  default:
+    driver: bridge
```
bowlofeggs, line 54: I seem to recall this file being available for download somewhere - is it possible for us to avoid checking it in to our repo and just using the existing one? I can't remember where it was, but if you also don't know I can try to find out.

devel/docker/settings/greenwave-settings.py
```diff
@@ -0,0 +1,16 @@
+HOST = '0.0.0.0'
+PORT = 6545
+DEBUG = False
+POLICIES_DIR = '/etc/greenwave/policies/'
+DIST_GIT_BASE_URL = 'https://src.fedoraproject.org'
+DIST_GIT_URL_TEMPLATE = '{DIST_GIT_BASE_URL}/{pkg_namespace}/{pkg_name}/raw/{rev}/f/gating.yaml'
+KOJI_BASE_URL = 'https://koji.fedoraproject.org/kojihub'
+BODHI_URL = 'https://bodhi/'
```
bowlofeggs: Should this be:

```suggestion
BODHI_URL = 'http://localhost:6543/'
```

?

devel/docker/settings/policy.yaml
```diff
@@ -0,0 +1,147 @@
...
+  - fedora-30
+  - fedora-29
+  - fedora-28
+  - fedora-27
+  - fedora-26
+decision_context: bodhi_update_push_stable
+blacklist: []
+subject_type: bodhi_update
+rules:
+  - !RemoteRule {}
```
bowlofeggs: I bet we could also download this file from Fedora infra upon vagrant up, so that we don't have to maintain our own copy. In fact, we also have a copy of this for our CI tests - it would probably be best if that file were also downloaded at test run time.

bowlofeggs (changes requested):

devel/docker/compose-services.yml
sebwoj: https://pagure.io/greenwave/blob/master/f/docker/docker-compose.yml

devel/docker/settings/policy.yaml
sebwoj: I noticed that this file is a bit outdated in CI so decided to create new one. But idea with auto-download for docker and CI is a good idea.

devel/docker/settings/greenwave-settings.py
sebwoj: Copy-paste from CI. But now I see that this setting is not used in GW. Removing it.

devel/ansible/roles/bodhi/files/.bashrc
sebwoj: > We could just make this all be part of bstop and bstart - what do you think?

I'm not sure about that. When I test my changes made in Bodhi, I restart(by bstop/bstart/brestart)
Bodhi many times, but external services(GW + WDB) does not need to be stopped/restarted/stopped each time.

> I don't mind it being separate, but maybe we should pick a less generic name if we do keep it separate. Maybe bstopdeps and bstartdeps or similar?

The name of the command should describe what command do, so first command creates services +  downloads db dumps + run dockers, and the second removes db dumps + removed docker services + removes images(to fetch latest each time). I'm ok with "bstopdeps" and "bstartdeps", but let's add another bremovedeps.
So:
- bstartdeps -> will create services +  download db dumps + run dockers
- brestartdeps -> (do we really need it? configuration changes and so on)
- bstopdeps -> will stop docker service
- bremovedeps -> will stop docker service + remove db dumps + remove docker services + removes images

What do you think?
Except above commands, do you think that running external services on vagrant setup (ansible) is a good idea?

bowlofeggs:
On Thu, 2019-05-02 at 15:03 +0000, Sebastian Wojciechowski wrote:
> So:
> bstartdeps -> will create services + download db dumps + run dockers
> brestartdeps -> (do we really need it? configuration changes and so
> on)
> bstopdeps -> will stop docker service
> bremovedeps -> will stop docker service + remove db dumps + remove
> docker services + removes images
> What do you think?

This sounds good to me.

devel/ansible/roles/bodhi/files/.bashrc
bowlofeggs: Yeah I think that argument about frequently restarting Bodhi and not wanting that to spend the time of restarting deps does make sense.

I don't think we need a bremovedeps. I also think it does make sense for the vagrant setup to run the deps by default without the user having to type bstartdeps - we already do that for bodhi itself, bodhi's consumer, rabbitmq, the database, etc.

bowlofeggs: All that said, a bremovedeps is fine since you already have it. Doesn't hurt anything.

devel/docker/compose-services.yml
bowlofeggs: Do you think we can use that file instead of keeping a copy here?

cverna: @sebwoj did you check if we can download that file ? Maybe that would be part of the vagrant provisioning ?

nphilipp:
Since both Randy and Clement are unavailable I'll continue with the review.

nphilipp:
I'll fix the conflicts and rebase as it makes reviewing it easier for me. :wink:

devel/ansible/roles/bodhi/files/.bashrc
nphilipp: Looks to me like this is resolved.

devel/docker/compose-services.yml
nphilipp: @sebwoj can we do this?

devel/docker/settings/greenwave-settings.py
nphilipp: Looks to me like this is resolved.

devel/docker/settings/policy.yaml
nphilipp: Looks to me like this is resolved.

devel/docker/compose-services.yml
sebwoj: Now this file is in the different location:
https://pagure.io/greenwave/blob/master/f/docker-compose.yml
But with this config, we have to use resultsDB(not used by Bodhi).
As far as I remember, ResultsDB preparation(downloading resultsDB DB dump + extacting + loading to docker) takes 30< min + 10 GB extra.
Besides, docker config from pagure GW repo creates GW docker based on development source code, instead of production one.

sebwoj:
@nphilipp Before review, I have to prepare code for parsing`https://infrastructure.fedoraproject.org/cgit/ansible.git/plain/roles/openshift-apps/greenwave/templates/fedora.yaml` before use.

devel/docker/compose-services.yml
nphilipp: Do you think we should continue using this previous version then which doesn't exist in the Greenwave repo anymore?

nphilipp:
> @nphilipp Before review, I have to prepare code for parsing`https://infrastructure.fedoraproject.org/cgit/ansible.git/plain/roles/openshift-apps/greenwave/templates/fedora.yaml` before use.

Should we move this PR back to WIP then?

For my understanding, why do you need to parse this file?

sebwoj:
code snippet:
```
decision_context: bodhi_update_push_testing
blacklist: []
subject_type: koji_build
{% if env == 'staging' %}
rules:
  - !PassingTestCaseRule {test_case_name: dist.rpmdeplint}
  - !RemoteRule {}
{% else %}
rules:
  - !RemoteRule {}
{% endif %}

--- !Policy
id: "taskotron_release_critical_tasks_for_stable"
product_versions:
{% if env == 'staging' %}
  - fedora-31
{% endif %}
  - fedora-30
```
Before use, we have to get rid of this `{% if` elements. As far as I understand the syntax, it is [ansible template](https://docs.ansible.com/ansible/latest/modules/template_module.html)

devel/docker/compose-services.yml
sebwoj: It exists, someone just moved it from one location to another:
FROM:
`https://pagure.io/greenwave/blob/master/f/docker/docker-compose.yml`
TO:
`https://pagure.io/greenwave/blob/master/f/docker-compose.yml`

But I think that we should use our, custom one(this from my PR), because of reasons I mentioned in my previous comment.
My docker-compose.yml config was created based on this from GW repo, but I customized it for Bodhi needs.

sebwoj:
Ready for review.

devel/docker/compose-services.yml
nphilipp: Fair enough.

Zlopez:
Solved conflict and rebased

Zlopez (dismissed):

devel/development.ini.example
```diff
@@ -59,6 +59,10 @@ openid_template = {username}.id.stg.fedoraproject.org
 # The cache warming is annoying during development because it takes so long. Let's disable it by default.
 warm_cache_on_start = false
 
+test_gating.required = True
+# WaiverDB and Greenwave in vagrant are provided via docker local services
+waiverdb_api_url = http://bodhi_user:pass@localhost:6544/api/v1.0
+greenwave_api_url = http://localhost:6545/api/v1.0
```
Zlopez, line 65: I tried this one now locally and I'm getting this json response when I'm trying to access this URL:
```
{"message":"The requested URL was not found on the server.  If you entered the URL manually please check your spelling and try again."}
```

I just did `bstartdeps` and tried to access the URL.

Did I missed something?

sebwoj: Production GW will give you the same response:
`https://greenwave-web-greenwave.app.os.fedoraproject.org/api/v1.0`

Try:
`http://localhost:6545/api/v1.0/about`
`http://localhost:6545/api/v1.0/policies`
`http://localhost:6545/api/v1.0/metrics`

Output should be the same as in production:
`https://greenwave-web-greenwave.app.os.fedoraproject.org/api/v1.0/metrics`
`https://greenwave-web-greenwave.app.os.fedoraproject.org/api/v1.0/about`
`https://greenwave-web-greenwave.app.os.fedoraproject.org/api/v1.0/policies`

Zlopez: In this case, everything looks OK

Zlopez (approved):

pypingou:
The PR has been approved by @Zlopez and CI is green, let's get it in :)

PR #3161, opened 2019-04-17
Document how to clear out composes when testing.

bowlofeggs:
Sometimes the Bodhi release engineer needs to delete composes that
were synchronized to staging from production when testing Bodhi
releases in staging. This commit documents how to do that.


abompard (approved):
Thanks!

PR #3162, opened 2019-04-17
Install psycopg2 from EPEL instead of pip.

bowlofeggs:


abompard (approved):
Yep, seems safer.

PR #3163, opened 2019-04-17
Remove an unnecessary workaround.

bowlofeggs:
This libsolv workaround is no longer needed.


abompard (approved):
:+1:

PR #3164, opened 2019-04-17
Document how to mark builds as signed in stg.

bowlofeggs:
We don't have build signing in staging, so this commit documents
how to fool Bodhi into thinking that a build has been signed so it
can be composed with bodhi-push.


abompard (approved):
Yep, that was it :-)

PR #3167, opened 2019-04-18
Handle possible duplicate messages in the consumers

### LINKED ISSUE #3069: Fedora Messaging may deliver messages multiple times
In Fedora Messaging, messages may get delivered multiple times (in case of deconnection for exemple). The consumers need to handle that.

Ideally the tasks would be idempotent, if not the consumer can check if the task to do has already been done (if the expected results have been produced), or any other method.

abompard:
Fixes #3069

In AMQP and thus Fedora Messaging, messages can be delivered multiple times in the event of a network issue or a crash, for example. This changeset ensures that the consumers are idempotent or will deduplicate messages.

### DIFF
--- a/bodhi/server/consumers/composer.py
+++ b/bodhi/server/consumers/composer.py
@@ -178,6 +178,9 @@ class ComposerHandler(object):
         If there are any security updates in the push, then those repositories
         will be executed before all others.
 
+        Duplicate messages: if a requested compose is already pending or has
+        already started, it will be skipped.
+
         Args:
             message: The message we are processing. This is how we know what compose jobs to run.
         """
@@ -236,6 +239,10 @@ class ComposerHandler(object):
                 composes = [Compose.from_dict(db, c) for c in msg['composes']]
             else:
                 raise ValueError('Unable to process message: {}'.format(msg))
+
+            # Filter out composes that are pending or have started, for example in
+            # case of duplicate messages.
+            composes = [c for c in composes if c.state == ComposeState.requested]
 
             for c in composes:
                 # Acknowledge that we've received the command to run these composes.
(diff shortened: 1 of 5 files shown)


bodhi/server/consumers/composer.py
```diff
@@ -236,6 +239,10 @@ def _get_composes(self, msg: dict):
             else:
                 raise ValueError('Unable to process message: {}'.format(msg))
 
+            # Filter out composes that are pending of have started, for example in
```
bowlofeggs:
```suggestion
            # Filter out composes that are pending or have started, for example in
```

bowlofeggs (changes requested):
Cool, just a tiny typo - I'll fix that, rebase, and then approve for Mergify.

bowlofeggs (approved):
Thanks!

bowlofeggs:
Oooh, Mergify caught an issue where this PR and another PR each passed independently, but fail when merged. Good thing we have it. I'm adding this patch to fix that unit test failure:

```
diff --git a/bodhi/tests/server/consumers/test_composer.py b/bodhi/tests/server/consumers/test_composer.py
index 333b9bcf..1d6c1842 100644
--- a/bodhi/tests/server/consumers/test_composer.py
+++ b/bodhi/tests/server/consumers/test_composer.py
@@ -356,7 +356,7 @@ That was the actual one''' % compose_dir
 
     def test__get_composes_duplicate(self):
         """Test _get_composes() when a duplicate message is received."""
-        msg = self._make_msg().body['msg']
+        msg = self._make_msg().body
         composes = self.handler._get_composes(msg)
         self.assertEqual(len(composes), 1)
         composes = self.handler._get_composes(msg)
```

bowlofeggs:

PR #3168, opened 2019-04-18
Automatic updates from tags

nphilipp:
This allows configuring releases for automatic updates from their respective candidate tag through the new `Release.create_automatic_updates` flag (which is settable via the CLI).

~~This allows configuring automatic updates for releases (e.g. Rawhide, see issue #3007) like this:~~

```
<releasename>.autoupdate.from_tags = tag1[, tag2...]
```

~~The configured release must be added to the database beforehand, otherwise the handler will bail out at some point (when it doesn't find the release in question).~~

As discussed, it also introduces a new enum value, `UpdateType.unspecified` because we don't know of what type the created updates are.

~~Open questions, why I marked this as WIP/draft:~~

- ~~Rawhide is different from branched releases. It doesn't have an associated version number for starters and it will probably be created once and stay in 'pending' forever. The only thing that will change with the current setup is, as I see it, the tag with which Rawhide builds get tagged with. Right now, one would have to give it a random, suitably high version number (e.g. 1000) to not violate the NOT NULL constraint on the DB column and avoid tracebacks from the `version_int` property. Shall we keep it like this, or allow NULL or empty versions (and let `version_int` cope with this)?~~

- ~~While working on this, I had the idea that it may be better to record information that maps tags to releases for automatic updates with the `Release` object instead of the instance configuration, to avoid having to change things in two places at branching (creating the new branched release, updating the Rawhide release in DB, tag->release mapping in configuration). This would require some changes to the DB schema, in the simplest case e.g. a `create_automatic_updates` flag in the release which would express that builds tagged into the respective `dist_tag` get an update created automatically. If we want more flexibility (e.g. from other tags associated with the release, or even more than one tag), it'd probably require more involved additions to the schema. Is doing it in configuration (with the req'd restart of the instance after changes) good enough for now, or shall I work on one of the options to specify this in the DB (and expose it via API)?~~


bodhi/server/models.py
```diff
@@ -790,8 +793,7 @@ def version_int(self):
         Returns:
             int: The version of the release.
         """
-        regex = re.compile(r'\D+(\d+)[CMF]?$')
-        return int(regex.match(self.name).groups()[0])
+        return int(self._version_int_regex.match(self.name).groups()[0])
```
cverna, line 802: I am curious, why this change was needed ?

bodhi/server/consumers/automatic_updates.py
```diff
@@ -0,0 +1,197 @@
...
+
+            rname = m.group('releasename')
+            subkey = m.group('subkey')
+
+            if subkey != 'from_tags':
+                continue
+
+            tags = []
+
+            for t in (x.strip() for x in v.split(',')):
```
cverna: Maybe renaming `t` to `tag` would make this a little bit easier to read. Explicit is better than implicit :-)

bodhi/server/consumers/automatic_updates.py
```diff
@@ -0,0 +1,197 @@
...
+        """
+        if not db_factory:
+            initialize_db(config)
+            self.db_factory = transactional_session_maker()
+        else:
+            self.db_factory = db_factory
+
+    @classmethod
+    def _read_configuration(cls):
+        """Extract release <-> tags information from the configuration once."""
```
cverna: I wonder if we should have this info available from the Release model object. Like we do for `mandatory_days_in_testing` (https://github.com/fedora-infra/bodhi/blob/develop/bodhi/server/models.py#L799).

Or we could directly get the config key for the release that is available in the msg body, any reason not to do that ?

bodhi/server/models.py
nphilipp: It's a (maybe a little over the top) optimization: previously, we manually compiled the regex every time the `version_int` property is accessed. Because the regex is constant, we can just do it once and reuse it.

bodhi/server/consumers/automatic_updates.py
nphilipp: I'll change it.

bodhi/server/consumers/automatic_updates.py
nphilipp: The purpose of this is to be able to find out the release from a tag into which a Koji build is tagged, it pulls that information for all configured releases/tags. I'm not sure how to feasibly do this in the context of a `Release` object, because we don't yet know which release it is when the message of a tagged build comes in. However, I could move this method and the related mappings to `Release` at the class level, then it's available if we need this info elsewhere. What do you think?

cverna: I think we already have the Release - Koji tag mapping in the Release object. For example the candidate_tag (https://github.com/fedora-infra/bodhi/blob/develop/bodhi/server/models.py#L744) could be used. 
For example for the rawhide release the candidate tag would currently be f31 and we branch f31 the release can be updated to use the f32 tag.

We could find which release the build is associated with using this tag.

bowlofeggs:
>     * Rawhide is different from branched releases. It doesn't have an associated version number for starters and it will probably be created once and stay in 'pending' forever. The only thing that will change with the current setup is, as I see it, the tag with which Rawhide builds get tagged with. Right now, one would have to give it a random, suitably high version number (e.g. 1000) to not violate the NOT NULL constraint on the DB column and avoid tracebacks from the `version_int` property. Shall we keep it like this, or allow NULL or empty versions (and let `version_int` cope with this)?

Rawhide does actually have a version number - it's currently 31. I think we will probably just set it to that number, and every time we branch we will raise the number and then create the branched release.

bowlofeggs:
>     * While working on this, I had the idea that it may be better to record information that maps tags to releases for automatic updates with the `Release` object instead of the instance configuration, to avoid having to change things in two places at branching (creating the new branched release, updating the Rawhide release in DB, tag->release mapping in configuration). This would require some changes to the DB schema, in the simplest case e.g. a `create_automatic_updates` flag in the release which would express that builds tagged into the respective `dist_tag` get an update created automatically. If we want more flexibility (e.g. from other tags associated with the release, or even more than one tag), it'd probably require more involved additions to the schema. Is doing it in configuration (with the req'd restart of the instance after changes) good enough for now, or shall I work on one of the options to specify this in the DB (and expose it via API)?

We do map tags to a release in the database today:

https://github.com/fedora-infra/bodhi/blob/5c4cc813979fba7df33e95859d2a9b2a88d0010c/bodhi/server/models.py#L774-L781

Or do you mean something different than that?

bodhi/server/buildsys.py
```diff
@@ -165,25 +179,20 @@ def listPackages(self) -> typing.List[typing.Mapping[str, typing.Union[int, str]
     @multicall_enabled
     def getBuild(self, build='TurboGears-1.0.2.2-2.fc17', other=False, testing=False):
         """Emulate Koji's getBuild."""
+        # needed to test against non-existent builds
+        if 'youdontknowme' in build:
```
bowlofeggs, line 183: heh

bodhi/server/consumers/automatic_updates.py
```diff
@@ -0,0 +1,212 @@
...
+        r'(?P<releasename>\w+)\.autoupdates\.(?P<subkey>(?:\w+\.)*\w+)', re.ASCII)
+
+    _version_number_re = re.compile(r'[^\d]*(?P<vernum>\d+).*', re.ASCII)
+
+    def __init__(self, db_factory: transactional_session_maker = None):
+        """
+        Initialize the Automatic Update Handler.
+
+        Args:
+            db_factory (bodhi.server.util.TransactionalSessionMaker or None): If given, used as the
```
bowlofeggs: We don't need to document the type here since you used the fancy new type hinting above. I'm actually in the middle of a (long) project to convert all of Bodhi to use type hinting instead of docblocks to document types.

```suggestion
            db_factory: If given, used as the
```

bodhi/server/consumers/automatic_updates.py
```diff
@@ -0,0 +1,212 @@
...
+    def rels_to_tags(self) -> typing.Dict[str, typing.List[str]]:
+        """Map release names to tag names."""
+        self._read_configuration()
+        return self._rels_to_tags
+
+    @property
+    def tags_to_rels(self) -> typing.Dict[str, str]:
+        """Map tag names to release names."""
+        self._read_configuration()
+        return self._tags_to_rels
```
bowlofeggs: This seems like it does something very similar to

https://github.com/fedora-infra/bodhi/blob/5c4cc813979fba7df33e95859d2a9b2a88d0010c/bodhi/server/models.py#L851-L875

Can we reuse the existing code here?

production.ini
```diff
@@ -442,6 +442,15 @@ fedora_epel.mandatory_days_in_testing = 14
...
 #f28.post_beta.critpath.min_karma = 2
 
+###
+### Automatic updates for releases
+###
+
+# You can specify that updates are automatically created for a release when
+# builds are tagged with certain tags. This can be used to enable automatic
+# gating tests which are tied to updates.
+#rawhide.autoupdates.from_tags = f31
```
bowlofeggs: I think it might be better to make it so we have a boolean for whether we want to automatically create updates. Then we can assume that the ```Release.candidate_tag``` tag is the tag to watch for new builds. This way the configuration is simpler and we can reuse more code.

bowlofeggs (dismissed):

bodhi/server/consumers/automatic_updates.py
nphilipp: I'll fix this (and reindent).

production.ini
nphilipp: Good, I'd say this resolves all questions I had regarding how to specify from which tags we want to create automatic updates, and into which release. Making it hinge on the candidate tag and a flag rather than being set in configuration makes it less flexible (but this seems an acceptable trade-off) and obviate the code processing configuration.

bodhi/server/consumers/automatic_updates.py
nphilipp: I'm not sure we even need that if we get rid of using configuration to specify which release gets automatic updates from which tag.

nphilipp:
The above is missing the bits allowing to configure automatic updates as a flag on `Release` from the respective candidate tag (instead of instance configuration). I'm working on it.

nphilipp:
The latest two commits should move the configuration of automatic updates from the instance configuration to the Release model, respectively the create_automatic_updates boolean.

nphilipp:
Hmm, looks like something failed in the CI tests.

```
java.lang.IllegalArgumentException: The supplied credentials are invalid to login
	at org.jenkinsci.plugins.pipeline.githubstatusnotification.GitHubStatusNotificationStep.getGitHubIfValid(GitHubStatusNotificationStep.java:263)
	at org.jenkinsci.plugins.pipeline.githubstatusnotification.GitHubStatusNotificationStep.getRepoIfValid(GitHubStatusNotificationStep.java:268)
	at org.jenkinsci.plugins.pipeline.githubstatusnotification.GitHubStatusNotificationStep.access$100(GitHubStatusNotificationStep.java:79)
	at org.jenkinsci.plugins.pipeline.githubstatusnotification.GitHubStatusNotificationStep$Execution.run(GitHubStatusNotificationStep.java:373)
	at org.jenkinsci.plugins.pipeline.githubstatusnotification.GitHubStatusNotificationStep$Execution.run(GitHubStatusNotificationStep.java:355)
	at org.jenkinsci.plugins.workflow.steps.AbstractSynchronousNonBlockingStepExecution$1$1.call(AbstractSynchronousNonBlockingStepExecution.java:47)
	at hudson.security.ACL.impersonate(ACL.java:290)
	at org.jenkinsci.plugins.workflow.steps.AbstractSynchronousNonBlockingStepExecution$1.run(AbstractSynchronousNonBlockingStepExecution.java:44)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)
Finished: FAILURE
```

Is there a way to resubmit?

bowlofeggs:
On Tue, 2019-05-07 at 05:42 -0700, Nils Philippsen wrote:
> Is there a way to resubmit?

Sadly there's not a formal way to get it to re-run. When things like
this happen, I make an empty commit, force push, then remove the empty
commit and force push again.

cverna:
> On Tue, 2019-05-07 at 05:42 -0700, Nils Philippsen wrote: Is there a way to resubmit?
> Sadly there's not a formal way to get it to re-run. When things like this happen, I make an empty commit, force push, then remove the empty commit and force push again.

I think you can also use `git commit --amend --no-edit` and force push

nphilipp:
This latest force push reorders the commits, so the changes to the base test case classes come first (because I need them for my current work).

bodhi/server/consumers/automatic_updates.py
```diff
@@ -0,0 +1,158 @@
...
+    The Bodhi Automatic Update Handler.
+
+    A consumer that listens for messages about tagged builds and creates
+    updates from them.
+    """
+
+    _auto_update_cfg_key_re = re.compile(
+        r'(?P<releasename>\w+)\.autoupdates\.(?P<subkey>(?:\w+\.)*\w+)', re.ASCII)
+
+    _version_number_re = re.compile(r'[^\d]*(?P<vernum>\d+).*', re.ASCII)
```
cverna: Are these 2 regular expression still needed ?

bodhi/server/consumers/automatic_updates.py
```diff
@@ -0,0 +1,158 @@
...
+            if mandatory not in msg:
+                missing.append(mandatory)
+        if missing:
+            raise BodhiException(f"Received incomplete tag message. Missing: {', '.join(missing)}")
+
+        btag = msg['tag']
+        bname = msg['name']
+        bversion = msg['version']
+        brelease = msg['release']
+        bnvr = f'{bname}-{bversion}-{brelease}'
```
cverna: I don't have a strong opinion here. but maybe we don't need the `bname`, `bversion` and `brelease` variable seems they are used only in the f-string to create `bnvr`.  I guess the following would just be as readable
```suggestion
        bnvr = f"msg['name']-msg['version']-msg['release']"
```

bodhi/server/consumers/automatic_updates.py
```diff
@@ -0,0 +1,158 @@
...
+            update = Update(
+                release=rel,
+                builds=[build],
+                notes=f"Automatic update for {bnvr}.",
+                type=UpdateType.unspecified,
+                stable_karma=3,
+                unstable_karma=-3,
+                user=user,
+            )
+            dbsession.add(update)
```
cverna, line 147: Don't we need to call `dbsession.commit()` after the add ?

cverna (commented):
Only a few comments. otherwise that looks fine to me

bodhi/server/consumers/automatic_updates.py
nphilipp: Oops, I'll remove that. :smiley:

bodhi/server/consumers/automatic_updates.py
nphilipp: Yeah, I think I only used it as a shortcut. I'll probably use old-style `str.format()` instead because it's less to type (and easier to read IMO). :wink:

bodhi/server/consumers/automatic_updates.py
nphilipp: You're right.

production.ini
```diff
@@ -442,6 +442,15 @@ fedora_epel.mandatory_days_in_testing = 14
 #f28.post_beta.critpath.num_admin_approvals = 0
 #f28.post_beta.critpath.min_karma = 2
 
+###
+### Automatic updates for releases
+###
+
+# You can specify that updates are automatically created for a release when
```
cverna: I think we don't need this setting any more

devel/development.ini.example
```diff
@@ -15,6 +15,7 @@ f7.status = post_beta
 f7.post_beta.mandatory_days_in_testing = 7
 f7.post_beta.critpath.num_admin_approvals = 0
 f7.post_beta.critpath.min_karma = 2
+rawhide.autoupdates.from_tags = f31
```
cverna: I don't think we need this any more

bodhi/server/models.py
```diff
@@ -759,6 +760,9 @@ class Release(Base):
             associated with this release.
         composed_by_bodhi (bool): The flag that indicates whether the release is composed by
             Bodhi or not. Defaults to True.
+        create_automatic_updates (bool): A flag indicating that updated should
```
cverna, line 763: We need a way to set this boolean using the cli for example. So that when Release engineering creates a new release they can set this boolean.

production.ini
nphilipp: I'll remove this...

devel/development.ini.example
nphilipp: ...and this.

bodhi/server/models.py
nphilipp: I'm working on this one, this needs additions/updates in a couple of places (e.g. tests).

nphilipp:
@bowlofeggs I'll dismiss your initial review, somehow it's still seeing it as unresolved even though all individual items are...

cverna:
This PR will need to be rebase on top of #3232. While testing it I found that we were missing a call to `         buildsys.setup_buildsystem(config) ` to initialize the build system but this is now taken care in #3232.

Otherwise it works :fireworks:

nphilipp:
Rebased on top of #3232.

nphilipp:
@cverna:

> While testing it I found that we were missing a call to ` buildsys.setup_buildsystem(config)` to initialize the build system but this is now taken care in #3232.

It's odd that this didn't cause any tests to fail... would this only matter with a real Koji instance as the buildsys? Should we make `DevBuildSys` fail if its `setup_buildsystem()` method isn't called before we use it from tests?

cverna:
> @cverna:
> 
> > While testing it I found that we were missing a call to ` buildsys.setup_buildsystem(config)` to initialize the build system but this is now taken care in #3232.
> 
> It's odd that this didn't cause any tests to fail... would this only matter with a real Koji instance as the buildsys? Should we make `DevBuildSys` fail if it's `setup_buildsystem()` method isn't called before we use it from tests?

Yes I think it is because we are using `DevBuildSys`, it might be a good idea to make it fail indeed in the tests. But maybe in a different PR ?

nphilipp:
> > @cverna:
> > > While testing it I found that we were missing a call to ` buildsys.setup_buildsystem(config)` to initialize the build system but this is now taken care in #3232.
> > 
> > 
> > It's odd that this didn't cause any tests to fail... would this only matter with a real Koji instance as the buildsys? Should we make `DevBuildSys` fail if it's `setup_buildsystem()` method isn't called before we use it from tests?
> 
> Yes I think it is because we are using `DevBuildSys`, it might be a good idea to make it fail indeed in the tests. But maybe in a different PR ?

Of course, this would be totally out of scope for this one.

nphilipp:
NB: This is based off of #3232 and shares the first two commits with #3228. I'd consider #3232 to be blocking it, but not the latter—it doesn't matter in which order they go in, I'll rebase the respective other accordingly and update its migration script.

nphilipp:
The latest push has these changes:

- Removes changes which add consulting instance configuration for automatic update policy only to remove it later.
- Ensures that the automatically created update is announced on the messaging bus, d'oh. And updates tests to mock sending of such updates.

nphilipp:
...meanwhile, I noticed a couple of things I need to change in this PR which Randy changed in #3232, e.g. concerning the message structure (fedmsg v. fedora-messaging).

nphilipp:
Latest push should address the messaging issues and improve one test.

nphilipp:
...and this round makes one consumer test expect the automatic update handler (in fact, mock it away).

bodhi/tests/server/base.py
```diff
@@ -136,10 +137,11 @@ class BaseTestCase(unittest.TestCase):
         'cors_connect_src': 'http://0.0.0.0:6543',
         'cors_origins_ro': 'http://0.0.0.0:6543',
         'cors_origins_rw': 'http://0.0.0.0:6543',
-        'sqlalchemy.url': DEFAULT_DB
+        'sqlalchemy.url': DEFAULT_DB,
+        'f17.autoupdates.from_tags': 'f17-updates-candidate',
```
cverna: I think this was used in the first implementation of this PR but should not be required anymore :smiley:

bodhi/tests/server/consumers/test_automatic_updates.py
```diff
@@ -0,0 +1,132 @@
...
+        super().setup_method(method)
+
+        self.release = self.create_release('47', create_automatic_updates=True)
+
+        body = {
+            'build_id': 442562,
+            'name': 'colord',
+            'tag_id': 214,
+            'instance': 's390',
+            'tag': 'f17-updates-candidate',
```
cverna, line 54: I would have expected this tag to be `f47-updates-candidate`

bodhi/tests/server/consumers/test_automatic_updates.py
```diff
@@ -0,0 +1,132 @@
...
+        msg = deepcopy(self.sample_message)
+        msg.body['release'] += '.testmissingnvr'
+        with pytest.raises(BodhiException):
+            self.handler(msg)
+
+    # Test messages that should be ignored.
+
+    def test_ignored_tag(self, caplog):
+        """Test messages re: tags not configured for automatic updates."""
+        caplog.set_level(logging.DEBUG)
```
cverna, line 170: The caplog fixture is nice :+1:

cverna:
One more thing we need to manage the `unspecified` type on the UI, it is currently not styled and the text is white on a white background so not ideal :)

I think it just need a new value here --> https://github.com/fedora-infra/bodhi/blob/5d84f0f167e0f5db44035006b630ce63d62b400e/bodhi/server/util.py#L726

Similar to what is done here --> https://github.com/fedora-infra/bodhi/blob/5d84f0f167e0f5db44035006b630ce63d62b400e/bodhi/server/util.py#L784

See image here https://paste.opensuse.org/view/raw/5218800

bodhi/tests/server/base.py
nphilipp: Oops. Will remove.

bodhi/tests/server/consumers/test_automatic_updates.py
nphilipp: Not sure if the '47' release isn't a typo...

bodhi/tests/server/consumers/test_automatic_updates.py
nphilipp: But doesn't work with test classes derived from `unittest.TestCase`, took me a while to figure out...

bodhi/tests/server/base.py
nphilipp: Removed.

bodhi/tests/server/consumers/test_automatic_updates.py
nphilipp: It wasn't, but it's not future-proof either (I expect there to be a Fedora 47 some day and who knows what out test data will look like then). I've changed it to retrieve the F17 release from the test DB and change the create_automatic_updates flag, and only if it doesn't exist, create it.

nphilipp:
The latest push fixes the above issues.

nphilipp:
And I've rebased this on top of #3247 because otherwise I couldn't test the latest commit fixing CSS classes for 'unspecified' updates.

cverna (approved):
:+1:  Looks good to me :fireworks: . Thanks @nphilipp

PR #3170, opened 2019-04-20
Add knowledge of branch freezes to the Release model

### LINKED ISSUE #105: don't allow updates to be stuck "between" repos while in "freeze"
Currently it is possible while in freeze to have update/s which are "stuck" between repos. 
- Maintainer submits an update.
- Update receives enough karma to get to stable before it get pushed to 'updates-testing' 
- Update don't appear in any repo untill freeze is over and it gets pushed to stable repos ('fedora'/'updates')

I think this will improve security (all updates are available in freeze) without manually downloading them from 'koji', also this helps with testing because not everyone downloads packages manually and more people testing things is better.

### LINKED ISSUE #1038: Bodhi should inform about freeze
[Miro Hrončok reported this issue in Bugzilla](https://bugzilla.redhat.com/show_bug.cgi?id=1026321)

> Description of problem:
> When there is any freeze that prevents the update to be pushed in stable, bodhi should inform about > that.
> 
> Steps to Reproduce:
> 1. While there is a freeze, mark an update as stable
> 
> Actual results:
> Bodhi says:
> This update has been submitted for stable by $USERNAME. 
> 
> But the update is not pushed, because of the freeze. When the user doesn't know about the freeze,
> (s)he is confused.
> 
> Expected results:
> Bodhi says:
> This update has been submitted for stable by $USERNAME.
> There is an ongoing freeze, this will be pushed to stable after the freeze is over. $LINK_TO_WIKI

### LINKED ISSUE #1563: Add knowledge of branch freezes to the Release model
We should add the ability for Bodhi to know when a branched Release is frozen. During freeze periods, the masher should avoid mashing updates that are request stable, but it should allow exceptions by looking for blocker bugs. This way, releng doesn't have to run ```bodhi-push``` two times (once for the f<branched>-updates-testing, and again for all other stable releases).

sebwoj:
fix #105
fix #1038
fix #1563

### DIFF
--- a/bodhi/server/services/releases.py
+++ b/bodhi/server/services/releases.py
@@ -30,6 +30,7 @@ from bodhi.server.models import (
     Update,
     UpdateStatus,
     UpdateType,
+    UpdateRequest,
     Build,
     BuildrootOverride,
     Package,
@@ -336,6 +337,22 @@ def save_release(request):
                             request.db,
                             'This update is marked obsolete because '
                             'the {} release is archived.'.format(u.release.name),
+                            author='bodhi',
+                        )
+                # Inform user that update requested for stable
+                # will be pushed to stable after the freeze is over.
+                if k == "state" and v == ReleaseState.frozen and \
+                        r.state != ReleaseState.frozen:
+                    updates = request.db.query(Update).filter(Update.release_id == r.id).filter(
+                        Update.request == UpdateRequest.stable
+                    ).filter(
+                        Update.locked == False
+                    ).all()
+                    for u in updates:
+                        u.comment(
+                            request.db,
+                            'There is an ongoing freeze; this will be pushed to'
+                            ' stable after the freeze is over.',
                             author='bodhi',
                         )
                 setattr(r, k, v)
--- a/bodhi/server/migrations/versions/c60d95eef4f1_add_frozen_release_state.py
+++ b/bodhi/server/migrations/versions/c60d95eef4f1_add_frozen_release_state.py
@@ -1,0 +1,67 @@
+# Copyright (c) 2019 Sebastian Wojciechowski.
+#
+# This file is part of Bodhi.
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+"""
+Add frozen release state.
+
+Revision ID: c60d95eef4f1
+Revises: b1fd856efcf6
+Create Date: 2019-04-20 08:44:52.083675
+"""
+from alembic import op
+from sqlalchemy import exc
+
+
+# revision identifiers, used by Alembic.
+revision = 'c60d95eef4f1'
+down_revision = 'b1fd856efcf6'
+
+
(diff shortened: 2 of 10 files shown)


bodhi/server/services/releases.py
```diff
@@ -338,6 +339,22 @@ def save_release(request):
...
+                        Update.request == UpdateRequest.stable
+                    ).filter(
+                        Update.locked == False
+                    ).all()
+                    for u in updates:
+                        u.comment(
+                            request.db,
+                            'There is an ongoing freeze, this will be pushed to'
+                            ' stable after the freeze is over.',
+                            author='bodhi',
```
bowlofeggs, line 356: Nice touch!

bodhi/server/services/releases.py
```diff
@@ -338,6 +339,22 @@ def save_release(request):
...
+                        r.state != ReleaseState.frozen:
+                    updates = request.db.query(Update).filter(Update.release_id == r.id).filter(
+                        Update.request == UpdateRequest.stable
+                    ).filter(
+                        Update.locked == False
+                    ).all()
+                    for u in updates:
+                        u.comment(
+                            request.db,
+                            'There is an ongoing freeze, this will be pushed to'
```
bowlofeggs:
```suggestion
                            'There is an ongoing freeze; this will be pushed to'
```

bowlofeggs (commented):
@mohanboddu - do you think it would be helpful if ```bodhi-push``` automatically filtered out stable updates for frozen releases when the --builds flag is not used?

sebwoj:
1) I forgot to mention that there are some places where `models.ReleaseState.frozen` could be(or not) inserted, but I'm not sure. Example:
https://github.com/fedora-infra/bodhi/blob/0c382987cc7ba1a08c58d227bb0388472773792d/bodhi/server/scripts/untag_branched.py#L77

2) > @mohanboddu - do you think it would be helpful if bodhi-push automatically filtered out stable updates for frozen releases when the --builds flag is not used?

Yeah, good question.

mohanboddu:
> @mohanboddu - do you think it would be helpful if `bodhi-push` automatically filtered out stable updates for frozen releases when the --builds flag is not used?

@bowlofeggs I dont exactly understand your question, we use --builds flag only to push certain builds to stable (not all eligible stable updates) for the frozen release.

We currently use --request flag to push only testing updates for the frozen release.

bowlofeggs:
@mohanboddu The idea is that you could just run bodhi-push the same during a freeze as you do during other times, and since Bodhi will know about freezes it will avoid pushing updates to stable by default.

sebwoj:
@bowlofeggs What do you think about @nirik idea to check state(daily script) using the following url:
https://infrastructure.fedoraproject.org/infra/ansible/vars/all/Frozen.yaml

We can use also:
https://infrastructure.fedoraproject.org/infra/ansible/vars/all/00-FedoraCycleNumber.yaml

bowlofeggs:
@sebwoj I'm unsure about that. The problem is that it ties upstream Bodhi into a particular way that Infrastructure does Ansible, which won't make sense for other deployments (believe it or not, there are other Bodhi deployments than Fedora's).

It's better to let the Ansible playbook use whatever we do here than it is for us to try to interpret the Ansible playbook's yaml files.

bowlofeggs (commented):
I'm happy to accept this as is. @sebwoj are you ready for this to get merged?

bodhi/server/services/releases.py
sebwoj: Thanks :)

sebwoj:
PR rebased.
I'm ok with current version, bur we should be sure about

> 1. I forgot to mention that there are some places where `models.ReleaseState.frozen` could be(or not) inserted, but I'm not sure. Example:
>    https://github.com/fedora-infra/bodhi/blob/0c382987cc7ba1a08c58d227bb0388472773792d/bodhi/server/scripts/untag_branched.py#L77

bowlofeggs:
@sebwoj to this day, I'm not sure what the untag_branched script is really for. I'm not sure whether it is used or not. So… I think we can ignore that for now.

bowlofeggs (approved):
Thanks!

PR #3171, opened 2019-04-22
Serve message schemas.

### LINKED ISSUE #3083: Bodhi's server needs to serve the message schemas
The message schemas have identifiers in them that are links to Bodhi's server. We need to make these links work so third party consumers can get the schemas.

bowlofeggs:
Bodhi now serves its own message schemas at /message-schemas/v1/
so that the URLs referenced in the schemas will not return HTTP
404 codes.

fixes #3083


cverna (approved):
:+1:

PR #3172, opened 2019-04-23
Disable libvirt qemu user sessions

mattiaverga:
Starting from F30, libvirt defaults to use user sessions for QEMU. This will prevent Bodhi VM to work since we require network customization.
This PR will disable qemu_use_session and restores the previous state.

See https://bugzilla.redhat.com/show_bug.cgi?id=1701755

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -73,6 +73,10 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
         # your development guest becoming corrupted (in which case you should only need to do a
         # vagrant destroy and vagrant up to get a new one).
         domain.volume_cache = "unsafe"
+
+        # Bodhi should not use user sessions since it requires network customization
+        # See https://bugzilla.redhat.com/show_bug.cgi?id=1701755
+        domain.qemu_use_session = false
     end
  end
 end


Vagrantfile
```diff
@@ -73,6 +73,10 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
         # your development guest becoming corrupted (in which case you should only need to do a
         # vagrant destroy and vagrant up to get a new one).
         domain.volume_cache = "unsafe"
+
+        # Bodhi should not use user sessions since it requires network customization
```
bowlofeggs, line 77: Is the network customization the port forwarding we do?

bowlofeggs (commented):
I don't mind the change, though I'd love it if there were a way for us to use the user session instead of the system session. I'll wait to see what you and/or Vit say about the port forwarding. If that's causing it, I wonder if there's a way we can achieve that with ssh instead or something…

Vagrantfile
mattiaverga: That is what I've understood from Vit reply. It makes sense to me that a user session cannot forward ports to the host system.

bowlofeggs (changes requested):
Hmm, this doesn't work with ```vagrant-libvirt-0.0.40-5.fc29.noarch```:

```
$ vagrant up
Bringing machine 'bodhi' up with 'libvirt' provider...
There are errors in the configuration of this machine. Please fix
the following errors and try again:

Libvirt Provider:
* The following settings shouldn't exist: qemu_use_session
```

We will need to find a solution that works for Fedora 28-31 (or at least, 29-31).

I used to have a ```Vagrantfile.example``` instead of a ```Vagrantfile``` that the developer had to copy into place before using it. This made it much easier to customize the guest how the developer liked it (adjusting memory, CPU, things like this, etc.). We could just go back to that and add a comment over this setting saying that Fedora < 30 needs to remove it? Or, we could write some Ruby to detect whether this setting is needed.

mattiaverga:
Since `qemu_use_session` has been introduced with vagrant 2.2, what about a
```
if Vagrant::VERSION >= 2.2
    domain.qemu_use_session = false
end
```
?

bowlofeggs:
On Sat, 2019-05-04 at 01:59 -0700, Mattia Verga wrote:
> Since qemu_use_session has been introduced with vagrant 2.2

Isn't it really a part of vagrant-libvirt, and not Vagrant? Perhaps
there's a way to test for the version of vagrant-libvirt? I don't know
Ruby…

bowlofeggs:
I recently upgraded to Fedora 30, and I was able to use Vagrant without this patch. Are you sure it's needed?

mattiaverga:
I hadn't retried in a while, now I've just removed the old VM and let bodhi recreate it with default settings, it seems to work.
I don't think this PR is needed anymore. Thanks.

PR #3174, opened 2019-04-24
Add a RateLimiter log filter and document it.

### LINKED ISSUE #969: Bodhi should e-mail crash tracebacks to admins
Debugging would be aided by adding some code to Bodhi that sends tracebacks to admins. Here are a few requirements:
1. We need count/rate limiting, so nobody gets 1,000 e-mails in their inbox.
2. The list of e-mail addresses should be a setting.
3. We should set this up for both the web interface and the masher.

bowlofeggs:
In order to send crash e-mails to admins with Bodhi 4, we need a
way to rate limit the e-mails so we don't send too many repeat
messages about the same problem. This commit introduces
bodhi.server.logging.RateLimiter, which can be used with Python's
logging system to filter messages via the SMTPHandler.

Unfortunately, it is not possible to use logging filters with
Python's configparser logging configs, which is what Pyramid uses
for logging configuration. Due to this, this commit also adds
documentation describing how deployments can use pyramid_sawing to
move the logging configuration into a YAML file. It also documents
the RateLimiter filter. This allows deployments to use the log
filter with Python's built-in SMTPHandler if they don't mind using
a plugin.

This commit also adjusts all places that were configuring logging
to use a new bodhi.server.logging.setup() function that is able to
use the pyramid_sawing YAML file if it is configured, or is able
to use the regular Pyramid INI configuration if the plugin is not
being used. This way there is one way to configure logging in
Bodhi across all processes.

fixes #969

### DIFF
--- a/bodhi/server/__init__.py
+++ b/bodhi/server/__init__.py
@@ -17,7 +17,7 @@
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
 """Initialize the Bodhi server."""
 from collections import defaultdict
-import logging
+import logging as python_logging
 
 from cornice.validators import DEFAULT_FILTERS
 from dogpile.cache import make_region
@@ -33,7 +33,7 @@ from bodhi.server import bugs, buildsys
 from bodhi.server.config import config as bodhi_config
 
 
-log = logging.getLogger(__name__)
+log = python_logging.getLogger(__name__)
 
 
 #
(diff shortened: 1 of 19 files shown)


bowlofeggs:
For the record, I'm not in love with this approach. It's unfortunate that Pyramid only supports configparser for logging configuration. However, it is at the option of the deployment still, so that's nice.

Here's a review for pyramid_sawing: https://bugzilla.redhat.com/show_bug.cgi?id=1702451

abompard (approved):
Yeah, it's sad that we need to add a dependency, but I guess there's no way around it.

bowlofeggs:

bowlofeggs:

PR #3178, opened 2019-04-25
Annotate bodhi.client.bindings.

bowlofeggs:

### DIFF
--- a/bodhi/client/bindings.py
+++ b/bodhi/client/bindings.py
@@ -33,6 +33,7 @@ import logging
 import os
 import re
 import textwrap
+import typing
 
 try:
     import dnf
@@ -44,6 +45,9 @@ import requests.exceptions
 
 from fedora.client import AuthError, OpenIdBaseClient, FedoraClientError, ServerError
 import fedora.client.openidproxyclient
+
+if typing.TYPE_CHECKING:  # pragma: no cover
+    import munch  # noqa: 401
 
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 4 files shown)


bodhi/client/bindings.py
```diff
@@ -103,10 +118,10 @@ def __unicode__(self):
...
 
 
-def errorhandled(method):
+def errorhandled(method: typing.Callable) -> typing.Callable:
     """Raise exceptions on failure. Used as a decorator for BodhiClient methods."""
     @functools.wraps(method)
-    def wrapper(*args, **kwargs):
+    def wrapper(*args, **kwargs) -> typing.Any:
         try:
             result = method(*args, **kwargs)
```
sebwoj, line 118: Since Bodhi is using Pyhon3.6, and [annotating variables](https://mypy.readthedocs.io/en/latest/cheat_sheet_py3.html#variables) was introduced in Python 3.6, we can use it for variables.
What do you think?

sebwoj (changes requested):

bodhi/client/bindings.py
bowlofeggs: @sebwoj oh nice, I didn't know about that. I don't think we need to annotate all variables since mypy can infer types in most cases, but I'll adjust the places I used the comment to do it this way.

bowlofeggs: I'll apply this patch:

```
diff --git a/bodhi/client/bindings.py b/bodhi/client/bindings.py
index f77aeeaa..11f85ab5 100644
--- a/bodhi/client/bindings.py
+++ b/bodhi/client/bindings.py
@@ -530,7 +530,7 @@ class BodhiClient(OpenIdBaseClient):
             page (int): Return a specific page of results.
                 (default:None)
         """
-        params = {}  # type: typing.MutableMapping[str, typing.Union[int, str, None]]
+        params: typing.MutableMapping[str, typing.Union[int, str, None]] = {}
         if user:
             params['user'] = user
         if packages:
```

bowlofeggs:
Strange, the f30 diff-cover test failed to be able to read ```.git```, even though it is passed with the -Z flag, and even though all the other diff-cover tests were able to read it.

bowlofeggs:
I opened https://github.com/fedora-infra/bodhi/pull/3211 to see if it would help with the sharing of the ```.git``` folder.

Zlopez (approved):
It's nice to see type hints used that much.

PR #3179, opened 2019-04-26
Eliminate __unicode__() methods.

bowlofeggs:
With Python 3 we do not need these anymore.


puiterwijk (approved):
LGTM

PR #3180, opened 2019-04-26
Test Fedora Messaging messages in CI

abompard:
This PR adds the infrastructure to send and receive Fedora Messaging messages in the CI environment, and test the messages that went through.


abompard:
What can be added to this:
- add a wrapper to send messages from the CI tests
- add tests for the consumers

abompard:
It looks like Jenkins is not running the integration tests, I don't know why. I'm looking into it but if you have an idea I'm all ears.

abompard:
Ah, that's because of #3157

abompard:
OK, the failures we have now are related to the DNF API breakage.

devel/ci/integration/rabbitmq/rabbitmqadmin
```diff
@@ -0,0 +1,1050 @@
...
+else:
+    from configparser import ConfigParser, NoSectionError
+    import http.client as httplib
+    import urllib.parse as urlparse
+    from urllib.parse import quote_plus
+    from urllib.parse import quote
+    def b64(s):
+        return base64.b64encode(s.encode('utf-8')).decode('utf-8')
+
+VERSION = '3.6.16'
```
bowlofeggs: Can we link to where this file comes from?

bowlofeggs: Or, would it make more sense to package it as an RPM, or maybe to just have the CI script download it from an authoritative source?

devel/ci/integration/tests/fixtures/rabbitmq.py
```diff
@@ -0,0 +1,70 @@
+# Copyright © 2018 Red Hat, Inc.
```
bowlofeggs:
```suggestion
# Copyright © 2019 Red Hat, Inc.
```

devel/ci/integration/tests/fixtures/rabbitmq.py
```diff
@@ -0,0 +1,70 @@
...
+@pytest.fixture(scope="session")
+def rabbitmq_container(docker_backend, docker_network):
+    """Fixture preparing and yielding a RabbitMQ container.
+
+    Args:
+        docker_backend (conu.DockerBackend): The Docker backend (fixture).
+        docker_network (str): The Docker network ID (fixture).
+
+    Yields:
+        conu.DockerContainer: The RabbitMQ container.
```
bowlofeggs: For new code, I'd like us to start using type annotations instead of docblocks to mark the types of things. So the above would be like this:

```
def rabbitmq_container(docker_backend: conu.DockerBackend, docker_network: str) -> conu.DockerContainer:
    """
    Fixture preparing and yielding a RabbitMQ container.

    Args:
        docker_backend: The Docker backend (fixture).
        docker_network: The Docker network ID (fixture).
    Yields:
        The RabbitMQ container.
```

devel/ci/integration/tests/fixtures/rabbitmq.py
```diff
@@ -0,0 +1,70 @@
...
+        print("Consumer not connected yet, retrying")
+        time.sleep(1)
+    else:
+        raise RuntimeError("The Fedora Messaging consumer did not connect in time")
+    yield container
+    container.kill()
+    container.delete()
+
+
+def _consumer_is_connected(container):
```
bowlofeggs: Let's add a docblock and type annotations here too.

devel/ci/integration/tests/test_messages.py
```diff
@@ -0,0 +1,23 @@
+import time
```
bowlofeggs, line 19: Let's add a license/copyright header on this.

devel/ci/integration/tests/test_messages.py
```diff
@@ -0,0 +1,23 @@
...
+
+def test_push_composer_start(bodhi_container, db_container, rabbitmq_container):
+    try:
+        output = bodhi_container.execute(["bodhi-push", "--username", "ci", "-y"])
+    except ConuException as e:
+        assert False, str(e)
+    output = "".join(line.decode("utf-8") for line in output)
+    assert output.endswith("Sending composer.start message\n")
+    # Give some time for the message to go around
+    time.sleep(2)
```
bowlofeggs, line 34: I wonder if we could use [asyncio.wait()](https://docs.python.org/3.6/library/asyncio-task.html#asyncio.wait) here so that we will wait *up to* a timeout but not always the length of the sleep here? This way we can react as the messages arrive, but also won't hang forever if they don't arrive. I once found a sleep statement in Bodhi that was causing its unit tests to take a good bit longer and got a pretty big boost by dropping it in d26d42a4468f07839ee649f4808c1ed45af7e382 ☺ What do you think?

bowlofeggs (changes requested):

devel/ci/integration/rabbitmq/rabbitmqadmin
abompard: Yeah that can be done, it's actually served from the RabbitMQ server itself.

abompard:
OK I've made some changes in the last commit (to be squashed before merge). I removed `rabbitmqadmin` because it's not actually necessary for what I was using it for.

I don't know how to address the `time.sleep()` issue, though. I don't think tests are run asynchronously in pytest. I could implement a listener as a fixture but that would bring the fedora-messaging stack into the test host, and then it would require tests to be run through Twisted. It may be doable but I'm not sure it's worth it.

abompard:
I rebased the PR, I think this is ready for re-review.
The CI error is caused by an upstream bug I think.

devel/ci/integration/bodhi/fedora-messaging.toml
```diff
@@ -0,0 +1,43 @@
+amqp_url = "amqp://rabbitmq/"
+callback = "bodhi.server.consumers:messaging_callback"
```
cverna:
```suggestion
callback = "bodhi.server.consumers:Consumer"
```
This was changed for bodhi 4.0

cverna:
@abompard  One small change needed otherwise LGTM

abompard:
OK, I implemented your change @cverna , and I rebased and squashed the commits. This should be good for re-review. Thanks.

cverna (approved):
LGTM

devel/ci/integration/tests/test_messages.py
abompard: I'm not sure how to do that. What event will trigger the completion of the future in `wait()`? I don't think we can bubble up that event from the container to the host, where the tests are run. Can we?

bowlofeggs: Yeah I didn't really consider the complexity of detecting changes from within the container.

bowlofeggs (approved):
Thanks!

PR #3182, opened 2019-04-26
Remove some unnecessary casts to str in bindings.

bowlofeggs:


puiterwijk (approved):
+1

PR #3188, opened 2019-04-28
Test builds path in CI

sebwoj:


devel/ci/integration/tests/test_bodhi.py
```diff
@@ -16,6 +16,7 @@
 # along with this program; if not, write to the Free Software
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
 
+import math
```
bowlofeggs, line 19: Since ```math``` is part of the standard library, we should put a blank line between it and ```psycopg2``` for PEP-8.

devel/ci/integration/tests/test_bodhi.py
```diff
@@ -507,3 +508,113 @@ def test_get_overrides_view(bodhi_container, db_container):
         with read_file(bodhi_container, "/httpdir/errorlog") as log:
             print(log.read())
         raise
+
+
+def test_get_build_json(bodhi_container, db_container):
+    """Test ``/builds/{nvr}`` path"""
+    # Fetch build(of latest update) from DB
```
bowlofeggs:
```suggestion
    # Fetch builds (of latest update) from DB
```

devel/ci/integration/tests/test_bodhi.py
```diff
@@ -507,3 +508,113 @@ def test_get_overrides_view(bodhi_container, db_container):
...
+        with conn.cursor() as curs:
+            curs.execute(query_updates)
+            update_id = curs.fetchone()[0]
+            curs.execute(query_builds, (update_id, ))
+            row = curs.fetchone()
+            nvr = row[0]
+            release_id = row[1]
+            signed = row[2]
+            type = row[3]
+            epoch = row[4]
```
bowlofeggs: You could probably express all of the above with one line:

```
nvr, release_id, signed, type, epoch = curs.fetchone()
```

Also, you might want to use a different variable name than ```type```, since that's a Python keyword.

devel/ci/integration/tests/test_bodhi.py
```diff
@@ -507,3 +508,113 @@ def test_get_overrides_view(bodhi_container, db_container):
...
+        print(http_response)
+        print(http_response.text)
+        with read_file(bodhi_container, "/httpdir/errorlog") as log:
+            print(log.read())
+        raise
+
+
+def test_get_builds_json(bodhi_container, db_container):
+    """Test ``/builds`` path"""
+    # Fetch builds(of latest update) from DB
```
bowlofeggs:
```suggestion
    # Fetch builds (of latest update) from DB
```

bowlofeggs (changes requested):

bowlofeggs (approved):
Thanks!

PR #3189, opened 2019-04-29
Correct comments in ci tests

sebwoj:


bowlofeggs:
The integration test actually passed, but CI failed to notify GitHub. I think the diff-cover failure might be fixed by the now merged https://github.com/fedora-infra/bodhi/pull/3211.

bowlofeggs:
Hmm, Mergify seems stuck.

PR #3191, opened 2019-04-29
Convert bodhi.client.bindings to use f-strings.

bowlofeggs:
There are a few spots that still use .format(). Those particular
places are using reusable templates and are thus not practical to
convert to using f-strings.


bowlofeggs:
The rawhide test failure is due to https://github.com/fedora-infra/bodhi/issues/3193

bowlofeggs:
Hmm, so is the f29-unit failure. Hopefully hawkey didn't push out a backwards incompatible change to the stable repositories…

pypingou (approved):

PR #3195, opened 2019-04-30
Annotate bodhi.server.ffmarkdown

sebwoj:

### DIFF
--- a/bodhi/server/ffmarkdown.py
+++ b/bodhi/server/ffmarkdown.py
@@ -22,6 +22,8 @@ Fedora-flavored Markdown.
 Author: Ralph Bean <rbean@redhat.com>
 """
 
+import typing
+
 from markdown.extensions import Extension
 import markdown.inlinepatterns
 import markdown.postprocessors
@@ -30,34 +32,38 @@ import pyramid.threadlocal
 
 from bodhi import MENTION_RE
 
+if typing.TYPE_CHECKING:  # pragma: no cover
+    import re  # noqa: 401
+    import xml  # noqa: 401
+
 
 BUGZILLA_RE = r'([a-zA-Z]+)(#[0-9]{5,})'
 
 
-def user_url(name):
+def user_url(name: str) -> str:
     """
     Return a URL to the given username.
 
     Args:
-        name (basestring): The username of the user we want a URL for.
+        name: The username of the user we want a URL for.
     Returns:
-        basestring: A URL to the requested user.
+        A URL to the requested user.
     """
     request = pyramid.threadlocal.get_current_request()
     return request.route_url('user', name=name)
 
 
-def bug_url(tracker, idx):
+def bug_url(tracker: str, idx: typing.Union[int, str]) -> typing.Optional[str]:
     """
     Return the URL for the given bug.
 
     Args:
-        tracker (basestring): Which bug tracker is being referenced. May be any of 'fedora',
-            'gcc', 'gnome', 'kde', 'mozilla', 'pear', 'perl', 'php', 'python', 'rh', 'rhbz'
-            or 'sourceware'.
-        idx (basestring or int): The bug number.
+        tracker: Which bug tracker is being referenced. May be any of 'fedora',
+            'gcc', 'gnome', 'kde', 'mozilla', 'pear', 'perl', 'php', 'python', 'rh', 'rhbz', or
+            'sourceware'.
+        idx: The bug number.
     Returns:
-        basestring: The URL of the given bug.
+        The URL of the given bug or None if tracker unsupported.
     """
     try:
         trackers = {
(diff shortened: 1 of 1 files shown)


bowlofeggs (commented):
This LGTM - is it still WIP or are you ready to merge?

bodhi/server/ffmarkdown.py
```diff
@@ -34,30 +38,29 @@
...
     """
     Return the URL for the given bug.
 
     Args:
-        tracker (basestring): Which bug tracker is being referenced. May be any of 'fedora',
-            'gcc', 'gnome', 'kde', 'mozilla', 'pear', 'perl', 'php', 'python', 'rh', 'rhbz'
-            or 'sourceware'.
-        idx (basestring or int): The bug number.
+        tracker: Which bug tracker is being referenced. May be any of 'fedora',
+            'gnome', 'kde', 'mozilla', 'pear', 'perl', 'php', 'python', 'rh', or 'rhbz'.
```
bowlofeggs: It looks like gcc and sourceware got dropped from this list, possibly during a rebase?

bodhi/server/ffmarkdown.py
```diff
@@ -22,6 +22,10 @@
 Author: Ralph Bean <rbean@redhat.com>
 """
 
+import re
+import typing
+import xml
```
bowlofeggs: We can move the ```re``` and ```xml``` imports to an ```if typing.TYPE_CHECKING``` block, and then reference them using strings below. This way it's clearer that this module doesn't directly use these libraries and that they are only referenced during type checking.

bodhi/server/ffmarkdown.py
```diff
@@ -83,14 +86,14 @@ def bug_url(tracker, idx):
 class MentionPattern(markdown.inlinepatterns.Pattern):
     """Match username mentions and point to their profiles."""
 
-    def handleMatch(self, m):
+    def handleMatch(self, m: re.Match) -> xml.etree.ElementTree.Element:
```
bowlofeggs:
```suggestion
    def handleMatch(self, m: 're.Match') -> 'xml.etree.ElementTree.Element':
```

bodhi/server/ffmarkdown.py
```diff
@@ -102,14 +105,14 @@ def handleMatch(self, m):
 class BugzillaPattern(markdown.inlinepatterns.Pattern):
     """Match bug tracker patterns."""
 
-    def handleMatch(self, m):
+    def handleMatch(self, m: re.Match) -> xml.etree.ElementTree.Element:
```
bowlofeggs:
```suggestion
    def handleMatch(self, m: 're.Match') -> 'xml.etree.ElementTree.Element':
```

bowlofeggs (changes requested):
Just a couple minor changes - I'll make them and rebase for you.

bowlofeggs:
I'm applying this patch:

```
diff --git a/bodhi/server/ffmarkdown.py b/bodhi/server/ffmarkdown.py
index c2ff2cd1..1d3160e4 100644
--- a/bodhi/server/ffmarkdown.py
+++ b/bodhi/server/ffmarkdown.py
@@ -22,9 +22,7 @@ Fedora-flavored Markdown.
 Author: Ralph Bean <rbean@redhat.com>
 """
 
-import re
 import typing
-import xml
 
 from markdown.extensions import Extension
 import markdown.inlinepatterns
@@ -34,6 +32,10 @@ import pyramid.threadlocal
 
 from bodhi import MENTION_RE
 
+if typing.TYPE_CHECKING:  # pragma: no cover
+    import re  # noqa: 401
+    import xml  # noqa: 401
+
 
 BUGZILLA_RE = r'([a-zA-Z]+)(#[0-9]{5,})'
 
@@ -57,7 +59,8 @@ def bug_url(tracker: str, idx: typing.Union[int, str]) -> typing.Optional[str]:
 
     Args:
         tracker: Which bug tracker is being referenced. May be any of 'fedora',
-            'gnome', 'kde', 'mozilla', 'pear', 'perl', 'php', 'python', 'rh', or 'rhbz'.
+            'gcc', 'gnome', 'kde', 'mozilla', 'pear', 'perl', 'php', 'python', 'rh', 'rhbz', or
+            'sourceware'.
         idx: The bug number.
     Returns:
         The URL of the given bug or None if tracker unsupported.
@@ -86,7 +89,7 @@ def bug_url(tracker: str, idx: typing.Union[int, str]) -> typing.Optional[str]:
 class MentionPattern(markdown.inlinepatterns.Pattern):
     """Match username mentions and point to their profiles."""
 
-    def handleMatch(self, m: re.Match) -> xml.etree.ElementTree.Element:
+    def handleMatch(self, m: 're.Match') -> 'xml.etree.ElementTree.Element':
         """
         Build and return an Element that links to the matched User's profile.
 
@@ -105,7 +108,7 @@ class MentionPattern(markdown.inlinepatterns.Pattern):
 class BugzillaPattern(markdown.inlinepatterns.Pattern):
     """Match bug tracker patterns."""
 
-    def handleMatch(self, m: re.Match) -> xml.etree.ElementTree.Element:
+    def handleMatch(self, m: 're.Match') -> 'xml.etree.ElementTree.Element':
         """
         Build and return an Element that links to the referenced bug.
 

```

bowlofeggs (approved):
Thanks!

PR #3197, opened 2019-04-30
Vagrant: Use the same number of CPUs as the host.

bowlofeggs:
Prior to this commit, the Vagrantfile always gave the guest 8
cores. This would not be ideal for systems with fewer or more than
8 cores. Now we just pick the number of cores that the host has.


pypingou (approved):

bowlofeggs:

bowlofeggs:

PR #3198, opened 2019-04-30
Replace the call to hawkey with a call to DNF in repo sanity check

puiterwijk:
libdnf/hawkey silently dropped the API we were using to do repo sanity checks.
So instead, let's just shell out to DNF and make it their problem.

### DIFF
--- a/bodhi/server/util.py
+++ b/bodhi/server/util.py
@@ -36,7 +36,6 @@ from pyramid.i18n import TranslationStringFactory
 import arrow
 import bleach
 import colander
-import hawkey
 import libcomps
 import libravatar
 import librepo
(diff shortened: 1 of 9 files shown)


bodhi/server/util.py
```diff
@@ -234,6 +235,7 @@ def get_critpath_components(collection='master', component_type='rpm', component
     return critpath_components
 
 
+
```
bowlofeggs: flake8 might get mad about there being 3 lines here instead of 2.

bodhi/server/util.py
```diff
@@ -249,62 +251,110 @@ def sanity_check_repodata(myurl, repo_type):
...
+        # "tests" is a list of tuples with (dnf args, expected output) to run.
+        # For every test, DNF is run with the arguments, and if the expected output is not found,
+        #  an error is raised.
+        tests = []
+
+        if repo_type in ('yum', 'source'):
+            tests.append((['list', 'available'], 'testrepo'))
+        elif repo_type == 'module':
+            tests.append((['module', 'list'], '.*'))
+        # TODO: Add tests that checks for a random build or module and gets its info.
```
bowlofeggs: Let's drop this and file an issue for it - it's a good idea ☺

bodhi/server/util.py
```diff
@@ -249,62 +251,110 @@ def sanity_check_repodata(myurl, repo_type):
...
+
+        for test in tests:
+            dnfargs, expout = test
+
+            # Make sure every DNF test runs in a new temp dir
+            testdir = tempfile.mkdtemp(dir=tmpdir)
+            output = sanity_check_repodata_dnf(testdir, myurl, *dnfargs)
+            if (expout == ".*" and len(output.strip()) != 0) or (expout in output):
+                return
```
bowlofeggs: Do you want to return here, or continue?

bowlofeggs (changes requested):

bodhi/server/util.py
puiterwijk: Yep, it did, and I fixed that :).

bodhi/server/util.py
puiterwijk: continue. This was me being silly.

bodhi/server/util.py
puiterwijk: I wanted to do this before I submitted the PR, but didn't get to it yet. I'll file the issue instead and try to do this tomorrow :).

bowlofeggs (commented):
Sweet - looks good! I will squash and then let mergify do its thing. Thanks so much!

bodhi/server/util.py
puiterwijk: And issue filed as #3199.

bowlofeggs (approved):
Thanks so much!

puiterwijk:
The latest change adds the one missing line of diff cover, and fixes some flake8 errors.
I'm hopeful this'll pass CI.

puiterwijk:
Had forgotten to run flake8 over the test :).
Additionally, newline in docblock added.

bodhi/server/util.py
```diff
@@ -249,62 +248,110 @@ def sanity_check_repodata(myurl, repo_type):
...
+    Because DNF does not provide any usable API they won't break in a year time, we just shell out
+    to run some repository actions on it.
+
+    Args:
+        tempdir (basestring): Temporary directory that will be removed at the end.
+        myurl (basestring): A path to a repodata directory.
+        dnf_args (list): A list of arguments after DNF is set up to use the repo under test.
+            Example would be ["search", "kernel"] or ["module", "list"].
+    Raises:
+        Exception: If the repodata is not valid or does not exist.
```
sebwoj, line 343: Actually, [function](https://docs.python.org/3/library/subprocess.html#subprocess.check_output) is returning [CalledProcessError](https://docs.python.org/3/library/subprocess.html#subprocess.CalledProcessError)

sebwoj (changes requested):

bodhi/server/util.py
puiterwijk: For now that's the only option, yes. But I have plans to add more checks, which might return other exceptions.
And the current docblock is technically correct:

```
>>> issubclass(subprocess.CalledProcessError, Exception)
True
```

bowlofeggs:

bowlofeggs:

PR #3200, opened 2019-04-30
Bodhi updates on Greenwave message

tdawson:
Bodhi updates updates_test_gating_status when Greenwave sends a message saying it's ready.

### DIFF
--- a/bodhi/server/consumers/greenwave.py
+++ b/bodhi/server/consumers/greenwave.py
@@ -1,0 +1,66 @@
+# Copyright © 2019 Red Hat, Inc.
+#
+# This file is part of Bodhi.
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along with
+# this program; if not, write to the Free Software Foundation, Inc., 51
+# Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+"""
+The "greenwave handler".
+
+This module is responsible for listening for messages from greenwave.
+It then updates the policies of the build that greenwave checked.
+"""
+
+import logging
+
+import fedora_messaging
+
+from bodhi.server.models import Build
+from bodhi.server.util import transactional_session_maker
+from bodhi.server.exceptions import BodhiException
+
+log = logging.getLogger(__name__)
+
+
+class GreenwaveHandler:
+    """
+    The Bodhi Greenwave Handler.
+
+    A fedora-messaging listener waiting for messages from greenwave about enforced policies.
+    """
+
+    def __init__(self):
+        """Initialize the GreenwaveHandler."""
+        self.db_factory = transactional_session_maker()
+
+    def __call__(self, message: fedora_messaging.api.Message):
+        """Handle messages arriving with the configured topic."""
+        msg = message.body
+        if not msg:
+            log.debug("Ignoring message without body.")
+            return
+
+        subject_identifier = msg.get("subject_identifier")
+
+        if subject_identifier is None:
+            raise BodhiException("Couldn't find subject_identifier in the Greenwave message")
+
+        with self.db_factory():
+
+            build = Build.get(subject_identifier)
+            if build is None:
+                raise BodhiException(f"Couldn't find build {subject_identifier} in DB")
+
+            log.info(f"Updating the test_gating_status for: {build.update.alias}")
+            build.update.update_test_gating_status()
--- a/bodhi/tests/server/consumers/test_consumers.py
+++ b/bodhi/tests/server/consumers/test_consumers.py
@@ -146,3 +146,14 @@ class TestConsumer(base.BaseTestCase):
         Consumer()(msg)
 
         handler.assert_called_once_with(msg)
+
+    @mock.patch('bodhi.server.consumers.GreenwaveHandler')
+    def test_messaging_callback_greenwave(self, Handler):
+        msg = Message(
+            topic="org.fedoraproject.prod.greenwave.decision.update",
+            body={}
+        )
+        handler = mock.Mock()
+        Handler.side_effect = lambda: handler
+        Consumer()(msg)
+        handler.assert_called_once_with(msg)
--- a/bodhi/tests/server/consumers/test_greenwave.py
+++ b/bodhi/tests/server/consumers/test_greenwave.py
@@ -1,0 +1,102 @@
+# Copyright © 2019 Red Hat, Inc.
+#
+# This file is part of Bodhi.
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+"""This test suite contains tests for the bodhi.server.consumers.greenwave module."""
+
+from unittest import mock
+
+from fedora_messaging.api import Message
+
+from bodhi.server import models
+from bodhi.server.consumers import greenwave
+from bodhi.server.config import config
+from bodhi.server.exceptions import BodhiException
+from bodhi.tests.server.base import BaseTestCase, TransactionalSessionMaker
+
+
+class TestGreenwaveHandler(BaseTestCase):
+    """Test class for the :func:`GreenwaveHandler` method."""
+
+    def setUp(self):
+        super().setUp()
+        self.sample_message = Message(
+            topic="org.fedoraproject.prod.greenwave.decision.update",
+            body={"subject_identifier": "bodhi-2.0-1.fc17"},
+        )
+        self.handler = greenwave.GreenwaveHandler()
+
+    @mock.patch.dict(config, [('greenwave_api_url', 'http://domain.local')])
+    def test_update_test_gating_status(self):
+        """
+        Assert that a greenwave message updates the gating tests status of an update.
+        """
+
+        self.handler.db_factory = TransactionalSessionMaker(self.Session)
+        update = self.db.query(models.Update).filter(
+            models.Build.nvr == 'bodhi-2.0-1.fc17').one()
+
+        # before the greenwave consumer run the gating tests status is None
+        assert update.test_gating_status is None
+
+        with mock.patch('bodhi.server.models.util.greenwave_api_post') as mock_greenwave:
+            greenwave_response = {
+                'policies_satisfied': True,
+                'summary': "all tests have passed"
+            }
+            mock_greenwave.return_value = greenwave_response
+            self.handler(self.sample_message)
+
+        # After the consumer run the gating tests status was updated.
+        assert update.test_gating_status == models.TestGatingStatus.passed
+
+    @mock.patch('bodhi.server.consumers.greenwave.log')
+    def test_greenwave_bad_message(self, mock_log):
+        """ Assert that the consumer ignores messages badly formed """
+
+        bad_message = Message(topic="", body={})
+        self.handler(bad_message)
+        self.assertEqual(mock_log.debug.call_count, 1)
+        mock_log.debug.assert_called_with("Ignoring message without body.")
+
+    def test_greenwave_message_missing_info(self):
+        """
+        Assert that the consumer raise an exception if we could not find the
+        subject_identifier in the message
+        """
+
+        bad_message = Message(topic="", body={"msg": {}})
+        with self.assertRaises(BodhiException) as context:
+            self.handler(bad_message)
+
+        self.assertIn(
+            "Couldn't find subject_identifier in the Greenwave message", context.exception.args
+        )
+
+    def test_greenwave_wrong_build_nvr(self):
+        """
+        Assert that the consumer raise an exception if we could not find the
+        subject_identifier (build nvr) in the DB.
+        """
+        self.handler.db_factory = TransactionalSessionMaker(self.Session)
+        self.sample_message.body["subject_identifier"] = "notapackage-2.0-1.fc17"
+
+        with self.assertRaises(BodhiException) as context:
+            self.handler(self.sample_message)
+
+        self.assertIn(
+            "Couldn't find build notapackage-2.0-1.fc17 in DB", context.exception.args
+        )
--- a/bodhi/server/consumers/__init__.py
+++ b/bodhi/server/consumers/__init__.py
@@ -36,6 +36,7 @@ except ImportError:  # pragma: no cover
     ComposerHandler = None  # pragma: no cover
 from bodhi.server.consumers.signed import SignedHandler
 from bodhi.server.consumers.updates import UpdatesHandler
+from bodhi.server.consumers.greenwave import GreenwaveHandler
 
 
 log = logging.getLogger('bodhi')
@@ -59,6 +60,7 @@ class Consumer:
             self.composer_handler = None
         self.signed_handler = SignedHandler()
         self.updates_handler = UpdatesHandler()
+        self.greenwave_handler = GreenwaveHandler()
 
     def __call__(self, msg: fedora_messaging.api.Message):  # noqa: D401
         """
@@ -92,6 +94,10 @@ class Consumer:
                or msg.topic.endswith('.bodhi.update.edit'):
                 log.debug('Passing message to the Updates handler')
                 self.updates_handler(msg)
+
+            if msg.topic.endswith('.greenwave.decision.update'):
+                log.debug('Passing message to the Greenwave handler')
+                self.greenwave_handler(msg)
         except Exception as e:
             error_msg = f'{str(e)}: Unable to handle message: {msg}'
             log.exception(error_msg)


tdawson:
I belive the code that reacts to the code is working, but my test is not.
For my test, I am grabbing the update associated with my message, and manually changing the test_gating_status.  This should (and does) change test_gating_passed to False. I then run a message through the function and it attempts to change it back.  I believe it doesn't, but only because I manually changed it in the first place.
When I do these changes manually, I don't get a successful change from normal, to false, and back to normal.
This is why this is still in Draft format while I figure out a correct way to do the test.

bodhi/tests/server/consumers/test_greenwave.py
```diff
@@ -0,0 +1,108 @@
...
+            },
+        )
+        self.handler = greenwave.GreenwaveHandler()
+
+    @mock.patch('bodhi.server.consumers.greenwave.Build')
+    def test_update_test_gating_status(self, mock_build_model):
+        """Assert that messages marking the build as signed updates the database"""
+        build = mock_build_model.get.return_value
+        update = models.Build.get("bodhi-3.13.3-1.fc30").update
+        update.test_gating_status = ""
```
bowlofeggs: This is an enum:

https://github.com/fedora-infra/bodhi/blob/3.14.0/bodhi/server/models.py#L638-L658

bowlofeggs:
```suggestion
        update.test_gating_status = models.TestGatingStatus.waiting
```

bowlofeggs (changes requested):

bowlofeggs:
Make sure you sign the DCO in your commit messages, as described here:

This is why the DCO check has failed.

bodhi/server/consumers/greenwave.py
```diff
@@ -0,0 +1,61 @@
+# Copyright © 2016-2019 Red Hat, Inc.
```
bowlofeggs:
```suggestion
# Copyright © 2019 Red Hat, Inc.
```

bodhi/server/consumers/greenwave.py
```diff
@@ -0,0 +1,61 @@
...
+
+import fedora_messaging
+
+from bodhi.server import initialize_db
+from bodhi.server.config import config
+from bodhi.server.models import Build
+from bodhi.server.models import Update
+from bodhi.server.util import transactional_session_maker
+
+log = logging.getLogger('bodhi')
```
bowlofeggs: Let's make the logger be a bit more specific so we know which module emitted it:

```suggestion
log = logging.getLogger(__name__)
```

bodhi/server/consumers/greenwave.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+class GreenwaveHandler(object):
```
bowlofeggs: Since we are Python 3 only, we don't need to subclass ```object``` anymore:

```suggestion
class GreenwaveHandler:
```

bodhi/server/consumers/greenwave.py
```diff
@@ -0,0 +1,61 @@
...
+    """
+    The Bodhi Greenwave Handler.
+
+    A fedora-messaging listener waiting for messages from greenwave about enforced policies.
+    """
+
+    def __init__(self):
+        """Initialize the GreenwaveHandler."""
+        initialize_db(config)
+        self.db_factory = transactional_session_maker()
```
bowlofeggs, line 45: It seems that you might not be using this anywhere - we probably do want to run each message through a database transaction.

bodhi/tests/server/consumers/test_greenwave.py
```diff
@@ -0,0 +1,104 @@
+# Copyright © 2016-2019 Red Hat, Inc.
```
bowlofeggs:
```suggestion
# Copyright © 2019 Red Hat, Inc.
```

bodhi/tests/server/consumers/test_greenwave.py
```diff
@@ -0,0 +1,104 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+"""This test suite contains tests for the bodhi.server.consumers.greenwave module."""
+
+from unittest import mock
+import unittest
+
+from fedora_messaging.api import Message
```
bowlofeggs, line 22: For PEP-8, there should be an empty line between third party imports and Bodhi imports:

```suggestion
from fedora_messaging.api import Message

```

bodhi/tests/server/consumers/test_greenwave.py
```diff
@@ -0,0 +1,104 @@
...
+        self.handler = greenwave.GreenwaveHandler()
+
+    @mock.patch('bodhi.server.consumers.greenwave.Update')
+    def test_update_test_gating_status(self, mock_build_model):
+        """Assert that messages marking the build as signed updates the database"""
+        update = mock_build_model.get.return_value.update
+        update.test_gating_status = models.TestGatingStatus.waiting
+
+        self.handler(self.sample_message)
+        self.assertTrue(update.test_gating_passed)
```
bowlofeggs: Is this test making a live web request to Greenwave? If so, that will not pass in CI (CI disables networking to ensure that the unit tests do not use network resources.)

bowlofeggs (dismissed):

bodhi/server/consumers/greenwave.py
tdawson: Verified this can safely be removed.
Removing.

bodhi/tests/server/consumers/test_greenwave.py
tdawson: I believe it is.
I'm currently researching how to mock that.

tdawson: The live request to Greenwave was made during 
`update.update_test_gating_status()`
I have mock'ed that, and it is no longer doing a live contact to greenwave during the tests.  and the test runs much faster now as well.

bodhi/server/consumers/greenwave.py
```diff
@@ -0,0 +1,55 @@
...
+
+    def __init__(self):
+        """Initialize the GreenwaveHandler."""
+        setup_logging()
+
+    def __call__(self, message: fedora_messaging.api.Message):
+        """Handle messages arriving with the configured topic."""
+        msg = message.body['msg']
+        subject_identifier = msg['subject_identifier']
+        build = Build.get(subject_identifier)
```
cverna: It might be good to manage the case when we don't find a build and raise a BodhiException for example see [here](https://github.com/fedora-infra/bodhi/blob/18740373fea2c50c268f1a96267104f3145ee119/bodhi/server/consumers/automatic_updates.py#L96)

bodhi/server/consumers/greenwave.py
```diff
@@ -0,0 +1,55 @@
...
+        update = Update
```
cverna: Is this needed ?

tdawson: Do you mean just
  update = Update ?
Or do you mean, why don't I just do
  update = Build.get(subject_identifier).update
For both of those, the answer is that btest was having a hissy fit because it said I wasn't using Update, which I have to import, or it doesn't work.

bodhi/server/consumers/greenwave.py
tdawson: Good point.  Working on it.

bodhi/server/consumers/greenwave.py
cverna: I meant `update=Update`, I would try to remove that and remove the `from bodhi.server.models import Update` and see what error you get

tdawson: If I remove both  `update=Update` and  `from bodhi.server.models import Update` then my tests fail with 
`  E           ModuleNotFoundError: No module named 'bodhi.server.consumers.greenwave.Update'; 'bodhi.server.consumers.greenwave' is not a package`

If I remove just the `update=Update` then my tests pass, but btest fails with 
`./bodhi/server/consumers/greenwave.py:30:1: F401 'bodhi.server.models.Update' imported but unused`

bodhi/tests/server/consumers/test_consumers.py
```diff
@@ -69,3 +69,15 @@ def test_messaging_callback_updates_edit(self, Handler):
         Handler.side_effect = lambda: handler
         messaging_callback(msg)
         handler.assert_called_once_with(msg)
+
+    @mock.patch('bodhi.server.consumers.greenwave.Update.update_test_gating_status')
```
cverna: We don't need that mock anymore :-)

bodhi/tests/server/consumers/test_greenwave.py
```diff
@@ -0,0 +1,76 @@
...
+                }
+                print ("MOCK_GREENWAVE START")
+                mock_greenwave.return_value = greenwave_response
+                print ("MOCK_GREENWAVE END")
+                print ("SELF HANDLER START")
+                self.handler(self.sample_message)
+                print ("SELF HANDLER END")
+
+        print ("BUILD START")
+        #build = self.db.query(models.Build).filter_by(builds="bodhi-2.0-1.fc17").first()
```
cverna: We can remove that comment :-)

cverna (changes requested):
I think we need to remove the mock on `     @mock.patch('bodhi.server.consumers.greenwave.Update.update_test_gating_status') ` for the test to pass.

Would be good to also remove the debug print statements :+1:

tdawson:
My comment on this pull request was hidden in the commit.
"Added several print statements to view what values are as tests go along.
I can see that the update is properly being processed, but I can not get the results back to the test."

This commit was not intended to be ready, in fact it fails the tests.  I pushed it up so that others could easily look at it and/or try it in vagrant.

All random print statements, and odd commented out lines will be removed.  Right now it's a work in progress.

tdawson:
Although the code in greenwave.py is working, the tests are not.  The central problem is that the tests don't seem to be able to grab the Update that greenwave.py is working on.
This last commit has comments on what I think is important parts of the code, hopefully allowing someone else to jump in and see the progress and hopefully be able to fix it.

cverna:
I think we will need to rebase this PR once #3232 is merged since it refactor how we call the consumers. Let's wait for that then we can see how to get the tests to pass :-)

tdawson:
Looks very good.  Thank you Clement.
I had to make a few minor changes, mostly whitespace, to make btest happy.
I have also squashed the commits, and gave it a good, long descriptive git commit message.
As long as all the tests pass, which I think they will, I think this might be ready.

bodhi/server/consumers/greenwave.py
```diff
@@ -0,0 +1,66 @@
...
+    def __init__(self):
+        """Initialize the GreenwaveHandler."""
+        self.db_factory = transactional_session_maker()
+
+    def __call__(self, message: fedora_messaging.api.Message):
+        """Handle messages arriving with the configured topic."""
+        msg = message.body.get("msg")
+        if msg is None:
+            log.debug("Ignoring message without 'msg'.")
+            return
```
nphilipp, line 52: With fedora-messaging instead of fedmsg, the interesting contents are directly in the body, see e.g. commit 884a0ccdc9f597e3eeb41442f9d5e71164ffb5e9. I.e., the above would simply be:

```
    msg = message.body
```

Ignoring messages without a `msg` block isn't necessary anymore.

bodhi/server/consumers/greenwave.py
```diff
@@ -0,0 +1,66 @@
...
+            raise BodhiException("Couldn't find subject_identifier in the Greenwave message")
+
+        with self.db_factory():
+
+            build = Build.get(subject_identifier)
+            if build is None:
+                raise BodhiException(f"Couldn't find build {subject_identifier} in DB")
+
+            log.info(f"Updating the test_gating_status for: {build.update.alias}")
+            build.update.update_test_gating_status()
```
nphilipp, line 66: I think this needs to be committed to the database, e.g. like this:

```
    with self.db_factory() as dbsession:
        ...
        build.update.update_test_gating_status()
        dbsession.commit()
```

@cverna?

bodhi/tests/server/consumers/test_greenwave.py
```diff
@@ -0,0 +1,106 @@
...
+
+from bodhi.server import models
+from bodhi.server.consumers import greenwave
+from bodhi.server.config import config
+from bodhi.server.exceptions import BodhiException
+from bodhi.tests.server.base import BaseTestCase, TransactionalSessionMaker
+
+
+class TestGreenwaveHandlerConsume(BaseTestCase):
+    """Test class for the :func:`GreenwaveHandler.consume` method."""
```
nphilipp: There is no `consume()` method :wink:, perhaps just make it "Test the greenwave handler." or "Test the :class:`GreenwaveHandler` class."?

bodhi/tests/server/consumers/test_greenwave.py
```diff
@@ -0,0 +1,106 @@
...
+    def setUp(self):
+        super().setUp()
+        self.sample_message = Message(
+            topic="org.fedoraproject.prod.greenwave.decision.update",
+            body={
+                "msg": {
+                    "subject_identifier": "bodhi-2.0-1.fc17",
+                }
+            },
+        )
```
nphilipp, line 39: Same as above, i.e. make it `... body={"subject_identifier": "bodhi-2.0-1.fc17"}, ...` instead.

bodhi/tests/server/consumers/test_greenwave.py
```diff
@@ -0,0 +1,106 @@
...
+            "Couldn't find subject_identifier in the Greenwave message" in context.exception.args
+        )
+
+    def test_greenwave_wrong_build_nvr(self):
+        """
+        Assert that the consumer raise an exception if we could not find the
+        subject_identifier (build nvr) in the DB.
+        """
+        self.handler.db_factory = TransactionalSessionMaker(self.Session)
+        self.sample_message.body["msg"]["subject_identifier"] = "notapackage-2.0-1.fc17"
```
nphilipp: And the same here, i.e. `... self.sample_message.body["subject_identifier"] = ...`

bodhi/tests/server/consumers/test_greenwave.py
```diff
@@ -0,0 +1,106 @@
...
+
+        with self.assertRaises(BodhiException) as context:
+            self.handler(self.sample_message)
+
+        self.assertTrue(
+            "Couldn't find build notapackage-2.0-1.fc17 in DB" in context.exception.args
+        )
```
nphilipp, line 102: The test class is derived from `unittest.TestCase`, so you can also use `self.assertIn("...", context.exception.args)` here. When my PR #3168 is merged, we can start using test classes not derived from `unittest.TestCase` with pytest which, among other goodies, would let you simply use asserts here, i.e. `assert "..." in ...` (but we're not there yet).

bodhi/server/consumers/greenwave.py
cverna: yes sounds good to me :+1:

tdawson: +1 from me too.
Always good to specifically push/commit/close the object and get everything written before you move onto the next one.  Good catch.

bodhi/server/consumers/greenwave.py
tdawson: Although I do believe you, I'm going to wait until I capture an actual greenwave fedora-message before doing this change.  I thought I was working off a fedora-messager message, but I guess it was fedmsg.

tdawson: After todays meeting, I agree with you.  I will be changing all messaging functions to get things straight from body, and not body['msg']

bodhi/tests/server/consumers/test_greenwave.py
tdawson: Ya ... I think this was a cut and paste from a different test, that I didn't fix up enough.
Fixing.

bodhi/tests/server/consumers/test_greenwave.py
tdawson: Fixing

bodhi/tests/server/consumers/test_greenwave.py
tdawson: Fixing

bodhi/tests/server/consumers/test_greenwave.py
tdawson: I'd like to leave this one as it currently is.  It works and I don't see the benifit of switching to asserIn.
Maybe after your code is merged and we can just use assert, we'll change this one.

bodhi/server/consumers/greenwave.py
tdawson: I'm getting all sorts of `"name 'dbsession' is not defined"`  and `"... object has no attribute 'dbsession' "`
I've tried lots of variations on dbsession (db.session, db_session, db, session)
I've tried having it be from things (self.dbsession, build.dbsession, self.db_factory.dbsession).
And a very good spread of the combination of the above two things.

cverna: @tdawson it is because you need to change the `with self.db_factory():` statement to `with self.db_factory() as dbsession:` then you can use `dbsession` :+1:

bodhi/tests/server/consumers/test_greenwave.py
nphilipp: Sure, I just didn't know if you were aware of it.

bodhi/server/consumers/greenwave.py
tdawson: That works.  Thanks

bowlofeggs:
On Fri, 2019-05-24 at 09:18 -0700, Nils Philippsen wrote:
>     with self.db_factory() as dbsession:
>         ...
>         build.update.update_test_gating_status()
>         dbsession.commit()

You don't need the commit() when you use the context manager - it will
commit for you when you __exit__() the with statement.

bowlofeggs:
On Tue, 2019-05-28 at 13:51 -0700, Troy Dawson wrote:
> I'd like to leave this one as it currently is. It works and I don't
> see the benifit of switching to asserIn.
> Maybe after your code is merged and we can just use assert, we'll
> change this one.

assertIn gives a clearer error message when the assertion fails. Rather
than seeing an error like "false is not true", you will see "this
string" not found in "that string" (or similar). It makes it easier to
debug.

bodhi/tests/server/consumers/test_greenwave.py
tdawson: After readying bowlofeggs comment below, and testing it out, it does give a much better debugging message.  I have changed both of my `self.assertTrue` with `self.assertIn`

bodhi/server/consumers/greenwave.py
tdawson: Although I did add the dbsession.commit(), and things worked, after reading bowlofeggs comment below, it looks like it is just adding a duplicate call.  So I have removed it.

cverna: > Although I did add the dbsession.commit(), and things worked, after reading bowlofeggs comment below, it looks like it is just adding a duplicate call. So I have removed it.

Ha nice did not know that :-)

abompard:
The current patch looks good to me, so +1, but since I'm not the one who has been reviewing this until now, maybe someone else should +1 it too.

bodhi/tests/server/consumers/test_greenwave.py
```diff
@@ -0,0 +1,102 @@
...
+        self.handler = greenwave.GreenwaveHandler()
+
+    @mock.patch.dict(config, [('greenwave_api_url', 'http://domain.local')])
+    def test_update_test_gating_status(self):
+        """
+        Assert that a greenwave message updates the gating tests status of an update.
+        """
+
+        self.handler.db_factory = TransactionalSessionMaker(self.Session)
```
nphilipp, line 48: Just an idea, this initialization of the test DB session could be moved to `TestGreenwaveHandler.setUp()` to be available in all test methods that might use it.

nphilipp (approved):
Looks good to me!

cverna (approved):
I wrote this PR with @tdawson, so I approve these changes :stuck_out_tongue:

abompard (approved):

bodhi/server/consumers/greenwave.py
```diff
@@ -0,0 +1,66 @@
...
+        subject_identifier = msg.get("subject_identifier")
+
+        if subject_identifier is None:
+            raise BodhiException("Couldn't find subject_identifier in the Greenwave message")
+
+        with self.db_factory():
+
+            build = Build.get(subject_identifier)
+            if build is None:
+                raise BodhiException(f"Couldn't find build {subject_identifier} in DB")
```
bowlofeggs, line 63: Raising exceptions here and elsewhere will cause a severe problem: https://github.com/fedora-infra/bodhi/issues/3302

Issue #3062, opened 2019-03-14
Analyze Bodhi's queries to look for opportunities to add indices


bowlofeggs:
I recently discovered that Bodhi wasn't indexing a field that seemed like an obvious one to index, so I wrote https://github.com/fedora-infra/bodhi/commit/285c66a00a3391f712f6ef9e55613f497575b7a9

It is likely that there are many other fields that should be indexed but are not. This is likely a fairly easy way to boost Bodhi's performance, essentially for free. We should analyze some of the common queries that Bodhi performs and look for opportunities to speed them up with indices.

bowlofeggs:
#3201 is planned for inclusion in the upcoming 4.0.0 release: #3221. This should help with performance.

bowlofeggs:
Bodhi 4.0.0 beta is built and deployed to staging:

https://copr.fedorainfracloud.org/coprs/bowlofeggs/bodhi-pre-release
https://bodhi.stg.fedoraproject.org/

PR #3201, opened 2019-04-30
Add an index to comments.update_id.

bowlofeggs:
Many queries will pull in update comments, and it was discovered
that the comments table's update_id column was not indexed. This
means that the comments table was being scanned repeatedly when
updates were being retrieved from the database.

This commit lowered the time to query for all updates from a given
user from 5638.934 ms to 14.911 ms in the Vagrant environment,
just 0.264% of the previous time to run the query.

re #3062


bowlofeggs:
As a bonus, this also makes the integration tests go much faster. They now run in about 3 minutes, where they took about 30 minutes before.

pypingou (approved):
nice

bowlofeggs:

bowlofeggs:

PR #3205, opened 2019-05-01
Create composes based on update's alias in bodhi-push

### LINKED ISSUE #3160: Stable updates push based on update id
During release freezes, Fedora QA will request Fedora RelEng to push some of the updates to push to stable if they fix blockers or FE's. For that we use

```
bodhi-push --builds 'nvr,nvr,...'
```

This is okay if the updates have only one build, but it becomes tiresome if there a multiple updates like [this update](https://bodhi.fedoraproject.org/updates/FEDORA-2019-ac2a21ff07) for example.

Can you make it to pass the update id's to push to stable? Something like

```
bodhi-push --updates 'update_id1,update_id2...'
```

sebwoj:
fix #3160


sebwoj:
@bowlofeggs What do you think?
UT will be next

bodhi/server/push.py
```diff
@@ -50,8 +50,28 @@ def update_sig_status(update):
                 click.echo('Build %s still unsigned' % build.nvr)
 
 
+def check_if_updates_and_builds_set(ctx, param, value):
```
bowlofeggs: For new code, we can start using type annotations for types instead of docblocks - it'll make the conversion easier later. It's also OK to do this for consistency, if you prefer.

bodhi/server/push.py
```diff
@@ -50,8 +50,28 @@ def update_sig_status(update):
...
+    """
+    Print an error to stderr if the user has set both the --updates and --builds flags.
+
+    Args:
+        ctx (click.core.Context): The Click context, used to find out if the other flags are set.
+        param (click.core.Option): The option being handled.
+        value (unicode): The value of the param flag.
+    Returns:
+        unicode: The value of the param flag.
```
bowlofeggs:
```suggestion
        str: The value of the param flag.
```

bodhi/server/push.py
```diff
@@ -50,8 +50,28 @@ def update_sig_status(update):
...
+        (param.name == 'updates' and ctx.params.get('builds', False))):
+            raise click.BadOptionUsage('ERROR: must specify only one of --updates or --builds')
+    return value
+
+
 @click.command()
-@click.option('--builds', help='Push updates for a comma-separated list of builds')
+@click.option('--builds', help='Push updates for a comma-separated list of builds',
+              callback=check_if_updates_and_builds_set)
+@click.option('--updates', help='Push updates for a comma-separated list of update\'s aliases',
```
bowlofeggs:
```suggestion
@click.option('--updates', help='Push updates for a comma-separated list of update aliases',
```

docs/user/man_pages/bodhi-push.rst
```diff
@@ -27,6 +27,10 @@ Options
 
     A comma-separated list of builds to include in the push.
 
+``--updates TEXT``
+
+    A comma-separated list of update's aliases to include in the push.
```
bowlofeggs:
```suggestion
    A comma-separated list of update aliases to include in the push.
```

bodhi/server/push.py
```diff
@@ -50,8 +50,28 @@ def update_sig_status(update):
                 click.echo('Build %s still unsigned' % build.nvr)
 
 
+def check_if_updates_and_builds_set(ctx, param, value):
+    """
+    Print an error to stderr if the user has set both the --updates and --builds flags.
```
bowlofeggs, line 58: It seems to me like it should be OK to mix these flags since they are OR'd together later - is there a problem you noticed when these flags are both used?

bowlofeggs (changes requested):

bodhi/server/push.py
sebwoj: When both flags will be provided, and both flags will be pointing to different updates, bodhi-push will not create any Composes.

bowlofeggs (approved):
Thanks!

PR #3207, opened 2019-05-01
Remove a workaround for an old dnf bug to workaround a new dnf bug.

bowlofeggs:
This commit removes a workaround for a dnf Bug our CI used to hit,
in order to work around a new dnf bug we just started hitting
today[0]. I do not know why removing this line fixes the current
issue, but it does and we don't need this line anymore anyway.

[0] https://bugzilla.redhat.com/show_bug.cgi?id=1705265


nirik:
Looks fine to me.

bowlofeggs:
Thanks @nirik!

PR #3213, opened 2019-05-04
User should be able to set Update.display_name

### LINKED ISSUE #1369: Add a short description field for an update
As discussed in #1351, in an effort to make the titles of updates more useful, it would be great provide a new field tied to an update allowing a package maintainer to add a short description to a multi-build update. With it falling back to the logic for truncating the list outlined in #1351, if a short description is not given. 

As an example the short description could simply be "GNOME", so the label would read:

    enhancement update in Fedora 25 for GNOME

instead of:

    enhancement update in Fedora 25 for nautilus, gnome-shell and 63 others

sebwoj:
fix #1369

### DIFF
--- a/bodhi/client/__init__.py
+++ b/bodhi/client/__init__.py
@@ -113,6 +113,8 @@ new_edit_options = [
                  help='Space or comma-separated list of required Taskotron tasks'),
     click.option('--suggest', help='Post-update user suggestion',
                  type=click.Choice(['logout', 'reboot'])),
+    click.option('--display-name',
+                 help='The name of the update'),
     staging_option]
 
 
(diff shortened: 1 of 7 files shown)


tdawson (commented):
This needs to be rebased.  But from everything I see and have tested, this looks good.

bowlofeggs:
On Tue, 2019-06-04 at 14:36 -0700, Troy Dawson wrote:
> This needs to be rebased.

We don't need to rebase pull requests unless there are conflicts;
Mergify does that for us when it does the merge.

tdawson (approved):
Then this looks good.  Thank you.
I am approving this pull request.

PR #3216, opened 2019-05-05
Correct annotations in server.bugs

sebwoj:


bowlofeggs:
Looks like ```mypy``` gets upset with this patch:

```
bodhi/server/bugs.py:321: error: Incompatible types in assignment (expression has type "FakeBugTracker", variable has type "Optional[Bugzilla]")
```

bowlofeggs (approved):
Thanks!

PR #3217, opened 2019-05-05
Annotate bodhi.server.mail

sebwoj:


bodhi/server/mail.py
```diff
@@ -354,14 +358,14 @@ def get_template(update, use_template='fedora_errata_template'):
     return templates
 
 
-def _send_mail(from_addr, to_addr, body):
+def _send_mail(from_addr: str, to_addr: str, body: str) -> None:
```
bowlofeggs, line 361: I wonder if mypy is able to infer the ```None``` return type when there isn't a ```return``` statement…

bowlofeggs (approved):
Thanks!

PR #3219, opened 2019-05-06
Use greenwave subject_identifier to store missing tests in the UI.

cverna:
This commmit makes use of the subject_identifier (build nvr) in
greenwave response to store the related missing tests.
The missing tests are then displayed in the list of automated test on
the UI.

### DIFF
--- a/bodhi/server/templates/update.html
+++ b/bodhi/server/templates/update.html
@@ -54,16 +54,16 @@ ${update.type} update for ${update.beautify_title(amp=True) | n}
     var handle_unsatisfied_requirements = function(data){
         $.each(data['unsatisfied_requirements'], function(i, requirement) {
             if (requirement.type == 'test-result-missing') {
-                if (missing_tests[requirement.item] === undefined)
-                    missing_tests[requirement.item] = [];
-                missing_tests[requirement.item].push(requirement);
+                if (missing_tests[requirement.subject_identifier] === undefined)
+                    missing_tests[requirement.subject_identifier] = [];
+                missing_tests[requirement.subject_identifier].push(requirement);
             }
             // the user may have already specified this in the required taskotron tests
             if ($.inArray(requirement.testcase, requirements) == -1) {
                 requirements.push(requirement.testcase);
             }
             $('#failed_requirements').append('<li class="list-group-item">' + requirement.testcase + '</li>');
-            
+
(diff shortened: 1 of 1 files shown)


cverna:
This is giving the following results with update FEDORA-2019-20da1ba9e2
image

nphilipp:
I'm not familiar with this part of the Bodhi code or the Greenwave API, so I'm hesitant to review it fully.

Anyway, reading the commit log, I'm not sure if I understand the reason for the change. I understand what `subject_identifier` is, but I can only guess that `item` is some ID with no informational value to the user? Perhaps describe why using the NVR as a key makes the two failing items be shown in the list of automated tests?

cverna:
> I'm not familiar with this part of the Bodhi code or the Greenwave API, so I'm hesitant to review it fully.
> 

I am not that familiar neither :-P

> Anyway, reading the commit log, I'm not sure if I understand the reason for the change. I understand what `subject_identifier` is, but I can only guess that `item` is some ID with no informational value to the user? Perhaps describe why using the NVR as a key makes the two failing items be shown in the list of automated tests?

So the current `requirement.item` is a JS object with key values pair inside, the code that build the row to display the tests results is using the build NVR to access the data from the `missing_test` object. In the current response from greenwave we have the build NVR in `requirement.subject_identifier` hence the change.
My guess is that either this code never worked or that greenwave used to return a build NVR in `requirement.item` and that has changed.

nphilipp:
> So the current `requirement.item` is a JS object with key values pair inside, the code that build the row to display the tests results is using the build NVR to access the data from the `missing_test` object. In the current response from greenwave we have the build NVR in `requirement.subject_identifier` hence the change.
> My guess is that either this code never worked or that greenwave used to return a build NVR in `requirement.item` and that has changed.

That's pretty interesting. Would you put that information into the commit message? Just for reference.

cverna:
> > So the current `requirement.item` is a JS object with key values pair inside, the code that build the row to display the tests results is using the build NVR to access the data from the `missing_test` object. In the current response from greenwave we have the build NVR in `requirement.subject_identifier` hence the change.
> > My guess is that either this code never worked or that greenwave used to return a build NVR in `requirement.item` and that has changed.
> 
> That's pretty interesting. Would you put that information into the commit message? Just for reference.

Done :smile:

nphilipp (approved):
Thanks, looks good to me!

PR #3220, opened 2019-05-07
tables.html  Rename first column in Updates list to Package

abitrolly:
This reflects that fact that the column lists package names.

See https://github.com/fedora-infra/bodhi/pull/2999#discussion_r279444431 for discussion.

Signed-off-by: Anatoli Babenia <anatoli@rainforce.org>

### DIFF
--- a/bodhi/server/templates/tables.html
+++ b/bodhi/server/templates/tables.html
@@ -4,7 +4,7 @@
 <%def name="updates(updates, display_user=True, display_request=True, display_release=True, display_karma=True)">
   <table class="table mb-0 p-1">
     <thead><tr>
-        <th>Package</th>
+        <th>Items</th>
         <th class="nowrap">Age</th>
         %if display_user:
         <th class="nowrap">Submitter</th>
--- a/bodhi/server/templates/tables.html
+++ b/bodhi/server/templates/tables.html
@@ -4,7 +4,7 @@
 <%def name="updates(updates, display_user=True, display_request=True, display_release=True, display_karma=True)">
   <table class="table mb-0 p-1">
     <thead><tr>
-        <th>Update</th>
+        <th>Package</th>
         <th class="nowrap">Age</th>
         %if display_user:
         <th class="nowrap">Submitter</th>


nphilipp:
As I mentioned [here](https://github.com/fedora-infra/bodhi/pull/2999#discussion_r282474926), the builds contained in an update can be something other than RPM packages. I'm not sure how to indicate this better, perhaps use "Update contents" as the column header?

bowlofeggs:
I believe the items shown in this column are actually Update titles. Currently, the title is a list of Build.nvrs, but https://github.com/fedora-infra/bodhi/pull/3213 will make it possible for the user to set their own titles. Thus, I think it's OK that this says it's a list of updates, because it is a list of update titles.

abitrolly:
Now I am sure that this interfaces element needs to be documented. Blender popups is the best example. Here a simple question mark that leads to description page would help.

ryanlerch:
i agree with @bowlofeggs here, that column is a list of update titles. for multi-package updates (as well as being able to set a title), changing this to package would be even more confusing.

abitrolly:
Rebased and renamed the column to `Items`. Item can be:
* package
* module
* container
One table row may container multiple different Items. Is that correct?

ryanlerch:
The updates list page has now changed a fair bit with #3500 it is no longer represented as a table, so there is no longer column headers:

screencapture-0-0-0-0-6543-updates-2019-09-12-15_59_31

abitrolly:
@ryanlerch how does a multipackage update look like in this table?

ryanlerch:
it displays in the list with the contents of the beautify_title() function. e.g.

***package1, package, 2 and XXX more***
or
***package1 and package 2***
or
***package1***
or
the update name, which is a relatively new field in the new update form that allows a packager to make an update name. It is designed to be used for large multi-package updates like GNOME updates, allowing the packager to add something like "GNOME 3.32.1" rather than just showing the first two packages alphabetically and having the rest hidden

abitrolly:
I would color code the type of update rather than status, because color allows to scan the list for specific types of updates quickly. For me it is more interesting that scanning list for packages with specific status.

It also makes sense to me to add automatic header "Multi-package update" if no custom header is set. Then add separate entries for each entry in multipackage update, and in CSV export add a separate column for mutipackage title, which will be empty for normal updates.

pypingou:
> I would color code the type of update rather than status, because color allows to scan the list for specific types of updates quickly. For me it is more interesting that scanning list for packages with specific status.

I actually am more interested in the status than the type so I guess it'll depend on people :)

Should we just close this PR? We can always open a ticket to continue discussing the current UI if we feel the need for it :)
But as it is I don't think we'll merge this PR since the issue it addresses no longer exists :)

abitrolly:
Yes. No need to keep it open.

PR #3223, opened 2019-05-08
Stop producing bodhi-dequeue-stable.

bowlofeggs:
The batching feature was dropped in Bodhi 4.0.0, and the code
backing the bodhi-dequeue-stable CLI was removed, but we forgot to
remove it from the setup.py so the CLI still gets generated.

This commit removes the CLI from setup.py.


cverna (approved):
:+1:

PR #3228, opened 2019-05-14
API: Create/edit updates from Koji tags

nphilipp:
This PR contains code to let users submit new updates or edit existing ones for a Koji (side) tag, i.e. it would pull the latest builds in the tag into the update and remember the tag with the update. For editing (not fully implemented yet), it would work like this: if the list of submitted builds is non-empty, do things just as before, if it is empty, consult Koji again for a list of builds.

NB: it uses the first two commits from PR#3168, so it would be good if this one would be merged first.

### DIFF
--- a/bodhi/server/migrations/versions/d3f8bd499ecd_add_from_tag_column_to_updates.py
+++ b/bodhi/server/migrations/versions/d3f8bd499ecd_add_from_tag_column_to_updates.py
@@ -1,0 +1,41 @@
+# Copyright (c) 2019 Red Hat, Inc.
+#
+# This file is part of Bodhi.
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+"""
+Add from_tag column to updates.
+
+Revision ID: d3f8bd499ecd
+Revises: c60d95eef4f1
+Create Date: 2019-05-14 15:10:54.769789
+"""
+from alembic import op
+import sqlalchemy as sa
+
+
+# revision identifiers, used by Alembic.
+revision = 'd3f8bd499ecd'
+down_revision = 'c60d95eef4f1'
+
+
+def upgrade():
+    """Add the from_tag column to the updates table."""
+    op.add_column('updates', sa.Column('from_tag', sa.UnicodeText(), nullable=True))
+
+
+def downgrade():
+    """Drop the from_tag column from the updates table."""
+    op.drop_column('updates', 'from_tag')
--- a/bodhi/server/models.py
+++ b/bodhi/server/models.py
@@ -1666,6 +1666,8 @@ class Update(Base):
             Greenwave integration was not enabled when the update was created.
         compose (Compose): The :class:`Compose` that this update is currently being composed in. The
             update is locked if this is defined.
+        from_tag (str): The koji tag from which the list of builds was
+            originally populated (if any).
     """
 
     __tablename__ = 'updates'
@@ -1738,6 +1740,9 @@ class Update(Base):
 
     # Greenwave
     test_gating_status = Column(TestGatingStatus.db_type(), default=None, nullable=True)
+
+    # Koji tag, if any, from which the list of builds was populated initially.
+    from_tag = Column(UnicodeText, nullable=True)
 
     def __init__(self, *args, **kwargs):
         """
--- a/bodhi/server/validators.py
+++ b/bodhi/server/validators.py
@@ -29,7 +29,8 @@ import pyramid.threadlocal
 import rpm
 
 from bodhi.server.config import config
-from . import log
+from bodhi.server.exceptions import BodhiException
+from . import buildsys, log
 from .models import (
     Build,
     Bug,
@@ -200,7 +201,7 @@ def validate_build_nvrs(request, **kwargs):
         request (pyramid.request.Request): The current request.
         kwargs (dict): The kwargs of the related service definition. Unused.
     """
-    for build in request.validated.get('builds', []):
+    for build in request.validated.get('builds') or []:  # cope with builds being None
         try:
             cache_nvrs(request, build)
         except ValueError:
@@ -216,6 +217,35 @@ def validate_build_nvrs(request, **kwargs):
 
 
 @postschema_validator
+def validate_builds_or_from_tag_exist(request, **kwargs):
+    """
+    Ensure that at least one of the  builds or from_tags parameters exist in the request.
+
+    Args:
+        request (pyramid.request.Request): The current request.
+        kwargs (dict): The kwargs of the related service definition. Unused.
+    """
+    builds = request.validated.get('builds')
+    from_tag = request.validated.get('from_tag')
+
+    if builds is None and from_tag is None:
+        request.errors.add('body', 'builds,from_tag',
+                           "You must specify either builds or from_tag.")
+
+    if builds is not None:
+        if not isinstance(builds, list):
+            request.errors.add('body', 'builds', "The builds parameter must be a list.")
+        elif len(builds) == 0:
+            request.errors.add('body', 'builds', "You may not specify an empty list of builds.")
+
+    if from_tag is not None:
+        if not isinstance(from_tag, str):
+            request.errors.add('body', 'from_tag', "The from_tags parameter must be a string.")
+        elif len(from_tag.strip()) == 0:
+            request.errors.add('body', 'from_tag', "You may not specify an empty from_tag.")
+
+
+@postschema_validator
 def validate_builds(request, **kwargs):
     """
     Ensure that the builds parameter is valid for the request.
@@ -226,10 +256,7 @@ def validate_builds(request, **kwargs):
     """
     edited = request.validated.get('edited')
     user = User.get(request.user.name)
-
-    if not request.validated.get('builds', []):
-        request.errors.add('body', 'builds', "You may not specify an empty list of builds.")
-        return
+    builds = request.validated.get('builds') or []  # cope with builds set to None
 
     if edited:
         up = request.db.query(Update).filter_by(alias=edited).first()
@@ -246,7 +273,7 @@ def validate_builds(request, **kwargs):
                 request.errors.add('body', 'builds',
                                    'Cannot edit stable updates')
 
-        for nvr in request.validated.get('builds', []):
+        for nvr in builds:
             # Ensure it doesn't already exist in another update
             build = request.db.query(Build).filter_by(nvr=nvr).first()
             if build and build.update is not None and up.alias != build.update.alias:
@@ -255,7 +282,7 @@ def validate_builds(request, **kwargs):
 
         return
 
-    for nvr in request.validated.get('builds', []):
+    for nvr in builds:
         build = request.db.query(Build).filter_by(nvr=nvr).first()
         if build and build.update is not None:
             request.errors.add('body', 'builds',
@@ -287,7 +314,7 @@ def validate_build_tags(request, **kwargs):
     else:
         valid_tags = tag_types['candidate']
 
-    for build in request.validated.get('builds', []):
+    for build in request.validated.get('builds') or []:
         valid = False
         tags = cache_tags(request, build)
         if tags is None:
@@ -1234,3 +1261,49 @@ def validate_request(request, **kwargs):
                         build.package.name, build.evr, target.description, other_build.evr))
                 request.errors.status = HTTPBadRequest.code
                 return
+
+
+@postschema_validator
+def validate_from_tag(request: pyramid.request.Request, **kwargs: dict):
+    """Check that an existing from_tag is valid and set `builds`.
+
+    Ensure that `from_tag` is a valid Koji tag and set `builds` to latest
+    builds from this tag if unset.
+
+    Args:
+        request: The current request.
+        kwargs: The kwargs of the related service definition. Unused.
+    """
+    koji_tag = request.validated.get('from_tag')
+
+    if koji_tag:
+        koji_client = buildsys.get_session()
+
+        if request.validated.get('builds'):
+            # Builds were specified explicitly, verify that the tag exists in Koji.
+            if not koji_client.getTag(koji_tag):
+                request.errors.add('body', 'from_tag', "The supplied from_tag doesn't exist.")
+                return
+
+            # Flag that `builds` wasn't filled from the Koji tag.
+            request.validated['builds_from_tag'] = False
+        else:
+            # Builds weren't specified explicitly, pull the list of latest NVRs here, as it is
+            # necessary for later validation of ACLs pertaining the respective components.
+            try:
+                request.validated['builds'] = [
+                    b['nvr'] for b in koji_client.listTagged(koji_tag, latest=True)
+                ]
+            except koji.GenericError as e:
+                if "invalid taginfo" in str(e).lower():
+                    request.errors.add('body', 'from_tag', "The supplied from_tag doesn't exist.")
+                else:
+                    raise BodhiException("Encountered error while requesting tagged builds from "
+                                         f"Koji: '{e}'") from e
+            else:  # no Koji error, request.validated['builds'] was filled
+                if not request.validated['builds']:
+                    request.errors.add('body', 'from_tag',
+                                       "The supplied from_tag doesn't contain any builds.")
+                else:
+                    # Flag that `builds` was filled from the Koji tag.
+                    request.validated['builds_from_tag'] = True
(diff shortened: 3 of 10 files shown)


nphilipp:
NB: This is based off of #3232 and shares the first two commits with #3168. I'd consider #3232 to be blocking it, but not the latter—it doesn't matter in which order they go in, I'll rebase the respective other accordingly and update its migration script.

nphilipp:
To bring some order into my PRs, this one is now based off of #3168, to avoid having to double-check common commits and migration script ordering between them.

bodhi/server/validators.py
```diff
@@ -315,7 +315,7 @@ def validate_build_tags(request, **kwargs):
     else:
         valid_tags = tag_types['candidate']
 
-    for build in request.validated.get('builds', []):
+    for build in request.validated.get('builds') or []:
```
cverna, line 317: Why this change ?

cverna:
@nphilipp You will probably want to squash your commits :smile:

bodhi/server/migrations/versions/d3f8bd499ecd_add_from_tag_column_to_updates.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+# revision identifiers, used by Alembic.
+revision = 'd3f8bd499ecd'
+down_revision = 'e8a059156d38'
+
+
+def upgrade():
+    """Add the from_tag column to the updates table."""
+    op.add_column('updates', sa.Column('from_tag', sa.UnicodeText(), nullable=True))
```
cverna, line 36: Would it be more explicit to call this `from_sidetag` ? Just plain `from_tag` might let people think that we are talking about releases tags (f30-updates, f30-updates-testing, ... ).

bodhi/server/models.py
```diff
@@ -121,7 +121,7 @@ def __json__(self, request=None):
         Return a JSON representation of this EnumSymbol.
 
         Args:
-            request (pyramid.util.Request): The current request.
+            request (pyramid.request.Request): The current request.
```
cverna: So these changes should be kept in separated commit from the side tag API.

bodhi/server/models.py
```diff
@@ -1602,6 +1602,8 @@ class Update(Base):
             Greenwave integration was not enabled when the update was created.
         compose (Compose): The :class:`Compose` that this update is currently being composed in. The
             update is locked if this is defined.
+        from_tag (unicode): The koji tag from which the list of builds was
```
cverna:
```suggestion
        from_tag (str): The koji tag from which the list of builds was
```
Since Bodhi is Python 3 now the unicode type does not exists anymore :-)

bodhi/server/validators.py
```diff
@@ -191,15 +191,18 @@ def cache_nvrs(request, build):
...
     """
     Ensure that the given builds reference valid Build objects.
 
     Args:
-        request (pyramid.util.Request): The current request.
+        request (pyramid.request.Request): The current request.
         kwargs (dict): The kwargs of the related service definition. Unused.
     """
-    for build in request.validated.get('builds', []):
+    builds = request.validated.get('builds')
```
cverna: I think this should also be in a separated commit, since it does not relate with the side tag change.

bodhi/server/validators.py
```diff
@@ -214,21 +217,47 @@ def validate_nvrs(request, **kwargs):
...
+        request (pyramid.request.Request): The current request.
         kwargs (dict): The kwargs of the related service definition. Unused.
     """
     edited = request.validated.get('edited')
     user = User.get(request.user.name)
-
-    if not request.validated.get('builds', []):
-        request.errors.add('body', 'builds', "You may not specify an empty list of builds.")
-        return
+    builds = request.validated.get('builds') or []  # cope with builds set to None
```
cverna, line 259: I thinks these changes should also be in a separated commit from the side tag one.

bodhi/server/validators.py
```diff
@@ -1232,3 +1261,46 @@ def validate_request(request, **kwargs):
...
+            # validation of ACLs pertaining the respective components.
+            try:
+                request.validated['builds'] = [
+                    b['nvr'] for b in koji_client.listTagged(koji_tag, latest=True)
+                ]
+            except koji.GenericError as e:
+                if "invalid taginfo" in str(e).lower():
+                    request.errors.add('body', 'from_tag', "The supplied from_tag doesn't exist.")
+                else:
+                    raise
```
cverna: Should we raise a BodhiException here ?

bodhi/server/validators.py
```diff
@@ -1232,3 +1261,46 @@ def validate_request(request, **kwargs):
...
+            else:
```
cverna: This is a bit hard to follow everything here :smile:, adding some comment might be a good idea to explain the logic.

bodhi/server/models.py
nphilipp: Yeah, I noticed that too while working on the CLI part of this. I'll add a tidying up commit to remove 
mentions of `unicode` and `basestr` and let the topical changes in the PR be based on that.

bodhi/server/migrations/versions/d3f8bd499ecd_add_from_tag_column_to_updates.py
nphilipp: I'm a very "from the bottom up" person I guess and called it `from_tag` and not `from_sidetag` because the code doesn't care if the tag you supply is a side tag or not—and AIUI can't distinguish between them either: it's just a matter of convention that we call some tags "side tags" because they're created ad-hoc and short-lived for the purpose of building packages "on the side". If you want the purpose of this API be better or explicitly reflected in the name of the flag, I'll happily make the change.

bodhi/server/models.py
nphilipp: Agreed.

bodhi/server/validators.py
nphilipp: I'm trying to remember why I changed this in the beginning... Perhaps I should remove it because I don't know what the purpose of this change was anymore.

bodhi/server/validators.py
nphilipp: Good question, I'll remove it.

bodhi/server/validators.py
nphilipp: Sounds like a good idea.

nphilipp: Probably like this because we're good Python 3 citizens :wink::

```
    raise BodhiException(...) from e
```

bodhi/server/validators.py
nphilipp: Mhm, `...  # normal update with explicit list of builds`.

bodhi/server/validators.py
nphilipp: It may not look like it but this one is directly related to the from-sidetag feature, it moves checking if the list of builds is empty to the (as of yet comment-less) `else` branch in `validate_from_tag()` below.

bodhi/server/migrations/versions/d3f8bd499ecd_add_from_tag_column_to_updates.py
cverna: I don't feel strongly on that one to be honest, so let's keep it as it is :+1:

bodhi/server/validators.py
cverna: Yes I think that would make the code easier to read

nphilipp:
> @nphilipp You will probably want to squash your commits

Of course -- do you think the first one "Use less ambiguous validator names." should be separate or part of the squashed feature?

bodhi/server/validators.py
nphilipp: I've looked into it and will remove it.

bodhi/server/validators.py
nphilipp: Heh, noticed that my proposed comment is wrong even, after looking at the greater context. I'll just move the code out of the superfluous `else:` branch (which is only necessary if the block should be handled by a `finally:` branch which we don't have here).

nphilipp:
@cverna This latest push should address your requests. Mind that the order GitHub lists the commits is wrong (why ever), the first three are:

```
bef23ece Correct location of request type.
53cffcf0 Don't mention obsolete string types.
6af83d8f Use less ambiguous validator names.
```

Again, I'm not sure if the last of these shouldn't be squashed with the feature commits. What do you think?

bodhi/tests/server/services/test_updates.py
```diff
@@ -275,6 +277,43 @@ def test_new_rpm_update_koji_error(self, *args):
...
 
+    @mock.patch.dict('bodhi.server.validators.config', {'acl_system': 'dummy'})
+    @mock.patch(**mock_uuid4_version1)
+    @mock.patch(**mock_valid_requirements)
+    def test_new_rpm_update_from_tag(self, *args):
+        """Test creating an update using builds from a Koji tag."""
+        # We don't want the new update to obsolete the existing one.
+        self.db.delete(Update.query.one())
+
+        update = self.get_update(builds=None, from_tag='f17-updates-candidate')
```
abompard, line 288: This fails in CI because the code does not expect builds to be `None`, only a list (that can be empty). I suggest to either use `[]` here or to handle in the validation code the case where `builds` is `None`.

abompard:
An improper testcase crashes CI (I've suggested possible fixes in a line comment), other than that the code looks good, thanks :-)

cverna:
> @cverna This latest push should address your requests. Mind that the order GitHub lists the commits is wrong (why ever), the first three are:
> 
> ```
> bef23ece Correct location of request type.
> 53cffcf0 Don't mention obsolete string types.
> 6af83d8f Use less ambiguous validator names.
> ```
> 
> Again, I'm not sure if the last of these shouldn't be squashed with the feature commits. What do you think?

I think renaming the validators can be squashed with the feature commit since it is kind of related.

bodhi/tests/server/services/test_updates.py
nphilipp: Where does it fail and how? AIUI, it uses `BaseTestCaseMixin.get_update()` which has this:

```
        if builds:
            if isinstance(builds, list):
                builds = ','.join(builds)
            if not isinstance(builds, str):
                builds = builds.encode('utf-8')
            update['builds'] = builds
```

I.e. if `builds` is set to `None`, it should simply be missing from the returned dictionary.

nphilipp:
@cverna > I think renaming the validators can be squashed with the feature commit since it is kind of related.

Will do.

bodhi/tests/server/services/test_updates.py
abompard: In the CI logs you'll see:
```
13:55:12  Traceback (most recent call last):
13:55:12    File "/usr/lib/python3.7/site-packages/pyramid/tweens.py", line 41, in excview_tween
13:55:12      response = handler(request)
13:55:12    File "/usr/lib/python3.7/site-packages/pyramid/router.py", line 148, in handle_request
13:55:12      registry, request, context, context_iface, view_name
13:55:12    File "/usr/lib/python3.7/site-packages/pyramid/view.py", line 667, in _call_view
13:55:12      response = view_callable(context, request)
13:55:12    File "/usr/lib/python3.7/site-packages/pyramid/config/views.py", line 169, in __call__
13:55:12      return view(context, request)
13:55:12    File "/usr/lib/python3.7/site-packages/pyramid/config/views.py", line 188, in attr_view
13:55:12      return view(context, request)
13:55:12    File "/usr/lib/python3.7/site-packages/pyramid/config/views.py", line 214, in predicate_wrapper
13:55:12      return view(context, request)
13:55:12    File "/usr/lib/python3.7/site-packages/pyramid/viewderivers.py", line 384, in authdebug_view
13:55:12      return view(context, request)
13:55:12    File "/usr/lib/python3.7/site-packages/pyramid/viewderivers.py", line 325, in secured_view
13:55:12      return view(context, request)
13:55:12    File "/usr/lib/python3.7/site-packages/pyramid/viewderivers.py", line 436, in rendered_view
13:55:12      result = view(context, request)
13:55:12    File "/usr/lib/python3.7/site-packages/pyramid/viewderivers.py", line 144, in _requestonly_view
13:55:12      response = view(request)
13:55:12    File "/usr/lib/python3.7/site-packages/cornice/service.py", line 493, in wrapper
13:55:12      validator(request, **args)
13:55:12    File "/bodhi/bodhi/server/validators.py", line 90, in validator
13:55:12      f(request, **kwargs)
13:55:12    File "/bodhi/bodhi/server/validators.py", line 203, in validate_build_nvrs
13:55:12      for build in request.validated.get('builds', []):
13:55:12  TypeError: 'NoneType' object is not iterable
```
The `builds` key is set to `None`, so `get()` does not fallback on `[]`, therefore the crash.

bowlofeggs:
Can you move the "Don't mention obsolete string types." commit to another pull request, since it's unrelated to the other changes here? Alternatively, please squash the other commits to one and then this can stay here.

bowlofeggs:
```flake8``` is upset about some lines in this PR:

```
11:08:37  ./bodhi/tests/server/test_validators.py:789:36: W504 line break after binary operator
11:08:37          assert (str(excinfo.value) ==
11:08:37                                     ^
11:08:37  ./bodhi/tests/server/test_validators.py:790:1: W191 indentation contains tabs
11:08:37  				"Encountered error while requesting tagged builds from Koji: 'foo'")
11:08:37  ^
11:08:37  ./bodhi/tests/server/test_validators.py:790:1: E101 indentation contains mixed spaces and tabs
11:08:37  				"Encountered error while requesting tagged builds from Koji: 'foo'")
11:08:37  ^
11:08:37  ./bodhi/tests/server/test_validators.py:790:5: E128 continuation line under-indented for visual indent
11:08:37  				"Encountered error while requesting tagged builds from Koji: 'foo'")
11:08:37      ^
11:08:37  ./bodhi/tests/server/test_validators.py:792:1: E101 indentation contains mixed spaces and tabs
11:08:37          # check type of wrapped exception
11:08:37  ^
```

File: bodhi/server/models.py (hunk 4 above)
Comment (bowlofeggs) [reply bodhi/server/models.py]:
```suggestion
        from_tag (str): The koji tag from which the list of builds was
```

bodhi/server/validators.py
```diff
@@ -191,15 +192,15 @@ def cache_nvrs(request, build):
...
     """
     Ensure that the given builds reference valid Build objects.
 
     Args:
-        request (pyramid.util.Request): The current request.
+        request (pyramid.request.Request): The current request.
         kwargs (dict): The kwargs of the related service definition. Unused.
     """
-    for build in request.validated.get('builds', []):
+    for build in request.validated.get('builds') or []:  # cope with builds being None
```
bowlofeggs, line 204: This would be more idiomatic:

```suggestion
    for build in request.validated.get('builds', []):
```

bodhi/server/validators.py
bowlofeggs: This would be more idiomatic:

```suggestion
    builds = request.validated.get('builds', [])
```

bodhi/server/validators.py
bowlofeggs:
```suggestion
    for build in request.validated.get('builds', []):
```

bodhi/server/validators.py
```diff
@@ -496,16 +523,16 @@ def validate_acls(request, **kwargs):
 
 
 @postschema_validator
-def validate_uniqueness(request, **kwargs):
+def validate_build_uniqueness(request, **kwargs):
```
bowlofeggs: I'm fine with renaming this, but it seems unrelated to the rest of the PR - can we do this in a separate pull request?

bowlofeggs (dismissed):

bowlofeggs:
It would also be ideal to move the commit "Correct location of request type." to its own PR. It would make this one simpler to review.

bodhi/server/validators.py
nphilipp: But it wouldn't be the same, that's why I replaced it (I want it to not croak at the `'builds'` key having `None` as a value). I could pull it apart like this to make it stand out more:

```
builds = request.validated.get('builds')
# cope with builds being None
if not builds:
    builds = []
for build in builds:
    ...
```

nphilipp:
I'll move the standalone commits out into their own PR.

bodhi/server/validators.py
nphilipp: @bowlofeggs I didn't make it clear this was a question: should I change this as I outlined, or add a comment?

bodhi/server/validators.py
nphilipp: Sure, I'll put this and the two type-change ones in their own PR (one PR for the three commits).

nphilipp:
The first three commits are covered in their own PR #3320.

nphilipp:
Apparently GH doesn't let me resolve this without dismissing. :disappointed:

bodhi/server/migrations/versions/d3f8bd499ecd_add_from_tag_column_to_updates.py
```diff
@@ -0,0 +1,41 @@
...
+Revises: 5703ddfe855d
+Create Date: 2019-05-14 15:10:54.769789
+"""
+from alembic import op
+import sqlalchemy as sa
+
+
+# revision identifiers, used by Alembic.
+revision = 'd3f8bd499ecd'
+down_revision = '5703ddfe855d'
```
cverna: You need to update the migration script to use the latest revision :smile:

cverna:
Overall this LGTM. I think you can squash your commits. 

@abompard, do you want to also take a second look at it ?

nphilipp:
The latest push squashes the commits into one.

bodhi/server/migrations/versions/d3f8bd499ecd_add_from_tag_column_to_updates.py
abompard: Yep, this should be rebased.

abompard (changes requested):

nphilipp:
Oops, missed that another migration went in meanwhile...

abompard (approved):
I think this one looks good now.

cverna (approved):
:fireworks:

PR #3230, opened 2019-05-15
Don't crash on invalid RSS requests.

### LINKED ISSUE #3227: UnboundLocalError: local variable 'key' referenced before assignment in RSS handler
Our RSS feed will crash if the requestor sends an invalid query:

```
2019-05-14 14:58:49,638 ERROR [bodhi.server][Dummy-3] Error caught.  Handling HTML response.
  File "/usr/lib/python3.7/site-packages/pyramid/tweens.py", line 39, in excview_tween
    response = handler(request)
  File "/usr/lib/python3.7/site-packages/pyramid/router.py", line 156, in handle_request
    view_name
  File "/usr/lib/python3.7/site-packages/pyramid/view.py", line 642, in _call_view
    response = view_callable(context, request)
  File "/usr/lib/python3.7/site-packages/pyramid/config/views.py", line 181, in __call__
    return view(context, request)
  File "/usr/lib/python3.7/site-packages/pyramid/viewderivers.py", line 390, in attr_view
    return view(context, request)
  File "/usr/lib/python3.7/site-packages/pyramid/viewderivers.py", line 368, in predicate_wrapper
    return view(context, request)
  File "/usr/lib/python3.7/site-packages/pyramid/viewderivers.py", line 462, in rendered_view
    request, result, view_inst, context)
  File "/usr/lib/python3.7/site-packages/pyramid/renderers.py", line 435, in render_view
    return self.render_to_response(response, system, request=request)
  File "/usr/lib/python3.7/site-packages/pyramid/renderers.py", line 458, in render_to_response
    result = self.render(value, system_values, request=request)
  File "/usr/lib/python3.7/site-packages/pyramid/renderers.py", line 454, in render
    result = renderer(value, system_values)
  File "/usr/lib/python3.7/site-packages/bodhi/server/renderers.py", line 74, in render
    feed_description = "All %s" % (key)
UnboundLocalError: local variable 'key' referenced before assignment
```

It would be better for it to return a 4xx response instead of a 5xx response when getting invalid RSS requests. In this case, the issue is that the request doesn't contain one of the keys in the big if/else block here:

https://github.com/fedora-infra/bodhi/blob/846bc15a53122908d7bd38c0d2230b7f17d53674/bodhi/server/renderers.py#L55-L66

It might be better to refactor this code to use a dictionary instead of a big if/else, but that's not necessary to solve this issue.

bowlofeggs:
fixes #3227

### DIFF
--- a/bodhi/server/renderers.py
+++ b/bodhi/server/renderers.py
@@ -16,10 +16,15 @@
 # along with this program; if not, write to the Free Software
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
 """Define special view renderers, such as RSS."""
+import logging
 import operator
 
 from pytz import utc
 from feedgen.feed import FeedGenerator
+from pyramid.exceptions import HTTPBadRequest
+
+
+log = logging.getLogger(__name__)
 
 
 def rss(info):
@@ -64,6 +69,10 @@ def rss(info):
         elif 'overrides' in data:
             key = 'overrides'
             feed_title = 'Update overrides'
+        else:
+            # This is a request we don't know how to render. Let's return BadRequest and log.
+            log.debug('Unable to render RSS feed for data: %s', data)
+            raise HTTPBadRequest('Invalid RSS feed request')
 
         feed_description_list = []
         for k in request.GET.keys():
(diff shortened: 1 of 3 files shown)


bowlofeggs:
The rawhide test failure is due to https://bugzilla.redhat.com/show_bug.cgi?id=1707981 and can be ignored.

bowlofeggs:
This patch will be included in Bodhi 4.0.0.

bowlofeggs:
I've moved this commit over to #3232.

PR #3232, opened 2019-05-16
Critical fixes for Bodhi 4.0.0

bowlofeggs:
There are several commits in this PR because they build upon each other. Please read their commit messages for details.


cverna:
We probably want also to update the dev environment config file --> https://github.com/fedora-infra/bodhi/blob/develop/devel/ansible/roles/bodhi/files/bodhi.toml#L8

bowlofeggs:
There are several open pull requests that are critical for 4.0.0. Some of them build upon others, and it's becoming annoying to keep the patches rebased correctly for my RPM builds of the betas, so I'm going to close the others and cherry pick their commits into this one.

bowlofeggs:
@cverna fixed!

abompard:
I'm a bit confused about the revert of the `blockingCallFromThread` uses. Since the message callback is running in a Twisted thread, it needs `blockingCallFromThread` to use the `fedora_messaging` API, no? What am I missing?

bowlofeggs:
@jeremycline - would you care to explain?

jeremycline:
> I'm a bit confused about the revert of the blockingCallFromThread uses. Since the message callback is running in a Twisted thread, it needs blockingCallFromThread to use the fedora_messaging API, no? What am I missing?

The callback is being run in the context of a normal thread, managed by Twisted (it's outside the event loop) by means of the ``deferToThread`` API. ``fedora_messaging.api.publish`` does not use Twisted at all, just the blocking connection provided by Pika. It can be run anywhere, but it's a synchronous call so it _shouldn't_ be run in the reactor loop as that blocks any other asynchronous calls from running. If the ``publish`` call made use of Twisted APIs, it would indeed need to use the ``blockingCallFromThread`` to schedule a call to it in the context of the reactor loop thread, but as it is it just runs a blocking IO call inside the event loop.

abompard:
Ah that's where I got confused, I thought the publish API had been moved to Twisted as well, but indeed only the CLI is defaulting to Twisted, not the API. Thanks for the clarification.

bowlofeggs:
The tests failed due to hitting the timeout, and I believe it is due to https://github.com/fedora-infra/fedora-messaging/issues/170

abompard:
I can try and investigate that, one possible explanation would be that publish calls are retried now and we should mock/disable that for unit tests.

bowlofeggs:
The test timeout should be fixed by https://github.com/fedora-infra/bodhi/pull/3243

bowlofeggs:
The ```rawhide_unit``` failure is due to https://bugzilla.redhat.com/show_bug.cgi?id=1707981 and can be ignored.

cverna (approved):
Reviewed each commit individually and they look good to me

bowlofeggs:
Cool, let's merge this bad boy then!

PR #3242, opened 2019-05-21
Make sure %{uid} in krb ccache gets replaced with the effective UID

puiterwijk:
This makes it so runs by different users get evaluated to different
ccache's.
This prevents issues when a process is run by another user, and then
the normal procedures would be unable to authenticate to buildsys
until the incorrectly-owned ccache is removed.

For example: bodhi composes might fail to log in to koji after an
admin used bodhi-shell to fix something else, because the bodhi-shell
might have run as root.


puiterwijk:
The `rawhide-unit` test got hit by an Infra outage where the registry broke. The `pr-head` looks like it's a segfault in.. bash?

bowlofeggs:
On Tue, 2019-05-21 at 23:57 -0700, Patrick Uiterwijk wrote:
> The pr-head looks like it's a segfault in.. bash?

The segfault is due to https://bugzilla.redhat.com/show_bug.cgi?id=1707981 and can be ignored.

bowlofeggs (approved):
Thanks!

PR #3243, opened 2019-05-22
Do not send fedora-messages in the unit tests

### LINKED ISSUE #3241: Bodhi isn't mocking all calls to AMQP publish
See https://github.com/fedora-infra/fedora-messaging/issues/170 for full detail, but tl;dr is that I discovered that Bodhi's tests are trying to publish real messages. Starting with 1.6.1, fedora-messaging now retries on publish failures, and this means that Bodhi's tests take a very long time now. It's so long that the tests time out in CI and are thus blocking progress on important pull requests, such as #3232, so this is critical to resolve.

bowlofeggs:
We recently started to notice a significant time increase in
running Bodhi's unit tests with fedora-messaging >= 1.6.1. It
turned out that that version added more retry logic when it could
not publish messages, and that many of our unit tests were
attempting to publish messages. This caused the tests to go from
about 3 minutes to over an hour.

This commit adds mocks to all the places that were attempting to
publish messages.

fixes #3241

### DIFF
--- a/bodhi/tests/server/consumers/test_composer.py
+++ b/bodhi/tests/server/consumers/test_composer.py
@@ -2011,25 +2011,26 @@ testmodule:master:20172:2
             with mock_sends(*[base_schemas.BodhiMessage] * 3):
                 self.handler(self._make_msg())
 
-        with self.db_factory() as session:
-            up = session.query(Update).one()
+        with mock_sends(*[base_schemas.BodhiMessage] * 3):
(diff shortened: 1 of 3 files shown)


abompard (approved):
Yes, looks correct.

bowlofeggs:
The ```rawhide_unit``` failure is due to https://bugzilla.redhat.com/show_bug.cgi?id=1707981 and can be ignored.

PR #3245, opened 2019-05-23
Backport 3232

bowlofeggs:
This is a backport of #3232 to 4.0.


bowlofeggs:
The flake8 test actually passed - it was just a failure in CentOS CI to notify GitHub of the result.

PR #3246, opened 2019-05-24
Add Celery

abompard:
This is a big PR, follow each commits to make it more understandable. In this PR:
- Celery is introduced to handle the long-running tasks
- the composer task and the handle_update task are converted to Celery (they were previously run by a Fedora Messaging callback)
- to test the update task, Ipsilon is introduced in the CI environment, which makes authenticated requests possible
- an additional CLI test is added, just as a bonus :-) (it shares some DB code with the task tests, but I'm not sure if it's a good idea to abstract it out)


bowlofeggs:
Since this is backwards incompatible, I've marked it as WIP for now so Mergify doesn't merge it yet. Once we merge a backwards incompatible patch to develop, the develop branch has to become Bodhi 5, so we might want to make sure we are ready to stop using develop for 4.x before merging this.

bowlofeggs:
Looks like there are some conflicts and test failures.

bowlofeggs:
There seem to be conflicts and test failures.

nphilipp:
I'll resolve the conflicts.

bodhi/server/config.py
```diff
@@ -155,7 +155,7 @@ def validate_path(value):
     Raises:
         ValueError: If os.path.exists returns False.
     """
-    if not os.path.exists(value):
+    if value is None or not os.path.exists(value):
         raise ValueError('"{}" does not exist.'.format(value))
```
nphilipp: How about this?

```
-    if not os.path.exists(value):
-￼        raise ValueError('"{}" does not exist.'.format(value))
￼+    if value is None or not os.path.exists(value):
+        raise ValueError(f'{value!r} does not exist.')
```

It would make it immediately obvious if `value` is `None` and not a string.

nphilipp: This would change the quotes from double to single and need to be reflected in the test(s).

bodhi/server/tasks/__init__.py
```diff
@@ -0,0 +1,85 @@
...
+
+from celery import Celery
+
+from bodhi.server import bugs, buildsys, initialize_db
+from bodhi.server.config import config
+from bodhi.server.util import pyfile_to_module
+
+
+# Workaround https://github.com/celery/celery/issues/5416
+if sys.version_info >= (3, 7):
```
nphilipp: We could apply the workaround only with versions of Celery where this isn't fixed:

```
if celery.version_info < (4, 3) and sys.version_info >= (3, 7):
    ...
```

What do you think?

bodhi/server/tasks/__init__.py
```diff
@@ -0,0 +1,85 @@
...
+    from re import Pattern
+    from celery.app.routes import re as routes_re
+    routes_re._pattern_type = Pattern
+
+log = logging.getLogger('bodhi')
+
+# The Celery app object.
+app = Celery()
+app.config_from_object(pyfile_to_module(config["celery_config"], "celeryconfig"))
```
nphilipp, line 40: This causes errors on import if the configuration file doesn't exist (e.g. when running the test suite on a system where Bodhi isn't installed):

```
=============================================== ERRORS ===============================================
____________________ ERROR collecting bodhi/tests/server/services/test_updates.py ____________________
../../../.virtualenvs/bodhi/lib/python3.7/site-packages/py/_path/local.py:701: in pyimport
    __import__(modname)
bodhi/tests/server/__init__.py:25: in <module>
    from bodhi.server.models import (
bodhi/server/models.py:47: in <module>
    from bodhi.server.tasks import handle_update
bodhi/server/tasks/__init__.py:40: in <module>
    app.config_from_object(pyfile_to_module(config["celery_config"], "celeryconfig"))
bodhi/server/util.py:1516: in pyfile_to_module
    with open(filename) as config_file:
E   FileNotFoundError: [Errno 2] Unable to load file (No such file or directory): '/etc/bodhi/celeryconfig.py'
========================================== warnings summary ==========================================
```

Maybe move it to `_do_init()` below?

bodhi/tests/server/tasks/__init__.py
```diff
@@ -0,0 +1,18 @@
+# Copyright © 2018 Red Hat, Inc.
```
nphilipp, line 1: > \# Copyright © 2018 Red Hat, Inc.

> \# Copyright © 2018

> # 2018

:grin:

bodhi/tests/server/test_util.py
```diff
@@ -1424,3 +1424,37 @@ def test___call___success(self, Session, log_exception):
...
+
+
+class TestPyfileToModule(unittest.TestCase):
+    """Test the pyfile_to_module() function."""
+
+    def setUp(self):
+        self.tempdir = tempfile.mkdtemp('bodhi')
+
+    def tearDown(self):
+        shutil.rmtree(self.tempdir)
```
nphilipp, line 1565: Since we use pytest, how about not deriving `TestPyfileToModule` from `unittest.TestCase`? The unittest compat disables a couple of pytest features, for us here the most obvious is being able to use plain asserts rather than `self.assertSomething(...)`...

bodhi/tests/server/test_util.py
```diff
@@ -1424,3 +1424,37 @@ def test___call___success(self, Session, log_exception):
...
+
+    def test_basic_call(self):
+        filepath = os.path.join(self.tempdir, "testfile.py")
+        with open(filepath, "w") as fh:
+            fh.write("FOO = 'bar'\n")
+        result = util.pyfile_to_module(filepath, "testfile")
+        self.assertEqual(getattr(result, "FOO"), "bar")
+        self.assertEqual(result.__file__, filepath)
+        self.assertEqual(result.__name__, "testfile")
```
nphilipp, line 1574: ...for instance these lines could then be written the following way:

```
        assert getattr(result, "FOO") == "bar"
        assert result.__file__ == filepath
        assert result.__name__ == "testfile"
```

I find that this is much easier to read, and not having to lookup what the respective `.assertSomething()` method is called is also a plus.

devel/ci/integration/bodhi/start.sh
```diff
@@ -1,8 +1,16 @@
 set -m
 
 # Bodhi backend
+echo "Starting Fedora Messaging consumers"
 fedora-messaging consume &
 
+# Celery
+echo "Starting Celery"
+pushd /  # Otherwise celeryconfig will be picked up from CWD (/bodhi)
```
nphilipp, line 9: > \# Otherwise celeryconfig will be picked up from CWD (/bodhi)

Oops, is there a way to make this more robust?

devel/ci/integration/bodhi/variants.rpm.xml.j2
```diff
@@ -0,0 +1,32 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE variants PUBLIC "-//Red Hat, Inc.//DTD Variants info//EN" "variants2012.dtd">
+<variants>
+    <variant id="Everything" name="Everything" type="variant">
+        <arches>
+            <arch>x86_64</arch>
+	    [% if release.version_int >= 25 or release.version_int == 6 %]
+		<arch>i386</arch>
```
nphilipp, line 8: I imagine the single-digit versions are for RHEL, CentOS. Is there a way we can emulate how the two "distro worlds" are usually distiniguished in RPM spec files? Like this:

```
%if 0%{?fedora} >= 25 || 0%{?rhel} == 6
...
%endif
```

devel/ci/integration/tests/utils.py
```diff
@@ -134,3 +163,25 @@ def get_sent_messages(rabbitmq_container):
...
+
+
+def get_task_results(container):
+    """Read a file in a container.
+
+    Args:
+        container (conu.DockerContainer): The container where the file is.
+        binary (bool): Whether the file should be opened in binary mode.
+
+    Yields:
```
nphilipp: Shouldn't this be "Returns"? People might look at the docstring and think this is a generator otherwise.

devel/ci/integration/tests/utils.py
```diff
@@ -54,6 +57,32 @@ def make_db_and_user(db_container, name, use_dump=False):
...
+        container: conu.DockerContainer, path: str, dir_not_empty: bool = False,
+        timeout: int = 120):
+    """Check that a file is in a container.
+
+    Args:
+        container: The container where the file is.
+        path: The file path in the container.
+        timeout: How long to wait before throwing an exception.
+    """
+    while timeout > 0:
```
nphilipp, line 70: I don't think this can ever be `False` with `if timeout == 0:" in the loop below. You could just move the latter check (and raise the exception) out of the loop:

```
while timeout > 0:
    ...
    timeout -= 1
if timeout == 0:
    raise ...
```

nphilipp (changes requested):
Can you please change at least the following I flagged?

- `bodhi/server/tasks/__init__.py`: make it free of side-effects, i.e. importable on systems where Bodhi isn't installed
- `Yields:` → `Returns:` to not confuse people

To me the rest isn't so important, i.e. either arguable whether it should be changed or not important enough to block this going in (can be fixed later).

bodhi/server/tasks/__init__.py
abompard: Yep, good point, thanks.

bodhi/server/config.py
abompard: Sure

bodhi/tests/server/tasks/__init__.py
abompard: Wait... What year is this? ;-)

bodhi/tests/server/test_util.py
abompard: I agree but since I mostly moved the test cases from `consumers` to `tasks`, I wanted to keep those changes to a minimum so git can follow the rename. I think the conversion to pytest can be the topic of another PR, if that's OK with you.

devel/ci/integration/bodhi/start.sh
abompard: I haven't found any. It's an unfortunate but common problem with apps that choose to use a python module for their configuration instead of a regular file: you're dependant on Python's import paths. If there's a better way, I'm all ears, though.

devel/ci/integration/bodhi/variants.rpm.xml.j2
abompard: Actually I just copied this file from our Fedora ansible repo, and I'd rather not make any changes to it so we can update it whenever we want/need to.

devel/ci/integration/tests/utils.py
abompard: Correct, I abusively cut-n-pasted the docstring ;-)

devel/ci/integration/tests/utils.py
abompard: Yep, agreed.

bodhi/server/tasks/__init__.py
abompard: That won't work because `app` is used as a decorator before the tasks are loaded, so it needs to be configured first.

I think the issue is that the development.ini file that is used in the vagrant box is copied from the file in `devel/`, so it needs to be updated with the changes that happened in this PR (unless you start with a fresh vagrant box on this branch)

bodhi/tests/server/test_util.py
nphilipp: Sure. :+1:

abompard: See #3357.

bodhi/server/tasks/__init__.py
nphilipp: Hmm. `Celery.config_from_object()` lets users specify module names as strings, which only tries to import the module when it's used for the first time. Let me cook something up using a PEP 302 custom finder loader and see if it helps.

nphilipp: As discussed, this just needs the appropriate configuration in `development.ini` (cf. `devel/development.ini.example`).

devel/ci/integration/bodhi/start.sh
nphilipp: We could use a custom module finder/loader such as I mentioned when we talked before, it would let us/celery import configuration from a defined module name but we get to choose the file that actually gets loaded by e.g. an absolute path in the configuration, like this (untested):

* in `bodhi/server/util.py`:

```
...
class PyFileAsModule:
    """Module finder/loader for arbitrary Python files."""

    def __init__(self, filename: str, modname: str):
        if modname in sys.modules:
            raise ValueError(f"Module {modname} exists in sys.modules.")

        if modname in sys.meta_path:
            raise ValueError(f"Module finder for {modname} registered in sys.meta_path.")

        self.filename = filename
        self.modname = modname

    @classmethod
    def register(cls, filename: str, modname: str) -> str:
        """Create and register a module finder/loader for Python files."""
        obj = cls(filename=filename, modname=modname)
        sys.meta_path.append(obj)
        return modname

    def find_module(self,
                    fullname: str,
                    path: str = None) -> typing.Optional['PyFileAsModule']:
        """Return self as the loader if responsible."""
        if fullname == self.modname:
            return self

    def load_module(self, fullname: str) -> types.ModuleType:
        """Return a module from a registered file."""
        if fullname in sys.modules:
            return sys.modules[fullname]

        try:
            m = types.ModuleType(fullname)
            m.__file__ = self.filename
            m.__loader__ = self
            sys.modules[fullname] = m

            with open(self.filename, 'r') as f:
                exec(f, sys.modules[fullname].__dict__)

            return m
        except Exception as e:
            if fullname in sys.modules:
                del sys.modules[fullname]
            raise e
```

* in `bodhi/server/tasks/__init__.py`:

```
...
from bodhi.server.util import PyFileAsModule
...
# The Celery app object.
app = celery.Celery()
app.config_from_object(PyFileAsModule.register(config["celery_config"], 'celeryconfig'))
...
```

bodhi/server/tasks/__init__.py
abompard: The `development.ini` file in Vagrant is created by copying the `devel/development.ini.example` file at provision time. Are you referring ton another file?

nphilipp:
With the latest push, some unit tests fail or hang because they try to send messages. I've fixed these in [my fork](https://github.com/nphilipp/bodhi/tree/abompard-nils--celery): https://github.com/nphilipp/bodhi/commit/bef1e7285d76cdeef8b693708e51703907f59575

@abompard if the changes look good to you, please apply, thanks!

abompard:
I have applied your changes, thanks Nils.

abompard:
I can't reproduce the unitttest errors locally, either in Vagrant or by running bodhi-ci. Any idea @nphilipp ?

nphilipp:
@abompard as discussed, here's the change to make coverage ignore the workaround for old celery versions: https://github.com/nphilipp/bodhi/commit/5133c5c2202feb50c4d2d2a9696639751184fe6c

Zlopez:
Solved conflict and rebased

Zlopez:
I'm now hitting this issue when trying to run `alembic upgrade head` on this branch:
```
  File "/usr/local/bin/alembic", line 11, in <module>
    load_entry_point('alembic', 'console_scripts', 'alembic')()
  File "/usr/lib/python3.7/site-packages/alembic/config.py", line 540, in main
    CommandLine(prog=prog).main(argv=argv)
  File "/usr/lib/python3.7/site-packages/alembic/config.py", line 534, in main
    self.run_cmd(cfg, options)
  File "/usr/lib/python3.7/site-packages/alembic/config.py", line 514, in run_cmd
    **dict((k, getattr(options, k, None)) for k in kwarg)
  File "/usr/lib/python3.7/site-packages/alembic/command.py", line 276, in upgrade
    script.run_env()
  File "/usr/lib/python3.7/site-packages/alembic/script/base.py", line 475, in run_env
    util.load_python_file(self.dir, "env.py")
  File "/usr/lib/python3.7/site-packages/alembic/util/pyfiles.py", line 90, in load_python_file
    module = load_module_py(module_id, path)
  File "/usr/lib/python3.7/site-packages/alembic/util/compat.py", line 156, in load_module_py
    spec.loader.exec_module(module)
  File "<frozen importlib._bootstrap_external>", line 728, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "/home/vagrant/bodhi/bodhi/server/migrations/env.py", line 27, in <module>
    from bodhi.server.models import Base
  File "/home/vagrant/bodhi/bodhi/server/models.py", line 48, in <module>
    from bodhi.server.tasks import handle_update
  File "/home/vagrant/bodhi/bodhi/server/tasks/__init__.py", line 40, in <module>
    app.config_from_object(pyfile_to_module(config["celery_config"], "celeryconfig"))
  File "/home/vagrant/bodhi/bodhi/server/util.py", line 1516, in pyfile_to_module
    with open(filename) as config_file:
FileNotFoundError: [Errno 2] Unable to load file (No such file or directory): '/etc/bodhi/celeryconfig.py'
```

Zlopez:
I fixed the issue with `vagrant provisioning` mentioned above and also fixed one test that was failing :-)

Zlopez:
I see that the tests are now failing on coverage, which is weird:
```
FAIL Required test coverage of 99% not reached. Total coverage: 99.25%
```

Zlopez:
Rebased and fixed conflict

nphilipp:
Picking this up again I noticed that the changes in the PR brought coverage down. The following push fixes these issues:

- One test checking greenwave messages being dispatched to the consumer was inadvertently removed, added it back.
- Another test which checked how the consumer copes without the composer being installed needed to be repurposed for the generic "handler throws exception" case (instead of being removed).

devel/ci/integration/bodhi/start.sh
nphilipp: Hmm, it seems Celery's way to find its configuration is broken enough for the above -- even if fixed -- not to work.

nphilipp (approved):
I think we've ironed out all issues, except unrelated stuff like rawhide-docs still failing.

PR #3247, opened 2019-05-28
Remove duplicate permission dict for RabbitMQ host.

nphilipp:
This made provisioning the Vagrant box fail lately:

```
TASK [bodhi : Create a guest user in RabbitMQ] *********************************
fatal: [bodhi]: FAILED! => {"changed": false, "msg": "Error parsing permissions: You can't have two permission dicts for the same vhost"}
```


cverna (approved):
LGTM

PR #3250, opened 2019-05-28
Doc that bodhi 4 dropped fedmsg_enabled setting

bowlofeggs:


puiterwijk (approved):
LGTM

PR #3254, opened 2019-05-29
Message consuming and sending now works in Vagrant

bowlofeggs:
This commit allows us to produce and consumer AMQP messages in
the Vagrant development environment, and makes the shell aliases
such that they interact with the fm-consumer process.


cverna (approved):
:+1:

PR #3256, opened 2019-05-29
Add a missing f for a format string

bowlofeggs:
nirik reported that Bodhi was printing this warning:

Warning: {update.get_title()} has unsigned builds and has been skipped

It turns out that we forgot to put the f in front of a format
string, so the template wasn't getting rendered.

This commit fixes that and adds an assertion that the warning is
rendered correctly.


cverna (approved):
:+1:

PR #3269, opened 2019-05-30
Drop grepping for id tags

puiterwijk:


docs/user/release_notes.rst
```diff
@@ -2,6 +2,26 @@
...
+Bug fixes
+^^^^^^^^^
+
+* Fix updateinfo <id> tag detection (:issue:`3261`).
+
+
+Contributors
+^^^^^^^^^^^^
+
+The following developers contributed to Bodhi 4.0.1:
```
cverna:
```suggestion
The following developers contributed to Bodhi 4.0.2:
```

docs/user/release_notes.rst
```diff
@@ -2,6 +2,26 @@
...
+v4.0.2
+------
+
+This is a bugfix release.
+
+
+Bug fixes
+^^^^^^^^^
+
+* Fix updateinfo <id> tag detection (:issue:`3261`).
```
bowlofeggs:
```suggestion
* Drop updateinfo <id> tag detection (:issue:`3269`).
```

bowlofeggs (changes requested):

bowlofeggs:
Maybe we should change the commit message to "Drop grepping for <id> tags" instead of "Fix", since we are actually dropping it.

docs/user/release_notes.rst
bowlofeggs: And Maybe we should s/Fix/Drop/ here too.

bowlofeggs (approved):
Thanks!

PR #3270, opened 2019-05-30
Add multiple DNF configs in tests

puiterwijk:

### DIFF
--- a/bodhi/server/util.py
+++ b/bodhi/server/util.py
@@ -22,6 +22,7 @@ from contextlib import contextmanager
 from urllib.parse import urlencode
 import functools
 import hashlib
+import itertools
 import json
 import os
 import pkg_resources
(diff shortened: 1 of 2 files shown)


bodhi/server/util.py
```diff
@@ -317,22 +318,53 @@ def sanity_check_repodata(myurl, repo_type):
             tests.append((['list', 'available'], 'testrepo'))
         elif repo_type == 'module':
             tests.append((['module', 'list'], '.*'))
+        else:
```
bowlofeggs: Would we not want to run this on "yum" repos?

puiterwijk: Yeah, that was my bug :).

bodhi/server/util.py
```diff
@@ -315,24 +316,54 @@ def sanity_check_repodata(myurl, repo_type):
...
+                else:
+                    raise RepodataException(
+                        f"DNF did not return expected output when running test!"
+                        + f" Test: {dnfargs}, config: {config}, expected: {expout},"
+                        + f" output: {output}")
+
+
+def build_cfg_combinations(options):
+    """
+    Builds the Cartesian product of cfg_options
```
abompard, line 349: `cfg_options` should probably be `options` here, since it's the function's arg name.

bodhi/server/util.py
```diff
@@ -315,24 +316,54 @@ def sanity_check_repodata(myurl, repo_type):
...
+        list: A list of all config option combinations to test.
+    """
+    raw_options = []
+    for option in options:
+        raw_option = []
+        for value in options[option]:
+            raw_option.append("%s=%s" % (option, value))
+        raw_options.append(raw_option)
+
+    return list(itertools.product(*raw_options))
```
abompard, line 363: Instead of converting the list elements to lists themselves later, here you may want to do something like:
```python
return [list(result) for result in itertools.product(*raw_options)]
```

abompard (commented):
It took me some time to understand how the `build_cfg_combinations()` function was working, but that's because I'm unfamiliar with `itertools.product()`. 
The code looks fine, I've made a couple nitpicky comments.

ryanlerch:
This one has unresolved changes that have stuck around for a while. Going to close.

Feel free to reopen if still working on this

PR #3275, opened 2019-05-31
Clarify the meaning of atomic commits

bowlofeggs:
There was some confusion over the meaning and purpose of the
requirement for atomic commits in Bodhi. This commit seeks to
clarify that confusion.


docs/developer/index.rst
```diff
@@ -24,9 +24,11 @@ Before you submit a pull request to Bodhi, please ensure that it meets these cri
   ``flake8`` utility to automatically check your code.
 * Add an entry to `docs/user/release_notes.rst`_ for any changes you make that should be in release
   notes.
-* Make sure your commits are atomic. Each commit should focus on one improvement or bug fix. If you
-  need to build upon changes that are related but aren't atomic, feel free to send more than one
-  commit in the same pull request.
+* Make sure your commits are atomic. With only rare exceptions, each improvement or bug fix should
+  have exactly one commit. This makes it much easier to peruse the git history to find out what
+  changes occurred, and is particularly valuable when commits need to be cherry picked. If you need
```
puiterwijk: `what changes occurred` could also match changes that were made during a pull request for fixing remarks, which I think is not what we want.
How about `find out which changes relate to a feature or bugfix implementation`?

puiterwijk (dismissed):

bowlofeggs:
@puiterwijk I am applying this patch to address your suggestion:

```
diff --git a/docs/developer/index.rst b/docs/developer/index.rst
index 44985d9e..b05b09e8 100644
--- a/docs/developer/index.rst
+++ b/docs/developer/index.rst
@@ -25,10 +25,10 @@ Before you submit a pull request to Bodhi, please ensure that it meets these cri
 * Add an entry to `docs/user/release_notes.rst`_ for any changes you make that should be in release
   notes.
 * Make sure your commits are atomic. With only rare exceptions, each improvement or bug fix should
-  have exactly one commit. This makes it much easier to peruse the git history to find out what
-  changes occurred, and is particularly valuable when commits need to be cherry picked. If you need
-  to build upon prior unmerged commits while fixing a different issue, feel free to send more than
-  one commit in the same pull request.
+  have exactly one commit. This makes it much easier to peruse the git history to find out which
+  changes relate to a feature or bugfix implementation, and is particularly valuable when commits
+  need to be cherry picked. If you need to build upon prior unmerged commits while fixing a
+  different issue, feel free to send more than one commit in the same pull request.
 * Your commit messages must include a Signed-off-by tag with your name and e-mail address,
   indicating that you agree to the
   `Developer Certificate of Origin <https://developercertificate.org/>`_. Bodhi uses version 1.1 of
```

abompard (approved):
Seems clear to me.

PR #3239, opened 2019-05-19
Use update.alias instead of update.title in updates rss link

sebwoj:
Signed-off-by: Sebastian Wojciechowski <s.wojciechowski89@gmail.com>

PR #3276, opened 2019-06-01
Add package list to the description of an update in rss feed

### LINKED ISSUE #3272: RSS feed gives no information about packages build
Since commit 60dc56c that RSS feed doesn't give information about the packages affected in a build.
Example from [Fedora30 Stable - RSS](https://bodhi.fedoraproject.org/rss/updates/?releases=F30&status=stable) :
```xml
<item>
  <title>FEDORA-2019-59d394e0fd</title>
  <link>https://bodhi.fedoraproject.org/updates/createrepo_c-0.14.1-1.fc30</link>
  <description>- Update to 0.14.1
- Add --pkgorigins mode for Koji
- Correct pkg count in headers if there were invalid pkgs (RhBug:1596211)
- Prevent exiting with 0 if errors occur while finalizing repodata.
  </description>
  <pubDate>Fri, 24 May 2019 12:20:49 +0000</pubDate>
</item>
```

Also the link is invalid as in #3248

Looking to the [Fedora30 Stable - Web UI](https://bodhi.fedoraproject.org/updates/?releases=F30&status=stable) it seems that it should be:
```
  <title>createrepo_c-0.14.1-1.fc30</title>
  <link>https://bodhi.fedoraproject.org/updates/FEDORA-2019-59d394e0fd</link>
```

mattiaverga:
* add the list of builds included in an update in the rss description of the update itself. Also transforms markdown code to html for better readability of the rss feed.
* correct update url

Fixes #3272 


bodhi/server/renderers.py
```diff
@@ -98,11 +100,21 @@ def link_dict(obj):
...
+                text += '# Update description:\n' + notes(*args, **kwargs)
+                return markup(None, text)
+            return describe
+
         getters = {
             'updates': {
                 'title': operator.itemgetter('alias'),
-                'link': linker('update', 'id', 'title'),
-                'description': operator.itemgetter('notes'),
+                'link': linker('update', 'id', 'alias'),
```
bowlofeggs, line 137: This change was also made in https://github.com/fedora-infra/bodhi/pull/3239

bowlofeggs (changes requested):
This looks reasonable - can we add some tests to make sure the expected info is getting rendered? We probably already have some tests on the RSS feed, so we might be able to just add some assertions if my probably claim turns out to be right :)

bodhi/server/renderers.py
mattiaverga: I've removed the commit from this PR

mattiaverga:
While working on this I've found a possible error in tests: https://github.com/fedora-infra/bodhi/blob/639e1406b9ba0f97b7800f0d740619d117a80d80/bodhi/tests/server/services/test_updates.py#L1587-L1590
The URL `/updates/bodhi-2.0-1.fc17` is wrong, but the test doesn't fail because a response status 406 is expected. Should the URL has to be corrected or it's like that by purpose?

bowlofeggs:
Hmm, that test does seem pretty odd. The name of the test doesn't imply that we'd be expecting that code. Perhaps the code is happening because the accept header is Atom and we can only do HTML, JSON, or RSS? In any case, it's not completely clear what the purpose of that test is, and yeah that URL is invalid.

mattiaverga:
I have created #3315 to correct the wrong URL and give a better description to the test.

bodhi/server/renderers.py
```diff
@@ -98,11 +100,21 @@ def link_dict(obj):
                 return dict(href=request.route_url(route, **{param: obj[key]}))
             return link_dict
 
+        def describe_update(notes, builds):
```
bowlofeggs: Can we add type annotations to describe what these arguments are? It seems like they might be callables based on how they are used below, but it would be good to describe their types with annotations and their purpose with a docblock. It would also be good to document the purpose of this function in a docblock.

bowlofeggs (dismissed):

bodhi/server/renderers.py
mattiaverga: Yeah, I admit it is quite contorted indeed. I have taken example from the above `linker` function and I had had some problems myself to follow the code stream.

I've added a docstring now, I hope it's clear enough...

szpak: @bowlofeggs Is there anything else that needs to be done in code to merge this PR?

szpak:
It would be really useful to have it merged...

mattiaverga:
Just to be clear: this will still show the update alias as title of the RSS entry, the list of the updated packages will be in the description of the RSS entry.

Should we use the update title as the RSS entry title?

mattiaverga:
Since I think the users want to read the package names in the RSS title, I've changed my PR to show `update.title` as the RSS title, while `update.alias` is now showed in the update description.
This is the actual result:
Screenshot_20190909_181146

ryanlerch (approved):
LGTM

didierm:
Issue #3272 was closed 19 days ago, but RSS feeds are still Borked : 

https://bodhi.fedoraproject.org/rss/updates/?release=F30&type=newpackage

image

mattiaverga:
@didierm the fix will be deployed in production in a future release

didierm:
Descriptive feeds are back , thanks.

PR #3281, opened 2019-06-02
Remove unused models.Build.get_url

sebwoj:


abompard (approved):
Yep, I don't see any other use either.

PR #3289, opened 2019-06-04
Correct grammar on a comment that Bodhi writes

bowlofeggs:
Bodhi had a grammatical error in the message it writes when an
update's test_gating_status field changes. This commit corrects
that error.

### DIFF
--- a/bodhi/server/models.py
+++ b/bodhi/server/models.py
@@ -3555,7 +3555,7 @@ class Update(Base):
         if value != old:
             target.comment(
                 Session(),
-                f"This update test gating status has been changed to '{value}'.",
+                f"This update's test gating status has been changed to '{value}'.",
                 author="bodhi",
             )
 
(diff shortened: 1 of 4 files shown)


tdawson (commented):
Very good catch.  It makes things more clear, as well as better grammer.
My only question is whether it should be "This update's test_gating_status has been changed to 'failed'" ?  So it's clear what variable has been changed.
But I guess it depends on who we expect to see this.  If it's an end user at the command line, leave out the underscore.  If it's an admin parsing logs, I'd appreciate the underscore because it's easier to grep.

bowlofeggs:
On Tue, 2019-06-04 at 14:08 -0700, Troy Dawson wrote:
> My only question is whether it should be "This update's
> test_gating_status has been changed to 'failed'" ? So it's clear what
> variable has been changed.
> But I guess it depends on who we expect to see this. If it's an end
> user at the command line, leave out the underscore. If it's an admin
> parsing logs, I'd appreciate the underscore because it's easier to
> grep.

This text is a comment that Bodhi writes onto the update, so end users
will be seeing it.

tdawson (approved):
If this is something that end users are seeing, then it looks good to me.
Approved.

PR #3295, opened 2019-06-06
Upgrade Vagrant to Fedora 30 and run ansible with Python 3.

cverna:
This commit update the vagrant base image to Fedora 30. Because of
the version of rabbitmq released in Fedora 30 the rabbitmq ansible module
does not work, so a patched copy is inculded in this commit.
This can be removed when https://github.com/ansible/ansible/pull/50381 is
released.

### DIFF
--- a/devel/ansible/roles/bodhi/tasks/rabbitmq.yml
+++ b/devel/ansible/roles/bodhi/tasks/rabbitmq.yml
@@ -24,6 +24,14 @@
   notify:
     - reload rabbitmq
 
+- name: Make sure /var/lib/rabbitmq is owned by the rabbitmq user
+  file:
+    path: /var/lib/rabbitmq
+    state: directory
+    recurse: yes
+    owner: rabbitmq
+    group: rabbitmq
+
 - name: start rabbitmq
   service: name=rabbitmq-server state=started enabled=yes
 
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -13,9 +13,9 @@ require 'etc'
 VAGRANTFILE_API_VERSION = "2"
 
 Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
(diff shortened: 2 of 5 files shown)


devel/ansible/roles/bodhi/tasks/rabbitmq.yml
```diff
@@ -24,6 +24,14 @@
   notify:
     - reload rabbitmq
 
+- name: Make sure /var/lib/rabbitmq is owned by the rabbitmq user
```
abompard, line 27: Is this tied to a bug in the F30 rabbitmq package? If so, it would be nice to add a reference to that bug so we can remove those lines when it's fixed.

abompard (commented):
One minor thing, otherwise it LGTM.

devel/ansible/roles/bodhi/tasks/rabbitmq.yml
cverna: > Is this tied to a bug in the F30 rabbitmq package? If so, it would be nice to add a reference to that bug so we can remove those lines when it's fixed.

To be honest I have not investigated further than that, the rabbitmq-server service was failing because of the wrong permission on this directory. I ll check if there is a bug for that in BZ

cverna: So for some reason `/var/lib/rabbitmq/.erlang.cookie` is created and owned by root before we try to start the service. I ll try to investigate that further

bowlofeggs:
Production is on F29, and I usually only update it every other release (i.e., I would normally skip F30 and wait for F31). If we do this, we won't be developing on the same release as production (though we will still have CI).

cverna:
> Production is on F29, and I usually only update it every other release (i.e., I would normally skip F30 and wait for F31). If we do this, we won't be developing on the same release as production (though we will still have CI).

Ha ok that's fine so let's close this, I mostly did that to update my vagrant box anyway.

PR #3299, opened 2019-06-06
Convert the client to use f-strings

bowlofeggs:


abompard (approved):
Looks good to me, the CI failures are unrelated.

Issue #3300, opened 2019-06-06
Bodhi's docs don't build with Sphinx 2.1.0


bowlofeggs:
Due to https://github.com/Cornices/cornice.ext.sphinx/issues/22 our autogenerated REST API docs won't build with Sphinx 2.1.0. I plan to pin us to Sphinx < 2.1 temporarily so that our pip CI tests will unblock our PRs, but we are going to have to solve this problem because eventually Rawhide is going to get that version of Sphinx and then we'll be in trouble.

Here are a few options we could consider:

0. Fix ```cornice_sphinx```. I have contributed a few patches before. I'm not sure how involved this fix will be.
1. Switch to Django - there are a lot of benefits to doing this, but it is also a lot of work and we don't likely have the time to do it before Rawhide gets the new Sphinx (or before F31 is branched).
2. Hand write our docs instead. We could use https://sphinxcontrib-httpdomain.readthedocs.io/en/stable/ to assist in formatting, and we could also put the docs in-line in the Python service files. This would be a lot of work up front, and would also require us to maintain them as we change the API (which is why keeping them in-line in the Python code would be wise - that way we are more likely to remember to adjust them.)

Other ideas welcome.

mattiaverga:
This has been fixed, I forgot to close

PR #3301, opened 2019-06-06
Avoid sphinx >= 2.1 in the pip CI test

bowlofeggs:
Sphinx 2.1.0 does not work with cornice_sphinx[0]. This is causing
our CI to fail, but fixing it is not easy. For now, we are going
to install an older Sphinx until we can figure out what to do to
work with Sphinx >= 2.1.0.

re #3300

[0] https://github.com/Cornices/cornice.ext.sphinx/issues/22

### DIFF
--- a/devel/ci/Dockerfile-pip
+++ b/devel/ci/Dockerfile-pip
@@ -37,6 +37,7 @@ RUN pip-3 install \
     pydocstyle \
     pytest \
     pytest-cov \
+    "sphinx<2.1" \
     sqlalchemy_schemadisplay \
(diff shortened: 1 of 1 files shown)


puiterwijk (approved):
Given that this is not used in production code, this gets a 90-day exemption from the `MUST NOT set maximum versions` statement in the [Fedora Infrastructure Application Security Policy, Dependencies section](https://fedora-infra-docs.readthedocs.io/en/latest/dev-guide/security_policy.html#dependencies).

PR #3307, opened 2019-06-11
Type annotate the client

bowlofeggs:

### DIFF
--- a/bodhi/client/__init__.py
+++ b/bodhi/client/__init__.py
@@ -281,6 +281,8 @@ def _save_override(url: str, user: str, password: str, staging: bool, edit: bool
         password: The user's password.
         staging: Whether to use the staging server or not.
         edit: Set to True to edit an existing buildroot override.
+        openid_api: The URL to an OpenID server that can be used to authenticate to the Bodhi
+            server.
         kwargs: Other keyword arguments passed to us by click.
     """
     client = bindings.BodhiClient(base_url=url, username=user, password=password, staging=staging,
@@ -453,6 +455,7 @@ def _validate_edit_update(
     The update argument can only be an update id.
 
     Args:
+        ctx: The click Context. Unused.
         param: The name of the parameter being validated. Unused.
         value: The value of the value being validated.
     Returns:
@@ -599,6 +602,7 @@ def query(url: str, debug: bool, mine: bool = False, rows: typing.Optional[int] 
                        True.
         mine: If the --mine flag was set
         debug: If the --debug flag was set
+        rows: How many rows to fetch
         kwargs: Other keyword arguments passed to us by click.
     """
     client = bindings.BodhiClient(base_url=url, staging=kwargs['staging'])
--- a/bodhi/client/__init__.py
+++ b/bodhi/client/__init__.py
@@ -600,8 +600,8 @@ def query(url: str, debug: bool, mine: bool = False, rows: typing.Optional[int] 
     Args:
         url: The URL of a Bodhi server to create the update on. Ignored if staging is
                        True.
-        mine: If the --mine flag was set
         debug: If the --debug flag was set
+        mine: If the --mine flag was set
         rows: How many rows to fetch
         kwargs: Other keyword arguments passed to us by click.
     """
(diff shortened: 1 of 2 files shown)


bodhi/client/__init__.py
```diff
@@ -65,17 +68,19 @@ def _warn_if_url_or_openid_and_staging_set(ctx, param, value):
...
 
     Args:
-        ctx (click.core.Context): The Click context. Unused.
-        param (click.core.Option): The option being handled. Unused.
-        value (bool): The value of the --debug flag.
+        ctx: The Click context. Unused.
+        param: The option being handled. Unused.
+        value: The value of the --debug flag.
     Returns:
         bool: The value of the --debug flag.
```
sebwoj: Let's remove bool.

bodhi/client/__init__.py
```diff
@@ -261,18 +267,18 @@ def wrapper(*args, **kwargs):
...
-        user (unicode): The username to authenticate as.
-        password (unicode): The user's password.
-        staging (bool): Whether to use the staging server or not.
-        edit (bool): Set to True to edit an existing buildroot override.
-        kwargs (dict): Other keyword arguments passed to us by click.
+        url: The URL of a Bodhi server to create the update on. Ignored if staging is True.
+        user: The username to authenticate as.
+        password: The user's password.
+        staging: Whether to use the staging server or not.
+        edit: Set to True to edit an existing buildroot override.
```
sebwoj, line 283: `openid_api` missing

bodhi/client/__init__.py
```diff
@@ -437,17 +441,19 @@ def new(user, password, url, debug, openid_api, **kwargs):
...
+        value: str) -> str:
     """
     Validate the update argument given to the updates edit command.
 
     The update argument can only be an update id.
 
     Args:
-        param (basestring): The name of the parameter being validated. Unused.
-        value (basestring): The value of the value being validated.
+        param: The name of the parameter being validated. Unused.
```
sebwoj, line 459: `ctx` is missing

bodhi/client/__init__.py
```diff
@@ -587,11 +592,11 @@ def query(url, debug, mine=False, rows=None, **kwargs):
...
 
     Args:
-        url (unicode): The URL of a Bodhi server to create the update on. Ignored if staging is
+        url: The URL of a Bodhi server to create the update on. Ignored if staging is
                        True.
-        mine (Boolean): If the --mine flag was set
-        debug (Boolean): If the --debug flag was set
-        kwargs (dict): Other keyword arguments passed to us by click.
+        mine: If the --mine flag was set
+        debug: If the --debug flag was set
```
sebwoj, line 603: `rows` is missing + what do you think about changing order of args in `Args` section to match the order of args in function?

bodhi/client/__init__.py
```diff
@@ -897,28 +899,29 @@ def overrides():
...
-        url (unicode): The URL of a Bodhi server to create the update on. Ignored if staging is
-                       True.
-        packages (unicode): If supplied, the overrides for these package are queried
-        expired (bool): If supplied, True returns only expired overrides, False only active.
-        releases (unicode): If supplied, the overrides for these releases are queried.
-        builds (unicode): If supplied, the overrides for these builds are queried.
-        rows (unicode): The limit of rows displayed per page for query result.
-        page (unicode): If supplied, returns the results for a specific page number.
-        kwargs (dict): Other keyword arguments passed to us by click.
+        user: If supplied, overrides for this user will be queried.
```
sebwoj, line 931: The same problem with args order here.

sebwoj (changes requested):
Only minor problems.

bodhi/client/__init__.py
bowlofeggs: I'm applying this patch:

```
diff --git a/bodhi/client/__init__.py b/bodhi/client/__init__.py
index 8004f065..549d20ae 100644
--- a/bodhi/client/__init__.py
+++ b/bodhi/client/__init__.py
@@ -82,7 +82,7 @@ def _set_logging_debug(
         param: The option being handled. Unused.
         value: The value of the --debug flag.
     Returns:
-        bool: The value of the --debug flag.
+        The value of the --debug flag.
     """
     if value:
         ch = logging.StreamHandler()
```

bodhi/client/__init__.py
bowlofeggs: For this and the other missing things, I will make a second commit - teeeechnically they aren't really part of switching to type annotation, but I agree they should be documented.

bodhi/client/__init__.py
bowlofeggs: I'll do a third commit to switch the arg order to match.

puiterwijk (approved):
LGTM

Zlopez:
Resolved conflict

PR #3315, opened 2019-06-15
Correct wrong test and rename test_get_single_update_rss_fail_406

mattiaverga:
Correct wrong URL + better name and description of what the test does.


cverna (approved):
:+1:

PR #3317, opened 2019-06-17
Do not ask for_original_spec_nvr results from greenwave.

cverna:
Greenwave already takes care of getting the tests results for original_spec_nvr
in resultsdb, hence bodhi does not need to ask for it.
See greenwave code (https://pagure.io/greenwave/blob/master/f/greenwave/resources.py#_95)

This will have for effect to improve the performance of bodhi's
requests to greenwave.

### DIFF
--- a/bodhi/server/models.py
+++ b/bodhi/server/models.py
@@ -1884,9 +1884,6 @@ class Update(Base):
         """
         # See discussion on https://pagure.io/greenwave/issue/34 for why we use these subjects.
         subject = [{'item': build.nvr, 'type': 'koji_build'} for build in self.builds]
-        # Additionally add some subject entries for "CI"
-        # https://pagure.io/greenwave/issue/61
-        subject.extend([{'original_spec_nvr': build.nvr} for build in self.builds])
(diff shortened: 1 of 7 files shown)


nphilipp (commented):
The changes look good to me, also considering recent email conversation concerning Bodhi/Greenwave performance.

Two thoughts about the commit message:

- The title contains a typo, I guess `for_original_spec_nvr` should be `for original_spec_nvr`.
- Referencing Greenwave code by branch name isn't stable. Perhaps link the current tip of master by hash instead? E.g.:

```
See the relevant code in Greenwave:

https://pagure.io/greenwave/blob/4e427729f711b73a3018614cf9c889efa3b80ce1/f/greenwave/resources.py#_95
```

cverna:
Update the commit message after @nphilipp comments.

nphilipp (approved):
Thanks!

Approved, as per discussion between @abompard, @cverna and myself.

abompard (approved):
After looking at why the need for `original_spec_nvr` arised, it's apparently been partly removed in the JS (`update.html`) so it's safe to finish removing it there.
I don't think we need it anywhere anymore.

PR #3319, opened 2019-06-17
Avoid NACK loops in the automatic updates consumer

nphilipp:
The first commit contains an otherwise unrelated fix which is needed for the reworked unit tests to be able to verify that the expected log messages are issued.

### DIFF
--- a/bodhi/server/scripts/approve_testing.py
+++ b/bodhi/server/scripts/approve_testing.py
@@ -35,7 +35,6 @@ from ..config import config
 
 
 logger = logging.getLogger('approve-testing')
-logging.basicConfig(level=logging.ERROR)
 
 
 def usage(argv):
@@ -65,6 +64,8 @@ def main(argv=sys.argv):
     Args:
         argv (list): A list of command line arguments. Defaults to sys.argv.
     """
+    logging.basicConfig(level=logging.ERROR)
+
     if len(argv) != 2:
         usage(argv)
 
--- a/bodhi/tests/server/scripts/test_approve_testing.py
+++ b/bodhi/tests/server/scripts/test_approve_testing.py
@@ -20,6 +20,7 @@ This module contains tests for the bodhi.server.scripts.approve_testing module.
 """
 from datetime import datetime, timedelta
 from io import StringIO
+import logging
 from unittest.mock import call, patch
 
 from fedora_messaging import api, testing as fml_testing
@@ -31,6 +32,9 @@ from bodhi.server.scripts import approve_testing
 from bodhi.tests.server.base import BaseTestCase
 
 
+# Don't muck around with global log level.
+@patch('bodhi.server.scripts.approve_testing.logging.basicConfig',
+       new=lambda *p, **k: None)
 class TestMain(BaseTestCase):
     """
     This class contains tests for the main() function.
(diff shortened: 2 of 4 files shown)


tdawson:
For my fix, Randy suggested I use log.debug instead of log.warning.
log.warning is going to send out an email to bodhi maintainers for each bad message.
We need to be able to look through debug logs and see why things didn't happen as we expected.   But if some other program is having issues, we don't need to spam the bodhi maintainers.

nphilipp:
@tdawson aah, I'll change that then.

bodhi/server/scripts/approve_testing.py
```diff
@@ -35,7 +35,9 @@
 
 
 logger = logging.getLogger('approve-testing')
-logging.basicConfig(level=logging.ERROR)
+if __name__ == '__main__':
```
abompard: That won't work, when the script is run as `bodhi-approve-testing` its `__name__` value is `bodhi.server.scripts.approve_testing`.
You could, for example, check that `logger.handlers` is an empty list or not. If it's empty, configure logging, otherwise do nothing.
This also looks like something that should be done in `main()`, not in the module itself.

abompard (changes requested):

bodhi/server/scripts/approve_testing.py
nphilipp: Thanks, I'll change it.

bowlofeggs:
On Mon, 2019-06-17 at 13:01 -0700, Troy Dawson wrote:
> log.warning is going to send out an email to bodhi maintainers for
> each bad message

Our production config is set so that error and exception both e-mail -
warning and info get logged (not e-mailed) and debug is currently not
turned on (we can easily turn it on when we want to debug something, of
course).

bodhi/server/scripts/approve_testing.py
```diff
@@ -68,6 +67,8 @@ def main(argv=sys.argv):
     if len(argv) != 2:
         usage(argv)
 
+    logging.basicConfig(level=logging.ERROR)
```
abompard: Hmm, if you do just that, what happens when the `main()` function is called in the unit tests? Does it change the log level globally for all the tests?

bowlofeggs:
On Wed, 2019-06-19 at 02:19 -0700, Aurélien Bompard wrote:
> logging.basicConfig(level=logging.ERROR)

We should try to use the /etc/bodhi/logging.yml that was introduced
with Bodhi 4.0.0, rather than hard coding logging configs.

nphilipp:
> On Wed, 2019-06-19 at 02:19 -0700, Aurélien Bompard wrote: logging.basicConfig(level=logging.ERROR)
> We should try to use the /etc/bodhi/logging.yml that was introduced with Bodhi 4.0.0, rather than hard coding logging configs.

@bowlofeggs Sure, but we should do it in a different PR.

nphilipp:
@abompard The latest push should address handling of the global log level in tests:

* It moves the `logging.basicConfig()` call into `main()` just as before.
* The `TestMain` class (which tests `approve_testing.main()`) patches `logging.BasicConfig()` to be a NOP.
* It adds a test for that, i.e. that invoking `approve_testing.main()` from the `TestMain` class actually leaves the log level alone.

bowlofeggs:
On Wed, 2019-06-19 at 08:00 -0700, Nils Philippsen wrote:
> @bowlofeggs Sure, but we should do it in a different PR.

Agreed. I've also been thinking it would be very nice to have a single
"process set up" function that all of Bodhi's various processes (web,
message consumers, and the litany of CLIs) would use. That function
could be the thing that sets up logging, database connections,
buildsystem, bugzilla, etc. It's been on my (huge) todo list for years
and just haven't gotten to it.

bodhi/tests/server/scripts/test_approve_testing.py
```diff
@@ -483,3 +487,32 @@ def test_usage(self, stdout, exit):
...
+
+        saved_log_level = logging.root.level
+        sentinel = logging.root.level = SentinelInt(logging.WARNING)
+
+        with patch('bodhi.server.scripts.approve_testing.initialize_db'):
+            with patch('bodhi.server.scripts.approve_testing.get_appsettings', return_value=''):
+                approve_testing.main(['nosetests', 'some_config.ini', 'testnoses'])
+
+        self.assertIs(logging.root.level, sentinel)
+        logging.root.level = saved_log_level
```
bowlofeggs, line 518: Heh, that's a pretty elaborate and clever test!

bowlofeggs (commented):
LGTM - I'll leave it for @abompard to formally approve.

abompard (approved):

bodhi/server/scripts/approve_testing.py
nphilipp: Yes, it would, but my latest changes address this.

PR #3320, opened 2019-06-17
Type and name fixes

nphilipp:
This PR contains the commits that fix type names and make validator names less ambiguous which are required, but not topical to PR#3228.


cverna (approved):
LGTM, thanks

PR #3330, opened 2019-06-19
Update Developer documentation

Zlopez:

### DIFF
--- a/docs/developer/index.rst
+++ b/docs/developer/index.rst
@@ -12,16 +12,16 @@ Before you submit a pull request to Bodhi, please ensure that it meets these cri
 
 * All tests must pass.
 * New code must have 100% test coverage. This one is particularly important, as we don't want to
-  deploy any broken code into production. After you've run ``btest``, you can verify your new code's
-  test coverage with ``bdiff-cover`` in the Vagrant environment, or
-  ``diff-cover coverage.xml --compare-branch=origin/develop --fail-under=100`` if you are not using
-  Vagrant.
+  deploy any broken code into production. At the end of ``btest`` run, you can see your code coverage.
 * New functions, methods, and classes must have docblocks that explain what the code block is, and
   describing any parameters it accepts and what it returns (if anything). You can use the
-  ``pydocstyle`` utility to automatically check your code for this.
-* Parameter and return value types should be declared using `type hints`_.
+  ``pydocstyle`` utility to automatically check your code for this. You can also run ``bci pydocstyle``
+  in Vagrant.
+* Parameter and return value types should be declared using `type hints`_. You can test this by running
+  ``bci mypy`` in Vagrant.
 * New code must follow `PEP-8 <https://www.python.org/dev/peps/pep-0008/>`_. You can use the
-  ``flake8`` utility to automatically check your code.
+  ``flake8`` utility to automatically check your code. Alternatively you can run ``bci flake8``
+  in Vagrant.
 * Add an entry to `docs/user/release_notes.rst`_ for any changes you make that should be in release
   notes.
 * Make sure your commits are atomic. Each commit should focus on one improvement or bug fix. If you


abompard (approved):

docs/developer/index.rst
```diff
@@ -12,16 +12,16 @@ Before you submit a pull request to Bodhi, please ensure that it meets these cri
 
 * All tests must pass.
 * New code must have 100% test coverage. This one is particularly important, as we don't want to
-  deploy any broken code into production. After you've run ``btest``, you can verify your new code's
-  test coverage with ``bdiff-cover`` in the Vagrant environment, or
-  ``diff-cover coverage.xml --compare-branch=origin/develop --fail-under=100`` if you are not using
-  Vagrant.
```
bowlofeggs, line 18: Why drop the documentation on how to run diff-cover for non-Vagrant users?

docs/developer/index.rst
```diff
@@ -12,16 +12,16 @@ Before you submit a pull request to Bodhi, please ensure that it meets these cri
...
-  ``pydocstyle`` utility to automatically check your code for this.
-* Parameter and return value types should be declared using `type hints`_.
+  ``pydocstyle`` utility to automatically check your code for this. You can also run ``bci pydocstyle``
+  in Vagrant.
+* Parameter and return value types should be declared using `type hints`_. You can test this by running
+  ``bci mypy`` in Vagrant.
 * New code must follow `PEP-8 <https://www.python.org/dev/peps/pep-0008/>`_. You can use the
-  ``flake8`` utility to automatically check your code.
+  ``flake8`` utility to automatically check your code. Alternatively you can run ``bci flake8``
+  in Vagrant.
```
bowlofeggs, line 24: Though it is true that you can use ```bci``` in the ways described above, I would not recommend it because we have bash shell aliases that will be faster (except for mypy). For example, pydocstyle and flake8 are both run by the ```bteststyle``` alias, which is significantly faster than the bci equivalent.

docs/developer/index.rst
Zlopez: I thougt the btest is also working outside Vagrant.

docs/developer/index.rst
Zlopez: I didn't know about bteststyle.

Zlopez:
@bowlofeggs Should I create another PR to address the issues?

This one is already merged.

abompard:
Whoops, looks like I approved too soon. Could you rework that please Michal?

docs/developer/index.rst
abompard: Nah it's a bash alias setup in vagrant

bowlofeggs:
On Wed, 2019-06-19 at 07:27 -0700, Zlopez wrote:
> @bowlofeggs Should I create another PR to address the issues?
> This one is already merged.

Yeah we probably should so people don't run the slow commands all the
time. The docs aren't incorrect, they just are inefficient
recommendations. bci is mostly intended to help you to reproduce
problems when they occur in CI locally so it's easier to fix. For the
most part developers don't need to do that and can just run all the
tests directly in the VM rather than containers, which is hugely more
efficient.

The current exception is mypy because we only run that on a subset of
the files at the moment and that list is defined in devel/ci/bodhi-ci
currently. So the bteststyle alias actually does run bci mypy just for
that one test. We could probably find a way to define the list of files
somewhere else and have the shell alias and bodhi-ci both use that
single list to eliminate that as well.

abompard:
@bowlofeggs Yeah apparently mypy can load its options from [a configuration file](https://mypy.readthedocs.io/en/stable/config_file.html).

bowlofeggs:
On Thu, 2019-06-20 at 01:10 -0700, Aurélien Bompard wrote:
> @bowlofeggs Yeah apparently mypy can load its options from a
> configuration file.

Ah nice - I hadn't noticed before that there was a files option in
there. I know you could give it settings (we actually are doing that)
but missed that I could define the enforcement path there too. I'll
send a PR today to switch to that so we can make the bteststyle go
fast.

bowlofeggs:
On Thu, 2019-06-20 at 01:10 -0700, Aurélien Bompard wrote:
> @bowlofeggs Yeah apparently mypy can load its options from a
> configuration file.

https://github.com/fedora-infra/bodhi/pull/3335

Issue #3248, opened 2019-05-28
Bodhi redirects web UI users to a non-existant URL after creating updates


bowlofeggs:
Bodhi is redirecting users to ```/updates/{title}``` instead of ```/updates/{alias}``` when they create updates. The title is no longer a valid way to access updates, so this leads the user to a 404. We should instead redirect them to the alias URL.

mattiaverga:
This is strange, the javascript that manages the update form is correctly set to redirect to `/updates/{alias}`:
https://github.com/fedora-infra/bodhi/blob/e42d7653a556a3e11ae0b961b6acc8772994037e/bodhi/server/static/js/update_form.js#L18
Indeed, the redirect works well in my local Vagrant VM.

bowlofeggs:
I wonder if the user that reported it had stale cached JS in their browser?

bowlofeggs:
I'm going to close this, as I haven't heard anything further about it. I think we can assume it was cached JS.

PR #3332, opened 2019-06-20
Use versioned URL for static resources

### LINKED ISSUE #1424: Blank dark-gray banner at the top of screen in chrome
It only shows up for me in chrome, but not firefox.

/cc @ryanlerch

mattiaverga:
This will make the static resources available under a dynamically named URL based on Bodhi version number. Instead of `{base_url}/static/` it will be, for example, `{base_url/static/v4.0.2/`.
It should avoid browser cache problems that arise after a Bodhi upgrade.

Fixes #1424 (and some other already closed like #3248 )


Zlopez:
Does this change need static content to be in the versioned folder?

mattiaverga:
> Does this change need static content to be in the versioned folder?

No, the files themselves will remain in `bodhi/server/static`, but Pyramid will map them to a URL which changes with every Bodhi version.
Currently it always maps `bodhi/server/static` directory to `{base_url}/static/`, with this change it will map the same dir to `{base_url}/static/{bodhi_version}/`, so after every Bodhi upgrade the user's browser will refresh its cache because it thinks to load a different file.

Zlopez (approved):

PR #3333, opened 2019-06-20
Update developer documentation

Zlopez:
This PR addresses issues in #3330

### DIFF
--- a/docs/developer/index.rst
+++ b/docs/developer/index.rst
@@ -13,15 +13,18 @@ Before you submit a pull request to Bodhi, please ensure that it meets these cri
 * All tests must pass.
 * New code must have 100% test coverage. This one is particularly important, as we don't want to
   deploy any broken code into production. At the end of ``btest`` run, you can see your code coverage.
+  If you are not using Vagrant environment you can check the code coverage by running
+  ``diff-cover coverage.xml --compare-branch=origin/develop --fail-under=100``.
 * New functions, methods, and classes must have docblocks that explain what the code block is, and
   describing any parameters it accepts and what it returns (if anything). You can use the
-  ``pydocstyle`` utility to automatically check your code for this. You can also run ``bci pydocstyle``
+  ``pydocstyle`` utility to automatically check your code for this. You can also run ``pydocstyle bodhi``
   in Vagrant.
 * Parameter and return value types should be declared using `type hints`_. You can test this by running
   ``bci mypy`` in Vagrant.
 * New code must follow `PEP-8 <https://www.python.org/dev/peps/pep-0008/>`_. You can use the
-  ``flake8`` utility to automatically check your code. Alternatively you can run ``bci flake8``
+  ``flake8`` utility to automatically check your code. Alternatively you can run ``flake8-3``
   in Vagrant.
+* If you want to run the three above at once you can use ``blint`` alias in Vagrant.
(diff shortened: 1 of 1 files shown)


docs/developer/index.rst
```diff
@@ -13,14 +13,16 @@ Before you submit a pull request to Bodhi, please ensure that it meets these cri
 * All tests must pass.
 * New code must have 100% test coverage. This one is particularly important, as we don't want to
   deploy any broken code into production. At the end of ``btest`` run, you can see your code coverage.
+  If you are not using Vagrant environment you can check the code coverage by running
+  ``diff-cover coverage.xml --compare-branch=origin/develop --fail-under=100``.
 * New functions, methods, and classes must have docblocks that explain what the code block is, and
   describing any parameters it accepts and what it returns (if anything). You can use the
-  ``pydocstyle`` utility to automatically check your code for this. You can also run ``bci pydocstyle``
+  ``pydocstyle`` utility to automatically check your code for this. You can also run ``bteststyle``
```
abompard: Actually, running `bteststyle` will run flake8, pydocstyle and mypy, so I think the document is a bit misleading here. The in-vagrant command to only run pydocstyle is `pydocstyle bodhi`.

docs/developer/index.rst
```diff
@@ -13,14 +13,16 @@ Before you submit a pull request to Bodhi, please ensure that it meets these cri
...
   in Vagrant.
 * Parameter and return value types should be declared using `type hints`_. You can test this by running
-  ``bci mypy`` in Vagrant.
+  ``bteststyle`` in Vagrant.
```
abompard: This can stay, running `bci mypy` is currently the only way to run *only* mypy in Vagrant

docs/developer/index.rst
```diff
@@ -13,14 +13,16 @@ Before you submit a pull request to Bodhi, please ensure that it meets these cri
...
 * New code must follow `PEP-8 <https://www.python.org/dev/peps/pep-0008/>`_. You can use the
-  ``flake8`` utility to automatically check your code. Alternatively you can run ``bci flake8``
+  ``flake8`` utility to automatically check your code. Alternatively you can run ``bteststyle``
```
abompard: Again, `bteststyle` will run more than flake8, the in-vagrant command to only run flake8 is `flake8-3`.

I think it would be better to add a bullet point that will inform the dev that running `bteststyle` will run flake8, then pydocstyle, and then mypy, so it's a good way to check all those linters in one command.

abompard (changes requested):

bowlofeggs:
On Thu, 2019-06-20 at 01:19 -0700, Aurélien Bompard wrote:
> I think it would be better to add a bullet point that will inform the
> dev that running bteststyle will run flake8, then pydocstyle, and
> then mypy, so it's a good way to check all those linters in one
> command.

Yeah and I also wonder if the name bteststyle is not great since mypy
isn't a style checker. Maybe we should rename it to blint or something.

bowlofeggs:
I submitted https://github.com/fedora-infra/bodhi/pull/3336 to rename it to blint.

bowlofeggs (approved):

Zlopez:
I will update this PR to reflect the change.

Zlopez:
Updated to reflect the change in #3336

abompard (approved):

PR #3334, opened 2019-06-20
Prepare for the upcoming 4.1 release

abompard:


docs/user/release_notes.rst
```diff
@@ -2,6 +2,77 @@
...
+
+Features
+^^^^^^^^
+
+* Add autopush and stable_days to an update.  When ``autopush`` is true, the
+  approve_testing cronjob will push the update to stable if it has meet the
+  testing requirements (``stable_days``).  ``stable_days`` cannot be smaller
+  than the release ``mandatory_days_in_testing`` (:issue:`2978`).
+* Add a login reminder for posting feedback.
+* Add package_manager enum and testing_repository string to Release model.
```
mattiaverga: Probably we need to mention that no default is provided, so if one wants Bodhi to display the install command for an update they need to manually edit the already existent releases after the database migration.

abompard: Good point.

abompard:
(not waiting for CI to run because I only changed the release notes)

PR #3335, opened 2019-06-20
Run mypy directly in Vagrant

bowlofeggs:
This commit converts the bteststyle Vagrant alias to run mypy
directly in Vagrant rather than running it in a container. This
makes the command significantly faster. To do this, we moved the
mypy configuration out of bodhi-ci and into setup.cfg.

re #3330


abompard (approved):

Zlopez (approved):

Zlopez:
Resolved conflict

PR #3340, opened 2019-06-20
Add type hinting to bodhi.server.config

bowlofeggs:


abompard:
That's weird, flake8 has decided to scan the whole venv...

Zlopez (approved):

PR #3341, opened 2019-06-20
Only format log messages when necessary

bowlofeggs:
It is inefficient to format log messages in the call to the logger.
Instead, we should let the logger format the strings if the logger
is at the right log level. This will improve performace for
deployments that only log exceptions or other high levels.

### DIFF
--- a/bodhi/server/bugs.py
+++ b/bodhi/server/bugs.py
@@ -57,7 +57,7 @@ class FakeBugTracker(object):
             args: The list of args passed to the method.
             kwargs: The kwargs passed to the method.
         """
-        log.debug('__noop__(%s)' % str(args))
+        log.debug('__noop__(%s)', str(args))
 
     comment = update_details = modified = close = on_qa = __noop__
 
@@ -78,7 +78,7 @@ class Bugzilla(object):
         user = config.get('bodhi_email')
         password = config.get('bodhi_password')
         url = config.get("bz_server")
-        log.info("Using BZ URL %s" % url)
+        log.info("Using BZ URL %s", url)
         if config['bugzilla_api_key']:
             self._bz = bugzilla.Bugzilla(url=url, api_key=config.get('bugzilla_api_key'),
                                          cookiefile=None, tokenfile=None)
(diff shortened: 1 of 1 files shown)


bodhi/server/bugs.py
```diff
@@ -176,10 +176,10 @@ def on_qa(self, bug_id: int, comment: str):
         try:
             bug = self.bz.getbug(bug_id)
             if bug.product not in config.get('bz_products'):
-                log.info("Skipping set on_qa on {0!r} bug #{1}".format(bug.product, bug_id))
+                log.info("Skipping set on_qa on %s bug #%s", bug.product, bug_id)
```
abompard: You may want to have `repr(bug.product)` here to replace the `{!r}`.

bodhi/server/bugs.py
```diff
@@ -209,7 +209,7 @@ def close(self, bug_id: int, versions: typing.Mapping[str, str], comment: str):
         try:
             bug = self.bz.getbug(bug_id)
             if bug.product not in config.get('bz_products'):
-                log.info("Skipping set closed on {0!r} bug #{1}".format(bug.product, bug_id))
+                log.info("Skipping set closed on %s bug #%s", bug.product, bug_id)
```
abompard: Same remark here.

bodhi/server/bugs.py
```diff
@@ -292,10 +292,10 @@ def modified(self, bug_id: typing.Union[int, str], comment: str):
         try:
             bug = self.bz.getbug(bug_id)
             if bug.product not in config.get('bz_products'):
-                log.info("Skipping set modified on {0!r} bug #{1}".format(bug.product, bug_id))
+                log.info("Skipping set modified on %s bug #%s", bug.product, bug_id)
```
abompard: Same remark here.

abompard (changes requested):

bodhi/server/bugs.py
ryanlerch: @abompard have updated @bowlofeggs commit with the suggested changes.

abompard (approved):

PR #3344, opened 2019-06-23
Make stable_days field in update form to use mandatory_days_in_testing

mattiaverga:
This will make the "stable days" field fo the new update form to display an "auto" placeholder when creating a new update, while when editing an existing update the minimum accepted value will be the mandatory_days_in_testing for that update.

It was quite confusing for me to have "Auto-request stable based on time" flagged and "Stable days" with an empty value (zero?). This should make things clearer.

### DIFF
--- a/bodhi/server/templates/new_update.html
+++ b/bodhi/server/templates/new_update.html
@@ -277,13 +277,13 @@ ${update.notes}
                       % endif
                     ></dd>
 
-                  <dt data-toggle="tooltip" title="This is the number of days an update needs to spend in testing to be automatically pushed to stable.">
+                  <dt data-toggle="tooltip" title="This is the number of days an update needs to spend in testing to be automatically pushed to stable. If empty it will be set to the minimum number of days set for the Release.">
                   Stable days</dt>
-                  <dd> <input type="number" name="stable_days" min="1"
+                  <dd> <input type="number" name="stable_days" placeholder="auto"
                       % if update:
-                      value="${update.stable_days}"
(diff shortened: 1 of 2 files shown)


bodhi/server/templates/new_update.html
```diff
@@ -277,13 +277,13 @@ <h2 class="float-left">Edit
...
+                  <dt data-toggle="tooltip" title="This is the number of days an update needs to spend in testing to be automatically pushed to stable. If empty it will be set to the minimum number of days set for the Release.">
                   Stable days</dt>
-                  <dd> <input type="number" name="stable_days" min="1"
+                  <dd> <input type="number" name="stable_days" placeholder="auto"
                       % if update:
-                      value="${update.stable_days}"
+                      min="${update.mandatory_days_in_testing}" value="${update.stable_days}"
                       % elif not update:
-                      value=""
+                      min="1" value=""
```
cverna: I think actually we want this to be 0 since we will have releases with 0 mandatory days in testing like rawhide.

cverna (commented):
Thanks, looks good just one small comment

cverna (approved):
:+1:

cverna:
@abompard I forgot to add the backport 4.1 label to this PR but I think it could go into that build, since it improve the UI for the autotime updates.
I leave it to you to decide if we want to include this one in the 4.1 or not

abompard:
I wonder if Mergify can pick it up if I add the tag even after it's merged or if I should just cherry-pick it. Testing that.

abompard:
It does! Way to go Mergify! :-)

PR #3346, opened 2019-06-25
Make sure that we mock call to fedora messaging in test_approve_testing.

### LINKED ISSUE #3338: The tests for autotime pushing take an enormous amount of time to run
I noticed that the unit tests are very slow in CI (and also in bci when run in Vagrant). I see that this test runs for an extreme amount of time (on the order of minutes):

https://github.com/fedora-infra/bodhi/blob/9be769b4dc53e3bd65d0beb94d6708f97c49e52d/bodhi/tests/server/scripts/test_approve_testing.py#L590-L609

The other autotime tests also take minutes to run, resulting in these new tests causing the CI to run for a very long time.

cverna:
Fixes #3338


abompard (approved):
Seems legit :-)

PR #3347, opened 2019-06-25
Authorize null mandatory days in testing

cverna:
This PR has 2 commits, the first commit makes sure that we can set a specific release mandatory days to 0 and the second one protect the code from trying to cast `None` to an integer.


abompard:
I think you can squash those commits, and then it's good to go

Zlopez (commented):
Looks good, just squash the commits :-)

cverna:
@abompard @Zlopez squashed :-)

Zlopez (approved):

Issue #2319, opened 2018-04-26
[L] API users should be able to merge side tags


bowlofeggs:
We need to add the ability for Bodhi to merge side tags. Bodhi should make sure the tests pass before merging, and it should also ensure there will be no conflicts due to the merge. Once the side tag update is merged, the update should either progress to stable (for Rawhide), or should go to updates-testing (for other releases).

If the merge fails, Bodhi should ensure that the packager has the information they need to resolve the conflict and try again. We will need to watch their side tag for new or removed builds and update the update accordingly by using a fedora-messaging consumer.

nphilipp:
@bowlofeggs:

> Bodhi should make sure the tests pass before merging, and it should also ensure there will be no conflicts due to the merge.

Is this functionality in Bodhi already, or are there issues/cards for tests passing and no merge conflicts?

> We will need to watch their side tag for new or removed builds and update the update accordingly by using a fedora-messaging consumer.

This sounds like it could be its own task, what do you think?

bowlofeggs:
On Mon, 2019-05-27 at 06:14 -0700, Nils Philippsen wrote:
> > Bodhi should make sure the tests pass before merging, and it should
> > also ensure there will be no conflicts due to the merge.
> 
> Is this functionality in Bodhi already, or are there issues/cards for
> tests passing and no merge conflicts?

Hey Nils!

We do have a "do the tests pass" in Bodhi already (the
Update.test_gating_status field), but we do not have any code for
merging things in Koji or detecting conflicts.

> > We will need to watch their side tag for new or removed builds and
> > update the update accordingly by using a fedora-messaging consumer.
> 
> This sounds like it could be its own task, what do you think?

Sure, we could split it up if you prefer. Feel free to file another
issue for this part of it and put it on the backlog. If you do so,
please link it here as well.

cverna:
As a general note merging tag sounds like something we want to do in a celery tasks, so this ticket really depends on #2851

cverna:
The script used by releng to merge side tags is (https://pagure.io/releng/blob/master/f/scripts/mass-tag.py)

mohanboddu:
I want to bring up a use case here:

[...]

PR #3353, opened 2019-06-26
Add option to create/edit updates side tags to CLI

### LINKED ISSUE #2325: [S] The CLI needs a way for users to create side tag updates
Similarly to #2324, the CLI also needs to be able to create side tag updates. This will also enable ```fedpkg``` to create side tag updates via the Bodhi CLI, if they like.

Zlopez:
Fixes #2325

### DIFF
--- a/bodhi/client/__init__.py
+++ b/bodhi/client/__init__.py
@@ -134,6 +134,9 @@ new_edit_options = [
                  type=click.Choice(['logout', 'reboot'])),
     click.option('--display-name',
                  help='The name of the update'),
+    click.option('--from-tag', help='Use builds from a Koji tag instead of specifying '
+                                    'them individually.',
+                 is_flag=True),
     staging_option]
 
 
@@ -395,7 +398,7 @@ def require_severity_for_security_update(type: str, severity: str):
 @click.option('--type', default='bugfix', help='Update type', required=True,
               type=click.Choice(['security', 'bugfix', 'enhancement', 'newpackage']))
 @add_options(new_edit_options)
-@click.argument('builds')
+@click.argument('builds_or_tag')
 @click.option('--file', help='A text file containing all the update details')
 @handle_errors
 @openid_option
@@ -419,9 +422,19 @@ def new(user: str, password: str, url: str, debug: bool, openid_api: str, **kwar
         openid_api: A URL for an OpenID API to use to authenticate to Bodhi.
         kwargs: Other keyword arguments passed to us by click.
     """
-
     client = bindings.BodhiClient(base_url=url, username=user, password=password,
                                   staging=kwargs['staging'], openid_api=openid_api)
+
+    # Because bodhi.server.services.updates expects from_tag to be string
+    # copy builds to from_tag and remove builds
+    if kwargs['from_tag']:
+        if len(kwargs['builds_or_tag'].split(' ')) > 1:
+            click.echo("ERROR: Can't specify more than one tag.", err=True)
+            sys.exit(1)
+        kwargs['from_tag'] = kwargs.pop('builds_or_tag')
+    else:
+        kwargs['builds'] = kwargs.pop('builds_or_tag')
+        del kwargs['from_tag']
 
     if kwargs['file'] is None:
         updates = [kwargs]
@@ -432,7 +445,7 @@ def new(user: str, password: str, url: str, debug: bool, openid_api: str, **kwar
     kwargs['notes'] = _get_notes(**kwargs)
 
     if not kwargs['notes'] and not kwargs['file']:
-        click.echo("ERROR: must specify at least one of --file, --notes, or --notes-file")
+        click.echo("ERROR: must specify at least one of --file, --notes, or --notes-file", err=True)
         sys.exit(1)
 
     for update in updates:
@@ -517,13 +530,32 @@ def edit(user: str, password: str, url: str, debug: bool, openid_api: str, **kwa
 
         kwargs['builds'] = [b['nvr'] for b in former_update['builds']]
         kwargs['edited'] = former_update['alias']
-        if kwargs['addbuilds']:
-            for build in kwargs['addbuilds'].split(','):
-                if build not in kwargs['builds']:
-                    kwargs['builds'].append(build)
-        if kwargs['removebuilds']:
-            for build in kwargs['removebuilds'].split(','):
-                kwargs['builds'].remove(build)
+        # Because bodhi.server.services.updates expects from_tag to be string
+        # copy builds to from_tag and remove builds
+        if kwargs['from_tag']:
+            if not former_update.get('from_tag', None):
+                click.echo(
+                    "ERROR: This update was not created from a tag."
+                    " Please remove --from_tag and try again.", err=True
+                )
+                sys.exit(1)
+            if kwargs['addbuilds'] or kwargs['removebuilds']:
+                click.echo(
+                    "ERROR: The --from-tag option can't be used together with"
+                    " --addbuilds or --removebuilds.", err=True
+                )
+                sys.exit(1)
+            kwargs['from_tag'] = former_update['from_tag']
+            del kwargs['builds']
+        else:
+            kwargs.pop('from_tag')
+            if kwargs['addbuilds']:
+                for build in kwargs['addbuilds'].split(','):
+                    if build not in kwargs['builds']:
+                        kwargs['builds'].append(build)
+            if kwargs['removebuilds']:
+                for build in kwargs['removebuilds'].split(','):
+                    kwargs['builds'].remove(build)
         del kwargs['addbuilds']
         del kwargs['removebuilds']
 
--- a/bodhi/tests/client/test___init__.py
+++ b/bodhi/tests/client/test___init__.py
@@ -757,6 +757,63 @@ class TestNew(unittest.TestCase):
         self.assertEqual(send_request.mock_calls, calls)
         self.assertEqual(bindings_client.base_url, 'http://localhost:6543/')
 
+    @mock.patch('bodhi.client.bindings.BodhiClient.csrf',
+                mock.MagicMock(return_value='a_csrf_token'))
+    @mock.patch('bodhi.client.bindings.BodhiClient.send_request',
+                return_value=client_test_data.EXAMPLE_UPDATE_MUNCH, autospec=True)
+    def test_from_tag_flag(self, send_request):
+        """
+        Assert correct behavior with the --from-tag flag.
+        """
+        runner = testing.CliRunner()
+
+        result = runner.invoke(
+            client.new,
+            ['--user', 'bowlofeggs', '--password', 's3kr3t', '--autokarma', 'fake_tag',
+             '--bugs', '1234567', '--from-tag', '--url',
+             'http://localhost:6543', '--notes', 'No description.'])
+
+        self.assertEqual(result.exit_code, 0)
+        expected_output = client_test_data.EXPECTED_UPDATE_OUTPUT.replace('example.com/tests',
+                                                                          'localhost:6543')
+        self.assertTrue(compare_output(result.output, expected_output + '\n'))
+        bindings_client = send_request.mock_calls[0][1][0]
+        calls = [
+            mock.call(
+                bindings_client, 'updates/', auth=True, verb='POST',
+                data={
+                    'close_bugs': False, 'stable_karma': None, 'csrf_token': 'a_csrf_token',
+                    'staging': False, 'autokarma': True, 'autotime': False, 'stable_days': None,
+                    'suggest': None, 'notes': 'No description.', 'request': None,
+                    'bugs': '1234567', 'requirements': None, 'unstable_karma': None, 'file': None,
+                    'notes_file': None, 'type': 'bugfix', 'severity': None, 'display_name': None,
+                    'from_tag': 'fake_tag'
+                }
+            ),
+            mock.call(
+                bindings_client,
+                'updates/FEDORA-EPEL-2016-3081a94111/get-test-results',
+                verb='GET'
+            )
+        ]
+        self.assertEqual(send_request.mock_calls, calls)
+        self.assertEqual(bindings_client.base_url, 'http://localhost:6543/')
+
+    def test_from_tag_flag_multiple_tags(self):
+        """
+        Assert correct behavior with the --from-tag and multiple tags.
+        """
+        runner = testing.CliRunner()
+
+        result = runner.invoke(
+            client.new,
+            ['--user', 'bowlofeggs', '--password', 's3kr3t', '--autokarma', 'fake tag',
+             '--bugs', '1234567', '--from-tag', '--url',
+             'http://localhost:6543', '--notes', 'No description.'])
+
+        self.assertEqual(result.exit_code, 1)
+        self.assertEqual(result.output, 'ERROR: Can\'t specify more than one tag.\n')
+
     def test_new_update_without_notes(self):
         """
         Assert providing neither --notes-file nor --notes parameters to new update request
@@ -783,7 +840,7 @@ class TestNew(unittest.TestCase):
              '--notes', 'bla bla bla', '--type', 'security'])
 
         self.assertEqual(result.exit_code, 2)
-        self.assertEqual(result.output, ('Usage: new [OPTIONS] BUILDS\n\nError: Invalid '
+        self.assertEqual(result.output, ('Usage: new [OPTIONS] BUILDS_OR_TAG\n\nError: Invalid '
                          'value for severity: must specify severity for a security update\n'))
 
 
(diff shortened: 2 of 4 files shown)


bodhi/client/__init__.py
```diff
@@ -129,6 +129,7 @@ def _set_logging_debug(ctx, param, value):
                  type=click.Choice(['logout', 'reboot'])),
     click.option('--display-name',
                  help='The name of the update'),
+    click.option('--from-tag', help='Builds will be handled as Koji tags', is_flag=True),
```
sebwoj: Let's update [docs](https://github.com/fedora-infra/bodhi/blob/develop/docs/user/man_pages/bodhi.rst)

bodhi/tests/client/test___init__.py
```diff
@@ -721,7 +721,7 @@ def test_close_bugs_flag(self, send_request):
                 return_value=client_test_data.EXAMPLE_UPDATE_MUNCH, autospec=True)
     def test_display_name_flag(self, send_request):
         """
-        Assert correct behavior with the --close-bugs flag.
+        Assert correct behavior with the --display-name flag.
```
sebwoj: This change is not related to this commit. Let's add second commit with this correction.

sebwoj (changes requested):
Only minor issues.

bodhi/client/__init__.py
```diff
@@ -134,6 +134,7 @@ def _set_logging_debug(
...
```
nphilipp: The help string is a little confusing, how about:

```
    help='Use builds from a Koji tag instead of specifying them individually.'
```

bodhi/client/__init__.py
```diff
@@ -419,10 +420,17 @@ def new(user: str, password: str, url: str, debug: bool, openid_api: str, **kwar
...
                                   staging=kwargs['staging'], openid_api=openid_api)
 
+    # Because bodhi.server.services.updates expects from_tag to be string
+    # copy builds to from_tag and remove builds
+    if kwargs['from_tag']:
+        kwargs['from_tag'] = kwargs['builds']
+        del(kwargs['builds'])
+    else:
+        del(kwargs['from_tag'])
+
```
nphilipp, line 438: I'd prefer it very much if we renamed `builds` to e.g. `builds_or_tag` and made the decision on `from_tag` when calling the API because it's more obvious what is what then.

Also (see below), we probably should enforce that users can only specify one tag here.

docs/user/man_pages/bodhi.rst
```diff
@@ -271,6 +271,11 @@ The ``updates`` command allows users to interact with bodhi updates.
 
         The name of the update
 
+    ``--from-tag``
+
+        If this flag is provided ``<builds>`` are handled as Koji tag. We support only single Koji tag
+        at once. Koji tag allows to bundle multiple builds together in one update.
+
```
nphilipp, line 278: Similar here (`builds` -> e.g. `builds_or_tag`, whatever click makes out of the chosen variable name). Also, I'd reword it like this:

```
If this flag is provided, `builds_or_tag` will be interpreted as a Koji tag and expand to all latest builds in it. Only a single tag can be provided.
```

Also, I don't see that we ensure only one tag is specified by the user, would you add this (see above)?

docs/user/man_pages/bodhi.rst
```diff
@@ -334,6 +339,11 @@ The ``updates`` command allows users to interact with bodhi updates.
...
```
nphilipp, line 347: Same here.

nphilipp (changes requested):

nphilipp:
Adding the `WIP` label so we don't forget squashing this -- I have no idea what mergify makes of the merge commits but I absolutely want to avoid them getting in, or mergify to be confused by them.

bodhi/client/__init__.py
Zlopez: This sounds reasonable.

bodhi/client/__init__.py
Zlopez: Should we enforce this on API side or CLI side?

nphilipp: IIRC, the API simply doesn't interpret `from_tag` so that it could mean multiple things. I'm actually not sure what would happen if we simply let the existing (`builds`) logic on the CLI side do what it does for a tag, too... Anyway, I guess it would be best to check for this ("enforce" is such a strong word :wink:) on the client side so the user gets tailored feedback immediately.

Zlopez:
Looking at this code today, I think that edit option is not viable for CLI. Edit does query on the beginning and the query never returns tag only builds according to API.

@nphilipp You did changes to API, what do you think?

nphilipp:
@Zlopez as discussed on IRC, the API endpoints return `from_tag` for me (Vagrant box in the current `develop` branch), the only caveat is that it can be `null` for normal (non-side-tag) updates. As far as the API is concerned, `from_tag` is only informative, i.e. the API lets you edit an update filled from a tag normally by specifying a list of builds and if I'm not completely off track the other way would work, too, i.e. edit a normal update to be filled by builds from a tag.

I guess this is what you're aiming at, that the CLI should only allow repopulating the builds from a tag if the update was created from and for this tag, right?

Zlopez:
This last update should address changes @nphilipp wanted.

Zlopez:
@nphilipp Today meeting made me thinking about this PR. How do you expect CLI should work when editing updates?

Current implementation is this:
1. Query update
2. Do any edit action
3. Check if the update was created using tag when `--from-tag` flag was used
a. If yes, remove builds from request and send tag (we will lose every edit related to builds list made in step 2)
b. If no, just sent the builds or throw error if `--from-tag` flag was used
4. Send request to API

If somebody wants to use `--addbuilds` or `--removebuilds` parameters in edit, should we allow that? Not sure what I should send in this case to edit current sidetag. We must take in consideration that we currently editing update not sidetag directly.

nphilipp:
/cc @ryanlerch @pypingou @cverna @abompard @bowlofeggs 

Sorry for the wide distribution, but I'd appreciate your input here:

> @nphilipp Today meeting made me thinking about this PR. How do you expect CLI should work when editing updates?
> 
> Current implementation is this:
> 
> 1. Query update
> 2. Do any edit action
> 3. Check if the update was created using tag when `--from-tag` flag was used
>    a. If yes, remove builds from request and send tag (we will lose every edit related to builds list made in step 2)

I think we shouldn't do this implicitly, i.e. without the user specifying the `--from-tag` flag again, just as we don't automatically update builds when editing an update just because a newer NVR is available. @ryanlerch what do you think from a UX perspective?

I can think of two ways to handle builds in a "side-tagged" update:

1. "traditionally", i.e. the list of builds is only changed explicitly by a user, either by specifying `--addbuilds`/`--removebuilds` or `--from-tag`, the latter would pull the current set of NVRs from Koji (server-side). This requires the least work, i.e. only some small changes in this PR and the web-UI-side task Ryan is working on.
2. "managed update", i.e. if an update is created from a side tag, the Bodhi server would listen to successful builds in this tag on the bus and update the list of build NVRs automatically. Users wouldn't be able to modify the list of builds in such an update. This would require writing an additional consumer.

I think we're far too late for option 2 if we want this feature to land in time for Flock, besides that we would need to iron out the kinks which corner cases inevitably would cause because we haven't really talked it through. That's not to say we couldn't implement this at a later point if we wanted to.

>    b. If no, just sent the builds or throw error if `--from-tag` flag was used
> 4. Send request to API
> 
> If somebody wants to use `--addbuilds` or `--removebuilds` parameters in edit, should we allow that? Not sure what I should send in this case to edit current sidetag. We must take in consideration that we currently editing update not sidetag directly.

In the CLI, you'd just ignore that the update was originally created from a side tag and submit the list of builds just as it is done for normal updates. I've drawn a flow diagram outlining how I think option 1 from above would be implemented CLI-side (this ignores all the other things you could change when editing an update):

create-edit-update-flow-cli

bodhi/client/__init__.py
```diff
@@ -527,6 +542,19 @@ def edit(user: str, password: str, url: str, debug: bool, openid_api: str, **kwa
         del kwargs['addbuilds']
         del kwargs['removebuilds']
 
+        # Because bodhi.server.services.updates expects from_tag to be string
+        # copy builds to from_tag and remove builds
+        if kwargs['from_tag']:
+            if not former_update.get('from_tag', None):
+                click.echo(
+                    "ERROR: This update is not created from tag."
+                    " Please remove --from_tag and try again")
```
nphilipp: Same as above, `err=True`.

bodhi/client/__init__.py
```diff
@@ -419,10 +422,22 @@ def new(user: str, password: str, url: str, debug: bool, openid_api: str, **kwar
...
                                   staging=kwargs['staging'], openid_api=openid_api)
 
+    # Because bodhi.server.services.updates expects from_tag to be string
+    # copy builds to from_tag and remove builds
+    if kwargs['from_tag']:
+        if len(kwargs['builds_or_tag'].split(' ')) > 1:
+            click.echo("ERROR: Can't specify more than one tag")
+            sys.exit(1)
+        kwargs['from_tag'] = kwargs['builds_or_tag']
+        del(kwargs['builds_or_tag'])
```
nphilipp: You could do these two lines in one:

```
        kwargs['from_tag'] = kwargs.pop('builds_or_tag')
```

bodhi/client/__init__.py
```diff
@@ -419,10 +422,22 @@ def new(user: str, password: str, url: str, debug: bool, openid_api: str, **kwar
...
     """
-
     client = bindings.BodhiClient(base_url=url, username=user, password=password,
                                   staging=kwargs['staging'], openid_api=openid_api)
 
+    # Because bodhi.server.services.updates expects from_tag to be string
+    # copy builds to from_tag and remove builds
+    if kwargs['from_tag']:
+        if len(kwargs['builds_or_tag'].split(' ')) > 1:
+            click.echo("ERROR: Can't specify more than one tag")
```
nphilipp: Please add `err=True` to the `click.echo()` call.

bodhi/client/__init__.py
```diff
@@ -527,6 +542,19 @@ def edit(user: str, password: str, url: str, debug: bool, openid_api: str, **kwa
...
+        # Because bodhi.server.services.updates expects from_tag to be string
+        # copy builds to from_tag and remove builds
+        if kwargs['from_tag']:
+            if not former_update.get('from_tag', None):
+                click.echo(
+                    "ERROR: This update is not created from tag."
+                    " Please remove --from_tag and try again")
+                sys.exit(1)
+            kwargs['from_tag'] = former_update['from_tag']
+            del(kwargs['builds'])
```
nphilipp: Also same, `kwargs.pop(...)` :smiley:

nphilipp (changes requested):
I'll take it that nobody objects to the flow I outlined in the diagram above, so let's implement it this way.

pypingou:
> I think we shouldn't do this implicitly, i.e. without the user specifying the --from-tag flag again, just as we don't automatically update builds when editing an update just because a newer NVR is available. 

I agree with this, though this has a consequence we need to be aware/pay attention to.

Imagine a bodhi update recording a build `foo-1.1` in side-tag `A`. Now if the packager builds `foo-1.2` in that same side-tag and doesn't update the bodhi update, we have a mismatch between what is in bodhi and what is in koji.
But when merging all the builds in koji will land in the base side-tag.
So to prevent this from happening, we need to double-check that the list of builds in the side-tag == the list of builds in the bodhi update before we merge that side-tag. If they don't, we raise an error to the user asking them to update the list of builds in the update.

nphilipp:
> Imagine a bodhi update recording a build `foo-1.1` in side-tag `A`. Now if the packager builds `foo-1.2` in that same side-tag and doesn't update the bodhi update, we have a mismatch between what is in bodhi and what is in koji.
> But when merging all the builds in koji will land in the base side-tag.
> So to prevent this from happening, we need to double-check that the list of builds in the side-tag == the list of builds in the bodhi update before we merge that side-tag. If they don't, we raise an error to the user asking them to update the list of builds in the update.

Yeah, I'll keep that in mind as I work on issue #2319 e.a.

Zlopez:
The last update should address latest requested changes and work as proposed by @nphilipp

bodhi/client/__init__.py
```diff
@@ -419,10 +422,20 @@ def new(user: str, password: str, url: str, debug: bool, openid_api: str, **kwar
...
+    # Because bodhi.server.services.updates expects from_tag to be string
+    # copy builds to from_tag and remove builds
+    if kwargs['from_tag']:
+        if len(kwargs['builds_or_tag'].split(' ')) > 1:
+            click.echo("ERROR: Can't specify more than one tag", err=True)
+            sys.exit(1)
+        kwargs['from_tag'] = kwargs.pop('builds_or_tag')
+    else:
+        kwargs['builds'] = kwargs.pop('builds_or_tag')
+        del(kwargs['from_tag'])
```
nphilipp: `del` is a keyword, so there's no need for parentheses: `del kwargs['from_tag']`

bodhi/client/__init__.py
```diff
@@ -517,15 +530,34 @@ def edit(user: str, password: str, url: str, debug: bool, openid_api: str, **kwa
...
-        del kwargs['addbuilds']
-        del kwargs['removebuilds']
+        # Because bodhi.server.services.updates expects from_tag to be string
+        # copy builds to from_tag and remove builds
+        if kwargs['from_tag']:
+            if not former_update.get('from_tag', None):
+                click.echo(
+                    "ERROR: This update is not created from tag."
+                    " Please remove --from_tag and try again", err=True
+                )
```
nphilipp, line 540: Grammar and missing punctuation could be fixed:

```
        click.echo(
            "ERROR: This update was not created from a tag. "
            "Please remove --from-tag and try again.", err=True
        )
```

bodhi/client/__init__.py
```diff
@@ -517,15 +530,34 @@ def edit(user: str, password: str, url: str, debug: bool, openid_api: str, **kwa
...
+                sys.exit(1)
+            if kwargs['addbuilds'] or kwargs['removebuilds']:
+                click.echo(
+                    "ERROR: Parameter --from-tag can't be called together with"
+                    "--addbuilds or --removebuilds parameters", err=True
+                )
```
nphilipp, line 546: Likewise:

```
        click.echo(
            "ERROR: The --from-tag option cannot be used together with "
            "--addbuilds or --removebuilds.", err=True
        )
```

bodhi/client/__init__.py
```diff
@@ -517,15 +530,34 @@ def edit(user: str, password: str, url: str, debug: bool, openid_api: str, **kwa
...
+                sys.exit(1)
+            kwargs['from_tag'] = former_update['from_tag']
+            del(kwargs['builds'])
```
nphilipp: Same as above (`del`).

bodhi/client/__init__.py
```diff
@@ -517,15 +530,34 @@ def edit(user: str, password: str, url: str, debug: bool, openid_api: str, **kwa
...
+            kwargs.pop('from_tag')
+            if kwargs['addbuilds']:
+                for build in kwargs['addbuilds'].split(','):
+                    if build not in kwargs['builds']:
+                        kwargs['builds'].append(build)
+            if kwargs['removebuilds']:
+                for build in kwargs['removebuilds'].split(','):
+                    kwargs['builds'].remove(build)
+        del(kwargs['addbuilds'])
+        del(kwargs['removebuilds'])
```
nphilipp: Ditto. :wink:

bodhi/tests/client/test___init__.py
```diff
@@ -757,6 +757,63 @@ def test_display_name_flag(self, send_request):
...
+                'updates/FEDORA-EPEL-2016-3081a94111/get-test-results',
+                verb='GET'
+            )
+        ]
+        self.assertEqual(send_request.mock_calls, calls)
+        self.assertEqual(bindings_client.base_url, 'http://localhost:6543/')
+
+    def test_from_tag_flag_multiple_tags(self):
+        """
+        Assert correct behavior with the --from-tag flag when multiple tags are specified.
```
nphilipp: "... ***if*** multiple tags..."

bodhi/client/__init__.py
```diff
@@ -419,10 +422,20 @@ def new(user: str, password: str, url: str, debug: bool, openid_api: str, **kwar
...
     """
-
     client = bindings.BodhiClient(base_url=url, username=user, password=password,
                                   staging=kwargs['staging'], openid_api=openid_api)
 
+    # Because bodhi.server.services.updates expects from_tag to be string
+    # copy builds to from_tag and remove builds
+    if kwargs['from_tag']:
+        if len(kwargs['builds_or_tag'].split(' ')) > 1:
+            click.echo("ERROR: Can't specify more than one tag", err=True)
```
nphilipp: Add punctuation? Like: "ERROR: Can't specify more than one tag."

bodhi/tests/client/test___init__.py
```diff
@@ -1838,6 +1895,133 @@ def test_addbuilds_removebuilds(self, send_request, query):
...
+        self.assertEqual(bindings_client.base_url, 'http://localhost:6543/')
+
+    @mock.patch('bodhi.client.bindings.BodhiClient.csrf',
+                mock.MagicMock(return_value='a_csrf_token'))
+    @mock.patch('bodhi.client.bindings.BodhiClient.query',
+                return_value=client_test_data.EXAMPLE_QUERY_MUNCH, autospec=True)
+    def test_from_tag_flag_no_tag(self, query):
+        """
+        Assert correct behavior with the --from-tag flag when update has none.
+        """
```
nphilipp, line 1953: "...***if*** the update wasn't created from a tag."

bodhi/tests/client/test___init__.py
```diff
@@ -1838,6 +1895,133 @@ def test_addbuilds_removebuilds(self, send_request, query):
...
+        bindings_client = query.mock_calls[0][1][0]
+        query.assert_called_with(
+            bindings_client, updateid='FEDORA-2017-c95b33872d')
+
+    @mock.patch('bodhi.client.bindings.BodhiClient.csrf',
+                mock.MagicMock(return_value='a_csrf_token'))
+    @mock.patch('bodhi.client.bindings.BodhiClient.query', autospec=True)
+    def test_from_tag_addbuilds(self, query):
+        """
+        Assert correct behavior with the --from-tag flag when --addbuilds is provided.
```
nphilipp: "...***if*** --addbuilds is ***used***."

bodhi/tests/client/test___init__.py
```diff
@@ -1838,6 +1895,133 @@ def test_addbuilds_removebuilds(self, send_request, query):
...
+        result = runner.invoke(
+            client.edit, ['FEDORA-2017-c95b33872d', '--user', 'bowlofeggs',
+                          '--password', 's3kr3t', '--from-tag',
+                          '--addbuilds', 'tar-1.29-4.fc25,nedit-5.7-1.fc25',
+                          '--notes', 'Updated package.',
+                          '--url', 'http://localhost:6543'])
+
+        self.assertEqual(result.exit_code, 1)
+        self.assertEqual(result.output, "ERROR: Parameter --from-tag can't be called together with"
+                                        "--addbuilds or --removebuilds parameters\n")
```
nphilipp: Adjust for the change above.

bodhi/tests/client/test___init__.py
```diff
@@ -1838,6 +1895,133 @@ def test_addbuilds_removebuilds(self, send_request, query):
...
+
+        result = runner.invoke(
+            client.edit, ['FEDORA-2017-c95b33872d', '--user', 'bowlofeggs',
+                          '--password', 's3kr3t', '--from-tag',
+                          '--notes', 'Updated package.',
+                          '--url', 'http://localhost:6543'])
+
+        self.assertEqual(result.exit_code, 1)
+        self.assertEqual(result.output, "ERROR: This update is not created from tag."
+                                        " Please remove --from_tag and try again\n")
```
nphilipp: Adjust for the change above.

bodhi/tests/client/test___init__.py
```diff
@@ -757,6 +757,63 @@ def test_display_name_flag(self, send_request):
...
+        runner = testing.CliRunner()
+
+        result = runner.invoke(
+            client.new,
+            ['--user', 'bowlofeggs', '--password', 's3kr3t', '--autokarma', 'fake tag',
+             '--bugs', '1234567', '--from-tag', '--url',
+             'http://localhost:6543', '--notes', 'No description.'])
+
+        self.assertEqual(result.exit_code, 1)
+        self.assertEqual(result.output, 'ERROR: Can\'t specify more than one tag\n')
```
nphilipp: Adjust for the change above.

bodhi/tests/client/test___init__.py
```diff
@@ -1838,6 +1895,133 @@ def test_addbuilds_removebuilds(self, send_request, query):
...
+        bindings_client = query.mock_calls[0][1][0]
+        query.assert_called_with(
+            bindings_client, updateid='FEDORA-2017-c95b33872d')
+
+    @mock.patch('bodhi.client.bindings.BodhiClient.csrf',
+                mock.MagicMock(return_value='a_csrf_token'))
+    @mock.patch('bodhi.client.bindings.BodhiClient.query', autospec=True)
+    def test_from_tag_removebuilds(self, query):
+        """
+        Assert correct behavior with the --from-tag flag when --addbuilds is provided.
```
nphilipp: "...***if --removebuilds is used***."

bodhi/tests/client/test___init__.py
```diff
@@ -1838,6 +1895,133 @@ def test_addbuilds_removebuilds(self, send_request, query):
...
+                          '--password', 's3kr3t', '--from-tag',
+                          '--removebuilds', 'nodejs-grunt-wrap-0.3.0-2.fc25',
+                          '--notes', 'Updated package.',
+                          '--url', 'http://localhost:6543'])
+
+        print(result.output)
+
+        self.assertEqual(result.exit_code, 1)
+        self.assertEqual(result.output, "ERROR: Parameter --from-tag can't be called together with"
+                                        "--addbuilds or --removebuilds parameters\n")
```
nphilipp: Adjust for the change above.

docs/user/man_pages/bodhi.rst
```diff
@@ -334,6 +339,12 @@ The ``updates`` command allows users to interact with bodhi updates.
 
         The name of the update
 
+    ``--from-tag``
+
+        If given, this flag will cause that ``<update>`` will be considered as created from Koji tag
+        and will use tag returned by query instead of builds.
```
nphilipp: Let's reword this: "If given, for updates that were created from a Koji tag, this will update the builds to the latest ones in the tag."

nphilipp (changes requested):
As discussed, only minor things:

- `del` is a keyword and doesn't need parentheses
- some grammar, punctuation, wording issues

nphilipp:
I took the liberty of fixing the couple text issues you missed, and shortening a couple of the docstrings where the summary line was too long.

nphilipp (approved):
Looks good to me, let's hope the CI pipeline gets finished over the weekend!

PR #3363, opened 2019-06-29
Add timeout flag to pytest

### LINKED ISSUE #3339: Get tests to fail if they take too long
We've had a number of times that tests started taking a very long time. Sometimes this happens when a pull request is merged, because we don't always notice during review that the tests took a long time to run.

We could look for testing plugins that can cause the tests to fail if they take too long. For example, this one looks possibly useful: https://pypi.org/project/pytest-timeout/ It would be nice if we could specify something like "no test should take longer than 1 second to run", but that plugin doesn't seem to have a way to express that unless the decorator can be used on the base test class (it doesn't document that you can do that). I'd prefer not to have to add that decorator to each individual test. It might be OK to just set a time limit on how long the entire suite runs if we can't find a better plugin than this one.

Having this would help prevent problems like https://github.com/fedora-infra/bodhi/issues/3338

sebwoj:
fix #3339


sebwoj:
At first, I inserted "--timeout=60" to this [config file](https://github.com/fedora-infra/bodhi/blob/develop/setup.cfg), but I have no idea how to install pytest-timeoout in CI enviroment. This is why this cli option is in every `pytest` invocation, except for CI.

abompard (changes requested):
To add pytest-timeout to CI, you need to add it to the installed packages in `devel/ci/cico.pipeline` (look for pytest there).
Also, I think it would be better to add the CLI option to `setup.cfg`, where options are already added to the command line.

sebwoj:
Yeah, this is the place I was looking for.
Thx @abompard

Zlopez (approved):

ryanlerch:
@sebwoj just one conflict on this one.

@abompard is your change request satisfied now too?

abompard (approved):

sebwoj:
@abompard It looks like integration tests are still missing pytest-timeout.

In the PR I wrote the following command:
`'python3.6 -m pip install \'pytest<4.0\' pytest-cov conu munch fedora-messaging pytest-timeout'`
but in CI(jenkins [console](https://ci.centos.org/job/bodhi-pipeline/job/PR-3363/9/execution/node/33/log/) ) the following command is performed:
`'python3.6 -m pip install '\''pytest<4.0'\'' pytest-cov conu munch fedora-messaging'`
Do you know how to fix this?

cverna:
Closing this due to inactivity

PR #3370, opened 2019-07-02
Correct test_composes_info in CI

sebwoj:
Signed-off-by: Sebastian Wojciechowski <s.wojciechowski89@gmail.com>

PR #3371, opened 2019-07-03
Unbreak various tests in CI

Zlopez:
This should fix the failing CI tests

### DIFF
--- a/devel/ci/integration/bodhi/Dockerfile-rawhide
+++ b/devel/ci/integration/bodhi/Dockerfile-rawhide
@@ -7,6 +7,9 @@ LABEL \
 
 # For integration testing we're using the infrastructure repo
 RUN curl -o /etc/yum.repos.d/infra-tags.repo https://infrastructure.fedoraproject.org/cgit/ansible.git/plain/files/common/fedora-infra-tags.repo
+# work around https://bugzilla.redhat.com/show_bug.cgi?id=1699396
+RUN rpm --import "$(awk -F= '{if ($0 ~ "gpgkey *=") {print $2;}}' < /etc/yum.repos.d/infra-tags.repo)"
+
 RUN dnf upgrade -y
 
 # Install Bodhi deps (that were not needed by the unittests container)
--- a/devel/ci/integration/bodhi/Dockerfile-rawhide
+++ b/devel/ci/integration/bodhi/Dockerfile-rawhide
@@ -8,7 +8,10 @@ LABEL \
 # For integration testing we're using the infrastructure repo
 RUN curl -o /etc/yum.repos.d/infra-tags.repo https://infrastructure.fedoraproject.org/cgit/ansible.git/plain/files/common/fedora-infra-tags.repo
 # work around https://bugzilla.redhat.com/show_bug.cgi?id=1699396
-RUN rpm --import "$(awk -F= '{if ($0 ~ "gpgkey *=") {print $2;}}' < /etc/yum.repos.d/infra-tags.repo)"
+# and deal with https://fedoraproject.org/wiki/Changes/Set_skip_if_unavailable_default_to_false
+RUN rpm --import "$(awk -F= '{if ($0 ~ "gpgkey *=") {print $2;}}' \
+        < /etc/yum.repos.d/infra-tags.repo)" && \
+    echo skip_if_unavailable=True >> /etc/yum.repos.d/infra-tags.repo
 
 RUN dnf upgrade -y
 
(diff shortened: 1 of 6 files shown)


nphilipp:
I've updated the PR:

- Added a description to the commit log about what went wrong here, i.e. backwards-incompatible changes in pytest 5.0 triggered the test failures. Because 5.0 dropped support for Python 2.x, we don't have it in Fedora yet, so the problem only showed up in some setups (Python modules installed with pip and up-to-date).
- Renamed the context manager object from `e` to `excinfo` to make it more explicit we're not dealing with an exception here.
- Rather than `str(e.value), used the `ExceptionInfo.exconly()` method which exposes name and message of the wrapped exception (and do this consistently in another test with the second commit).

pypingou:
Ok, I'm confused the test results show as red here but visiting: https://ci.centos.org/job/bodhi-pipeline/job/PR-3371/3/ I see ``Test Result (no failures)``

is it still running?

nphilipp:
> Ok, I'm confused the test results show as red here but visiting: https://ci.centos.org/job/bodhi-pipeline/job/PR-3371/3/ I see `Test Result (no failures)`
> 
> is it still running?

Right now it still shows 6 as "expected". While the tests are in flight, I always look into the [Pipeline Steps](https://ci.centos.org/job/bodhi-pipeline/job/PR-3371/3/flowGraphTable) and, of the ones marked red, check their log (terminal symbol beside the red bubble). Basically, we have one issue in the pip-integration/-build tests with an unsigned package, and the sphinx problem in rawhide-docs.

pypingou (approved):
nice work tracking this down! :)

nphilipp:
With #3373 in flight I have had some time to think about `str(excinfo.value)` vs. `excinfo.exconly()`, especially that the latter isn't the same as the former but also includes the exception name. @Zlopez do you have an opinion?

Zlopez:
@nphilipp Will `excinfo.exconly()` work with current tests?

pypingou:
> Will excinfo.exconly() work with current tests?

Yes, we checked that, it was added in 2007 (iirc)

nphilipp:
As discussed, let's broaden the scope of this PR and include:

* Work around Fedora Infra RPM repo key import issue (#3373)
* Add `skip_if_unavailable` to the infra-tags repo on Rawhide (due to https://fedoraproject.org/wiki/Changes/Set_skip_if_unavailable_default_to_false from F31 on, this only triggered now)
* Clamp `pydocstyle < 4.0` because https://github.com/PyCQA/pydocstyle/issues/370 trips over the nested function in bodhi/server/renderers.py

nphilipp:
Oh, and as discussed I've replaced `excinfo.exconly()` with `str(excinfo.value)` as it's closer to what was done originally.

devel/ci/integration/bodhi/Dockerfile-rawhide
```diff
@@ -7,6 +7,12 @@ LABEL \
 
 # For integration testing we're using the infrastructure repo
 RUN curl -o /etc/yum.repos.d/infra-tags.repo https://infrastructure.fedoraproject.org/cgit/ansible.git/plain/files/common/fedora-infra-tags.repo
+# work around https://bugzilla.redhat.com/show_bug.cgi?id=1699396
+RUN rpm --import "$(awk -F= '{if ($0 ~ "gpgkey *=") {print $2;}}' < /etc/yum.repos.d/infra-tags.repo)"
+# deal with https://fedoraproject.org/wiki/Changes/Set_skip_if_unavailable_default_to_false
+RUN dnf install -y dnf-plugins-core
+RUN dnf config-manager infrastructure-tags --setopt skip_if_unavailable=True --save
```
pypingou: Every command adds a layer in the container, do you know if Randy and Aurélien preferred the one command / line approach? Otherwise, it is often recommended, to have multiple command per "RUN" using ``\`` to delimit them.

nphilipp: I don't know their preference. In the Dockerfiles there are instances of one RUN line containing multiple commands (though not many), but also multiple consecutive RUN lines each containing a single command.

nphilipp:
For reference, here's an example of `pydocstyle` tripping: https://ci.centos.org/job/bodhi-pipeline/job/PR-3371/6/execution/node/329/log/

pypingou (approved):

Zlopez:
So only thing failing now will be rawhide docs test? Or this is a new issue?

nphilipp:
> So only thing failing now will be rawhide docs test? Or this is a new issue?

No, it's that `cornice_sphinx` is incompatible with the latest `sphinx` version: https://github.com/Cornices/cornice.ext.sphinx/issues/22

Zlopez:
I see that there are plenty of integration tests failing, but this will be probably related to https://github.com/fedora-infra/bodhi/pull/3370

nphilipp:
As discussed: considering that we verified the changes in this PR to fix what they're supposed to do and that they aren't causing the remaining and new CI pipeline failures, I'll attempt to merge them manually, which should make fixing the outstanding issues easier as you don't have to dig through the errors this one is fixing.

PR #3372, opened 2019-07-04
Move the logic to models.py and migrate the script to argparse

pypingou:
There are two commits in this PR, the first one move the logic for moving an update from pending to stable directly, out of the approve_testing script and into ``bodhi/server/models.py`` as a method of the ``Update`` objects.
This will allow to re-use this heuristic in other places of bodhi without duplicating the code.

The second commit move the approve_testing script to argparse to make it easier to support a ``--debug`` option allowing to change the logging level from ``ERROR`` to ``INFO``, thus giving the user running the script more info about what is going on.

### DIFF
--- a/bodhi/server/models.py
+++ b/bodhi/server/models.py
@@ -3653,6 +3653,41 @@ class Update(Base):
                 author="bodhi",
             )
 
+    def move_to_stable(self, db):
+        """
+        Move the update to stable if it was set to not spend any time in testing.
+
+        Args:
+            db (sqlalchemy.orm.session.Session): A database session.
+        Raises:
+            sqlalchemy.exc.SQLAlchemyError: If the database doesn't behave as expected.
+        """
+        if not self.release.mandatory_days_in_testing and not self.autotime:
+            # If this release does not have any testing requirements and is not autotime,
+            # skip it
+            log.debug(f"{self.release.name} doesn't have mandatory days in testing")
+            return
+
+        # If this update was already commented, skip it
+        if self.has_stable_comment:
+            return
+
+        # If updates have reached the testing threshold, say something! Keep in mind
+        # that we don't care about karma here, because autokarma updates get their request set
+        # to stable by the Update.comment() workflow when they hit the required threshold. Thus,
+        # this function only needs to consider the time requirements because these updates have
+        # not reached the karma threshold.
+        if self.meets_testing_requirements:
+            log.info(f'{self.alias} now meets testing requirements')
+            self.comment(db, str(config.get('testing_approval_msg')), author='bodhi')
+
+            notifications.publish(update_schemas.UpdateRequirementsMetStableV1.from_dict(
+                dict(update=self)))
+
+            if self.autotime and self.days_in_testing >= self.stable_days:
+                log.info(f"Automatically marking {self.alias} as stable")
+                self.set_request(db=db, action=UpdateRequest.stable, username="bodhi")
+
 
 event.listen(
     Update.test_gating_status,
--- a/bodhi/server/scripts/approve_testing.py
+++ b/bodhi/server/scripts/approve_testing.py
@@ -22,7 +22,7 @@ The script is responsible for commenting on updates after they reach the mandato
 spent in the testing repository.
 """
 
-import os
+import argparse
 import sys
 import logging
 
@@ -36,20 +36,7 @@ from ..config import config
 logger = logging.getLogger('approve-testing')
 
 
-def usage(argv):
-    """
-    Print usage information and exit with code 1.
-
-    Args:
-        argv (list): The arguments that were passed to the CLI from sys.argv.
-    """
-    cmd = os.path.basename(argv[0])
-    print('usage: %s <config_uri>\n'
-          '(example: "%s development.ini")' % (cmd, cmd))
-    sys.exit(1)
-
-
-def main(argv=sys.argv):
+def main(argv=None):
     """
     Comment on updates that are eligible to be pushed to stable.
 
@@ -59,16 +46,23 @@ def main(argv=sys.argv):
     pushed to stable.
 
     This function is the entry point for the bodhi-approve-testing console script.
-
-    Args:
-        argv (list): A list of command line arguments. Defaults to sys.argv.
     """
-    logging.basicConfig(level=logging.ERROR)
+    parser = argparse.ArgumentParser(
+        description='Comment on updates that are eligible to be pushed to stable.'
+    )
+    parser.add_argument('config_uri', help='Path to the configuration file to use')
+    parser.add_argument(
+        '--verbose', dest='verbose', action='store_true', default=False,
+        help='Increase the verbosity level of the output')
 
-    if len(argv) != 2:
-        usage(argv)
+    args = parser.parse_args(argv)
 
-    settings = get_appsettings(argv[1])
+    if args.verbose:
+        logging.basicConfig(level=logging.INFO)
+    else:
+        logging.basicConfig(level=logging.ERROR)
+
+    settings = get_appsettings(args.config_uri)
     initialize_db(settings)
     db = Session()
     buildsys.setup_buildsystem(config)
@@ -85,3 +79,7 @@ def main(argv=sys.argv):
         db.rollback()
         Session.remove()
         sys.exit(1)
+
+
+if __name__ == '__main__':
+    main()
--- a/bodhi/server/scripts/approve_testing.py
+++ b/bodhi/server/scripts/approve_testing.py
@@ -28,9 +28,8 @@ import logging
 
 from pyramid.paster import get_appsettings
 
-from bodhi.messages.schemas import update as update_schemas
-from bodhi.server import Session, initialize_db, notifications, buildsys
-from ..models import Update, UpdateStatus, UpdateRequest
+from bodhi.server import Session, initialize_db, buildsys
+from ..models import Update, UpdateStatus
 from ..config import config
 
 
@@ -78,33 +77,8 @@ def main(argv=sys.argv):
         testing = db.query(Update).filter_by(status=UpdateStatus.testing,
                                              request=None)
         for update in testing:
-            if not update.release.mandatory_days_in_testing and not update.autotime:
-                # If this release does not have any testing requirements and is not autotime,
-                # skip it
-                print(f"{update.release.name} doesn't have mandatory days in testing")
-                continue
-
-            # If this update was already commented, skip it
-            if update.has_stable_comment:
-                continue
-
-            # If updates have reached the testing threshold, say something! Keep in mind
-            # that we don't care about karma here, because autokarma updates get their request set
-            # to stable by the Update.comment() workflow when they hit the required threshold. Thus,
-            # this function only needs to consider the time requirements because these updates have
-            # not reached the karma threshold.
-            if update.meets_testing_requirements:
-                print(f'{update.alias} now meets testing requirements')
-                update.comment(db, str(config.get('testing_approval_msg')), author='bodhi')
-
-                notifications.publish(update_schemas.UpdateRequirementsMetStableV1.from_dict(
-                    dict(update=update)))
-
-                if update.autotime and update.days_in_testing >= update.stable_days:
-                    print(f"Automatically marking {update.alias} as stable")
-                    update.set_request(db=db, action=UpdateRequest.stable, username="bodhi")
-
-                db.commit()
+            update.move_to_stable(db)
+            db.commit()
 
     except Exception as e:
         print(str(e))
--- a/bodhi/tests/server/scripts/test_approve_testing.py
+++ b/bodhi/tests/server/scripts/test_approve_testing.py
@@ -21,6 +21,7 @@ This module contains tests for the bodhi.server.scripts.approve_testing module.
 from datetime import datetime, timedelta
 from io import StringIO
 import logging
+import sys
 from unittest.mock import call, patch
 
 from fedora_messaging import api, testing as fml_testing
@@ -131,8 +132,9 @@ class TestMain(BaseTestCase):
     @patch('bodhi.server.models.Update.comment', side_effect=IOError('The DB died lol'))
     @patch('bodhi.server.scripts.approve_testing.Session.remove')
     @patch('bodhi.server.scripts.approve_testing.sys.exit')
+    @patch('bodhi.server.models.log')
     @patch('sys.stdout', new_callable=StringIO)
-    def test_exception_handler(self, stdout, exit, remove, comment):
+    def test_exception_handler(self, stdout, log, exit, remove, comment):
         """The Exception handler prints the Exception, rolls back and closes the db, and exits."""
         update = self.db.query(models.Update).all()[0]
         update.date_testing = datetime.utcnow() - timedelta(days=15)
@@ -154,15 +156,19 @@ class TestMain(BaseTestCase):
             self.db,
             ('This update can be pushed to stable now if the maintainer wishes'),
             author='bodhi')
-        self.assertEqual(stdout.getvalue(),
-                         f'{update.alias} now meets testing requirements\nThe DB died lol\n')
+
+        self.assertEqual(
+            [call.info(f'{update.alias} now meets testing requirements')],
+            log.mock_calls)
+        self.assertEqual(stdout.getvalue(), 'The DB died lol\n')
         remove.assert_called_once_with()
 
     @patch('bodhi.server.models.Update.comment', side_effect=[None, IOError('The DB died lol')])
     @patch('bodhi.server.scripts.approve_testing.Session.remove')
     @patch('bodhi.server.scripts.approve_testing.sys.exit')
+    @patch('bodhi.server.models.log')
     @patch('sys.stdout', new_callable=StringIO)
-    def test_exception_handler_on_the_second_update(self, stdout, exit, remove, comment):
+    def test_exception_handler_on_the_second_update(self, stdout, log, exit, remove, comment):
         """
         Ensure, that when the Exception is raised, all previous transactions are commited,
         the Exception handler prints the Exception, rolls back and closes the db, and exits.
@@ -195,9 +201,14 @@ class TestMain(BaseTestCase):
             ('This update can be pushed to stable now if the maintainer wishes'),
             author='bodhi',)
         self.assertEqual(comment.call_args_list, [comment_expected_call, comment_expected_call])
-        self.assertEqual(stdout.getvalue(),
-                         (f'{update2.alias} now meets testing requirements\n'
-                          f'{update.alias} now meets testing requirements\nThe DB died lol\n'))
+        self.assertEqual(
+            [
+                call.debug(f'Set alias for bodhi2-2.0-1.fc17 to {update2.alias}'),
+                call.info(f'{update2.alias} now meets testing requirements'),
+                call.info(f'{update.alias} now meets testing requirements'),
+            ],
+            log.mock_calls)
+        self.assertEqual(stdout.getvalue(), 'The DB died lol\n')
         remove.assert_called_once_with()
 
     def test_non_autokarma_critpath_update_meeting_karma_requirements_gets_one_comment(self):
@@ -477,6 +488,8 @@ class TestMain(BaseTestCase):
         with patch('bodhi.server.scripts.approve_testing.initialize_db'):
             with patch('bodhi.server.scripts.approve_testing.get_appsettings', return_value=''):
                 approve_testing.main(['nosetests', 'some_config.ini', 'testnoses'])
+
+        cmd = sys.argv[0].rsplit('/')[-1]
 
         self.assertEqual(
             stdout.getvalue(),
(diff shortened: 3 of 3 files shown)


bodhi/server/models.py
```diff
@@ -3642,6 +3642,43 @@ def comment_on_test_gating_status_change(target, value, old, initiator):
                 author="bodhi",
             )
 
+    def move_to_stable(self, db):
+        """
+        Move the update to stable if it was set to not spend any time in testing.
+
+        Args:
+            db (sqlalchemy.orm.session.Session): A database session.
```
Zlopez, line 3661: Do you need the session to be passed as argument?

Zlopez (commented):
Only one comment, otherwise looks good

bodhi/server/models.py
nphilipp: It's how methods working on a database session are usually implemented in Bodhi.

bodhi/server/models.py
```diff
@@ -3642,6 +3642,43 @@ def comment_on_test_gating_status_change(target, value, old, initiator):
...
+        Raises:
+            sqlalchemy.exc.SQLAlchemyError: If the database doesn't behave as expected.
```
nphilipp, line 3663: Rather than mentioning the types in the docstring, would you add type hints? Our Python baseline is 3.6 and (AIUI) we want new code to have type hints rather than the same info in docstrings (but haven't gotten around to documenting this, ouch).

bodhi/server/models.py
```diff
@@ -3642,6 +3642,43 @@ def comment_on_test_gating_status_change(target, value, old, initiator):
...
+            self.comment(db, str(config.get('testing_approval_msg')), author='bodhi')
+
+            notifications.publish(update_schemas.UpdateRequirementsMetStableV1.from_dict(
+                dict(update=self)))
+
+            if self.autotime and self.days_in_testing >= self.stable_days:
+                log.info(f"Automatically marking {self.alias} as stable")
+                self.set_request(db=db, action=UpdateRequest.stable, username="bodhi")
+
+            db.commit()
```
nphilipp: I don't think the session should be committed here, rather than where the method is used.

bodhi/tests/server/scripts/test_approve_testing.py
```diff
@@ -154,15 +155,19 @@ def test_exception_handler(self, stdout, exit, remove, comment):
             self.db,
             ('This update can be pushed to stable now if the maintainer wishes'),
             author='bodhi')
-        self.assertEqual(stdout.getvalue(),
-                         f'{update.alias} now meets testing requirements\nThe DB died lol\n')
+
+        self.assertEqual(
+            [call.debug(f'{update.alias} now meets testing requirements')],
+            log.mock_calls)
```
nphilipp, line 202: To cope with debugging output added elsewhere in `models`, I'd prefer something like this:

```
self.assertIn(
    call.info(f'{update.alias} now meets testing requirements'),
    log.mock_calls)
```

bodhi/tests/server/scripts/test_approve_testing.py
```diff
@@ -195,9 +200,14 @@ def test_exception_handler_on_the_second_update(self, stdout, exit, remove, comm
...
-        self.assertEqual(stdout.getvalue(),
-                         (f'{update2.alias} now meets testing requirements\n'
-                          f'{update.alias} now meets testing requirements\nThe DB died lol\n'))
+        self.assertEqual(
+            [
+                call.debug(f'Set alias for bodhi2-2.0-1.fc17 to {update2.alias}'),
+                call.debug(f'{update2.alias} now meets testing requirements'),
+                call.debug(f'{update.alias} now meets testing requirements'),
+            ],
+            log.mock_calls)
```
nphilipp: Similar here:

```
for msg_call in (call.debug(f'Set alias for bodhi2-2.0-1.fc17 to {update2.alias}'),
            call.info(f'{update2.alias} now meets testing requirements'),
            call.info(f'{update.alias} now meets testing requirements')):
    self.assertIn(msg_call, log.mock_calls)
```

bodhi/tests/server/scripts/test_approve_testing.py
```diff
@@ -469,19 +523,20 @@ def test_subsequent_comments_after_initial_push_comment(self):
         self.assertEqual(cmnts[2].text, config.get('testing_approval_msg'))
 
     @patch('sys.exit')
-    @patch('sys.stdout', new_callable=StringIO)
-    def test_usage(self, stdout, exit):
+    @patch('sys.stderr', new_callable=StringIO)
+    def test_usage(self, stderr, exit):
         """
         Assert that the usage message is returned to the user if not exactly 2 arguments are given
```
nphilipp: "... 1 argument ..."?

bodhi/tests/server/scripts/test_approve_testing.py
```diff
@@ -154,15 +155,63 @@ def test_exception_handler(self, stdout, exit, remove, comment):
...
             author='bodhi')
-        self.assertEqual(stdout.getvalue(),
-                         f'{update.alias} now meets testing requirements\nThe DB died lol\n')
+
+        self.assertEqual(
+            [call.info(f'{update.alias} now meets testing requirements')],
+            log.mock_calls)
+        self.assertEqual(stdout.getvalue(), 'The DB died lol\n')
+        # Check the logging level when debug is False
+        self.assertEqual(log.getEffectiveLevel(), 20)
```
nphilipp, line 165: This fails in local tests, it returns 30, i.e. `log.WARNING`. Maybe some other test messes around with logging levels?

I'm not sure what the best way to fix this would be:

* Either intercept/mock out the methods of `models.log` that can set the log level so it isn't actually globally changed in tests.
* Or, instead of using `mock.patch` to get hold of the log object, use the `caplog` fixture of pytest. See  https://docs.pytest.org/en/latest/logging.html#caplog-fixture
* Something else?

bodhi/tests/server/scripts/test_approve_testing.py
```diff
@@ -469,19 +523,20 @@ def test_subsequent_comments_after_initial_push_comment(self):
...
-                approve_testing.main(['nosetests', 'some_config.ini', 'testnoses'])
+                approve_testing.main(['some_config.ini', 'testnoses'])
 
         self.assertEqual(
-            stdout.getvalue(),
-            'usage: nosetests <config_uri>\n(example: "nosetests development.ini")\n')
-        exit.assert_called_once_with(1)
+            stderr.getvalue(),
+            'usage: py.test-3 [-h] [--debug] config_uri\n'
+            'py.test-3: error: unrecognized arguments: testnoses\n')
```
nphilipp: This fails in local tests because I call pytest as `pytest` instead of `py.test-3`. Either take the actual executable name into account here, or perhaps use a regex match?

Zlopez (approved):

bodhi/server/scripts/approve_testing.py
```diff
@@ -60,16 +46,23 @@ def main(argv=sys.argv):
...
+    )
+    parser.add_argument('config_uri', help='Path to the configuration file to use')
+    parser.add_argument(
+        '--debug', dest='debug', action='store_true', default=False,
+        help='Increase the verbosity level of the output')
+
+    args = parser.parse_args(argv)
+
+    if args.debug:
+        logging.basicConfig(level=logging.INFO)
```
nphilipp, line 61: This feels odd, that `--debug` sets the log level to `INFO`. Should this be a `--verbose` flag instead?

nphilipp (changes requested):
I spotted some changes in the second commit fixing stuff from the first (e.g. `call.debug()` vs. `call.info()`), not sure how to flag this appropriately here. :)

bodhi/server/scripts/approve_testing.py
pypingou: wfm

bodhi/tests/server/scripts/test_approve_testing.py
pypingou: Already have the fix for this :)

bodhi/tests/server/scripts/test_approve_testing.py
pypingou: Good catch

bodhi/tests/server/scripts/test_approve_testing.py
pypingou: I have the same problem locally so I suspect something is messing up w/ logs somewhere :(

bodhi/server/models.py
pypingou: If we don't commit, there is no need to pass the db object so that makes this comment moot :)

bodhi/server/models.py
pypingou: Simplifies things :)

bodhi/server/models.py
pypingou: Actually, the db object is needed in a few places, so nm my last comment (just the Raises is to be removed)

bodhi/tests/server/scripts/test_approve_testing.py
pypingou: We would have to have this debug info be added in the code path used in this part of the code, so do you still want this?

bodhi/server/models.py
nphilipp: > Actually, the db object is needed in a few places, so nm my last comment (just the Raises is to be removed)

Do you know if implicit flushes could still raise `SQLAlchemyError`?

bodhi/server/models.py
nphilipp: Heh. It's just that we don't ever commit to the DB elsewhere in the model classes. :)

bodhi/server/models.py
Zlopez: Yes, they can. I was doing something like this in Anitya tests.

pypingou: ``.flush()`` will raise any exception that ``.commit()`` would

bodhi/tests/server/scripts/test_approve_testing.py
nphilipp: We're calling `self.comment()` and `self.set_request()` from within `Update.move_to_stable()` so I'd not totally rule out that some time down the road, someone adds debugging in some unrelated place which might break on this -- then -- unrelated test.

bodhi/tests/server/scripts/test_approve_testing.py
nphilipp: 👍

bodhi/server/scripts/approve_testing.py
nphilipp: Heh, if anybody is ever confused about this, I'll send them your way. :stuck_out_tongue:

bodhi/tests/server/scripts/test_approve_testing.py
pypingou: okido :)

bodhi/tests/server/scripts/test_approve_testing.py
nphilipp: Shall I give option 2 (pytest caplog fixture) a try?

bodhi/server/scripts/approve_testing.py
pypingou: I've adjusted it! :D

bodhi/server/models.py
nphilipp: So I'd leave the `Raises: ...` block in because implicit flushes are a thing.

bodhi/tests/server/scripts/test_approve_testing.py
pypingou: Feel free to give it a try, but from my quick testing I couldn't make it work, if you've played with it before you may have better luck though :)

bodhi/server/models.py
pypingou: okido

bodhi/server/scripts/approve_testing.py
```diff
@@ -78,36 +71,15 @@ def main(argv=sys.argv):
...
+            db.commit()
 
     except Exception as e:
         print(str(e))
         db.rollback()
         Session.remove()
         sys.exit(1)
+
+
+if __name__ == '__main__':
```
cverna, line 84: I don't think we need this, this script is installed by setup.py and points to its main function (https://github.com/fedora-infra/bodhi/blob/develop/setup.py#L232)

bodhi/server/models.py
```diff
@@ -3653,6 +3653,41 @@ def comment_on_test_gating_status_change(target, value, old, initiator):
                 author="bodhi",
             )
 
+    def move_to_stable(self, db):
```
cverna, line 3656: This is misleading since it will not actually move the update to stable, but request the update to be move to stable by the next compose. 
The effect is that bodhi will tag the koji build with the release `Pending stable tag`. For the rawhide use case we still need to find out how the update status will be updated to stable after the nightly compose.

Maybe we should keep the name as `approve_testing` or something in that spirit.

bodhi/server/models.py
```diff
@@ -3653,6 +3653,41 @@ def comment_on_test_gating_status_change(target, value, old, initiator):
...
+        """
+        Move the update to stable if it was set to not spend any time in testing.
```
cverna, line 3658: See above comment :-), also the code does more than that in some case it will not request the update to be pushed to stable.

cverna (commented):
Added a few comments, overall I am still not convince we need this change at this point of time. What I would do is change the frequency of the approve testing cron job to have it run every 10 min for example and let people play with that in staging.
Based on the feedbacks we should consider improving the experience or not :-).

Just by curiosity what is the need to have the build in the rawhide buildroot available immediately ? Is that for building dependent packages ? if that's the case would that not be covered by the multi build update ?

bodhi/server/scripts/approve_testing.py
pypingou: It is present in ``skopeo_lite.py``, ``check_policies.py`` and ``sar.py`` but I'm fine with removing it, I just find it  convenient.

pypingou:
> Just by curiosity what is the need to have the build in the rawhide buildroot available immediately ? Is that for building dependent packages ? if that's the case would that not be covered by the multi build update ?

One of the requirement is to not impact the current workflow in rawhide for package that do not opt-in. Waiting 10 minutes, it impacting the current workflow. People expect their package to land in rawhide in a few minutes after they have built it.

bodhi/server/models.py
pypingou: This makes me think that this is not working or going to work as we want.

If you check: ``koji -s https://koji.stg.fedoraproject.org/kojihub buildinfo python-arrow-0.12.1-10.fc31`` you will see that this build is still in ``f31-updates-testing`` while the build is 8 days old.

If the move to stable only happens at the compose time, since we are not doing compose in bodhi for rawhide, it means updates will be stuck in ``f31-updates-testing``, potentially, in a pending state in the UI, but nothing moves the update further :(

cverna: yes also the f31 release in staging is not configure properly, the `Pending stable tag` should be `f31-pending` and not `f31-updates-testing` so that the builds get signed by robosignatory and tagged with the `f31` tag.

I guess we need a consumer that listen to the end of compose messages, and set the updates status to `STABLE` if the compose was successful.

cverna:
> > Just by curiosity what is the need to have the build in the rawhide buildroot available immediately ? Is that for building dependent packages ? if that's the case would that not be covered by the multi build update ?
> 
> One of the requirement is to not impact the current workflow in rawhide for package that do not opt-in. Waiting 10 minutes, it impacting the current workflow. People expect their package to land in rawhide in a few minutes after they have built it.

In my opinion this is the perfect example where getting early feedback from users is useful, we have a good enough solution which makes a slight impact to the current workflow instead of assuming that this is wrong we should ask users if they can live with that or not. If not we can improve the solution, if they can live with it then happy days we saved ourselves some work :-)

bodhi/server/scripts/approve_testing.py
cverna: Yeah, I am just trying to be conservative and make sure we don't break things that nobody was aware of :stuck_out_tongue:

pypingou:
> In my opinion this is the perfect example where getting early feedback from users is useful, we have a good enough solution which makes a slight impact to the current workflow instead of assuming that this is wrong we should ask users if they can live with that or not. If not we can improve the solution, if they can live with it then happy days we saved ourselves some work :-)

I agree but the only place where we can have early feedback is in staging. Once we turn this on in production, everyone will be impacted by this change.
And in staging things are still not working and asking for feedback there means asking people to do work in two places, one of which is thrown away (might still be worth doing, the first point in this sentence is what worries me most).

bodhi/server/models.py
pypingou: This means we need to figure out exactly which update went into every compose.
To avoid this work which is tedious and error-prone, we agreed that we would mark the updates as STABLE once they are available in the rawhide buildroot, just as it is today (once the build is in the buildroot, there is nothing else to consider for rawhide).

So the flow is:
- build in koji
- automatically create an update in bodhi in a PENDING state
- query greenwave to know if that update should be gated or not (did they opt-in?) if not, skip the next step
- wait for greenwave to announce its decision
- inform the user about greenwave's decision by a comment on the udpate
- if greenwave lets the update go through, push the update to the ``-pending`` tag and mark the update as STABLE.

cverna: > This means we need to figure out exactly which update went into every compose.
> To avoid this work which is tedious and error-prone, we agreed that we would mark the updates as STABLE once they are available in the rawhide buildroot, just as it is today (once the build is in the buildroot, there is nothing else to consider for rawhide).
> 
> So the flow is:
> 
>     * build in koji
> 
>     * automatically create an update in bodhi in a PENDING state
> 
>     * query greenwave to know if that update should be gated or not (did they opt-in?) if not, skip the next step
> 
>     * wait for greenwave to announce its decision
> 
>     * inform the user about greenwave's decision by a comment on the udpate
> 
>     * if greenwave lets the update go through, push the update to the `-pending` tag and mark the update as STABLE.

Ha ok yes that makes things simpler :-). So we just need something like that in the approve_testing script : 

```
if update.release.mandatory_days_in_testing == 0:
    update.status =  UpdateStatus.stable

```
after this line https://github.com/fedora-infra/bodhi/blob/develop/bodhi/server/scripts/approve_testing.py#L105.

We have to keep in mind that the build might not be in stable tho depending on the signing status. Also I think users are used to be have the build available in the repo when it is marked as stable in bodhi so that might get them confused. 

Maybe we should approach the packagers that have originally commented on the Change proposal and get there inputs on these decisions.

nphilipp:
> Just by curiosity what is the need to have the build in the rawhide buildroot available immediately ? Is that for building dependent packages ? if that's the case would that not be covered by the multi build update ?

AIUI, the multi build update involves using a side tag, so this is, at the very least, a change of process for pkg maintainers, if not more manual work. This is anecdotal of course, but when I have my packager hat on the sequence of 1) build something in Rawhide 2) koji wait-repo 3) build something else that depends on it (lather, rinse, repeat) is a very common one. Everything that makes it longer or more tedious won't be appreciated. :wink:

> Maybe we should approach the packagers that have originally commented on the Change proposal and get there inputs on these decisions.

Is this a thread on devel list? If yes, from when (circa)?

pypingou:
> Is this a thread on devel list? If yes, from when (circa)?

It was done a bit in emails after the original thread on the rawhide package gating change proposal and followed up by an IRC meeting where we gathered some more inputs leading to the refinement of the change proposal and it being merged back into one proposal instead of two.

pypingou:
> > Just by curiosity what is the need to have the build in the rawhide buildroot available immediately ? Is that for building dependent packages ? if that's the case would that not be covered by the multi build update ?
> 
> AIUI, the multi build update involves using a side tag, so this is, at the very least, a change of process for pkg maintainers, if not more manual work. This is anecdotal of course, but when I have my packager hat on the sequence of 1) build something in Rawhide 2) koji wait-repo 3) build something else that depends on it (lather, rinse, repeat) is a very common one. Everything that makes it longer or more tedious won't be appreciated. wink

Agreed, especially since this is going to also impact the person not opting-in in gating. At this point, however, I'm ok with running the cron job every 5 minutes or 2 minutes even to see if we can get the flow to work in staging and hopefully we can then improve this in time for production.

bodhi/server/models.py
bowlofeggs: We should not pass the db session in new Bodhi code. I spent the past three years trying to get rid of that pattern. You can get the db session by instantiating ```bodhi.server.models.Session```.

bodhi/server/scripts/approve_testing.py
bowlofeggs: Yeah let's not put this code - it's not easy to test and is also not necessary, and in fact also cannot be used.

bodhi/server/scripts/approve_testing.py
```diff
@@ -22,35 +22,21 @@
 spent in the testing repository.
 """
 
-import os
+import argparse
```
bowlofeggs, line 25: Let's not use argparse - we use click everywhere else that we need a CLI lib.

bowlofeggs (changes requested):
I don't think we need a ```--debug``` flag. We have config files to change logging levels - that's how they should be changed.

The PR message says we are moving the code for other uses (I suppose future uses?) - what are those?

cverna:
Let's close this, we can reopen it, if we still think that this is needed

PR #3377, opened 2019-07-10
Automatically create the rawhide updates in a TESTING state

pypingou:
For updates that are automatically created for rawhide there is no
need to have them go through the PENDING state. We can just directly
create them in the TESTING state, let them be tested by CI and let
the approve_testing cron job move them to STABLE once they have
passed their CI.

### DIFF
--- a/bodhi/server/consumers/automatic_updates.py
+++ b/bodhi/server/consumers/automatic_updates.py
@@ -28,7 +28,7 @@ import fedora_messaging
 
 from bodhi.server import buildsys
 from bodhi.server.models import Build, ContentType, Package, Release
-from bodhi.server.models import Update, UpdateRequest, UpdateType, User
+from bodhi.server.models import Update, UpdateStatus, UpdateType, User
 from bodhi.server.util import transactional_session_maker
 
 
@@ -141,10 +141,15 @@ class AutomaticUpdateHandler:
                 stable_karma=3,
                 unstable_karma=-3,
                 user=user,
+                status=UpdateStatus.testing,
             )
 
-            log.debug("Setting request for new update.")
-            update.set_request(dbsession, UpdateRequest.testing, owner_name)
+            # Comment on the update that it was automatically created.
+            update.comment(
+                dbsession,
+                str("This update was automatically created"),
+                author="bodhi",
+            )
 
             log.debug("Adding new update to the database.")
             dbsession.add(update)
--- a/bodhi/tests/server/consumers/test_automatic_updates.py
+++ b/bodhi/tests/server/consumers/test_automatic_updates.py
@@ -22,12 +22,10 @@ from unittest import mock
 import logging
 
 from fedora_messaging.api import Message
-from fedora_messaging.testing import mock_sends
 import pytest
 
-from bodhi.messages.schemas.update import UpdateMessage
 from bodhi.server.consumers.automatic_updates import AutomaticUpdateHandler
-from bodhi.server.models import Build, Release, Update, UpdateType, User
+from bodhi.server.models import Build, Release, Update, UpdateStatus, UpdateType, User
 from bodhi.tests.server import base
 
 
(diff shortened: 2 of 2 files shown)


Zlopez:
These tests are currently failing:
```
bodhi.tests.server.consumers.test_automatic_updates.TestAutomaticUpdateHandler.test_consume
bodhi.tests.server.consumers.test_automatic_updates.TestAutomaticUpdateHandler.test_missing_user
bodhi.tests.server.consumers.test_automatic_updates.TestAutomaticUpdateHandler.test_existing_user
bodhi.tests.server.consumers.test_automatic_updates.TestAutomaticUpdateHandler.test_duplicate_message
```

pypingou:
I expect more tests will fail but I can't look into them right now

Zlopez:
Don't worry I could fix them

bodhi/server/consumers/automatic_updates.py
```diff
@@ -141,10 +141,9 @@ def __call__(self, message: fedora_messaging.api.Message) -> None:
                 stable_karma=3,
                 unstable_karma=-3,
                 user=user,
+                status=UpdateStatus.testing,
             )
 
-            log.debug("Setting request for new update.")
-            update.set_request(dbsession, UpdateRequest.testing, owner_name)
```
Zlopez, line 147: `set_request` is also sending notifications about update. Shouldn't we sent one here now?

cverna: The notification was there to notify the update owner that the update was ready to be pushed to testing. This is not valid any more since we are going to create the update in the testing status directly.

If we feel the need to notify the user that the update was automatically created we can maybe add a comment to the update.

pypingou: @cverna an opinion on this one?

cverna: So thinking about it last night :-), I think we need to add comment to the update so that we notify the packager that an update was automatically created. Otherwise I think that would be silent.

See https://github.com/fedora-infra/bodhi/blob/develop/bodhi/server/scripts/approve_testing.py#L98 for an example on how to add an update comment

pypingou: +1 on having a notification, it seems that comment from "bodhi", "autoqa" and "taskotron" aren't sending notifications though, (bodhi/server/models.py line 3096), not quite sure how to go around this :(

pypingou: I went with "anonymous" for now to see how it looks, happy to change it

cverna: > +1 on having a notification, it seems that comment from "bodhi", "autoqa" and "taskotron" aren't sending notifications though, (bodhi/server/models.py line 3096), not quite sure how to go around this :(

I am pretty sure that `bodhi` user commenting on an update will send a email to the owner of the update. I am not sure we need the fedora-messaging here. 
Do we ?

pypingou: Hm, indeed it does email, just not fedora-messaging.
I kinda think I would like to have fedora-messaging there but this can wait for later.

So it means: changing back anonymous to bodhi and removing all the tests testing fedora-messaging, right?

Zlopez (approved):

bowlofeggs:
We should not have made this change. As [documented](https://bodhi.fedoraproject.org/docs/user/update_states.html), the testing status means that the builds have been pushed out to the updates-testing repository. These builds have not been pushed out. We have a ```test_gating_status``` field to reflect that builds are being tested by CI - there is no reason for the ```status``` field to need to say testing too. Also, lots of places in Bodhi rely on the state reflecting, well, the state of the build.

We should revert this change.

cverna:
> We should not have made this change. As [documented](https://bodhi.fedoraproject.org/docs/user/update_states.html), the testing status means that the builds have been pushed out to the updates-testing repository. 

This works well in the non rawhide case, but we lacked documentation on what the updates flow would be for rawhide. We have tried to take the decision that was the best at that time. 

>These builds have not been pushed out. We have a `test_gating_status` field to reflect that builds are being tested by CI - there is no reason for the `status` field to need to say testing too. Also, lots of places in Bodhi rely on the state reflecting, well, the state of the build.

So would you leave the update in Pending until the nightly rawhide compose creates the repo ?
> 
> We should revert this change.

Once we have the automation needed for multi-build update we should be able to create the update in Pending directly. But the update will be move to testing once all the builds have been signed by robosignatory.

If you could produce a flow diagram of how you think it should work that would be really helpful :smiley:.

PR #3378, opened 2019-07-11
Mark the update as stable if they are not meant to go through testing

pypingou:
In regular releases we have:
- pending-testing
- testing
- pending-stable
- stable

For the rawhide release in Fedora, we have:
- testing
- stable
Updates that are in testing and whose gating status is "green", are
pushed to stable via the approve_testing script. On "regular"
releases they will go from testing to pending-stable, on "rawhide"
we want them to go from testing to stable directly since the compose
process that normally moves the build from pending-stable to stable
is not run by bodhi in rawhide.

This commit also adds a comment precising that setting
mandatory_days_in_testing to 0 on a non-rawhide release may lead to
unexpected behaviour in this part of the code, since non-rawhide
releases are composed within bodhi and thus having the updates marked
as stable outside of the compose process may leads to some issues.

### DIFF
--- a/bodhi/server/scripts/approve_testing.py
+++ b/bodhi/server/scripts/approve_testing.py
@@ -103,6 +103,11 @@ def main(argv=sys.argv):
                 if update.autotime and update.days_in_testing >= update.stable_days:
                     print(f"Automatically marking {update.alias} as stable")
                     update.set_request(db=db, action=UpdateRequest.stable, username="bodhi")
+                    # For updates that are not included in composes run by bodhi itself,
+                    # mark them as stable
+                    if not update.release.composed_by_bodhi:
+                        update.status = UpdateStatus.stable
+                        update.request = None
 
                 db.commit()
 
--- a/bodhi/tests/server/scripts/test_approve_testing.py
+++ b/bodhi/tests/server/scripts/test_approve_testing.py
@@ -635,6 +635,32 @@ class TestMain(BaseTestCase):
                     approve_testing.main(['nosetests', 'some_config.ini'])
 
         self.assertEqual(update.request, models.UpdateRequest.stable)
+        self.assertEqual(update.status, models.UpdateStatus.testing)
+
+    @patch.dict(config, [('fedora.mandatory_days_in_testing', 0)])
+    def test_autotime_update_zero_day_in_testing_no_gated_gets_pushed_to_rawhide(self):
+        """
+        Ensure that an autotime update with 0 mandatory_days_in_testing that meets
+        the test requirements gets pushed to stable in rawhide.
+        """
+        update = self.db.query(models.Update).all()[0]
+        update.autokarma = False
+        update.autotime = True
+        update.request = None
+        update.release.composed_by_bodhi = False
+        update.stable_karma = 10
+        update.stable_days = 0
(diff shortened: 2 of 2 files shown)


bodhi/server/scripts/approve_testing.py
```diff
@@ -103,6 +103,12 @@ def main(argv=sys.argv):
                 if update.autotime and update.days_in_testing >= update.stable_days:
                     print(f"Automatically marking {update.alias} as stable")
                     update.set_request(db=db, action=UpdateRequest.stable, username="bodhi")
+                    # For updates that do not stay in testing at all, mark the update as stable
+                    # Do note that if we ever configure a release which is composed by bodhi with
+                    # mandatory_days_in_testing = 0, its updates will end up being marked as stable
+                    # before the compose. This may lead to some unexpected behavior.
+                    if update.release.mandatory_days_in_testing == 0:
```
cverna:
```suggestion
                    if not update.release.composed_by_bodhi:
```

Turns out we have this handy boolean flag on the release object which was added ([here](https://github.com/fedora-infra/bodhi/issues/2317)), we should use it here.

bodhi/server/scripts/approve_testing.py
```diff
@@ -103,6 +103,12 @@ def main(argv=sys.argv):
...
+                        update.status = UpdateStatus.stable
```
cverna, line 109: I am not 100% sure but we might need to set the update.request to None also.

cverna (changes requested):
It will also need new tests to verify the if condition.

--- 
Thinking about it since we have the `composed_by_bodhi` boolean I think we are sure here that the release is `rawhide` so we could probably tag the build in the update with the stable tag.

bodhi/server/scripts/approve_testing.py
pypingou: Oh nice this should alleviate most of the concerns the comment speaks about, cool!

bodhi/server/scripts/approve_testing.py
pypingou: Do we want to do it now or try this way and adjust later if needed?

cverna: Yes I think we need to do it looking at what the composer does for normal updates. https://github.com/fedora-infra/bodhi/blob/develop/bodhi/server/consumers/composer.py#L664

Zlopez: What about the `update.locked` flag?

pypingou: indeed, :+1:

cverna: > What about the `update.locked` flag?

I think this is not needed ( reading the comment here https://github.com/fedora-infra/bodhi/blob/develop/bodhi/server/models.py#L1646 )

pypingou: > What about the `update.locked` flag?

We don't lock them in our process, they are locked by the compose process itself so packagers do not update them while the compose is running :)

Zlopez (approved):
Everything looks fine

cverna (approved):
LGTM

PR #3379, opened 2019-07-11
Change the tag flow in bodhi for rawhide.

pypingou:
The first flow we thought of was:
- builds are picked from the candidate_tag
- once un-gated, they are pushed into the pending_stable_tag

However, by doing this we realized we wanted/needed a different
flow for two reasons:
- rawhide does not have the notion of "pending", the builds are
  directly in testing. So the updates we create for rawhide are
  directly in the testing status.
- we want builds to the signed before they are tested so the CI
  system really tests what is being shipped.

So we're changing the flow to be:
- builds are picked from pending_testing_tag (where they are pushed
  by robotsignatory)
- once un-gated, they are pushed into the pending_stable_tag (where
  something will push them into their stable tag)
  -> The update are marked as stable at the time their builds are
     pushed into the pending_stable_tag for rawhide.

### DIFF
--- a/bodhi/server/consumers/automatic_updates.py
+++ b/bodhi/server/consumers/automatic_updates.py
@@ -100,7 +100,7 @@ class AutomaticUpdateHandler:
 
         with self.db_factory() as dbsession:
             rel = dbsession.query(Release).filter_by(create_automatic_updates=True,
-                                                     candidate_tag=btag).first()
+                                                     pending_testing_tag=btag).first()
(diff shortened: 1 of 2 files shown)


cverna:
That looks good to me. FYI the rawhide-docs fails because of https://github.com/fedora-infra/bodhi/issues/3300.

I ll let a second reviewer approve the PR.

Zlopez (approved):
LGTM

PR #3391, opened 2019-07-16
Moar coverage

nphilipp:
We recently ran into problems regarding minimum coverage in both the develop and 4.1 branches. This PR fixes some low(ish) hanging fruit and some tests that broke on me while working on it (the first two commits). It increases coverage by ~0.15 percentage points in both branches, so should give us some breathing room.

### DIFF
--- a/bodhi/server/scripts/skopeo_lite.py
+++ b/bodhi/server/scripts/skopeo_lite.py
@@ -689,9 +689,14 @@ class Copier(object):
             self.src.download_blob(digest, size, self.dest.ensure_blob_path(digest))
         elif isinstance(self.src, DirectoryEndpoint) and isinstance(self.dest, RegistryEndpoint):
             self.dest.upload_blob(digest, size, self.src.get_blob_path(digest))
-        elif isinstance(self.src, RegistryEndpoint) and isinstance(self.dest, RegistryEndpoint):
-            if self.src.registry == self.dest.registry:
-                self.dest.link_blob(digest, self.src.repo)
+        else:
+            # The copy() function below ensures that the following two asserts don't fail.
+            assert (isinstance(self.src, RegistryEndpoint)
+                    and isinstance(self.dest, RegistryEndpoint))
+            assert self.src.registry == self.dest.registry
+
+            self.dest.link_blob(digest, self.src.repo)
+
         # Other forms of copying are not needed currently, and not implemented
 
     def _copy_manifest(self, info, toplevel=False):
@@ -761,6 +766,11 @@ def copy(src, dest, src_creds, src_tls_verify, src_cert_dir, dest_creds, dest_tl
     src = parse_spec(src, src_creds, src_tls_verify, src_cert_dir)
     dest = parse_spec(dest, dest_creds, dest_tls_verify, dest_cert_dir)
 
+    # Copier._copy_blob() assumes this:
+    # - The src and dest registries are of types RegistryEndpoint or DirectoryEndpoint, at least one
+    #   must be a RegistryEndpoint.
+    # - If both are a RegistryEndpoint, they must be the same so linking is possible, therefore we
+    #   copy src reg -> local dir -> dest reg if this isn't the case.
     if src.registry != dest.registry:
         tempdir = tempfile.mkdtemp()
         try:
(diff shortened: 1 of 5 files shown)


bodhi/server/scripts/skopeo_lite.py
```diff
@@ -692,7 +692,11 @@ def _copy_blob(self, digest, size):
         elif isinstance(self.src, RegistryEndpoint) and isinstance(self.dest, RegistryEndpoint):
             if self.src.registry == self.dest.registry:
                 self.dest.link_blob(digest, self.src.repo)
-        # Other forms of copying are not needed currently, and not implemented
+            else:  # pragma: no cover
```
Zlopez: I'm not sure if these kind of changes should be done in PR, that is increasing the coverage.

bodhi/server/util.py
```diff
@@ -312,7 +312,7 @@ def sanity_check_repodata(myurl, repo_type):
 
         if repo_type in ('yum', 'source'):
             tests.append((['list', 'available'], 'testrepo'))
-        elif repo_type == 'module':
+        else:  # repo_type == 'module', verified above
```
Zlopez, line 315: If there will be new `repo_type` in future (for example 'flatpak') we could easily forget about this change.

Zlopez (commented):
I have a few comments, otherwise it looks good

bodhi/server/scripts/skopeo_lite.py
```diff
@@ -692,7 +692,11 @@ def _copy_blob(self, digest, size):
...
+                raise RuntimeError("Copier can't copy between registries directly.")
+        else:  # pragma: no cover
+            # Other forms of copying are not needed currently, and not implemented
+            raise RuntimeError(f"Can't copy between registries: {self.src!r}, {self.dest!r}")
```
nphilipp: I'm not sure about introducing `pragma: no cover` here because we don't use it anywhere else. Alternatively, I could replace the conditions that can't fail with asserts.

bodhi/server/scripts/skopeo_lite.py
nphilipp: See my own comment there: "I'm not sure about introducing pragma: no cover here because we don't use it anywhere else. Alternatively, I could replace the conditions that can't fail with asserts." What do you think?

bodhi/server/util.py
nphilipp: The existing code would just silently do nothing in this case :wink:. I'll add asserts to guard against this.

bodhi/server/scripts/skopeo_lite.py
Zlopez: `pragma: no cover` is only recommended, if the code can't be reached or can't be tested by any way

bodhi/server/util.py
pypingou: Raise an exception if the repo_type isn't supported?

bodhi/server/scripts/skopeo_lite.py
nphilipp: Yeah, this would have been the case, but the way I did it obviates introducing it at all. :smile:

bodhi/server/util.py
nphilipp: Yup, just through using asserts which don't (shouldn't) affect coverage.

Zlopez:
@nphilipp 
It looks like `bodhi.tests.server.scripts.test_skopeo_lite` is failing again.

bodhi/server/scripts/skopeo_lite.py
```diff
@@ -689,10 +689,16 @@ def _copy_blob(self, digest, size):
...
+        else:
+            # The copy() function below ensures that the following two asserts don't fail.
+            assert (isinstance(self.src, RegistryEndpoint)
+                    and isinstance(self.dest, RegistryEndpoint))
+            assert self.src.registry == self.dest.registry
+
+            self.dest.link_blob(digest, self.src.repo)
+
         # Other forms of copying are not needed currently, and not implemented
+            raise RuntimeError(f"Can't copy between registries: {self.src!r}, {self.dest!r}")
```
Zlopez: This line is currently breaking up the `bodhi/tests/server/scripts/test_skopeo_lite.test_skopeo_copy_link` test

nphilipp: Oops, so much for "fixing" my changes :)

bodhi/server/scripts/skopeo_lite.py
nphilipp: Current push has the asserts instead.

Zlopez (approved):
Working as expected

PR #3394, opened 2019-07-17
Go to stable directly for rawhide-like releases.

pypingou:
For releases that do not go through testing and aren't composed by
bodhi, go to stable directly.
This means, move the update's build into update.release.stable_tag and
remove the update.release.pending_testing_tag (where the update was when
it was picked up) and update.release.pending_stable_tag which was just
added when the update was marked pending stable.


Zlopez (approved):
Looks good to me

nphilipp (approved):

PR #3397, opened 2019-07-17
Fix automatically create gated updates if gating is on

pypingou:
When bodhi automatically creates new updates for certain releases,
these updates need to be marked as waiting for CI if that bodhi
instance is configured for gating.
Otherwise, they end up with a test_gating_status that is `None`
and that breaks the gating later on.

### DIFF
--- a/bodhi/server/consumers/automatic_updates.py
+++ b/bodhi/server/consumers/automatic_updates.py
@@ -27,10 +27,10 @@ import logging
 import fedora_messaging
 
 from bodhi.server import buildsys
-from bodhi.server.models import Build, ContentType, Package, Release
+from bodhi.server.config import config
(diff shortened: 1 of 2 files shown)


Zlopez (approved):

Zlopez:
`pip-flake8` error:
```
16:13:14  ./bodhi/tests/server/consumers/test_automatic_updates.py:84:42: E711 comparison to None should be 'if cond is None:'
16:13:14          assert update.test_gating_status == None
```

pypingou:
> `pip-flake8` error:
> 
> ```
> 16:13:14  ./bodhi/tests/server/consumers/test_automatic_updates.py:84:42: E711 comparison to None should be 'if cond is None:'
> 16:13:14          assert update.test_gating_status == None
> ```

Fixed, thanks :)

PR #3399, opened 2019-07-17
Get gating status from Greenwave for created updates

nphilipp:
This should let us do the right thing if no tests are required for this
update, i.e. Greenwave won't send us any updates either.


nphilipp:
parallel PR for 4.1: #3400

Zlopez (approved):

PR #3400, opened 2019-07-17
Get gating status from Greenwave for created updates

nphilipp:
This should let us do the right thing if no tests are required for this
update, i.e. Greenwave won't send us any updates either.

(cherry picked from commit a6fc97117ee8139d5aabe8093ccc1820fee100c2)

### DIFF
--- a/bodhi/server/consumers/automatic_updates.py
+++ b/bodhi/server/consumers/automatic_updates.py
@@ -146,8 +146,10 @@ class AutomaticUpdateHandler:
             if config.get('test_gating.required'):
                 log.debug(
                     'Test gating required is enforced, marking the update as '
-                    'waiting on test gating')
+                    'waiting on test gating and updating it from Greenwave to '
+                    'get the real status.')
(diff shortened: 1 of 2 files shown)


Zlopez (approved):

bodhi/server/consumers/automatic_updates.py
```diff
@@ -146,8 +146,10 @@ def __call__(self, message: fedora_messaging.api.Message) -> None:
             if config.get('test_gating.required'):
                 log.debug(
                     'Test gating required is enforced, marking the update as '
-                    'waiting on test gating')
+                    'waiting on test gating and updating it from Greenwave to '
+                    'get the real status.')
                 update.test_gating_status = TestGatingStatus.waiting
+                update.update_test_gating_status()
```
pypingou, line 152: I was wondering if we need to do this after the commit() or not?

pypingou: reading the code again, I think this is fine as is

PR #3402, opened 2019-07-18
Ensure correct order of automatic update comments

### LINKED ISSUE #3401: [s] Fix comment order: https://bodhi.stg.fedoraproject.org/updates/FEDORA-2019-a0e3451113
The comment that the update was created automatically appears after the comments generated by checking greenwave.

nphilipp:
Previously, gating status was checked before the comment about creating
the update was added. Because the timestamp of comments defaults to
datetime.utcnow() (evaluated when the object is created), this caused
the creation comment to be sorted after the two comments concerning the
test gating status.

fixes: #3401


pypingou (approved):
Nice

Zlopez (approved):

PR #3415, opened 2019-07-23
When automatically creating an update, obsolete the pending ones

pypingou:
If an update is automatically created, there could still be one
that was manually created by the user and which is laying around
in a "Pending" state (the one automatically created being created
in a "Testing" state).
So to not have these updates laying around doing nothing forever,
we can just obsolete them.

But we do not want to obsolete the ones in Testing as the new build
may fail CI while the old one passes and thus would land in the
rawhide buildroot.

### DIFF
--- a/bodhi/server/consumers/automatic_updates.py
+++ b/bodhi/server/consumers/automatic_updates.py
@@ -151,6 +151,10 @@ class AutomaticUpdateHandler:
                 author="bodhi",
             )
 
+            # Obsolete all the updates that are in pending and having this
+            # build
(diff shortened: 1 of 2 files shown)


pypingou:
This still needs some tests :)

pypingou:
After more thoughts, this may actually not work since ``automatic_updates`` is going to try to create an update with a build for which there is already an update existing, so it may crash before it got a chance to obsolete the "Pending" one.

pypingou:
Closing this one, the code remains here, but I don't think it solves the issue I was trying to address.

PR #3417, opened 2019-07-24
Handle situation where there is an existing update but in Pending

### LINKED ISSUE #3414: Running fedpkg update on master leads to a stuck update
If you run `fedpkg update` in master with rawhide gating enable, the update that is created is in "Pending" state.
This means, the update will remain "stuck" in bodhi since the workflow assumes the update is created in "Testing".

For now we could "fix" it by "manually" move all the "Pending" updates to "Testing" every once in a while but this is clearly not a long term viable solution.

We can do better than cron job! :D

pypingou:
If the user manually creates the bodhi update for a release for which the
update should have been automatically created, that update will stay in
a Pending state with nothing moving it forward.
When we finally receive the message from koji about a build being tagged
in the tag of interest, we will discard it since there is already an
update for it.
This results in the update being basically stuck in bodhi.

With this commit, if are notified of a build tagged in a koji tag of
interest, we will check if there is an existing update and if there is
one that is in "Pending" state, we will move it to "Testing" and check
its gating status, effectively allowing to move the update to the next
stage from where it can be picked by the regular workflow.

Fixes https://github.com/fedora-infra/bodhi/issues/3414

### DIFF
--- a/bodhi/server/consumers/automatic_updates.py
+++ b/bodhi/server/consumers/automatic_updates.py
@@ -109,7 +109,22 @@ class AutomaticUpdateHandler:
             bcls = ContentType.infer_content_class(Build, kbuildinfo)
             build = bcls.get(bnvr)
             if build:
-                log.info(f"Build, active update for {bnvr} exists already, skipping.")
+                if build.update.status == UpdateStatus.pending:
+                    log.info(
+                        f"Build, active update for {bnvr} exists already "
+                        "in Pending, moving it along.")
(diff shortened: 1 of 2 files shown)


nphilipp:
Looks good to me, the explanation of the problem in the commit log is very helpful! As discussed on IRC, I'll reword the commit log, log messages, etc. a little.

nphilipp (approved):
Looks good to me!

PR #3421, opened 2019-07-25
Deal with message handler exceptions individually

nphilipp:
Previously, if one handler raised an exception it would cause other
handlers interested in the same message suffix to be skipped.


nphilipp:
This is WIP/draft because of the discussion that needs to happen in #3422 first.

abompard:
Apparently the discussion has reached a consensus, are you still willing to work on this @nphilipp ?

bodhi/server/consumers/__init__.py
```diff
@@ -61,18 +67,21 @@ def __call__(self, msg: fedora_messaging.api.Message):  # noqa: D401
...
-            if msg.topic.endswith('.buildsys.tag'):
-                log.debug('Passing message to the Signed handler')
-                self.signed_handler(msg)
+        error_handlers_msgs = []
 
-                log.debug('Passing message to the Automatic Update handler')
-                self.automatic_update_handler(msg)
+        for handler_info in self.handler_infos:
+            try:
+                if msg.topic.endswith(handler_info.topic_suffix):
```
abompard: There's little chance that the `endswith()` method raise an exception, so you could write all that as:
```py
for handler_info in self.handler_infos:
    if not msg.topic.endswith(handler_info.topic_suffix):
        continue
    log.debug(...)
    try:
        handler_info.handler(msg)
    except Exception as e:
        log.exception(...)
        error_handlers_msgs.append(....)
```
And you'll get less indentation levels! :-)

nphilipp: Good catch, I'll change it.

PR #3423, opened 2019-07-25
Fix a couple tests on F31+

abompard:
Those tests fail in F31+ because the newer version of the OpenID client library opens the configuration file differently.


pypingou:
Jenkins seems not happy about these tests :(

abompard:
@pypingou Jenkins is happy now, care to re-review?

pypingou (approved):

PR #3426, opened 2019-07-28
Use more recent Fedora version in placeholder text

robert-scheck:


pypingou:
Looks like the tests aren't happy :)
But this seems related to https://github.com/fedora-infra/bodhi/issues/3427 not to your changes.

We'll need to look into this

abompard (approved):
Thanks.

Issue #3431, opened 2019-07-29
Reduce the amount of emails sent by bodhi in rawhide


pypingou:
Bodhi currently sends a lot of emails (5 to 7) for each rawhide update going through.

We need to reduce this.

pypingou:
With #3435 the number of emails goes down from 5 or 7 to 2 or 3 (update created, (CI failed), update going to stable).

PR #3433, opened 2019-07-29
Only send an email for the comments added when CI fails

pypingou:
Relates to https://github.com/fedora-infra/bodhi/issues/3431

### DIFF
--- a/bodhi/server/models.py
+++ b/bodhi/server/models.py
@@ -3066,7 +3066,8 @@ class Update(Base):
         return
 
     def comment(self, session, text, karma=0, author=None, karma_critpath=0,
-                bug_feedback=None, testcase_feedback=None, check_karma=True):
+                bug_feedback=None, testcase_feedback=None, check_karma=True,
+                email_notification=True):
         """Add a comment to this update.
 
         If the karma reaches the 'stable_karma' value, then request that this update be marked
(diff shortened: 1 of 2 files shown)


bodhi/server/models.py
```diff
@@ -3706,6 +3708,9 @@ def comment_on_test_gating_status_change(target, value, old, initiator):
         """
         Place comment on the update when ``test_gating_status`` changes.
 
+        Only notify the users by email if the new status is not in ``waiting``,
```
abompard: This looks like it's saying the opposite from what it's doing. Shouldn't `not` be removed? Also, what about the `waiting` state, it's not in the code below.

bodhi/tests/server/test_models.py
```diff
@@ -4164,6 +4165,37 @@ def test_comment_on_test_gating_status_change(self):
...
+        self.assertEqual(len(self.obj.comments), 1)
+
+        # Let's set test_gating_status to 'waiting' once again.
+        self.obj.test_gating_status = TestGatingStatus.waiting
+
+        # Check that still only one email was sent:
+        self.assertEqual(mail.send.call_count, 1)
+
+        # We should have still only one comment about test_gating_status change.
+        self.assertEqual(len(self.obj.comments), 2)
```
abompard, line 4197: I don't understand, you want to test that there's still only one comment, but you assert that the length is 2 here.

bodhi/server/models.py
pypingou: sorry missed to update the comment in both sections, will adjust

abompard:
Obsoleted by #3435

PR #3446, opened 2019-08-01
Fix the automatic_updates consumer when the build exists but not in an update

pypingou:
There are situation apparently where the build exists in bodhi's
database but is not associated with an update (that's the case
when the user created a buildroot override with that build).

So we need to check if the build is associated to an update before
we can check the status of that update.

### DIFF
--- a/bodhi/server/consumers/automatic_updates.py
+++ b/bodhi/server/consumers/automatic_updates.py
@@ -108,7 +108,7 @@ class AutomaticUpdateHandler:
 
             bcls = ContentType.infer_content_class(Build, kbuildinfo)
             build = bcls.get(bnvr)
-            if build:
+            if build and build.update:
                 if build.update.status == UpdateStatus.pending:
                     log.info(
                         f"Build, active update for {bnvr} exists already "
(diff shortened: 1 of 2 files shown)


bodhi/server/consumers/automatic_updates.py
```diff
@@ -108,7 +108,7 @@ def __call__(self, message: fedora_messaging.api.Message) -> None:
 
             bcls = ContentType.infer_content_class(Build, kbuildinfo)
             build = bcls.get(bnvr)
-            if build:
+            if build and build.update:
```
nphilipp, line 111: This looks to be in the wrong place, we definitely want it to return on line 128 and not try to create the same build again if it exists.

nphilipp (changes requested):
Apart from fixing that it should avoid to create the build if it exists (but `build.update` is `None`), we should add a test case.

nphilipp:
FYI, apart from adding a test, I've brought the automatic updates consumer tests back in order (a commit of its own) and reworded the commit log.

bodhi/tests/server/consumers/test_automatic_updates.py
```diff
@@ -89,6 +89,41 @@ def test_consume(self, caplog):
...
+        # check if the update exists...
+        update = self.db.query(Update).filter(
+            Update.builds.any(Build.nvr == self.sample_nvr)
+        ).first()
+
+        # ...and some of its properties
+        assert update is not None
+        assert update.type == UpdateType.unspecified
+        assert update.status == UpdateStatus.testing
+        assert update.test_gating_status is None
```
pypingou, line 120: Shouldn't that status be `waiting`?

nphilipp: Valid objection. I guess why the tests succeed (for me at least) is because we don't actually ask Greenwave in tests, i.e. don't update anything. But we should mock this part, too. I'll get on it.

nphilipp: Then again, no: it's because in these tests, test gating isn't required so the field won't be set. All is well. :smiley:

nphilipp (approved):
LGTM.

PR #3448, opened 2019-08-02
Improve the performance of validate_build_uniqueness.

cverna:
Instead of looping twice against all the build to check
for uniqueness use a set to record the previously seen
build. If a build is already in the set it means that we
found a duplicate.
Doing the same for the tuple (Package, Release) which
allow us to detect multiple builds for a package on the
same release.

Performance testing against an update with 103 builds.
Before change : 25,9s for the request to complete
After change : 5.6s for the request to complete


cverna:
See #3449 for the CI failure

pypingou (approved):
Nice :)

PR #3453, opened 2019-08-05
Overhaul the new update form

ryanlerch:
Redesigned the new update form to make it easier to use. Also simplifies the adding of builds so we search for builds in koji directly without having to specify a package name first.

screencapture-localhost-6543-updates-new-2019-08-05-17_42_50


mattiaverga:
It looks nice and a lot more polished than before to me, just a few suggestions:

- it would be nice if the bugs list could display some bug information like #2821 . Moreover, we could fetch the bug list directly from bugzilla, by passing the bugzilla base url and the bz_products list from bodhi config, so that the user interface is not made specifically for Fedora only and a single bugzilla query can fetch all the information required instead of having the bug list fetched from apps.fedoraproject.org and then query bugzilla for every bug.
- the CLI has now the ability to create an update from a side tag, so the should make the update form also capable of doing that.

ryanlerch:
> It looks nice and a lot more polished than before to me, just a few suggestions:
> 
> * it would be nice if the bugs list could display some bug information like #2821 . Moreover, we could fetch the bug list directly from bugzilla, by passing the bugzilla base url and the bz_products list from bodhi config, so that the user interface is not made specifically for Fedora only and a single bugzilla query can fetch all the information required instead of having the bug list fetched from apps.fedoraproject.org and then query bugzilla for every bug.
Yeah, this was something that i was going to loop back on, because the current call to apps only pull 8 bugs, and freeform the rest. That said, i didnt realise #2821 was there too -- sorry!

The builds adder is actually using the same selectize.js control as the new bugs adder (just configured and styled a different). The bugs adder should probably be the same style, and use the code in #2821 for pulling from BZ directly rather than using the fcomm connector thing on app.fp.org.

> * the CLI has now the ability to create an update from a side tag, so the should make the update form also capable of doing that.

Yeah, that one is next on my list :) the goal of this PR was to try to clean up this form a bit before adding that new functionality :)

ryanlerch:
Added a bunch of upgraded features, mainly pulling additional data from bugzilla.

thanks to @mattiaverga and #2821 for some of the code in there too!

cverna:
@ryanlerch it seems that the unit tests are failing in the CI environment. See the tests results --> https://ci.centos.org/job/bodhi-pipeline/job/PR-3453/4/testReport/

bodhi/server/templates/master.html
```diff
@@ -38,8 +38,14 @@
     <link href="${request.static_url('bodhi:server/static/css/site.css') }" rel="stylesheet" />
     <link href="${request.static_url('bodhi:server/static/css/datagrepper-feed.css') }" rel="stylesheet" />
     <link href="${request.static_url('bodhi:server/static/css/panel.css') }" rel="stylesheet" />
+    <link href="${request.static_url('bodhi:server/static/selectize/selectize.bootstrap3.css') }" rel="stylesheet" />
     <script src="${request.static_url('bodhi:server/static/js/jquery-1.10.2.min.js')}"></script>
     <script src="${request.static_url('bodhi:server/static/js/Chart-0.2.0.min.js')}"></script>
+    <script type="text/javascript">
+      var settings = {
+          bz_products: ["${ '","'.join(config['bz_products']) | n }"]
+      };
```
mattiaverga, line 48: Probably we could make use of a `bz_baseurl` entry in Bodhi config, so that we don't hardcode the Bugzilla base URL in the two ajax calls in update_form.js

bodhi/server/static/js/update_form.js
```diff
@@ -26,314 +26,192 @@ $(document).ready(function() {
...
+                            type: 'error',
+                        });
+                        $("#bugs-card .selectize-control").removeClass("loading")
+                        callback();
+                    } else {
+                        // Check Bug product
+                        var product = data.bugs[0].product;
+                        if (settings.bz_products.indexOf(product) == -1) {
+                            messenger.post({
+                                message: 'Bug #' + data.bugs[0].id + ' doesn\'t seem to refer to Fedora or Fedora EPEL.\nAre you sure you want to reference it in this update? Bodhi will not be able to operate on this bug!',
```
mattiaverga: Should we remove the reference to Fedora and be more generic?

ryanlerch: Okay, updated this string to read:

    doesn\'t seem to refer to a product that Bodhi manages updates for.\nAre you sure you want to reference it in this update? Bodhi will not be able to operate on this bug!

bodhi/server/templates/master.html
ryanlerch: added an item to the config to add the BZ rest url, and referenced it in the script. thaznks!

ryanlerch:
jenkies test

bodhi/server/static/js/forms.js
```diff
@@ -119,6 +119,16 @@ Form.prototype.data = function() {
         }
     });
 
+    $(this.idx + " select").each(function() {
+        var value = $(this).val();
+        console.log(value)
```
cverna: I guess this was used for debug, we should probably remove it ?

cverna (commented):
Overall LGTM, just one small comment

bodhi/server/static/js/forms.js
ryanlerch: Correct! removed.

Thanks!

cverna (approved):

PR #3457, opened 2019-08-09
add option to vagrant provisioning to use stg infra

ryanlerch:
This adds an option when provisioning the bodhi devel
envrionment with vagrant to set up the config to use
the fedora staging infra.

to use this option, the box is provisioned with

    vagrant --use-staging up

note that talking to koji stg requires a krb ticket, so
before starting bodhi, get a new ticket with:

    kinit <fasusername>@STG.FEDORAPROJECT.ORG

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -7,8 +7,37 @@
 # cp devel/Vagrantfile.example Vagrantfile
 # vagrant up
 
-require 'etc'
+# There is also an option when provisioning to setup bodhi
+# to use the fedora stg infratucture for koji and src.stg (for package acls)
+# To use this, use the command:
+#
+# vagrant --use-staging up
+# 
+# vagrant will prompt you for for fas username for auth to koji. Additionally
+# after the box is finished, you will need to aquire a kerberos ticket periodically
+# with `kinit <fasusername>@STG.FEDORAPROJECT.ORG` for bodhi to be able to auth with
+# the stg koji.
 
+require 'etc'
+require 'getoptlong'
+
+opts = GetoptLong.new(
+  [ '--use-staging', GetoptLong::OPTIONAL_ARGUMENT ]
+)
+use_staging_infra = false
+
+opts.each do |opt, arg|
+  case opt
+    when '--use-staging'
+        use_staging_infra=true
+  end
+end
+
+if use_staging_infra
+    print "Enter your FAS username for krb auth to staging: "
+    fasusername = STDIN.gets.chomp
+    print "\n"
+end
 
 VAGRANTFILE_API_VERSION = "2"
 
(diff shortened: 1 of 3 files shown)


Vagrantfile
```diff
@@ -54,12 +83,19 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
  # config.vm.synced_folder ".dnf-cache", "/var/cache/dnf", type: "sshfs", sshfs_opts_append: "-o nonempty"
 
  # Comment this line if you would like to disable the automatic update during provisioning
- config.vm.provision "shell", inline: "sudo dnf upgrade -y"
+ #config.vm.provision "shell", inline: "sudo dnf upgrade -y"
```
abompard: Is this intended for this change? It seems unrelated.

devel/ansible/roles/bodhi/tasks/main.yml
```diff
@@ -144,6 +144,10 @@
   args:
       creates: /home/vagrant/bodhi/development.ini
 
+- name: Change development.ini to make bodhi use the staging infrastructure
+  import_tasks: configure_stg.yml
+  when: fas is defined
```
abompard: Using the `fas` variable for that seems a bit non-intuitive, I think it'd be better if we defined a `staging` variable or a `env` variable and set it to `staging`, and use that here.

abompard (commented):
A couple remarks, nothing big.

Vagrantfile
ryanlerch: yup! an unintended change. turned this off during testing. sorry!

devel/ansible/roles/bodhi/tasks/main.yml
ryanlerch: the fas variable is what is passed into ansible from vagrant (the changes to the vagrantfile now prompts the user for the fas username for the confgiure_stg.yml to add to the config).

I'll pass in the use_staging_infra boolean too, change the fas variable to `staging_fas` and change the when line to 

    when: (staging_fas is defined) and use_staging_infra

for extra clarity here.

does that sound ok?

ryanlerch:
@abompard updated. comments about the fixes are replies to your inline comments.

devel/ansible/roles/bodhi/tasks/main.yml
abompard: Yes, thanks :-)

Vagrantfile
abompard: No problem, but it seems so be still here...

ryanlerch: gah, sorry!

i uncommented it, then commented it back out again to speed up testing of the other change, then forgot to uncomment it again! sorry!

just updated the commit again to make sure it is NOT commented out!

abompard:
The CI failure comes from new deprecation warnings in Rawhide that mess up the output, I think they can be ignored for now.

abompard (approved):

PR #3458, opened 2019-08-12
Display the number of updates that passed or ignored gating per release.

cverna:
This commit add the number of updates that passed or ignored the
gating tests in the release overview.
It is also adding the possibility to filter updates using the
gating status in the url.


cverna:
Example from the vagrant environment 
Screenshot_2019-08-12 Fedora Updates System

nphilipp (approved):
Looks good to me.

PR #3461, opened 2019-08-13
Make the integration tests aware of an update's display_name

abompard:
This should fix the integration test failures in CI


cverna (approved):
LGTM

Issue #3429, opened 2019-07-29
Rawhide updates ruining the bodhi.fedoraproject.org/users/<me> experience


hroncok:
I was using https://bodhi.fedoraproject.org/users/churchyard to see my recent updates and check whether they are still in testing.

Since I don't usually have more than 10 updates in testing at the same time, the page was perfect.

However I do dozens (hundreds?) rawhide builds a week and now rawhide updates are swamping the page, making it essentially useless for me.

I'd appreciate, if I could hide the rawhide updates (maybe via an URL argument, such as `?release=!F31`). Thanks.

cverna:
Does https://bodhi.fedoraproject.org/updates/?user=churchyard&status=testing works for you ? Rawhide update will not stay long in testing unless the tests are failing.

hroncok:
Can I see testing + pending?

hroncok:
Also, this doesn't list the buildroot overrides at all, so from one simple dashboard page, I would need to use more.

cverna:
Ok then I guess we need some way to filter the updates per releases. Currently you can do something like that https://bodhi.fedoraproject.org/updates/?user=churchyard&release=F30 but it seems limited to one release.

ryanlerch:
> Can I see testing + pending?

Just opened up #3463  that lets you specify multiple status params in the updates query, so it will let you do testing + pending.

The same code that renders the /updates/ page is used to render the update list on the user profile page, so what i am thinking here is to add a new box at the top that shows just the pending and testing updates. (perhaps only showing this box if the user page is your page too maybe)

Filtering by if it is an auto update or not could be problematic if one of your auto updates doo get stuck in pending or testing, you probably still want to know about that.

hroncok:
> Filtering by if it is an auto update or not could be problematic if one of your auto updates doo get stuck in pending or testing, you probably still want to know about that.

I think you are right here. Thanks!

ryanlerch:
The new dashboard view now shows your pending and testing updates on the index (when logged in)

![screencapture-0-0-0-0-6543-2019-09-12-15_58_29](https://user-images.githubusercontent.com/592259/64757812-79554780-d576-11e9-90cf-98c36881c2a9.png)

#3500 is in develop branch now.

going to close this one, feel free to re-open if this doesnt meet your expectaions, @hroncok

hroncok:
Looks awesome!

PR #3463, opened 2019-08-14
allow multiple status params for update list views

ryanlerch:
Previously, when querying a list of updates, only a
single status parameter could be specified, e.g.:

https://bodhi.fedoraproject.org/updates/?user=churchyard&status=testing

this commit changes that so multiple status parameters
can be supplied, for example:

https://bodhi.fedoraproject.org/updates/?user=churchyard&status=testing&status=pending

Related: #3429


cverna (approved):
LGTM

PR #3466, opened 2019-08-14
Send a message when an update is ready to be tested.

### LINKED ISSUE #3428: Have bodhi send a dedicated fedora-messaging message when an update is ready to be tested
Updates are manually created in a "Pending" state, where they are signed before they are pushed to a "Testing" where they are tested by the CI system.
So we can't trigger the CI system off new updates being created.

The easiest solution is thus to send a dedicated fedora-messaging message to tell the CI System: "Now this update can be tested".

This need coordination with the OSCI team maintaining the CI system.

abompard:
Fixes #3428


Zlopez (approved):
LGTM

PR #3467, opened 2019-08-14
Update the greenwave policy for F32

abompard:
This will fix CI


abompard:
The policy is directly imported from the Fedora infra Ansible repo.

cverna (approved):
LGTM

abompard:
I'm adding a commit that is unrelated to greenwave but prevent the CI from passing. This way we'll see if we actually get a green CI after this PR.

PR #3474, opened 2019-08-19
Allow to configure a release without an override tag.

### LINKED ISSUE #3447: Do not allow buildroot override in rawhide
If the user creates a buildroot override in rawhide it ends up messing with our entire workflow. The builds are no longer in their appropriate tags, they appear in bodhi's database but aren't linked to any update (cf PR: #3446) that leads to basically builds being stuck.

Preventing buildroot override creation in rawhide should just solve that :)

cverna:
A few releases do not make use of the buildroot override
(eg container, rawhide, flatpak). This commit makes it possible
to create a release without the override tag.
If an user tries to create a buildroot override for such a release
an error message will be displayed.

Fixes #3447


bodhi/server/validators.py
```diff
@@ -1091,6 +1091,11 @@ def _validate_override_build(request, nvr, db):
 
             build.release = release
 
+        if not build.release.override_tag:
+            request.errors.add("body", "nvr", "Cannot create a buildroot override"
+                               " Release associated to build does not support it.")
```
nphilipp: The error message doesn't read well, how about:

```
"Cannot create a buildroot override because the release associated with the build does not support it."
```

nphilipp (approved):
Looks good to me, except the error message string (see below).

nphilipp (changes requested):
Oops, missed the one error message string.

bodhi/server/validators.py
cverna: Yes that's much better :smile: . I have updated the PR

nphilipp (approved):
Thanks!

PR #3481, opened 2019-08-22
Create additional side tags on multi build tag

### LINKED ISSUE #3473: Update from side tag: create appropriate auxiliary tags and tag builds accordingly
# Story

As a user of multi-build gating,

I want that Bodhi creates corresponding `*-pending-signing` and `*-testing` tags (if they don't exist already) when an update is created or edited from a side-tag, and tags the builds into the `*-pending-signing` tag and out of the side-tag,

so that robosignatory can pick the build up from there and sign it.

# Background

@pypingou's multi-build update workflow diagram \[[download](https://user-images.githubusercontent.com/820624/63262799-34940480-c287-11e9-9594-e177ce7e6d4d.png)\]:

Zlopez:
Create new side tags from initial side tag and move builds to newly
created sidetag.

Closes #3473

### DIFF
--- a/bodhi/server/buildsys.py
+++ b/bodhi/server/buildsys.py
@@ -126,6 +126,7 @@ class DevBuildsys:
         cls.__added__ = []
         cls.__tagged__ = {}
         cls.__rpms__ = []
+        cls.__tags__ = []
 
     def multiCall(self):
         """Emulate Koji's multiCall."""
@@ -487,7 +488,7 @@ class DevBuildsys:
                  'parent_id': 6438, 'maxdepth': None, 'noconfig': False, 'child_id': 6441,
                  'nextdepth': None, 'filter': [], 'currdepth': 4}]
 
-    def addTag(self, tag: str, **opts):
+    def createTag(self, tag: str, **opts):
         """Emulate tag adding."""
         if 'parent' not in opts:
             raise ValueError('No parent in tag options')
--- a/bodhi/server/models.py
+++ b/bodhi/server/models.py
@@ -982,6 +982,34 @@ class Release(Base):
         """
         return config.get(f'{self.setting_prefix}.status', None)
 
+    def get_testing_side_tag(self, from_tag: str) -> str:
+        """
+        Return the testing side tag for this ``Release``.
+
+        Args:
+            from_tag: Name of side tag from which ``Update`` was created.
+
+        Returns:
+            Testing side tag used in koji.
+        """
+        side_tag_postfix = config.get(
+            f'{self.setting_prefix}.koji-testing-side-tag', "-testing")
+        return from_tag + side_tag_postfix
+
+    def get_pending_signing_side_tag(self, from_tag: str) -> str:
+        """
+        Return the testing side tag for this ``Release``.
+
+        Args:
+            from_tag: Name of side tag from which ``Update`` was created.
+
+        Returns:
+            Testing side tag used in koji.
+        """
+        side_tag_postfix = config.get(
+            f'{self.setting_prefix}.koji-pending-signing-side-tag', "-pending-signing")
+        return from_tag + side_tag_postfix
+
 
 class TestCase(Base):
     """
--- a/bodhi/server/services/updates.py
+++ b/bodhi/server/services/updates.py
@@ -26,7 +26,7 @@ from sqlalchemy import func, distinct
 from sqlalchemy.sql import or_
 from requests import RequestException, Timeout as RequestsTimeout
 
-from bodhi.server import log, security
+from bodhi.server import log, security, buildsys
 from bodhi.server.exceptions import BodhiException, LockedUpdateException
 from bodhi.server.models import (
     Update,
@@ -440,6 +440,8 @@ def new_update(request):
     empty, the list of builds will be filled with the latest builds in this
     Koji tag. This is done by validate_from_tag() because the list of builds
     needs to be available in validate_acls().
+    Ensure that related tags ``from_tag``-pending-signing and ``from_tag``-testing
+    exists and if not create them in Koji.
 
     Args:
         request (pyramid.request): The current request.
@@ -492,8 +494,8 @@ def new_update(request):
         # the builds as signed.
         request.db.commit()
 
-        # After we commit the transaction, we need to get the builds and releases again, since they
-        # were tied to the previous session that has now been terminated.
+        # After we commit the transaction, we need to get the builds and releases again,
+        # since they were tied to the previous session that has now been terminated.
         builds = []
         releases = set()
         for nvr in build_nvrs:
@@ -536,6 +538,24 @@ def new_update(request):
 
             if len(releases) > 1:
                 result = dict(updates=updates)
+
+            if from_tag:
+                koji = buildsys.get_session()
+                for update in updates:
+                    # Validate that <koji_tag>-pending-signing and <koji-tag>-testing exists,
+                    # if not create them
+                    side_tag_pending_signing = update.release.get_pending_signing_side_tag(
+                        from_tag)
+                    side_tag_testing = update.release.get_testing_side_tag(from_tag)
+                    if not koji.getTag(side_tag_pending_signing):
+                        koji.createTag(side_tag_pending_signing, parent=from_tag)
+                    if not koji.getTag(side_tag_testing):
+                        koji.createTag(side_tag_testing, parent=from_tag)
+
+                    to_tag = side_tag_pending_signing
+                    # Move every new build to <from_tag>-pending-signing tag
+                    update.add_tag(to_tag)
+
     except LockedUpdateException as e:
         log.warning(str(e))
         request.errors.add('body', 'builds', "%s" % str(e))
--- a/bodhi/server/validators.py
+++ b/bodhi/server/validators.py
@@ -204,6 +204,16 @@ def validate_build_nvrs(request, **kwargs):
     for build in request.validated.get('builds') or []:  # cope with builds being None
         try:
             cache_nvrs(request, build)
+            if request.validated.get('from_tag'):
+                n, v, r = request.buildinfo[build]['nvr']
+                release = request.db.query(Release).filter(or_(Release.name == r,
+                                                               Release.name == r.upper(),
+                                                               Release.version == r)).first()
+                if release and release.composed_by_bodhi:
+                    request.errors.add(
+                        'body', 'builds',
+                        f"Can't create update from tag for release"
+                        f" '{release.name}' composed by Bodhi.")
         except ValueError:
             request.validated['builds'] = []
             request.errors.add('body', 'builds', 'Build does not exist: %s' % build)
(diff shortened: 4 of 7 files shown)


bodhi/server/services/updates.py
```diff
@@ -536,6 +538,23 @@ def new_update(request):
 
             if len(releases) > 1:
                 result = dict(updates=updates)
+
+            if from_tag:
+                koji = buildsys.get_session()
+                # Validate that <koji_tag>-pending-signing and <koji-tag>-testing exists,
+                # if not create them
+                koji_tag_pending_signing = from_tag + "-pending-signing"
+                koji_tag_testing = from_tag + "-testing"
```
nphilipp: I'm not sure if we want to hardcode "-pending-signing" and "-testing" or put it in configuration.

@cverna, @abompard?

bodhi/server/services/updates.py
```diff
@@ -536,6 +538,23 @@ def new_update(request):
...
+                if not koji.getTag(koji_tag_pending_signing):
+                    koji.createTag(koji_tag_pending_signing, parent=from_tag)
+                if not koji.getTag(koji_tag_testing):
+                    koji.createTag(koji_tag_testing, parent=from_tag)
+
+                to_tag = from_tag + "-pending-signing"
```
nphilipp:
```
    to_tag = koji_tag_pending_signing
```

:wink:

Zlopez:
Rebased and addressed comments from @nphilipp

bodhi/server/buildsys.py
```diff
@@ -448,7 +449,7 @@ def getTag(self, taginfo, **kw):
                 'perm': None, 'id': 246, 'arches': None,
                 'maven_include_all': False, 'perm_id': None}
 
-    def addTag(self, tag: str, **opts):
+    def createTag(self, tag: str, **opts):
```
cverna, line 491: I think we still want `addTag` here. I would have assumed that this is used in some test.

bodhi/server/services/updates.py
```diff
@@ -536,6 +538,24 @@ def new_update(request):
...
+                    'koji_pending_signing_side_tag')
+                koji_tag_testing = from_tag + config.config.get('koji_testing_side_tag')
+                if not koji.getTag(koji_tag_pending_signing):
+                    koji.createTag(koji_tag_pending_signing, parent=from_tag)
+                if not koji.getTag(koji_tag_testing):
+                    koji.createTag(koji_tag_testing, parent=from_tag)
+
+                to_tag = koji_tag_pending_signing
+                # Move every new build to <from_tag>-pending-signing tag
+                for build in builds:
```
cverna: I think you could use the update `add_tag` method here (https://github.com/fedora-infra/bodhi/blob/develop/bodhi/server/models.py#L2845).

bodhi/server/buildsys.py
Zlopez: It's not, I checked the whole project before doing this change.
I did it, because the koji API doesn't have any method `addTag`.

bodhi/server/services/updates.py
Zlopez: I didn't noticed that update has `add_tag` method.

Zlopez:
Use `update.add_tag` instead on `koji.moveBuild`.

bodhi/server/config.py
```diff
@@ -377,6 +377,12 @@ class BodhiConfig(dict):
         'koji_hub': {
             'value': 'https://koji.stg.fedoraproject.org/kojihub',
             'validator': str},
+        'koji_pending_signing_side_tag': {
```
cverna: We have per release configuration settings for example (https://github.com/fedora-infra/bodhi/blob/develop/production.ini#L430).

I would feel better if this was something that we configure only for a release (rawhide in that case). That will give us less chances to break things :smile:

bodhi/server/config.py
```diff
@@ -377,6 +377,12 @@ class BodhiConfig(dict):
...
+            'value': '-pending-signing',
+            'validator': str},
+        'koji_testing_side_tag': {
```
cverna: same here

bodhi/server/config.py
Zlopez: Could I set default for those settings?

Zlopez: It looks like I can set defaults for `rawhide.koji_pending_signing_side_tag`, so I will change it.

cverna: Do we need defaults tho ? if nothing is configured then we should not do anything. What do you think ?

Zlopez: I think it is better to have default value then checking, if the configuration exists every time you want to work with it.

cverna: I don't have a strong opinion on that, but if we want a default then let's use something similar to the mandatory_days_in_testing. (ie use fedora.your_setting_here instead of rawhide).

See https://github.com/fedora-infra/bodhi/blob/1a48548d911566fb54a187f41bc74c61f4430572/bodhi/server/models.py#L861

I think longer term we might want to be able to create updates from side tag for more than rawhide.

Zlopez: This looks good. I will add something similar for side tags.

bodhi/server/models.py
```diff
@@ -977,6 +977,34 @@ def setting_status(self) -> typing.Optional[str]:
         """
         return config.get(f'{self.setting_prefix}.status', None)
 
+    def get_testing_side_tag(self, from_tag: str) -> str:
```
cverna, line 985: Nice :+1:

bodhi/server/services/updates.py
```diff
@@ -536,6 +538,23 @@ def new_update(request):
 
             if len(releases) > 1:
                 result = dict(updates=updates)
+
+            if from_tag:
+                koji = buildsys.get_session()
+                for update in updates:
+                    # Validate that <koji_tag>-pending-signing and <koji-tag>-testing exists,
+                    # if not create them
+                    koji_tag_pending_signing = update.release.get_pending_signing_side_tag(from_tag)
```
cverna: Nitpick here but I think it makes more sense to call this side_tag rather than koji_tag.

```suggestion
                    side_tag_pending_signing = update.release.get_pending_signing_side_tag(from_tag)
```

bodhi/server/services/updates.py
```diff
@@ -536,6 +538,23 @@ def new_update(request):
...
+                    koji_tag_testing = update.release.get_testing_side_tag(from_tag)
```
cverna:
```suggestion
                    side_tag_testing = update.release.get_testing_side_tag(from_tag)
```

cverna:
Just 2 smalls changes and I think this will be ready :smiley:

bodhi/server/services/updates.py
```diff
@@ -536,6 +538,23 @@ def new_update(request):
 
             if len(releases) > 1:
                 result = dict(updates=updates)
+
+            if from_tag:
```
cverna, line 542: Wondering if we should restrict that to the release that are not composed by bodhi (ie rawhide). I don't think we want to allow create update from side tag for all releases yet.
@pypingou  what do you think ?

pypingou: >  I don't think we want to allow create update from side tag for all releases yet.

This sounds wise, let's start with rawhide only, it should be fairly straight forward to change this later on. This would allow using chain-build in stable releases which is a side-effect of gating rawhide packages that we are interested in, but I'd rather we do this in small steps :)

cverna: @Zlopez then to do that you will need to check that the update release is not composed by bodhi. You can check here how it was done --> https://github.com/fedora-infra/bodhi/blob/develop/bodhi/server/scripts/approve_testing.py#L116

Zlopez: Ok, I will update it

Zlopez:
It looks like plenty of pip tests are now failing. According to the output, it looks like this is caused by change in mock module and some of the tests are not returning expected output.

Zlopez:
Another issue I see is related to `sqlalchemy`, I'm not sure what exactly was changed, but the error is rather confusing one `LookupError: "testing" is not among the defined enum value"`

But looking at the Enum it doesn't make sense:
```
self = Enum('testing', 'obsolete', 'unpush', 'revoke', 'stable', name='ck_update_request')
elem = <testing>
```

bodhi/server/models.py
```diff
@@ -982,6 +982,34 @@ def setting_status(self) -> typing.Optional[str]:
...
+        Return the testing side tag for this ``Release``.
+
+        Args:
+            from_tag: Name of side tag from which ``Update`` was created.
+
+        Returns:
+            Testing side tag used in koji.
+        """
+        side_tag_postfix = config.get(
+            '{}.koji-testing-side-tag'.format(self.setting_prefix), "-testing")
```
cverna: should we use f-strings here ?

bodhi/server/models.py
```diff
@@ -982,6 +982,34 @@ def setting_status(self) -> typing.Optional[str]:
...
+        Return the testing side tag for this ``Release``.
+
+        Args:
+            from_tag: Name of side tag from which ``Update`` was created.
+
+        Returns:
+            Testing side tag used in koji.
+        """
+        side_tag_postfix = config.get(
+            '{}.koji-pending-signing-side-tag'.format(self.setting_prefix), "-pending-signing")
```
cverna: should we use f-strings here ?

bodhi/server/services/updates.py
```diff
@@ -536,6 +538,25 @@ def new_update(request):
...
+                koji = buildsys.get_session()
+                for update in updates:
+                    if update.release.composed_by_bodhi:
```
cverna: we want to do the following if the release is **not** composed by bodhi.

Zlopez: I thought you want to create it if it is composed by bodhi. My mistake, I will fix it.

bodhi/server/models.py
Zlopez: Do we have some standard here? I saw `.format()` in some places, `%s` in others and I also saw f-string somewhere.

cverna: we are slowly moving to f-strings, I think all new code should be using f-string.

Zlopez: Fixed

bodhi/server/models.py
Zlopez: Fixed

bodhi/server/services/updates.py
Zlopez: Fixed

Zlopez:
@cverna Could you look at this again

cverna:
> @cverna Could you look at this again

Yes, so as mentioned during today's call we should probably check if the release is not composed by bodhi in a validator. If a user tries to create an update using from_tag on a stable release bodhi should not create the update and let the user know that this is currently not supported.

cverna (commented):
LGTM, I ll let squash your commit before approving the PR

cverna (approved):

PR #3483, opened 2019-08-23
determine a release for builds in a sidetag update

### LINKED ISSUE #3480: Unable to create a sidetag update: no release associated with build
When trying to create a new update based on a sidetag, it fails, with Bodhi unable to associate a release to the sidetag.

Specifically, the error is:

    Cannot find release associated with build: adwaita-blue-gtk-theme-3.22.21.0-10.fc30, tags: ['f30-build-side-7749']

This is being generated by https://github.com/fedora-infra/bodhi/blob/develop/bodhi/server/validators.py#L158

Also, note that the tests for using the `from_tag` field in  server/services/update.py 's new_update do work, but only because it is testing using a valid candidate tag which bodhi can determine a release for.

 Possibly related to: #3473

ryanlerch:
Previously, when trying to create an update from a side tag,
it would fail, as bodhi was unable to determine the release
for build(s) for that sidetag.

This commit adds code to the validate_from_tag validator to
pull the inheritance data from koji of a sidetag, and puts it
into a new per-request cache item in the request (`from_tag_inherited`)
One or more of the tags in a sidetags inhertiance structure will match
with one of the tags defined in the Bodhi Release.

Then, when validating all the builds to determine the releases, this
list is added to the regular tags to help determine the bodhi release
of the build.

Fixes: #3480

### DIFF
--- a/bodhi/tests/server/services/test_updates.py
+++ b/bodhi/tests/server/services/test_updates.py
@@ -285,13 +285,13 @@ class TestNewUpdate(BaseTestCase):
         # We don't want the new update to obsolete the existing one.
         self.db.delete(Update.query.one())
 
-        update = self.get_update(builds=None, from_tag='f17-updates-candidate')
+        update = self.get_update(builds=None, from_tag='f17-build-side-7777')
         with fml_testing.mock_sends(update_schemas.UpdateRequestTestingV1):
             r = self.app.post_json('/updates/', update)
 
         up = r.json_body
-        self.assertEqual(up['title'], 'TurboGears-1.0.2.2-3.fc17')
-        self.assertEqual(up['builds'][0]['nvr'], 'TurboGears-1.0.2.2-3.fc17')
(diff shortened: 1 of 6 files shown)


bodhi/tests/server/services/test_updates.py
```diff
@@ -285,13 +285,13 @@ def test_new_rpm_update_from_tag(self, *args):
         # We don't want the new update to obsolete the existing one.
         self.db.delete(Update.query.one())
 
-        update = self.get_update(builds=None, from_tag='f17-updates-candidate')
+        update = self.get_update(builds=None, from_tag='f17-build-side-7777')
```
Zlopez, line 288: These changes will probably conflict with https://github.com/fedora-infra/bodhi/pull/3481, it will be nice to merge this first

cverna:
I had a look and it looks good to me, but I ll appreciate if someone else gives it another review :smiley:

Zlopez (approved):
Looks good to me

PR #3487, opened 2019-08-27
add support for creating sidetag updates to webui

### LINKED ISSUE #2324: [S] The UI needs a way for users to create side tag updates
Once the [API can create side tags](https://github.com/fedora-infra/bodhi/issues/3009), we should rework the UI so that users can create side tag updates or create traditional updates. We might be able to just add a new entry to the "Create" menu in the top bar that creates a side tag update, while reusing some of the existing create update form, but presenting a few different choices. For example, a side tag update should prompt the user for a release and a display name, and should not prompt the user for builds.

ryanlerch:
This adds the ability for a user to create a new
update, using a side-tag rather than specifying
a list of builds.

Resolves: #2324

This one is also dependent on getting #3483 merged first, so filing this as a WIP.

### DIFF
--- a/bodhi/server/__init__.py
+++ b/bodhi/server/__init__.py
@@ -307,6 +307,8 @@ def main(global_config, testing=None, session=None, **settings):
     config.add_route('search_packages', '/search/packages')
     config.add_route('latest_candidates', '/latest_candidates')
     config.add_route('latest_builds', '/latest_builds')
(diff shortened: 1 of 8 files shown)


Zlopez (commented):
Looks good, but I see that some tests are failing

Zlopez (approved):
Looks good to me

Issue #3476, opened 2019-08-19
Move successfully tested update from testing to stable and remove side and aux tags.


nphilipp:
# Story

As a user of multi-build gating,

when the tests of an update created for a side tag are finished successfully (i.e. an according gating decision has been announced on the message bus), I want Bodhi to:

* add a comment documenting the decision to the update
* tag its builds into the stable tag and out of the `*-testing` tag of the side tag
* change the status of an update created for a side-tag to stable
* and remove the side tag and its auxiliary tags

so that the update and the build artifacts contained in it are made available for the general public.

# Background

pypingou's multi-build update workflow diagram \[[download](https://user-images.githubusercontent.com/820624/63262799-34940480-c287-11e9-9594-e177ce7e6d4d.png)\]:

![Multi-builds - Rawhide Package Gating Flow.png](https://user-images.githubusercontent.com/820624/63262799-34940480-c287-11e9-9594-e177ce7e6d4d.png)

This card corresponds to the bottom-most bubble in the Bodhi column.

NB: This diagram is about the Rawhide workflow. Other inputs are going into this last step, like pushing the update because it spent enough time in testing (`Update.autotime`, which happens to default to 0 in Rawhide), or because it has gathered enough karma (`Update.autokarma`).

nphilipp:
Not sure how far I'll get with this before going on vacation but I'll give it a shot.

nphilipp:
After my first implementation attempt and some discussion, it seems that the first three items work already, in the greenwave consumer and the approve_testing script. I've implemented the last item in the script, too: PR #3489 which is WIP right now because there are no tests yet.

PR #3489, opened 2019-08-28
Clear side and aux tags on move to stable

nphilipp:
This implements what's missing for #3476, i.e. re-/untagging the builds and removing the side tag and its auxiliary tags (-pending-signing and -testing).

~~I haven't gotten around to writing tests for it yet, and~~ the pattern for the aux tags (`<sidetag>-pending-signing` and `<sidetag>-testing`) is hardcoded, not configurable (not sure if we need this). ~~This is why it has the WIP label, which can be removed once more testing and lookovers went over this than I can do before going on vacation tomorrow.~~

The first two commits are for (IMO) minor cleanups and test out well in unit tests (integration tests still ongoing). ~~I'll see if I can add some tests for it before going on vacation.~~ Please "do the needful" while I'm away, but look at it thoroughly because my eyes are beginning to glaze over. :wink:

### DIFF
--- a/bodhi/server/buildsys.py
+++ b/bodhi/server/buildsys.py
@@ -17,12 +17,12 @@
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
 """Define tools for interacting with the build system and a fake build system for development."""
 
-from threading import Lock
 import logging
 import time
 import typing
 import os
 from functools import wraps
+from threading import Lock
 
 import backoff
 import koji
@@ -504,9 +504,19 @@ class DevBuildsys:
         opts['perm_id'] = 1
         self.__tags__.append((tag, opts))
 
-    def deleteTag(self, tagid):
+    def deleteTag(self, tagid: typing.Union[str, int]):
         """Emulate tag deletion."""
-        del self.__tags__[tagid]
+        if isinstance(tagid, str):
+            for tid, tinfo in self.__tags__:
+                if tagid == tid:
+                    self.__tags__.remove((tid, tinfo))
+                    return
+        else:
+            del self.__tags__[tagid]
+
+    def removeSideTag(self, sidetag):
+        """Emulate side tag and build target deletion."""
+        pass
 
     def getRPMHeaders(self, rpmID: str,
                       headers: typing.Any) -> typing.Union[typing.Mapping[str, str], None]:
--- a/bodhi/server/scripts/approve_testing.py
+++ b/bodhi/server/scripts/approve_testing.py
@@ -22,12 +22,12 @@ The script is responsible for commenting on updates after they reach the mandato
 spent in the testing repository.
 """
 
-import datetime
 import os
 import sys
 import logging
 
 from pyramid.paster import get_appsettings
+from sqlalchemy import func
 
 from bodhi.messages.schemas import update as update_schemas
 from bodhi.server import Session, initialize_db, notifications, buildsys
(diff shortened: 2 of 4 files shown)


nphilipp:
The latest push tests the new codepaths, removing WIP.

bodhi/server/scripts/approve_testing.py
```diff
@@ -51,6 +52,20 @@ def usage(argv):
     sys.exit(1)
 
 
+def clear_side_aux_tags(client_session, update):
+    for tag in (
+        update.from_tag + '-pending-signing',
```
Zlopez: I made them configurable in #3481, you can call it like this `update.release.get_pending_signing_side_tag(update.from_tag)`.
Similar for testing side tag `update.release.get_testing_side_tag(update.from_tag)`.

Configuration is done as for any release specific config:
`rawhide.koji-testing-side-tag`
`rawhide.koji-pending-signing-side-tag`

Zlopez:
Updated PR with changes from #3481

bodhi/server/buildsys.py
```diff
@@ -92,6 +93,8 @@ class DevBuildsys:
                    'tag_id': 19,
                    'task_id': 127621}
 
+    _side_tag_re = re.compile(r'(?P<base>.*)-side-(?P<id>[0-9]+)')
```
pypingou: I'm not seeing this being used anywhere, am I missing something?

cverna: Good catch :-) removed

bodhi/server/buildsys.py
```diff
@@ -506,7 +506,16 @@ def createTag(self, tag: str, **opts):
 
     def deleteTag(self, tagid):
         """Emulate tag deletion."""
-        del self.__tags__[tagid]
+        if isinstance(tagid, str):
+            for tid, tinfo in self.__tags__:
+                if tagid == tid:
```
pypingou, line 511: Aren't we comparing string to int here?

bodhi/server/buildsys.py
```diff
@@ -506,7 +506,16 @@ def createTag(self, tag: str, **opts):
 
     def deleteTag(self, tagid):
         """Emulate tag deletion."""
```
pypingou, line 508: Should we adjust the docstring to specify the tagid can be either a string or an int and the behavior in each case?

bodhi/server/scripts/approve_testing.py
```diff
@@ -110,18 +111,43 @@ def main(argv=sys.argv):
 
                 if update.autotime and update.days_in_testing >= update.stable_days:
                     print(f"Automatically marking {update.alias} as stable")
-                    update.set_request(db=db, action=UpdateRequest.stable, username="bodhi")
+                    # For now only rawhide update can be created using side tag
+                    # Do not ad the release.pending_stable_tag if the update
```
pypingou: ad -> add

pypingou: I'm also not seeing where we're checking that the update was not created from a side tag

bodhi/server/scripts/approve_testing.py
```diff
@@ -110,18 +111,43 @@ def main(argv=sys.argv):
...
+
+                            update.remove_tag(pending_signing_tag)
+                            update.remove_tag(testing_tag)
+                            update.remove_tag(update.from_tag)
+
+                            koji = buildsys.get_session()
+                            koji.deleteTag(pending_signing_tag)
+                            koji.deleteTag(testing_tag)
+
+                            koji.removeSideTag(update.from_tag)
```
pypingou, line 146: Hm, should we document why we use deleteTag() in the first two and removeSideTag() in the third one?

pypingou (commented):
A few comments, mostly nickpick it looks good overall

bodhi/server/scripts/approve_testing.py
cverna: Isn't the name of the method enough ? `pending_signing_tag` and `testing_tag` are tags and `update.from_tag` is a side tag.

bodhi/server/scripts/approve_testing.py
cverna: At the update creation we check that only update against a release that is not composed by bodhi can be created by side tag. So here it is enough to check that the release is composed by bodhi or not.

bodhi/server/buildsys.py
cverna: This is the dummy build system, and in that case `self.__tags__` is a list of string tuples. So we should compare string with string.

bodhi/server/buildsys.py
cverna: will do

bodhi/server/scripts/approve_testing.py
pypingou: Is it worth mentioning this in the comment then? (My question was triggered by it)

bodhi/server/scripts/approve_testing.py
cverna: @mohanboddu do you know if there is an actual difference in Koji between side tags and tags ? Could we just use deleteTag to delete a side tag ?

cverna: Discussed on IRC the removeSideTag API also remove the build target on koji side. I have added a comment to make this explicit.

pypingou (approved):
Thanks :)

PR #3491, opened 2019-08-30
Fix enums with SQLAlchemy >= 1.3.8

abompard:
This commit fixes our Enum implementation for SQLAlchemy 1.3.8+

See [this changelog entry](https://docs.sqlalchemy.org/en/14/changelog/changelog_13.html#change-ac119f6307026142f7a0ccbf81065f25) for details.

### DIFF
--- a/bodhi/server/models.py
+++ b/bodhi/server/models.py
@@ -221,7 +221,11 @@ class DeclEnumType(SchemaType, TypeDecorator):
         self.enum = enum
         self.impl = Enum(
             *enum.values(),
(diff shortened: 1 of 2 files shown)


pypingou (approved):
nice catch :)

Zlopez (approved):
It looks like it was only small change

abompard:
Yeah it's a very small change but it took me a while to pin it down.

Zlopez:
Could we run tests again? It looks like there was some issue with duffy node

Issue #3061, opened 2019-03-14
[L] - Koji calls should be moved to a backend task worker


bowlofeggs:
Creating or editing updates with many builds takes a long time. Even 50 builds can take an extreme about of time, to the point of causing API timeouts. There is likely more than one cause for this (indices come to mind), but one item that comes to mind that would help us is to move Koji API calls from the web server to backend task workers instead.

This would help with another problem too: right now we make our Koji API calls in the middle of a database transaction. If the database transaction is rolled back after Koji write calls were made, nothing will roll back those Koji calls. I believe this happened yesterday with https://bodhi.fedoraproject.org/updates/FEDORA-EPEL-2019-9e9f81e581 - it seems that Bodhi untagged one of its builds from being a buildroot override (https://bodhi.fedoraproject.org/overrides/python3-cryptography-2.3-2.el7), but for some reason rolled back the transaction after doing this. This caused Bodhi to be in an inconsistent state because it recorded the override as being active even though it was untagged in Koji. If we instead waited to commit the transaction before queueing tasks to do backend work like this, we could ensure that we have better consistency. Of course, we'd need good error handling and state tracking around making sure that the Koji calls make it to Koji (otherwise we also end up being inconsistent), but that seems like an easier place to deal with it than what we do today. Plus, it'll make the web API faster.

I am putting this on the CI Gating project kanban board because we will have much larger updates with Rawhide Gating than we typically do today, and this will be a performance bottleneck if we don't solve it.

We need to finish #2851 before we can work on this issue.

cverna:
Scope
--------

Move the write calls to koji in a celery tasks to benefit from celery retry features. We don't want to queue the celery task until the database transaction is successfully committed.

Notes : We currently have a mechanism in place for this for sending fedora-messaging. There might be possibility to reuse this. (Sqlalchemy trigger/hook)

nphilipp:
## Story Time…

As a package maintainer creating or editing updates with many builds, …

1. I want that the Bodhi UI/API doesn't block on the related Koji API calls,

    so that I don't run into timeouts.

2. I want that related Koji API calls are only performed after the transaction has been committed to the database,

    to avoid inconsistencies between Bodhi and Koji.

## Acceptance Criteria

- [ ] A Celery task worker exists which performs the changes related to a newly created or edited update in Koji.
- [ ] Creating an update or editing it doesn't perform the Koji changes in a blocking fashion, but delegates them to the task worker after the transaction has been committed to the database.

[...]

PR #3498, opened 2019-09-11
Check if it is safe to tag a multi build update in stable.

cverna:
This commit add the logic to check if a multi build update
has builds that conflicts with the stable tag. To do that we
are checking for each package present in the update if a more
recent build exists in the stable tag. If that is the case we
consider that this package build is conflicting and the update
will not be pushed to stable.

If none of the build present in the update are conflicting then
we proceed with pushing the update to stable.

### DIFF
--- a/bodhi/server/models.py
+++ b/bodhi/server/models.py
@@ -1458,6 +1458,10 @@ class Build(Base):
         """
         return self._get_kojiinfo()['id']
 
+    def get_creation_time(self) -> datetime:
+        """Return the creation time of the build."""
+        return datetime.fromisoformat(self._get_kojiinfo()['creation_time'])
+
     def unpush(self, koji):
         """
         Move this build back to the candidate tag and remove any pending tags.
@@ -1482,6 +1486,21 @@ class Build(Base):
                     'Moving %s from %s to %s' % (
                         self.nvr, tag, release.candidate_tag))
                 koji.moveBuild(tag, release.candidate_tag, self.nvr)
+
+    def is_latest(self) -> bool:
+        """Check if this is the latest build available in the stable tag."""
+        koji_session = buildsys.get_session()
+        # Get the latest builds in koji in a tag for a package
+        koji_builds = koji_session.getLatestBuilds(
+            self.update.release.stable_tag,
+            package=self.package.name
+        )
+
+        for koji_build in koji_builds:
+            build_creation_time = datetime.fromisoformat(koji_build['creation_time'])
+            if self.get_creation_time() < build_creation_time:
+                return False
+        return True
 
 
 class ContainerBuild(Build):
@@ -2924,6 +2943,19 @@ class Update(Base):
             koji.untagBuild(tag, build.nvr, force=True)
         if return_multicall:
             return koji.multiCall()
+
+    def find_conflicting_builds(self) -> list:
+        """
+        Find if there are any builds conflicting with the stable tag in the update.
+
+        Returns a list of conflicting builds, empty is none found.
+        """
+        conflicting_builds = []
+        for build in self.builds:
+            if not build.is_latest():
+                conflicting_builds.append(build.nvr)
+
+        return conflicting_builds
 
     def modify_bugs(self):
         """
--- a/bodhi/tests/server/scripts/test_approve_testing.py
+++ b/bodhi/tests/server/scripts/test_approve_testing.py
@@ -915,6 +915,41 @@ class TestMain(BasePyTestCase):
         assert update.request is None
         assert update.autotime == False
 
+    @patch("bodhi.server.buildsys.DevBuildsys.getLatestBuilds", return_value=[{
+        'creation_time': '2007-08-25 19:38:29.422344'}])
+    def test_update_conflicting_build_not_pushed(self, build_creation_time):
+        """
+        Ensure that an update that have conflicting builds will not get pushed.
+        """
+        update = self.db.query(models.Update).all()[0]
+        update.autokarma = False
+        update.autotime = True
+        update.request = None
+        update.stable_karma = 1
+        update.stable_days = 7
+        update.date_testing = datetime.utcnow() - timedelta(days=8)
+        update.status = models.UpdateStatus.testing
+        update.release.composed_by_bodhi = False
+
+        # Clear pending messages
+        self.db.info['messages'] = []
+        self.db.commit()
+
+        with patch('bodhi.server.scripts.approve_testing.initialize_db'):
+            with patch('bodhi.server.scripts.approve_testing.get_appsettings', return_value=''):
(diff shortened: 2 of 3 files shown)


bodhi/server/models.py
```diff
@@ -1483,6 +1487,20 @@ def unpush(self, koji):
                         self.nvr, tag, release.candidate_tag))
                 koji.moveBuild(tag, release.candidate_tag, self.nvr)
 
+    def is_latest(self) -> bool:
+        """Check if this is the latest build available in the stable tag."""
+        koji_session = buildsys.get_session()
+        builds = koji_session.getLatestBuilds(
```
Zlopez: Will this return every build that is not in stable yet?

bodhi/server/models.py
```diff
@@ -1483,6 +1487,20 @@ def unpush(self, koji):
...
+            self.update.release.stable_tag,
+            package=self.package.name
+        )
+
+        for build in builds:
+            build_creation_time = datetime.fromisoformat(build['creation_time'])
```
Zlopez: This will be probably better :-)
```suggestion
            build_creation_time = build.get_creation_time()
```

Zlopez (commented):
Clear and readable :+1: Just a few comments

bodhi/server/models.py
cverna: No that will return the latest builds in the stable tag (ie f31) for the specified package name. It is a more efficient way to know if there is a build in the stable tag, rather than getting all the builds tagged with f31.

bodhi/server/models.py
cverna: That would be neat but here build is a dictionary retruned by koji and not a Build type from Bodhi, the name are confusing here so I might rename it in koji_build.

bodhi/server/models.py
Zlopez: Nice to know

bodhi/server/models.py
Zlopez: Yeah, it's confusing

pypingou:
I like the code, it is looking quite nice I have a couple of comments though:
a) should we move this to an async process? side-tags and thus updates with 100s of builds will take a certain amount of time to query koji
b) two runs of the cron job will result in two comments no? Should we try to prevent this?

cverna:
> I like the code, it is looking quite nice I have a couple of comments though:
> a) should we move this to an async process? side-tags and thus updates with 100s of builds will take a certain amount of time to query koji

I think this covered by https://github.com/fedora-infra/bodhi/issues/3061.

> b) two runs of the cron job will result in two comments no? Should we try to prevent this?

Indeed that will results in a lot of comments :-(, I ll check how we can avoid that

pypingou:
> > I like the code, it is looking quite nice I have a couple of comments though:
> > a) should we move this to an async process? side-tags and thus updates with 100s of builds will take a certain amount of time to query koji
 
> I think this covered by #3061.

Indeed it is, though since we're adding a call here, should we get started on it? Or would we rather keep this all in ticket #3061 ?

cverna:
> > > I like the code, it is looking quite nice I have a couple of comments though:
> > > a) should we move this to an async process? side-tags and thus updates with 100s of builds will take a certain amount of time to query koji
> 
> > I think this covered by #3061.
> 
> Indeed it is, though since we're adding a call here, should we get started on it? Or would we rather keep this all in ticket #3061 ?

I would prefer to keep it out of the scope, to keep this PR simple.

pypingou:
> I would prefer to keep it out of the scope, to keep this PR simple.

wfm

bodhi/tests/server/scripts/test_approve_testing.py
```diff
@@ -915,6 +915,41 @@ def test_autotime_update_no_autokarma_negative_karma_not_pushed(self):
         assert update.request is None
         assert update.autotime == False
 
+    @patch("bodhi.server.buildsys.DevBuildsys.getLatestBuilds", return_value=[{
+        'creation_time': '2007-08-25 19:38:29.422344'}])
+    def test_update_conflicting_build_not_pushed(self, build_creation_time):
+        """
+        Ensure that an autotime update which negative karma does not get pushed.
```
pypingou: Looks like we forgot to update this one :)

bodhi/tests/server/scripts/test_approve_testing.py
```diff
@@ -915,6 +915,41 @@ def test_autotime_update_no_autokarma_negative_karma_not_pushed(self):
...
+                    approve_testing.main(['nosetests', 'some_config.ini'])
+
+        assert update.status == models.UpdateStatus.testing
+
+        bodhi = self.db.query(models.User).filter_by(name='bodhi').one()
+        cmnts = self.db.query(models.Comment).filter_by(update_id=update.id, user_id=bodhi.id)
+        assert cmnts.count() == 2
+        assert cmnts[0].text == config.get('testing_approval_msg')
+        assert cmnts[1].text == "This update cannot be pushed to stable. "\
+            f"These builds bodhi-2.0-1.fc17 have a more recent build in koji's "\
```
pypingou: you don't need to make this line a f-string since there is no variable in it :)

pypingou (commented):
One docstring we should update
One nitpick that is up to you

One question about the tests though: I only see one patch for the setting `creation_time` we get from koji, shouldn't there be two?

cverna:
> One docstring we should update
> One nitpick that is up to you

Good catch :smile: 

> 
> One question about the tests though: I only see one patch for the setting `creation_time` we get from koji, shouldn't there be two?

We need only one because the pseudo buildsystem we use returns the same value for getLatestBuilds and getBuild. So to make sure that the creation_time are different I really only need to mock the returned value of one of these call.

If you think that mocking both call would make it clearer, I am happy to do it.

pypingou:
>>  One question about the tests though: I only see one patch for the
>> setting creation_time we get from koji, shouldn't there be two?

> We need only one because the pseudo buildsystem we use returns the same
> value for getLatestBuilds and getBuild. So to make sure that the
> creation_time are different I really only need to mock the returned value
> of one of these call.
> 
> If you think that mocking both call would make it clearer, I am happy to
> do it.

Let's not mock something that doesn't need to be, but maybe we could mention it
in the docstring or a comment?

pypingou:
Code still looks fine to me but:
- Let's open a ticket to trac the issue of adding a comment everytime the cron runs
- Jenkins seems less happy than me :(

cverna:
> Code still looks fine to me but:
> 
>     * Let's open a ticket to trac the issue of adding a comment everytime the cron runs
> 
https://github.com/fedora-infra/bodhi/issues/3514
>     * Jenkins seems less happy than me :(

I think this should clear tomorrow once we have a new db dump.

pypingou (approved):

PR #3504, opened 2019-09-13
Document what the update states mean for rawhide

pypingou:


docs/user/update_states.rst
```diff
@@ -26,6 +26,13 @@ Once submitted to Bodhi, updates move through the following states:
 
 :ref:`unpushed`: The update has been removed from testing.
 
+.. note:: | In the case of Fedora rawhide, the states described here are a little bit changed
+           as rawhide has only two states (testing, stable).
```
cverna, line 30: This is no longer true for multi build update, since we have pending. Pending is the state where the updates wait for all the builds to be signed. We likely want to unify the single build update to do the same thing.

pypingou: :+1: for this, do we want to update the doc to the current or the future state?

cverna: Let's go with the current state and let's try to update that once we modify the single build update :-)

cverna (approved):
:+1:

PR #3509, opened 2019-09-19
Fix integration tests on Rawhide

abompard:
This should fix the integration test issues we're having on Rawhide.


ryanlerch (approved):
Looks good!

PR #3516, opened 2019-09-25
allow only 1 update per side tag

### LINKED ISSUE #3484: Should we stop a user from creating two sidetag updates with the same sidetag
Currently, it is possible for a user to create two different updates from the same sidetag. This may be an edge-case, as if the builds are the same when creating the update, bodhi will not allow the update to be created, as the builds validator detects an update already has those builds.

However, it is possible if a user pushes new builds to a sidetag, to create a new update from the same sidetag.

The discussion here is should we update the validators to determine if any existing update has the same from_tag value, and block the update from being created?

ryanlerch:
previously, a user was able to create an update
from a sidetag that was already used in an existing
update. this commit adds a validation check to make
sure a user cannot create an update from a side tag
that is already used in an update.

fixes: #3484

### DIFF
--- a/bodhi/server/validators.py
+++ b/bodhi/server/validators.py
@@ -1291,6 +1291,18 @@ def validate_from_tag(request: pyramid.request.Request, **kwargs: dict):
     koji_tag = request.validated.get('from_tag')
 
     if koji_tag:
+        # check if any existing updates use this side tag
+        update = request.db.query(Update).filter_by(from_tag=koji_tag).first()
+        if update:
+            if request.validated.get('edited') == update.alias:
+                # existing update found, but it is the one we are editing, so keep going
+                pass
+            else:
+                request.errors.add('body', 'from_tag', "Update already exists using this side tag")
+                # don't run any more validators
+                request.validated = []
+                return
+
         koji_client = buildsys.get_session()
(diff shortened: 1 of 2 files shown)


bodhi/server/validators.py
```diff
@@ -1291,6 +1291,18 @@ def validate_from_tag(request: pyramid.request.Request, **kwargs: dict):
     koji_tag = request.validated.get('from_tag')
 
     if koji_tag:
+        # check if any existing updates use this side tag
+        update = request.db.query(Update).filter_by(from_tag=koji_tag).first()
+        if update:
+            if request.validated['edited'] and request.validated['edited'] == update.alias:
```
pypingou: Is ``request.validated`` a dict?
If so, you could simplify this as:
````
if request.validated.get('edited') == update.alias:
````

If there is no `'edited'` in the `request.validated`, it'll be `None` and thus not equal to the alias

pypingou (commented):
One comment, otherwise it's nice and straight forward :)

bodhi/server/validators.py
ryanlerch: thanks! commit updated.

pypingou (approved):
Nice! :)

PR #3517, opened 2019-09-25
turn off autokarma and autotime for automatic updates

### LINKED ISSUE #3424: Automatic rawhide updates should not list karma limits
I think that the highlighted part of an update overview page should not be visible in an automated rawhide update:

ryanlerch:
having autokarma and autotime does not really make
much sense for automatic updates, so this turns it off.

Additionally, the recently merged UI refresh hides some
items if they are turned off, making the ui cleaner for
the user.

Fixes: #3424

### DIFF
--- a/bodhi/server/consumers/automatic_updates.py
+++ b/bodhi/server/consumers/automatic_updates.py
@@ -156,6 +156,7 @@ class AutomaticUpdateHandler:
                 type=UpdateType.unspecified,
                 stable_karma=3,
                 unstable_karma=-3,
+                autokarma=False,
                 user=user,
                 status=UpdateStatus.testing,
             )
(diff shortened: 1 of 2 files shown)


bodhi/server/consumers/automatic_updates.py
```diff
@@ -156,6 +156,8 @@ def __call__(self, message: fedora_messaging.api.Message) -> None:
                 type=UpdateType.unspecified,
                 stable_karma=3,
                 unstable_karma=-3,
+                autokarma=False,
+                autotime=False,
```
cverna: We need autotime otherwise the update will not get pushed to stable automatically. For rawhide the time threshold is set to 0 days :-).

I agree that autokarma is probably not needed

ryanlerch: okies, updated the commit, removing the autotime. thanks!

cverna (approved):
LGTM

PR #3518, opened 2019-09-25
add automatic_updates_blacklist

### LINKED ISSUE #3445: Builds coming from the releng user should not create updates in rawhide.
The releng user is used to create composes in koji. The koji builds are tagged with the f31-updates-candidate tag which makes bodhi create updates for it.

The automatic update consumer needs to check if the build comes from the `releng` user and ignore it.

@mohanboddu and @nirik also think that the build should be tagged directly to f31.

ryanlerch:
adds a new config item, automatic_updates_blacklist, which is
a list of users to not process auto updates from

fixes: #3445 

### DIFF
--- a/bodhi/server/consumers/automatic_updates.py
+++ b/bodhi/server/consumers/automatic_updates.py
@@ -91,6 +91,11 @@ class AutomaticUpdateHandler:
             log.debug(f"Koji build info for {bnvr} doesn't contain 'owner_name'.")
             return
 
+        if kbuildinfo['owner_name'] in config.get('automatic_updates_blacklist'):
+            log.debug(f"{bnvr} owned by {kbuildinfo['owner_name']} who is listed in "
+                      "automatic_updates_blacklist, skipping.")
+            return
+
         # some APIs want the Koji build info, some others want the same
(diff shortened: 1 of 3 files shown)


bodhi/server/consumers/automatic_updates.py
```diff
@@ -91,6 +91,10 @@ def __call__(self, message: fedora_messaging.api.Message) -> None:
             log.debug(f"Koji build info for {bnvr} doesn't contain 'owner_name'.")
             return
 
+        if kbuildinfo['owner_name'] in config.get('automatic_updates_blacklist'):
+            log.debug(f"{bnvr} owned by a user in automatic_updates_blacklist, skipping.")
```
pypingou: maybe we should also log the `owner_name`?

pypingou (commented):
One small suggestion, otherwise nice and simple :)

bodhi/server/consumers/automatic_updates.py
ryanlerch: done, and updated the commit! thanks!

pypingou (approved):
Nice :)

PR #3519, opened 2019-09-25
Unify rawhide simple build update with multi build update.

### LINKED ISSUE #3513: Create single build updates in pending status for rawhide
Now that we have the automation in place to automatically move an update from pending to testing we should use that for single build updates too.

For that we need to change the automatic_update consumer (https://github.com/fedora-infra/bodhi/blob/develop/bodhi/server/consumers/automatic_updates.py) to create the update in the pending status. 

And make sure that the signed consumer will move the updates from pending to testing when the build is signed.

cverna:
This commits changes how bodhi automatically creates single build
updates for rawhide. Instead of creating the update directly in the
testing status it creates it in pending. The update is then moved in
testing when the build is signed.
Once the update is moved to testing bodhi sends a fedora-messaging
message to trigger the CI tests. Results of these tests will be received
by the greenwave consumer.

Fixes #3513

### DIFF
--- a/bodhi/server/consumers/automatic_updates.py
+++ b/bodhi/server/consumers/automatic_updates.py
@@ -28,7 +28,7 @@ import fedora_messaging
 
 from bodhi.server import buildsys
 from bodhi.server.config import config
-from bodhi.server.models import Build, ContentType, Package, Release, TestGatingStatus
+from bodhi.server.models import Build, ContentType, Package, Release
 from bodhi.server.models import Update, UpdateStatus, UpdateType, User
 from bodhi.server.util import transactional_session_maker
 
@@ -105,7 +105,7 @@ class AutomaticUpdateHandler:
 
         with self.db_factory() as dbsession:
             rel = dbsession.query(Release).filter_by(create_automatic_updates=True,
-                                                     pending_testing_tag=btag).first()
+                                                     candidate_tag=btag).first()
             if not rel:
                 log.debug(f"Ignoring build being tagged into {btag!r}, no release configured for "
                           "automatic updates for it found.")
@@ -114,22 +114,7 @@ class AutomaticUpdateHandler:
             bcls = ContentType.infer_content_class(Build, kbuildinfo)
             build = bcls.get(bnvr)
             if build and build.update:
-                if build.update.status == UpdateStatus.pending:
-                    log.info(
-                        f"Build, active update for {bnvr} exists already "
-                        "in Pending, moving it along.")
-                    build.update.status = UpdateStatus.testing
-                    build.update.request = None
-                    dbsession.add(build)
-                    if config.get('test_gating.required'):
-                        log.debug(
-                            'Test gating is required, marking the update as waiting on test '
-                            'gating and updating it from Greenwave to get the real status.')
-                        build.update.test_gating_status = TestGatingStatus.waiting
-                        build.update.update_test_gating_status()
-                    dbsession.commit()
-                else:
-                    log.info(f"Build, active update for {bnvr} exists already, skipping.")
+                log.info(f"Build, active update for {bnvr} exists already, skipping.")
                 return
 
             if not build:
@@ -163,7 +148,7 @@ class AutomaticUpdateHandler:
                 unstable_karma=-3,
                 autokarma=False,
                 user=user,
-                status=UpdateStatus.testing,
+                status=UpdateStatus.pending,
             )
 
             # Comment on the update that it was automatically created.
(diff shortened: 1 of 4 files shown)


bodhi/server/consumers/automatic_updates.py
```diff
@@ -157,7 +141,7 @@ def __call__(self, message: fedora_messaging.api.Message) -> None:
                 stable_karma=3,
                 unstable_karma=-3,
                 user=user,
-                status=UpdateStatus.testing,
+                status=UpdateStatus.pending,
```
ryanlerch, line 151: does the request have to be set to testing here too? or does that not matter?

cverna: The request bit does only really matter when the release is composed by Bodhi (which is not the case for rawhide). 
Bodhi's composer which runs every night query all the updates that are in pending status with the request set to testing, the list of update that is returned will then be composed (build added to the update-testing repo). The same is true for updates in testing status with the request set to stable.

So in the case of rawhide we don't have to worry about the request.

bodhi/server/consumers/signed.py
```diff
@@ -110,7 +110,7 @@ def __call__(self, message: fedora_messaging.api.Message):
             log.info("Build %s has been marked as signed" % build_nvr)
 
             # If every build in update is signed change status to testing
-            if build.update.from_tag and build.update.signed():
+            if not build.update.release.composed_by_bodhi and build.update.signed():
                 log.info("Every build in update is signed, set status to testing")
```
pypingou, line 115: I think we should call greenwave to see if the update should remain in testing or not here.

Basically: move the update from Pending to Testing, check with greenwave immediately if the update should be gated and it not directly move it to Stable.

pypingou: Cf the lack deleted block in ``automatic_updates.py``

cverna: > I think we should call greenwave to see if the update should remain in testing or not here.
> 
> Basically: move the update from Pending to Testing, check with greenwave immediately if the update should be gated and it not directly move it to Stable.

Even if the update is no gated don't we want to wait for the tests to finishing running before moving to stable ? That way we get the same experience if an update is gated or not ( ie one will not be faster than the other)

pypingou: > Even if the update is no gated don't we want to wait for the tests to finishing running before moving to stable ? 

This assumes there will be test running which is currently not the case most often :)

If we want to minimize the impact of gating and I think if we do not want to gate a build if there is no reasons to :)

cverna: > > Even if the update is no gated don't we want to wait for the tests to finishing running before moving to stable ?
> 
> This assumes there will be test running which is currently not the case most often :)

Ok so if I read the code correctly then we don't need to call greenwave in that case, since by default update.test_gating_status is None and if the status is None we consider the tests passed (see code below). Which means that the update would be pushed to Stable the next time the cronjob runs.

https://github.com/fedora-infra/bodhi/blob/develop/bodhi/server/models.py#L2449

> 
> If we want to minimize the impact of gating and I think if we do not want to gate a build if there is no reasons to :)

cverna: > > > Even if the update is no gated don't we want to wait for the tests to finishing running before moving to stable ?
> > 
> > 
> > This assumes there will be test running which is currently not the case most often :)
> 
> Ok so if I read the code correctly then we don't need to call greenwave in that case, since by default update.test_gating_status is None and if the status is None we consider the tests passed (see code below). Which means that the update would be pushed to Stable the next time the cronjob runs.

Hum this can then be a problem if the update has tests and is gated, since the test_gating_status will be None. So I guess there is no way around having to call greenwave at that point.
> 
> https://github.com/fedora-infra/bodhi/blob/develop/bodhi/server/models.py#L2449
> 
> > If we want to minimize the impact of gating and I think if we do not want to gate a build if there is no reasons to :)

bodhi/server/consumers/signed.py
```diff
@@ -110,10 +111,17 @@ def __call__(self, message: fedora_messaging.api.Message):
...
 
                 build.update.status = UpdateStatus.testing
                 build.update.date_testing = func.current_timestamp()
                 build.update.request = None
+
+                if config.get("test_gating.required"):
```
Zlopez, line 121: Do you think it's better to have this in `signed.py` instead of `automatic_updates.py`?

Zlopez (commented):
Just one question, otherwise looks fine to me.

bodhi/server/consumers/signed.py
cverna: Since tests are now triggered only when we move the update in testing, I think it makes more sense to set the TestGatingStatus in the same place. Having in in `automatic_updates.py` would work but then we would split the logic related to moving an update to testing in 2 differents consumers.

Zlopez: Ok, this makes sense.

Zlopez (approved):
LGTM

PR #3521, opened 2019-09-26
remove critpath_karma from the UI

### LINKED ISSUE #2183: The CLI cannot leave critpath karma
The web UI can leave critpath karma, but the CLI cannot. We should bring the CLI to feature parity.

### LINKED ISSUE #2194: Consider whether to remove the critpath_karma field
Bodhi has a ```critpath_karma``` field that does not seem to be used for anything other than display in the UI. I started a [discussion](https://lists.fedoraproject.org/archives/list/devel@lists.fedoraproject.org/thread/I2KT7Z4BCYP3ZJMVAKTEV5Y2TGDECRHQ/#IEI5SNVTP4TWRMOXSK5MEIM57W4DJQSG) to find out about its history. Perhaps we can remove it to clean some code up and to make the UI less confusing.

This field is certainly confusing to users. I've observed users giving critical path karma instead of normal karma in updates, which can lead to bad results. For example, three users gave [this update](https://bodhi.fedoraproject.org/updates/FEDORA-2018-ef0b897e5d) critpath karma -1's, which were not recorded as karma on the update and thus would not stop it from going to stable automatically. IMO, the critpath karma field is harmful in this example.

If we do decide to remove it, consider whether it makes sense to keep the data in the database for history, or to drop it due to it not being useful.

ryanlerch:
for a long time, there has been a critpath_karma field
in an updates comments that users could use on critical path
updates. However, it was not used by bodhi to do anything.

This PR removes the critpath_karma from the ui, but leaves
everything else for now. see #2194 for more information.

Additionally, the bodhi cli never had support for changing
the critpath_karma.

Fixes: #2194
Fixes: #2183

### DIFF
--- a/bodhi/server/models.py
+++ b/bodhi/server/models.py
@@ -4141,6 +4141,7 @@ class Comment(Base):
     Attributes:
         karma (int): The karma associated with this comment. Defaults to 0.
         karma_critpath (int): The critpath karma associated with this comment. Defaults to 0.
+            **DEPRECATED** no longer used in the UI
         text (str): The text of the comment.
         timestamp (datetime.datetime): The time the comment was created. Defaults to
             the return value of datetime.utcnow().
--- a/bodhi/server/templates/fragments.html
+++ b/bodhi/server/templates/fragments.html
@@ -85,7 +85,7 @@ ${statusmap[status]}\
             ${comment.user.name}
           </a>
           % if comment.text:
-            % if comment.karma or comment.karma_critpath or comment.bug_feedback or any([True for feedback in comment.unique_testcase_feedback if feedback.karma]):
+            % if comment.karma or comment.bug_feedback or any([True for feedback in comment.unique_testcase_feedback if feedback.karma]):
               commented &amp; provided feedback
             % else:
               commented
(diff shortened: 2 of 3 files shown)


pypingou:
:+1: on the idea

I wonder if we should go the extra step and remove the field from the DB entirely.

@abompard is the next release going to be a X or a .Y release? How far should we go with this?

ryanlerch:
@pypingou in the thread in #2194 @AdamWill said that there might be a use for the field, but it still needs to be figured out, that is why i left it in there for now...

pypingou (approved):

bodhi/server/models.py
```diff
@@ -4141,6 +4141,7 @@ class Comment(Base):
     Attributes:
         karma (int): The karma associated with this comment. Defaults to 0.
         karma_critpath (int): The critpath karma associated with this comment. Defaults to 0.
+            **DEPRECATED** no longer used in the UI
```
pypingou, line 4149: This is going to go in as is but I was wondering if we should link to the ticket here as a reason why it's deprecated but not removed (we can address this in another PR though).

Issue #3288, opened 2019-06-04
Users should be able to re-run update tests with the CLI


tdawson:
User should be able to re-run update tests via the command line interface.
Very Minimal - All tests are re-run
Minimal - A user can choose to re-run all OpenQA, or Fedora-CI, or Taskatron tasks.
Ideal - A user can choose to either run all tests, or choose which individual tests to run.

tdawson:
A preliminary investigation has already been done concerning this. For more details on who to contact, and possible ways to implement this, please see issue #2323

The proper API's need to be in place before we can implement this.  The cannot be implemented until the OpenQA (#3284 ) Fedora-CI (#3285 ) and/or Taskotron (#3286 ) API's are implemented.    It is probably possible to implement this after just one of these API's is implemented, but not before.

Zlopez:
#3526 is only implementing the Very Minimal implementation, so I will left this one open.

pypingou:
Minimal as in: the user cannot choose which test should be re-run, they are all re-triggered.

Issue #3287, opened 2019-06-04
Users should be able to re-run update tests with the Web UI


tdawson:
User should be able to re-run update tests in the web interface.
Very Minimal - All tests are re-run
Minimal - A user can choose to re-run all OpenQA, or Fedora-CI, or Taskatron tasks.
Ideal - A user can choose to either run all tests, or choose which individual tests to run.

tdawson:
A preliminary investigation has already been done concerning this. For more details on who to contact, and possible ways to implement this, please see issue #2323

The proper API's need to be in place before we can implement this.  The cannot be implemented until the OpenQA (#3284 ) Fedora-CI (#3285 ) and/or Taskotron (#3286 ) API's are implemented.    It is probably possible to implement this after just one of these API's is implemented, but not before.

mattiaverga:
This should now be possible with Bodhi 5.7.1

Issue #3286, opened 2019-06-04
Add ability to re-run Taskotron tests on an Update


tdawson:
A user may want to re-run the Bodhi Update tests. Taskotron is one of the places that the tests are run. There is currently nothing in bodhi that allows Taskotron tests to be re-run. Implement an API to allow an update's Taskotron tests to be re-run.
Minimum - re-run all the tests.
Ideal - Have an option to re-run all tests, or just specific tests.

tdawson:
A preliminary investigation has already been done concerning this.  For more details on who to contact, and possible ways to implement this, please see issue #2323

pypingou:
Considering that taskotron is going away, I am not sure we should invest much time in this

Zlopez:
This is a part of https://github.com/fedora-infra/bodhi/issues/3288 and https://github.com/fedora-infra/bodhi/issues/3287 now. So I'm closing this.

Issue #3284, opened 2019-06-04
Add ability to re-run OpenQA tests on an Update


tdawson:
A user may want to re-run the Bodhi Update tests.  OpenQA is one of the places that the tests are run.  There is currently nothing in bodhi that allows OpenQA tests to be re-run.  Implement an API to allow an update's OpenQA tests to be re-run.
Minimum - re-run all the tests.
Ideal - Have an option to re-run all tests, or just specific tests.

tdawson:
A preliminary investigation has already been done concerning this.  For more details on who to contact, and possible ways to implement this, please see issue #2323

Zlopez:
This is a part of https://github.com/fedora-infra/bodhi/issues/3288 and https://github.com/fedora-infra/bodhi/issues/3287 now. So I'm closing this.

PR #3524, opened 2019-09-27
Add API call to retrigger update tests

Zlopez:
This commit adds API call which will publish bodhi.update.status.testing
message to re-trigger tests again.

This should help with implementation of #3284, #3285, #3286, #3287
and #3288.


cverna (approved):
LGTM

PR #3525, opened 2019-10-01
Improve Unit Test coverage

ryanlerch:
100% test coverage for server/views/generic.py 
100% test coverage for server/util.py 
100% test coverage for server/views/validators.py

### DIFF
--- a/bodhi/server/util.py
+++ b/bodhi/server/util.py
@@ -119,19 +119,6 @@ def build_evr(build):
     return tuple(map(str, (build['epoch'], build['version'], build['release'])))
 
 
-def link(href, text):
-    """
-    Return an HTML anchor tag for the given href using the given text.
-
-    Args:
-        href (str): The URL to link.
-        text (str): The text to render for the link.
-    Returns:
-        str: The requested anchor tag.
-    """
-    return '<a href="%s">%s</a>' % (href, text)
-
-
 class memoized(object):
     """Decorator that permanently caches a function's return value each time it is called.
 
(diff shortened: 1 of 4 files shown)


ryanlerch:
Throwing the WIP label on this one, as i am going to try to 100% another one :)

bodhi/server/validators.py
```diff
@@ -454,7 +454,7 @@ def validate_acls(request, **kwargs):
             release = cache_release(request, build)
             if release is None:
                 return
-        elif 'update' in request.validated:
+        else:
```
cverna: Is this change needed ? I would be cautious about changing the logic in this validator :-)

ryanlerch: with the nature of the checker, there never seems to be a case where we would be able to get this far, and not have update in request.validated. I'll pull this change out though, but validators.py won't be 100% covered anymore...

bodhi/server/util.py
```diff
@@ -119,19 +119,6 @@ def build_evr(build):
     return tuple(map(str, (build['epoch'], build['version'], build['release'])))
 
 
-def link(href, text):
```
Zlopez, line 122: Why is this removed? It is not called anymore?

Zlopez (commented):
Just one comment, otherwise looks good.

bodhi/server/util.py
ryanlerch: yeah, this one is unused now.

Zlopez (approved):

PR #3526, opened 2019-10-01
Add option to trigger tests to CLI

Zlopez:
This is a minimal implementation of #3288.

### DIFF
--- a/bodhi/client/__init__.py
+++ b/bodhi/client/__init__.py
@@ -924,6 +924,36 @@ def waive(update: str, show: bool, test: typing.Iterable[str], comment: str, url
         print_resp(resp, client)
 
 
+@updates.command(name="trigger-tests")
+@click.argument('update')
+@openid_option
+@staging_option
+@url_option
+@debug_option
+@handle_errors
+def trigger_tests(update: str, url: str, openid_api: str, **kwargs):
+    # User Docs that show in the --help
+    """
+    Trigger tests for existing update in testing status.
+
+    UPDATE: The alias of the update (e.g. FEDORA-2017-f8e0ef2850)
+    """
+    # Developer Docs
+    """
+    The update argument must be an update id..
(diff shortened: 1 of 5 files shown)


bodhi/client/__init__.py
```diff
@@ -924,6 +924,36 @@ def waive(update: str, show: bool, test: typing.Iterable[str], comment: str, url
...
 
 
+@updates.command()
+@click.argument('update')
+@openid_option
+@staging_option
+@url_option
+@debug_option
+@handle_errors
+def trigger(update: str, url: str, openid_api: str, **kwargs):
```
cverna: I don't find trigger to be very explicit on what this command does. As a user typing `bodhi update trigger`does not really speek to me.
Maybe something like `bodhi update trigger-tests` or `bodhi update start-tests` would be better. What do you think ?

Zlopez: I wasn't sure how to use option that has two words in it. Looked through bodhi commands and didn't found any. This is why I went for one word option.

bodhi/client/__init__.py
```diff
@@ -924,6 +924,36 @@ def waive(update: str, show: bool, test: typing.Iterable[str], comment: str, url
         print_resp(resp, client)
 
 
+@updates.command()
```
cverna:
```suggestion
@updates.command(name="start-tests")
```

bodhi/client/__init__.py
cverna: I think this can be done has suggested above.

bodhi/client/__init__.py
Zlopez: Thanks

Zlopez:
Addressed @cverna comments and changed option to `trigger-tests`

cverna (approved):

PR #3528, opened 2019-10-02
tidy up the masthead / header a bit

ryanlerch:

### DIFF
--- a/bodhi/server/services/updates.py
+++ b/bodhi/server/services/updates.py
@@ -428,7 +428,7 @@ def query_updates(request):
     # when rendering the html, so we add this here.
     if request.accept.accept_html():
         return_values.update(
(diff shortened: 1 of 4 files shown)


bodhi/server/services/updates.py
```diff
@@ -428,7 +428,7 @@ def query_updates(request):
     # when rendering the html, so we add this here.
     if request.accept.accept_html():
         return_values.update(
-            types=reversed(list(bodhi.server.models.UpdateType.values())),
```
cverna, line 431: why do we remove that here ?

ryanlerch: the reversed was causing some issues with the tests, and TBH, the order here for types doesnt really matter, so i removed it.

cverna (approved):
LGTM

PR #3532, opened 2019-10-04
Add Release as a required item to the Update message schema

### LINKED ISSUE #3515: Include the releases in the message schema of updates
For the CI pipeline it would be helpful if `bodhi.update.status.testing` messages were to include the release information.

I was surprised that this was not already the case and I saw on datagrepper that quite a few messages actually already include this information (for example: https://apps.fedoraproject.org/datagrepper/id?id=2019-0612255c-501f-40e1-91b5-6026044cbea3&is_raw=true&size=extra-large) but I'm not seeing it in the message schema.

Could we either fix the message schema in all places or at least for `bodhi.update.status.testing` messages?

Thanks

ryanlerch:
This adds ReleaseV1 message schema snippet, and adds it
to the Updatev1 release schema, to ensure that all update
related messages contain the Release information about
a particular update.

Fixes #3515

### DIFF
--- a/bodhi/messages/schemas/base.py
+++ b/bodhi/messages/schemas/base.py
@@ -173,6 +173,32 @@ class BuildV1(typing.NamedTuple):
         }
 
 
+class ReleaseV1(typing.NamedTuple):
+    """
+    A model for referencing a Release.
+
+    Attributes:
+        name: The name of the release.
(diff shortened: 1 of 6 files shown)


pypingou (approved):
Thanks!

PR #3536, opened 2019-10-07
Prevent user from creating an update from a tag that is not a sidetag

siddharthvipul:
+ Prevent user from creating an update if the tag doesn't have a side tag
+ test for feature

### DIFF
--- a/bodhi/server/validators.py
+++ b/bodhi/server/validators.py
@@ -1323,10 +1323,18 @@ def validate_from_tag(request: pyramid.request.Request, **kwargs: dict):
         koji_client = buildsys.get_session()
         taginfo = koji_client.getTag(koji_tag)
 
+        if not taginfo:
+            request.errors.add('body', 'from_tag', "The supplied from_tag doesn't exist.")
+            return
+
+        # prevent user from creating an update if tag is not a side tag
+        if not taginfo.get('extra', {}).get('sidetag'):
+            request.errors.add('body', 'from_tag', "The supplied tag is not a side tag.")
+            return
+
         # add all the inherited tags of a sidetag to from_tag_inherited
-        if taginfo and taginfo['extra']['sidetag']:
-            for tag in koji_client.getFullInheritance(koji_tag):
-                request.from_tag_inherited.append(tag['name'])
+        for tag in koji_client.getFullInheritance(koji_tag):
+            request.from_tag_inherited.append(tag['name'])
 
         if request.validated.get('builds'):
             # Builds were specified explicitly, verify that the tag exists in Koji.
--- a/bodhi/tests/server/test_validators.py
+++ b/bodhi/tests/server/test_validators.py
@@ -843,8 +843,8 @@ class TestValidateFromTag(BasePyTestCase):
 
         This makes the validator a no-op."""
         del self.request.validated['from_tag']
-
         validators.validate_from_tag(self.request)
+        assert not self.request.errors
 
     # Error conditions
 
@@ -901,3 +901,23 @@ class TestValidateFromTag(BasePyTestCase):
 
         # check type of wrapped exception
         assert isinstance(excinfo.value.__cause__, koji.GenericError)
+
+    def test_with_unknown_tag(self):
+        """Test to prevent users to create an update from a tag that doesn't exist"""
+        with mock.patch('bodhi.server.validators.buildsys.get_session',
+                        self.mock_get_session_for_class(self.UnknownTagDevBuildsys)):
+            validators.validate_from_tag(self.request)
+
+        assert self.request.errors == [
+            {'location': 'body', 'name': 'from_tag',
+             'description': "The supplied from_tag doesn't exist."}
+        ]
+
+    def test_without_sidetag(self):
+        """A tag that is not a sidetag should add an error to the request"""
+        self.request.validated['from_tag'] = 'no-side-tag'
+        validators.validate_from_tag(self.request)
+        assert self.request.errors == [
+            {'location': 'body', 'name': 'from_tag',
+             'description': "The supplied tag is not a side tag."}
+        ]
--- a/bodhi/server/buildsys.py
+++ b/bodhi/server/buildsys.py
@@ -464,7 +464,7 @@ class DevBuildsys:
                     'id': 7777}
 
         return {'maven_support': False, 'locked': False, 'name': taginfo,
-                'perm': None, 'id': 246, 'arches': None,
+                'extra': {}, 'perm': None, 'id': 246, 'arches': None,
                 'maven_include_all': False, 'perm_id': None}
 
     def getFullInheritance(self, taginfo, **kw):


bodhi/server/validators.py
```diff
@@ -1323,6 +1323,9 @@ def validate_from_tag(request: pyramid.request.Request, **kwargs: dict):
         koji_client = buildsys.get_session()
         taginfo = koji_client.getTag(koji_tag)
 
+        if not taginfo['extra'].get('sidetag'):
+            request.errors.add('body', 'from_tag', "The supplied tag doesn't contain a side tag")
+
```
ryanlerch: Are we sure that only sidetags have extra in there? While the sidetag plugin doesnt have an isSideTag or similar, there is listSideTags() which gives a list of all the sidetags.

bodhi/server/validators.py
```diff
@@ -1323,6 +1323,9 @@ def validate_from_tag(request: pyramid.request.Request, **kwargs: dict):
         koji_client = buildsys.get_session()
         taginfo = koji_client.getTag(koji_tag)
 
+        if not taginfo['extra'].get('sidetag'):
+            request.errors.add('body', 'from_tag', "The supplied tag doesn't contain a side tag")
```
nphilipp: The error message is a bit off, and validators need to return after encountering and setting the error. Also, it would crash if the tag doesn't exist (`koji_client.getTag(koji_tag)` would return `None` then).

I'd change it like this, i.e. fix the message and error handling in your code and consolidate it with the lines below:

```
    if not taginfo:
        request.errors.add('body', 'from_tag', "The supplied tag doesn't exist.")
        return

    # The implementation in KojiHub guarantees that 'extra' exists.
    if not taginfo['extra'].get('sidetag'):
        request.errors.add('body', 'from_tag', "The supplied tag isn't a side tag.")
        return

    # add all the inherited tags of a sidetag to from_tag_inherited
    for tag in koji_client.getFullInheritance(koji_tag):
        request.from_tag_inherited.append(tag['name'])
```

This would probably need another test for the "supplied tag doesn't exist" case.

bodhi/tests/server/test_validators.py
```diff
@@ -790,8 +790,15 @@ def test_without_from_tag(self):
 
         This makes the validator a no-op."""
         del self.request.validated['from_tag']
+        validators.validate_from_tag(self.request)
+        assert self.request.errors == "The supplied from_tag doesn't exist"
```
nphilipp: This assertion is wrong. :wink: Firstly, `self.request.errors` is a list of dicts (see test methods beneath `# Error conditions`. Secondly, it should be empty because not having a `from_tag` at all is valid -- it's a normal update then. Good catch that the assertion is missing, though! It should be something like `assert not self.request.errors`.

bodhi/tests/server/test_validators.py
```diff
@@ -790,8 +790,15 @@ def test_without_from_tag(self):
...
 
+    def test_with_sidetag(self):
+        """Test to prevent users to create an update from a tag that is not a side tag"""
         validators.validate_from_tag(self.request)
+        assert self.request.validated['from_tag']['extra'].get('sidetag')
+        assert not self.request.errors
+
```
nphilipp: This is already tested in `test_known_without_builds()` above.

bodhi/tests/server/test_validators.py
```diff
@@ -848,3 +855,14 @@ def test_unknown_koji_genericerror(self):
 
         # check type of wrapped exception
         assert isinstance(excinfo.value.__cause__, koji.GenericError)
+
+    def test_without_sidetag(self):
+        """A tag without a sidetag should add an error to the request"""
```
nphilipp: """A tag **which isn't** a side tag..."""

bodhi/tests/server/test_validators.py
```diff
@@ -848,3 +855,14 @@ def test_unknown_koji_genericerror(self):
...
+        with mock.patch('bodhi.server.validators.buildsys.get_session',
+                        self.mock_get_session_for_class(self.UnknownTagDevBuildsys)):
+            validators.validate_from_tag(self.request)
+
+        assert self.request.errors == [
+            {'location': 'body', 'name': 'from_tag',
+             'description': "The supplied tag doesn't contain a side tag"}
```
nphilipp: Adapt to changes above:

```
    'description': "The supplied tag isn't a side tag."}
```

nphilipp (changes requested):
Hi Vipul, I found some issue with error handling and the tests. Would you please check this? Thanks!

bodhi/server/validators.py
nphilipp: The `extra` dict is always in `taginfo`, the KojiHub code takes care of that. For a side tag, it would e.g. look like this (see https://koji.stg.fedoraproject.org/koji/taginfo?tagID=7715 or its corresponding `koji_client.getTag(...)` API output):

```
'extra': {'sidetag_user': 'nphilipp', 'sidetag': True, 'sidetag_user_id': 11},
```

For a normal tag, the `extras` dict would be missing the `sidetag` field, i.e. `taginfo['extra'].get('sidetag')` returns `False`.

bodhi/server/validators.py
siddharthvipul: makes much more sense,
Thank you for this :)

bodhi/tests/server/test_validators.py
siddharthvipul: changed

bodhi/tests/server/test_validators.py
```diff
@@ -907,3 +907,25 @@ def test_unknown_koji_genericerror(self):
...
+        assert self.request.errors == [
+            {'location': 'body', 'name': 'from_tag',
+             'description': "The supplied from_tag doesn't exist."}
+        ]
+
+    def test_without_sidetag(self):
+        """A tag that is not a sidetag should add an error to the request"""
+        with mock.patch('bodhi.server.validators.buildsys.get_session',
+                        self.mock_get_session_for_class(self.UnknownTagDevBuildsys)):
+            validators.validate_from_tag(self.request)
```
nphilipp: This test shouldn't mock the Koji client with one which always behaves as if the tag is unknown -- it is about known tags which aren't side tags. I.e., it should look like this:

````
    def test_without_sidetag(self):
        """A tag that is not a sidetag should add an error to the request"""
        self.request.validated['from_tag'] = 'no-side-tag'

        validators.validate_from_tag(self.request)

        assert self.request.errors == [ 
            {'location': 'body', 'name': 'from_tag',
             'description': "The supplied tag is not a side tag."}
        ]
````

For this to work, the mocked Koji client (from `bodhi/server/buildsys.py`) needs to return the `extra` dict which we assume above, or the code needs to cope with it missing from the API response. To be safe, we should do both :wink::

````
diff --git a/bodhi/server/buildsys.py b/bodhi/server/buildsys.py
index 6e9048bc..dd198f9d 100644
--- a/bodhi/server/buildsys.py
+++ b/bodhi/server/buildsys.py
@@ -464,7 +464,7 @@ class DevBuildsys:
                     'id': 7777}
 
         return {'maven_support': False, 'locked': False, 'name': taginfo,
-                'perm': None, 'id': 246, 'arches': None,
+                'extra': {}, 'perm': None, 'id': 246, 'arches': None,
 
     def getFullInheritance(self, taginfo, **kw):
diff --git a/bodhi/server/validators.py b/bodhi/server/validators.py
index c96919b4..389cf8e9 100644
--- a/bodhi/server/validators.py
+++ b/bodhi/server/validators.py
@@ -1328,7 +1328,7 @@ def validate_from_tag(request: pyramid.request.Request, **kwargs: dict):
             return
 
         # prevent user from creating an update if tag is not a side tag
-        if not taginfo['extra'].get('sidetag'):
+        if not taginfo.get('extra', {}).get('sidetag'):
             request.errors.add('body', 'from_tag', "The supplied tag is not a side tag.")
             return
 
````

nphilipp (changes requested):

nphilipp (approved):
Looks good to me, thanks!

PR #3538, opened 2019-10-08
fix test_get_updates_view failure on integration tests

ryanlerch:
Occasionally, the data in the DB dump would cause the
devel.ci.integration.tests.test_bodhi.test_get_updates_view test to
fail. This would occur when one of the latest 20 updates had the
optional "display_name" field set for the update. Previously, the test
in test_get_updates_view only checked for the update title as
contstructed from the build names, not taking into account the
display_name if one was set. THis commit fixes this issue.


cverna (approved):
:+1: thanks

PR #3539, opened 2019-10-08
Reload update page when commenting

### LINKED ISSUE #3502: When leaving a comment with karma, the karma is not updatedin the webUI
After leaving a comment, bodhi updates the update page with JS, adding the comment to the bottom of the comment list. However, if a user adds karma, the karma total for the update is not changed on the page until a refresh of the page is done.

ryanlerch:
Previously, when adding a comment to an update, the javascript
dynamically added the comment to the bottom of the list, but did not
update the karma. In this commit, we just reload the entire page rather
than fuss with adding the information correctly after the async comment
form returns a success.

Fixes: #3502


pypingou (approved):
Nice and simple

PR #3543, opened 2019-10-09
add new version_hash property to Update model

### LINKED ISSUE #3497: Add versioning for updates
Basically we need  a way to uniquely identify an update as a specific list of builds (1 or more).
If the update is edited (builds added/removed), that list of builds change and thus the version should be different.

In the world of CI, we want to test the updates (uniquely identified by their alias/updateid), but since the builds in the update may change, we currently cannot distinguish test results from an update containing the build A-1 and B-1 from the same update edited to contain the builds A-1 and B-2.

Adding a `version` field to the update should allow for this distinction.

Note: that field should appear on the message sent to the bus.

ryanlerch:
This adds a new version_hash property to the Update model, which is a
SHA1 hash of all the builds in an update.

Fixes: #3497

### DIFF
--- a/bodhi/server/models.py
+++ b/bodhi/server/models.py
@@ -1758,7 +1758,7 @@ class Update(Base):
 
     __tablename__ = 'updates'
     __exclude_columns__ = ('id', 'user_id', 'release_id')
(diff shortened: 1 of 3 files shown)


pypingou (approved):
Thanks !

PR #3548, opened 2019-10-10
Add changelog entry to update notes for automatic updates

### LINKED ISSUE #3192: Automatic updates should be filled with information from the RPM changelog
This is a continuation of issue #3007 and covers items deemed not that high of a priority, not needed for the MVP.

As a Bodhi user, when an update is created automatically because a build was tagged into a certain tag, I want that new entries in the changelog are evaluated and:

- discovered Bugzilla tickets are associated with the update
- changelog entries are used to pre-fill the update notes

so that the update is more informational.

Thoughts:

- We need to figure out what the new changelog entries are. Between builds, we could e.g. have more than one entry. Old entries could even be removed, or existing metadata changed (e.g. when a previously wrong changelog date or weekday is fixed) which makes discovery of "what is new" harder.
- When a new build triggers a new update, it should probably obsolete and already existing (but active) update of the same package, and inherit Bugzilla tickets and update notes. With manually created updates, the notes could just be merged and the maintainer would sort out commonalities etc. How would we handle this automatically?

ryanlerch:
This abstracts out the code in mail.py that generates the changelog entry, and puts it in util.py so we can use it in other places. Then, this new abstraction is used to add a changelog entry to the updates notes of automatic updates.

Fixes: #3192

### DIFF
--- a/bodhi/server/consumers/automatic_updates.py
+++ b/bodhi/server/consumers/automatic_updates.py
@@ -30,7 +30,7 @@ from bodhi.server import buildsys
 from bodhi.server.config import config
 from bodhi.server.models import Build, ContentType, Package, Release
 from bodhi.server.models import Update, UpdateStatus, UpdateType, User
-from bodhi.server.util import transactional_session_maker
+from bodhi.server.util import generate_changelog, transactional_session_maker
 
 log = logging.getLogger('bodhi')
 
@@ -139,10 +139,21 @@ class AutomaticUpdateHandler:
                 dbsession.add(user)
 
             log.debug(f"Creating new update for {bnvr}.")
+            changelog = generate_changelog(build)
+            if changelog:
+                notes = f"""Automatic update for {bnvr}.
+
+##### **Changelog**
+
+```
+{changelog}
+```"""
+            else:
+                notes = f"Automatic update for {bnvr}."
             update = Update(
                 release=rel,
                 builds=[build],
-                notes=f"Automatic update for {bnvr}.",
+                notes=notes,
                 type=UpdateType.unspecified,
                 stable_karma=3,
                 unstable_karma=-3,
--- a/bodhi/server/util.py
+++ b/bodhi/server/util.py
@@ -103,6 +103,36 @@ def get_rpm_header(nvr, tries=0):
     raise ValueError("No rpm headers found in koji for %r" % nvr)
 
 
+def generate_changelog(build: 'models.Build') -> typing.Optional[str]:
+    """
+    Generate a changelog for a given build.
+
+    Args:
+        build: the build to create a changelog for.
+    Returns:
+        A changelog of changes between the given build, and the previous one.
+        Or returns None if the build type doesn't have the get_latest() method.
+    """
+    # Find the most recent update for this package, other than this one
+    try:
+        lastpkg = build.get_latest()
+    except AttributeError:
+        # Not all build types have the get_latest() method, such as ModuleBuilds.
+        return None
+
+    # Grab the RPM header of the previous update, and generate a ChangeLog
+    oldh = get_rpm_header(lastpkg)
+    oldtime = oldh['changelogtime']
+    text = oldh['changelogtext']
+
+    if not text:
+        oldtime = 0
+    elif isinstance(oldtime, list):
+        oldtime = oldtime[0]
+
+    return build.get_changelog(oldtime)
+
+
 def build_evr(build):
     """
     Return a tuple of strings of the given build's epoch, version, and release.
--- a/bodhi/tests/server/consumers/test_automatic_updates.py
+++ b/bodhi/tests/server/consumers/test_automatic_updates.py
@@ -92,6 +92,37 @@ class TestAutomaticUpdateHandler(base.BasePyTestCase):
 
         assert not any(r.levelno >= logging.WARNING for r in caplog.records)
 
+    @pytest.mark.parametrize('changelog', (True, None, ""))
+    @mock.patch('bodhi.server.consumers.automatic_updates.generate_changelog')
+    def test_changelog(self, mock_generate_changelog, changelog):
+        """Assert that update notes contain the changelog if it exists."""
+        if changelog:
+            # fill the changelog here rather than in the decorator
+            changelog = ('* Sat Aug  3 2013 Fedora Releng <rel-eng@lists.fedoraproject.org> - 2\n'
+                         '- Added a free money feature.\n* Tue Jun 11 2013 Randy <bowlofeggs@fpo>'
+                         ' - 2.0.1-2\n- Make users ☺\n')
+
+        mock_generate_changelog.return_value = changelog
+
+        # process the message
+        self.handler(self.sample_message)
+
+        # check if the update exists...
+        update = self.db.query(Update).filter(
+            Update.builds.any(Build.nvr == self.sample_nvr)
+        ).first()
+
+        if changelog:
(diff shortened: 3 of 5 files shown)


bodhi/server/util.py
```diff
@@ -103,6 +103,36 @@ def get_rpm_header(nvr, tries=0):
...
+def generate_changelog(build):
+    """
+    Generate a changelog for a given build.
+
+    Args:
+        build (bodhi.server.models.Build): the build to create a changelog for.
+    Returns:
+        str or None: a changelog of changes between the given build, and the previous one.
+            or returns None if the build type doesn't have the get_latest() method.
+    """
```
nphilipp: We want to use type hints in new (non-testing) code (see the [developer docs](https://bodhi.fedoraproject.org/docs/developer/index.html)), i.e.:

```
def generate_changelog(build: models.Build) -> typing.Optional[str]:
    """
    Generate a changelog for a given build.

    Args:
         build: the build to create a changelog for.
    Returns:
        A changelog of changes between the given build, and the previous one,
        or None if the build type doesn't have the get_latest() method.
    """
```

bodhi/server/util.py
```diff
@@ -103,6 +103,36 @@ def get_rpm_header(nvr, tries=0):
...
+        lastpkg = build.get_latest()
+    except AttributeError:
+        # Not all build types have the get_latest() method, such as ModuleBuilds.
+        return None
+
+    # Grab the RPM header of the previous update, and generate a ChangeLog
+    oldh = get_rpm_header(lastpkg)
+    oldtime = oldh['changelogtime']
+    text = oldh['changelogtext']
+    del oldh
```
nphilipp: Is there a reason to delete the `oldh` reference explicitly? It should fall out of context when returning from the function anyway.

bodhi/tests/server/consumers/test_automatic_updates.py
```diff
@@ -91,6 +91,70 @@ def test_consume(self, caplog):
...
+        self.handler(self.sample_message)
+
+        # check if the update exists...
+        update = self.db.query(Update).filter(
+            Update.builds.any(Build.nvr == self.sample_nvr)
+        ).first()
+
+        print(update.notes)
+        assert "##### **Changelog**" not in update.notes
+
```
nphilipp: Instead of repeating the almost same test method three times, you could use the `@pytest.mark.parametrize()` decorator to cater for the differences:

````
    @pytest.mark.parametrize('changelog', (True, None, "")
    @mock.patch('bodhi.server.consumers.automatic_updates.generate_changelog')
    def test_changelog(self, mock_generate_changelog, changelog):
        """Assert that update notes contain the changelog if it exists."""
        if changelog:
            # fill the changelog here rather than in the decorator
            changelog = ('* Sat Aug  3 2013 Fedora Releng <rel-eng@lists.fedoraproject.org> - 2.1.0-1\n'
                                  '- Added a free money feature.\n* Tue Jun 11 2013 Randy <bowlofeggs@fpo>'
                                  ' - 2.0.1-2\n- Make users ☺\n')

        [... run the things ...]

        if changelog:
            assert update.notes == f"""Automatic update for colord-1.3.4-1.fc26.

##### **Changelog**

```
{changelog}
```"""
        else:  # no changelog
            assert update.notes == "Automatic update for colord-1.3.4-1.fc26."
````

File: bodhi/server/consumers/automatic_updates.py [hunk 4]
```diff
@@ -139,10 +139,21 @@ def __call__(self, message: fedora_messaging.api.Message) -> None:
...
 
             log.debug(f"Creating new update for {bnvr}.")
+            changelog = generate_changelog(build)
+            if changelog:
+                notes = f"""Automatic update for {bnvr}.
+
+##### **Changelog**
+
+```
+{generate_changelog(build)}
```
nphilipp: You could use `changelog` directly instead of calling `generate_changelog()` twice here.

nphilipp (changes requested):
Looks good overall, only some minor things I'd change. Thanks!

bodhi/server/util.py
ryanlerch: no reason, except that is how it was done in the code i copied over from mail.py. have removed it now

bodhi/tests/server/consumers/test_automatic_updates.py
ryanlerch: done, thanks!

bodhi/server/consumers/automatic_updates.py
ryanlerch: done!

bodhi/server/util.py
```diff
@@ -103,6 +103,36 @@ def get_rpm_header(nvr, tries=0):
     raise ValueError("No rpm headers found in koji for %r" % nvr)
 
 
+def generate_changelog(build) -> typing.Optional[str]:
```
nphilipp: The type description for the `build` parameter is missing:

```
    def generate_changelog(build: models.Build) -> typing.Optional[str]:
```

bodhi/server/util.py
```diff
@@ -103,6 +103,36 @@ def get_rpm_header(nvr, tries=0):
...
+    """
+    Generate a changelog for a given build.
+
+    Args:
+        build: the build to create a changelog for.
+    Returns:
+        a changelog of changes between the given build, and the previous one.
+        or returns None if the build type doesn't have the get_latest() method.
```
nphilipp: Capitalize the first words in the sentences?

nphilipp (changes requested):
Only minor nitpicking this time. :wink:

bodhi/server/util.py
ryanlerch: Reallt not sure what this is supposed to fix here.

I tried adding this in:

def generate_changelog(build: models.Build) -> typing.Optional[str]:
NameError: name 'models' is not defined

Then tried adding models not in the conditional import of typing.TYPE_CHECKING, but stiull getting another error:

def generate_changelog(build: models.Build()) -> typing.Optional[str]:
AttributeError: module 'bodhi.server.models' has no attribute 'Build'

ryanlerch: okies, figured out it needs to be quoted out:

```
def generate_changelog(build: 'models.Build') -> typing.Optional[str]:

```

nphilipp: Ah right, I didn't notice that the `models` module is only imported conditionally. My bad.

nphilipp (approved):
Looks good to me, thanks!

nphilipp:
@ryanlerch I'm not sure why the integration tests fail here (something timed out), on my Vagrant box they succeed. I'll "empty-amend"\* and force-push so the CI pipeline conducts the tests again.

\*: i.e. `git commit --amend --no-edit && git push --force` which will just update the timestamp of the latest commit, to make the CI pipeline do the tests again.

abompard (changes requested):
Since there will be some user-visible change, I think that this change should be listed in the next release notes. Please add a file in the `news` directory as explained in [this documentation](https://github.com/fedora-infra/bodhi/blob/develop/docs/developer/index.rst#release-notes). Thanks!

ryanlerch:
> Since there will be some user-visible change, I think that this change should be listed in the next release notes. Please add a file in the `news` directory as explained in [this documentation](https://github.com/fedora-infra/bodhi/blob/develop/docs/developer/index.rst#release-notes). Thanks!

DOne! thanks!

abompard (approved):

PR #3550, opened 2019-10-10
Move the --user and --password arguments in a group with --openid-api

pypingou:
This way action which allow to set the --openid-api arguments will not
forget to add the --user and --password ones (as did the waive and
request-tests).
This commit also fixes the waive and request-tests action so they allow
logging in from the CLI via the --user and --password arguments as do
other actions requiring to be logged in.

### DIFF
--- a/bodhi/client/__init__.py
+++ b/bodhi/client/__init__.py
@@ -96,12 +96,17 @@ url_option = click.option('--url', envvar='BODHI_URL', default=bindings.BASE_URL
                           help=('URL of a Bodhi server. Ignored if --staging is set. Can be set '
                                 'with BODHI_URL environment variable'),
                           callback=_warn_if_url_or_openid_and_staging_set)
-openid_option = click.option(
-    '--openid-api', envvar='BODHI_OPENID_API',
-    default=openidproxyclient.FEDORA_OPENID_API,
-    help=('URL of an OpenID API to use to authenticate to Bodhi. Ignored if --staging is set. Can '
-          'be set with BODHI_OPENID_API environment variable'),
-    callback=_warn_if_url_or_openid_and_staging_set)
+openid_options = [
+    click.option('--user'),
+    click.option('--password', hide_input=True),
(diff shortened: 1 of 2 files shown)


nphilipp (commented):
Would it make sense to add `openid_options` to `new_edit_options`, `save_edit_options` and `release_options` instead of adding the `@add_options(openid_options)` decorator to every function that uses these three?

pypingou:
> Would it make sense to add `openid_options` to `new_edit_options`, `save_edit_options` and `release_options` instead of adding the `@add_options(openid_options)` decorator to every function that uses these three?

`new_edit_options`, `save_edit_options` and `release_options` are all lists, so we could expend the lists where they are defined but then we end up with some method which are explicit about including the `openid_options` and some which aren't because they bundle these options in others (new_edit_options...).

Zlopez (approved):
This looks much better

nphilipp:
> > Would it make sense to add `openid_options` to `new_edit_options`, `save_edit_options` and `release_options` instead of adding the `@add_options(openid_options)` decorator to every function that uses these three?
> 
> `new_edit_options`, `save_edit_options` and `release_options` are all lists, so we could expend the lists where they are defined but then we end up with some method which are explicit about including the `openid_options` and some which aren't because they bundle these options in others (new_edit_options...).

Yeah, this could be more confusing than helpful.

nphilipp:
@pypingou I'm not sure why the integration tests fail on f31 and Rawhide, they succeed in my Vagrant box. I'll force-push the branch without changes to retrigger the tests.

abompard (commented):
Do you think that this change should be listed in the next release notes? If so, please add a file in the `news` directory as explained in [this documentation](https://github.com/fedora-infra/bodhi/blob/develop/docs/developer/index.rst#release-notes). Thanks!

abompard (dismissed):
I think this is worthy of an entry in the release notes because new options have been added to the CLI.

PR #3565, opened 2019-10-11
Introduce Towncrier to manage the release notes

### LINKED ISSUE #3562: The release notes are a pain to write
Before cutting a release, composing the release notes from all individual commits and PRs is a pain. It could be solved by the introduction of a tool like [towncrier](https://github.com/hawkowl/towncrier/).

abompard:
This should help a lot with the next releases.

There are changes in the developer guidelines, please review them and create the files in the `news` directory accordingly.
Also, when reviewing PRs, please make sure that the submitter did not forget to add a news fragment and a migration fragment when appropriate, as well as their author fragment, in the `news` directory.

Fixes #3562


news/get-authors.py
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env python3
```
Zlopez, line 1: I recommend to add some description of this file. You could copy it from Anitya - https://github.com/release-monitoring/anitya/blob/master/news/get-authors.py

docs/developer/releases.rst
```diff
@@ -50,29 +50,23 @@ How to make a release
...
-      ``--compare-branch`` flag being passed to ``diff-cover``.
-
-#. Raise the version to the appropriate value in ``setup.py`` and ``docs/conf.py``.
-#. Use ``git log`` to find all commits that are new in this release and use the commit messages to
-   write release notes in ``docs/user/release_notes.rst``. You can filter ``git log`` to show commit
-   that differ between two git refs to help. For example, to see differences between ``3.12`` and
-   ``3.13``, you could use ``git log 3.12..3.13``. Be aware that any commits that were cherry picked
-   to ``3.12`` will show up in this example and we wouldn't want to mark those commits as being new
-   in ``3.13``.
+If you are making a new major or minor release:
```
Zlopez, line 53: Does bodhi uses separate package for fedora messaging schemes with separate versioning?

If yes, there should be also mentioned here.

docs/developer/index.rst
```diff
@@ -82,6 +80,39 @@ Before you submit a pull request to Bodhi, please ensure that it meets these cri
...
+* If it is not present already, add a file in the ``news`` directory named ``username.author``
+  where ``username`` is the first part of your commit's email address, and containing the name
+  you want to be credited as.
+* If you think that the changes you make should be in release notes, add a file in the ``news``
+  directory with the format explained below.
+
+Release Notes
+-------------
+
+To add entries to the release notes, create a file in the ``news`` directory in the
```
Zlopez, line 95: I would recommend to switch `type` and `source` order, because I always need to check if it's `source.type` or `type.source`. This is caused by type being explained first and then source comes to play.

docs/developer/index.rst
```diff
@@ -82,6 +80,39 @@ Before you submit a pull request to Bodhi, please ensure that it meets these cri
   `github.com/fedora-infra/bodhi <https://github.com/fedora-infra/bodhi/>`_, or you may e-mail a
   patch to the
   `mailing list <https://lists.fedoraproject.org/archives/list/bodhi@lists.fedorahosted.org/>`_.
+* If your changes contain database migrations, you must add a file named ``summary.migration``
+  in the ``news`` directory, where ``summary`` is a short textual description of the change
+  (or the issue number you're fixing).
+* If it is not present already, add a file in the ``news`` directory named ``username.author``
```
Zlopez, line 86: I don't think this is needed if you have `get-authors.py` script.

Zlopez (commented):
Just few comments, otherwise looks :+1:

news/get-authors.py
abompard: Added, thanks.

docs/developer/releases.rst
abompard: No, all sub-packages have the same version, including the message schemas.

docs/developer/index.rst
abompard: Yeah that's more readable, I used to use the other order because I wanted to introduce the author type first because it was a special case in the source part, but now I've reordered the docs and that's no longer necessary. I'll switch them.

docs/developer/index.rst
abompard: Yeah, that was more in case the get-author script makes mistakes, which I've seen it do because some authors use weird or inconsistent email addresses in their git config (like `someone+github@example.com`), so I wanted to offer a way out.

Also, I think it's more and information that's part of the patch than part of the release process commit. I see the get-authors script as a safety net to avoid forgetting someone, not as the plan A for doing things.

PR #3566, opened 2019-10-11
Update.signed is a property do not try to call it

cverna:


abompard (changes requested):
I would really love to see some unit tests to make sure this does not happen again, if possible. Thanks!

abompard (approved):
Thanks! :-)

PR #3569, opened 2019-10-15
Remove metrics page.

ryanlerch:
The Metrics page was initially created before bodhi had the concept of multiple content types and releases, and as such it is kind of not useful much any more in it's current format. 

Old view:
Screenshot from 2019-10-15 15-02-24

Graphs like these might be better suited to the releases list page, or on the page for individual releases.

This PR removes the metrics view.


Zlopez:
Is this information available somewhere else in Bodhi, or do we lose it when removing this graph?

ryanlerch:
yeah, the individual release pages have these breakdowns, just with so many releases now, with varying amounts of updates, this graph is no longer useful (and having a whole endpoint for a single graph is overkill too IMHO

Zlopez (approved):

Issue #3368, opened 2019-07-02
Greenwave consumer: process received message for single-build updates


nphilipp:
Currently, the Greenwave consumer receives a message, then uses `Update.update_test_gating_status()` which queries Greenwave in turn. If an update only contains a single build, we should just use the information from the received message because it's less error-prone.

PR #3571, opened 2019-10-16
Greenwave consumer : use the message data for single build updates

abompard:
See #3368.


bodhi/server/consumers/greenwave.py
```diff
@@ -68,5 +68,28 @@ def __call__(self, message: fedora_messaging.api.Message):
...
+        Extract gating information from the Greenwave message and return it.
+
+        Returns:
+            TestGatingStatus:
+                - TestGatingStatus.ignored if no tests are required
+                - TestGatingStatus.failed if policies are not satisfied
+                - TestGatingStatus.passed if policies are satisfied, and there
+                  are required tests
+        """
+        if not msg['policies_satisfied']:
```
pypingou, line 95: just to be on the safe side: ``msg.get("policies_satisfied")`` here?

pypingou: or better, we check if that key is in the dict first and bail if it isn't then move on to your code as is :)

abompard: Sure, why not, I added that.

abompard:
@pypingou , are you happy with the changes?

bodhi/server/consumers/greenwave.py
```diff
@@ -50,6 +50,7 @@ def __call__(self, message: fedora_messaging.api.Message):
             log.debug("Ignoring message without body.")
             return
 
+        print(msg)
```
pypingou: debugging left over?

pypingou:
one comment the rest looks fine :)

bodhi/server/consumers/greenwave.py
abompard: Indeed! It should be OK now, sorry.

pypingou (approved):
Thanks!

PR #3572, opened 2019-10-16
Don't upgrade packages when building containers for CI

abompard:
There's currently an update that breaks our CI containers, don't upgrade packages when building them.
Also, according to Clément, upgrading packages in containers is a bad practice.

### DIFF
--- a/devel/ci/Dockerfile-f29
+++ b/devel/ci/Dockerfile-f29
@@ -1,9 +1,6 @@
 FROM registry.fedoraproject.org/fedora:29
 LABEL maintainer="Randy Barlow <bowlofeggs@fedoraproject.org>"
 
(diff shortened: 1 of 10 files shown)


cverna (approved):

pypingou:
I propose we merge this one manually as it is supposed to fix the currently failing tests

abompard:
Nah it's not fixing them actually, they still fail with the change. I wonder if it may be a real bug somewhere.

ryanlerch:
is the test_results/ directory supposed to be included here?

abompard:
> is the test_results/ directory supposed to be included here?

No, this PR is just a giant mess a the moment because I can't test locally. Please ignore for now. (also I added files I shouldn't have added)

Issue #3357, opened 2019-06-27
Convert the testcases to PyTest


abompard:
Some testcase classes, for example in `bodhi/tests/server/test_util.py`, derive from `unittest.TestCase`. We use pytest, so they could be migrated to derive from `bodhi.tests.server.base.BaseTestCase`. That would allow us to use the simpler `assert` constructs instead of the `self.assert*()` methods.

mm-git01:
Has this been resolved ? can i take this ?

pypingou:
It has not and it is not a small task so feel free to one file at a time and submit PR as you go :)

mm-git01:
> It has not and it is not a small task so feel free to one file at a time and submit PR as you go :)

cool,will do it. Btw i was pointed to bodhi by vipul-siddhart. I have been a fedora contributor for almost a year+ now and was a google summer of code intern this year with fedora. Want to do more infra work moving forward.

mm-git01:
i took a look at the code today,this is definitely gonna take some time.I'll push thing one by one as they are done.

Navneet78:
Hello I am new here. It would be my first PR can I do this?

mm-git01:
> Hello I am new here. It would be my first PR can I do this?

hey,i am kinda working on this rn. Since this is your first PR and pretty easy,you can take this :smile: 

@pypingou can you please point me to some other issue that i can look into that needs to be fixed.

Navneet78:
Thank you @Pac23

ryanlerch:
@Pac23 @Navneet78 I have started creating a little project for this issue, breaking down each file that needs to be changed into it's own issue.

Feel free to check out the project here, and grab one of these if you are still interested:

https://github.com/fedora-infra/bodhi/projects/7

nphilipp:
@abompard @ryanlerch 

BTW, `bodhi.tests.server.base.BaseTestCase` is the legacy class derived from `unittest.TestCase`, we want stuff to be converted over to `BasePyTestCase` instead—both these classes incorporate `BaseTestCaseMixin` which contains the meat, i.e. the database session with test data in it, and the WSGI TestApp where we can test exposed web API. At some point, when everything has been migrated away from `unittest.TestCase`, we can get rid of `BaseTestCase` and migrate the mixin into `BasePyTestCase`.

nphilipp:
@abompard @ryanlerch 

[...]

PR #3574, opened 2019-10-17
convert test_bugs.py to BaseTestCase

ryanlerch:
Related: #3357


Zlopez (commented):
Why the change from `self.assertSomething` to `assert`?

ryanlerch:
> Why the change from `self.assertSomething` to `assert`?

@abompard talked about this in the issue #3357 so i did it.

Zlopez (approved):

Zlopez:
@ryanlerch I wasn't looked at the original issue, thanks for explanation.

PR #3580, opened 2019-10-17
fix using Vagrant options

nphilipp:
Using GetoptLong in the Vagrantfile will trip over any other Vagrant
options, making e.g. 'vagrant up --provision' fail. Check for the
presence of the --use-staging option manually and remove it from ARGV
(so the main option parser doesn't trip over it).


abompard (approved):

ryanlerch:
thanks @nphilipp :)

PR #3581, opened 2019-10-17
Skip update without builds when creating compose

### LINKED ISSUE #3471: Update without build can be created
It appears that an update without build can be created (https://bodhi.fedoraproject.org/updates/FEDORA-2019-e4e61bb84b). 
We need to make sure that the validators are not letting an update without a build being created.

In that case the update made the compose failed.

Zlopez:
Fixes #3471

### DIFF
--- a/bodhi/server/models.py
+++ b/bodhi/server/models.py
@@ -3997,6 +3997,9 @@ class Compose(Base):
                 continue
             # ASSUMPTION: For now, updates can only be of a single type.
             ctype = None
+            if not update.builds:
+                log.info(f"No builds in {update.alias}. Skipping.")
+                continue
             for build in update.builds:
                 if ctype is None:
                     ctype = build.type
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -36,3 +36,4 @@ nosetests.xml
 virtenv
 .vscode
 test_results
+.projectile
```
abompard: Could you explain what this is please?

Zlopez: This is Emacs projectile module. I started to use it recently.

pypingou: You can add this to ~/.gitignore and this will then be ignored in all your git repositories

Zlopez: @pypingou This will be probably better solution, I hope it is working on the Silverblue now. Last time I tried it, it didn't worked.

Zlopez: `~/.gitignore` is still ignored, but I removed it from `.gitignore`

pypingou (approved):
Looks fine to me

PR #3611, opened 2019-10-18
Don't derive test classes in test_bugs.py from BaseTestCase.

nphilipp:
On the one hand, this still derives from `unittest.TestCase`, which
breaks use of some PyTest features in hard to debug ways. On the other
hand, it also adds unnecessary overhead in form of the test database
that it creates for every called test method.


ryanlerch (approved):

Zlopez (approved):

ryanlerch (changes requested):
Looks like there are 2 self.asserts that i missed in the initial commit on this file, that still worked because i changed them to basetestcase...

https://github.com/fedora-infra/bodhi/pull/3611/files#diff-5d496e504a611a29cbe4cb4888b45c27R298-R301

ryanlerch (approved):

PR #3613, opened 2019-10-18
Migrate test_schemas.py to PyTest

### LINKED ISSUE #3596: Convert bodhi/tests/server/test_schemas.py to PyTest
[bodhi/tests/server/test_schemas.py](https://github.com/fedora-infra/bodhi/blob/develop/bodhi/tests/server/test_schemas.py) still contains classes that derive from `unittest.TestCase`. We use PyTest, so they could be migrated to not derive from `unittest.TestCase`, or derive from `bodhi.tests.server.base.BasePyTestCase` in case they need to use the test database or WSGI app. 

#3612

robshelly:
Removed dependency on unittest.TestCase.

Fixes: [#3596](https://github.com/fedora-infra/bodhi/issues/3596)

Signed-off-by: Rob Shelly <rshelly@redhat.com>


cverna:
@robshelly Hi thanks for that your commit message needs to have the "Signed-off-by:" statement. You can do that by doing `git commit -s`.

To amend your current commit you can just run `git commit --amend -s`

bodhi/tests/server/test_schemas.py
```diff
@@ -21,7 +21,7 @@
 from bodhi.server import schemas
 
 
-class TestSchemas(unittest.TestCase):
+class TestSchemas():
```
nphilipp: In the interests of consistency (and because you have to sign-off on your commit anyway), would you remove the parens?, I.e.:

````
class TestSchemas:
````

nphilipp (changes requested):

cverna (approved):
LGTM, thanks for that :-)

nphilipp (approved):

cverna (commented):
@robshelly You also need to remove the `import unittest` statement since we don't use it anymore :-).

ryanlerch (changes requested):
the `unittest` module is still being imported, but is unused. This is causing the pip-flake8 CI test to fail.

ryanlerch (approved):

PR #3615, opened 2019-10-18
Convert bodhi/tests/server/test_alembic.py to PyTest

### LINKED ISSUE #3599: Convert bodhi/tests/server/test_alembic.py to PyTest
[bodhi/tests/server/test_alembic.py](https://github.com/fedora-infra/bodhi/blob/develop/bodhi/tests/server/test_alembic.py) still contains classes that derive from `unittest.TestCase`. We use PyTest, so they could be migrated to not derive from `unittest.TestCase`, or derive from `bodhi.tests.server.base.BasePyTestCase` in case they need to use the test database or WSGI app. 

#3612

Fan:
Updated test case to use pytest

Fixes: #3599


requirements.txt
```diff
@@ -24,3 +24,11 @@ PyYAML
...
 sqlalchemy
 waitress
+flake8
+pytest
+pytest-cov
+webtest
+conu >= 0.5.0
+munch
+psycopg2
+requests
```
nphilipp: This change shouldn't be part of this PR (`git commit -a` considered harmful? :wink:), and these modules don't belong in `requirements.txt` anyway, they're only needed for testing, not running Bodhi.

bodhi/tests/server/test_alembic.py
```diff
@@ -22,14 +22,15 @@
...
+
+from bodhi.tests.server.base import BasePyTestCase
 
 
 REPO_PATH = os.path.abspath(
     os.path.join(os.path.dirname(__file__), '..', '..', '..'))
 
 
-class TestAlembic(unittest.TestCase):
+class TestAlembic(BasePyTestCase):
```
nphilipp: This wasn't derived from `BaseTestCase` before, i.e. it doesn't need the test database etc. and the incurred overhead which comes with `BasePyTestCase`. Just don't inherit from `unittest.TestCase`, i.e. make it:

````
class TestAlembic:
    ...
````

File: bodhi/tests/server/test_alembic.py [hunk 3]
```diff
@@ -54,5 +55,5 @@ def test_alembic_history(self):
 
         stdout = proc2.communicate()[0]
         stdout = stdout.strip().split(b'\n')
-        self.assertEqual(len(stdout), 1)
+        assert (len(stdout)) == (1)
```
nphilipp: You can get rid of most of the parentheses here:

````
        assert len(stdout) == 1
````

**Review by nphilipp — CHANGES_REQUESTED** [review]

File: requirements.txt (hunk 1 above)
Comment (Fan) [reply requirements.txt]: My bad! Got too used to committing all...

**Comment by Fan** [comment]
Should have updated everything now!

**Review by nphilipp — APPROVED** [review]
Thanks!

File: bodhi/tests/server/test_alembic.py [hunk 4]
```diff
@@ -22,14 +22,15 @@
 
 import os
 import subprocess
-import unittest
+
+from bodhi.tests.server.base import BasePyTestCase
```
ryanlerch:
```suggestion
```

ryanlerch: this looks like it is lo longer used, and importing it is causing the flake8 style tests to fail in the CI.

ryanlerch (changes requested):

ryanlerch (approved):

PR #3621, opened 2019-10-21
List items in RSS field in descending order

mattiaverga:
The RSS renderer should generate items with the most recent on top, so it should reverse the data list in input.


Zlopez (changes requested):
Could you add a news file for towncrier? See https://github.com/fedora-infra/bodhi/blob/develop/docs/developer/index.rst#release-notes

Zlopez (approved):
LGTM

PR #3624, opened 2019-10-21
Drop the use of the 'method' parameter in setup_method(), and require minimum PyTest version.

nphilipp:


nphilipp:
Hmm, it looks as if `setup_method()` still requires the `method` parameter, only for `setup_module()`, the `module` parameter is optional. Oops.

Issue #3625, opened 2019-10-21
Update.pushed needs to be set to True in the signed consumer when an update is moved to testing.


cverna:
A rawhide update is moved from pending to testing in the signed consumer, when moving the update the update.pushed is not set which then makes it impossible to unpush an update from the UI.

StephenCoady:
@cverna @pypingou I think I have added the change necessary locally. Could one of you guys give me permission to assign tickets, please? Pushing branch now and will create a PR shortly.

PR #3628, opened 2019-10-22
set update.pushed to true when moved to testing in the signed consumer

StephenCoady:
Added build.update.pushed = True for the signed consumer so that it can be unpushed. See https://github.com/fedora-infra/bodhi/issues/3625

Signed-off-by: Stephen Coady <scoady@redhat.com>

### DIFF
--- a/bodhi/server/consumers/signed.py
+++ b/bodhi/server/consumers/signed.py
@@ -118,6 +118,7 @@ class SignedHandler(object):
                 build.update.status = UpdateStatus.testing
                 build.update.date_testing = func.current_timestamp()
                 build.update.request = None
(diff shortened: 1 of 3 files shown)


cverna (commented):
Looks great :-) could you just adjust one of the test to make sure that we check that the boolean is correctly set.

Zlopez (changes requested):
Could you add news file. See https://github.com/fedora-infra/bodhi/blob/develop/docs/developer/index.rst#release-notes

cverna (approved):
:+1: thanks

Zlopez (approved):

PR #3629, opened 2019-10-22
Change the message sent when an update is ready to be tested

pypingou:
This commit changes the content of the message sent when an update is
ready to be tested by a CI system to match the schema already used by
the Fedora CI system.

More information about the request can be found at:
https://pagure.io/fedora-ci/general/issue/70

### DIFF
--- a/bodhi/messages/schemas/update.py
+++ b/bodhi/messages/schemas/update.py
@@ -570,7 +570,7 @@ class UpdateRequirementsMetStableV1(UpdateMessage):
         return f'{self.update.alias} has met stable testing requirements'
 
 
-class UpdateReadyForTestingV1(UpdateMessage):
+class UpdateReadyForTestingV1(BodhiMessage):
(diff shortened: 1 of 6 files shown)


Zlopez (commented):
Do we want to have news file for this? Otherwise this looks good

pypingou:
> Do we want to have news file for this? Otherwise this looks good

This is fixing something that was added for 5.0, so I assume it already is listed in the release note.

Zlopez:
According to the nature of the change, this will be probably backwards incompatible for any consumer who consumed this message.

pypingou:
> According to the nature of the change, this will be probably backwards incompatible for any consumer who consumed this message.

Indeed, except for the fact that this message was added in 5.0 which hasn't been released yet :)

And that the topic changed as well, so if you filter for a specific topic, you won't receive the new messages

Zlopez (approved):

PR #3643, opened 2019-10-24
Migrate server/test__init__.py to PyTest

### LINKED ISSUE #3591: Convert bodhi/tests/server/test___init__.py to PyTest
[bodhi/tests/server/test___init__.py](https://github.com/fedora-infra/bodhi/blob/develop/bodhi/tests/server/test___init__.py) still contains classes that derive from `unittest.TestCase`. We use PyTest, so they could be migrated to not derive from `unittest.TestCase`, or derive from `bodhi.tests.server.base.BasePyTestCase` in case they need to use the test database or WSGI app. 

#3612

Zlopez:

Fixes #3591

### DIFF
--- a/bodhi/tests/server/test___init__.py
+++ b/bodhi/tests/server/test___init__.py
@@ -18,7 +18,6 @@
 """This test suite contains tests for bodhi.server.__init__."""
 from unittest import mock
 import collections
-import unittest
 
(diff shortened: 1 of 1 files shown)


ryanlerch (approved):
LGTM

PR #3645, opened 2019-10-24
Migrate scripts/test_initializedb.py to PyTest

### LINKED ISSUE #3587: Convert bodhi/tests/server/scripts/test_initializedb.py to PyTest
[bodhi/tests/server/scripts/test_initializedb.py](https://github.com/fedora-infra/bodhi/blob/develop/bodhi/tests/server/scripts/test_initializedb.py) still contain classes that derive from `unittest.TestCase`. We use PyTest, so they could be migrated to not derive from `unittest.TestCase`, or derive from `bodhi.tests.server.base.BasePyTestCase` in case they need to use the test database or WSGI app. 

#3612

Zlopez:

Fixes #3587


bodhi/tests/server/scripts/test_initializedb.py
```diff
@@ -19,12 +19,12 @@
 
 from io import StringIO
 from unittest import mock
-import unittest
+import pytest
```
ryanlerch, line 23:
```suggestion

import pytest
```

ryanlerch: pytest is a third party module, so we need a newline between it and the standard modules, otherwise it will fail the style tests.

ryanlerch (changes requested):

ryanlerch (approved):

PR #3657, opened 2019-10-25
Fix a XSS issue

abompard:
There is an XSS issue in Bodhi. The production instance in Fedora has been hotfixed.


pypingou (approved):

pypingou:
Working on CSP headers will also help us with this but that's not a small task

drunkwcodes (approved):

PR #3658, opened 2019-10-26
[WIP] Implement Content-Security-Policy header

mattiaverga:
This is mostly based on the old, retired, PR #2533 
It's a work-in-progress and I would like your thoughts before working more on it.

What has been done:

- [x] Set a Content-Security-Policy header
- [x] Setup a random nonce value to allow inline scripts
- [x] Upgrade jquery and chartjs because old versions were causing problems with CSP
- [x] Move all inline styles to use css classes
- [x] Move all href="javascript:" stuff to alternative scripts

What to do:
- [x] ~~Figure out a "script-src" block that still happens, but I can't figure out what's the blocked source~~ that was due to the "plasma integration" Firefox extension - not a Bodhi fault
- [x] Fix charts in the Release page
- [x] Write some tests to ensure CSP and nonce generator work as expected
- [ ] Test!


pypingou:
Thank you very much for starting on this, it is not a small task and seeing the first step done on this is highly appreciated.

I just would like to make sure we do not push this as part of the 5.0 release as that release will contain a lot of changes related to rawhide packages gating and having been through the CSP dance in the pagure project, I know that getting CSP to work properly is going to take us some time and probably a few releases.
So I'd rather we keep the rawhide packages gating features and the CSP features apart until we're sure the first one works fine and we can focus on the second one.

I hope this is ok for you as well :)

mattiaverga:
@pypingou Sure. 
Meanwhile, can I start to push just the inline styles and `javascript:` urls removal and the upgrade of jquery and chartjs in a separate PR to prepare the work? @ryanlerch is it ok for you?

I also need the Chartjs upgrade to polish the Release details page (I would like to give the actual graph a better readability and fix some glitches).

mattiaverga:
FYI I've rebased this old PR to the latest master branch, but the new searchbar component is not compatible with CSP (it uses inline styles, while we may want to disallow that).
So I'll let this as WIP, meanwhile I've filed a bug upstream https://github.com/running-coder/jquery-typeahead/issues/507

PR #3671, opened 2019-10-29
Update ChartJS and redesign Release page.

mattiaverga:
- Update ChartJS
- Move the timeline graph on the right column, so that it is wider and more readable
- Add a color for "enhancement" type to distinguish it from "unspecified" type; also add a test for `type2color()` which wasn't tested at all
- Split tables in the left column, add captions, show all the properties of the release like the CLI output
- Add two new graphs (updates by type and updates by status) and fix links and icons not being showed
- Lastly, add a placeholder in case Javascript is disabled.

immagine


ryanlerch (approved):
Thanks!

This was the one i kinda left alone, thanks for the refresh!

LGTM

pypingou (approved):
If @ryanlerch is fine, so am I :)

PR #3674, opened 2019-10-29
Migrate tasks/test_composer.py to PyTest

### LINKED ISSUE #3590: Convert bodhi/tests/server/tasks/test_composer.py to PyTest
[bodhi/tests/server/tasks/test_composer.py](https://github.com/fedora-infra/bodhi/blob/develop/bodhi/tests/server/tasks/test_composer.py) still contains classes that derive from `unittest.TestCase`. We use PyTest, so they could be migrated to not derive from `unittest.TestCase`, or derive from `bodhi.tests.server.base.BasePyTestCase` in case they need to use the test database or WSGI app. 

#3612

Zlopez:

Fixes #3590

### DIFF
--- a/bodhi/tests/server/tasks/test_composer.py
+++ b/bodhi/tests/server/tasks/test_composer.py
@@ -26,9 +26,9 @@ import os
 import shutil
 import tempfile
 import time
-import unittest
 from http.client import IncompleteRead
 
+import pytest
 from click import testing
 from fedora_messaging import api
 from fedora_messaging.testing import mock_sends
@@ -79,7 +79,7 @@ mock_absent_taskotron_results = {
 }
 
 
-class TestTask(unittest.TestCase):
+class TestTask:
     @mock.patch("bodhi.server.tasks.bugs")
     @mock.patch("bodhi.server.tasks.buildsys")
     @mock.patch("bodhi.server.tasks.initialize_db")
@@ -96,7 +96,7 @@ class TestTask(unittest.TestCase):
         handler.run.assert_called_with(api_version=42, data={"foo": "bar"})
 
 
-class TestCheckpoint(unittest.TestCase):
+class TestCheckpoint:
     """Test the checkpoint() decorator."""
     def test_with_return(self):
         """checkpoint() should raise a ValueError if the wrapped function returns anything."""
@@ -108,10 +108,10 @@ class TestCheckpoint(unittest.TestCase):
             def dont_wrap_me_bro(self):
                 return "I told you not to do this. Now look what happened."
 
-        with self.assertRaises(ValueError) as exc:
+        with pytest.raises(ValueError) as exc:
             TestClass().dont_wrap_me_bro()
 
-        self.assertEqual(str(exc.exception), 'checkpointed functions may not return stuff')
+            assert 'checkpointed functions may not return stuff' in str(exc.value)
 
 
 @mock.patch('bodhi.server.push.initialize_db', mock.MagicMock())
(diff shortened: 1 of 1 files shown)


ryanlerch (approved):

ryanlerch:
the unit tests on this one seem to be failing now :(

nphilipp:
I'll rebase this on the current HEAD to restart the tests.

bodhi/tests/server/tasks/test_composer.py
```diff
@@ -108,10 +108,10 @@ def __init__(self):
             def dont_wrap_me_bro(self):
                 return "I told you not to do this. Now look what happened."
 
-        with self.assertRaises(ValueError) as exc:
+        with pytest.raises(ValueError) as exc:
             TestClass().dont_wrap_me_bro()
 
-        self.assertEqual(str(exc.exception), 'checkpointed functions may not return stuff')
+        assert 'checkpointed functions may not return stuff' in str(exc)
```
nphilipp: The `pytest.raises()` context manager is a wrapper around a potentially thrown exception similar to `unittest.TestCase.assertRaises()`, the assertion should look like this:

````
    assert 'checkpointed functions may not return stuff' in exc.value
````

This needs to be fixed in a couple other cases below, too.

File: bodhi/tests/server/tasks/test_composer.py [hunk 2]
```diff
@@ -170,7 +170,6 @@ def setUp(self):
         mock_config = mock.patch.dict(
             'bodhi.server.tasks.composer.config', test_config)
         mock_config.start()
-        self.addCleanup(mock_config.stop)
```
nphilipp, line 173: This should be moved into `teardown_method()`, like this:

````
    def setup_method(self, method):
        ...
        self.mock_config = mock.patch.dict(...)
        self.mock_config.start()
    ...
    def teardown_method(self, method):
        super().teardown_method(method)
        self.mock_config.stop()
        ...
````

File: bodhi/tests/server/tasks/test_composer.py [hunk 3]
```diff
@@ -2344,14 +2324,14 @@ def test_request_not_stable(self, Popen):
                        'testcontainer2:1.0.1-1.fc28container'):
             for dtag in [source.split(':')[1], source.split(':')[1].split('-')[0], 'testing']:
                 mock_call = mock.call(
-                    [config['skopeo.cmd'], 'copy',
+                    [config['skopeo.cmd'], 'copy', '--dest-tls-verify=false',
```
nphilipp: This looks like it's mistakenly copied from `test_skopeo_extra_copy_flags()`. It breaks this and a couple of other tests for me, at any rate. :wink:

nphilipp (changes requested):
Here are the changes I made to fix the unit tests:

[PR3590-changes.patch.txt](https://github.com/fedora-infra/bodhi/files/3824115/PR3590-changes.patch.txt)

bodhi/tests/server/tasks/test_composer.py
Zlopez: Thanks, I couldn't find out how to call this.

bodhi/tests/server/tasks/test_composer.py
Zlopez: I noticed this today, when checking other PRs.

bodhi/tests/server/tasks/test_composer.py
Zlopez: I will remove it than.

bodhi/tests/server/tasks/test_composer.py
```diff
@@ -2656,10 +2642,8 @@ def test_primary_arches_undefined(self):
 
         url = t._get_master_repomd_url('aarch64')
 
-        self.assertEqual(
-            url,
+        assert url ==\
```
nphilipp: Please add a space between `==` and the backslash (see PEP8: "Always surround these binary operators with a single space on either side..."), here and in similar code below. Sorry for not spotting this earlier.

nphilipp (changes requested):

bodhi/tests/server/tasks/test_composer.py
Zlopez: Why flake8 didn't catch this?

Zlopez: This should be fixed now, I found a few more in the code so I fixed them too.

nphilipp: I think flake8 didn't catch this because of the backslash line-continuation.

nphilipp (approved):

PR #3677, opened 2019-10-30
Do not double enquote the update title

pypingou:
Basically, if we specify amp=True on update.beautify_title(), it will
replace the & sign with its html quoted version: &amp;
But since we're encoding the output of this function into html again,
the & in &amp; get encoded again resulting in a "&amp;amp;" string which
is not valid html and is breaking the test suite.

Thanks to Nils for getting the idea of what is going on!

### DIFF
--- a/bodhi/server/models.py
+++ b/bodhi/server/models.py
@@ -2672,8 +2672,8 @@ class Update(Base):
 
             "package1, package, 2 and XXX more"
 
-        If the "amp" parameter is specified it will replace the and with and
-        &amp; html entity
+        If the "amp" parameter is specified it will replace the "and" with an
+        "&" entity.
 
         If the "nvr" parameter is specified it will include name, version and
         release information in package labels.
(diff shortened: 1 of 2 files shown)


Zlopez (approved):
You are missing some tests, otherwise this looks like it's solving the integration tests issues.

Zlopez:
At least some of them are now solved. F29 and F30 are still failing :-(

Zlopez:
Oh no, all the integration tests are still failing, but with different issue now. So at least the double quote issue is gone.

pypingou:
I've dropped the possibility to get ``&amp;`` entirely from the model as all the places where this is used are in templates where they are then escape again via the ``|h``.

pypingou:
Btw, you asked earlier what does ``|h``, this is its definition:
````
h : HTML escaping, provided by markupsafe.escape(string)
````
Source: https://docs.makotemplates.org/en/latest/filtering.html?highlight=escape#expression-filtering

Zlopez:
> Btw, you asked earlier what does `|h`, this is its definition:
> 
> ```
> h : HTML escaping, provided by markupsafe.escape(string)
> ```
> 
> Source: https://docs.makotemplates.org/en/latest/filtering.html?highlight=escape#expression-filtering

I did? I don't remember, but thanks for explanation.

pypingou:
@Zlopez sorry that was @nphilipp who did

Issue #3641, opened 2019-10-24
Split up DB models module


nphilipp:
## Story

As a developer contributing to Bodhi,

I'd like the `bodhi/server/models.py` module to be converted into a package, with one module per DB model class (group) and its tests split up likewise,

so that it's easier to work with.

## Background

Currently, `models.py` is by far our largest Python module, with a size of 180kB and almost 4700 lines. It contains 21 database model classes, along with auxiliary enumeration types, etc. and is pretty easy to get lost in. Likewise, its unit test module `test_models.py` is almost the same size.

## Acceptance Criteria

- [ ] `models.py` is converted into a `database` package, split up along the lines of base classes and auxiliary code and DB models (together with their respective Enum types)
- [ ] the package exposes all classes used elsewhere in the project, i.e. `__init__.py` imports them, so that consumers of the code have to be changed only minimally
- [ ] `test_models.py` is split up similarly

mattiaverga:
@nphilipp I'd like to work on this, if you don't plan to do it yourself.

chunkSpadunk:
Is this already fixed I would like to work on this if the issue is still open

Solar-Rays:
Can I work on this issue if it is still open?

ssemanich:
can i work on this issue

mattiaverga:
Hi @Solar-Rays and @ssemanich, yes the issue is still open. I'll assign it to Solar-Rays, since they expressed their interest earlier, but since it is quite a big task maybe you can coordinate and work on it together.

Solar-Rays:
Thank you. @ssemanich we can work on this together for sure. How best would you like to collaborate on this? Via Github here or on a different platform? 

Please do let me know. Cheers

ssemanich:
i think via github

ssemanich:
@nphilipp and @mattiaverga
 am stack somewhere how should i separate the db. models classes from base classes. Another thing is should the enums types also be in the db.model classes or they should have their file

nphilipp:
Hi! And sorry for the late response.

Contrary to what I wrote over 4 years ago 😉, I would refactor `models.py` a bit more, i.e. make a `database` package that contains different modules and subpackages – of course, this requires adaptation in the current consumers of `models.py`. I’ve updated the original comment accordingly.

To begin, here’s the list of classes in the module:

```
nils@makake:~/src/fedora-infra/bodhi (develop)> grep ^class bodhi-server/bodhi/server/models.py 
[...]

PR #3681, opened 2019-11-02
Convert models module into a package

mattiaverga:
As in #3641 I've split up the models.py module into a package.

This has been forked from 44a4b71efabaaa81fd3725dd5601fab6ef4504fd. Since models.py has been removed, the changes made to it after the base commit will not be automatically inherited. So this PR will need to be updated with the latest changes just before the merge.

I would suggest to postpone the tests split in a successive PR, so that we're sure everything is right in this first part.
BTW, I also found in several places in the tests and in `bodhi/server/views/generic.py` that several modules were referenced from models instead of being imported from themselves (i.e. `models.Session` instead of `bodhi.server.session`, or `bodhi.server.models.mail` instead of `bodhi.server.mail`): I have corrected that.


mattiaverga:
JENKINS_TEST

mattiaverga:
jenkies test

mattiaverga:
mmm tests are failing because `bodhi/tests/server/test_models.py::TestUpdateEdit::test_new_builds_log_when_release_has_no_signing_tag` hits timeout without any clue.
That's odd, testing locally with pytest everything passes. Any idea?

mattiaverga:
It seems that `bodhi/tests/server/test_models.py::TestUpdateEdit::test_new_builds_log_when_release_has_no_signing_tag` reaches timeout because `handle_update.delay()` never gets executed by Celery. This is very similar to what I described in #3698 .
Running the tests directly in the Vagrant environment gives no problem, the test only gets stuck in CI, which have a different broker URL in Celery config file. However, I can't explain why it only happens with the models module splitted.

mattiaverga:
Retiring since it's severely outdated

PR #3691, opened 2019-11-06
Speed up hide_existing=true when listing candidates

### LINKED ISSUE #3687: Call to get candidates in new update form is now slow
Previously, the ajax call we used to populate the builds chooser in the new update form took a couple seconds to load. However, since the introduction of this fix:

https://github.com/fedora-infra/bodhi/pull/3670

this call now takes upwards of 30seconds on stg. (not sure if this will be quicker on prod, but it is pretty slow atm.) the endpoint is : https://bodhi.stg.fedoraproject.org/latest_candidates?hide_existing=true&prefix=

not sure what is taking so long here, it doenst look like there are any additional calls to koji that is slowing it down...

nphilipp:
Previously, the database was queried for each build returned from koji, which could result in thousands of individual queries.

Instead, query the database once for known builds associated with an update and filter out returned builds found in this set. This set isn't filtered for active releases because a build might be associated with an update for an archived release but still inherited into one of the tags we query Koji for. It performs well enough at the moment but the set can easily be cached if not.

Fixes: #3687

------

This PR also reverts the respective workaround. To test the performance, configure your testing instance to query koji.stg instead of using the mocked instance, e.g. by using `vagrant up --use-staging`  or having something like this in your `development.ini` (substitute your FAS user in `krb_principal`, also you might need to install `krb5-workstation` and acquire a matching ticket in the Vagrant box):

````
[app:main]
buildsystem = koji
koji_hub = https://koji.stg.fedoraproject.org/kojihub
koji_web_url = https://koji.stg.fedoraproject.org/koji/
krb_principal = nphilipp@STG.FEDORAPROJECT.ORG
acl_system = pagure
pagure_url = https://src.stg.fedoraproject.org/
...
````

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -15,7 +15,7 @@ width = 80
 show-source = True
 max-line-length = 100
 exclude = .git,.tox,dist,*egg,build,tools
-ignore = E712,W503
+ignore = E711,E712,W503
 # Configure flake8-import-order
 application-import-names = bodhi
 import-order-style = pep8
--- a/bodhi/server/static/js/update_form.js
+++ b/bodhi/server/static/js/update_form.js
@@ -200,7 +200,7 @@ $(document).ready(function() {
                     $builds_search_selectize.disable()
                 }
                 $.ajax({
-                    url: '/latest_candidates?hide_existing=false&prefix=' + encodeURIComponent(query),
+                    url: '/latest_candidates?hide_existing=true&prefix=' + encodeURIComponent(query),
                     type: 'GET',
                     error: function() {
                         messenger.post({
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

**Comment by abompard** [comment]
This looks sane. There's a minor flake8 error, and also I think it'd be useful to add a test to cover the case you describe, where a build in an archived release is inherited, so that we don't accidentally change that later (like, when looking for more optimizations... ;-) )

**Comment by nphilipp** [comment]
> I think it'd be useful to add a test...

Agreed.

File: setup.cfg [hunk 1]
```diff
@@ -15,7 +15,7 @@ width = 80
 show-source = True
 max-line-length = 100
 exclude = .git,.tox,dist,*egg,build,tools
-ignore = E712,W503
+ignore = E711,E712,W503
```
abompard, line 18: Hmm I'm not sure we should disable it globally. Does SQLAlchemy accept the `is not None` construct? If it does we can use it, if not I think it's better to exclude that check on the specific line.

ryanlerch (approved):

PR #3694, opened 2019-11-08
Output the update install command in bug comments.

### LINKED ISSUE #740: Change to bugzilla updates text for quicker testing feedback
Forwarded from https://fedorahosted.org/bodhi/ticket/791

Based on the fedora-devel discussion ("Bodhi front page after login"):

What about scrapping the existing bugzilla message, and instead posting the direct link to koji to download the package (as soon as the package is signed):

"""To test the update now run:

dnf upgrade ​https://koji.fp.o/update-yyyy/path-to.rpm ​https://koji.fp.o/update-yyyy/path-to-another.

rpm ​https://koji.fp.o/update-yyyy/path-to-yet-another.rpm

After that go to ​https://bodhi.fp.o/update-yyyy/ to leave karma. """

dnf will skip packages that are not already installed with 'upgrade'. Since this would be only run manually, and the dependent packages would be downloaded from mirrors, this shouldn't cause too much load on koji servers. The advantage would be that the message would be actionable and feedback from the bug reporters would be as fast as possible.

mattiaverga:
- provide update install command directly in bugs comments (if `update.install_command` is available).
- change the format of `stable_bug_msg` and `testing_bug_msg` so that they now uses some placeholders instead of a bunch of `%s` (this is highlighted in the release notes, so that if a user has set custom messages in their Bodhi instance, they need to correct them during the upgrade).
- `testing_bug_msg` is no more a suffix of `stable_bug_msg`.
- raise a ValueError if we try to post a default message to a bug associated to an update which isn't in Stable or Testing status.

Here it is an example of a bug comment:
```
FEDORA-2019-53a68bf913 has been pushed to the Fedora 31 testing repository.
In short time you'll be able to install the update with the following command:
`sudo dnf upgrade --enablerepo=updates-testing --advisory=FEDORA-2019-53a68bf913`
You can provide feedback for this update here: http://localhost:6543/updates/FEDORA-2019-53a68bf913

See also https://fedoraproject.org/wiki/QA:Updates_Testing for more information on how to test updates.
```

Fixes #740


Zlopez (approved):
Nice change
:+1:  for the news file

PR #3695, opened 2019-11-10
removed classes unittest.TestCase

### LINKED ISSUE #3605: Convert bodhi/tests/messages/schemas/test_update.py to BaseTestCase
[bodhi/tests/messages/schemas/test_update.py](https://github.com/fedora-infra/bodhi/blob/develop/bodhi/tests/messages/schemas/test_update.py) still contains classes that derive from unittest.TestCase. We use pytest, so they could be migrated to derive from bodhi.tests.server.base.BaseTestCase.

This also allows the use of the simpler assert constructs instead of the self.assert*() methods.

#3574

subhamkrai:
Signed-off-by: subhamkrai <subhamkumarrai03@gmail.com>

Fixes: #3605

### DIFF
--- a/bodhi/tests/messages/schemas/test_update.py
+++ b/bodhi/tests/messages/schemas/test_update.py
@@ -15,7 +15,6 @@
 # 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 """Unit tests for the update message schemas."""
 
-import unittest
 
 from bodhi.messages.schemas.base import BuildV1, ReleaseV1, UpdateV1, UserV1
 from bodhi.messages.schemas.update import (
@@ -36,7 +35,7 @@ from bodhi.messages.schemas.update import (
 from bodhi.tests.messages.utils import check_message
 
 
-class UpdateMessageTests(unittest.TestCase):
+class TestUpdateMessage:
     """A set of unit tests for classes in :py:mod:`bodhi_messages.schemas.update`"""
 
     def test_eject_v1(self):


bodhi/tests/messages/schemas/test_update.py
```diff
@@ -36,7 +36,7 @@
 from bodhi.tests.messages.utils import check_message
 
 
-class UpdateMessageTests(unittest.TestCase):
+class UpdateMessageTests:
```
nphilipp: In order to be recognized as a test, the class name has to start with `Test`. Would you rename it to `TestUpdateMessage`? Thanks!

nphilipp (changes requested):

bodhi/tests/messages/schemas/test_update.py
subhamkrai: Yes sure I'll

bodhi/tests/messages/schemas/test_update.py
```diff
@@ -15,7 +15,7 @@
 # 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 """Unit tests for the update message schemas."""
 
-import unittest
+
```
nphilipp, line 18: I don't know why I didn't spot this earlier: this additional empty line makes flake8 complain (which is run during our automatic tests) -- there need to be two blank lines between the docstring of a module and the imports, this makes it three. Would you please remove it? Thanks!

nphilipp (changes requested):

nphilipp (approved):
LGTM, thanks!

PR #3067, opened 2019-03-17
Remove unused comment in home.html

sebwoj:
Signed-off-by: Sebastian Wojciechowski <s.wojciechowski89@gmail.com>

PR #3699, opened 2019-11-10
Remove unused badge_ids setting from config

sebwoj:
This commit removes config setting used in the commented code, which in turn was removed in PR #3067.


Zlopez (approved):
This value is only in config, so no harm removing it.

PR #3500, opened 2019-09-12
Create a Dashboard for logged in users and refresh the webUI

ryanlerch:
This re-arranges the webUI a little bit to make way for a
dashboard page for logged in users, that shows their active
updates and overrides. When logged in, the home page looks like this:

screencapture-0-0-0-0-6543-2019-09-12-15_58_29

and when not logged in, like this:

screencapture-0-0-0-0-6543-2019-09-12-16_00_12

The updates list page is also refreshed, looking like this:
screencapture-0-0-0-0-6543-updates-2019-09-12-15_59_31

And all the counts on the front page are now moved to the Releases page, and a releases menuitem now in the top menu:
screencapture-0-0-0-0-6543-releases-2019-09-12-15_58_57

The actual page for an update has also been refreshed, removing a lot of the colours that were used for everything, with colour only really used for karma, and the status of the update itself.

screencapture-0-0-0-0-6543-updates-FEDORA-2019-095d8600a1-2019-09-12-16_13_14

PR #3700, opened 2019-11-10
Remove unused max_update_length_for_ui setting from config

sebwoj:
This commit removes config setting used in the code removed in PR #3500.


Zlopez (approved):
I can see that this is really only in the config, so no harm removing this.

PR #3702, opened 2019-11-11
Test overrides rss path in CI

sebwoj:


Zlopez (approved):
Coverage raising is always good

PR #3709, opened 2019-11-12
Make test_push_composer_start_more reliable

### LINKED ISSUE #3698: CI failings, probably celery worker isn't started
I did some investigation on the CI failings, it seems that Celery worker isn't started. I think that `/etc/bodhi/start.sh` is not executed when docker image changes from `f31-integration-build-bodhi` to `f31-integration`
https://github.com/fedora-infra/bodhi/blob/6794e5b9b5b2ecca77e43e7df596e8b569b84448/devel/ci/integration/bodhi/Dockerfile-f31#L65
I tried to change that from CMD to ENTRYPOINT, but it never executes - I cannot even see a simple echo command.

That leads the 
https://github.com/fedora-infra/bodhi/blob/6794e5b9b5b2ecca77e43e7df596e8b569b84448/devel/ci/integration/tests/test_bodhi_tasks.py#L48
line fail for timeout.
Note that the other tasks that rely on celery `.delay()` are not tested for a successful completion so the other tests don't fail...

AdamSaleh:
The test relies on having pending composes that will run pungi

We only check there are composes, we should be checking that their type
implies running pungi as well, so either module or rpm content_type.
I hope adding more thorough query should help with the test.

The new query:

```
select distinct
        c.release_id, c.request, state, b.type
      from
        composes as c
      inner join updates as u on (c.release_id = u.release_id and c.request = u.request )
      inner join builds as b on (b.update_id = u.id)
      where b.type in ('rpm', 'module') and state <> 'failed';
```

On failure, it will print the result of the query to test-log.

Fixes #3698

 Maybe :smile:

### DIFF
--- a/devel/ci/integration/tests/test_bodhi_tasks.py
+++ b/devel/ci/integration/tests/test_bodhi_tasks.py
@@ -35,17 +35,33 @@ def test_push_composer_start(bodhi_container, db_container, rabbitmq_container):
     output = "".join(line.decode("utf-8") for line in output)
     assert output.endswith("Requesting a compose\n")
     # Check how many composes should be done
-    query = "SELECT COUNT(*) FROM composes"
+    # because we later check for pungi.log, we need to see if there are composes that apply
+    # we can get these from the type field of the builds table,
+    # after joining with the composes table through updates table.
+    query = """
+      SELECT DISTINCT
+        c.release_id, c.request, state, b.type
+      FROM
+        composes AS c
+      INNER JOIN updates AS u ON (c.release_id = u.release_id AND c.request = u.request )
+      INNER JOIN builds AS b ON (b.update_id = u.id)
+      WHERE b.type IN ('rpm', 'module') AND state <> 'failed';
+    """
     db_ip = db_container.get_IPv4s()[0]
     conn = psycopg2.connect("dbname=bodhi2 user=postgres host={}".format(db_ip))
     with conn:
         with conn.cursor() as curs:
             curs.execute(query)
-            composes_count = curs.fetchone()[0]
-    if composes_count == 0:
+            composes = curs.fetchall()
+    if not composes:
         pytest.skip("We can't test whether composes were run, there are none pending")
+
     # Give some time for the message to go around and the command to be run.
-    wait_for_file(bodhi_container, "/tmp/pungi-calls.log")
+    try:
+        wait_for_file(bodhi_container, "/tmp/pungi-calls.log")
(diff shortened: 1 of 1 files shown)


devel/ci/integration/tests/test_bodhi_tasks.py
```diff
@@ -35,17 +35,34 @@ def test_push_composer_start(bodhi_container, db_container, rabbitmq_container):
     output = "".join(line.decode("utf-8") for line in output)
     assert output.endswith("Requesting a compose\n")
     # Check how many composes should be done
-    query = "SELECT COUNT(*) FROM composes"
+    # because we later check for pungi.log, we need to see if there are composes that apply
+    # we can get these from the type field of the builds table,
+    # after joining with the composes table through updates table.
+    query = """
```
Zlopez, line 41: Could you write the SQL keywords in uppercase, so it is more readable?

devel/ci/integration/tests/test_bodhi_tasks.py
```diff
@@ -35,17 +35,34 @@ def test_push_composer_start(bodhi_container, db_container, rabbitmq_container):
...
+        c.release_id, c.request, state, b.type
+      from
+        composes as c
+      inner join updates as u on (c.release_id = u.release_id and c.request = u.request )
+      inner join builds as b on (b.update_id = u.id)
+      where b.type in ('rpm', 'module') and state <> 'failed';
+    """
     db_ip = db_container.get_IPv4s()[0]
     conn = psycopg2.connect("dbname=bodhi2 user=postgres host={}".format(db_ip))
+    print("dbname=bodhi2 user=postgres host={}".format(db_ip))
```
Zlopez: This is debug print or do you have this here on purpose?

Zlopez (commented):
Few comments

devel/ci/integration/tests/test_bodhi_tasks.py
AdamSaleh: Yes, this is a left-over debug print.

devel/ci/integration/tests/test_bodhi_tasks.py
AdamSaleh: I am actually used to writing sql lower-case, but I can change it :)

Zlopez: It's usually more readable uppercase, especially inside python string.

Zlopez (approved):
:+1:

PR #3712, opened 2019-11-13
util/generate_changelog: Build#get_latest can return None

relrod:
Handle this case and default the last build time to 0 to prevent a crash.

I'd like to point out that this is why gradual type systems do not work.

### DIFF
--- a/bodhi/server/util.py
+++ b/bodhi/server/util.py
@@ -121,14 +121,14 @@ def generate_changelog(build: 'models.Build') -> typing.Optional[str]:
         return None
 
     # Grab the RPM header of the previous update, and generate a ChangeLog
-    oldh = get_rpm_header(lastpkg)
-    oldtime = oldh['changelogtime']
-    text = oldh['changelogtext']
+    oldtime = 0
+    if lastpkg is not None:
+        oldh = get_rpm_header(lastpkg)
+        oldtime = oldh['changelogtime']
+        text = oldh['changelogtext']
 
-    if not text:
-        oldtime = 0
-    elif isinstance(oldtime, list):
-        oldtime = oldtime[0]
+        if text and isinstance(oldtime, list):
+            oldtime = oldtime[0]
 
     return build.get_changelog(oldtime)
 
--- a/bodhi/server/util.py
+++ b/bodhi/server/util.py
@@ -121,17 +121,19 @@ def generate_changelog(build: 'models.Build') -> typing.Optional[str]:
         return None
 
     # Grab the RPM header of the previous update, and generate a ChangeLog
-    if lastpkg is not None:
+
+    def _get_oldtime(lastpkg):
+        if lastpkg is None:
+            return 0
         oldh = get_rpm_header(lastpkg)
+        if not oldh['changelogtext']:
+            return 0
         oldtime = oldh['changelogtime']
-        text = oldh['changelogtext']
-
-        if not text:
-            oldtime = 0
-        elif isinstance(oldtime, list):
+        if isinstance(oldtime, list):
             oldtime = oldtime[0]
+        return oldtime
 
-    return build.get_changelog(oldtime)
+    return build.get_changelog(_get_oldtime(lastpkg))
 
 
 def build_evr(build):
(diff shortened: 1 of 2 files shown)


cverna:
@relrod thanks for this, would you like to work on adding tests for that change ? If not I can have a look at it :small_airplane:

nphilipp:
> I'd like to point out that this is why gradual type systems do not work.

Heh, we don't use type hints as a way to do static typing in Python, more as a formalized way of documenting what types are expected to be passed into callables and what is returned. We're only gradually introducing type hints rather than mentioning the types in docstrings with new code (or when someone does it passing by), therefore we have a mix of both. And like all kinds of documentation, it can be wrong or incomplete. ;)

bodhi/server/util.py
```diff
@@ -121,14 +121,14 @@ def generate_changelog(build: 'models.Build') -> typing.Optional[str]:
...
-        oldtime = 0
-    elif isinstance(oldtime, list):
-        oldtime = oldtime[0]
+    oldtime = 0
+    if lastpkg is not None:
+        oldh = get_rpm_header(lastpkg)
+        oldtime = oldh['changelogtime']
+        text = oldh['changelogtext']
+
+        if text and isinstance(oldtime, list):
```
abompard: This fails in the unit tests. If `text` is falsy but `oldtime` is a list, the first element will not be extracted and the program will crash in `get_changelog()` because it does not expect a list.

abompard (changes requested):

bodhi/server/util.py
relrod: That might be true, but that's how it was in the old Bodhi before this crash was introduced.

https://github.com/fedora-infra/bodhi/commit/54210f7e1932b763103ad0a38c3bb818862bc081#diff-2a0b3d27a369a1dd870f69deba3a9b40L344-L352

If `text` is falsy there, it defaults to 0 (which I do above in this patch), otherwise if it's truey and `oldtime` is a list, it takes the head of it.

abompard: Thanks. FYI my hotfix looks like this:
```diff
diff --git a/bodhi/server/util.py b/bodhi/server/util.py
index aa611a8c5..1e2232aef 100644
--- a/bodhi/server/util.py
+++ b/bodhi/server/util.py
@@ -121,14 +121,13 @@ def generate_changelog(build: 'models.Build') -> typing.Optional[str]:
         return None
 
     # Grab the RPM header of the previous update, and generate a ChangeLog
-    oldh = get_rpm_header(lastpkg)
-    oldtime = oldh['changelogtime']
-    text = oldh['changelogtext']
-
-    if not text:
-        oldtime = 0
-    elif isinstance(oldtime, list):
-        oldtime = oldtime[0]
+    oldtime = 0
+    if lastpkg is not None:
+        oldh = get_rpm_header(lastpkg)
+        if oldh['changelogtext']:
+            oldtime = oldh['changelogtime']
+            if isinstance(oldtime, list):
+                oldtime = oldtime[0]
 
     return build.get_changelog(oldtime)
 
```
It's basically the same except I only overwrite oldtime with changelogtime if it's necessary. A cleaner solution would be to have a small `_getoldtime()` function, but eh :-)

bodhi/server/util.py
```diff
@@ -121,13 +121,14 @@ def generate_changelog(build: 'models.Build') -> typing.Optional[str]:
         return None
 
     # Grab the RPM header of the previous update, and generate a ChangeLog
-    oldtime = 0
```
abompard: Hmm, I think this should stay, no? If you don't set it here, when `lastpkg` is `None`, `oldtime` will be unset, no?

abompard (changes requested):

bodhi/server/util.py
relrod: Bleh, just go with what you wrote above.

relrod: (Yes you're right, but just go with what you wrote, it's cleaner)

abompard:
I took the liberty of fixing this and adding a few unit tests, I hope you don't mind Ricky. We currently have some bad issues in prod and we'd like to cut a new release ASAP, and this fix need to be in. Thanks for pointing it out!

abompard (approved):

PR #3722, opened 2019-11-14
Web UI for triggering tests for update

AdamSaleh:
Added a simple modal in UI to trigger update_trigger_tests, so that u…sers can rerun tests in ui.

Signed-off-by: Adam Saleh <asaleh@redhat.com>


bodhi/server/util.py
```diff
@@ -726,6 +726,19 @@ def can_waive_test_results(context, update):
...
 
+def can_trigger_tests(context, update):
+    """
+    Return True or False if we should be able to trigger tests.
+
+    Args:
+        context (mako.runtime.Context): The current template rendering context. Unused.
+        update (bodhi.server.models.Update): The Update on which we are going to waive test results.
+    Returns:
+        bool: Indicating if the test results can be waived on the given update.
```
sebwoj: Let's change this description to reflect real meaning of return value.

bodhi/server/templates/update.html
```diff
@@ -1090,6 +1121,39 @@ <h3 class="modal-title" id="waiveModalLabel">Waive All Test Results</h3>
       });
     })
 % endif
+
+% if self.util.can_trigger_tests(update):
```
sebwoj, line 1127: Instead of creating additional function, we can use:
`if request.registry.settings.get('test_gating.required')`

sebwoj (changes requested):

StephenCoady (approved):

PR #3758, opened 2019-11-14
Convert bodhi/tests/server/functional/test_users.py to PyTest

### LINKED ISSUE #3728: Convert bodhi/tests/server/functional/test_users.py to PyTest
[`bodhi/tests/server/functional/test_users.py`](https://github.com/fedora-infra/bodhi/blob/develop/bodhi/tests/server/functional/test_users.py) still contains classes that derive from `unittest.TestCase`. We use PyTest, so they could be migrated to not derive from `unittest.TestCase`, or derive from `bodhi.tests.server.base.BasePyTestCase` in case they need to use the test database or WSGI app.

In summary:

sebwoj:
fix #3728

### DIFF
--- a/bodhi/tests/server/functional/test_users.py
+++ b/bodhi/tests/server/functional/test_users.py
@@ -23,10 +23,10 @@ from bodhi.server.models import Update, User
 from bodhi.tests.server import base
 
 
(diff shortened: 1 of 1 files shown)


nphilipp (approved):
Thanks!

PR #3761, opened 2019-11-14
Convert tests/server/consumers/test_consumers.py to PyTest

### LINKED ISSUE #3725: Convert bodhi/tests/server/consumers/test_consumers.py to PyTest
[`bodhi/tests/server/consumers/test_consumers.py`](https://github.com/fedora-infra/bodhi/blob/develop/bodhi/tests/server/consumers/test_consumers.py) still contains classes that derive from `unittest.TestCase`. We use PyTest, so they could be migrated to not derive from `unittest.TestCase`, or derive from `bodhi.tests.server.base.BasePyTestCase` in case they need to use the test database or WSGI app.

In summary:

sebwoj:
fix #3725

### DIFF
--- a/bodhi/tests/server/consumers/test_consumers.py
+++ b/bodhi/tests/server/consumers/test_consumers.py
@@ -1,4 +1,4 @@
-# Copyright © 2019 Red Hat, Inc.
+# Copyright © 2019 Red Hat, Inc. and others.
 #
 # This file is part of Bodhi.
 # This program is free software; you can redistribute it and/or
@@ -17,19 +17,19 @@
 """Test the bodhi.server.consumers package."""
 from unittest import mock
 
+import pytest
 from fedora_messaging.api import Message
 from fedora_messaging.exceptions import Nack
 
 from bodhi.server import config
 from bodhi.server.consumers import Consumer, HandlerInfo, signed
-from bodhi.tests.server import base
 
 
 @mock.patch.dict(
     config.config,
     {'pungi.cmd': '/usr/bin/true', 'compose_dir': '/usr/bin/true',
      'compose_stage_dir': '/usr/bin/true'})
-class TestConsumer(base.BaseTestCase):
+class TestConsumer:
     """Test class for the Consumer class."""
 
     @mock.patch('bodhi.server.consumers.bugs.set_bugtracker')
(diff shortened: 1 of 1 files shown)


bodhi/tests/server/consumers/test_consumers.py
```diff
@@ -29,7 +30,7 @@
     config.config,
     {'pungi.cmd': '/usr/bin/true', 'compose_dir': '/usr/bin/true',
      'compose_stage_dir': '/usr/bin/true'})
-class TestConsumer(base.BaseTestCase):
+class TestConsumer:
```
nphilipp, line 32: These tests don't need the test database or the web app. Good catch!

Please remove the `from bodhi.tests.server import base` line above, as it's not needed anymore:

````
20:33:34  ./bodhi/tests/server/consumers/test_consumers.py:26:1: F401 'bodhi.tests.server.base' imported but unused
20:33:34  from bodhi.tests.server import base
20:33:34  ^
````

File: bodhi/tests/server/consumers/test_consumers.py [hunk 2]
```diff
@@ -40,8 +41,8 @@ def test__init___(self, info, initialize_db, setup_buildsystem, set_bugtracker):
         """Test the __init__() method."""
         consumer = Consumer()
 
-        self.assertTrue(any(x.name == 'Signed' and isinstance(x.handler, signed.SignedHandler)
-                            for x in consumer.handler_infos))
+        assert any(x.name == 'Signed' and isinstance(x.handler, signed.SignedHandler) \
+                            for x in consumer.handler_infos)
```
nphilipp: Remove the backslash at the end of the line and make the indent of the second line match up (`for x` should be flush with `x.name` above):

````
20:33:34  ./bodhi/tests/server/consumers/test_consumers.py:44:87: E502 the backslash is redundant between brackets
20:33:34          assert any(x.name == 'Signed' and isinstance(x.handler, signed.SignedHandler) \
20:33:34                                                                                        ^
20:33:34  ./bodhi/tests/server/consumers/test_consumers.py:45:29: E127 continuation line over-indented for visual indent
20:33:34                              for x in consumer.handler_infos)
20:33:34                              ^
````

nphilipp (changes requested):

nphilipp (approved):
Looks good to me, thanks!

PR #3774, opened 2019-11-18
Add back the ability to add abitairy text as a build

### LINKED ISSUE #3707: Bodhi 5.0: new way to add build is terrible
Please, switch back to previous way

Being able to simply add build is important.

Current version is slow, and don't work well when trying to paste package name in input.

Old way is also better because allow to copy/paste NEVR during the build, without anything tagged yet.

### LINKED ISSUE #3765: No longer able to enter the NVR of a build in the create update dialog
In the past one could just enter the NVR of a build into an update without waiting for it to query koji for available builds.

At the moment I can't edit an update because the query to koji is repeatedly timing out. When the timeout occurs you have to just refresh the page and hope you get lucky and the query completes, but it turns filing an update via the web UI into a multi-minute event which is very annoying.

ryanlerch:
the call to get the list of candidate builds from koji
seems to be timing out a lot, so adding back the ability
to just add an abitairy build to the builds list.

Fixes: #3765 
Fixes: #3707

### DIFF
--- a/bodhi/server/static/js/update_form.js
+++ b/bodhi/server/static/js/update_form.js
@@ -131,19 +131,7 @@ $(document).ready(function() {
         var $builds_search_selectize = $('#builds-search').selectize({
             valueField: 'nvr',
             labelField: 'nvr',
-            create: function(input, callback){
-                if (input in $builds_search_selectize.options){
-                    callback($builds_search_selectize.options[input])
-                } else {
-                    messenger.post({
-                        message: input+' is not tagged in koji as a candidate build',
-                        type: 'error',
-                    });
-                    callback();
-                }
-                console.log();
-                callback()
-            },
+            create: true,
             searchField: ['nvr', 'tag_name', 'owner_name'],
             preload: true,
             plugins: ['remove_button','restore_on_backspace'],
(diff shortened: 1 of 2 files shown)


bodhi/server/static/js/update_form.js
```diff
@@ -195,12 +183,12 @@ $(document).ready(function() {
                 $('#builds-search-selectized').attr("placeholder", "");
             },
             load: function(query, callback) {
-                if (query == ""){
-                    console.log("query -"+query+"-");
-                    $builds_search_selectize.disable()
-                }
+                //if (query == ""){
```
cverna: Should we remove this commented lines ? Or is it valuable to keep ?

bodhi/server/static/js/update_form.js
```diff
@@ -211,7 +199,7 @@ $(document).ready(function() {
                     },
                     success: function(res) {
                         $('#builds-search-selectized').attr("placeholder", "search and add builds");
-                        $builds_search_selectize.enable()
+                        // $builds_search_selectize.enable()
```
cverna: Same here :-)

bodhi/server/static/js/update_form.js
```diff
@@ -179,7 +167,7 @@ $(document).ready(function() {
             },
             onInitialize: function(){
                 // make sure the placeholder shows when items already exist when page loads
-                $('#builds-search-selectized').attr("placeholder", "loading candidate builds...");
+                // $('#builds-search-selectized').attr("placeholder", "loading candidate builds...");
```
cverna: And here too ?

bodhi/server/static/js/update_form.js
ryanlerch: Same again! Sorry

bodhi/server/static/js/update_form.js
ryanlerch: and same again. sorry again!

cverna (approved):

PR #3764, opened 2019-11-15
Linkify update aliases in comments

mattiaverga:
This PR converts a simple update alias into a link to the update itself and beautify a link to an update by shortening it to the update alias.

example 1: `FEDORA-2019-1a2b3c4d5e` is converted into `<a href="{base_address}updates/FEDORA-2019-1a2b3c4d5e">FEDORA-2019-1a2b3c4d5e</a>`

example 2: `{base_address}updates/FEDORA-2019-1a2b3c4d5e` is converted into `<a href="{base_address}updates/FEDORA-2019-1a2b3c4d5e">FEDORA-2019-1a2b3c4d5e</a>`

example 3: `https://fake.bodhi.site/updates/FEDORA-2019-1a2b3c4d5e` is NOT converted (and it follows standard markup rules)

Fixes #776

Signed-off-by: Mattia Verga <mattia.verga@tiscali.it>

PR #3775, opened 2019-11-18
make bodhi comments render as markdown (again)

### LINKED ISSUE #3723: Unrendered Markdown in comments section
In Bodhi 5.0.0, there seems to be some unrendered Markdown in the info messages in the comments section of an update. See the following screenshot (see also [live](https://bodhi.fedoraproject.org/updates/FEDORA-2019-fdf8e84f60)):

Screenshot_2019-11-14 FEDORA-2019-fdf8e84f60 — bugfix update for libreport — Fedora Updates System

ryanlerch:
previously, in the old UI, the bodhi user comments rendered
as markdown in the updates view. This was mainly used to provide
clickable links in the bodhi comments other updates. However with the
introduction of the new UI, this regressed and the comments were just
shown as text, not rendered markdown.

This commit changes the layout of bodhi comments so we can present the
rendered markdown properly.

Fixes: #3723

### DIFF
--- a/bodhi/server/templates/fragments.html
+++ b/bodhi/server/templates/fragments.html
@@ -79,14 +79,12 @@ ${statusmap[status]}\
 % if comment.user.name == "bodhi":
 <div class="row">
         <div class="col-12 text-muted font-size-09">
-          <div class="card border-0">
-            <div class="card-body border-0 p-2">
-                <i class="fa fa-info-circle"></i>
-                <span class="font-weight-bold">${comment.text | n}</span>
-                <a href="${comment.url()}" class="notblue float-right" data-toggle="tooltip" title="${comment.timestamp.strftime("%Y-%m-%d %H:%M:%S")} (UTC)">
-                    ${util.age(comment.timestamp)}
-                  </a>
-            </div>
+          <div class="d-flex p-2 align-items-top">
+            <i class="fa fa-info-circle pt-2"></i>
+            <div class="font-weight-bold">${util.markup(comment.text, bodhi=False) | n}</div>
+            <a href="${comment.url()}" class="notblue pt-1 ml-auto text-nowrap" data-toggle="tooltip" title="${comment.timestamp.strftime("%Y-%m-%d %H:%M:%S")} (UTC)">
+              ${util.age(comment.timestamp)}
+            </a>
           </div>
         </div>
 </div>


bodhi/server/templates/fragments.html
```diff
@@ -79,14 +79,12 @@
 % if comment.user.name == "bodhi":
```
AdamSaleh, line 79: Question, what is the reason to have a different rendering for comments by 'bodhi'? Reading the code, it seems this would work fine for ordinary users.

bodhi/server/templates/fragments.html
```diff
@@ -79,14 +79,12 @@
...
-            <div class="card-body border-0 p-2">
-                <i class="fa fa-info-circle"></i>
-                <span class="font-weight-bold">${comment.text | n}</span>
-                <a href="${comment.url()}" class="notblue float-right" data-toggle="tooltip" title="${comment.timestamp.strftime("%Y-%m-%d %H:%M:%S")} (UTC)">
-                    ${util.age(comment.timestamp)}
-                  </a>
-            </div>
+          <div class="d-flex p-2 align-items-top">
+            <i class="fa fa-info-circle pt-2"></i>
+            <div class="font-weight-bold">${util.markup(comment.text) | n}</div>
```
mattiaverga: In #3764 I've added a new option to util.markup() to not use the Bodhi markup extensions. I think we should use that option here to avoid update links already posted in markup form to be modified by the new extension.

bodhi/server/templates/fragments.html
ryanlerch: the "bodhi" user is the one that adds the status update comments for an update. So all comments by this user are rendered differently in the comment stream (and also not counted in the comment counts on the update list page).

Screenshot from 2019-11-19 15-15-26

bodhi/server/templates/fragments.html
ryanlerch: roger that! we will have to merge this one after yours goes in though. So will add WIP.

abompard: PR #3764 is merged, you can adjust this now if need be.

ryanlerch:
this one is ready for re-review now too.

abompard (approved):
LGTM

PR #3793, opened 2019-11-22
Enable handling bugs by default for the Security Response product

### LINKED ISSUE #3789: Bodhi complains that it doesn't handle certain types of bugs (regression)
When filing security updates for Fedora, there are 2 bugs which must be included. The Fedora tracker for that security issue, and the security team tracker. With the bodhi update, it complains that it doesn't handle that type of bug for the security team tracker. Examples:

Fedora bug that bodhi handles:
https://bugzilla.redhat.com/show_bug.cgi?id=1775116

Tracker that it does not (and should):
https://bugzilla.redhat.com/show_bug.cgi?id=1775074

mattiaverga:
Bugs for the `Security Response` Bugzilla product have never been handled by Bodhi, but they should.

Fixes #3789 

### DIFF
--- a/bodhi/server/config.py
+++ b/bodhi/server/config.py
@@ -276,7 +276,7 @@ class BodhiConfig(dict):
             'value': 'dev',
             'validator': str},
         'bz_products': {
-            'value': ['Fedora', 'Fedora EPEL', 'Fedora Modules'],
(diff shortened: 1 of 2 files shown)


mattiaverga:
As commented in the Issue, we mustn't re-enable Bodhi actions on Security Response bugs, I'll retire this PR.

Issue #3790, opened 2019-11-22
Bring back space delimited bug number entry.


jmflinuxtx:
With the latest Bodhi upate, I can no longer drop in a list of space delimited bug numbers, they must be entered one at a time and select the found results.  This is a huge annoyance when dealing with updates that fix a large number of bugs (today's kernel updates address 40 different security bugs).  Please bring back the ability to drop a list of bug numbers in the bugs field.

pypingou:
@ryanlerch what do you think of this one?

mattiaverga:
Just as a note, adding a comma separated list of bugs seems to work.

jmflinuxtx:
1774933,1774934,1774937,1774939,1774946,1774947,1774949,1774950,1774957,1774958,1774963,1774965,1774972,1774973,1774976,1774978,1774988,1774989,1774998,1775002,1775021,1775023,1775010,1775011,1775015,1775016,1775042,1775043,1775047,1775048,1775050,1775051,1775956,1775110,1775097,1775115,1775074,1775116,1775063,1775117
Seems to just hang there forever. (Yes, this is a real bug list on a current update)

ryanlerch:
yeah, let's pull the code that pulls down the bug titles for now, it seems to make this hang

ryanlerch:
not sure what is causing the hang here... other than the annoying messages that should soon be resolved by #3794 pasting the comma separated list seems to work for me.

I am adding it to a new update, @jmflinuxtx are you editing an exisitng update or adding to a new update?

PR #3796, opened 2019-11-23
Fix JS error when removing bug from the list

mattiaverga:
In the  New Update form, when a user create a bug by entering the bug # and then removes it by the remove button, JS crashes because it misses `component`, `product` and `version` properties:
`TypeError: item.component is undefined`

This PR will make the callback memorize those properties. Another solution would be to just make selectize forget about the removed bug by setting `persist: false` in its options... you have the choice.

Could be somehow related to #3790 

### DIFF
--- a/bodhi/server/static/js/update_form.js
+++ b/bodhi/server/static/js/update_form.js
@@ -82,7 +82,11 @@ $(document).ready(function() {
                                 type: 'error',
                             });
                         }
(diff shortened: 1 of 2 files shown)


ryanlerch (approved):

abompard:
CI errors are unrelated, merging.

PR #3798, opened 2019-11-25
add celery tasks, now call these from models

StephenCoady:
fyi @nphilipp 

Signed-off-by: Stephen Coady <scoady@redhat.com>

### DIFF
--- a/bodhi/server/models.py
+++ b/bodhi/server/models.py
@@ -46,7 +46,7 @@ from bodhi.messages.schemas import (buildroot_override as override_schemas,
 from bodhi.server import bugs, buildsys, log, mail, notifications, Session, util
 from bodhi.server.config import config
 from bodhi.server.exceptions import BodhiException, LockedUpdateException
-from bodhi.server.tasks import handle_update
+from bodhi.server.tasks import handle_update, tag_update_builds_task
 from bodhi.server.util import (
     avatar as get_avatar, build_evr, get_critpath_components,
     get_rpm_header, header, tokenize, pagure_api_get)
@@ -2314,7 +2314,6 @@ class Update(Base):
         """
         db = request.db
         buildinfo = request.buildinfo
-        koji = request.koji
         up = db.query(Update).filter_by(alias=data['edited']).first()
         del(data['edited'])
 
@@ -2411,18 +2410,7 @@ class Update(Base):
                 })
 
             # Add the pending_signing_tag to all new builds
-            for build in new_builds:
-                if up.from_tag:
-                    # this is a sidetag based update. use the sidetag pending signing tag
-                    side_tag_pending_signing = up.release.get_pending_signing_side_tag(up.from_tag)
-                    koji.tagBuild(side_tag_pending_signing, build)
-                elif up.release.pending_signing_tag:
-                    # Add the release's pending_signing_tag to all new builds
-                    koji.tagBuild(up.release.pending_signing_tag, build)
-                else:
-                    # EL6 doesn't have these, and that's okay...
-                    # We still warn in case the config gets messed up.
-                    log.warning('%s has no pending_signing_tag' % up.release.name)
+            tag_update_builds_task.delay(update=up, builds=new_builds)
 
         # And, updates with new or removed builds always get their karma reset.
         # https://github.com/fedora-infra/bodhi/issues/511
@@ -4692,7 +4680,6 @@ class BuildrootOverride(Base):
             The new updated override.
         """
         db = request.db
-
         edited = data.pop('edited')
         override = cls.get(edited.id)
 
--- a/bodhi/server/services/updates.py
+++ b/bodhi/server/services/updates.py
@@ -26,7 +26,7 @@ from sqlalchemy import func, distinct
 from sqlalchemy.sql import or_
 from requests import RequestException, Timeout as RequestsTimeout
 
-from bodhi.server import log, security, buildsys
+from bodhi.server import log, security
 from bodhi.server.exceptions import BodhiException, LockedUpdateException
 from bodhi.server.models import (
     Update,
@@ -39,6 +39,7 @@ from bodhi.server.models import (
     Package,
     Release,
 )
+from bodhi.server.tasks import handle_side_and_related_tags_task
 import bodhi.server.schemas
 import bodhi.server.services.errors
 import bodhi.server.util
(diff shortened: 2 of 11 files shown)


bodhi/server/models.py
```diff
@@ -2390,18 +2389,7 @@ def edit(cls, request, data):
...
-                    side_tag_pending_signing = up.release.get_pending_signing_side_tag(up.from_tag)
-                    koji.tagBuild(side_tag_pending_signing, build)
-                elif up.release.pending_signing_tag:
-                    # Add the release's pending_signing_tag to all new builds
-                    koji.tagBuild(up.release.pending_signing_tag, build)
-                else:
-                    # EL6 doesn't have these, and that's okay...
-                    # We still warn in case the config gets messed up.
-                    log.warning('%s has no pending_signing_tag' % up.release.name)
+            tag_build(new_builds, up)
```
abompard: Celery tasks are called using the `.delay()` method, please check out how it's done with the `handle_update()` task.

bodhi/server/services/updates.py
```diff
@@ -567,22 +568,7 @@ def new_update(request):
...
-                    if not koji.getTag(side_tag_signing_pending):
-                        koji.createTag(side_tag_signing_pending, parent=from_tag)
-                    if not koji.getTag(side_tag_testing_pending):
-                        koji.createTag(side_tag_testing_pending, parent=from_tag)
-                        koji.editTag2(side_tag_testing_pending, perm="autosign")
-
-                    to_tag = side_tag_signing_pending
-                    # Move every new build to <from_tag>-signing-pending tag
-                    update.add_tag(to_tag)
+                create_tag(updates, from_tag)
```
abompard: Same comment here, you'll have to use the `.delay()` method.

bodhi/server/tasks/__init__.py
```diff
@@ -83,3 +83,50 @@ def handle_update(api_version: int, **kwargs):
     _do_init()
     handler = UpdatesHandler()
     handler.run(api_version=api_version, data=kwargs)
+
+
+@app.task(name="create_tag")
+def create_tag(updates: list, from_tag):
```
abompard: Is `from_tag` supposed to be a string? If so, please type it accordingly.

bodhi/server/tasks/__init__.py
```diff
@@ -83,3 +83,50 @@ def handle_update(api_version: int, **kwargs):
...
+            koji.createTag(side_tag_testing_pending, parent=from_tag)
+            koji.editTag2(side_tag_testing_pending, perm="autosign")
+
+        to_tag = side_tag_signing_pending
+        # Move every new build to <from_tag>-signing-pending tag
+        update.add_tag(to_tag)
+
+
+@app.task(name="tag_build")
+def tag_build(builds: list, update):
```
abompard: Please type the `update` argument.

abompard (dismissed):
Once the functions are moved to tasks, I'll expect some unit tests to fail (because the corresponding work would be done accordingly). I think you'll also have to reorder those parts of the unit tests, possibly create tests for the new tasks in the proper test module.

One more thing, I think it's worth adding an entry in the release notes saying that these actions will now be done asynchronously, please add a news fragment.

mattiaverga:
@abompard should these two new tasks be worked by the `has_koji_mount` celery queue?
https://github.com/fedora-infra/bodhi/blob/983f7c74e832a37cc1c842937208d5536bad9864/celeryconfig.py#L19-L24

abompard:
> @abompard should these two new tasks be worked by the `has_koji_mount` celery queue?

Unless I'm mistaken this code is only using the Koji API, so it can be executed from anywhere, we don't need the actual mount.

nphilipp:
The type hints introduced a cyclic import (models -> tasks -> models) for which I have a fix locally. I noticed, however, that (expectedly) some existing tests block on when calling the new tasks. I'll mock these calls out and instead add unit tests testing `create_tag`, `task_build` in isolation modeled after existing tests for Celery tasks.

nphilipp:
This fixes the tests breaking because of the introduction of using Celery tasks for Koji tagging. Needs unit tests for the new Celery tasks, though, so still WIP.

bodhi/tests/server/tasks/test_handle_side_and_related_tags.py
```diff
@@ -0,0 +1,49 @@
+from unittest.mock import patch
+
+from bodhi.server import buildsys, models
+from bodhi.server.tasks import handle_side_and_related_tags_task
+from bodhi.server.tasks.handle_side_and_related_tags import main as hdle_side_and_related_tags_main
```
nphilipp: Whoa that's a long line. Perhaps shorten the name to `handle_srtags_main` here and below (as you're shortening `handle` anyway to make it fit)?

bodhi/tests/server/tasks/test_handle_side_and_related_tags.py
```diff
@@ -0,0 +1,49 @@
...
+
+class TestTask(BasePyTestCase):
+    """Test the task in bodhi.server.tasks."""
+
+    @patch("bodhi.server.tasks.buildsys")
+    @patch("bodhi.server.tasks.initialize_db")
+    @patch("bodhi.server.tasks.config")
+    @patch("bodhi.server.tasks.handle_side_and_related_tags.main")
+    def test_task(self, main_function, config_mock, init_db_mock, buildsys):
+
```
nphilipp: Hmm, that blank line looks weird. Did you have some code previously and remove it later?

bodhi/tests/server/tasks/test_tag_update_builds.py
```diff
@@ -0,0 +1,58 @@
...
+
+class TestTask(BasePyTestCase):
+    """Test the task in bodhi.server.tasks."""
+
+    @patch("bodhi.server.tasks.buildsys")
+    @patch("bodhi.server.tasks.initialize_db")
+    @patch("bodhi.server.tasks.config")
+    @patch("bodhi.server.tasks.tag_update_builds.main")
+    def test_task(self, main_function, config_mock, init_db_mock, buildsys):
+
```
nphilipp: Same blank line here. I don't think there are explicit rules but it looks weird and is inconsistent with the other methods.

bodhi/server/tasks/__init__.py
```diff
@@ -28,7 +28,7 @@
 from bodhi.server.util import pyfile_to_module
 
 if typing.TYPE_CHECKING:
-    from bodhi.server.models import Update
+    from bodhi.server.models import Update  # pragma: no cover
```
nphilipp: This could go to `if typing.TYPE_CHECKING:` above, so any other "typing-only" imports are covered.

nphilipp (approved):
Looks good to me!

PR #3810, opened 2019-11-28
Migrate updates service tests to pytest

### LINKED ISSUE #3740: Convert bodhi/tests/server/services/test_updates.py to PyTest
[`bodhi/tests/server/services/test_updates.py`](https://github.com/fedora-infra/bodhi/blob/develop/bodhi/tests/server/services/test_updates.py) still contains classes that derive from `unittest.TestCase`. We use PyTest, so they could be migrated to not derive from `unittest.TestCase`, or derive from `bodhi.tests.server.base.BasePyTestCase` in case they need to use the test database or WSGI app.

In summary:

nphilipp:
In the course, wrap `unittest.TestCase.assertMultiLineEqual()` as a function in `bodhi.tests`.

fixes: #3740

### DIFF
--- a/bodhi/tests/server/services/test_updates.py
+++ b/bodhi/tests/server/services/test_updates.py
@@ -21,6 +21,7 @@ from html import escape
 from unittest import mock
 from urllib import parse as urlparse
 import copy
+import re
 import textwrap
 import time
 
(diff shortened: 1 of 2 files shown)


nphilipp:
Putting this on the CI Gating board because it blocks #3819.

bodhi/tests/server/services/test_updates.py
```diff
@@ -3002,11 +2984,10 @@ def test_edit_update(self, *args):
 
         Karma has been reset.
         """).strip()
-        self.assertMultiLineEqual(up['comments'][-1]['text'], comment)
-        self.assertEqual(len(up['builds']), 1)
-        self.assertEqual(up['builds'][0]['nvr'], 'bodhi-2.0.0-3.fc17')
-        self.assertEqual(self.db.query(RpmBuild).filter_by(nvr='bodhi-2.0.0-2.fc17').first(),
-                         None)
+        assert_multiline_equal(up['comments'][-1]['text'], comment)
```
pypingou, line 2987: what is the difference between this method from unittest and doing a basic ``assert foo == bar``?

nphilipp: It's used for string with a couple of lines. In case the assert fails, it doesn't dump the whole strings at you but shows a diff.

pypingou (approved):
Looks fine to me

Issue #3814, opened 2019-11-28
Cannot de-select the "auto-request stable" options or change thresholds.


opoplawski:
https://bodhi.fedoraproject.org/updates/FEDORA-EPEL-2019-dac149ad76/edit
I want to de-select the auto-request stable options, but the save button is not responsive.  Nor can I change the karma threshold or number of days.

mattiaverga:
I think that's because the javascript form crashes due to the presence of `"` in some bug titles. After the provided patch is applied it should work as expected.
Same problem as in #3714

PR #3817, opened 2019-11-28
[5.1-Backports] Correctly handle bug titles with double quote in new update form

### LINKED ISSUE #3714: Cannot edit builds/bugs in an update
After update to bodhi-5.0.0, I cannot edit builds nor bugs for an update.

- I haven't found any way of editing it in GUI (when clicking "edit", there is a list of builds and a list of bugs, but it cannot be edited or at least it's not clear how).
- In CLI, I get "Password:" prompt even though I have valid kerberos ticket.

mattiaverga:
Double quotes in bug titles need to be escaped with `\\` to be compatible with JSON.parse().
Fix #3714 #3814 

### DIFF
--- a/bodhi/server/util.py
+++ b/bodhi/server/util.py
@@ -1303,3 +1303,14 @@ def pyfile_to_module(
         e.strerror = 'Unable to load file (%s)' % e.strerror
         raise
     return d
+
+
+def json_escape(text: str) -> str:
+    """Escape double quotes for JSON.parse compatibility.
+
+    Args:
+        text: The text to sanitize.
+    Returns:
+        Escaped text.
+    """
+    return text.replace('"', '\\"')
--- a/bodhi/server/templates/new_update.html
+++ b/bodhi/server/templates/new_update.html
@@ -1,4 +1,7 @@
 <%inherit file="master.html"/>
+<%
+  from bodhi.server.util import json_escape
+%>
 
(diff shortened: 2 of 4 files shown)


bodhi/server/util.py
```diff
@@ -1303,3 +1303,14 @@ def pyfile_to_module(
         e.strerror = 'Unable to load file (%s)' % e.strerror
         raise
     return d
+
+
+def json_escape(text: str) -> str:
+    """Escape double quotes for JSON.parse compatibility.
+
+    Args:
+        text (str): The text to sanitize.
```
sebwoj: Since you are using type annotations, let's drop 'str' here.

bodhi/server/util.py
```diff
@@ -1303,3 +1303,14 @@ def pyfile_to_module(
...
+    Returns:
+        str: Escaped text.
```
sebwoj: Same here.

sebwoj (changes requested):

bodhi/server/util.py
mattiaverga: corrected, thanks

bodhi/server/util.py
mattiaverga: corrected

ryanlerch (approved):

PR #3819, opened 2019-11-29
Handle side tag updates in stable releases.

### LINKED ISSUE #3800: Workflow for updates from side tags in stable releases
Story
=====

As a user of Bodhi,

I want updates from side tags to behave like normal updates with many packages,

so they aren't left dangling, but can be tested and moved to stable like other updates.

Background
==========

* People have created updates from side tags in stable releases and found them to be stuck, not moved to testing: #3783 
* There's a card on the Community Fire Team™ board for it: https://teams.fedoraproject.org/project/community-fire-team/kanban
* There are two options when to delete the side-tag in Koji:
  * When the update is submitted. This would let people fix things by building normally and adding the builds to the update and is likely the easier option which is why we'll go with it (we can always make it more complicated if and when people come screaming and yelling at us).
  * When the update is moved to stable. This allows chain-building until the update is moved to stable, but probably splits the changes across more places in the code and requires people to continue building into the side-tag and refresh builds from it.

Acceptance Criteria
===============

For an update from a side-tag in a stable release:

* [ ] Bodhi should not create auxiliary tags to the side-tag.

* [ ] To be treated as a bog-standard update from multiple builds from here on out, builds in the update should be re-tagged with the `pending_signing_tag` of the release to be picked up for signing by robosignatory.

* [ ] The corresponding side tag is deleted when the update is submitted.

* [ ] Unit tests covering the implementation of the new workflow exist.

nphilipp:
In releases after the Bodhi activation point, updates will be funneled into the normal workflow and the side tag deleted.

@ryanlerch: For these updates, the side-tag info & UI will not be shown and because the PR is also based on #3810 (another PyTest migration), which is why I want to raise both to your attention.

fixes: #3800

### DIFF
--- a/bodhi/server/services/updates.py
+++ b/bodhi/server/services/updates.py
@@ -467,8 +467,12 @@ def new_update(request):
     empty, the list of builds will be filled with the latest builds in this
     Koji tag. This is done by validate_from_tag() because the list of builds
     needs to be available in validate_acls().
-    Ensure that related tags ``from_tag``-pending-signing and ``from_tag``-testing
-    exists and if not create them in Koji.
+
+    If the release is composed by Bodhi (i.e. a branched or stable release
+    after the Bodhi activation point), ensure that related tags
+    ``from_tag``-pending-signing and ``from_tag``-testing exists and if not
+    create them in Koji. If the state of the release is not `pending`, add its
+    pending-signing tag and remove it if it's a side tag.
 
     Args:
         request (pyramid.request): The current request.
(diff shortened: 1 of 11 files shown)


bodhi/server/services/updates.py
```diff
@@ -467,8 +467,11 @@ def new_update(request):
     empty, the list of builds will be filled with the latest builds in this
     Koji tag. This is done by validate_from_tag() because the list of builds
     needs to be available in validate_acls().
-    Ensure that related tags ``from_tag``-pending-signing and ``from_tag``-testing
-    exists and if not create them in Koji.
+
+    If the state of the release is `pending`, ensure that related tags
```
pypingou: Composed by bodhi rather than the release state actually

pypingou (changes requested):

bodhi/server/services/updates.py
nphilipp: Good catch!

bodhi/server/templates/new_update.html
```diff
@@ -71,7 +71,7 @@ <h5 class="font-weight-bold my-1 ml-2">Builds</h5>
                           </div>
                       </div>
                       % endif
-                      % if update and update.from_tag:
+                      % if update and update.from_tag and not update.release.composed_by_bodhi:
```
cverna, line 74: I have a hard time to understand this change, do this mean that we will only show this button for rawhide ?

nphilipp: Yes, it does. Because the update (and its builds) will be moved into the normal workflow, i.e. re-tagged with the pending-signing/-testing tag, and the side-tag will be removed, having the button serves no purpose because the tag doesn't exist anymore at this point.

cverna: Ok I misunderstood the purpose of this PR. I am not sure I get the use case. How does someone creates an update from a side tag in a stable release then ?

pypingou: > How does someone creates an update from a side tag in a stable release then ?

The user creates the side-tag, builds in it, creates an update from that side-tag.
At this point, all the builds are merged back to the traditional koji tag (the default signing-pending) and the side-tag is removed.
If the user needs to fix something for that update, they will have to go the buildroot override way as they do today.

pypingou (dismissed):

bodhi/server/templates/new_update.html
cverna: > > How does someone creates an update from a side tag in a stable release then ?
> 
> The user creates the side-tag, builds in it, creates an update from that side-tag.
> At this point, all the builds are merged back to the traditional koji tag (the default signing-pending) and the side-tag is removed.
> If the user needs to fix something for that update, they will have to go the buildroot override way as they do today.

Ok I missed that this was for editing the update and not creating it. Looking at the name of the template I assumed that this was only used to create new updates :see_no_evil:

cverna:
Are there any changes needed at the cli level ?

nphilipp:
> Are there any changes needed at the cli level ?

Not functionally, but I should probably update one of the error messages:

    "ERROR: This update was not created from a tag. Please remove --from_tag and try again."

Adding something like "... or the tag doesn't exist anymore. Please ..." should do it.

nphilipp:
Hmm, we probably return `from_tag` set in the JSON even for updates for a release composed by Bodhi. I'll add some code to the client to cope with this.

bodhi/server/templates/new_update.html
nphilipp: There are some confusingly named templates, methods, etc. around new updates vs. editing. :slightly_frowning_face:

nphilipp:
@cverna the latest push lets the client cope with updates originally created from a side-tag (which at the point of editing doesn't exist anymore).

cverna:
Code looks good, let's squash the commit before merging

nphilipp:
#3810 needs to go in before that anyway :)

nphilipp:
@cverna I've squashed the topical commits, i.e. everything but the "fix typo" one and rebased.

cverna (approved):
LGTM

nphilipp:
```
15:14:36  14:13:29.339 container.py      INFO   running command ['ls', '/tmp/pungi-calls.log']
15:14:36  14:13:29.544 container.py      INFO   ls: cannot access '/tmp/pungi-calls.log': No such file or directory
15:14:36  14:13:29.553 container.py      ERROR  command failed
15:14:36  14:13:29.553 container.py      INFO   exec metadata: {'ID': '64543a497c1dd81b3c763b7611b30c111de5b60e016ca875f5a368eb36f13e7f', 'Running': False, 'ExitCode': 2, 'ProcessConfig': {'tty': False, 'entrypoint': 'ls', 'arguments': ['/tmp/pungi-calls.log'], 'privileged': False, 'user': 'bodhi'}, 'OpenStdin': False, 'OpenStderr': True, 'OpenStdout': True, 'CanRemove': False, 'ContainerID': '7f75c8319ec34269eb7e5072d74ee201b7526671beb72a4972df9061980a7036', 'DetachKeys': '', 'Pid': 8334}
```

*sighs*

PR #3820, opened 2019-12-02
In Cornice 4.0, schema arguments must be instances

abompard:
See https://github.com/Cornices/cornice/releases/tag/4.0.0, the check for being a `MappingSchema` instance only works on instances and not on classes.

### DIFF
--- a/bodhi/server/services/builds.py
+++ b/bodhi/server/services/builds.py
@@ -60,7 +60,7 @@ def get_build(request):
     return build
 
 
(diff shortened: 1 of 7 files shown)


pypingou (approved):
If this is backward compatible (which the tests will tell us), it's fine for me! :)

pypingou:
jenkies test

abompard:
This change is incompatible with cornice-sphinx, which we use to build the docs. And upstream Cornice said they'd issue a fix to allow usage of classes instead of instances, so I think we can drop this PR.

PR #3836, opened 2019-12-06
Correct typo in DevBuildsys.listTagged docblock

sebwoj:


mattiaverga (changes requested):
I don't think we need a news entry here, can you remove that file and rebase?

mattiaverga (approved):

PR #3847, opened 2019-12-10
Try to avoid timeout error when requesting latest_candidates with hide_existing=true

### LINKED ISSUE #3841: bodhi fails to contact koji in prod
````
curl "https://bodhi.fedoraproject.org/latest_candidates?hide_existing=true&prefix=shutter"
<html><body><h1>504 Gateway Time-out</h1>
The server didn't respond in time.
</body></html>
````

Since bodhi 5.1 this is no longer a blocker but it's also definitely not a nice UX.

Originally reported at: https://pagure.io/fedora-infrastructure/issue/8451

### PR DESCRIPTION
**Pull request by mattiaverga**
Refine the query used to build up the list of builds we want to exclude from the results: if I'm correct, the `candidate_tag` is removed from a build when the update is pushed to testing, so those builds are already filtered from the koji response.

Results from Vagrant test, from the actual query:
```
>>> associated_build_nvrs = set(row[0] for row in db.query(m.Build.nvr).filter(m.Build.update_id != None))
>>> len(associated_build_nvrs)
217947
```
and the refined query:
```
>>> associated_build_nvrs = set(row[0] for row in db.query(m.Build.nvr).join(m.Update).filter(m.Update.status == m.UpdateStatus.pending))
>>> len(associated_build_nvrs)
210
```
This should hopefully close #3841 

In a side commit, I've added a delay to the javascript used to query koji, so that we don't fire a query for every letter we type in the search field.

### DIFF
--- a/bodhi/server/static/js/update_form.js
+++ b/bodhi/server/static/js/update_form.js
@@ -188,6 +188,7 @@ $(document).ready(function() {
                 // make sure the placeholder disappears when focused
                 $('#builds-search-selectized').attr("placeholder", "");
             },
+            loadThrottle: 800,
             load: function(query, callback) {
                 $.ajax({
                     url: '/latest_candidates?hide_existing=true&prefix=' + encodeURIComponent(query),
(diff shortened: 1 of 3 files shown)


pypingou:
I'd like a second set of eyes on this one (preferably from @nphilipp or @ryanlerch who touched that part of the code recently) but the code and logic look fine to me :)

cverna (approved):
+1

PR #3853, opened 2019-12-12
Allow task id to be null in the bodhi.update.status.testing message s…

cverna:
…chema.

In the case of modular build there is no Koji task so
we need to allow the task_id to be null.

### DIFF
--- a/bodhi/messages/schemas/update.py
+++ b/bodhi/messages/schemas/update.py
@@ -662,7 +662,7 @@ class UpdateReadyForTestingV1(BodhiMessage):
                     },
                     'task_id': {
                         'description': 'Task ID of the koji build.',
-                        'type': 'integer',
+                        'type': ['null', 'integer'],
                     },
                     'component': {
                         'description': 'Name of the component tested.',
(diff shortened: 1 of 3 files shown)


pypingou:
I'm :+1: to the change but give me a little time before merging this, I'd like to see if I can get someone from the CI folks

cverna:
Closing this, since the CI pipeline will require a different message for modules. I ll submit another PR that send this message only for RPM updates.

pypingou:
I am re-opening this PR and letting it in.

The gist of it is that even RPM builds can be imported into koji (just like we do for module builds), releng has done it a few times in the past and even though it's an exceptional measure there is nothing telling us it will never happen again.
If we keep ``task_id`` required and import an RPM build in the future, bodhi will crash as it does today and things are stuck.
If we allow ``task_id`` to be ``None``, the CI pipeline will crash and the update will be gated. Considering this would be exceptional circumstances, this would be something we can waive through (not ideal but also very exceptional).

pypingou (approved):

PR #3857, opened 2019-12-12
Change a log level for validation errors

abompard:
Because ERROR-level logs get sent by email and this is really not necessary.


pypingou (approved):
:fireworks:

PR #3861, opened 2019-12-13
Use shared forms code also for comments.

### LINKED ISSUE #3837: Comment submission is slow on updates with many builds, in extreme cases request times out and user experience is confusing (spinner stops spinning, no page refresh, no error shown, comment not yet visible on manual refresh)
Lately when I add a comment to an update, it often doesn't appear. I press the "Add Comment & Feedback" button, it shows a spinner for a while, then it goes back to 'ready to press' state - but the rest of the page is not refreshed, I do not see my comment or karma appear. Sometimes if I do a hard refresh (shift-ctrl-r) at this point the changes will appear, but sometimes even this doesn't make them show up. So I submit the comment again (and the same thing happens)...then I hard refresh again a few seconds or minutes later, and I see both comments.

I've seen quite a few duplicated comments on updates from other people, too, from which I infer that I'm not the only one getting bitten by this.

mattiaverga:
This will make comments submission to use existing shared code with other form components (new-update, override).

It could help to fix #3837 

### DIFF
--- a/bodhi/server/static/js/comment_form.js
+++ b/bodhi/server/static/js/comment_form.js
@@ -2,7 +2,7 @@
 
 $(document).ready(function() {
     var base = document.baseURI;
-    CommentsForm = function() {};
+    var CommentsForm = function() {};
     CommentsForm.prototype = new Form("#new_comment", base + "comments/");
     CommentsForm.prototype.success = function(data) {
         Form.prototype.success.call(this, data);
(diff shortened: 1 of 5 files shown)


AdamWill:
This may address the UI issues (my frontend knowledge isn't good enough to know at a glance), but it'd also be good to make it so submitting a comment to a bug with a lot of builds *doesn't* take a long time, I think. There's no intrinsic reason why it should, I don't think.

cverna (approved):
LGTM, thanks

cverna:
@Mergifyio rebase

cverna:
@Mergifyio rebase

cverna:
jenkies test

PR #3864, opened 2019-12-15
Upgrade Vagrant box to f30

mattiaverga:
I'd rather wanted to upgrade it to f31, but it seems docker [doesn't run on that](https://github.com/docker/for-linux/issues/792) by default, so I've upgrade to f30.

### DIFF
--- a/devel/ansible/roles/bodhi/tasks/rabbitmq.yml
+++ b/devel/ansible/roles/bodhi/tasks/rabbitmq.yml
@@ -6,6 +6,12 @@
   with_items:
       - rabbitmq-server
 
+- name: Start rabbitmq
+  service:
+    name: rabbitmq-server
+    state: started
+    enabled: yes
+
 - name: Create RabbitMQ systemd override directory
   file:
     path: /etc/systemd/system/rabbitmq-server.service.d/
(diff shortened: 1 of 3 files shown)


devel/ansible/roles/bodhi/tasks/rabbitmq.yml
```diff
@@ -6,6 +6,12 @@
   with_items:
       - rabbitmq-server
 
+- name: Start rabbitmq
+  service:
+    name: rabbitmq-server
+    state: started
+    enabled: yes
```
abompard, line 13: This should probably be moved below the copy of the `override.conf` file below, because systemd will make use of it.

mattiaverga: While trying to upgrade Vagrant to F31 I discovered that trying to start the server after the limits override and the plugin activation makes it crash. I had to firstly start the server and then do the customization.
I did not try if that happens also in F30: I could check, but both the limits override and the plugin activation triggers a service restart with a `systemd daemon-reload`.

For the same reason I changed the user creation step: with the `permissions` command the step failed with a strange `unknown host: Vhost`.

mattiaverga: I've just tried: also in F30, doing the customization before starting the server makes rabbitmq-server.service unable to start. Instead starting the service first and then doing the customization and reload/restart the service works.

cverna (approved):

cverna:
@mattiaverga could rebase your branch and force push to trigger the CI. Not sure why the last run failed.

PR #3866, opened 2019-12-17
responses 0.10.8 breaks the unit tests.

cverna:
The 0.10.8 version of responses adds the text/plain content-type to
the response headers. This is breaking skopeo_lite which does not
expect this content type in the response.

Upstream issue https://github.com/getsentry/responses/issues/290


abompard (approved):
Let's keep an eye on the upstream bug to revert the version limit when it's fixed.

cverna:
This was already fixed upstream \o/

PR #3870, opened 2019-12-23
Fix celery task names in celeryconfig.py

abompard:
The task names in the example config were wrong.


pypingou (approved):

abompard:
The integration test error seemed to be transient, trying again.

PR #3872, opened 2019-12-23
Fix integration tests

abompard:
This PR is an attempt to fix the integration tests, which currently fail for no bodhi-related reason.


abompard:
Well apparently the integration tests stopped failing... There's a situation in the prod DB that we probably haven't handled, but it'll have to wait for next time.
The commit in this PR is still useful, though.

PR #3874, opened 2019-12-26
Sent UpdateReadyForTestingV1 only for rpm

### LINKED ISSUE #3855: UpdateReadyForTestingV1 needs to be sent only for rpm build updates
For other content (modules, flatpaks, containers) a new message will be needed.

We should check the `update.content_type` in `_ready_for_testing` method and only send the message if the contentType is rpm.

https://github.com/fedora-infra/bodhi/blob/develop/bodhi/server/models.py#L3900

sebwoj:
fix #3855


cverna:
@Mergifyio rebase

cverna (approved):
LGTM

PR #3889, opened 2020-01-07
Don't use deprecated unescape function in tests.

QuLogic:
Signed-off-by: Elliott Sales de Andrade <quantum.analyst@gmail.com>


cverna (approved):
Thanks

PR #3890, opened 2020-01-07
Always show search box.

### LINKED ISSUE #1455: new theme requires locating tiny magnifying glass to show search box
the primary thing I do on bodhi is search for a package, but this is made considerably more clumsy since I have to locate and click a tiny magnifying glass in order to get to the search box. is search a secondary feature for your typical users?? can't it always be displayed? if not, could you provide a cookie to remember whether the user has opened the search box?

QuLogic:
Fixes #1455.

Signed-off-by: Elliott Sales de Andrade <quantum.analyst@gmail.com>

### DIFF
--- a/bodhi/server/static/css/site.css
+++ b/bodhi/server/static/css/site.css
@@ -58,46 +58,6 @@ hr {
     display: inline;
 }
 
-/* Make the navbar collapse take effect at larger sizes than it normally would
- * so we don't crowd out the search widget...
- */
-@media (max-width: 991px) {
-    .navbar-header {
-        float: none;
-    }
-    .navbar-toggle {
-        display: block;
-    }
-    .navbar-collapse {
-        border-top: 1px solid transparent;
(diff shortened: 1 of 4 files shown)


cverna:
@ryanlerch could you take a look at this PR ? thanks

mattiaverga:
Just my personal opinion here: the searchbar seems a little too prominent.
What about moving it in a row below the first masthead, reducing its width and height, like in the following mockup? The free space left in the bar could be used in future for a newsticker (for example for announcing Bodhi's outages for maintenance or releases in freeze state).
immagine

QuLogic:
I disagree on its prominence; as noted in #1455, this is a primary method of interaction so it should be prominent.

Also, if I shrank the search bar to the size in your mockup, the results popup is also that width, which is too small to be useful. And unfortunately I do not know the right CSS incantations to make it work right.

cverna:
Closing in favour of #3963

PR #3892, opened 2020-01-07
Add quick release filters for listing updates

mattiaverga:
This PR adds three quick filters to the update list search form. The `all current`, `all pending` and `all archived` are just wrappers to select all releases in a specific state. They can be combined by themselves or with specific release names as usual.

The third commit in this PR also changes the default link to the Updates page to only list updates from `current` and `pending` releases to lower the query results. On my local test machine the update page generation speed has decreased from 1.9s to 1.2s.

Signed-off-by: Mattia Verga mattia.verga@tiscali.it

### DIFF
--- a/bodhi/server/templates/updates.html
+++ b/bodhi/server/templates/updates.html
@@ -71,6 +71,9 @@ ${parent.css()}
                                 "user",
                                 "updateid",
                                 "alias"]
+              macroCatReleases = [('__current__', 'all current'),
+                                  ('__pending__', 'all pending'),
+                                  ('__archived__', 'all archived')]
               searchterm=''
               parameters = request.params.dict_of_lists()
               if 'search' in parameters:
@@ -136,6 +139,14 @@ ${parent.css()}
                   <div class="col-8 pl-1">
                       <select id="updatereleases" name="releases" multiple="multiple">
                           <option value="">&nbsp;</option>
+                          <optgroup label="quick filters">
+                          % for macrocat in macroCatReleases:
+                            <option value="${macrocat[0]}"
+                            % if 'releases' in parameters and macrocat[0] in parameters['releases']:
+                            selected="selected"
+                            % endif
+                            >${macrocat[1]}</option>
+                          %endfor
                           % for rstatus in ['current', 'pending', "archived"]:
                             <optgroup label="${rstatus}">
                             % for value in releases[rstatus]:
--- a/bodhi/tests/server/services/test_updates.py
+++ b/bodhi/tests/server/services/test_updates.py
@@ -1204,6 +1204,15 @@ class TestEditUpdateForm(BasePyTestCase):
 
 @mock.patch('bodhi.server.models.handle_update', mock.Mock())
 class TestUpdatesService(BasePyTestCase):
+    @pytest.fixture
+    def create_quick_filters_data(self):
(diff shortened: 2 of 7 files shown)


cverna:
What do you think about renaming the filter to `ga, pre-release, eol` ? or something along these lines.
The current, pending and archived are kind of internal status for the releases, if the above are clearer from an end user perspective.

bodhi/tests/server/services/test_updates.py
```diff
@@ -1204,6 +1204,16 @@ def test_xss(self):
...
 class TestUpdatesService(BasePyTestCase):
+    @pytest.fixture
+    def create_quick_filters_data(self):
+        self.create_release('20')
+        self.create_release('40')
+        Release.query.filter_by(version='20').first().state = ReleaseState.archived
+        Release.query.filter_by(version='40').first().state = ReleaseState.pending
+        self.create_update([('really_old-1.2.3-1.fc20')], 'F20')
+        self.create_update([('shiny_new-7.8.9-1.fc40')], 'F40')
+        self.db.flush()
```
cverna: I don't think that you need the flush here if you do a commit just after.

bodhi/server/templates/updates.html
```diff
@@ -71,6 +71,9 @@ <h3 class="font-weight-bold m-0 d-flex align-items-center">Updates
                                 "user",
                                 "updateid",
                                 "alias"]
+              macroCatReleases = [('all-current', 'all current'),
```
cverna: Not sure about the `all-` prefix, if think if you choose in a filter `pending` it is obvious that you are going to get all the pending releases.

What do you think ?

mattiaverga: You're right, I'll change that

mattiaverga:
> What do you think about renaming the filter to `ga, pre-release, eol` ? or something along these lines.
> The current, pending and archived are kind of internal status for the releases, if the above are clearer from an end user perspective.

Since the release page already name the categories as `Current Releases, Pending Releases, Archived Releases` I think it's better to maintain the same names. Or change both, if the `ga, pre-release, eol` names are better.

bodhi/server/templates/updates.html
mattiaverga: The more I think about that, the less sure I am about. The filter `?releases=all-current,F15` is more readable than `?releases=current,F15`, where `current` seems a name of a release.

I'm thinking about adopting RH Bugzilla style and change the macro filters to `__current__`, `__pending__` and `__archived__`, which should also automatically prevent that one of those names conflicts with a true release name (for example, if Bodhi is used outside Fedora and a release is named 'current').

What do you think?

cverna: Sounds good to me :+1:

cverna:
> > What do you think about renaming the filter to `ga, pre-release, eol` ? or something along these lines.
> > The current, pending and archived are kind of internal status for the releases, if the above are clearer from an end user perspective.
> 
> Since the release page already name the categories as `Current Releases, Pending Releases, Archived Releases` I think it's better to maintain the same names. Or change both, if the `ga, pre-release, eol` names are better.

Ok lets keep it as it is then (Current, Pending, Archived) :)

cverna (approved):
Thanks

PR #3894, opened 2020-01-08
Take precautions to avoid DB transaction leaks

abompard:
Use transactional_session_maker in the celery tasks to ensure that DB sessions are removed when the task is done.
Only pass scalar as celery task arguments to prevent celery from having to emit SQL

### DIFF
--- a/bodhi/server/models.py
+++ b/bodhi/server/models.py
@@ -2431,6 +2431,9 @@ class Update(Base):
 
         up.date_modified = datetime.utcnow()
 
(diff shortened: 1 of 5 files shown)


cverna:
Looks good to me, all the celery tasks are not in 5.1 yet so we won't be able to backport this PR cleanly. It might be easier to make a PR per commit.

abompard:
Good point, I'll split the PR in 2

abompard:
See #3901 and #3902

PR #3896, opened 2020-01-08
Fix broken JS in update page

mattiaverga:
#3722 broke javascript on the update page. A `});` was missing.


cverna (approved):
Thanks for catching this

PR #3901, opened 2020-01-10
Use transactional_session_maker in the celery tasks

abompard:
This ensures that DB sessions are removed when the task is done.


abompard:
I'm unable to reproduce locally the failures that CI sees. I'll keep working on this, sorry for the delay.

cverna (approved):

PR #3902, opened 2020-01-10
Only pass scalar arguments to celery tasks

abompard:
This avoids the celery enqueuer emitting SQL queries to resolve
attributes, and therefore opening new transactions.


cverna (approved):
:+1:

PR #3903, opened 2020-01-11
Add search filters in the Overrides page

mattiaverga:
Currently, there's no easy way to reach the overrides page and apply filters in it. The only link available is in the homepage when a user is logged in and links to the user's overrides. To apply filters you need to manually write them in the URL.

This PR adds search filters to the overrides page (just like the updates page) and a link in the navbar which brings to a list of all active overrides.

This PR depends on #3892 


cverna (approved):
LGTM

Issue #3854, opened 2019-12-12
Create proper live and readiness endpoints for bodhi to be used in Open Shift configuration.


ryanlerch:
Currently we use :8080/ for both, we should have proper endpoints that represent the container is alive and should not be killed (usually /healthz/live) and that the container is ready to serve requests (usually /healthz/ready)

PR #3910, opened 2020-01-15
Health endpoints

AdamSaleh:
Should implement part of the leftover work for end-of-the year Bodhi CI initiative, notably  #3854

Liveness endpoint works always.
Readyness endpoint at least tries to query a database.

### DIFF
--- a/bodhi/server/views/generic.py
+++ b/bodhi/server/views/generic.py
@@ -470,6 +470,36 @@ def api_version(request):
     return dict(version=bodhi.server.util.version())
 
 
+@view_config(route_name='liveness', renderer='json')
+def liveness(request):
+    """
+    Return 'ok' as a sign of being alive.
+
+    Args:
+        request (pyramid.request.Request): The current request.
+    Returns:
+        str: 'ok'
+    """
+    return 'ok'
+
+
+@view_config(route_name='readyness', renderer='json')
+def readyness(request):
+    """
+    Return 200 if the app can query the db, to signify being ready.
+
+    Args:
+        request (pyramid.request.Request): The current request.
+    Returns:
+        dict: A dictionary with list of services checked.
+    """
+    try:
+        request.db.execute("SELECT 1")
+        return dict(db_session=True)
+    except Exception:
+        raise Exception("App not ready, is unable to execute a trivial select.")
+
+
 @notfound_view_config(append_slash=True)
 def notfound_view(context, request):
     """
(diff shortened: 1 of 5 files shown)


cverna:
Thanks for that, this will need some unit tests in order to pass the CI.

bodhi/server/views/generic.py
```diff
@@ -465,6 +465,40 @@ def api_version(request):
     return dict(version=bodhi.server.util.version())
 
 
+@view_config(route_name='api_version', renderer='json')
```
mattiaverga: the route name is wrong

bodhi/server/views/generic.py
```diff
@@ -465,6 +465,40 @@ def api_version(request):
...
+def liveness(request):
+    """
+    Return the Bodhi API version.
```
mattiaverga: docstrings are wrong

bodhi/server/views/generic.py
```diff
@@ -465,6 +465,40 @@ def api_version(request):
...
+
+    Args:
+        request (pyramid.request.Request): The current request.
+    Returns:
+        dict: A dictionary with a "version" key indexing a string of the Bodhi version.
+    """
+    return "ok"
+
+
+@view_config(route_name='api_version', renderer='json')
```
mattiaverga: same here, route name and docstrings are wrong

bodhi/server/views/generic.py
```diff
@@ -465,6 +465,40 @@ def api_version(request):
...
+        request (pyramid.request.Request): The current request.
+    Returns:
+        dict: A dictionary with a "version" key indexing a string of the Bodhi version.
+    """
+    ready = True
+    try:
+        db.session.execute("SELECT 1")
+    except:
+        ready = False
+    if ready:
```
mattiaverga: I think this if/else statement is unnecessary, all the work can be done inside the try/except statement

bodhi/server/views/generic.py
```diff
@@ -465,6 +465,40 @@ def api_version(request):
...
+def liveness(request):
+    """
+    Return the Bodhi API version.
+
+    Args:
+        request (pyramid.request.Request): The current request.
+    Returns:
+        dict: A dictionary with a "version" key indexing a string of the Bodhi version.
+    """
+    return "ok"
```
mattiaverga: `return True` seems better to me - just my opinion

mattiaverga (dismissed):
There are several errors due to excessive copy/pasting. Also it misses tests.

cverna:
That also need a file in the `new` directory (https://github.com/fedora-infra/bodhi/blob/develop/docs/developer/index.rst#release-notes)

cverna (approved):

PR #3917, opened 2020-02-01
Prevent updates being stuck in Testing from sidetags

### LINKED ISSUE #3912: Can't push Rawhide update manually
When I tried pushing update with merged side tag, it took some while, where I did not know what is going on. During that time, because I was impatient, I have disabled "Auto-request stable based on karma?", "Auto-request stable based on time?" and "Require testcases". Then it eventually got into testing, but there was no option to push it into stable.

I was told by @cverna, that I should re-enable "Auto-request stable based on time?", which seems to enable the auto push, but this is not obvious. It should either not be possible to disable everything or there should be option to push this manually.

[1]: https://bodhi.fedoraproject.org/updates/FEDORA-2020-223644200d

### LINKED ISSUE #3914: Wrong update instructions for rawhide
If one manually creates an update for rawhide, it doesn't auto-migrate.  So we need to chase karma, and the "how to install" instructions don't work, which is a problem.

Concretely, I have [this update](https://bodhi.fedoraproject.org/updates/FEDORA-2020-d5d4c21e87).  It happens to be pulling in a side-tag.  The instructions at the top are to run 

```sh
sudo dnf upgrade --enablerepo=updates-testing --advisory=FEDORA-2020-d5d4c21e87
```

As far as I can tell, this will never work because the packages don't go there.  I'm actually not sure what the right invocation is, and would kind of like to know so that I can ask people to test.

mattiaverga:
Prevent updates being stuck in Testing from sidetags. Also don't show wrong update command.

Fix #3912 
Fix #3914 


cverna (approved):
LGTM, not sure why the CI is failing yet

mattiaverga:
I'll try to retrigger tests. Let's see if they're going to fail again.

PR #3924, opened 2020-02-08
Fix bodhi-client unit tests failing

mattiaverga:
For some reason the two client tests that requires a mocked user input for entering their username have started to failing.
I'm trying to find where the issue is. The pip-unit still uses a F29 image, so let's upgrade it to F31.
Another difference with other dockerfiles is that pip-unit installs pytest from pypi and gets 5.3.5, while in all other dockerfiles pytest comes from fedora repositories and is < 5. But I have upgraded pytest to 5.3.5 locally and I can't reproduce the failure, so I don't think that's the real problem.

### DIFF
--- a/bodhi/tests/client/test___init__.py
+++ b/bodhi/tests/client/test___init__.py
@@ -879,7 +879,7 @@ real_open = open
 def fake_open_no_session_cache(*args, **kwargs):
     """Fake open so that it looks like we have no session cache."""
     if args[0] == fedora.client.openidbaseclient.b_SESSION_FILE:
-        return mock.mock_open(read_data='{}')(*args, **kwargs)
+        return mock.mock_open(read_data=b'{}')(*args, **kwargs)
     return real_open(*args, **kwargs)
 
 
--- a/devel/ci/Dockerfile-pip
+++ b/devel/ci/Dockerfile-pip
@@ -1,8 +1,5 @@
-FROM registry.fedoraproject.org/fedora:29
+FROM registry.fedoraproject.org/fedora:31
 LABEL maintainer="Randy Barlow <bowlofeggs@fedoraproject.org>"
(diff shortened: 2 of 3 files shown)


mattiaverga:
So, it seems that tests which requires running a query with `--mine` are failing at `bindings.BodhiClient` initialization:
https://github.com/fedora-infra/bodhi/blob/1be3f91c9da1b7ba1f7556999091511a674ad317/bodhi/client/__init__.py#L645
But I can't say why. 
It calls
https://github.com/fedora-infra/bodhi/blob/1be3f91c9da1b7ba1f7556999091511a674ad317/bodhi/client/bindings.py#L206-L207
which throws this exception:
`(<class 'AttributeError'>, AttributeError("'str' object has no attribute 'decode'"), <traceback object at 0x7f96eac35640>)`

Another odd thing is that client tests are not performed cleanly. The `--debug` flag set by
https://github.com/fedora-infra/bodhi/blob/1be3f91c9da1b7ba1f7556999091511a674ad317/bodhi/tests/client/test___init__.py#L509
is inherited by other tests.

mattiaverga:
@cverna I see that a couple of weeks ago fedora-python 1.0.0 was released and these failures seems related to `fedora.client.OpenIdBaseClient` initialization.
Do you think the two things can be related? Only pip-units seems affected.

mattiaverga:
Well, the fix was easier than I thought... now the PR is ready to be reviewed.

But where do those `f29-*` expected tests come from? Is there some setting that need to be changed in centos-ci jenkins? I can't find anything that requires f29 tests in Bodhi configs.

cverna (approved):

pypingou:
All tests passed by f29 which are no longer needed.
Going to merge this one manually

PR #3930, opened 2020-02-16
Allow revoking the push to stable action.

### LINKED ISSUE #3921: There's no way to stop an update (being) pushed to stable and get it back to testing
1. Click *Push to stable*
 2. Realize this was not intended, you want to keep this in testing for some reason
 3. There is no *Unpush from stable, keep in testing* action available

As a workaround, one can unpush the update entirely and push it back to testing. That can be very confusing and possibly prevents additional timely push to stable.

mattiaverga:
Fix #3921 


cverna (approved):

hroncok:
Thanks!

PR #3931, opened 2020-02-18
Fix changelog entry in automatic rpm updates notes

mattiaverga:
Since Bodhi 5.0 automatically created RPM updates (for rawhide) should get useful changelog notes from rpm headers, but they don't.
This should fix notes creation and also simplify the code by moving the appropriate function from utils to the RpmBuild model.


bodhi/server/models.py
```diff
@@ -1635,9 +1639,9 @@ def get_latest(self):
         # generate ChangeLogs against those.
         latest = None
         evr = self.evr
-        for tag in [self.update.release.stable_tag, self.update.release.dist_tag]:
-            builds = koji_session.getLatestBuilds(
-                tag, package=self.package.name)
+        for tag in [self.release.stable_tag, self.release.dist_tag]:
+            builds = koji_session.listTagged(
```
cverna, line 1643: Why this changed was needed ?

bodhi/server/models.py
```diff
@@ -1671,6 +1676,17 @@ def get_changelog(self, timelimit=0):
         if not isinstance(when, list):
             when = [when]
 
+        if lastupdate:
+            lastpkg = self.get_latest()
+            if lastpkg is not None:
+                oldh = get_rpm_header(lastpkg)
+                if oldh['changelogtext']:
```
cverna, line 1683: Should we use get here ? or are we sure that the `changelogtext` key will always be there ?
```suggestion
                if oldh.get('changelogtext'):
```

bodhi/server/models.py
```diff
@@ -1649,13 +1653,14 @@ def get_latest(self):
                 break
         return latest
 
-    def get_changelog(self, timelimit=0):
+    def get_changelog(self, timelimit=0, lastupdate=False):
```
cverna, line 1656: Should we default this to `True` since the 2 calls we have are passing `True` ?

bodhi/tests/server/tasks/test_composer.py
```diff
@@ -540,7 +540,8 @@ def test_tag_ordering(self, *args):
 
         with self.db_factory() as session:
             firstupdate = session.query(Update).one()
-            build = RpmBuild(nvr=otherbuild, package=firstupdate.builds[0].package, signed=True)
+            build = RpmBuild(nvr=otherbuild, package=firstupdate.builds[0].package,
+                             release=firstupdate.release, signed=True)
```
cverna, line 544: why do we need to pass the release now ?

cverna (commented):
A few questions, but the overall looks good thanks for looking at this

bodhi/server/models.py
mattiaverga: This line is the main reason of why currently the changelog isn't produced: Koji's `getLatestBuilds()` returns a list of only the latest build of every package for the specified tag. That means that by filtering for a specific package we only get a list with a single entry, the latest build of that package for that tag (this build itself), so we will never get the "first build older than this one" in the next step.

...I hope I've been clear...

bodhi/server/models.py
mattiaverga: The `False` choice has been made to preserve the actual behavior of the function. Also setting `lastupdate` to `True` will overwrite the value from `timelimit`, so I think it's safer to let it `False` by default.

bodhi/server/models.py
mattiaverga: mmm I need to check if `changelogtext` is always present or not. Here I simply reused the old code that was in `bodhi.utils`.

bodhi/tests/server/tasks/test_composer.py
mattiaverga: Because of the change in `models.RpmBuild.get_latest()`:
```
for tag in [self.update.release.stable_tag, self.update.release.dist_tag]:
```
has been changed to 
```
for tag in [self.release.stable_tag, self.release.dist_tag]:
```
and a Build object is supposed to always have a Release associated. Previously the release was acquired from the update by linking Build -> Update -> Release, now it is acquired directly from Build -> Release and the mocked Build here in the tests was not happy with that.

bodhi/server/models.py
mattiaverga: I've just checked, `changelogtext` will always be there

cverna (approved):
Thanks :+1:

PR #3935, opened 2020-02-26
Fix the unit and integration tests.

cverna:
Lower the total coverate to 99% since there is little value in
chasing .xx%.
The postgresql container now requires to specify the AUTH METHOD to
trust in order to allow connections without a password.


mattiaverga:
It seems I have introduced a regression in 634a3d9980ddb8497cf17644043d863cc692898b
We must manually install `python3-pip` into `Dockerfile-pip`:
https://github.com/fedora-infra/bodhi/blob/51e8d2ed3c3410d5052042069cedac95391b4900/devel/ci/Dockerfile-pip#L4-L20

cverna:
> It seems I have introduced a regression in [634a3d9](https://github.com/fedora-infra/bodhi/commit/634a3d9980ddb8497cf17644043d863cc692898b)
> We must manually install `python3-pip` into `Dockerfile-pip`:
> 
> https://github.com/fedora-infra/bodhi/blob/51e8d2ed3c3410d5052042069cedac95391b4900/devel/ci/Dockerfile-pip#L4-L20

Yes that fixes the unit tests, the integration tests are hanging I am not sure why yet, still looking at it.

mattiaverga:
Apart from flake8 errors, it seems there is at least one test failing (`devel/ci/integration/tests/test_bodhi_cli.py::test_updates_query_details`), while `devel/ci/integration/tests/test_bodhi_tasks.py::test_push_composer_start` gets stuck.

`test_updates_query_details` fails due to:
```
>       assert "summary" in greenwave_result
E       AssertionError: assert 'summary' in {'message': 'Cannot find any applicable policies for koji_build subjects at gating point bodhi_update_push_stable in fedora-33'}
```



`test_push_composer_start` gets stuck at
https://github.com/fedora-infra/bodhi/blob/51e8d2ed3c3410d5052042069cedac95391b4900/bodhi/server/push.py#L201-L203
which never returns.

cverna:
@Mergifyio refresh

PR #3939, opened 2020-03-03
Use images from DockerHub instead of registry.fp.o.

cverna:


cverna:
Use DockerHub to have reliable image pull (https://pagure.io/fedora-infrastructure/issue/8710)

pypingou (approved):
Nice and easy, but a pity we can't use our own resources here :(

PR #3940, opened 2020-03-03
Make sure we are sending the messages before triggering a celery task.

### LINKED ISSUE #3904: bodhi occasionally fails to publish to fedora-messaging when it ought to (`update.request.testing`, `update.edit` at least)
I noticed that openQA tested a 389-ds-base update for Fedora 31, but not an update for the same package for Fedora 30 that was submitted a few minutes later. Looking into it, it seems we got a [bodhi.update.request.testing message for the F31 update](https://apps.fedoraproject.org/datagrepper/id?id=2020-87b4ee60-1e32-4b6b-ac6f-02f043aa9c8a&is_raw=true&size=extra-large), but none was published for [the F30 update](https://bodhi.fedoraproject.org/updates/FEDORA-2020-fb48098a19). I can't see any reason why a message should not have been published, so I'm filing this bug.

openQA won't test updates unless it sees a message for them, so this is restricting test coverage.

cverna:
We need to make sure to commit the database session before
triggering a celery task, but since the fedora-messaging messages
are sent on the database session commit let's commit this one after
the messages were added to the queue.

Fixes #3904

### DIFF
--- a/bodhi/server/models.py
+++ b/bodhi/server/models.py
@@ -2444,11 +2444,11 @@ class Update(Base):
         # Store the update alias so Celery doesn't have to emit SQL
         update_alias = up.alias
 
-        # Commit the changes in the db before calling a celery task.
-        db.commit()
-
         notifications.publish(update_schemas.UpdateEditV1.from_dict(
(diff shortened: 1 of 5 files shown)


pypingou (approved):
Looks good to me

PR #3942, opened 2020-03-05
Do not use the compose that are already done in the integration test

cverna:


Zlopez (approved):
LGTM

Issue #3922, opened 2020-02-06
Setup development environment


cverna:
# Summary
* Follow the [developer documentation](https://bodhi.fedoraproject.org/docs/developer/vagrant.html) on how to setup an environment using Vagrant.
* Run the tests confirm that the environment is working.
* Improve the documentation if needed.

## Background
Bodhi is a complex system that interface with many other applications, the development environment based on vagrant allows to have an integrated environment with developer tools available.

There are three main tasks for this ticket:

## Follow setup instructions for a development environment

Run through the existing [documentation](https://bodhi.fedoraproject.org/docs/developer/vagrant.html) to set up a development environment. Completing this helps you iterate new changes and test them. When you complete this, post a screenshot of bodhi running locally in your browser. 💻 📷

## Run the tests to confirm that the environment is working.

Once you have the vagrant environment running, you can run the unit-tests and integration tests. This is important because it will allow you to check if the changes you make to Bodhi works.

## Improve documentation if needed

If everything correctly works, you don't need to do anything more. 💯

If the steps didn't work for you, compare what you did change to get it working to the documentation. If something is unclear or a step is missing, submit a pull request with the improvement. Additions don't need to be long – it's better to write in a more direct and simple way.

StarWonder67:
@cverna Is there any public chat forum where I can reach out to the mentors? I am an outreachy applicant and would like to contribute to this project.

cverna:
> @cverna Is there any public chat forum where I can reach out to the mentors? I am an outreachy applicant and would like to contribute to this project.

@StarWonder67 yes you can join our IRC channel using the riot.im app https://riot.im/app/#/room/#freenode_#bodhi:matrix.org

karmadolkar:
hello @cverna ! I'm Karma Dolkar from Indian Institute of Technology, Roorkee (India). I'm an Outreachy 2020 applicant and very keen to contribute to this project.

Here's a screenshot of Bodhi running locally in my browser: 
 
![Screenshot from 2020-03-08 00-29-43](https://user-images.githubusercontent.com/48379515/76150812-51354400-60d4-11ea-83bc-8a4ac9df0f4b.png)

karmadolkar:
In my opinion, the documentation should include commands specific to commonly used OS, and not just Fedora (example: libvirtd is not a service in Ubuntu that requires enabling). I had to Google a bit regarding the Ubuntu commands so I thought I'd share a suggestion. :smile:  @cverna

[...]

PR #3947, opened 2020-03-09
doc change for setting up environment in Ubuntu

karmadolkar:
#3922 @cverna

### DIFF
--- a/docs/developer/vagrant.rst
+++ b/docs/developer/vagrant.rst
@@ -11,6 +11,9 @@ get started, simply use these commands::
     $ sudo systemctl enable libvirtd
     $ sudo systemctl start libvirtd
 
+If you're on Ubuntu, use this in place of the first command::
+		$ sudo apt install ansible libvirt-dev vagrant-libvirt vagrant-sshfs
+
 Check out the code and run ``vagrant up``::
 
     $ git clone https://github.com/fedora-infra/bodhi


docs/developer/vagrant.rst
```diff
@@ -11,6 +11,9 @@ get started, simply use these commands::
     $ sudo systemctl enable libvirtd
     $ sudo systemctl start libvirtd
 
+If you're on Ubuntu, use this in place of the first command::
```
cverna, line 14:
```suggestion
On distribution using a Debian base like Ubuntu, use this in place of the first command::
```

docs/developer/vagrant.rst
```diff
@@ -11,6 +11,9 @@ get started, simply use these commands::
...
+		$ sudo apt install ansible libvirt-dev vagrant-libvirt vagrant-sshfs
```
cverna, line 15: The indentation seems a little bit off, I think that it should be 4 spaces to look like the rest of the document

cverna (commented):
Thanks, just couple comments but looks good :+1:

cverna:
@karmadolkar Could you squash you commits to a single commit ? (https://gist.github.com/patik/b8a9dc5cd356f9f6f980#the-harder-and-less-flexible-way). 

Also you will need to sign off you commit using the `-s` option in the git commit command `git commit -s -m "This is signed off commit message `

accakks:
Hi @karmadolkar also an outreachy applicant here, this was really helpful for me as well, thankyou.

@cverna @karmadolkar However while using 2nd command, I got error
`Failed to enable unit: Unit file libvirtd.service does not exist.`

Installing virt-manager fixed this for me. 
```
sudo apt install virt-manager
sudo virt-manager

```
I understand we might not wanna give out instructions on installing everything, but we can add a heads up? (Example change below in bold)

> Before you get started, ensure that your host machine has virtualization extensions enabled in its BIOS so the guest doesn’t go slower than molasses. **Also make sure you have virt-manager installed** To get started, simply use these commands: ........

@cverna Let me know if this looks like something to add? 
(Because terminal message doesn't say that 'try installing vert-manager')

cverna (approved):

PR #3950, opened 2020-03-10
server: Close the db session when the approve_testing tasks succeed

### LINKED ISSUE #3849: Close the db session when the approve_testing tasks succeed
Currently we close the connection only when an exception happen. We should had a finally close to the try statement to close the connection in all the other cases.

https://github.com/fedora-infra/bodhi/blob/develop/bodhi/server/tasks/approve_testing.py

richardgreg:
Add a finally close to the try statement to close the connection in
all the other cases.

Fixes: #3849

Signed-off-by: Richard Gregory <richardgrecoson@gmail.com>

### DIFF
--- a/bodhi/server/tasks/approve_testing.py
+++ b/bodhi/server/tasks/approve_testing.py
@@ -47,6 +47,8 @@ def main():
                 db.commit()
     except Exception:
         log.exception("There was an error approving testing updates.")
+    finally:
+        db_factory._end_session()
 
 
 def approve_update(update: Update, db: Session):


richardgreg:
@cverna do you think there is something I can do differently in order for all tests to pass?

cverna:
> @cverna do you think there is something I can do differently in order for all tests to pass?

Yes not sure why the integration tests are failing now. I ll give it a look today and hopefully we can merge this after :)

cverna (approved):
Thanks

PR #3951, opened 2020-03-10
Make sure to escape the description of the error in the error template.

cverna:
We need to escape the error description since it can display
input from the user.


mattiaverga (approved):
There's a line too long for flake8 tests, everything else is good.

PR #3956, opened 2020-03-12
Associate bugs in changelog to rawhide updates.

### LINKED ISSUE #3925: RFE: Autoclose Rawhide bugs
I have just build rubygem-its. Because I have referenced bugzilla ticket in the changelog as well as in git log \[[1]\], I would expect Bodhi to associate this ticket with the Bodhi update \[[2]\] and handle the bug properly, i.e. close it as soon as the update lands in Rawhide. Unfortunately, this is probably not the case ATM.

[1]: https://src.fedoraproject.org/rpms/rubygem-rspec-its/c/3714b55268393f5398de3a56db57e7b19a32a2c7?branch=master
[2]: https://bodhi.fedoraproject.org/updates/FEDORA-2020-60bb269473

mattiaverga:
This will make possible to associate bugs to automatic rawhide updates by reading the notes from the changelog.

To add a bug use the following format: `fix(es)|close(s) (fedora|epel|rh|rhbz)#BUG_ID` (case insensitive).
For example: `- This will fix rhbz#123456`
or `- Closes EPEL#987654`

Fixes #3925 

### DIFF
--- a/bodhi/server/consumers/automatic_updates.py
+++ b/bodhi/server/consumers/automatic_updates.py
@@ -23,13 +23,14 @@ tagged with certain tags.
 """
 
 import logging
+import re
 
 import fedora_messaging
 
 from bodhi.server import buildsys
 from bodhi.server.config import config
-from bodhi.server.models import Build, ContentType, Package, Release
-from bodhi.server.models import Update, UpdateStatus, UpdateType, User
+from bodhi.server.models import (
+    Bug, Build, ContentType, Package, Release, Update, UpdateStatus, UpdateType, User)
 from bodhi.server.util import transactional_session_maker
 
 log = logging.getLogger('bodhi')
@@ -140,7 +141,9 @@ class AutomaticUpdateHandler:
 
             log.debug(f"Creating new update for {bnvr}.")
             changelog = build.get_changelog(lastupdate=True)
+            closing_bugs = []
             if changelog:
+                log.debug("Adding changelog to update notes.")
                 notes = f"""Automatic update for {bnvr}.
 
 ##### **Changelog**
@@ -148,11 +151,23 @@ class AutomaticUpdateHandler:
 ```
 {changelog}
 ```"""
+
+                for b in re.finditer(config.get('bz_regex'), changelog, re.IGNORECASE):
+                    idx = int(b.group(1))
+                    log.debug(f'Adding bug #{idx} to the update.')
+                    bug = Bug.get(idx)
+                    if bug is None:
+                        bug = Bug(bug_id=idx)
+                        dbsession.add(bug)
+                        dbsession.flush()
+                    if bug not in closing_bugs:
+                        closing_bugs.append(bug)
             else:
                 notes = f"Automatic update for {bnvr}."
             update = Update(
                 release=rel,
                 builds=[build],
+                bugs=closing_bugs,
                 notes=notes,
                 type=UpdateType.unspecified,
                 stable_karma=3,
--- a/bodhi/server/config.py
+++ b/bodhi/server/config.py
@@ -278,6 +278,10 @@ class BodhiConfig(dict):
         'bz_products': {
             'value': ['Fedora', 'Fedora EPEL', 'Fedora Modules'],
             'validator': _generate_list_validator(',')},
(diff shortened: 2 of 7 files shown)


cverna:
This looks great, I wonder if we should add something in the docs tho ? so that it is easy to point people to what syntax is supported.
What do you think ?

mattiaverga:
> This looks great, I wonder if we should add something in the docs tho ? so that it is easy to point people to what syntax is supported.
> What do you think ?

Yes, I thought about that, but I'm unsure where to put this information... should I add a new section specifically for rawhide updates?

Another thing that just came to mind, I should probably put the regex as a config parameter, so that other Bodhi instances can customize that for their needs. I'll do that in the weekend if you agree (I'm really busy at work and don't have much time in these days).

cverna:
> > This looks great, I wonder if we should add something in the docs tho ? so that it is easy to point people to what syntax is supported.
> > What do you think ?
> 
> Yes, I thought about that, but I'm unsure where to put this information... should I add a new section specifically for rawhide updates?

Yeah maybe a something after the update states description. 

> 
> Another thing that just came to mind, I should probably put the regex as a config parameter, so that other Bodhi instances can customize that for their needs. I'll do that in the weekend if you agree (I'm really busy at work and don't have much time in these days).

Sounds good about adding the regex to the configuration. And no stress about when you can work on that :-). 



> > This looks great, I wonder if we should add something in the docs tho ? so that it is easy to point people to what syntax is supported.
> > What do you think ?
> 
> Yes, I thought about that, but I'm unsure where to put this information... should I add a new section specifically for rawhide updates?
> 
> Another thing that just came to mind, I should probably put the regex as a config parameter, so that other Bodhi instances can customize that for their needs. I'll do that in the weekend if you agree (I'm really busy at work and don't have much time in these days).

bodhi/server/consumers/automatic_updates.py
```diff
@@ -140,19 +141,33 @@ def __call__(self, message: fedora_messaging.api.Message) -> None:
...
                 notes = f"""Automatic update for {bnvr}.
 
 ##### **Changelog**
 
 ```
 {changelog}
 ```"""
+
+                for b in re.finditer(config.get('bz_regex'), changelog, re.IGNORECASE):
+                    idx = int(b.group(1))
```
cverna, line 156: Is there a risk for b.group(1) to be None in case nothing is found ? If so that would make the casting break here

cverna (commented):
Just one comment otherwise Looks good to me

bodhi/server/consumers/automatic_updates.py
mattiaverga: If the regex doesn't match, we will never enter the for loop. If it matches, then `b.group(1)` cannot be None.

cverna: Cool just wanted to make sure of that :+1:

cverna (approved):

voxik:
Just wonder, does it also support `Resolves: rhbz#123546`, because this is format used RH internally.

mattiaverga:
> Just wonder, does it also support `Resolves: rhbz#123546`, because this is format used RH internally.

Not currently, but we can easily add support for that by customizing `bz_regex` in Bodhi config. We can do that either by editing the default regex or customize `production.ini` when we will deploy this in production, @cverna what's your preference?

I think this should work:
`r'(?:fix(?:es)?|close(?:s)|resolve(?:s)(?:\:)?)\s(?:fedora|epel|rh(?:bz)?)#(\d{5,})'`

cverna:
> > Just wonder, does it also support `Resolves: rhbz#123546`, because this is format used RH internally.
> 
> Not currently, but we can easily add support for that by customizing `bz_regex` in Bodhi config. We can do that either by editing the default regex or customize `production.ini` when we will deploy this in production, @cverna what's your preference?
> 
> I think this should work:
> `r'(?:fix(?:es)?|close(?:s)|resolve(?:s)(?:\:)?)\s(?:fedora|epel|rh(?:bz)?)#(\d{5,})'`

I don't have a strong preference, having it in the default will make it less likely to be forgotten when deploying the change, but changing the configuration is also fine with me.

PR #3959, opened 2020-03-14
Set the min expire date to tomorrow in UI

### LINKED ISSUE #744: creation of buildroot override fails with "Expiration_date : Expiration date in the past"
It's currently 23:10 in my timezone, I presume that wherever the server is, it is already after midnight. The default value set in "Expiration data" box is "tomorrow", which apparently is already today.
After changing the date to one day later submitting went fine.

mattiaverga:
Fix #744


cverna (approved):
Thanks

PR #3960, opened 2020-03-15
up.date_testing calculated in case of reset

### LINKED ISSUE #3493: Karma limits reached too early after being reset
In [this update](https://bodhi.fedoraproject.org/updates/FEDORA-2019-729bba50f9), karma was added as follows:
* -1
* "This update has been pushed to testing."
* "Bodhi is disabling automatic push to stable due to negative karma. The maintainer may push manually if they determine that the issue is not severe."
* +1
* New builds added; "Karma has been reset."
* "This update has been pushed to testing."
* +1 (repeat of the first +1)
* +1 (reverting the original -1)
* "This update can be pushed to stable now if the maintainer wishes"

The stable limits are 3, and there _are_ 3 +1's, but the first one was before the reset, so there should only be +2 karma. It also shows only +2 karma, and it's only been in testing 2 days, so that last comment should not have been made.

karmadolkar:
ref #3493 


bodhi/server/models.py
```diff
@@ -2430,8 +2432,6 @@ def edit(cls, request, data):
 
         # And, updates with new or removed builds always get their karma reset.
         # https://github.com/fedora-infra/bodhi/issues/511
```
cverna: This comment is not relevant here any more, we should probably move it with `data['karma_critipath'] =0` line.

cverna:
There is a unit test failing `bodhi.tests.server.services.test_updates.TestUpdatesService.test_meets_testing_requirements_since_karma_reset_critpath`, I think you need to change it to not use `update.date_testing` since this can now be none, using `utcnow` should be fine.

karmadolkar:
https://github.com/fedora-infra/bodhi/blob/develop/bodhi/tests/server/services/test_updates.py#L5525 

Do I remove this line?

cverna:
> https://github.com/fedora-infra/bodhi/blob/develop/bodhi/tests/server/services/test_updates.py#L5525
> 
> Do I remove this line?

You can see the error --> https://ci.centos.org/job/bodhi-pipeline/job/PR-3960/1/testReport/junit/bodhi.tests.server.services.test_updates/TestUpdatesService/test_meets_testing_requirements_since_karma_reset_critpath/

You can also run the tests locally in the vagrant box. I will let you search a little bit on your own, let me know if you need more pointers.

karmadolkar:
For running the test I executed something like:
`pytest-3 bodhi/tests/server/services/test_updates::TestUpdatesService::test_meets_testing_requirements_since_karma_reset_critpath`
However, file not found error came up.

Also, the reason for the error in the test was that None type and DateTime type couldn't be added. So, as you suggested, we should replace 
https://github.com/fedora-infra/bodhi/blob/develop/bodhi/tests/server/services/test_updates.py#L5548
with 
`self.date_testing = datetime.utcnow() + timedelta(days=7)`

mattiaverga:
> For running the test I executed something like:
> `pytest-3 bodhi/tests/server/services/test_updates::TestUpdatesService::test_meets_testing_requirements_since_karma_reset_critpath`
> However, file not found error came up.
> 
You're missing the `.py` extension in the filename.
Also I usually run local tests with few parameters that makes tests output less cluttered:
`pytest-3 --no-cov --disable-warnings bodhi/tests/server/services/test_updates.py::TestUpdatesService::test_meets_testing_requirements_since_karma_reset_critpath`

karmadolkar:
Yep!! Now it's working fine. I'll make a commit right now. :smile:

bodhi/server/models.py
cverna: @karmadolkar Just this left ^^. We should either remove that comment or move it with the code.

karmadolkar:
It is throwing an Assertion Error for the CI tests.

cverna:
> It is throwing an Assertion Error for the CI tests.

Yeah :( I don't think it is related to your change. Bodhi CI tests are not super stable unfortunately

bodhi/tests/server/services/test_updates.py
```diff
@@ -5545,7 +5545,7 @@ def test_meets_testing_requirements_since_karma_reset_critpath(self, *args):
```
mattiaverga: We should check that `date_testing` gets blanked just like `karma` is reset to zero, since this is the scope of this PR. Can you add an assert statement? I don't think we need a separate test for that, we can do this just here.

mattiaverga (dismissed):

bodhi/tests/server/services/test_updates.py
karmadolkar: I did the change and tried seeing if the failing CI tests are working. 

`pytest-3 --no-cov --disable-warnings devel/ci/integration/tests/test_bodhi_cli.py::TestUpdatesService::test_meets_testing_requirements_since_karma_reset_critpath`

It's throwing a file not found error.

mattiaverga: You can't run integration tests with pytest.
You need to use `bci` command from the vagrant bodhi shell. Use `bci --help` to list the available commands.

karmadolkar:
In the vagrant shell, I ran `flake8-3` command to check for the failing PEP-8 test, which is now working (at least locally it was). I also ran `btest` in vagrant to check for 100% code coverage.

cverna (approved):
:+1: This looks good to me. We just a a new file in /news for the release notes (see https://bodhi.fedoraproject.org/docs/developer/index.html#release-notes) . And we need @mattiaverga to also give it another look to check that his comments were addressed.

karmadolkar:
@cverna the Should the file name be PR3960.bug? What about its content?

mattiaverga:
It looks good to me also.
I'll wait to clear the request of changes until the news file is added, so that mergify will not close this PR (I think...).

@karmadolkar look at https://bodhi.fedoraproject.org/docs/developer/index.html#release-notes
The file should be named `3493.bug` and contains a brief description of what has been changed. For example "Reset `Update.date_testing` when editing builds"

news/3493.bug
```diff
@@ -0,0 +1 @@
+Reset update.date_testing when editing builds.
```
mattiaverga: https://bodhi.fedoraproject.org/docs/developer/index.html#release-notes
"It should not end with a `.` (full stop)"

PR #3963, opened 2020-03-19
Use jquery-typeahead for bodhi searchbar

### LINKED ISSUE #1455: new theme requires locating tiny magnifying glass to show search box

### LINKED ISSUE #2828: No candidates available in devel environment
When running a devel environment in Vagrant, no candidate builds are available for creating an upgrade and no candidates are showed in `/latest_candidates` page.

If I'm correct, the data which should provide those candidates is in `bodhi/server/buildsys.py`. 
https://github.com/fedora-infra/bodhi/blob/ccec39aa26465581a955a3de4c14190455fa2dbd/bodhi/server/buildsys.py#L429-L446
I think the problem is that `DevBuildsys.__tagged__` is empty, so no build is returned.

### LINKED ISSUE #3962: Search bar on landing page apparently does nothing
Using either Firefox or Epiphany, the search bar on [bodhi's homepage](https://bodhi.fedoraproject.org/?term=foo) seems to do nothing. It just reloads the current page and does not perform any search. In contrast, the search bar on [the updates page](https://bodhi.fedoraproject.org/updates/?search=foo) works perfectly fine. It would be nice if the search bar on the homepage would search for updates like the search bar on the updates pages does. Otherwise, it should be removed.

mattiaverga:
This will replace the current search component (twitter-typeahead) with an updated one (jquery-typeahead) and will make the search field always visible:
immagine

twitter-typeahead hasn't been updated for many years now, I think it's time to switch to something more updated. I've chosen jquery-typeahead because it's simpler and it also fits well for our searchbar grouping feature.

What this PR does:

- replace twitter-typeahead with jquery-typeahead in bodhi searchbar
- responsive navbar
- always show the search input field without faders
- populate `latest_candidates` list in devel environment
- use twitter-typeahead in the new overrides form (currently broken)

I did not remove the `bodhi/server/static/vendor/typeahead` directory for the moment.

This will supersede  #3890 
Fix #1455
Fix #3962 
Fix #2828 


cverna (approved):
Looks good to me

Issue #3944, opened 2020-03-06
builds seem to cumulate in updates-pending tag, an untagging bug?


kparal:
We have extremely high numbers of builds in `updates-pending` tags. For example:
```
$ koji list-tagged f31-updates-pending | wc -l
1493
$ koji list-tagged f32-updates-pending | wc -l
6894
```
The F32 numbers could be caused by F32 freeze, even though the number seems excessively high even when frozen. But F31 numbers are definitely wrong, because the last F31 updates push was **today**, and yet there are 1500 builds in there(!). Some builds are even duplicated/triplicated:
```
$ koji list-tagged f31-updates-pending | grep -E '^(Cython|anaconda|binutils|buildah|exiv2|hgview)'
Cython-0.29.13-1.fc31                     f31-updates-pending   ignatenkobrain
Cython-0.29.13-2.fc31                     f31-updates-pending   limb
anaconda-31.21-1.fc31                     f31-updates-pending   jkonecny
anaconda-31.22-1.fc31                     f31-updates-pending   m4rtink
binutils-2.32-22.fc31                     f31-updates-pending   nickc
binutils-2.32-23.fc31                     f31-updates-pending   nickc
binutils-2.32-24.fc31                     f31-updates-pending   nickc
buildah-1.11.0-0.11.dev.git60d5480.fc31   f31-updates-pending   lsm5
buildah-1.11.0-0.15.dev.git2df08f0.fc31   f31-updates-pending   lsm5
buildah-1.11.0-0.17.dev.git3f5436f.fc31   f31-updates-pending   lsm5
exiv2-0.27.2-0.2.RC2.fc31                 f31-updates-pending   rdieter
exiv2-0.27.2-1.fc31                       f31-updates-pending   rdieter
hgview-1.10.5-1.fc31                      f31-updates-pending   kiilerix
hgview-1.10.5-2.fc31                      f31-updates-pending   kiilerix
hgview-1.10.5-4.fc31                      f31-updates-pending   kiilerix
```
It looks like there's some bug regarding untagging, and the builds cumulate in this tag. I found this problem because our Taskotron's rpmdelint task downloads the whole tag contents, and our test machines started running out of disk space. I'll have to disable rpmdeplint until this is resolved somehow (cc @tflink).

mattiaverga:
I've noticed that if an update is still in pending->testing and gets enough karma to reach the stable state, it will go to pending->stable and `testing-pending` tag is not removed from its builds.

For example:
[...]

PR #3964, opened 2020-03-20
removed pending_testing tag when self.request is still in UpdateReque…

karmadolkar:
…st.testing

Ref #3944 


bodhi/server/models.py
```diff
@@ -2866,6 +2866,8 @@ def set_request(self, db, action, username):
             self.add_tag(self.release.pending_signing_tag)
         elif action is UpdateRequest.stable:
             self.add_tag(self.release.pending_stable_tag)
+            if self.request is UpdateRequest.testing:
+                self.remove_tag(self.release.pending_testing_tag)
```
mattiaverga, line 2876: We should check this really happens in tests. You can use [this test](https://github.com/fedora-infra/bodhi/blob/23505cf56ff64b83a65da7ffb728e5976a6afd31/bodhi/tests/server/test_models.py#L3338).

I think you'll need to use `mock.assert_called_with()`... take your time to learn how to use the mock library.

mattiaverga: ah, and as pointed out by lgtm-bot comment, use `==` instead of `is` ;-)

karmadolkar: Hi! Sorry for the late response. 

So we need to check if we encountered a case where an update went from pending to stable directly (due to fulfillment of the karma criteria). This is done using the assert statement on the test you gave. If this is true, then we know for sure that the action required for the update to go to stable but the request was to go to testing, i.e. the update is in the pending state wanting to go stable. Hence, the pending-testing tag is needed to be removed.

Could you please help me out with this? 

```
from unittest import mock
from mock import Mock
#to line 3351
mock = Mock()
mock.set_request.assert_called_once_with(self=self.obj, db=self.db , action=UpdateRequest.stable, username=req.user.name)
```

I ran the test using `pytest-3 --no-cov --disable-warnings bodhi/tests/server/test_models.py::TestUpdate::test_set_request_pending_stable` and `AssertionError: Expected 'set_request' to be called once. Called 0 times.` appeared.

karmadolkar:
Hi, sorry for a few silly errors in my last doubt. I called the wrong object in the wrong way, that was causing the error. The test was working locally. @mattiaverga 

Edit: There was a flake8 error due to a whitespace in the commas between arguments. I've corrected it

karmadolkar:
@mattiaverga @cverna kindly look into this :)

bodhi/tests/server/test_models.py
```diff
@@ -3349,6 +3349,9 @@ def test_set_request_pending_stable(self):
 
         assert self.obj.request == UpdateRequest.stable
         assert self.obj.status == UpdateStatus.pending
+        self.obj = mock.Mock()
+        self.obj.set_request(self.db, UpdateRequest.stable, req.user.name)
+        self.obj.set_request.assert_called_once_with(self.db, UpdateRequest.stable, req.user.name)
```
cverna: I think we should rather check that the remove_tag method was called since this the was we are adding in this change

cverna:
@karmadolkar one comment, you will also need to add a file in the news folder for the release notes :)

cverna (approved):
LGTM

Issue #3946, opened 2020-03-08
Make CI compose tests more reliable


mattiaverga:
In the past days we had some integration tests failing. It seems that when there's a failed compose we get different results from the expected:

**devel.ci.integration.tests.test_bodhi_cli.test_composes_list**
```
    def test_composes_list(bodhi_container, db_container):
        """Test ``bodhi composes list``"""
        result = _run_cli(bodhi_container, ["composes", "list"])
        assert result.exit_code == 0
        # Parse command output
        updates_by_compose = {}
        output_parser = re.compile(r"[\*\s]?([\w-]+)\s*:\s+(\d+) updates \((\w+)\)")
        for line in result.output.splitlines():
            match = output_parser.match(line)
            assert match is not None
            updates_by_compose[match.group(1)] = (int(match.group(2)), match.group(3))
        # Look in the DB for what is expected
        expected = {}
        query = """SELECT
        r.name, c.request, COUNT(u.id), c.state
        FROM composes c
        JOIN releases r ON r.id = c.release_id
        JOIN updates u ON u.release_id = r.id AND u.request = c.request
        WHERE (r.state = 'current' OR r.state = 'pending') AND u.locked = TRUE
        GROUP BY r.name, c.request, c.state
        """
        db_ip = db_container.get_IPv4s()[0]
        conn = psycopg2.connect("dbname=bodhi2 user=postgres host={}".format(db_ip))
        with conn:
            with conn.cursor() as curs:
                curs.execute(query)
                for record in curs:
                    compose = "{}-{}".format(record[0], record[1])
                    expected[compose] = (record[2], record[3])
        conn.close()
>       assert updates_by_compose == expected
E       AssertionError: assert {'F33C-testin...(0, 'failed')} == {}
E         Left contains more items:
E         {'F33C-testing': (0, 'failed')}
E         Full diff:
E         - {'F33C-testing': (0, 'failed')}
E         + {}

devel/ci/integration/tests/test_bodhi_cli.py:205: AssertionError
```

**devel.ci.integration.tests.test_bodhi_tasks.test_push_composer_start**
```
    def test_push_composer_start(bodhi_container, db_container, rabbitmq_container):
        try:
            output = bodhi_container.execute(
                ["bodhi-push", "--username", "ci", "-y"],
                exec_create_kwargs={"environment": {"PYTHONWARNINGS": "ignore"}},
            )
        except ConuException as e:
            assert False, str(e)
[...]

PR #3965, opened 2020-03-21
Make CI tests more reliable

mattiaverga:
This should avoid some (not all) of the errors pointed out in #3946 
I have revised the logic behind those two tests, please check if it makes sense to you.

### DIFF
--- a/devel/ci/integration/tests/test_bodhi_cli.py
+++ b/devel/ci/integration/tests/test_bodhi_cli.py
@@ -349,6 +349,7 @@ def test_updates_query_details(bodhi_container, db_container, greenwave_containe
         "FROM updates "
         "JOIN users ON updates.user_id = users.id "
         "JOIN releases ON updates.release_id = releases.id "
+        "WHERE releases.state = 'current' "
         "ORDER BY date_submitted DESC LIMIT 1"
     )
     query_comments = (
--- a/devel/ci/integration/tests/test_bodhi_cli.py
+++ b/devel/ci/integration/tests/test_bodhi_cli.py
@@ -186,11 +186,11 @@ def test_composes_list(bodhi_container, db_container):
     # Look in the DB for what is expected
     expected = {}
     query = """SELECT
-    r.name, c.request, COUNT(u.id), c.state
+    r.name, c.request, SUM(CASE WHEN u.locked = TRUE THEN 1 ELSE 0 END), c.state
     FROM composes c
     JOIN releases r ON r.id = c.release_id
     JOIN updates u ON u.release_id = r.id AND u.request = c.request
-    WHERE (r.state = 'current' OR r.state = 'pending') AND u.locked = TRUE
+    WHERE (r.state = 'current' OR r.state = 'pending')
     GROUP BY r.name, c.request, c.state
     """
     db_ip = db_container.get_IPv4s()[0]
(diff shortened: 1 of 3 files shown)


cverna:
Thanks for working on this, changes looks good to me :+1: 

Do you want to push another commit to fix the flake8 error ?

cverna (approved):

mattiaverga:
Now it randomly fails because of:
```
06:08:25.629 container.py      INFO   running command ['bodhi', 'updates', 'request', '--url', 'http://localhost:8080', '--openid-api', 'http://id.dev.fedoraproject.org/api/v1/', '--user', 'guest', '--password', 'ipsilon', 'FEDORA-2020-8e09d34dc3', 'stable']
06:08:36.623 container.py      INFO   ACL validation mechanism was unable to determine ACLs.
06:08:36.758 container.py      ERROR  command failed
```
It seems `http://id.dev.fedoraproject.org/api/v1/` address isn't reachable from by browser also. Shouldn't be `https://id.stg.fedoraproject.org/api/v1/`?

cverna:
> Now it randomly fails because of:
> 
> ```
> 06:08:25.629 container.py      INFO   running command ['bodhi', 'updates', 'request', '--url', 'http://localhost:8080', '--openid-api', 'http://id.dev.fedoraproject.org/api/v1/', '--user', 'guest', '--password', 'ipsilon', 'FEDORA-2020-8e09d34dc3', 'stable']
> 06:08:36.623 container.py      INFO   ACL validation mechanism was unable to determine ACLs.
> 06:08:36.758 container.py      ERROR  command failed
> ```
> 
> It seems `http://id.dev.fedoraproject.org/api/v1/` address isn't reachable from by browser also. Shouldn't be `https://id.stg.fedoraproject.org/api/v1/`?

So id.dev.fp.o in the integration test is just an ipsilon container that is run locally so if the container is up and running the test should be able to access it.

See https://github.com/fedora-infra/bodhi/blob/develop/devel/ci/integration/tests/fixtures/ipsilon.py

cverna:
> > Now it randomly fails because of:
> > ```
> > 06:08:25.629 container.py      INFO   running command ['bodhi', 'updates', 'request', '--url', 'http://localhost:8080', '--openid-api', 'http://id.dev.fedoraproject.org/api/v1/', '--user', 'guest', '--password', 'ipsilon', 'FEDORA-2020-8e09d34dc3', 'stable']
> > 06:08:36.623 container.py      INFO   ACL validation mechanism was unable to determine ACLs.
> > 06:08:36.758 container.py      ERROR  command failed
> > ```
> > 
> > 
> > It seems `http://id.dev.fedoraproject.org/api/v1/` address isn't reachable from by browser also. Shouldn't be `https://id.stg.fedoraproject.org/api/v1/`?
> 
> So id.dev.fp.o in the integration test is just an ipsilon container that is run locally so if the container is up and running the test should be able to access it.
> 
> See https://github.com/fedora-infra/bodhi/blob/develop/devel/ci/integration/tests/fixtures/ipsilon.py

Ok so the `test_updates_request` fails because it is using an update that does not have a build associated with it --> FEDORA-2020-8e09d34dc3.
We could try to make sure that we have builds in the update used in the test.

The second test `test_updates_query_details` seems to fail for another reason, looking into it

mattiaverga:
I've tweaked `test_updates_request` to only search for updates in testing, so that we'll not retrieve weird updates in strange statuses like that one.

I've also changed `test_updates_query_details`so that if there are no comments to the update, karma will be forced to 0.

cverna:
Ha forgot I approved that before, anyhow the tests are passing now :)

mattiaverga:
Yeah, but I don't think this is definitively fixed... at least the `fix_composes_list` didn't fix the error yesterday.
However I have identified that specific problem and I'll try to reproduce and fix it locally.

PR #3967, opened 2020-03-23
Get the update object in the celery worker from the database.

### LINKED ISSUE #3966: Celery worker used to tag builds are failing.
We have the following exception when trying to tag a side-tag update

```
[2020-03-23 13:46:12,096: ERROR/ForkPoolWorker-4] Task handle_side_and_related_tags[0049d0e8-6830-41e0-97c1-b4208714b70f] raised unexpected: AttributeError("'dict' object has no attribute 'release'")
  File "/usr/lib/python3.7/site-packages/celery/app/trace.py", line 382, in trace_task
    R = retval = fun(*args, **kwargs)
  File "/usr/lib/python3.7/site-packages/celery/app/trace.py", line 641, in __protected_call__
    return self.run(*args, **kwargs)
  File "/usr/lib/python3.7/site-packages/bodhi/server/tasks/__init__.py", line 134, in handle_side_and_related_tags_task
    main(updates, from_tag)
  File "/usr/lib/python3.7/site-packages/bodhi/server/tasks/handle_side_and_related_tags.py", line 39, in main
    release = update.release
AttributeError: 'dict' object has no attribute 'release'
```

cverna:
This commit makes sure that we get the update object from
the database in the celery tasks.

Fixes #3966

### DIFF
--- a/bodhi/server/tasks/handle_side_and_related_tags.py
+++ b/bodhi/server/tasks/handle_side_and_related_tags.py
@@ -20,14 +20,13 @@
 import logging
 import typing
 
-from bodhi.server import buildsys
+from bodhi.server import buildsys, util
 from bodhi.server.models import Update
-
 
 log = logging.getLogger(__name__)
 
 
-def main(updates: typing.List[Update], from_tag: str):
+def main(updates: typing.List[dict], from_tag: str):
     """Handle side-tags and related tags for updates in Koji.
 
     Args:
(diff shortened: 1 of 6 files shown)


bodhi/server/tasks/tag_update_builds.py
```diff
@@ -36,16 +36,19 @@ def main(update: Update, builds: typing.List[str]):
...
-            # Add the release's pending_signing_tag to all new builds
-            koji.tagBuild(update.release.pending_signing_tag, build)
-        else:
-            # EL6 doesn't have these, and that's okay...
-            # We still warn in case the config gets messed up.
-            log.warning(f'{update.release.name} has no pending_signing_tag')
-    koji.multiCall()
+    db_factory = util.transactional_session_maker()
+    with db_factory() as session:
+        update = session.query(Update).filter_by(alias=update["alias"]).first()
```
pypingou, line 41: Should we care for when update is ``None`` (ie: the alias provided did not link to an entry in the DB)?
If the worker are over-loaded, we could have a race condition between when the task was put in the queue and when the worker gets to it (and have someone delete the update in between).

bodhi/server/tasks/handle_side_and_related_tags.py
```diff
@@ -20,46 +20,48 @@
...
-            side_tag_testing_pending = update.release.get_testing_side_tag(from_tag)
-            if not koji.getTag(side_tag_signing_pending):
-                koji.createTag(side_tag_signing_pending, parent=from_tag)
-            if not koji.getTag(side_tag_testing_pending):
-                koji.createTag(side_tag_testing_pending, parent=from_tag)
-                koji.editTag2(side_tag_testing_pending, perm="autosign")
+    db_factory = util.transactional_session_maker()
+    with db_factory() as session:
+        for update_alias in updates:
+            update = session.query(Update).filter_by(alias=update_alias["alias"]).first()
```
pypingou, line 40: Same as below, should we check that ``update is not None``?

pypingou (commented):
One question (present two times), otherwise looks fine :)

bodhi/server/tasks/tag_update_builds.py
cverna: I don't think it is possible to delete an update other than directly from the database. I am not sure it is really needed to handle that case ? What do you think ?

pypingou: I thought it was, maybe it did and got removed.
That does lower the chances of running into this quite a bit indeed.

Let me know if you want me to approve the PR as is or if you want to update it

PR #3969, opened 2020-03-24
Only pass scalar argument to celery (part 2).

cverna:
This avoids the celery enqueuer emitting SQL queries to resolve
attributes, and therefore opening new transactions.

### DIFF
--- a/bodhi/server/tasks/handle_side_and_related_tags.py
+++ b/bodhi/server/tasks/handle_side_and_related_tags.py
@@ -20,49 +20,67 @@
 import logging
 import typing
 
+import koji  # noqa: 401
+
 from bodhi.server import buildsys, util
 from bodhi.server.models import Update
 
 log = logging.getLogger(__name__)
 
 
-def main(updates: typing.List[dict], from_tag: str):
+def main(aliases: typing.List[str], from_tag: str):
     """Handle side-tags and related tags for updates in Koji.
 
     Args:
-        updates: one or more Update objects
+        aliases: one or more Update alias
         from_tag: the tag into which the builds were built
     """
-    koji = buildsys.get_session()
     db_factory = util.transactional_session_maker()
-    with db_factory() as session:
-        for update_alias in updates:
-            update = session.query(Update).filter_by(alias=update_alias["alias"]).first()
-            if update is not None:
-                release = update.release
-                if not release.composed_by_bodhi:
-                    # Before the Bodhi activation point of a release, keep builds tagged
-                    # with the side-tag and its associate tags. Validate that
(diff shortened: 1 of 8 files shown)


bodhi/server/tasks/handle_side_and_related_tags.py
```diff
@@ -20,49 +20,56 @@
...
-def main(updates: typing.List[dict], from_tag: str):
-    """Handle side-tags and related tags for updates in Koji.
-
-    Args:
-        updates: one or more Update objects
-        from_tag: the tag into which the builds were built
-    """
-    koji = buildsys.get_session()
+def main(aliases: typing.List[str], from_tag: str):
+    """Handle side-tags and related tags for updates in Koji."""
```
pypingou: We lost a quite a bit of the doc string here, is that intended?

pypingou (commented):
Aside from the docblock getting much small this looks fine to me.

I would recommend maybe expanding on the change in the commit message as it's not very clear why we're doing so. We basically changed from a list of dic to a list of string, both of which shouldn't involved the DB. I figure the issue is in building the dict which does require DB connections, and I think this may be nice to be documented in the commit message.

bodhi/server/tasks/handle_side_and_related_tags.py
cverna: > We lost a quite a bit of the doc string here, is that intended?

Yes we are now typing the arguments there is no need to duplicate that in the doc string. In my opinion it is better to have that info in only one place.

cverna: ha just realize the type of args was not in the orginal docstring. I guess we can keep it them :)

cverna:
Comment addressed :)

pypingou (approved):
Thanks! :)

bodhi/server/tasks/tag_update_builds.py
```diff
@@ -20,37 +20,56 @@
...
+                update.from_tag)
+            koji.tagBuild(side_tag_pending_signing, build)
+        elif update.release.pending_signing_tag:
+            # Add the release's pending_signing_tag to all new builds
+            koji.tagBuild(update.release.pending_signing_tag, build)
+        else:
+            # EL6 doesn't have these, and that's okay...
+            # We still warn in case the config gets messed up.
+            log.warning(f'{update.release.name} has no pending_signing_tag')
+        koji.multiCall()
```
mattiaverga: This must go out of the for loop

cverna: indeed :)

cverna:
Merging this since the CI and flake8 failure are unrelated. Once this is deployed in prod I ll look at the CI failures.

Issue #3209, opened 2019-05-02
Document the full set of bug trackers that can be referenced in Bodhi's markdown


bowlofeggs:
While reviewing #3173 today, I realized that we don't document the full list of bug trackers that Bodhi supports in its markdown:

https://github.com/fedora-infra/bodhi/blob/3ea106ae8a074f26fb6982c133ac85299704782c/bodhi/server/ffmarkdown.py#L65-L74

We do have a link under the update comment box that pops up a modal that mentions a few of these:

![Screenshot from 2019-05-02 12-30-30](https://user-images.githubusercontent.com/354506/57091075-5f594d80-6cd6-11e9-847a-16a1993cb81f.png)

That model is here:

https://github.com/fedora-infra/bodhi/blob/3ea106ae8a074f26fb6982c133ac85299704782c/bodhi/server/templates/fragments.html#L218-L243

We should also document this in our Sphinx project.

So to solve this ticket, two things should be done:

- [ ] Document the trackers in the modal.
- [ ] Add a section to our Sphinx docs about Bodhi markdown, and list these there as well.

bowlofeggs:
P.S. We can have the modal just link to the Sphinx docs - that might be better so we only have one place we need to maintain the list.

bowlofeggs:
Perhaps an even better option is to have the list automatically generated from the code.

binayakbehera999:
I would like to work on this @bowlofeggs . I am new to open source, could you guide how to start with this issue?

richardgreg:
@cverna pls assign me to this. Thanks.

richardgreg:
>We do have a link under the update comment box that pops up a modal that mentions a few of these:

@bowlofeggs There is a help icon in the update's comment box that alludes to Fedora-Flavored Markdown and that's it. No link to `/fragments.html`

PR #3970, opened 2020-03-24
webui: fix broken link for privacy policy

### LINKED ISSUE #3971: broken comments help (missing markdown link, broken privacy link)
This is how the comments help looks now:
bodhi

"Fedora-flavored Markdown" is no longer a link - it used to be a link, even though I don't know where exactly it pointed. Now, nobody can figure out what the Fedora enhancements are.

Privacy policy link is broken, you can see its actual value in the bottom left corner.

 bodhi-5.2.2

richardgreg:


mattiaverga:
I don't understand what are you trying to fix. Also, this change is wrong.

richardgreg:
@mattiaverga 
>I don't understand what are you trying to fix.

Inside of an update's comment box, the link to `this privacy policy` in the help icon (top right) is broken.

>Also, this change is wrong.

Yeah I noticed that. I didn't get the chance try it in my local machine because I'm finding it difficult to log into the app from `localhost:6543` and interact with the comment box. Some things I've tried already is removing `stg` from the values of `openid.provider openid.url openid_template` in `development.ini.example`, but it's not working. I'm trying to open an account  in `https://admin.stg.fedoraproject.org/accounts` but I have not received a confirmation email yet. I'm not sure how long it should be taking. Do you have any suggestions?

mattiaverga:
Now I see... I've never noticed that broken link. It's really an awful case of mixing multiple quotes and double quotes.
I think the best (or the only) solution is to assign `request.registry.settings.get('privacy_link')` to a variable and use that inside the href property.

About opening an account into staging fas, it's been a long time I did... not sure how long it takes to receive the confirmation email, but I suppose not a lot of time.

richardgreg:
@mattiaverga okay.
Also wrapping the strings in docstrings seems to not break anything, but so does the current implementation. But in this case, I notice that using single or double quotes interchangeably around `href` no longer raises any error.

No matter the approach, the main problem for me now is that I can't test the functionality because `Legal` and `Privacy Policy` links don't show up when running Bodhi in local machine. Is this a normal feature?

mattiaverga:
> 
> No matter the approach, the main problem for me now is that I can't test the functionality because `Legal` and `Privacy Policy` links don't show up when running Bodhi in local machine. Is this a normal feature?

You need to add `privacy_link = whatever` in `development.ini` under `[app:main]` section.

bodhi/server/templates/update.html
```diff
@@ -219,8 +219,9 @@ <h4>How to install</h4>
                   <div class="col pr-0">
                     <%
                       help_content = "Comments support Fedora-Flavored Markdown"
+                      privacy_link = request.registry.settings.get('privacy_link')
```
mattiaverga: I think it's better to assign the variable only after having checked `request.registry.settings.get('privacy_link')` is set. I would move this line inside the if statement below.

cverna (approved):
LGTM

richardgreg:
@cverna I opened this PR before #3971 was filed. I see that the 'fedora-flavoured markdown' help not popping up is also part of the issue. I talked about the pop-up when starting work on #3209. Is there any particular reason why the feature was retracted? If there is no problem I can work-on putting it back.

mattiaverga:
I would prefer to get rid of that modal popup and link that help to the doc page you are working on in #3985 (just opening it in a new window/tab). That will avoid us to maintain the help text in two different places.

If @cverna agrees to that, I think we can merge this PR as is and then @richardgreg can change #3985 to add the missing link.

cverna:
> I would prefer to get rid of that modal popup and link that help to the doc page you are working on in #3985 (just opening it in a new window/tab). That will avoid us to maintain the help text in two different places.

:+1: 

> 
> If @cverna agrees to that, I think we can merge this PR as is and then @richardgreg can change #3985 to add the missing link.

Sounds good to me

PR #3975, opened 2020-03-27
Ignore builds associated with unpushed Updates when validating

### LINKED ISSUE #1809: Can't create an update with the same build as was in an unpushed update
If you unpush an update with some build, you can't create a new update with the same build NVR. That makes it problematic for certain use cases, e.g. like merging two updates together.

See this discussion for further details:
https://lists.fedoraproject.org/archives/list/devel@lists.fedoraproject.org/thread/JCOLM4HHMSY24J734KASUT2ZIUNIS3AW/

### LINKED ISSUE #3715: Cannot create new update after unpushing old one
To work around the issue with [not being able to edit updates](https://github.com/fedora-infra/bodhi/issues/3714), I wanted to unpush the updates ([f30](https://bodhi.fedoraproject.org/updates/FEDORA-2019-7cafbe66ba), [f31](https://bodhi.fedoraproject.org/updates/FEDORA-2019-94393775ec)) and create new ones with some of the builds in higher version. However, when clicking "Submit", I am getting the following error: "Builds : Update for dnf-plugins-core-4.0.11-1.fc30 already exists"

Even though I think that editing the update would be better approach, I don't understand why unpushed update should influence a creation of new update.

bigjust:
Currently, if you create an Update with a build, then unpush it, you
cannot create a new Update with that same build. This commit ignores
builds assigned to Unpushed commits when validating duplicate builds.

Resolves #3715, #1809

### DIFF
--- a/bodhi/server/validators.py
+++ b/bodhi/server/validators.py
@@ -295,9 +295,10 @@ def validate_builds(request, **kwargs):
     for nvr in builds:
         build = request.db.query(Build).filter_by(nvr=nvr).first()
         if build and build.update is not None:
-            request.errors.add('body', 'builds',
-                               "Update for {} already exists".format(nvr))
-            return
+            if build.update.status != UpdateStatus.unpushed:
+                request.errors.add('body', 'builds',
+                                   "Update for {} already exists".format(nvr))
+                return
 
 
 @postschema_validator
--- a/bodhi/tests/server/services/test_updates.py
+++ b/bodhi/tests/server/services/test_updates.py
@@ -172,6 +172,17 @@ class TestNewUpdate(BasePyTestCase):
         assert 'Update for bodhi-2.0-1.fc17 already exists' in res
 
     @mock.patch(**mock_valid_requirements)
+    def test_unpushed_update(self, *args):
+
+        unpushed_update = self.create_update(['whoopsie-1.0.0-1.fc17'])
+        unpushed_update.status = UpdateStatus.unpushed
+        self.db.commit()
+
+        res = self.app.post_json('/updates/', self.get_update('whoopsie-1.0.0-1.fc17'))
+
+        assert res.json['alias'] != unpushed_update.alias
+
+    @mock.patch(**mock_valid_requirements)
     def test_invalid_requirements(self, *args):
         update = self.get_update()
         update['requirements'] = 'rpmlint silly-dilly'
--- a/bodhi/tests/server/services/test_updates.py
+++ b/bodhi/tests/server/services/test_updates.py
@@ -173,12 +173,13 @@ class TestNewUpdate(BasePyTestCase):
 
     @mock.patch(**mock_valid_requirements)
     def test_unpushed_update(self, *args):
-
+        """Allow posting a duplicate build if the old update is unpushed."""
         unpushed_update = self.create_update(['whoopsie-1.0.0-1.fc17'])
         unpushed_update.status = UpdateStatus.unpushed
         self.db.commit()
 
-        res = self.app.post_json('/updates/', self.get_update('whoopsie-1.0.0-1.fc17'))
+        with fml_testing.mock_sends(update_schemas.UpdateRequestTestingV1):
+            res = self.app.post_json('/updates/', self.get_update('whoopsie-1.0.0-1.fc17'))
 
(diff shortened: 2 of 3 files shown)


mattiaverga:
Just a quick note: since an unpushed update can still be re-pushed to testing, we must be sure that if a newer update with the same build exists, the old one cannot be revived (maybe we should obsolete it).

bigjust:
I'll try that approach. 

@mattiaverga Also is there a reason a user can't obsolete updates manually?

mattiaverga:
@bigjust unpushing an update will cause all associated builds to be pushed back to `*-candidate` tag. Obsoleting an update will cause all associated build to have all their tag removed.

So we can't obsolete the older update. It must be possible to create a new update with the same build already present in an older unpushed update, but at the same time, when unpushing an update we must check that a build in that update isn't already associated to an active update (pending, testing or stable).
I don't know if that check already happens, I'll need to make some tests.

mattiaverga (changes requested):
Well, after having a deeper look into this, when the same build is posted in a new update it also gets removed from the older one.
So I think we can safely allow this. Can you also add a news fragment as described at https://bodhi.fedoraproject.org/docs/developer/index.html#release-notes to the PR?

mattiaverga (approved):

mattiaverga:
You should rebase your commit on top of develop branch and sign the DCO.

bodhi/server/validators.py
```diff
@@ -295,9 +295,10 @@ def validate_builds(request, **kwargs):
     for nvr in builds:
         build = request.db.query(Build).filter_by(nvr=nvr).first()
         if build and build.update is not None:
-            request.errors.add('body', 'builds',
-                               "Update for {} already exists".format(nvr))
-            return
+            if build.update.status is not UpdateStatus.unpushed:
```
mattiaverga: Values compared using 'is not' when equivalence is not the same as identity. Use '!=' instead.

bodhi/tests/server/services/test_updates.py
```diff
@@ -171,6 +171,15 @@ def test_duplicate_update(self, *args):
                                  status=400)
         assert 'Update for bodhi-2.0-1.fc17 already exists' in res
 
+    @mock.patch(**mock_valid_requirements)
+    def test_unpushed_update(self, *args):
+
+        unpushed_update = self.create_update(['whoopsie-1.0.0-1.fc17'])
+        unpushed_update.status = UpdateStatus.unpushed
+        self.db.commit()
+
```
mattiaverga: ./bodhi/tests/server/services/test_updates.py:180:1: W293 blank line contains whitespace

bodhi/tests/server/services/test_updates.py
```diff
@@ -171,6 +171,15 @@ def test_duplicate_update(self, *args):
...
         assert 'Update for bodhi-2.0-1.fc17 already exists' in res
 
+    @mock.patch(**mock_valid_requirements)
+    def test_unpushed_update(self, *args):
+
+        unpushed_update = self.create_update(['whoopsie-1.0.0-1.fc17'])
+        unpushed_update.status = UpdateStatus.unpushed
+        self.db.commit()
+        
+        res = self.app.post_json('/updates/', self.get_update('whoopsie-1.0.0-1.fc17'))
```
mattiaverga: ./bodhi/tests/server/services/test_updates.py:181:9: F841 local variable 'res' is assigned to but never used

We're not checking anything here... we could check that the newly created update has a different alias than the older and that `whoopsie-1.0.0-1.fc17` doesn't belong anymore to the old update.

mattiaverga (changes requested):

bodhi/tests/server/services/test_updates.py
bigjust: We're implicitly checking that the Update creates and doesn't fail due to the same build being included in an Obsolete Update. I can make this check explicit, and this test certainly fails without the logic added in validators.py

ernestask:
So, there isn’t a whole lot to address here, it seems. Is the work going to be finished sometime soon?

mattiaverga (approved):
I've pushed a small fix to pass unit tests, overall it looks good.

PR #3977, opened 2020-03-28
[WIP] test for bodhi timeout on new update

mattiaverga:
Just some changes to find out why bodhi sometimes doesn't return after creating a new update.
Do not merge this!

### DIFF
--- a/bodhi/server/models.py
+++ b/bodhi/server/models.py
@@ -2299,17 +2299,21 @@ class Update(Base):
 
         log.debug("Adding new update to the db.")
         db.add(up)
-        log.debug("Triggering db commit for new update.")
-        db.commit()
-
-        if not data.get("from_tag"):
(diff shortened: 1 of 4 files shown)


cverna:
> Just some changes to find out why bodhi sometimes doesn't return after creating a new update.
> Do not merge this!

Hey @mattiaverga, I looked a bit at that last week and from the logs it looked like the `handle_update` celery task never returned. After looking a bit on internet it seems that this can happen if the rabbitmq broker becomes unvailable, by default the task will just wait forever.
I have made some change to the celery config, so that we retry the connection with the broker 3 times and then if it is still not working just timeout the task.

See  https://infrastructure.fedoraproject.org/cgit/ansible.git/commit/?id=9e0810dd300fad89f6d63c1598dbe3b6a9c36fea.

I ll keep monitoring the logs, to see if that fixed the problem or not.

PR #3980, opened 2020-03-30
Do not bound the database session to the TransactionalSessionMaker cl…

### LINKED ISSUE #3979: sqlalchemy.orm.exc.DetachedInstanceError exception in the composer task.
Since we have released Bodhi 5.2 we are getting a lot of sqlalchemy.orm.exc.DetachedInstanceError exception in the composer task leading to the compose failing and needing to be resumed.

cverna:
…ass.

Since Threads are sharing the memory binding the session object to the class
make it possible for threads to override a previous session leading to unexpected
behaviours.

Fixes #3979

### DIFF
--- a/bodhi/server/util.py
+++ b/bodhi/server/util.py
@@ -874,16 +874,16 @@ class TransactionalSessionMaker(object):
         Exceptions or rolls back the transaction. In either case, it also will close and remove the
         Session.
         """
-        self.session = Session()
+        session = Session()
         try:
-            yield self.session
-            self.session.commit()
+            yield session
+            session.commit()
         except Exception as e:
             # It is possible for session.rollback() to raise Exceptions, so we will wrap it in an
             # Exception handler as well so we can log the rollback failure and still raise the
             # original Exception.
             try:
-                self.session.rollback()
+                session.rollback()
             except Exception:
                 log.exception('An Exception was raised while rolling back a transaction.')
             raise e
@@ -897,7 +897,6 @@ class TransactionalSessionMaker(object):
         This has been split off the main __call__ method to make it easier to
         mock it out in unit tests.
         """
-        self.session.close()
         Session.remove()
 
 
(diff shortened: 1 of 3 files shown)


bodhi/server/util.py
```diff
@@ -897,7 +897,6 @@ def _end_session(self):
         This has been split off the main __call__ method to make it easier to
         mock it out in unit tests.
         """
-        self.session.close()
         Session.remove()
```
cverna, line 900: By default Session.remove() calls Session.close() so we don't need to call it twice here.

news/3979.bug
```diff
@@ -0,0 +1,4 @@
+Do not bound the database session to the TransactionalSessionMaker class.
+Since Threads are sharing the memory binding the session object to the class
+make it possible for threads to override a previous session leading to unexpected
```
pypingou: "...to the class, it makes it possible for threads..."?

news/3979.bug
```diff
@@ -0,0 +1,4 @@
+Do not bound the database session to the TransactionalSessionMaker class.
```
pypingou: Should we specify which database session?

pypingou (commented):
Two small comments on the news file, otherwise :+1: for me

news/3979.bug
cverna: This is the session created in the TransactionalSessionMaker class, no sure how I can be more specific. Idea ?

news/3979.bug
cverna: +1 will update that

news/3979.bug
pypingou: What about:
````
Do not bound the database session created using TransactionalSessionMaker class to the object created
````

news/3979.bug
pypingou: If we go via the suggestion above, then we may want to drop the "to the class" in the line above this one

pypingou (approved):
Thanks! :)

PR #3981, opened 2020-03-30
Avoid using a database session in the tag_update_builds_task.

cverna:
In the tag_update_builds_task we do not need a
database session we just need to pass the list of
builds and the koji tag to be used.

### DIFF
--- a/bodhi/server/tasks/tag_update_builds.py
+++ b/bodhi/server/tasks/tag_update_builds.py
@@ -20,56 +20,25 @@
 import logging
 import typing
 
-import koji  # noqa: 401
-
-from bodhi.server import buildsys, util
-from bodhi.server.models import Update
+from bodhi.server import buildsys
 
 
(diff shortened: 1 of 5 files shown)


bodhi/server/tasks/tag_update_builds.py
```diff
@@ -22,54 +22,25 @@
...
     """
-    koji = buildsys.get_session()
-    koji.multicall = True
-    db_factory = util.transactional_session_maker()
     try:
-        with db_factory() as session:
-            update = session.query(Update).filter_by(alias=alias).first()
-        if update is not None:
-            tag_builds(update, builds, koji)
+        kc = buildsys.get_session()
```
pypingou, line 37: The rename of the variable makes the diff a little harder to read

pypingou (approved):
Nice, this does make the overall code a lot more readable and one-less-db-session++

PR #3983, opened 2020-03-30
Avoid using a database session in the handle side tag task.

cverna:
In handle_side_and_related_tags_task we don't modify any object in the
database. Therefore we don't need to have a database session in the task.
We can just pass the value needed as argument.

### DIFF
--- a/bodhi/server/tasks/handle_side_and_related_tags.py
+++ b/bodhi/server/tasks/handle_side_and_related_tags.py
@@ -20,67 +20,49 @@
 import logging
 import typing
 
-import koji  # noqa: 401
+from bodhi.server import buildsys
 
-from bodhi.server import buildsys, util
-from bodhi.server.models import Update
 
 log = logging.getLogger(__name__)
 
 
-def main(aliases: typing.List[str], from_tag: str):
+def main(builds: typing.List[str],
+         pending_signing_tag: str,
+         from_tag: str,
+         pending_testing_tag: typing.Optional[str]):
     """Handle side-tags and related tags for updates in Koji.
 
     Args:
-        aliases: one or more Update alias
-        from_tag: the tag into which the builds were built
+        builds: a list of builds to be tagged.
+        pending_signing_tag: the pending signing tag to apply on the builds.
+        from_tag: the tag into which the builds were built.
+        pending_testing_tag: the pending_testing_tag to create if not None.
     """
-    db_factory = util.transactional_session_maker()
-    koji = buildsys.get_session()
     try:
-        with db_factory() as db:
-            for alias in aliases:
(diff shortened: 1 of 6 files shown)


bodhi/server/tasks/handle_side_and_related_tags.py
```diff
@@ -20,67 +20,49 @@
...
+                koji.createTag(pending_signing_tag, parent=from_tag)
+            if not koji.getTag(pending_testing_tag):
+                log.info(f"Create {pending_testing_tag} in koji")
+                koji.createTag(pending_testing_tag, parent=from_tag)
+                koji.editTag2(pending_testing_tag, perm="autosign")
+        else:
+            # If we don't provide a pending_testing_tag, then we have merged the
+            # side tag into the release pending_signing tag.
+            # We can remove the side tag.
+            koji.removeSideTag(from_tag)
```
cverna, line 59: I want to double check that we can call this before actually tagging the builds.

bodhi/server/tasks/handle_side_and_related_tags.py
```diff
@@ -20,67 +20,49 @@
...
-from bodhi.server.models import Update
 
 log = logging.getLogger(__name__)
 
 
-def main(aliases: typing.List[str], from_tag: str):
+def main(builds: typing.List[str],
+         pending_signing_tag: str,
+         from_tag: str,
+         pending_testing_tag: typing.Optional[str]):
```
pypingou, line 32: If it's optional shouldn't it have a default value?

pypingou:
One nitpick but otherwise :+1: for me.

Let me know once you've double checked :)

bodhi/server/tasks/handle_side_and_related_tags.py
cverna: Optional means it can be either the type specified (str here) or None. So I think having the default value is fine.

bodhi/server/tasks/handle_side_and_related_tags.py
cverna: So I have checked this and that will remove the tag from all the builds tagged with it. Since at this point the list of builds is persisted in bodhi. I believe this is ok to do that step here.

@pypingou do you see any issue with that ?

pypingou: Since we have merged the side-tags into the pending_signing the builds shouldn't end up tag-less, so :+1: this should be fine

pypingou (approved):
My questions have been answer, :+1: for me

mattiaverga:
The test failure is unrelated, I'm going to manually merge this.

PR #3985, opened 2020-04-02
Document the full set of bug trackers that can be referenced in Bodhi's markdown

### LINKED ISSUE #3209: Document the full set of bug trackers that can be referenced in Bodhi's markdown

richardgreg:
Also added a section to Bodhi's Sphinx docs about Bodhi markdown,
and list the bug trackers there as well.

Fixes: #3209

### DIFF
--- a/bodhi/server/templates/update.html
+++ b/bodhi/server/templates/update.html
@@ -218,7 +218,7 @@ if can_edit and update.release.composed_by_bodhi:
                 <div class="row">
                   <div class="col pr-0">
                     <%
-                      help_content = "Comments support Fedora-Flavored Markdown"
+                      help_content = "Comments support <a href='docs/user/fedora-flavored-markdown.html'>Fedora-Flavored Markdown</a>"
                       if request.registry.settings.get('privacy_link'):
                         privacy_link = request.registry.settings.get('privacy_link')
                         help_content = help_content + "<br/><br/>Comments are governed under <a href='{}'>this privacy policy</a>.".format(privacy_link)
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -35,6 +35,7 @@ User Guide
    user/update_states
    user/buildroot_overrides
    user/automatic_updates
+   user/fedora-flavored-markdown
    user/man_pages/index
 
 .. toctree::
--- a/docs/user/fedora-flavored-markdown.rst
+++ b/docs/user/fedora-flavored-markdown.rst
@@ -1,0 +1,114 @@
+========================
+Fedora-Flavored Markdown
+========================
+
+Description
+===========
+
+Text fields in Bodhi2 support an `enhanced`_ version of `markdown`_. This is a cheat
+sheet for your reference.
+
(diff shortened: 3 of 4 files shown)


mattiaverga:
Can you move the new page under `docs/user/`? The `man_pages` is for CLI commands.

Also, if cverna agrees to what I proposed in #3970 , you will need to edit the link to `Fedora-Flavoured Markdown` to this page and we can remove the old code from `fragments.html`.

richardgreg:
> Can you move the new page under `docs/user/`? The `man_pages` is for CLI commands.
> 
> Also, if cverna agrees to what I proposed in #3970 , you will need to edit the link to `Fedora-Flavoured Markdown` to this page and we can remove the old code from `fragments.html`.

Okay!

bodhi/server/templates/update.html
```diff
@@ -218,7 +218,7 @@ <h4>How to install</h4>
                 <div class="row">
                   <div class="col pr-0">
                     <%
-                      help_content = "Comments support Fedora-Flavored Markdown"
+                      help_content = "Comments support <a href='https://bodhi.fedoraproject.org/docs/user/fedora-flavored-markdown.html'>Fedora-Flavored Markdown</a>"
```
mattiaverga: It is better to use a relative URL here
`<a href='docs/user/fedora-flavored-markdown.html'>`

docs/index.rst
```diff
@@ -36,6 +36,11 @@ User Guide
    user/buildroot_overrides
    user/man_pages/index
 
+.. toctree::
+   :maxdepth: 1
```
mattiaverga: No need to duplicate toctree and maxdepth. If you don't want to expand subchapters, you can move  `user/fedora-flavored-markdown` just above `user/release_notes`, otherwise you can put the line between `user/buildroot_overrides` and `user/man_pages/index`

news/richardgrecoson.author
```diff
@@ -0,0 +1 @@
+Richard O. Gregory
```
mattiaverga: This file is already present, no need to add it again

mattiaverga (changes requested):

docs/index.rst
richardgreg: Sounds good!

mattiaverga (approved):
LGTM thanks

PR #3987, opened 2020-04-06
expand the validate_builds() error message to print a url 

### LINKED ISSUE #3953: expand the validate_builds() error messages to print the URL of the conflicting update
When I run `fedpkg update` after a crash (https://github.com/fedora-infra/bodhi/issues/3952) that left the system in an indeterminate state, I see this error message:
```
Could not execute update: Could not generate update request: Update for python-execnet-1.7.1-3.el7 already exists
```

It would be great to extend this error message to print the URL for the update to save the user some clicks. This would make it easier for new users to understand and navigate Bodhi.

```
... already exists at https://bodhi.fedoraproject.org/updates/FEDORA-EPEL-2020-1ccdc325aa
```

jennizalwango:
Added tests for validate_builds()
Expanded the error message for validate_builds()

Fixes [# 3953](https://github.com/fedora-infra/bodhi/issues/3953)

Signed-off-by: Jenny Zalwango jennyzalwango12@gmail.com


bodhi.template.last
```diff
@@ -0,0 +1,42 @@
+[foo-1-1.fc30]
```
cverna: This file should not be committed, it was only used by you during development :-)

cverna: Same for the following file `bodhi/server/__pycache__/buildsys.cpython-37.pyc.139873619268272 `

bodhi/server/validators.py
```diff
@@ -272,8 +273,9 @@ def validate_builds(request, **kwargs):
         up = request.db.query(Update).filter_by(alias=edited).first()
         if not up:
             request.errors.add('body', 'builds',
-                               'Cannot find update to edit: %s' % edited)
+                               'Cannot find update to edit: %s' % urllib.request.urlopen(edited))
```
cverna: instead of using urllib, you should be able to get the url using the `request.route_url` [method](https://docs.pylonsproject.org/projects/pyramid/en/latest/api/request.html#pyramid.request.Request.route_url).

bodhi/server/validators.py
```diff
@@ -272,8 +273,9 @@ def validate_builds(request, **kwargs):
...
             return
+            print('edited')
```
cverna: This should be removed :)

bodhi/tests/server/test_validators.py
```diff
@@ -619,6 +619,42 @@ def test_update_found_but_not_update_object(self, mock_update_get):
         assert request.errors.status == exceptions.HTTPBadRequest.code
 
 
+class TestValidateBuilds(BasePyTestCase):
```
cverna, line 622: I think we should actually test that we have an error and that the error message contains the url. Here we just seems to make sure we don't have an error.
What do you think ?

bodhi/server/validators.py
jennizalwango: Okay, l will modify it to instead use `request.route_url` method as advised

bodhi.template.last
jennizalwango: okay l will work on them

bodhi/tests/server/test_validators.py
jennizalwango: Actually your right, let me look into it.

jennizalwango:
@cverna  l implemented most of the feedback, however, l am a bit stuck at the part of testing that the error message printed includes a URL.
l get a name error 
Screenshot from 2020-04-11 20-38-59

cverna:
> @cverna l implemented most of the feedback, however, l am a bit stuck at the part of testing that the error message printed includes a URL.
> l get a name error
> Screenshot from 2020-04-11 20-38-59

seems like it should be printed_url = self.request instead of just request

jennizalwango:
Thanks @cverna
Currently my power is off and my machine blacked out but will definitely
look into it once power comes back

jennizalwango:
@cverna  Kindly note that l have finalized with the feedback given. And l have worked on it kindly have a look on it Thanks

bodhi/server/validators.py
```diff
@@ -272,9 +272,9 @@ def validate_builds(request, **kwargs):
         up = request.db.query(Update).filter_by(alias=edited).first()
         if not up:
             request.errors.add('body', 'builds',
-                               'Cannot find update to edit: %s' % edited)
+                               'Cannot find update to edit: %s' % request.route_url(edited))
             return
-
+
```
cverna: I don't think we need to add a new line here.

.gitignore
```diff
@@ -35,4 +35,3 @@ composer/
 nosetests.xml
 virtenv
 .vscode
-test_results
```
cverna: why are you removing this ?

cverna (commented):
There is still one file that we don't want to commit in that PR ( bodhi/server/__pycache__/buildsys.cpython-37.pyc.139873619268272)

.gitignore
jennizalwango: ohh l think l removed it unknowingly:smile:  l will add it then push again

bodhi/server/validators.py
jennizalwango: Let me remove it

.gitignore
jennizalwango: Done with this

jennizalwango:
@cverna  l am having trouble, l am not certain why some of the checks are failing on the PR and yet local the tests pass successful, ls there anything that perhaps is not doing being done right from my end? Anything that l can do for all the checks to pass please?

cverna:
@jennizalwango you can see which tests are failing here https://ci.centos.org/job/bodhi-pipeline/job/PR-3987/7/#showFailuresLink

From giving a quick look at it they should also fail locally.

cverna:
Closing this due to lack of activity

PR #3989, opened 2020-04-07
Split handle_update task into two celery tasks for bugs and testcases

### LINKED ISSUE #3972: Bugs no longer set to MODIFIED when update is submitted
IIRC, it used to be the case that when an update that fixes a bug was submitted, Bodhi would post a comment on the bug and set it to MODIFIED. However, it no longer seems to do this. Bodhi only seems to touch the bug for the first time when the update is pushed to testing, at which point it goes straight to ON_QA. See for e.g. https://bugzilla.redhat.com/show_bug.cgi?id=1817031 (as I write this, an update has been submitted, but there is no comment from Bodhi on the bug) or https://bugzilla.redhat.com/show_bug.cgi?id=1809717 (a case where you can see the bug go straight from NEW to ON_QA).

mattiaverga:
This will split the `updates_handler` celery task in two separate tasks for bugs and testcases, so that we can safely retry to fetch testcase details or work on bugs in case of temporary connection problems.

The other logic behind handling the update is moved in the two methods under Update model that used `updates_handler` - `edit()` and `set_request()` - without too much code duplication (only the test gating check was in common by the two).
We may want to move the greenwave query to a separate celery task also, maybe?

### DIFF
--- a/bodhi/server/models.py
+++ b/bodhi/server/models.py
@@ -30,7 +30,7 @@ import uuid
 from urllib.error import URLError
 
 from simplemediawiki import MediaWiki
-from sqlalchemy import (and_, Boolean, Column, DateTime, event, ForeignKey,
+from sqlalchemy import (and_, Boolean, Column, DateTime, event, func, ForeignKey,
                         Integer, or_, Table, Unicode, UnicodeText, UniqueConstraint)
 from sqlalchemy.ext.declarative import declarative_base
 from sqlalchemy.orm import class_mapper, relationship, backref, validates
@@ -45,8 +45,8 @@ from bodhi.messages.schemas import (buildroot_override as override_schemas,
                                     errata as errata_schemas, update as update_schemas)
 from bodhi.server import bugs, buildsys, log, mail, notifications, Session, util
 from bodhi.server.config import config
-from bodhi.server.exceptions import BodhiException, LockedUpdateException
-from bodhi.server.tasks import handle_update, tag_update_builds_task
+from bodhi.server.exceptions import BodhiException, ExternalCallException, LockedUpdateException
+from bodhi.server.tasks import fetch_test_cases_task, tag_update_builds_task, work_on_bugs_task
 from bodhi.server.util import (
     avatar as get_avatar, build_evr, get_critpath_components,
     get_rpm_header, header, tokenize, pagure_api_get)
@@ -1130,7 +1130,7 @@ class Package(Base):
         Args:
             db (sqlalchemy.orm.session.Session): A database session.
         Raises:
-            BodhiException: When retrieving testcases from Wiki failed.
+            ExternalCallException: When retrieving testcases from Wiki failed.
         """
         if not config.get('query_wiki_test_cases'):
             return
@@ -1148,7 +1148,7 @@ class Package(Base):
             try:
                 response = wiki.call(query)
             except URLError:
-                raise BodhiException('Failed retrieving testcases from Wiki')
+                raise ExternalCallException('Failed retrieving testcases from Wiki')
             members = [entry['title'] for entry in
                        response.get('query', {}).get('categorymembers', {})
                        if 'title' in entry]
@@ -2451,22 +2451,36 @@ class Update(Base):
 
         up.date_modified = datetime.utcnow()
 
-        # Store the update alias so Celery doesn't have to emit SQL
-        update_alias = up.alias
-
         notifications.publish(update_schemas.UpdateEditV1.from_dict(
             message={'update': up, 'agent': request.user.name, 'new_bugs': new_bugs}))
 
-        # Commit the changes in the db before calling a celery task.
+        # If editing a Pending update, all of whose builds are signed, for a release
+        # which isn't composed by Bodhi (i.e. Rawhide), move it directly to Testing.
+        if not up.release.composed_by_bodhi and up.status == UpdateStatus.pending \
+                and up.signed:
+            log.info("Every build in the update is signed, set status to testing")
+            up.status = UpdateStatus.testing
+            up.date_testing = func.current_timestamp()
+            up.request = None
+            log.info(f"Update status of {up.alias} has been set to testing")
+
+        if config['test_gating.required']:
+            log.info(f"Updating test gating status of {up.alias}")
+            up.update_test_gating_status()
+
+        # Commit the changes in the db before calling celery tasks.
         db.commit()
 
-        handle_update.delay(
-            api_version=2, action='edit',
-            update_alias=update_alias,
-            agent=request.user.name,
-            new_bugs=new_bugs
-        )
+        log.info("Deferring working on bugs and fetching test cases to celery")
+        alias = up.alias
+        if not bool(config.get('bodhi_email')):
+            log.warning("Not configured to handle bugs")
+        else:
+            work_on_bugs_task.delay(alias, new_bugs)
 
+        fetch_test_cases_task.delay(alias)
+
+        log.info(f"Done editing {up.alias}")
         return up, caveats
 
     @property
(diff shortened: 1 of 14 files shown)


bodhi/server/models.py
```diff
@@ -2451,22 +2451,31 @@ def edit(cls, request, data):
...
         db.commit()
 
-        handle_update.delay(
-            api_version=2, action='edit',
-            update_alias=update_alias,
-            agent=request.user.name,
-            new_bugs=new_bugs
-        )
+        log.info("Deferring working on bugs and fetching test cases to celery")
+        work_on_bugs.delay(up.alias, new_bugs)
```
cverna: We had some weird bugs with celery serializing the objects when passing directly up.alias to the tasks. I ll recommend doing something like 
```suggestion
       update_alias = up.alias
        work_on_bugs.delay(update_alias, new_bugs)
```

bodhi/server/models.py
```diff
@@ -2451,22 +2451,31 @@ def edit(cls, request, data):
...
+        fetch_test_cases.delay(up.alias)
```
cverna: same here 
```suggestion
        fetch_test_cases.delay(update_alias)
```

bodhi/server/tasks/__init__.py
```diff
@@ -69,23 +69,6 @@ def compose(api_version: int, **kwargs):
     composer.run(api_version=api_version, data=kwargs)
 
 
-@app.task(name="handle_update")
-def handle_update(api_version: int, **kwargs):
```
cverna: We might still have tasks calling this function waiting in the queue when we update to the new release of bodhi, I am not sure if we want to keep this around even if we don't do anything in this function.

bodhi/server/tasks/fetch_test_cases.py
```diff
@@ -0,0 +1,49 @@
...
+import logging
+
+from bodhi.server.exceptions import BodhiException
+from bodhi.server.util import transactional_session_maker
+
+
+log = logging.getLogger(__name__)
+
+
+def main(alias: str):
```
cverna, line 29: This makes this tasks much smaller and nicer :+1:

bodhi/server/tasks/work_on_bugs.py
```diff
@@ -0,0 +1,92 @@
...
+
+    If handle_bugs is not True, return and do nothing.
+
+    Args:
+        update: The update that the bugs are associated with.
+        bugs: A list of bodhi.server.models.Bug instances that we wish to act on.
+    """
+    from bodhi.server.models import Bug, Update, UpdateType
+
+    handle_bugs = bool(config.get('bodhi_email'))
```
cverna: Could we check this before triggering the tasks ? It sounds not really useful to do that in the tasks, what do you think ?

bodhi/server/tasks/work_on_bugs.py
```diff
@@ -0,0 +1,92 @@
...
+    if not handle_bugs:
+        log.warning("Not configured to handle bugs")
+        return
+
+    log.info("Got %i bugs to sync for %r" % (len(bugs), alias))
```
cverna: we can use f-strings :smile:

bodhi/server/tasks/work_on_bugs.py
```diff
@@ -0,0 +1,92 @@
...
+        for bug_id in bugs:
+            bug = Bug.get(bug_id)
+            # Sanity check
+            if bug is None or bug not in update.bugs:
+                update_bugs_ids = [b.bug_id for b in update.bugs]
+                update.update_bugs(update_bugs_ids + [bug_id], session)
+                # Now, after update.update_bugs, bug with bug_id should exists in DB
+                bug = Bug.get(bug_id)
+
+            log.info("Getting RHBZ bug %r" % bug.bug_id)
```
cverna: let's use f-string here too

bodhi/server/tasks/work_on_bugs.py
```diff
@@ -0,0 +1,92 @@
...
+            try:
+                rhbz_bug = bug_module.bugtracker.getbug(bug.bug_id)
+
+                log.info("Updating our details for %r" % bug.bug_id)
```
cverna: let's use f-string here too

bodhi/server/tasks/work_on_bugs.py
```diff
@@ -0,0 +1,92 @@
...
+                bug.update_details(rhbz_bug)
+                log.info("  Got title %r for %r" % (bug.title, bug.bug_id))
```
cverna: let's use f-string here too

bodhi/server/tasks/work_on_bugs.py
```diff
@@ -0,0 +1,92 @@
...
+
+                # If you set the type of your update to 'enhancement' but you
+                # attach a security bug, we automatically change the type of your
+                # update to 'security'. We need to do this first, so we don't
+                # accidentally comment on stuff that we shouldn't.
+                if not update.type == UpdateType.security and bug.security:
+                    log.info("Setting our UpdateType to security.")
+                    update.type = UpdateType.security
+
+                log.info("Commenting on %r" % bug.bug_id)
```
cverna: let's use f-string here too

bodhi/server/tasks/work_on_bugs.py
```diff
@@ -0,0 +1,92 @@
...
+                comment = config['initial_bug_msg'] % (
+                    update.alias, update.release.long_name, update.abs_url())
+
+                log.info("Modifying %r" % bug.bug_id)
```
cverna: let's use f-string here too

bodhi/tests/server/tasks/test_fetch_test_cases.py
```diff
@@ -0,0 +1,78 @@
...
+    @patch("bodhi.server.tasks.fetch_test_cases.main")
+    def test_task(self, main_function, config_mock, init_db_mock, buildsys):
+        fetch_test_cases_task('foo')
+        config_mock.load_config.assert_called_with()
+        init_db_mock.assert_called_with(config_mock)
+        buildsys.setup_buildsystem.assert_called_with(config_mock)
+        main_function.assert_called_with('foo')
+
+
+class TestFetchTestCases(BaseTaskTestCase):
```
cverna, line 50: Should we check that the `fetch_test_cases` method is called when everything is fine ?

bodhi/server/tasks/__init__.py
mattiaverga: I have added back the task and tests.

bodhi/server/tasks/__init__.py
```diff
@@ -143,3 +144,23 @@ def tag_update_builds_task(tag: str, builds: typing.List[str]):
     log.info("Received an order to tag builds for an update")
     _do_init()
     main(tag, builds)
+
+
+@app.task(name="work_on_bugs", autoretry_for=(ExternalCallException,),
```
mattiaverga: I had to add this `ExternalCallException` because I think we don't want to keep retrying the task if the update is not found in the database and a generic BodhiException is raised.

cverna (approved):

cverna:
FYI I have deployed this on the staging instance

mattiaverga:
I still haven't got the occasion to try the new tasks in staging because Koji staging has been down for a few days.

cverna:
> I still haven't got the occasion to try the new tasks in staging because Koji staging has been down for a few days.

Yeah I am not sure what is happening there. I have given a try to the new tasks in my vagrant box tho and the tasks triggered correctly.

mattiaverga:
Now that koji staging is working again, I see some rawhide updates are getting stuck in pending.
The builds never goes after pending-signing tag...

PR #3992, opened 2020-04-08
side tag update composed by bodhi needs the candidate tag.

cverna:
During the compose builds that don't have the candidate tag
are ejected from the compose. So we need to apply the candidate tag
on the builds coming from the side tag.

### DIFF
--- a/bodhi/server/services/updates.py
+++ b/bodhi/server/services/updates.py
@@ -593,10 +593,13 @@ def new_update(request):
                         # After the Bodhi activation point of a release, add the pending-signing tag
                         # of the release to funnel the builds back into a normal workflow for a
                         # stable release.
+                        pending_signing_tag = u.release.pending_signing_tag
+                        candidate_tag = u.release.candidate_tag
                         handle_side_and_related_tags_task.delay(
                             builds=builds,
-                            pending_signing_tag=u.release.pending_signing_tag,
-                            from_tag=from_tag)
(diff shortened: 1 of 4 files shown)


bodhi/server/services/updates.py
```diff
@@ -593,10 +593,13 @@ def new_update(request):
                         # After the Bodhi activation point of a release, add the pending-signing tag
                         # of the release to funnel the builds back into a normal workflow for a
                         # stable release.
+                        pending_signing_tag = u.release.pending_signing_tag
+                        candidate_tag = u.release.candidate_tag
                         handle_side_and_related_tags_task.delay(
```
pypingou, line 598: I'm a bit wondering if we shouldn't simply have two arguments: ``tags_to_add`` and ``tags_to_remove`` both accepting lists, it may make it easier to understand the task's code itself and then the documentation about which tags to add or remove is moved to the place where the task is created.

This can be another PR though

pypingou (approved):
After discussing this on IRC, I think this is the right approach

bodhi/server/services/updates.py
cverna: yeah I thought about it, but since we check which tag is None in the tasks to know what to do (ie create or remove tag) it is easier to have the tags being explicit.

We could tho have a boolean to specify which case we want, or maybe a different celery task. 1 for rawhide and 1 for non rawhide.

PR #3995, opened 2020-04-09
Ignore celery results we don't use.

cverna:
It is recommended in the celery documentation to ignore tasks results
that are not used. Currently the only task's results used are for
the tasks triggered by celery-beat, the other task's results then can
be ignored.

In production it was also observed that some tasks hanged and never
returned making the web request hang too, which lead to having database
session left open.

This commit prevents such a scenario since triggering a task with no results
will return immediately.

### DIFF
--- a/bodhi/server/tasks/__init__.py
+++ b/bodhi/server/tasks/__init__.py
@@ -49,7 +49,7 @@ def _do_init():
     bugs.set_bugtracker()
 
 
-@app.task(name="compose")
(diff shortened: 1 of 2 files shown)


pypingou (approved):
Nice and easy :)

cverna:
> Nice and easy :)

Apparently not that easy :-), we seems to be using the results in the integration tests :-(

pypingou:
Oh well, you know my opinion about these :)

PR #3996, opened 2020-04-11
[WIP] Revise the number of messages/comments sent by 'approve_testing.py'

### LINKED ISSUE #3846: When update is autopushed, don't send both 'update can now be pushed' and 'update has been pushed' notifications
When an update has autopush-for-time enabled, at the point the time threshold for pushing is reached, Bodhi adds *two* comments and sends *two* emails basically simultaneously: the first to say the update can now be pushed to stable, and the second to say it has been pushed stable.

This is entirely unnecessary. If autopush by time is disabled it makes sense to add a comment/send a mail saying the update "can now be pushed stable", but it makes no sense to send two mails if autopush is enabled. I don't really care in detail how it's implemented, but in this situation, Bodhi should add *one* comment and send *one* email to describe the situation (the update reached its autopush threshold and was pushed automatically).

richardgreg:
I added a conditional statement so 'approve_testing' doesn't send
an unnecassary comment/message saying 'the update can now be
pushed to stable if maintainer wishes' even though 'autotime' is set.

Fixes: #3846

### DIFF
--- a/bodhi/server/tasks/approve_testing.py
+++ b/bodhi/server/tasks/approve_testing.py
@@ -78,14 +78,17 @@ def approve_update(update: Update, db: Session):
     log.info(f'{update.alias} now meets testing requirements')
     # Only send email notification about the update reaching
     # testing approval on releases composed by bodhi
-    update.comment(
-        db,
-        str(config.get('testing_approval_msg')),
-        author='bodhi',
-        email_notification=update.release.composed_by_bodhi
-    )
-    notifications.publish(update_schemas.UpdateRequirementsMetStableV1.from_dict(
-        dict(update=update)))
+    if not update.autotime or update.meets_testing_requirements:
+        update.comment(
+            db,
+            str(config.get('testing_approval_msg')),
+            author='bodhi',
+            email_notification=update.release.composed_by_bodhi
+        )
+        notifications.publish(update_schemas.UpdateRequirementsMetStableV1.from_dict(
+            dict(update=update)))
+    # Only post a comment informing the maintainer that the update can now be pushed
+    # to stable if autotime is not set and/or update has met testing requirements.
     if update.autotime and update.days_in_testing >= update.stable_days:
         log.info(f"Automatically marking {update.alias} as stable")
         # For now only rawhide update can be created using side tag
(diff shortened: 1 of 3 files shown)


richardgreg:
@cverna I'd like some feedback.

bodhi/server/tasks/approve_testing.py
```diff
@@ -78,14 +78,18 @@ def approve_update(update: Update, db: Session):
...
     # testing approval on releases composed by bodhi
-    update.comment(
-        db,
-        str(config.get('testing_approval_msg')),
-        author='bodhi',
-        email_notification=update.release.composed_by_bodhi
-    )
-    notifications.publish(update_schemas.UpdateRequirementsMetStableV1.from_dict(
-        dict(update=update)))
+    if not update.autotime or update.karma >= update.stable_karma:
```
cverna: So I am not a big fan of having the check for karma here since this adding yet another place where we check for it. Ideally we should only have one place where we check if the update meet the requirements the `meets_testing_requirements` method. 

Maybe we could add the comment in that method instead of here, but we need to check where this method is called then.

richardgreg: Right. The comment is already in `meets_testing_requirements`

bodhi/server/tasks/approve_testing.py
```diff
@@ -78,14 +78,17 @@ def approve_update(update: Update, db: Session):
...
     # testing approval on releases composed by bodhi
-    update.comment(
-        db,
-        str(config.get('testing_approval_msg')),
-        author='bodhi',
-        email_notification=update.release.composed_by_bodhi
-    )
-    notifications.publish(update_schemas.UpdateRequirementsMetStableV1.from_dict(
-        dict(update=update)))
+    if not update.autotime or update.meets_testing_requirements:
```
cverna, line 81: I feel that we could probably simplify the logic here. In line 76 we check if update.meets_testing_requirements is False and then return, maybe we could check if it is True instead and execute the code below. What do you think ?

richardgreg: It seems if we do that `test_autotime_update_zero_day_in_testing_fail_gating_is_not_pushed` fails.

richardgreg:
Retiring this PR in case someone else wants to have a go at it :)

PR #3997, opened 2020-04-13
Add url_name to User class to provide safe URLs

### LINKED ISSUE #3993: Bad characters in username
There's a bot with a bad username of `packagerbot/os-master01.phx2.fedoraproject.org` that makes CI tests failing.
https://bodhi.fedoraproject.org/users/packagerbot/os-master01.phx2.fedoraproject.org

I'm pushing a PR to safe check CI tests, but do we want to make Bodhi safe to bad usernames like this? Since usernames are from outside world, should we modify them in a safe way before storing in the database?

mattiaverga:
Since some usernames contain illegal characters to be used in URLs, we will use this sanitized internal username. The original username is preserved and it's still displayed to the frontend user.

I've also removed the previous workaround to get CI tests working, since this change should prevent those errors.

Fixes #3993 

### DIFF
--- a/bodhi/server/migrations/versions/4cf90f327ed6_add_a_url_name_to_user.py
+++ b/bodhi/server/migrations/versions/4cf90f327ed6_add_a_url_name_to_user.py
@@ -1,0 +1,49 @@
+# Copyright (c) 2020 Mattia Verga
+#
+# This file is part of Bodhi.
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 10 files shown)


mattiaverga:
Well, I've just realized that I don't like this solution at all... too much overcomplication for something that should really not happen. Since real usernames are from FAS and they should be compatible with openID, they cannot contain characters like `/` or `.`.
I think we should ask to change that bot `packagerbot/os-master01.phx2.fedoraproject.org` username to something that follows the standard.

cverna:
@AdamSaleh @pypingou any ideas there ? I am not sure renaming is possible in FAS even if we wanted to. Otherwise I think we can just accept that this user is broken :(

pypingou:
The issue is that this is not a problem in FAS, that username is coming from the kerberos keytab generated for that bot (whose username is simply: ``packagerbot``). Maybe someone in our infra would know how to give that keytab a prettier username, but I don't know that this is possible with our current keytab role.

PR #3998, opened 2020-04-14
docs: Reference the state that happens when an update is revoked

### LINKED ISSUE #2902: Inexisting update state(Revoked) in docs
https://bodhi.fedoraproject.org/docs/user/update_states.html#revoked

As far as I know "revoke" is only action.
So, can we remove it from docs, or the following line:
https://github.com/fedora-infra/bodhi/blob/develop/bodhi/server/models.py#L2587
should be changed?

richardgreg:
Fixes: #2902


cverna (approved):
Thanks

PR #3999, opened 2020-04-15
Use 'exact' search for koji

tkopecek:
NVR is passed to koji search page. There is no reason to expect, that
this NVR is a glob. Koji would be happier if it can use "exact" search
instead of "glob".

### DIFF
--- a/bodhi/server/templates/update.html
+++ b/bodhi/server/templates/update.html
@@ -427,7 +427,7 @@ if can_edit and update.release.composed_by_bodhi:
               % for build in update.builds:
               <%
               koji_web_url = request.registry.settings.get('koji_web_url').strip('/') + '/'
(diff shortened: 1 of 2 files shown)


mattiaverga:
It seems reasonable to me.

Can you take care of the failing tests? There are three tests that expects the `glob` parameter to be present in the HTML output.

tkopecek:
Test fixed :-)

cverna:
@tkopecek thanks :-) one last little thing, could you please squash the 2 commits into one and use the "Signed-off-by" tag in that commit.
This is explained in more details in the contribution guidelines --> https://bodhi.fedoraproject.org/docs/developer/index.html#contribution-guidelines

mattiaverga (approved):

PR #4004, opened 2020-04-20
Ensure we don't use a critpath update in test

mattiaverga:

### DIFF
--- a/bodhi/server/buildsys.py
+++ b/bodhi/server/buildsys.py
@@ -387,7 +387,7 @@ class DevBuildsys:
                  'name': 'f28F'},
             ]
         else:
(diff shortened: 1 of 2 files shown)


cverna (approved):

mattiaverga:
Well, now we have 
`Cannot find release associated with build: hugo-0.69.0-2.fc32~bootstrap, tags: ['f32~bootstrap-updates-candidate', 'f32~bootstrap', 'f32~bootstrap-updates-testing']`

It seems Go packaging macros create that horrible dist tag: https://docs.fedoraproject.org/en-US/packaging-guidelines/#bootstrapping
And they call that a "nice side-effect"... however, it seems those updates flow regularly in Bodhi, not sure why it fails in CI tests.

PR #4005, opened 2020-04-20
Modify default bz_regex to support RH standards

mattiaverga:
Follows discussion in #3956 


cverna (approved):

voxik:
I just wonder, is the RegExp case sensitive? I hope it is not. Sorry for being late :angel:

mattiaverga:
> I just wonder, is the RegExp case sensitive? I hope it is not. Sorry for being late angel

Nope, case insensitive.
I made it that way in the first PR, so ti works with `- Fix #34452 and also fix #66778`.

Hopefully, no one will put in the changelog something like `- Doesn't fix #12345` :stuck_out_tongue_winking_eye:

voxik:
Thanks for confirmation.

PR #4006, opened 2020-04-20
docs: Add 'frozen' state to Bodhi documentation

### LINKED ISSUE #3505: Add a banner during Freezes
Just before every release, I see people wondering why their packages are not pushed to stable. If Bodhi showed a message _somewhere_, this would probably be less common.

richardgreg:
Fixes: #3505

### DIFF
--- a/docs/user/update_states.rst
+++ b/docs/user/update_states.rst
@@ -10,6 +10,8 @@ Once submitted to Bodhi, updates move through the following states:
 
 :ref:`testing-stable`: The package is ready to go stable, and will wait until the next update push.
 
+:ref:`frozen`: The package is held back in the testing repository for additional checks.
+
 :ref:`stable`: The package has been released to the main updates repository.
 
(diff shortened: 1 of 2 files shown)


richardgreg:
Reading through #3505 it seems the 'frozen banner' issue is solved by Sebastian. All that's remaining is for `frozen` to be added to docs.

I think it wise for someone with a more intimate knowledge of Bodhi to provide all the details of what's really going on in the frozen state so it can go in the `Frozen` paragraph.

richardgreg:
@cverna what do you think?

cverna:
> @cverna what do you think?
Looks good, Frozen is when we stabilize a release before the release Beta or GA (Generaly Available), in such state all updates are blocked and release engineering will only push the updates that have been given a freeze break exception to fix a bug.

If you can make something from that to add in the Frozen description that would be great :-)

cverna (approved):
Thanks

PR #4009, opened 2020-04-22
WebUI: Make Test Cases look clickable.

### LINKED ISSUE #4003: Test Cases dont look like they are clickable
It's hard to make out if the test cases are cilckable and should point to wiki.
https://bodhi.fedoraproject.org/updates/FEDORA-2020-c014bea351

User should be able to clearly understand that the test cases are clickable, the colour of text for link to be changed maybe

richardgreg:
Fixes: #4003

### DIFF
--- a/bodhi/server/util.py
+++ b/bodhi/server/util.py
@@ -675,7 +675,7 @@ def testcase_link(context, test, short=False):
     """
     url = config.get('test_case_base_url') + test.name
     display = test.name.replace('QA:Testcase ', '')
-    link = "<a target='_blank' href='%s' class='notblue'>%s</a>" % (url, display)
+    link = f"<a target='_blank' href='{url}' class='font-weight-bolder'>{display}</a>"
     if not short:
         link = "Test Case " + link
     return link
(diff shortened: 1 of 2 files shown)


richardgreg:
The original author intentionally made it 'not blue'. I wonder why though?

mattiaverga (changes requested):
> The original author intentionally made it 'not blue'. I wonder why though?

I suppose that's because light blue text + small font size + white background is not the best for readability... I would suggest to change the `notblue` class with `font-weight-bolder` to enhance the readability.

Also, let's use f-strings while editing the text.

news/4003.dev
```diff
@@ -0,0 +1 @@
+Make Test Cases look clickable.
```
mattiaverga: Ah, sorry, I didn't notice in my previous review the news fragment type: this isn't a development related change, a filename suffix of `.bug` should be fine.

mattiaverga (changes requested):

cverna (approved):
LGTM

mattiaverga (approved):

PR #4010, opened 2020-04-23
Remove url column from Bug model

### LINKED ISSUE #1995: bodhi.server.models.Bug.url is inaccessible
The Bug model has a url database column, but the column cannot be accessed via the model because the model defines a [```url``` property](https://github.com/fedora-infra/bodhi/blob/3.0.0/bodhi/server/models.py#L2740-L2742) that overrides it.

We should ensure that this column really cannot be reached, and if so we should just drop it since it is unused.

### LINKED ISSUE #4008: Docs build failures with Sphinx 3
bodhi fails to build from source on Fedora with Sphinx 3 series.

Downstream bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=1823739

Error:
Warning, treated as error:
/builddir/build/BUILD/bodhi-5.1.1/bodhi/server/models.py:docstring of bodhi.server.models.Bug.url:1:duplicate object description of bodhi.server.models.Bug.url, other instance in developer/docblocks/bodhi.server.models, use :noindex: for one of them
make: *** [Makefile:51: html] Error 2
make: Leaving directory '/builddir/build/BUILD/bodhi-5.1.1/docs'
error: Bad exit status from /var/tmp/rpm-tmp.A6849v (%build)

Sphinx 3 changelog: https://www.sphinx-doc.org/en/master/changes.html#release-3-0-0-released-apr-06-2020

mattiaverga:
It seems that the Bug model still reference `url` as a column in the database, but that has already been removed:
```
>>> query = "SELECT * FROM bugs WHERE TRUE LIMIT 1"
>>> with conn:
...     with conn.cursor() as curs:
...             curs.execute(query)
...             row = curs.fetchone()
... 
>>> row
(52767, 1670508, 'Review Request: ckb-next - driver for Corsair RGB keyboards and mice', False, False)
```
(fields are id, bug_id, title, security, parent)

So I think we can safely remove that reference from the model without any database migration (trying to drop the `url` column results in an error, I've double checked).

Fixes #1995
an possibly fixes #4008 

### DIFF
--- a/bodhi/server/models.py
+++ b/bodhi/server/models.py
@@ -4377,8 +4377,6 @@ class Bug(Base):
         bug_id (int): The bug's id.
         title (str): The description of the bug.
         security (bool): True if the bug is marked as a security issue.
-        url (str): The URL for the bug. Inaccessible due to being overridden by the url
-            property (https://github.com/fedora-infra/bodhi/issues/1995).
         parent (bool): True if this is a parent tracker bug for release-specific bugs.
     """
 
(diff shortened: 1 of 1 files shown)


cverna (approved):
Thanks for looking at this

PR #4013, opened 2020-04-30
If an update include no builds, use alias as title

### LINKED ISSUE #4012: no clickable link for updates with no builds
If an update include no builds, `get_title` will return an empty string and there will be no clickable title in UI.

mattiaverga:
Fix #4012 

### DIFF
--- a/bodhi/server/models.py
+++ b/bodhi/server/models.py
@@ -2659,9 +2659,13 @@ class Update(Base):
                 title += " more"
 
                 return title
+            elif len(self.builds) == 0:
+                return self.alias
             else:
                 return " and ".join([build_label(build) for build in self.builds])
         else:
+            if len(self.builds) == 0:
+                return self.alias
             all_nvrs = [x.nvr for x in self.builds]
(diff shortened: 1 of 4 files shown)


bodhi/server/models.py
```diff
@@ -2659,9 +2659,13 @@ def build_label(build):
                 title += " more"
 
                 return title
+            elif len(self.builds) == 0:
```
cverna, line 2676: Since when there is no builds the return value is the same of beautify or not beautify, we could check if `self.builds == 0` right before the `if beautify` and return the alias, that would let us avoid duplicating that code.

mattiaverga: Not exactly: if `self.display_name != ''` it will take precedence over `self.builds == 0`, but only when `beautify`... thus I couldn't put the `if` before checking `beautify` value.

cverna: ah yeah that allows packager to put a custom name to the update.

cverna (approved):

PR #4017, opened 2020-05-09
Improve Json API response times

### LINKED ISSUE #271: /updates/?packages queries timing out
```
http get https://bodhi.fedoraproject.org/updates/\?packages\=gnome-shell
```

The query is causing a ton of CPU churn it seems, and never returns.

This is a re-post of fedora-infra/fedmsg_meta_fedora_infrastructure#314, which was filed there accidentally.

### LINKED ISSUE #301: Investigate performance issues

### LINKED ISSUE #1899: /updates 504 Gateway Timeouts
Queries for Fedora 24 through Fedora 26 security/stable package updates are hitting 504 a lot.

Example:
https://bodhi.fedoraproject.org/updates/?release=F26&status=stable&type=security&rows_per_page=1000&submitted_before=2017-10-14&submitted_since=2015-01-01

The above query works just fine if you reduce the rows_per_page to 200. But that's not reasonable.
https://bodhi.fedoraproject.org/updates/?release=F26&status=stable&type=security&rows_per_page=1000&submitted_before=2017-10-14&submitted_since=2015-01-01

Shouldn't there be a response even while it's trying to assemble results? Even 350 breaks it. If there's a better way to pull all (not some) of the updates, I'd love to hear it.

### LINKED ISSUE #3204: [L] - Design new API response schemas to improve performance
I spent some time performance analysing Bodhi's queries yesterday, and I discovered that Bodhi spends an enormous amount of time serializing the results of the query. For example, ``` $ bodhi updates query --status stable ``` takes about 5 seconds, but its database query only accounts for about 0.1 seconds. More than 4 seconds are spent serializing the data.

The problem is that we just hand the results of the query as database objects back to Pyramid, which then calls ```__json___()``` on our models, which ends up serializing every field on the matched objects. Some of those fields are foreign objects, and some fields on those foreign objects are foreign objects, so this ends up loading a bunch of data in.

The fix is to rethink the API. We should reduce the number of fields we serializing in our responses. It may be good to allow the user to specify which fields they would like to retrieve.

mattiaverga:
The changes here aim to improve Bodhi's JSON API response times.
At database/sqlalchemy level, I've changed some problematic relationships load modes. Some were set to be `lazy='joined'`, but that caused to load very large amounts of data from the db.
In some cases I've set the relationship to use inner joins (where the referenced object id was not nullable).

The PR include a database change to set user_id in comments table to be not nullable (I've checked the db and all comments are correctly linked to a user).
There could be some other places where this seems possible: for example all comments should be linked to an update... but there are some in the db for which this isn't true (don't know why).

At object serialization level, I've added a `__relation_fields__` property to BodhiBase object along with relevant object. The scope is to select what attributes should be serialized when the object is loaded as relation inside another object.

# Performance tests
I've made some dummy tests to measure performance improvements of this change.
## Query times
The code used to conduct this test is (from bodhi-shell):
```
def query_time(query):
    t0=time.time()
    query.all()
    print(time.time() - t0)

models = (m.Build, m.Comment, m.BuildrootOverride, m.Package, m.Release, m.Update, m.User)
for model in models:
    query = s.query(model)
    query = query.limit(1000)
    query_time(query)
```
The results are:
| Class | Before changes | After changes | Improvement |
|---|---|---|---|
| Build | 1,51109004020691 | 0,182040214538574 | -87,95% |
| Comment | 0,390185356140137 | 0,038461446762085 | -90,14% |
| BuildrootOverride | 1,15033030509949 | 0,199055910110474 | -82,70% |
| Package | 0,037990093231201 | 0,036218881607056 | -4,66% |
| Release | 0,005536556243896 | 0,005467653274536 | -1,24% |
| Update | 1,27048087120056 | 0,099624633789063 | -92,16% |
| User | 8,3333420753479 | 0,02573299407959 | -99,69% |

## JSON lists
| URL | Before changes | After changes | Improvement |
|---|---|---|---|
| http://localhost:6543/builds/?rows_per_page=1000 | 13s | 4,6s | -64,62% |
| http://localhost:6543/comments/?rows_per_page=1000 | 55s | 31s | -43,64% |
| http://localhost:6543/overrides/?rows_per_page=1000 | 6,1s | 0,6s | -90,16% |
| http://localhost:6543/packages/?rows_per_page=1000 | 186ms | 170ms | -8,60% |
| http://localhost:6543/releases/?rows_per_page=1000 | 75ms | 103ms | +37,33% |
| http://localhost:6543/updates/?rows_per_page=1000 | 136s | 26s | -80,88% |
| http://localhost:6543/users/?rows_per_page=1000 | 11s | 2,5s | -77,27% |

Because the JSON schema is changed, some tools that depends on this can break. So, I've tagged the PR as backwards incompatible.

Some longstanding issues that should be fixed by this:
fix #271
fix #301 
fix #1899
fix #3204 

### DIFF
--- a/bodhi/server/models.py
+++ b/bodhi/server/models.py
@@ -300,6 +300,8 @@ class BodhiBase(object):
         __exclude_columns__ (tuple): A list of columns to exclude from JSON
         __include_extras__ (tuple): A list of methods or attrs to include in JSON
         __get_by__ (tuple): A list of columns that :meth:`.get` will query.
+        __relation_fields__ (tuple): A list of columns to include in JSON when
+            the object is referenced as relation inside another object.
         id (int): An integer id that serves as the default primary key.
         query (sqlalchemy.orm.query.Query): a class property which produces a
             Query object against the class and the current Session when called.
@@ -308,6 +310,7 @@ class BodhiBase(object):
     __exclude_columns__ = ('id',)
     __include_extras__ = tuple()
     __get_by__ = ()
+    __relation_fields__ = ('id',)
 
     id = Column(Integer, primary_key=True)
 
@@ -366,7 +369,8 @@ class BodhiBase(object):
         return self._to_json(self, request=request, exclude=exclude, include=include)
 
     @classmethod
-    def _to_json(cls, obj, seen=None, request=None, exclude=None, include=None):
+    def _to_json(cls, obj, seen=None, request=None, exclude=None, include=None,
+                 relation_fields=None):
         """
         Return a JSON representation of obj.
 
@@ -381,6 +385,9 @@ class BodhiBase(object):
             include (iterable or None): An iterable of strings naming the extra attributes to
                 include in the JSON representation of the model. If None (the default), the class's
                 __include_extras__ attribute will be used.
+            __relation_fields__ (iterable or None): A list of columns to include in JSON when
+                the object is referenced as relation inside another object. If None (the default),
+                the class's __relation_fields__ attribute will be used.
         Returns:
             dict: A dict representation of the model suitable for serialization.
         """
@@ -391,28 +398,42 @@ class BodhiBase(object):
 
         if exclude is None:
             exclude = getattr(obj, '__exclude_columns__', [])
+        if include is None:
+            include = getattr(obj, '__include_extras__', [])
         properties = list(class_mapper(type(obj)).iterate_properties)
         rels = [p.key for p in properties if isinstance(p, RelationshipProperty)]
         attrs = [p.key for p in properties if p.key not in rels]
-        d = dict([(attr, getattr(obj, attr)) for attr in attrs
-                  if attr not in exclude and not attr.startswith('_')])
-
-        if include is None:
-            include = getattr(obj, '__include_extras__', [])
-
-        for name in include:
-            attribute = getattr(obj, name)
-            if callable(attribute):
-                attribute = attribute(request)
-            d[name] = attribute
-
-        for attr in rels:
-            if attr in exclude:
-                continue
-            target = getattr(type(obj), attr).property.mapper.class_
-            if target in seen:
-                continue
-            d[attr] = cls._expand(obj, getattr(obj, attr), seen, request)
+        if not relation_fields:
+            d = dict([(attr, getattr(obj, attr)) for attr in attrs
+                     if attr not in exclude and not attr.startswith('_')])
+            for name in include:
+                attribute = getattr(obj, name)
+                if callable(attribute):
+                    attribute = attribute(request)
+                d[name] = attribute
+            for attr in rels:
+                if attr in exclude:
+                    continue
+                target = getattr(type(obj), attr).property.mapper.class_
+                if target in seen:
+                    continue
+                d[attr] = cls._expand(obj, getattr(obj, attr), seen, request)
+        else:
+            d = dict([(attr, getattr(obj, attr)) for attr in attrs if attr in relation_fields])
+            for name in include:
+                if name not in relation_fields:
+                    continue
+                attribute = getattr(obj, name)
+                if callable(attribute):
+                    attribute = attribute(request)
+                d[name] = attribute
+            for attr in rels:
+                if attr not in relation_fields:
+                    continue
+                target = getattr(type(obj), attr).property.mapper.class_
+                if target in seen:
+                    continue
+                d[attr] = cls._expand(obj, getattr(obj, attr), seen, request)
 
         for key, value in d.items():
             if isinstance(value, datetime):
@@ -440,7 +461,9 @@ class BodhiBase(object):
         if hasattr(relation, '__iter__'):
             return [cls._expand(obj, item, seen, req) for item in relation]
         if type(relation) not in seen:
-            return cls._to_json(relation, seen + [type(obj)], req)
+            relation_fields = getattr(relation, '__relation_fields__', [])
+            return cls._to_json(relation, seen + [type(obj)], req,
+                                relation_fields=relation_fields)
         else:
             return relation.id
 
@@ -799,6 +822,7 @@ class Release(Base):
     __tablename__ = 'releases'
     __exclude_columns__ = ('id', 'builds')
     __get_by__ = ('name', 'long_name', 'dist_tag')
+    __relation_fields__ = ('id', 'name', 'long_name')
 
     name = Column(Unicode(10), unique=True, nullable=False)
     long_name = Column(Unicode(25), unique=True, nullable=False)
@@ -824,6 +848,10 @@ class Release(Base):
 
     package_manager = Column(PackageManager.db_type(), default=PackageManager.unspecified)
     testing_repository = Column(UnicodeText, nullable=True)
+
+    # One-to-many relationships
+    # builds backref from Build
+    # composes backref from Compose
 
     @property
     def critpath_min_karma(self) -> int:
@@ -1024,11 +1052,17 @@ class TestCase(Base):
 
     __tablename__ = 'testcases'
     __get_by__ = ('name',)
+    __exclude_columns__ = ('id', 'package', 'feedback',)
+    __relation_fields__ = ('id', 'name',)
 
     name = Column(UnicodeText, nullable=False)
 
+    # Many-to-one relationships
     package_id = Column(Integer, ForeignKey('packages.id'))
-    # package backref
+    # package backref from Package
+
+    # One-to-many relationships
+    # feedback backref from TestCaseKarma
 
 
 class Package(Base):
@@ -1051,12 +1085,14 @@ class Package(Base):
     __tablename__ = 'packages'
     __get_by__ = ('name',)
     __exclude_columns__ = ('id', 'test_cases', 'builds',)
+    __relation_fields__ = ('id', 'name',)
 
     name = Column(UnicodeText, nullable=False)
     requirements = Column(UnicodeText)
     type = Column(ContentType.db_type(), nullable=False)
 
-    builds = relationship('Build', backref=backref('package', lazy='joined'))
+    # One-to-many relationships
+    builds = relationship('Build', backref=backref('package', lazy='joined', innerjoin=True))
     test_cases = relationship('TestCase', backref='package', order_by="TestCase.id")
 
     __mapper_args__ = {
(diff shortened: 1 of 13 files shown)


bodhi/server/models.py
```diff
@@ -1348,17 +1384,22 @@ class Build(Base):
     """
 
     __tablename__ = 'builds'
-    __exclude_columns__ = ('id', 'package', 'package_id', 'release',
-                           'update_id', 'update', 'override')
+    __exclude_columns__ = ('id', 'package', 'release', 'update', 'override')
     __get_by__ = ('nvr',)
+    __relation_fields__ = ('id', 'nvr', 'type')
```
cverna: I would probably add the signed field here since it is quite important info to have about a build

cverna:
Overall this is looking good, I just want to check the fedora-messaging message sent, since I think this is using the json serialization. If that's the case we need to check the messages schema and also be noisy about the changes in case someone relies on a field that we are going to remove.

bodhi/server/models.py
mattiaverga: Field added

mattiaverga:
> Overall this is looking good, I just want to check the fedora-messaging message sent, since I think this is using the json serialization. If that's the case we need to check the messages schema and also be noisy about the changes in case someone relies on a field that we are going to remove.

Well, while writing this PR I gave a look to some tools that use Bodhi Json APIs, like fedora-easy-karma and qa-stats and I think they should not be broken by this.
About message schemes, I hope the unit tests we run are reliable enough...

cverna:
Ok so this impacts the messages sent. I did a bit of checking in vagrant and for example

The bodhi.update.edit message in prod https://apps.fedoraproject.org/datagrepper/id?id=2020-b28bbe36-4715-48b7-94bc-dd70d30a64ae&is_raw=true&size=extra-large

in vagrant : 

```
Topic: bodhi.update.edit
Headers: {
    "fedora_messaging_rpm_gl2ps": true,
    "fedora_messaging_schema": "bodhi.update.edit.v1",
    "fedora_messaging_severity": 20,
    "fedora_messaging_user_cverna": true,
    "sent-at": "2020-05-25T07:34:02+00:00"
}
Body: {
    "agent": "cverna",
    "new_bugs": [],
    "update": {
        "alias": "FEDORA-2020-db7697713b",
        "autokarma": true,
        "autotime": true,
        "bugs": [],
        "builds": [
            {
                "id": 257971,
                "nvr": "gl2ps-1.4.1-6.fc33",
                "signed": false,
                "type": "rpm"
            }
        ],
        "close_bugs": true,
        "comments": [
            {
                "id": 1317535,
                "karma": 0,
                "text": "This update has been submitted for testing by cverna. ",
                "timestamp": "2020-05-20 10:28:07",
                "user": {
                    "id": 91,
                    "name": "bodhi"
                }
            },
            {
                "id": 1317536,
                "karma": 0,
                "text": "This update's test gating status has been changed to 'greenwave_failed'.",
                "timestamp": "2020-05-20 10:28:10",
                "user": {
                    "id": 91,
                    "name": "bodhi"
                }
            },
            {
                "id": 1317537,
                "karma": 0,
                "text": "This update's test gating status has been changed to 'waiting'.",
                "timestamp": "2020-05-20 10:28:10",
                "user": {
                    "id": 91,
                    "name": "bodhi"
                }
            },
            {
                "id": 1317538,
                "karma": 0,
                "text": "cverna edited this update.",
                "timestamp": "2020-05-20 10:32:12",
                "user": {
                    "id": 91,
                    "name": "bodhi"
                }
            },
            {
                "id": 1317539,
                "karma": 0,
                "text": "This update's test gating status has been changed to 'greenwave_failed'.",
                "timestamp": "2020-05-20 10:32:15",
                "user": {
                    "id": 91,
                    "name": "bodhi"
                }
            },
            {
                "id": 1317540,
                "karma": 0,
                "text": "cverna edited this update.",
                "timestamp": "2020-05-20 10:33:36",
                "user": {
                    "id": 91,
                    "name": "bodhi"
                }
            },
            {
                "id": 1317541,
                "karma": 0,
                "text": "cverna edited this update.",
                "timestamp": "2020-05-20 10:33:48",
                "user": {
                    "id": 91,
                    "name": "bodhi"
                }
            },
            {
                "id": 1317547,
                "karma": 0,
                "text": "cverna edited this update.",
                "timestamp": "2020-05-25 07:34:01",
                "user": {
                    "id": 91,
                    "name": "bodhi"
                }
            }
        ],
        "compose": null,
        "content_type": "rpm",
        "critpath": false,
        "date_approved": null,
        "date_modified": "2020-05-25 07:34:01",
        "date_pushed": null,
        "date_stable": null,
        "date_submitted": "2020-05-20 10:28:07",
        "date_testing": null,
        "display_name": "",
        "from_tag": null,
        "karma": 0,
        "locked": false,
        "meets_testing_requirements": false,
        "notes": "sada\n\n\n\n",
        "pushed": false,
        "release": {
            "id": 40,
            "long_name": "Fedora 33",
            "name": "F33"
        },
        "request": "testing",
        "require_bugs": true,
        "require_testcases": true,
        "requirements": null,
        "severity": "unspecified",
        "stable_days": 7,
        "stable_karma": 3,
        "status": "pending",
        "suggest": "unspecified",
        "test_cases": [],
        "test_gating_status": "greenwave_failed",
        "title": "gl2ps-1.4.1-6.fc33",
        "type": "unspecified",
        "unstable_karma": -3,
        "updateid": "FEDORA-2020-db7697713b",
        "url": "http://localhost:6543/updates/FEDORA-2020-db7697713b",
        "user": {
            "id": 3442,
            "name": "cverna"
        },
        "version_hash": "6801e410c07ada5b006fac30e5fd66c76e51fd68"
    }
}
```

I am not sure why the tests are still working (I have not looked at it yet), but if we want to go ahead with that change I think we need to create a v2 of the message schema

mattiaverga:
I suppose the message tests are completely unreliable: the UpdateV1 schema description is completely different from the update object produced in the UpdateEditV1 message :hushed: (even before this PR)
Am I wrong?

Should I rewrite all object schemas from scratch and mark them V2?
What to do with schemas V1? Should they be removed, or we keep both V1 and V2?
Does that mean that every change we make to a model, also the message schema of that object should be changed?
Or we care only of fields described in the schema?

cverna:
> I suppose the message tests are completely unreliable: the UpdateV1 schema description is completely different from the update object produced in the UpdateEditV1 message hushed (even before this PR)
> Am I wrong?

Yeah it looks like it is using the schema in the `expected` part of the tests so if the schema change the tests will still be valid since the expected value is dynamic :(

https://github.com/fedora-infra/bodhi/blob/develop/bodhi/tests/messages/schemas/test_update.py#L58

> 
> Should I rewrite all object schemas from scratch and mark them V2?

So it looks like all the class have the version in there name for example `UpdateEditV1` we should probably create a `UpdateEditV2` on so on but that's not really scalable

> What to do with schemas V1? Should they be removed, or we keep both V1 and V2?

I guess we should always keep the latest version -1 so in our case V1 and V2.

> Does that mean that every change we make to a model, also the message schema of that object should be changed?

> Or we care only of fields described in the schema?

I think it is all the model changes that impact the json representation of the object. The base message schema loads the json representation of the object (https://github.com/fedora-infra/bodhi/blob/develop/bodhi/messages/schemas/base.py#L88). I am not sure this is great tbh, if we are going to create a V2 it might be better to decouple this and have the message independent from the json representation of the object.

What do you think ?

ccing @abompard  also which might give us some advice from the fedora-messaging side on how to manage these schemas.

mattiaverga:
> 
> I think it is all the model changes that impact the json representation of the object. The base message schema loads the json representation of the object (https://github.com/fedora-infra/bodhi/blob/develop/bodhi/messages/schemas/base.py#L88). I am not sure this is great tbh, if we are going to create a V2 it might be better to decouple this and have the message independent from the json representation of the object.
> 
> What do you think ?
> 
Yeah, that's exactly what I was thinking about the message schemas. Till now, we have been breaking the schemas all the time when we did a change to a model.
Maybe we can merge this change, then start working on new message schemas that don't rely on the model's json representation, so that we're sure they will stay stable in future and only change when we want to change them. 
The effort with GraphQL may solve this. I don't know if it worth to work on schemas V2 still based on json serialization.

Or maybe I can try to split out the database changes to see if the performance gain is enough without touching the object serialization part.

mattiaverga:
Ok, so, I made some tests.
Only by tweaking the joins type in the models I can get the same improvement in the database queries (as I expected) and a good performance improvement in json serialization of all models but the Update model, without changing the json output schema.

The Update model is the most problematic, since it basically throws in all other models... but the main problem are the update comments. By serializing the Update model, we serialize all the comments for each update, each comment serialize a bunch of other models... that's a long waterfall that requires a lot of time.
The logic solution would be to put only a list of comment_id when quering the Update model, then if the end user wants to access comment details they can query for that. But I'm pretty sure this would break at least fedora-easy-karma.

Let me know if you want me to open a new PR with only the model join changes that don't break message schemas and postpone the json changes when we find a way to have stable schemas which don't depend on model serialization.

cverna:
> Ok, so, I made some tests.
> Only by tweaking the joins type in the models I can get the same improvement in the database queries (as I expected) and a good performance improvement in json serialization of all models but the Update model, without changing the json output schema.
> 
> The Update model is the most problematic, since it basically throws in all other models... but the main problem are the update comments. By serializing the Update model, we serialize all the comments for each update, each comment serialize a bunch of other models... that's a long waterfall that requires a lot of time.
> The logic solution would be to put only a list of comment_id when quering the Update model, then if the end user wants to access comment details they can query for that. But I'm pretty sure this would break at least fedora-easy-karma.
> 
> Let me know if you want me to open a new PR with only the model join changes that don't break message schemas and postpone the json changes when we find a way to have stable schemas which don't depend on model serialization.

+1 to do that in a phase approach with model join changes first and then the json changes + schemas

abompard:
Hey all! The idea with versionned Fedora Messaging schemas, is to have the new messages published on a different topic (`bodhi.update.v2` instead of `bodhi.update`, for example) and to have both messages sent at the same time for "a while". What "a while" is is up to you, you could decide to keep it for one or more major versions, etc. The point is to give time to consumers to update their subscription from v1 to v2, and thus handle the new format.

I would also recommend adding a deprecation warning, in the docs and in the code, when the old schema is used.

Making a new class for V2 seems like the easiest way to do that, after all it should not happen so often, but you can try to share stuff with inheritance or do even more clever things. Not sure it's worth it though.
Hope that helps.

Issue #4018, opened 2020-05-11
Add a graphql endpoint to bodhi


cverna:
We want to have the possibility to use GraphQL to query bodhi. First we need to setup a new route to give access to the GraphQL ressources.

This can be done in the following steps: 

- [x] Add [Graphene](https://docs.graphene-python.org/en/latest/) dependency to the development environment

- [x] Add a new `graphql` endpoint that route to a graphql service, and just returns "Hello World!"

- [x] Add basic tests to that new endpoint.

Notes
------

To add a new dependency in the vagrant box using pip look at an example --> https://github.com/fedora-infra/bodhi/blob/develop/devel/ansible/roles/bodhi/tasks/main.yml#L102

For the tests to run we also need to update `requirements.txt` --> https://github.com/fedora-infra/bodhi/blob/develop/requirements.txt and for the the CI tests we need to update the container images in --> https://github.com/fedora-infra/bodhi/tree/develop/devel/ci

The GraphQL service can be added here --> https://github.com/fedora-infra/bodhi/tree/develop/bodhi/server/services

The `graphql` endpoint can be added in https://github.com/fedora-infra/bodhi/blob/develop/bodhi/server/__init__.py#L297

A example of a service test can be found here --> https://github.com/fedora-infra/bodhi/blob/develop/bodhi/tests/server/services/test_releases.py

PR #4023, opened 2020-05-17
Add a graphql endpoint to Bodhi

karmadolkar:
Ref #4018 


bodhi/server/services/graphql.py
```diff
@@ -0,0 +1,64 @@
+# Copyright © 2014-2019 Red Hat Inc., and others.
```
cverna: Let's update the copyright date here

bodhi/server/services/graphql.py
```diff
@@ -0,0 +1,64 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+"""Defines API endpoints related to GraphQL objects."""
+
+import math
```
cverna: I don't think we need that :-)

bodhi/server/services/graphql.py
```diff
@@ -0,0 +1,64 @@
...
+from bodhi.server.validators import (
+    validate_tags,
+    validate_enums,
+    validate_updates,
+    validate_packages,
+    validate_release,
+)
+import bodhi.server.schemas
+import bodhi.server.services.errors
+
```
cverna, line 20: There are many imports that we don't really need now so we should remove them. You can run `flake8 bodhi/server/services/graphql.py` in vagrant and you will have a list of all the unused imports.

bodhi/tests/server/services/test_graphql.py
```diff
@@ -0,0 +1,38 @@
+# Copyright © 2014-2019 Red Hat, Inc. and others.
```
cverna: same here let's update the date

bodhi/tests/server/services/test_graphql.py
```diff
@@ -0,0 +1,38 @@
...
+from fedora_messaging import testing as fml_testing
+
+from bodhi import server
+from bodhi.server.config import config
+from bodhi.server.models import (
+    Build, PackageManager, Release, ReleaseState, UpdateType,
+    User, Update, UpdateStatus, UpdateRequest)
+from bodhi.server.util import get_absolute_path
+from bodhi.tests.server import base, create_update
+from bodhi.server.views import generic
```
cverna: Also here we have a lot of unused imports :)

bodhi/server/services/graphql.py
```diff
@@ -0,0 +1,64 @@
...
+graphql = Service(name='graphql', path='/graphql', description='graphql service')
```
cverna, line 21: I think we also need to add that service in https://github.com/fedora-infra/bodhi/blob/develop/bodhi/server/__init__.py#L297

bodhi/server/services/graphql.py
```diff
@@ -0,0 +1,64 @@
...
+def graphql_get(request):
+    """
+    Return "Hello World".
+
+    Args:
+        request (pyramid.Request): The current request.
+    Returns:
+        str: A string "Hello World".
+    """
+    return "Hello World"
```
cverna: In the end we don't specially need to have graphene here so we could probably just remove the dependency. Sorry for adding that to the task.

bodhi/server/services/graphql.py
cverna: I missed the path argument there so I think we don't need to add the route :+1:

bodhi/server/services/graphql.py
```diff
@@ -0,0 +1,34 @@
+# Copyright © 2014-2020 Red Hat Inc., and others.
```
cverna: let's remove 2014 there since this file is created in 2020

bodhi/tests/server/services/test_graphql.py
```diff
@@ -0,0 +1,26 @@
+# Copyright © 2014-2020 Red Hat, Inc. and others.
```
cverna: same here
```suggestion
# Copyright © 2020 Red Hat, Inc. and others.
```

cverna (approved):
:+1:

PR #4026, opened 2020-05-18
Don't use dummy_threading in tests, the module is gone in Python 3.9

### LINKED ISSUE #4007: bodhi fails to build with Python 3.9: _dummy_thread and dummy_threading modules have been removed

hroncok:
The suggested usage of `dummy_threading` is:

    try:
        import threading
    except ImportError:
        import dummy_threading as threading

To support installations without the threading module:
  https://docs.python.org/3.6/library/dummy_threading.html

Installations without the threading module are no longer possible since Python 3.7:
  https://docs.python.org/3.7/library/dummy_threading.html

And hence this module was removed from Python 3.9.
  https://bugs.python.org/issue37312

In the tests the dummy_thread module was used in a non-traditional way to
prevent the ComposerThread class from spawning new threads.
Instead, we reset the ComposerThread.start() and ComposerThread.join() methods.

Fixes https://github.com/fedora-infra/bodhi/issues/4007
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1814243

### DIFF
--- a/bodhi/tests/server/tasks/test_composer.py
+++ b/bodhi/tests/server/tasks/test_composer.py
@@ -19,7 +19,6 @@ from unittest import mock
 from urllib.error import HTTPError, URLError
 import urllib.parse as urlparse
 import datetime
-import dummy_threading
 import errno
 import json
 import os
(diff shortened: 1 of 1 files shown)


cverna:
Merging this since the integration test failures are unrelated. Thanks

PR #4035, opened 2020-05-23
Fix for search results not selectable.

mattiaverga:
Clicking on results of the new search bar in Bodhi stg doesn't work as expected (sometimes it works, sometimes not).
This JS hack should make it work reliably.


cverna (approved):

cverna:
I have deployed this in stg and it seems to be much better :)

Issue #4024, opened 2020-05-18
Allow to query Releases using Graphql


cverna:
We want to query all releases using the `/graphql` endpoint.

Steps : 

- [x] Add grapqhql-sqlalchemy to the dev environment
- [x] Configure `/graphql` endpoint to use graphiql (see https://docs.graphene-python.org/projects/sqlalchemy/en/latest/tutorial/#creating-graphql-and-graphiql-views-in-flask) need to research for a generic solution or https://github.com/graphql-python/webob-graphql.
- [x] Create a new graphql_schemas.py module under `bodhi/server`
- [x] Add a Release class to the graphql_schemas.py using graphql-sqlalchmy examples
- [x] Add a Query class to the graphql service
- [x] Within the Query class add a method to query all releases (use the graphql_schemas.py class)
- [x] Test everything :smile: 

Expected results:

Using graphiql query all the releases, expected output is the same as `http https://bodhi.fedoraproject.org/releases/`

PR #4036, opened 2020-05-25
[WIP]Allow to query Releases using GraphQL

karmadolkar:
ref #4024 


bodhi/server/services/graphql.py
```diff
@@ -32,3 +42,23 @@ def graphql_get(request):
         str: A string "Hello World".
     """
     return "Hello World"
+
+
+@view_config()
+def graphql_view(request):
+    context = {'session': request.session}
+    return serve_graphql_request(request, schema, context_value=context)
```
StephenCoady: this line can be deleted

bodhi/server/services/graphql.py
```diff
@@ -32,3 +42,23 @@ def graphql_get(request):
...
+
+    # Setting graphiql=True
+    return serve_graphql_request(request, schema, graphiql=True, context_value=context)
+
+
+
+class Query(graphene.ObjectType):
+    release = graphene.List(Release)
```
StephenCoady: if you call this `allReleases` or something similar then it will be more obvious to the user what this does

bodhi/server/services/graphql.py
```diff
@@ -32,3 +42,23 @@ def graphql_get(request):
...
     return "Hello World"
+
+
+@view_config()
+def graphql_view(request):
+    context = {'session': request.session}
+    return serve_graphql_request(request, schema, context_value=context)
+
+    # Setting graphiql=True
+    return serve_graphql_request(request, schema, graphiql=True, context_value=context)
```
StephenCoady: I also think the flag here should be `graphiql_enabled=True`.

It doesn't matter for now as the default is `True` but it will matter once we set this based on a variable :)

StephenCoady (commented):
Great job so far! I left some small comments.

StephenCoady:
Also, it looks like you have some unused imports in the schema and service file, so the CI will fail. 

If you need to fix those, you can run `flake8 .` :)

bodhi/server/graphql_schemas.py
```diff
@@ -0,0 +1,72 @@
...
+"""Defines schemas related to GraphQL objects."""
+from graphene import relay
+from graphene_sqlalchemy import SQLAlchemyObjectType
+
+from bodhi.server.models import Release as ReleaseModel
+
+
+class Release(SQLAlchemyObjectType):
+    """
+    Type object representing a distribution release, such as Fedora 27.
```
cverna: I think in the doc string here we should just reference that this is using the Release object from models.py so that we don't have to maintain the Attributes list and documentation here.

bodhi/server/services/graphql.py
```diff
@@ -17,18 +17,56 @@
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
 """Defines API endpoints related to GraphQL objects."""
 from cornice import Service
+from webob_graphql import serve_graphql_request
+import graphene
```
cverna: Let's move this import at the top of the import 
```suggestion
import graphene
```
```suggestion
import graphene

from cornice import Service
from webob_graphql import serve_graphql_request
```

bodhi/server/services/graphql.py
```diff
@@ -17,18 +17,56 @@
...
 @graphql.get()
 def graphql_get(request):
     """
-    Return "Hello World".
+    Perform a GET request.
+
+    Args:
+        request (pyramid.Request): The current request.
+    Returns:
+        The response from Bodhi to the request.
```
cverna: The `graphql` response to the query

bodhi/server/services/graphql.py
```diff
@@ -17,18 +17,56 @@
...
+    """
+    context = {'session': request.session}
+
+    # Setting graphiql_enabled=True
```
cverna: Let's remove this comment

bodhi/server/services/graphql.py
```diff
@@ -17,18 +17,56 @@
...
+    return serve_graphql_request(request, schema, graphiql_enabled=True, context_value=context)
```
cverna: I think we want to have the `graphiql_enabled` value to be configure in the configuration so that we can have it set to `True` in the development environment and `False` in production.

bodhi/server/services/graphql.py
```diff
@@ -17,18 +17,56 @@
...
+
+
+@graphql.post()
+def graphql_post(request):
```
cverna: I am not 100% sure here (@StephenCoady might look at that), but I am not sure that we need 2 function here we could have only one with the following declaration

```
@graphql.post()
@graphql.get()
def graphql_endpoint(request):
....
````

**Comment by cverna** [comment]
We also need to update the tests :smile:

File: devel/ansible/roles/bodhi/tasks/main.yml [hunk 10]
```diff
@@ -109,6 +109,21 @@
       name: jinja2-cli
       executable: pip3
 
+- name: pip install WebOb-GraphQL
```
cverna: here you can use a loop in ansible instead of repeating the tasks 4 times. See https://docs.ansible.com/ansible/latest/user_guide/playbooks_loops.html#iterating-over-a-simple-list

cverna:
When trying to use the endpoint with `graphiql` and using the following query 
```
{
  allReleases {
    id
    name
    longName
  }
}
```
I don't get any releases in the response, I would have expected to see a json list of all the releases.

But I might be doing something wrong :-)


image

karmadolkar:
> I don't get any releases in the response, I would have expected to see a json list of all the releases.

Yes, I think a resolver function for each field is required to fetch the data.

bodhi/server/services/graphql.py
cverna: For adding this in the configuration you can look at an example here --> 
import the configuration : 
https://github.com/fedora-infra/bodhi/blob/develop/bodhi/server/models.py#L47
get the value of a configuration key: 
https://github.com/fedora-infra/bodhi/blob/develop/bodhi/server/models.py#L1094
add the configuration key to the file : 
https://github.com/fedora-infra/bodhi/blob/develop/production.ini#L322
and https://github.com/fedora-infra/bodhi/blob/develop/devel/development.ini.example

cverna: the configuration key could be something like `enable_graphiql`

bodhi/server/services/graphql.py
StephenCoady: This is correct, function with both `post` and `get` will work the same

bodhi/server/services/graphql.py
```diff
@@ -17,18 +17,56 @@
...
+    # Setting graphiql_enabled=True
+    return serve_graphql_request(request, schema, graphiql_enabled=True, context_value=context)
+
+
+class Query(graphene.ObjectType):
+    """Allow querying objects."""
+
+    allReleases = graphene.List(Release)
+
+    def resolve_releases(self, info):
```
StephenCoady: Looks like @cverna was right, graphene handles the mapping under the hood for us, so `resolve_allReleases` will work here.

bodhi/tests/server/services/test_graphql.py
```diff
@@ -15,12 +15,95 @@
...
-        """Ensure that "Hello World" is returned"""
-        res = self.app.get('/graphql')
-        assert res.body == b'"Hello World"'
+        """Ensure that a GraphQL response is returned"""
+        res = self.app.get('/graphql', status="*", expect_errors=True)
+        assert res.body == b'{"errors":[{"message":"Must provide query string."}]}'
+
+    def test_allReleases(self):
+        """Testing allReleases"""
+        client = Client(Release)
```
StephenCoady: So `Release` here isn't actually your schema. The `schema` you want to pass to your test client here is the collection of queries and mutations you have created. Graphene sqlalchemy actually creates this for you when you call `graphene.Schema(query=Query)`. If you check `graphql.py` you've already done that there so you can use the `schema` variable :)

karmadolkar:
Riot isn't working on my end right now (Connectivity to server has been lost). :( I'll text there as soon as it works!

bodhi/server/graphql_schemas.py
```diff
@@ -0,0 +1,34 @@
...
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+"""Defines schemas related to GraphQL objects."""
+from graphene import relay, Field, String
+from graphene_sqlalchemy import SQLAlchemyObjectType
+
+from bodhi.server.models import Release as ReleaseModel
+
+
+class Release(SQLAlchemyObjectType):
+    """Type object representing a distribution release from bodhi.server.models, such as Fedora 27."""
```
StephenCoady: this is the line that is breaking your lint, if you make it multi-line the flake8 run should pass

bodhi/tests/server/services/test_graphql.py
```diff
@@ -15,12 +15,47 @@
...
+            pending_stable_tag='f22-updates-pending',
+            override_tag='f22-override',
+            branch='f22',
+            package_manager=PackageManager.dnf,
+            testing_repository='updates-testing')
+
+        self.db.add(release)
+        self.db.commit()
+
+        executed = client.execute("""{  allReleases{  name  }}""")
```
StephenCoady, line 52: It would be nice to test for some of the trickier fields like `state` or `package_manager` (or both!). You could request it from allReleases and make sure it matched the value provided to db.add(). I think that could be a separate test.

bodhi/server/graphql_schemas.py
```diff
@@ -0,0 +1,34 @@
...
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+"""Defines schemas related to GraphQL objects."""
+from graphene import relay, Field, String
+from graphene_sqlalchemy import SQLAlchemyObjectType
+
+from bodhi.server.models import Release as ReleaseModel
+
+
+class Release(SQLAlchemyObjectType):
+    """Type object representing a distribution release from bodhi.server.models, like Fedora 27"""
```
StephenCoady: this is the line that is breaking your lint, if you make it multi-line the flake8 run should pass

StephenCoady (changes requested):

cverna:
This lgtm, @StephenCoady if you want to give it another look :-)

cverna (approved):

StephenCoady (approved):
looks good to me. great work :)

PR #4037, opened 2020-05-25
Making sure that builds of side tags update are marked as signed.

### LINKED ISSUE #4032: Side tag updates are not automatically pushed to testing
When I create a side tag update for a stable Fedora, it sits in pending until I manually click "push to testing". This is inconsistent with regular updates and a tad confusing.

See also https://pagure.io/fedora-infrastructure/issue/8930

cverna:
Fixes #4032.

### DIFF
--- a/bodhi/server/consumers/signed.py
+++ b/bodhi/server/consumers/signed.py
@@ -89,7 +89,9 @@ class SignedHandler(object):
                 log.info('Build is not assigned to release, skipping')
                 return
 
-            if build.update and build.update.from_tag:
+            if build.update \
+               and build.update.from_tag \
+               and not build.update.release.composed_by_bodhi:
                 koji_testing_tag = build.release.get_testing_side_tag(build.update.from_tag)
                 if tag != koji_testing_tag:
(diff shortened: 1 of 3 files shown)


bodhi/server/consumers/signed.py
```diff
@@ -89,7 +89,9 @@ def __call__(self, message: fedora_messaging.api.Message):
                 log.info('Build is not assigned to release, skipping')
                 return
 
-            if build.update and build.update.from_tag:
+            if build.update and \
+               build.update.from_tag and not \
+               build.update.release.composed_by_bodhi:
```
pypingou: For readability I tend to put the and on the new line and I would encourage it even more for the ``not`` as it is pretty important :)

pypingou:
One nitpick, let me know what you think of it, happy to approve afterward

bodhi/server/consumers/signed.py
cverna: updated :)

pypingou (approved):
Looking good!

PR #4043, opened 2020-05-30
Migrate TestCase-Package relationship to TestCase-Build

### LINKED ISSUE #824: Updates list all testcases, not just those from a proper category
[This anaconda update](https://bodhi.fedoraproject.org/updates/FEDORA-2016-cd26edfec7) lists this test case:
[User:Pschindl/Draft/Testcase Anaconda custom partitioning](https://fedoraproject.org/wiki/User:Pschindl/Draft/Testcase_Anaconda_custom_partitioning?redirect=no)

We tried to get rid of that by renaming that page, but the original page remained as a redirect, and is still listed on the bodhi page. But I found out that it's not the core issue.

The core issue is that our [package test plan SOP](https://fedoraproject.org/wiki/QA:SOP_package_test_plan_creation) says that only testcases listed in `Category:Package_(packagename)_test_cases` should be displayed. So for anaconda, only these test cases should be visible:
https://fedoraproject.org/wiki/Category:Package_anaconda_test_cases

It seems that Bodhi does not search the available testcases according to the categories, but just by matching titles of all available test cases on the whole wiki (isn't that slow?). That's not correct. Please only show testcases from the correct category, according to the SOP, thanks.

CCing @AdamWill to make sure I understood the intended workflow and the SOP correctly :-)

### LINKED ISSUE #1794: Package.fetch_test_cases() never removes tests
It seems that QA tests that are removed from the Wiki do not get removed in Bodhi. I took a peek at the [code](https://github.com/fedora-infra/bodhi/blob/2.10.1/bodhi/server/models.py#L665-L707) that synchronizes Bodhi's Package test cases with the Wiki, and I observed that it adds new test cases it finds in the wiki to Packages, but it does not remove old ones:

```
        for test in set(list_categorymembers(wiki, cat_page)):
            case = db.query(TestCase).filter_by(name=test).first()
            if not case:
                case = TestCase(name=test, package=self)
                db.add(case)
                db.flush()
```

There should also be a mechanism in there to find tests that are associated with the Package that are no longer found by ```list_categorymembers()```. It would probably also be good to clean up orphaned tests when unlinking them from a Package leaves them linked to no other Packages (and they might have a 1-1 relationship anyway, not sure...)

Here's an example update that should not have https://fedoraproject.org/wiki/User:Pschindl/Draft/Testcase_Custom_partitioning?rd=User:Pschindl/Draft/Testcase_Anaconda_custom_partitioning associated with it anymore:

https://bodhi.fedoraproject.org/updates/FEDORA-2017-5361e86979

@AdamWill reported this problem via IRC today.

### LINKED ISSUE #2979: Missing TestCases in some packages
Example:

> Package name: initial-setup
> Package id  : 4877

> Package name: gnome-initial-setup
> Package id  : 5876

For both packages, `fetch_test_cases` returns testcase name:`QA:Testcase base initial setup`.
But  we keep only one DB object with name `QA:Testcase base initial setup`, binded to package with id 5876(gnome-initial-setup).
Calling `fetch_test_cases`, Bodhi is trying to find TC with proper name, but omit to check related package id binded to TC. When several packages have TC with the same name, only first package running `fetch_test_cases` will get TC.

mattiaverga:
This will also allow to remove testcases no longer listed in the wiki.
Fix #824 
Fix #1794 
Fix #2979 

### DIFF
--- a/bodhi/server/migrations/versions/325954bac9f7_link_testcases_to_builds.py
+++ b/bodhi/server/migrations/versions/325954bac9f7_link_testcases_to_builds.py
@@ -1,0 +1,89 @@
+# Copyright (c) 2020 Mattia Verga
+#
+# This file is part of Bodhi.
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+"""
+Link testcases to builds.
+
+Revision ID: 325954bac9f7
+Revises: ff834fa4f23e
+Create Date: 2020-05-22 16:56:05.451523
+"""
+from alembic import op
+import sqlalchemy as sa
+
+
+# revision identifiers, used by Alembic.
+revision = '325954bac9f7'
+down_revision = 'ff834fa4f23e'
+
+
+def upgrade():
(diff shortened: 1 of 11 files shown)


cverna (approved):
LGTM, thanks for looking at this

PR #4047, opened 2020-06-02
Use podman as default to run the ci.

### LINKED ISSUE #3142: Get bodhi-ci's --container-runtime podman option working
```devel/ci/bodhi-ci``` has a ```--container-runtime``` flag that can be set to ```podman```. It hasn't ever worked fully, but may work better now with newer versions of podman. Let's see if we can get it working as well as we have ```docker``` working.

cverna:
The integration tests are still using docker since they
interface directly with the docker daemon. But the rest of the
tests run during the CI are now using podman.

Fixes #3142

### DIFF
--- a/devel/ci/bodhi-ci
+++ b/devel/ci/bodhi-ci
@@ -111,7 +111,7 @@ concurrency_option = click.option(
     help=('Number of concurrent processes to run. Integration test runs are separately limited due '
           'to resource contention. Defaults to the number of cores detected'))
 container_runtime_option = click.option(
-    '--container-runtime', '-c', default='docker', type=click.Choice(['docker', 'podman']),
+    '--container-runtime', '-c', default='podman', type=click.Choice(['docker', 'podman']),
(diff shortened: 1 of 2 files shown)


mattiaverga:
While I was looking at this some weeks ago, I've seen that we now have catatonit packaged in F >= 31
We may want to install it as well to support the `--init` option (?):
https://github.com/fedora-infra/bodhi/blob/efb276469ad1fdba967efe5a9c574ca4e8769491/devel/ci/bodhi-ci#L612-L614

cverna:
> While I was looking at this some weeks ago, I've seen that we now have catatonit packaged in F >= 31
> We may want to install it as well to support the `--init` option (?):
> 
> https://github.com/fedora-infra/bodhi/blob/efb276469ad1fdba967efe5a9c574ca4e8769491/devel/ci/bodhi-ci#L612-L614

Ah yes I missed this one :+1: .

I thought the integration tests would work, but they are failing because the container images are built by podman and therefore not available to docker :(

PR #4048, opened 2020-06-02
Make comments.user_id not nullable and tweak some joins

mattiaverga:
This splits off joins changes from #4017 
Setting the `user_id` field for comments to be not nullable makes possible to use faster inner joins. Also, I've disable the `lazy='joined'` relationship property in the Update model against builds and comments: this will avoid to unconditionally load all the builds and all the comments associated to the update when we retrieve it from the database.
In my opinion, this should give better performance. Locally I can see some speed improvements when loading some pages, like builds or users lists, but unfortunately the updates page list is not getting substantial improvement.

### DIFF
--- a/bodhi/server/migrations/versions/f50dc199039c_make_comments_user_id_not_nullable.py
+++ b/bodhi/server/migrations/versions/f50dc199039c_make_comments_user_id_not_nullable.py
@@ -1,0 +1,45 @@
+# Copyright (c) 2020 Mattia Verga
+#
+# This file is part of Bodhi.
+#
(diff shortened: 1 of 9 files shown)


mattiaverga:
I'm not sure if these CI test failures are real or not, I'm not able to reproduce them locally...
It would be nice to see in staging if this change can fix https://pagure.io/fedora-infrastructure/issue/8949

cverna (approved):
I ran the tests locally and every was green, so going to merge this manually.

Thanks for the PR.

PR #4052, opened 2020-06-14
Always use koji.multiCall for untag/unpush for better handle updates …

mattiaverga:
…with a lot of builds


cverna (approved):
LGTM

PR #4053, opened 2020-06-14
Correct infra-tags.repo url

mattiaverga:
Use the new pagure ansible repo for downloading the infra-tag.repo file to fix ci failures.
Also remove a workaround no more needed with F>=31


cverna:
I have incorporated these changes in https://github.com/fedora-infra/bodhi/pull/4057 with the addition of removing f30 and also configure cgroups v1 on the vagrant box so that we can keep docker.

mattiaverga:
Great, I'll retire this one

PR #4063, opened 2020-06-20
Disable manual creation of updates for releases not composed by Bodhi

### LINKED ISSUE #4058: Disable manual creation of updates for releases not composed by Bodhi
For whatever reason, it may happen that sometimes Bodhi misses a fedora-message for a newly created build for Rawhide. This means that the update isn't automatically created and the build doesn't get tagged into rawhide.

Now, a user may be tempted to create that update manually, but that's the wrong thing to do: since Bodhi doesn't compose Rawhide, the update will never be pushed to testing/stable.

I think that in such situation all that the users should do is to remove the "updates-candidate" from the affected build and then re-add the same tag. This will trigger another fedora-message that (hopefully) will be caught by Bodhi so that the update will be automatically created as usual.

So we should disallow users to post such updates (if they don't come from side-tags) and document the correct way in docs.

Any thoughts?

mattiaverga:
Fix #4058 


cverna (approved):
LGTM

PR #4064, opened 2020-06-20
Add a periodic task to check sign status of Builds in pending Updates

mattiaverga:
It may happen that Bodhi misses fedora-messaging messages announcing builds
have been signed.
In these cases, the Update remains stuck in pending until a manual intervention.
This script will daily cycle through builds of Updates in pending status and update
the signed status in the db to match the tags found in Koji.

### DIFF
--- a/bodhi/server/tasks/check_signed_builds.py
+++ b/bodhi/server/tasks/check_signed_builds.py
@@ -1,0 +1,85 @@
+# Copyright © 2017 Red Hat, Inc.
+#
+# This file is part of Bodhi.
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+
+"""
+Avoid Updates being stuck in pending.
+
+It may happen that Bodhi misses fedora-messaging messages announcing builds
+have been signed.
+In these cases, the Update remain stuck in pending until a manual intervention.
+
+This script will cycle through builds of Updates in pending status and update
+the signed status in the db to match the tags found in Koji.
+"""
+
+import logging
+from datetime import datetime, timedelta
+
+from bodhi.server import buildsys, models
+from bodhi.server.config import config
+from bodhi.server.util import transactional_session_maker
+
+
+log = logging.getLogger(__name__)
+
+
+def main():
+    """Check build tags and sign those we missed."""
+    db_factory = transactional_session_maker()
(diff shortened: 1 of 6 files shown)


cverna:
> It may happen that Bodhi misses fedora-messaging messages announcing builds
> have been signed.

So Bodhi will not miss messages since they are queue on the rabbitmq broker until consumed by the signed consumer, we have had a problem with normal release update using a sidetag not having the build signed because of a bug in the consumer (https://github.com/fedora-infra/bodhi/pull/4037).

> In these cases, the Update remains stuck in pending until a manual intervention.
> This script will daily cycle through builds of Updates in pending status and update
> the signed status in the db to match the tags found in Koji.

Generally I think that having that periodic tasks would not hurt, but in the same time I am not 100% sure it will be super useful so I am a bit on the fence about it.

@mattiaverga what are your thoughts ?

> 
> Signed-off-by: Mattia Verga [mattia.verga@tiscali.it](mailto:mattia.verga@tiscali.it)

mattiaverga:
mmm I'm quite sure I saw some updates not from side tags which Bodhi considered not signed, but with builds perfectly signed. I have now fixed most of them, but as example look at https://bodhi.fedoraproject.org/updates/FEDORA-2020-c546c08554

Another effect of the script is to obsolete empty updates, we have quite a few of these stuck in pending, just look at the last page of the pending updates list:
https://bodhi.fedoraproject.org/updates/?search=&status=pending&releases=__current__&releases=__pending__&page=4
This doesn't hurt anything, it's just to have things cleaner.

However, I will accept any decision you make about this.

cverna:
> mmm I'm quite sure I saw some updates not from side tags which Bodhi considered not signed, but with builds perfectly signed. I have now fixed most of them, but as example look at https://bodhi.fedoraproject.org/updates/FEDORA-2020-c546c08554
> 
> Another effect of the script is to obsolete empty updates, we have quite a few of these stuck in pending, just look at the last page of the pending updates list:
> https://bodhi.fedoraproject.org/updates/?search=&status=pending&releases=__current__&releases=__pending__&page=4
> This doesn't hurt anything, it's just to have things cleaner.

Ok that sounds good to me :-)

> 
> However, I will accept any decision you make about this.

bodhi/server/scripts/compat.py
```diff
@@ -94,6 +94,18 @@ def check_policies(config_uri):
     _trigger_task(config_uri, "check_policies_task")
 
 
+@click.command()
+@config_uri_argument
+@click.version_option(message='%(version)s')
+def check_signed_builds(config_uri):
```
cverna: We had this as a way to keep backward compatibility with older script run by cron jobs. Since this tasks is new I am not 100% sure we need that compatibility layer

bodhi/server/tasks/check_signed_builds.py
```diff
@@ -0,0 +1,84 @@
...
+from bodhi.server.util import transactional_session_maker
+
+
+log = logging.getLogger(__name__)
+
+
+def main():
+    """Check build tags and sign those we missed."""
+    db_factory = transactional_session_maker()
+    older_than = datetime.utcnow() - timedelta(days=2)
```
cverna: Should we get the timedelta value from the config ? that way it is easier to change if needed ?

cverna (commented):
Couple comments otherwise lgtm

bodhi/server/tasks/check_signed_builds.py
mattiaverga: Good idea, I'll change that.

cverna (approved):

PR #4079, opened 2020-07-02
Added metrics endpoint for scraping by Prometheus.

AdamSaleh:
Introduces prometheus-client to expose an endpoint for  metrics collection.
Uses a tween to hook into all of the incomming requests and gather basic information,
to compute metrics on number of requests grouped by endpoints and their result status. 

Signed-off-by: Adam Saleh <asaleh@redhat.com>

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -14,6 +14,7 @@ graphene
 graphene-sqlalchemy
 jinja2
 markdown
+prometheus_client
 psycopg2
 py3dns  # Should be required by pyLibravatar https://bugs.launchpad.net/pylibravatar/+bug/1661218
 pylibravatar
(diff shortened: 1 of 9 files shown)


requirements.txt
```diff
@@ -14,6 +14,7 @@ graphene
 graphene-sqlalchemy
 jinja2
 markdown
+prometheus_client
```
cverna, line 17: We also need this on the vagrant box --> https://github.com/fedora-infra/bodhi/blob/develop/devel/ansible/roles/bodhi/tasks/main.yml#L4

cverna (commented):
Just one small change needed otherwise looks good thanks.

requirements.txt
AdamSaleh: Done!

cverna (approved):

Issue #4077, opened 2020-07-01
Graphql : Query specific releases


cverna:
Now that we are able to query all the releases from bodhi using Graphql we should add another query that allow users to filter the results.

I think we should be able to query releases using the following attributes : 

* name
* id_prefix
* state
* composed_by_bodhi

Expected results:

We want to be able to use a query similar to 
```
{
  releases(state: "pending") {
    name
    version
  }
}
```
or

```
{
  releases(id_prefix: "FEDORA") {
    name
    state
  }
}
```

PR #4082, opened 2020-07-08
Allow querying specific releases

karmadolkar:
Ref #4077 

### DIFF
--- a/bodhi/server/services/graphql.py
+++ b/bodhi/server/services/graphql.py
@@ -21,7 +21,7 @@ from cornice import Service
 from webob_graphql import serve_graphql_request
 
 from bodhi.server.config import config
-from bodhi.server.graphql_schemas import Release
+from bodhi.server.graphql_schemas import Release, ReleaseModel
 
 
 graphql = Service(name='graphql', path='/graphql', description='graphql service')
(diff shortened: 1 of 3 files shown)


bodhi/server/services/graphql.py
```diff
@@ -48,11 +48,107 @@ class Query(graphene.ObjectType):
     """Allow querying objects."""
 
     allReleases = graphene.List(Release)
+    specificReleases = graphene.Field(lambda: graphene.List(Release),
```
cverna: Let's just use the following fields `name, id_prefix, state, composed_by_bodhi`

bodhi/server/services/graphql.py
```diff
@@ -48,11 +48,107 @@ class Query(graphene.ObjectType):
...
     def resolve_allReleases(self, info):
         """Answer Queries by fetching data from the Schema."""
         query = Release.get_query(info)  # SQLAlchemy query
         return query.all()
 
+    def resolve_specificReleases(self, info, **args):
+        """Resolving releases with a given argument"""
+        query = Release.get_query(info)
+
+        id_prefix = args.get("id_prefix")
```
cverna, line 65: same here we can really focus on the following fields `name, id_prefix, state, composed_by_bodhi`

bodhi/server/services/graphql.py
```diff
@@ -48,11 +48,37 @@ class Query(graphene.ObjectType):
     """Allow querying objects."""
 
     allReleases = graphene.List(Release)
+    specificReleases = graphene.Field(
```
cverna: I am not a big fan of `specificReleases` for the name of the query, it is a bit long to type for me :P. @StephenCoady do you have an idea on how we could name that ?

bodhi/tests/server/services/test_graphql.py
```diff
@@ -92,3 +92,63 @@ def test_enumfields(self):
...
+            "data": {
+                "specificReleases": [{
+                    "name": "F17"
+                }, {
+                    "name": "F22"
+                }]
+            }
+        }
+
+        executed = client.execute("""{  specificReleases(state: "current"){  name  }}""")
```
cverna: I would also add a test case where we have multiple attributes something like 
```suggestion
        executed = client.execute("""{  specificReleases(state: "pending", composedByBodhi: false){  name  }}""")
```

cverna (commented):
Looking good just a few comments :-)

bodhi/server/services/graphql.py
StephenCoady: Yeah I'd agree. Lets make it descriptive but simple, how about `getRelease` ?

bodhi/tests/server/services/test_graphql.py
```diff
@@ -92,3 +92,64 @@ def test_enumfields(self):
                 }]
             }
         }
+
+    def test_getRelease(self):
+        """Testing getRelease."""
+        client = Client(schema)
```
cverna, line 70: We could use a setup_method in the TestGraphQLService class to declare the clien and release variable (see https://github.com/fedora-infra/bodhi/blob/develop/bodhi/tests/server/test_config.py#L55). That way you don't have to copy/paste that to every tests and you can use self.client and self.release.

cverna (approved):
LGTM

StephenCoady (approved):
sorry for the delay reviewing. yeah this looks good to me. I've tried it out too and its working as expected :) well done.

PR #4086, opened 2020-07-19
Querying Updates using arguments

karmadolkar:
Returns null on the GraphiQL client as of now. Uses `stable_karma`, `unstable_karma`, `stable_days`, `status`, `request`, `pushed`, `critpath`, `date_approved`, `alias`, and `user_id` as arguments to Query.

@cverna @StephenCoady 

### DIFF
--- a/bodhi/server/services/graphql.py
+++ b/bodhi/server/services/graphql.py
@@ -21,8 +21,7 @@ from cornice import Service
 from webob_graphql import serve_graphql_request
 
 from bodhi.server.config import config
-from bodhi.server.graphql_schemas import Release, ReleaseModel
-
+from bodhi.server.graphql_schemas import Release, ReleaseModel, Update, UpdateModel
 
 graphql = Service(name='graphql', path='/graphql', description='graphql service')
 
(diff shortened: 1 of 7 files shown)


bodhi/server/services/graphql.py
```diff
@@ -48,11 +48,91 @@ class Query(graphene.ObjectType):
...
+        if composed_by_bodhi is not None:
+            query = query.filter(ReleaseModel.composed_by_bodhi == composed_by_bodhi)
+
+        state = args.get("state")
+        if state is not None:
+            query = query.filter(ReleaseModel.state == state)
+
+        return query.all()
+
+    def resolve_getUpdates(self, info, **args):
```
cverna, line 90:
```suggestion
    def resolve_getUpdate(self, info, **args):
```

StephenCoady:
This looks good to me so far. You still need to rename the resolver as suggested above though. Also, I think you have linting issues.

As a side note, we may need to move some of this logic to separate files. I think its fine for now but as soon as we add a couple more resolvers then the file is going to become pretty difficult to follow. Maybe we could nest everything under a `services/graphql/` folder. 

Also, the way you had to do types is fine, I suspect its because `type` is a built in function in python.

StephenCoady (changes requested):
Almost there! Sorry to be difficult, and I realise I was the one to suggest the name in the first place, but could we make `getRelease` and `getUpdate` plural? so `getReleases` etc. I think its more descriptive since its possible to get back multiple releases and updates.

Other than that I'm happy and will approve :+1:

StephenCoady (approved):
looks good to me!

Issue #4092, opened 2020-07-29
We want to query all users using the `/graphql` endpoint.


StephenCoady:
Steps: 

- [ ] Add a `User` class to the graphql_schemas.py
- [ ] Add a resolver to the graphql service
- [ ] Add tests

PR #4093, opened 2020-08-02
Query all users using the /graphql endpoint

karmadolkar:
Ref #4092.

### DIFF
--- a/bodhi/server/graphql_schemas.py
+++ b/bodhi/server/graphql_schemas.py
@@ -19,7 +19,7 @@
 from graphene import relay, Field, String
 from graphene_sqlalchemy import SQLAlchemyObjectType
 
-from bodhi.server.models import Release as ReleaseModel, Update as UpdateModel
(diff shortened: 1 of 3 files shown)


ryanlerch:
Removed the [WIP] from the title, and marked this as a draft PR in github here.

@karmadolkar is this one still a WIP (draft)? or ready for reviewing?

AdamSaleh:
@mattiaverga this has been sitting here almost a year and in production we are patching all of the graphql code out.

In our other projects we have been looking at openapi/swagger rather than graphql, so I would rath phase this experiment out and talk about refactoring the api to have nice machine-readable documentation, without the necesity to rewrite it in a new paradigm. What do you think?

This is of course with apologies to @karmadolkar whose code we let sit here for such a long time.

mattiaverga:
Yeah, as I wrote in other tickets, lately graphql has been unmaintained and the v3 stack needed by Python 3.10 is broken due to the lack of graphql-sqlalchemy upgrade.

I don't think anyone is using Bodhi graphql APIs yet, so I would assume it's safe to drop them.

PR #4100, opened 2020-08-11
Do not let update through when bodhi fails to talk to greenwave

### LINKED ISSUE #4089: Do not let update through when bodhi fails to talk to greenwave
Now that bodhi and greenwave have been together for a while and that we can see that appart from the unplanned outage, they are performing well. I believe it is time to stop letting bodhi push update through when it fails to talk to greenwave.

The current flow is as follow:
- bodhi queries greenwave
- if greenwave returns an error code or times-out
- bodhi sets the gating status to "failed"
- if there is a push, bodhi will take update that are in gating status "passed" and "failed", therefore letting through updates that should potentially be gated, but we do not know for sure because for some reason we failed to get a clear answer from greenwave.

There are two ways we can address this:
- Stop setting the gating status to "failed" when we fail to talk to greenwave (pros: less noise on the update, less notifications, less confusions to the users, cons: we don't know why the gating status has not been updated)
- Stop including updates with gating status "failed" in the push (pros/cons: the opposite as above)

cverna:
This commit removes the greenwave_failed status of
the test_gating_status update attribute. Instead in case of
a failure to communicate to greenwave we set the status to
waiting and rely on the fedora-messaging greenwave consumer
to update the gating status.

The greenwave_failed status is still present in the database
table and old updates having that status are not changed.

Fixes #4089

### DIFF
--- a/bodhi/server/models.py
+++ b/bodhi/server/models.py
@@ -606,7 +606,6 @@ class TestGatingStatus(DeclEnum):
     running = 'running', 'Running'
     passed = 'passed', 'Passed'
     failed = 'failed', 'Failed'
-    greenwave_failed = 'greenwave_failed', 'Greenwave failed to respond'
 
 
 class UpdateType(DeclEnum):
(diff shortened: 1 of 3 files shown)


pypingou (approved):
Thanks for handling this!

PR #4102, opened 2020-08-12
Fix 404 pages for bot users with nonstandard characters in usernames

### LINKED ISSUE #3993: Bad characters in username

mattiaverga:
This PR changes the default Pyramid's regex for replacement markers allowing usernames to have slashes.
https://docs.pylonsproject.org/projects/pyramid/en/latest/narr/urldispatch.html#route-pattern-syntax
I've used a rather permissive regex here (all non-whitespace characters) so anything behind `/users/` will be considered to be the username to search. I don't see any security concerns or other troubles for that, correct me if I'm wrong.

Fixes https://bodhi.fedoraproject.org/users/bpeck/jenkins-continuous-infra.apps.ci.centos.org
Fixes #3993 

### DIFF
--- a/bodhi/server/services/user.py
+++ b/bodhi/server/services/user.py
@@ -32,7 +32,7 @@ import bodhi.server.services.errors
 import bodhi.server.services.updates
 
 
-user = Service(name='user', path='/users/{name}',
+user = Service(name='user', path=r'/users/{name:\S+}',
                description='Bodhi users',
                # These we leave wide-open since these are only GETs
                cors_origins=bodhi.server.security.cors_origins_ro)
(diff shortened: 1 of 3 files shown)


cverna:
Do we want to add a unit tests for it ?

cverna (approved):

PR #4104, opened 2020-08-13
Fix encoding of package and user names in search results

mattiaverga:


cverna:
Do we want a news file for that ?

cverna (approved):

PR #4108, opened 2020-08-21
Do not force critpath packages to stay in testing for long pre-beta

pypingou:
If a release is not composed by bodhi, we should not force packages in
critpath to stay 14 days in testing. Otherwise, between branching and
beta, when bodhi is still in "rawhide" mode, critpath packages get stuck
for 14 days.

### DIFF
--- a/bodhi/server/models.py
+++ b/bodhi/server/models.py
@@ -1993,7 +1993,7 @@ class Update(Base):
         :return: The number of mandatory days in testing.
         :rtype:  int
         """
-        if self.critpath:
+        if self.critpath and self.release.composed_by_bodhi:
             return config.get('critpath.stable_after_days_without_negative_karma')
 
         days = self.release.mandatory_days_in_testing


pypingou:
This should do, though a better fix maybe to not update the critpath field pre-beta, but I didn't find where this is done in bodhi (and I'm running out of time to search it right now).

Feel free to accept or reject if you find a better solution.

Here is an example update that was stuck until I edited its critpath value in the DB: https://bodhi.fedoraproject.org/updates/FEDORA-2020-bac5ae64cf

mattiaverga:
I think we must override the karma critpath setting also to make this work, but I'm not completely sure.

Anyway, do we want to perform any check for rawhide releases? There's really no testing repository, so I don't think there's a reason to have any update stuck in testing. If gating is successful, I think any new update should be pushed directly to stable, shouldn't it?

bodhi/server/models.py
```diff
@@ -1993,7 +1993,7 @@ def mandatory_days_in_testing(self):
         :return: The number of mandatory days in testing.
         :rtype:  int
         """
-        if self.critpath:
+        if self.critpath and self.release.composed_by_bodhi:
```
cverna, line 1996: we should probably add a test for that new condition. What do you think ?

mattiaverga:
So, I would close this one in favor of #4135 

We will need to adjust the configuration template in ansible for pre-beta:
https://pagure.io/fedora-infra/ansible/blob/master/f/roles/bodhi2/base/templates/production.ini.j2#_586
and post-beta:
https://pagure.io/fedora-infra/ansible/blob/master/f/roles/bodhi2/base/templates/production.ini.j2#_594

mattiaverga:
See https://pagure.io/fedora-infra/ansible/pull-request/318

PR #4111, opened 2020-08-22
Set update.stable_days to 0 for Releases not composed by Bodhi itself

mattiaverga:

### DIFF
--- a/bodhi/server/models.py
+++ b/bodhi/server/models.py
@@ -2317,6 +2317,8 @@ class Update(Base):
             # For rawhide updates make sure autotime push is enabled
             # https://github.com/fedora-infra/bodhi/issues/3912
             data['autotime'] = True
+            data['stable_days'] = 0
+            log.debug("Overriding autotime settings for rawhide update.")
 
         up = Update(**data, release=release)
 
--- a/bodhi/tests/server/services/test_updates.py
+++ b/bodhi/tests/server/services/test_updates.py
@@ -336,8 +336,10 @@ class TestNewUpdate(BasePyTestCase):
         # Let's test what happens if the user sets autotime to False
         # Rawhide workflow must ignore this setting
         update['autotime'] = False
+        update['stable_days'] = 7
         with mock.patch('bodhi.server.buildsys.DevBuildsys.getTag', self.mock_getTag):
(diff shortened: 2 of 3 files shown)


bodhi/server/models.py
```diff
@@ -2317,6 +2317,8 @@ def new(cls, request, data):
             # For rawhide updates make sure autotime push is enabled
             # https://github.com/fedora-infra/bodhi/issues/3912
             data['autotime'] = True
+            data['stable_days'] = 0
```
cverna, line 2320: I don't think that should be hardcoded, it should be something we can configure using the configuration file maybe ? Something like we the mandatory_days_in_testing --> https://pagure.io/fedora-infra/ansible/blob/master/f/roles/bodhi2/base/templates/production.ini.j2#_597

What do you think ?

mattiaverga: That's about the same question I asked in #4108 : is there any reason for releases not composed by Bodhi to have updates stuck in testing, since no testing repository exists?

The scope of this PR is to prevent users being able to manually set `stable_days` on the update, if we don't want to hardcode there's no need to create another setting, I think we could just override user input with the release default.

cverna: OK sounds good, I was lacking a bit of context :-) :+1:

cverna (approved):

pypingou:
On Tue, Aug 25, 2020 at 09:19:22AM -0700, Mattia Verga wrote:
>    @mattiaverga commented on this pull request.
> 
>      ----------------------------------------------------------------------
> 
>    In [1]bodhi/server/models.py:
> 
>  > @@ -2317,6 +2317,8 @@ def new(cls, request, data):
>               # For rawhide updates make sure autotime push is enabled
>               # https://github.com/fedora-infra/bodhi/issues/3912
>               data['autotime'] = True
>  +            data['stable_days'] = 0
> 
>    That's about the same question I asked in [2]#4108 : is there any reason
>    for releases not composed by Bodhi to have updates stuck in testing, since
>    no testing repository exists?

There is in the context of gating, in rawhide with gating updates are created
directly in testing and then go to stable if gating allows.
So updates can be stuck in testing if their tests fail (iirc there is/was one
for F33) and this is normal/expected.

PR #4119, opened 2020-08-31
Allow searching updates by `from_side_tag` boolean.

mattiaverga:
Add form field in webUI for the new parameter and for gating status.

### DIFF
--- a/bodhi/client/__init__.py
+++ b/bodhi/client/__init__.py
@@ -591,6 +591,11 @@ def edit(user: str, password: str, url: str, debug: bool, openid_api: str, **kwa
 @click.option('--packages', help='Query by package name(s)')
 @click.option('--content-type', help='Query updates based on content type',
               type=click.Choice(['rpm', 'module']))  # And someday, container.
+@click.option('--from-side-tag/--not-from-side-tag', default=None,
+              help='List only updates created from side-tag / not from side-tag')
+@click.option('--gating', help='Filter by gating status', default=None,
+              type=click.Choice(['passed', 'failed', 'ignored', 'waiting',
+                                 'running', 'queued', 'greenwave_failed']))
 @click.option('--pushed', is_flag=True, default=None,
               help='Filter by pushed updates')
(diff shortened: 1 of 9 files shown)


mattiaverga:
Work In Progress
I'm going to add the same functionality to the CLI

bodhi/client/__init__.py
```diff
@@ -591,6 +591,11 @@ def edit(user: str, password: str, url: str, debug: bool, openid_api: str, **kwa
 @click.option('--packages', help='Query by package name(s)')
 @click.option('--content-type', help='Query updates based on content type',
               type=click.Choice(['rpm', 'module']))  # And someday, container.
+@click.option('--from-side-tag/--not-from-side-tag', default=None,
+              help='List only updates created from side-tag / not from side-tag')
+@click.option('--gating', help='Filter by gating status', default=None,
+              type=click.Choice(['passed', 'failed', 'ignored', 'waiting',
+                                 'running', 'queued', 'greenwave_failed']))
```
cverna, line 591: I think we have removed the greenwave_failed status.

docs/user/man_pages/bodhi.rst
```diff
@@ -404,6 +404,18 @@ The ``updates`` command allows users to interact with bodhi updates.
...
 
+    ``--from-side-tag``
+
+        Query for updates created from a side-tag.
+
+    ``--not-from-side-tag``
+
+        Query for updates not created from a side-tag.
+
+    ``--gating [passed | failed | ignored | waiting | running | queued | greenwave_failed]``
```
cverna, line 415: same thing here about greenwave_failed

cverna (commented):
A small comment otherwise looks good

bodhi/client/__init__.py
cverna: Actually we still have that in the db since older update can still have that status, so we should have that value here :+1:

cverna (approved):

PR #4125, opened 2020-09-12
Set testing request for updates from side-tags for non-rawhide releases

### LINKED ISSUE #4087: updates created from side tags are not automatically submitted for testing like single-package updates
After creating updates for stable Fedora releases, e.g.:
https://bodhi.fedoraproject.org/updates/FEDORA-2020-c5e96618e7
https://bodhi.fedoraproject.org/updates/FEDORA-2020-4163779303
I expected they would be submitted to testing automatically, like single-package updates. After two days I noticed they were still in "pending" state and found I had to submit them for testing manually.
This is a surprising difference and not easily discoverable if it's the first time you're doing it. Please make the experience consistent with single-package updates.

mattiaverga:
Side-tag rawhide updates are pushed directly to testing by the signed consumer.
For side-tag non-rawhide updates, the push is done by the composer, if the update.request is set to testing. But the request is never set.

This PR will make the signed consumer to set the request to testing to those updates when all builds are signed. It could also be set before all builds are signed, but I think it's nicer to emit the message on fedmsg only when it's really ready to be processed by the composer.

fixes #4087 

### DIFF
--- a/bodhi/server/consumers/signed.py
+++ b/bodhi/server/consumers/signed.py
@@ -28,7 +28,7 @@ import fedora_messaging
 from sqlalchemy import func
 
 from bodhi.server.config import config
-from bodhi.server.models import Build, UpdateStatus, TestGatingStatus
+from bodhi.server.models import Build, UpdateRequest, UpdateStatus, TestGatingStatus
 from bodhi.server.util import transactional_session_maker
 
 log = logging.getLogger('bodhi')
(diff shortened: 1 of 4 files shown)


bodhi/server/services/updates.py
```diff
@@ -612,6 +612,12 @@ def new_update(request):
                             pending_signing_tag=pending_signing_tag,
                             from_tag=from_tag,
                             candidate_tag=candidate_tag)
+                        # Set request to testing
+                        # https://github.com/fedora-infra/bodhi/issues/4087
+                        log.debug(f"Setting request for new update {u.alias} from side-tag.")
+                        db = request.db
+                        req = UpdateRequest.testing
+                        u.set_request(db, req, request.user.name)
```
mattiaverga: I've realized this may not be the best: since the tagging process is made asynchronously by a celery task, we probably end in a situation were the set_request is performed before the celery task tagging process, thus sending a notification on the bus before builds get tagged.

Should I move the set_request action inside the celery task? If so I will need to pass the update alias and the user.name to the task and I'll need to open a database session inside within the celery task.
Is there any other idea on how we can achieve this?

cverna (approved):

Issue #4118, opened 2020-08-31
Graphql : Query specific BuildrootOverride


cverna:
I think we should be able to query builds using the following attributes :

  build nvr (relationship)
  submitter username (relationship User)
  submission_date
  expiration_date

PR #4129, opened 2020-09-21
[WIP]Querying specific buildroot overrides

karmadolkar:
Ref #4118 

### DIFF
--- a/bodhi/server/services/graphql.py
+++ b/bodhi/server/services/graphql.py
@@ -21,7 +21,15 @@ from cornice import Service
 from webob_graphql import serve_graphql_request
 
 from bodhi.server.config import config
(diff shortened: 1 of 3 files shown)


bodhi/server/services/graphql.py
```diff
@@ -60,6 +60,13 @@ class Query(graphene.ObjectType):
         date_approved=graphene.String(), alias=graphene.String(),
         user_id=graphene.Int(), release_name=graphene.String())
 
+    resolve_getBuildrootOverrides = graphene.Field(
```
cverna:
```suggestion
    getBuildrootOverrides = graphene.Field(
```

cverna (commented):
One small thing to change, then we need to add tests and also update the documentation

cverna (approved):

PR #4132, opened 2020-09-26
Add F33 to tests pipeline

mattiaverga:


cverna:
LGTM, I think we can also drop F31 now but that can be done in another PR

cverna (approved):

mattiaverga:
I think mergify will not merge automatically this PR because it doesn't see f33-diff-cover results.
We'll need to manually merge it.

PR #4133, opened 2020-09-27
Grant side-tag owners rights to builds tagged into it

### LINKED ISSUE #4014: sidetag user experience can be improved
It seems that only a creator of a sidetag can submit it as a bodhi request, however, there is no way to assign more than one person to a sidetag management. As result, if sidetag creator does not possess permissions to _all_ components built in the sidetag, nobody will be able to submit the resulting request to bodhi.

I encounter this twice already when doing complex updates of packages with dependent ABI changes.

I'd sugges several changes, not necessarily all falling into bodhi alone:
- allow side-tag creator to add more users to manage the sidetag. Right now they cannot edit sidetag at all, they cannot do anything with it
- allow proven packagers to submit sidetags they do not own
- if a packager has rights to all packages built in sidetag, allow them to submit the sidetag to bodhi

It should not matter who created the sidetag, only a content of the tag should influence who can submit it.

mattiaverga:
This introduces some changes on ACLs validator. Since this is quite important change, here some explanation.

Firstly, I moved the check about user to be admin or packager out of the builds for loop. If the user is in an admin group, just validate they're rights; if the user is not in packager group, just fail. I see no reason to check that inside the for loop.

Then, the main change is when an update comes from a side-tag. Here we no more validate user rights against Pagure commit access on every build, but just check from Koji if the user is the creator of the side-tag. If user owns the side-tag, they can create the update from the side-tag even if there are builds for which they don't have commit access to. This will also prevents users from creating updates from side-tags they don't own (except provenpackagers).

The rationale here is that user A creates a side-tag and update a library into that. Then asks user B to rebuild a package that depends on that library inside the side-tag. When all dependent packages are rebuilt, user A can push a side-tag update which will update also packages they don't have commit access normally.

Fixes #4014 

### DIFF
--- a/bodhi/server/validators.py
+++ b/bodhi/server/validators.py
@@ -396,6 +396,8 @@ def validate_acls(request, **kwargs):
         request.errors.add('cookies', 'user', 'No ACLs for anonymous user')
         return
     user = User.get(request.user.name)
+    user_groups = [group.name for group in user.groups]
+    acl_system = config.get('acl_system')
     committers = []
     groups = []
 
@@ -408,11 +410,10 @@ def validate_acls(request, **kwargs):
     # (either the explicitly listed ones or on the ones associated with the
     # pre-existing update).. and so we'll do some conditional branching below
     # to handle those two scenarios.
-
-    # This can get confusing.
-    # TODO -- we should break these two roles out into two different clearly
-    # defined validator functions like `validate_acls_for_builds` and
-    # `validate_acls_for_update`.
+    #
+    # In case of a side-tag update, we don't validate against the builds,
+    # but we require the submitter to be who created the side-tag.
+    # The 'sidetag_owner' field is set by `validate_from_tag()`.
 
     builds = None
     if 'builds' in request.validated:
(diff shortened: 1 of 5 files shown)


bodhi/server/validators.py
```diff
@@ -427,6 +428,43 @@ def validate_acls(request, **kwargs):
...
+    mandatory_groups = config['mandatory_packager_groups']
+    for mandatory_group in mandatory_groups:
+        if mandatory_group not in user_groups:
+            error = (f'{user.name} is not a member of "{mandatory_group}", which is a '
+                     f'mandatory packager group')
+            request.errors.add('body', 'builds', error)
+            return
+
+    # If we try to create or edit a side-tag update, check if user owns the side-tag
+    if 'from_tag' in request.validated and request.validated.get('from_tag') is not None:
```
cverna: would `if request.validated.get('from_tag') is not None` be enough here ? Why do we need to check that "from_tag" is in request.validated ?

bodhi/tests/server/services/test_updates.py
```diff
@@ -101,7 +101,7 @@ class TestNewUpdate(BasePyTestCase):
     def mock_getTag(cls, tag, *kwargs):
         if tag == 'f17-build-side-7777':
             return {'maven_support': False, 'locked': False, 'name': 'f17-build-side-7777',
-                    'extra': {'sidetag_user': 'dudemcpants', 'sidetag': True},
+                    'extra': {'sidetag_user': 'guest', 'sidetag': True},
```
cverna, line 104: why did we need to change this ? just curious :)

bodhi/server/validators.py
mattiaverga: Yeah, you're right. The Schema provides a `missing=None` directive, so I think it's safe to suppose that parameter is always present in the request.

bodhi/tests/server/services/test_updates.py
mattiaverga: Because of `test_new_rpm_update_from_tag()`.
There the user guest tries to create a new side-tag update and it must succeed. But if the side-tag is owned by dudemcpants, as the mocked getTag replies, the test will fail.

I could modify the test by creating a new 'dudemcpants' user and make it create the update, but that's more work and I'm lazy :-p

cverna: :+1:

cverna (approved):
LGTM

PR #4135, opened 2020-10-03
Allow overriding `critpath.stable_after_days_without_negative_karma` …

mattiaverga:
…based on release status

A different approach than #4108 
It turned out that `critpath.stable_after_days_without_negative_karma` can actually be overriden by release status (and in fact it is in ansible config template), but this is not used by Bodhi.


cverna (approved):
Nice one :+1:

PR #4139, opened 2020-10-13
Update README.rst

laniku:
Add instructions for the Chromium browser as it is also widely used.


laniku:
Strange, a lot of checks failed

PR #4141, opened 2020-10-19
fix typo in bodhi/server/config.py

rayandas:
String literals are implicitly concatenated in a literal iterable definition when they are not separated by a comma.
361: 'value': ['proventesters', `'provenpackager,' 'releng',` 'security_respons', 'packager', 'bodhiadmin'] may lead to undesirable results.


rayandas:
Ah! it's not signed off. Closing.

PR #4140, opened 2020-10-18
Fix integration builds for F33 and Rawhide

mattiaverga:
also fix #4020 

Signed-off-by: Mattia Verga <mattia.verga@tiscali.it>

PR #4150, opened 2020-10-25
Fix to get integration tests run again

mattiaverga:
My previous commit fixed the integration-build, but tests were not running as expected.
This PR should fix them.

### DIFF
--- a/devel/ci/integration/bodhi/Dockerfile-f33
+++ b/devel/ci/integration/bodhi/Dockerfile-f33
@@ -21,7 +21,7 @@ RUN dnf install -y \
 RUN groupadd -r bodhi && \
     useradd  -r -s /sbin/nologin -d /home/bodhi/ -m -c 'Bodhi Server' -g bodhi bodhi
 # Install it
-RUN python3 setup.py install && pip3 install koji
+RUN python3 setup.py bdist_wheel && pip3 install ./dist/*.whl && pip3 install koji
 
 # Configuration
 RUN mkdir -p /etc/bodhi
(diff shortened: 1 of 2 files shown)


AdamSaleh:
I see several `koji: PermissionError: [Errno 13] Permission denied: 'dummy-test-package-gloster-0-1879.fc34.src.rpm'` ... is this the problem of the user we are running the tests as?

How did you test this locally?

mattiaverga:
> I see several `koji: PermissionError: [Errno 13] Permission denied: 'dummy-test-package-gloster-0-1879.fc34.src.rpm'` ... is this the problem of the user we are running the tests as?
> 
> How did you test this locally?

That's a different error and it's not in the scope of this PR.

In #4140 I changed the way bodhi is installed in the integration container from:
```
RUN python3 setup.py build && pip3 install --no-use-pep517 . && pip3 install koji
```
to
```
RUN python3 setup.py install && pip3 install koji
```
because the integration-build steps for F33 and Rawhide were failing (probably due to the fact pip got upgraded to 20.2, while in F32 and lower was still 20.1).
That change fixed the integration-build step, but then I noticed that all integration unit tests were unable to run.

This PR will make the unit tests run again, then we could start looking at why the download test (and a couple of others) are failing (but this happens also in F32 and lower).
Unfortunately we have to handle two problems here, one is Bodhi's setup.py file which generates more than one .egg file, and pip developers don't want to support that, the other is that, as usually, our integration tests get broken very frequently...

AdamSaleh (approved):
Now I understand this is the stage 'At last, the tests are running, we will make them pass afterwards'.

mattiaverga:
BTW if you're willing to look at the download test failure, it happens because actually koji tries to download builds, but that shouldn't happen as the test is supposed replace the koji executable with a mock...

AdamSaleh:
@mattiaverga thanks for the hint, Pingou literaly asked me last week if I could look at Bodhi tests to make them more reliable, so I will start with those :-)

PR #4152, opened 2020-10-30
Few performance fix.

rayandas:
This PR adds few performance fixes.

### DIFF
--- a/bodhi/server/mail.py
+++ b/bodhi/server/mail.py
@@ -314,7 +314,7 @@ def get_template(update: 'Update', use_template: str = 'fedora_errata_template')
         # Add this update's referenced Bugzillas
         i = 1
         info['references'] = ""
-        if len(update.bugs):
+        if update.bugs:
             info['references'] = "References:\n\n"
             parent = True in [bug.parent for bug in update.bugs]
             for bug in update.bugs:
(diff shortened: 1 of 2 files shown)


mattiaverga (changes requested):
Thank you!
I don't think there should be any performance improvement in using implicit `{}` instead of explicit `set()`, but I do prefer the explicit way because I find it to be better readable and more straightforward.
Not sure what are the preferences of other devs here, but I think it would be better if you can remove the two commits on `server/views/generic.py`.

The other changes look good to me.

pypingou:
> I don't think there should be any performance improvement in using implicit {} instead of explicit set(), but I do prefer the explicit way because I find it to be better readable and more straightforward.

:+1: for me, ``{}`` can easily be mixed up with dict, so I tend to use ``set()`` explicitey as well

rayandas:
@mattiaverga @pypingou I revert the changes in `server/views/generic.py`

mattiaverga:
> @mattiaverga @pypingou I revert the changes in `server/views/generic.py`

Instead of pushing a new commit to revert changes, you should edit/remove the previous commits with `git rebase -i`. Look at the git docs and by searching on the internet how to use that command.

rayandas:
@mattiaverga removed the commits.

AdamSaleh:
While I am not sure about the performance, I *really* like the code changes :-)
It might be useful to enable checks for these, i.e. https://pypi.org/project/flake8-comprehensions/ ?

mattiaverga (approved):

PR #4153, opened 2020-11-10
Fixing integration tests

AdamSaleh:
So far:

[x] fixed koji mock in integration test
[x] removed F31
[x] fixed the ipslon mock based on @cverna s advice
[x] improved some of the exception handling
[x] only running integration tests for f33 for now


pypingou:
> I don't think we install from pip anywhere

We don't but testing with pip allows to find out issues with new version of the dependencies before they make it to RPM and the stable Fedora releases.

cverna:
> So far:
> 
> [x] fixed koji mock in integration test
> [x] removed F31
> 
> To do:
> 
> [ ] fix the ipslon mock, last 3 test consistently fail on 404 response from the ipsilon mock, logs suggest problem with badly initialized nonce?

for ipsilon I think you need to change the version of python in the Dockerfile. We manually copy a file and the path contains the python version see https://github.com/fedora-infra/bodhi/commit/43467b47f2c4e35ac06e23c2e70c18ce030ef19b

> 
> Possibly to do:
> 
> [ ] Disable more targets? i.e. I don't think we install from pip anywhere, and we might not even use F32? Need to doublecheck, but leaving just current stable (or current deployment target + current stable) should provide us with enough coverage.

devel/ci/integration/ipsilon/Dockerfile
```diff
@@ -1,7 +1,7 @@
 FROM fedora
```
cverna: we should probably tag this so that we are not surprised when the python version changes. I think `FROM fedora:33` would be better.

cverna:
You will need to sign your commit for the DCO to pass :)

AdamSaleh:
Plan is to have the two targets skipped and then tackle those after this is merged.

* https://github.com/fedora-infra/bodhi/issues/4158
* https://github.com/fedora-infra/bodhi/issues/4163

devel/ci/bodhi-ci
```diff
@@ -1199,7 +1199,10 @@ def _build_jobs_list(
                     release=release, depends_on=build_job, buffer_output=buffer_output),
             ])
             jobs.extend(integration_build_jobs)
-        if command in ('all', 'integration'):
+        # we will be skipping integration job due to issues wth alembic
+        # https://github.com/fedora-infra/bodhi/issues/4158
+        # https://github.com/fedora-infra/bodhi/issues/4163
+        if command in ('all', 'integration') and release != 'pip' and release != 'f32':
```
cverna, line 1205: integration tests are not blocking the merge, so we could in theory just leave them fail :(

devel/ci/integration/greenwave/start.sh
```diff
@@ -1,6 +1,7 @@
 set -m
 
-fedora-messaging consume &
+# the command on the next line is not present, possibly a version mismatch?
+# fedora-messaging consume &
```
cverna: hum that seems wrong, we should try to understand why it is missing

devel/ci/integration/tests/test_bodhi_cli.py
```diff
@@ -478,12 +478,13 @@ def test_updates_download(bodhi_container, db_container):
     # something we can track.
     koji_mock = "#!/bin/sh\necho TESTING CALL $0 $@\n"
     with replace_file(bodhi_container, "/usr/bin/koji", koji_mock):
-        result = _run_cli(bodhi_container, cmd)
+        with replace_file(bodhi_container, "/usr/local/bin/koji", koji_mock):
```
cverna, line 485: do we use 2 different binary for the koji cli ?

cverna:
A few comments but other looks good

devel/ci/integration/tests/test_bodhi_cli.py
AdamSaleh: It seems that was f33/rawhide split? Or f32/f33? Can double-check.

devel/ci/integration/greenwave/start.sh
AdamSaleh: I can try to pin the version of the image.

devel/ci/bodhi-ci
AdamSaleh: No. In my experience, if they are failing and not blocking the merge, these are bad tests. 

Either they are important and should be passing and blocking. Or they are not important and shouldn't be there.

cverna: Agree, I am fine going that way but we should consider the future of these tests, are we going to invest in them or not.

devel/ci/integration/tests/test_bodhi_cli.py
mattiaverga: It's because of https://pagure.io/koji/issue/912
Koji it's installed from RPM package, but it isn't found because it doesn't provide egg info. So Bodhi will install Koji from Pypi and the Koji-Pypi executable is used.

AdamSaleh: So, we keep the workaround? :-)

devel/ci/integration/greenwave/start.sh
AdamSaleh: The image *is* pinned and after I removed it, I noticed integration tests still work. Reverted and will investigate in another issue.

devel/ci/bodhi-ci
AdamSaleh: I am leaning in the way of "lets fix pip, keep just pip, rawhide and current stable and agressively(?) upgrade once new version is stable"

devel/ci/integration/tests/test_bodhi_cli.py
cverna: Thanks @mattiaverga for finding that issue, I knew I saw it in the past :+1: 

Yeah let's keep the workaround, maybe we can add the koji issue as comment to explain why we did that ?

AdamSaleh: Ok, will add Mattia's comment.

cverna (approved):

PR #4126, opened 2020-09-13
Enable builds synch button in side-tag updates edit page for non-rawh…

mattiaverga:
…ide releases

Fixes #4122 

Signed-off-by: Mattia Verga <mattia.verga@tiscali.it>

PR #4159, opened 2020-11-18
Allow editing side-tag updates from cli for current releases

mattiaverga:
Now that side-tag updates are allowed for current releases, allow CLI to edit those updates.

Also include a small addition to #4126 for the webUI form page


cverna (approved):
:+1:

PR #4165, opened 2020-11-26
Release 5.6

cverna:

### DIFF
--- a/bodhi/__init__.py
+++ b/bodhi/__init__.py
@@ -17,7 +17,7 @@
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
 """Fedora's update manager."""
 
-__version__ = "5.5"
+__version__ = "5.6.0"
 
 # This is a regular expression used to match username mentions in comments.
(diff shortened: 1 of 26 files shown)


AdamSaleh (changes requested):
We are hitting rate limits on docker when running integration tests. I'd rather we change these to i.e. quay?

bodhi/__init__.py
```diff
@@ -17,7 +17,7 @@
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
 """Fedora's update manager."""
 
-__version__ = "5.5"
+__version__ = "5.6"
```
mattiaverga: Maybe this should set to `5.6.0` as Kevin had to patch the version number for packaging Bodhi in Koji

cverna: :+1:

mattiaverga (approved):
LGTM

I think we can handle the integration tests fix post release. Postgres official docker image is only on dockerhub, I can't find any feasible replacement in quay (or fedora registry).  It will not be a quick fix I suppose.

PR #4166, opened 2020-11-29
Split bodhi_messages in its own directory structure

mattiaverga:
This is the first step to split Bodhi components in their own packages, starting with bodhi-messages.
I think this approach can solve the problem with the benefit of having all the components still in the same repository, so we can test interoperability when making changes to one component.

I've put bodhi-messages in its own directory structure, with its tests and its own setup.py file. Some setuptools parameters are set in the main (common) bodhi package, so we can update and reuse them for all components (for example, the version tag).

This change is targeted for a Bodhi v6 release, since other packages that import `bodhi.messages` will need to change the import statement to `bodhi_messages`.

Let me know what you think about this. If it's ok I will proceed with splitting the other components.


bodhi/server/services/updates.py
```diff
@@ -20,6 +20,7 @@
 import copy
 import math
 
+from bodhi_messages.schemas import update as update_schemas
```
cverna, line 23: I think this import still should be at line 64. any reasons it was moved up ?

cverna (commented):
Overall this looks really good, thanks for looking at this

bodhi/server/services/updates.py
mattiaverga: Because when it was `bodhi.messages` it was identified as a local package, but now `bodhi_messages` is identified as Third Party.

cverna: Oh makes sense :+1:

cverna:
So I am happy to approve this, which means that we would move to a bodhi v6 on the develop branch. How does that sounds ?

mattiaverga:
Sounds good, but would you mind pushing the current devel branch to a 5.6 bugfix release first?
I think #4173 is quite urgent, we're actually not removing side-tags from koji and I think it can impact the infrastructure resources.

Next I'll try to split bodhi-client in a separate package.

PR #4167, opened 2020-11-30
Fix two reflected XSS vulnerabilities

puiterwijk:
This removes two cases where input from the query parameter could be reflected back,
without auto-escaping, which resulted in a possible XSS attack surface.

Reported-by: @hexdefined
CVE: CVE-2020-15855


cverna (approved):
:+1:

PR #4168, opened 2020-11-30
Remove koji from the python requirements.

cverna:
Currently the koji rpm does not install all the files needed
to be a valid python module see (https://pagure.io/koji/issue/912)
So for the time being we should just drop that requirement

### DIFF
--- a/devel/ci/integration/bodhi/Dockerfile-f32
+++ b/devel/ci/integration/bodhi/Dockerfile-f32
@@ -21,7 +21,7 @@ RUN dnf install -y \
 RUN groupadd -r bodhi && \
     useradd  -r -s /sbin/nologin -d /home/bodhi/ -m -c 'Bodhi Server' -g bodhi bodhi
 # Install it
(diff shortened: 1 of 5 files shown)


mattiaverga:
However, doing so will break Bodhi when freshly installed from Pypi... :-/
Not sure what's the best.

BTW we should make our pip Dockerfile rely on Bodhi requirements.txt file instead of manually installing python3-koji (and other python packages) from RPMs...

cverna:
> However, doing so will break Bodhi when freshly installed from Pypi... :-/
> Not sure what's the best.

Ha yeah :( did not think about this :( :(. I guess we can keep it like that now. I have patched the rpms to remove these dependencies so that we can build it.

> 
> BTW we should make our pip Dockerfile rely on Bodhi requirements.txt file instead of manually installing python3-koji (and other python packages) from RPMs...

:+1:

mattiaverga:
So, I think we can drop this one

PR #4169, opened 2020-12-01
WIP: Moved the reference of postgres image from dockerhub to quay

AdamSaleh:
Moved the reference of postgres image from dockerhub to quay to avoid the rate-limiting.

For this purpose I have created https://quay.io/organization/bodhi-ci and pushed the postgres image there.

Right now I am admin, but as it is an organization, it should be easy to add more people.

And @pypingou should take note I had the great idea of using the packagerbot email address as the organization e-mail.

This should be merged only once we agree that
* this is a good idea
* who should have the admin rights to the organization


mattiaverga:
+1 for me
Just a note: maybe it's better to name the organization as fedora-infra.

cverna (approved):

PR #4148, opened 2020-10-25
Fix DetachedInstanceError in automatic_updates handler

mattiaverga:
Fixes #4147

Signed-off-by: Mattia Verga <mattia.verga@tiscali.it>

PR #4170, opened 2020-12-01
We must pass a list of bug ids to work_on_bugs

mattiaverga:
Another error in the automatic updates handler when calling work_on_bugs celery task: we are passing a list of Bug objects, instead we must pass a list of bug ids.

### DIFF
--- a/bodhi/server/consumers/automatic_updates.py
+++ b/bodhi/server/consumers/automatic_updates.py
@@ -208,6 +208,7 @@ class AutomaticUpdateHandler:
                 log.error(f'Problem obsoleting older updates: {e}')
 
             alias = update.alias
+            buglist = [b.bug_id for b in update.bugs]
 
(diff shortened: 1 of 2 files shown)


AdamSaleh (approved):
Like the small change. On the other hand, is there a way how could we have caught it in tests? Like, had this never worked? Did the BZ api change?

mattiaverga:
> Like the small change. On the other hand, is there a way how could we have caught it in tests? Like, had this never worked? Did the BZ api change?

It never worked. In #4148 I've pushed a fix without realize that also the closing_bugs list was wrong.
We just test the `work_on_bugs` method, but when testing the automatic updates handler we mock the `work_on_bugs_task` celery call, so we never caught this problem.

ryanlerch:
This PR / Issue was included in the 5.7.0 release of Bodhi, and is now deployed in Fedora Staging. Please verify and test that this works as expected.

PR #4176, opened 2021-01-06
Avoid double comment at same time

### LINKED ISSUE #1403: Bodhi is now sending two identical emails when a package meets its karma requirement (and autopush is enabled)
Recently, Bodhi began behaving differently: instead of sending an email listing the details of the final karmic input that caused a package to meet requirements to be pushed to stable, and then a second email indicating the package has been submitted for stable, it now sends two copies of the latter. I'm not sure if this is the result of someone trying to fix #1210, or some other cause, but it seems broken for every package I've commented on of late.

### LINKED ISSUE #3846: When update is autopushed, don't send both 'update can now be pushed' and 'update has been pushed' notifications

mattiaverga:
Bodhi will avoid sending both 'update can now be pushed' and 'update has been pushed' notifications at the same time.
Fixes #3846

### DIFF
--- a/bodhi/server/tasks/approve_testing.py
+++ b/bodhi/server/tasks/approve_testing.py
@@ -80,14 +80,17 @@ def approve_update(update: Update, db: Session):
         log.info(f"{update.alias} has not met testing requirements - bailing")
         return
     log.info(f'{update.alias} now meets testing requirements')
-    # Only send email notification about the update reaching
-    # testing approval on releases composed by bodhi
-    update.comment(
-        db,
-        str(config.get('testing_approval_msg')),
-        author='bodhi',
-        email_notification=update.release.composed_by_bodhi
-    )
+    # If the update is going to be pushed automatically to stable, do not
+    # double comment that the maintainer can push it manually (#3846)
+    if not update.autotime or update.days_in_testing < update.stable_days:
+        # Only send email notification about the update reaching
+        # testing approval on releases composed by bodhi
+        update.comment(
+            db,
+            str(config.get('testing_approval_msg')),
+            author='bodhi',
+            email_notification=update.release.composed_by_bodhi
+        )
     notifications.publish(update_schemas.UpdateRequirementsMetStableV1.from_dict(
         dict(update=update)))
     if update.autotime and update.days_in_testing >= update.stable_days:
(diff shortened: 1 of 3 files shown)


news/3846.bug
```diff
@@ -0,0 +1 @@
+Bodhi will avoid sending both 'update can now be pushed' and 'update has been pushed' notifications at the same time
```
pypingou: + "on updates pushed automatically" ?

bodhi/server/tasks/approve_testing.py
```diff
@@ -80,14 +80,17 @@ def approve_update(update: Update, db: Session):
...
-    # testing approval on releases composed by bodhi
-    update.comment(
-        db,
-        str(config.get('testing_approval_msg')),
-        author='bodhi',
-        email_notification=update.release.composed_by_bodhi
-    )
+    # If the update is going to be pushed automatically to stable, do not
+    # double comment that the maintainer can push it manually (#3846)
+    if not update.autotime or update.days_in_testing < update.stable_days:
```
pypingou, line 85: I understand the first part of the if statement, the second part surprises me though, especially because we were not checking this here before. Isn't it being checked earlier on? Maybe as part of "meets_testing_requirements"?

mattiaverga: yeah, it is quite tricky...
There can be a case where autotime is on, the update has not yet been enough time in testing, but it meets testing requirements.

The `meets_testing_requirements` evaluates to True 
```if self.karma >= self.release.critpath_min_karma```
so for current releases, an update can be manually pushed to stable after reaching +2 karma, regardless of how much time has been in testing, even if autopush is set to automatically push after +3 karma...

mattiaverga:
@pypingou have all your comments been addressed? Can this PR be merged?

pypingou (approved):
All my questions have been answered, thanks!

PR #4181, opened 2021-01-16
Use Pagure's hascommit endpoint API to validate ACLs

mattiaverga:
This will allow collaborators which have access only to specific branches to create/edit updates for the releases where they have access to. Fixes https://pagure.io/fedora-infrastructure/issue/9510

Note that this requires the new Pagure endpoint API `hascommit` (https://pagure.io/pagure/pull-request/5087) and we'll need to wait for that to land in production.

Please also check carefully how we handle Pagure branch / release.branch association. As far as I know, we have:

- a 1:1 association for RPMs
- a 1:1 association for Containers (at least, this is what I observed)
- Flatpaks always use ~~master~~ **stable** branch (which is configurable in my PR)
- Modules don't have any association, so we just use master branch here (configurable) to mimic the old validation mechanism.

Is that right?
Also, https://pagure.io/fedora-infrastructure/issue/9566 needs to be fixed.

### DIFF
--- a/bodhi/server/config.py
+++ b/bodhi/server/config.py
@@ -51,6 +51,37 @@ def get_configfile() -> typing.Optional[str]:
     return configfile
 
 
+def _generate_dict_validator(value: str) -> dict:
+    """Return a dict version of value.
+
+    This function accept a string with comma separated tuples in the form `key:value`
+    and ensure it can be correctly imported as a dict. It will return a dict.
+    If it cannot do that, it will raise ValueError.
+
+    Args:
+        value: The value to be validated as a dict.
+    Returns:
+        The dict interpretation of value.
+    Raises:
+        ValueError: If value cannot be interpreted as a dict.
+    """
+    if isinstance(value, str):
+        values_list = [idx.strip() for idx in value.split(',') if idx.strip()]
+        dictvalue = dict()
+        for v in values_list:
+            try:
+                k, v = v.split(':')
+                dictvalue[k.strip()] = v.strip()
+            except Exception:
+                raise ValueError(f'"{value}" cannot be interpreted as a dict.')
+        value = dictvalue
+
+    if not isinstance(value, dict):
+        raise ValueError(f'"{value}" cannot be interpreted as a dict.')
+
+    return value
+
+
 def _generate_list_validator(
         splitter: str = ' ', validator: typing.Callable[[typing.Any], typing.Any] = str) \
         -> typing.Callable[[typing.Union[str, typing.List]], typing.Any]:
@@ -457,8 +488,14 @@ class BodhiConfig(dict):
         'openid_template': {
             'value': '{username}.id.fedoraproject.org',
             'validator': str},
-        'pagure_flatpak_namespace': {
-            'value': 'modules',
+        'pagure_namespaces': {
+            'value': ('rpm:rpms, module:modules, container:container, flatpak:flatpaks'),
+            'validator': _generate_dict_validator},
+        'pagure_flatpak_main_branch': {
+            'value': 'stable',
+            'validator': str},
+        'pagure_module_main_branch': {
+            'value': 'master',
             'validator': str},
         'pagure_url': {
             'value': 'https://src.fedoraproject.org/pagure/',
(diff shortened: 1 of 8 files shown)


bodhi/server/config.py
```diff
@@ -460,6 +460,12 @@ class BodhiConfig(dict):
         'pagure_flatpak_namespace': {
             'value': 'modules',
             'validator': str},
+        'pagure_flatpak_main_branch': {
+            'value': 'master',
+            'validator': str},
+        'pagure_module_main_branch': {
+            'value': 'master',
```
pypingou, line 498: By the time this code lands in production, the default branch should be "main", should we adjust the default value already?

bodhi/server/models.py
```diff
@@ -1154,6 +1154,36 @@ def get_pkg_committers_from_pagure(self):
...
+            namespace = config.get('pagure_flatpak_namespace')
+            # branchname override for all releases
+            branchname = config.get('pagure_flatpak_main_branch')
+        elif self.type == ContentType.module:
+            namespace = self.type.name + 's'
+            # There's no direct map between stream and branch names so override branchname
+            # for the moment
+            branchname = config.get('pagure_module_main_branch')
+        else:
+            namespace = self.type.name + 's'
```
pypingou: Food for thoughts, we could do something like: https://pagure.io/fedora-infra/toddlers/blob/main/f/toddlers.toml.example#_135 (ie: a dict that maps the ContentType.name to the dist-git names), this way if there is ever another namespace rename in the future, we can simply adjust the configuration file without having to touch the code.

We could have something like: ``namespace = namespace_dict[self.type.name]`` where ``namespace_dict`` is retrieved from the config.

pypingou:
Couple of comments but nothing blocking per say

bodhi/server/models.py
mattiaverga: yup, good idea, I'll update the PR

bodhi/server/config.py
mattiaverga: I've read that this has been reverted back to `master`, if I've not misunderstood

pypingou: For modules yes, for flatpak it's "stable', for container and rpm it's "rawhide"

bodhi/server/config.py
```diff
@@ -457,8 +485,14 @@ class BodhiConfig(dict):
         'openid_template': {
             'value': '{username}.id.fedoraproject.org',
             'validator': str},
-        'pagure_flatpak_namespace': {
-            'value': 'modules',
+        'pagure_namespaces': {
+            'value': ('rpm:rpms, module:modules, container:container, flatpak:modules'),
```
pypingou: ``flatpak:modules``?

bodhi/server/models.py
```diff
@@ -1154,6 +1146,28 @@ def get_pkg_committers_from_pagure(self):
...
+        """
+        pagure_url = config.get('pagure_url')
+        namespace = config.get('pagure_namespaces')[self.type.name]
+        if self.type == ContentType.flatpak:
+            # branchname override for all releases
+            branchname = config.get('pagure_flatpak_main_branch')
+        elif self.type == ContentType.module:
+            # There's no direct map between stream and branch names so just
+            # use main branch
+            branchname = config.get('pagure_module_main_branch')
```
pypingou, line 1166: Should we document here why we override the default branch for modules and flatpaks and not the others?

pypingou:
Couple of comments but we're getting into the nitpicks :)

bodhi/server/config.py
mattiaverga: Yep, it is currently set as `modules` by default (see the two removed lines above), then redefined to `flatpak` in ansible config template.
I think we can now make it to be `flatpak` by default, I'm going to change this line.

bodhi/server/models.py
mattiaverga: Isn't the comment above the override enough?

pypingou: aouch, ENEEDGLASSES :(

pypingou (approved):
All good for me :)

mattiaverga:
Thanks, so I'll leave the WIP label to prevent this from being merged until https://pagure.io/fedora-infrastructure/issue/9566 is worked out.

nirik:
That downstream ticket should be fixed now (aside from docs).

ryanlerch:
This PR / Issue was included in the 5.7.0 release of Bodhi, and is now deployed in Fedora Staging. Please verify and test that this works as expected.

PR #4183, opened 2021-01-25
Require diff-cover<4.2.0 in pip dockerfile

mattiaverga:
It should hopefully fix pip tests.


mattiaverga:
This is not needed anymore, dependency chain has been fixed in Pypi

PR #4186, opened 2021-02-27
Clear request status when release goes EOL

### LINKED ISSUE #4039: Request not cleared on updates when release goes EOL
When a release is archived, its updates statuses are set to "Obsolete", but we should also clear the Update request:
https://bodhi.fedoraproject.org/updates/FEDORA-2020-2855a8a590

andreamisuraca:
Hi @mattiaverga, is this what you meant with issue https://github.com/fedora-infra/bodhi/issues/4039?

### DIFF
--- a/bodhi/server/services/releases.py
+++ b/bodhi/server/services/releases.py
@@ -386,6 +386,7 @@ def save_release(request):
                     ).all()
                     for u in updates:
                         u.status = UpdateStatus.obsolete
+                        u.request = None
                         u.comment(
                             request.db,
                             'This update is marked obsolete because '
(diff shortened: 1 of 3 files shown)


mattiaverga:
Hi @andreamisuraca,
we already manually set the UpdateStatus to Obsolete, so we need to clear any existent request on the Update (after the Release is archived, no request will be processed for those updates). So I think setting it to None should make it.

andreamisuraca:
> Hi @andreamisuraca,
> we already manually set the UpdateStatus to Obsolete, so we need to clear any existent request on the Update (after the Release is archived, no request will be processed for those updates). So I think setting it to None should make it.

Hi @mattiaverga, thanks for the hint, could you please have a look at it now?

mattiaverga (approved):
LGTM

Pip and integraton tests failures are expected (sigh).

PR #4190, opened 2021-03-11
Avoid daily increment of <updated_date>

### LINKED ISSUE #4189: updated_date on updateinfo is changed constantly, without any changes to the errata
Referred here by: https://pagure.io/releng/issue/10051#comment-719399

It appears that the "updated_date" fields on errata / advisories / updateinfo is updated daily, even without any changes to the actual metadata or package lists. This causes problems for repository management tools such as Pulp, which sees that the "updated_date" has changed, and that therefore the repository has changed, and that therefore the repository metadata needs to be regenerated each time (unnecessarily).

"updated_date" should preferably only be changed when the advisory itself has been updated. Changing it daily without changing the advisory, devalues its usefulness as a component of the metadata.

It would be very difficult to simply ignore "updated_date" because we semi-frequently encounter errata that were updated without changing the "version" metadata appropriately - and it becomes deeply problematic to work with updateinfo / errata / advisories if neither of these fields can be relied upon.

This seems to be true of both EPEL and Fedora.  This issue was reported to us by a user and filed in our issue here: https://pulp.plan.io/issues/8229

EPEL is more of a concern to us simply due to the number of Pulp / Katello / Satellite users who want to manage EPEL content.

dralley:
When writing out the "updated_date" field in the metadata, if the update
has never been modified in Bodhi, rather than using utcnow(), fall back
to using issued_date. This avoids constant churn of the "updated_date".

closes #4189

### DIFF
--- a/bodhi/server/metadata.py
+++ b/bodhi/server/metadata.py
@@ -238,7 +238,8 @@ class UpdateInfoMetadata(object):
         if update.date_modified:
             rec.updated_date = update.date_modified
         else:
-            rec.updated_date = datetime.utcnow()
+            # If there is no date_modified, use date_submitted (#4189)
+            rec.updated_date = update.date_submitted
 
         col = cr.UpdateCollection()
         col.name = update.release.long_name.encode('utf-8')
--- a/bodhi/server/metadata.py
+++ b/bodhi/server/metadata.py
@@ -16,7 +16,6 @@
 # with this program; if not, write to the Free Software Foundation, Inc.,
 # 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 """Create metadata files when composing repositories."""
-from datetime import datetime
 import logging
 import os
 import shelve
@@ -233,12 +232,12 @@ class UpdateInfoMetadata(object):
             # Sometimes we only set the date_pushed after it's pushed out, however,
             # it seems that Satellite does not like update entries without issued_date.
             # Since we know that we are pushing it now, and the next push will get the data
-            # correctly, let's just insert utcnow().
-            rec.issued_date = datetime.utcnow()
+            # correctly, let's just insert "date submitted".
+            rec.issued_date = update.date_submitted
         if update.date_modified:
             rec.updated_date = update.date_modified
         else:
-            # If there is no date_modified, use date_submitted (#4189)
+            # Likewise, if there is no date_modified, use date_submitted
             rec.updated_date = update.date_submitted
 
         col = cr.UpdateCollection()
--- a/bodhi/tests/server/test_metadata.py
+++ b/bodhi/tests/server/test_metadata.py
@@ -16,7 +16,7 @@
 # with this program; if not, write to the Free Software Foundation, Inc.,
 # 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 
-from datetime import datetime, timedelta
+from datetime import datetime
 from hashlib import sha256
 from os.path import join, exists, basename
 from unittest import mock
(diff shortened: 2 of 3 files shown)


dralley:
This fixes some strange behavior described in the linked issue, where most (but not all) of the UpdateInfo in Fedora and CentOS have a constantly-changing <updated_date>. 

This is because ```rec.updated_date``` was set to ```datetime.utcnow()``` if  ```update.date_modified``` is ```null``` - and it is always ```null``` unless the update has been edited in Bodhi.

mattiaverga (changes requested):
It seems odd the the relevant test doesn't fail after your changes:
https://github.com/fedora-infra/bodhi/blob/31bfc2ee62b8e96f2eaa8f4553f502caf28ed3ac/bodhi/tests/server/test_metadata.py#L127-L142

Can you have a look at it and change it in accordance with the new behavior?
Also, I'm wondering if we shouldn't set `rec.updated_date` to `update_date_submitted` instead of `date_pushed`... what do you think?

Don't forget to add a new file as explained [here](https://bodhi.fedoraproject.org/docs/developer/#release-notes)

dralley:
@mattiaverga Thanks for taking a look!

If ```date_submitted``` can always be expected to be there (unlike ```date_pushed``` as per the inline comment here [0]), then it might make more sense.  However, I'm neither a regular user nor developer of Bodhi so my opinion is not worth much!

Regarding the test passing, looking at the code in [0], it a actually does make sense that it would pass since the test case doesn't initialize ```date_pushed```, which then gets filled in as ```utcnow()```.  And then due to my patch, ```updated_date``` gets the same value - so they both end up filled in by ```utcnow()```.

[0] https://github.com/fedora-infra/bodhi/pull/4190/files#diff-d8236f709f77c2189f8e6ebc696a39e7aca1a9e6a5522300a7e9adf558deb62cL230-L242

dralley:
@mattiaverga Will wait on the ```date_submitted``` decision before messing with the tests

mattiaverga:
Just my personal opinion, but I think it would make sense to use `date_submitted`. I've double checked and that is always set.

dralley:
@mattiaverga Regarding the notice here: https://github.com/fedora-infra/bodhi/blob/16b126a04eb69a159aa8ee885b61661ef3467960/bodhi/server/metadata.py#L233-L236

Do you think perhaps we should do the same with ```issued_date``` if ```date_pushed``` is not set?  It would be more deterministic to use "real" dates as much as possible.

dralley:
I went ahead and added it, I can squash the commits and update the message if you agree

mattiaverga (approved):

mattiaverga:
Thanks for looking at this!

dralley:
@mattiaverga Thank you!

Do you happen to know when this will be deployed live?  Or if not, is there a way I could be notified when it does?  We would like to close the issue in our downstream issue tracker.

mattiaverga:
> Do you happen to know when this will be deployed live? Or if not, is there a way I could be notified when it does? We would like to close the issue in our downstream issue tracker.

I don't know. There has been no "project leader" here lately, I'm just trying to move things forward.

I'll ask Fedora infrastructure folks if it's ok for me to publish a new Bodhi release with the latest changes, then we will need someone to deploy it on prod. But I think the deployment needs to wait the end of freeze period (not sure about that).

I'll let you know.

mattiaverga:
@dralley Bodhi 5.7.0 has been deployed yesterday in prod. Let us know any further problem.

dralley:
@mattiaverga Thank you!

PR #4199, opened 2021-03-30
Require setuptools, bodhi/server/__init__.py imports from pkg_resources

hroncok:


ryanlerch:
@hroncok this one is passing all checks now other than DCO.

ryanlerch (approved):

hroncok:
I've done the change via web UI and there is no way to edit the commit message to include the incantation. Feel free to do that when merging via the squash option. (I consider the requirement unnecessary, but if that is the only way to merge my PR, do it.)

ryanlerch:
Yeah, i'm not comfortable doing it this way. Probably the easiest way around here is to close this PR, and open another PR via the edit file github functionality, and add "Signed-off-by: YOUR NAME \<YOURNAME@EXAMPLE.COM\>" as the first line of the commit body.

hroncok:
https://github.com/fedora-infra/bodhi/pull/4205

PR #4200, opened 2021-04-03
Ensure updates with stable_comment are pushed to stable automatically

### LINKED ISSUE #4042: Update was not auto-pushed to stable after 7 days
[This update](https://bodhi.fedoraproject.org/updates/FEDORA-2020-189a42d65a) should have been pushed to stable after 7 days, but wasn't. I'm not sure if this was a one time issue.

It received +2 karma, the threshold was +3. So it received feedback, but not enough to reach the threshold. Maybe it's related to that?

mattiaverga:
The `approve_testing` task was set to ignore updates which already had
stable_comment.
This was not correct and prevented some updates from being pushed to
stable when they reached the stable_days threshold.

As an example, consider an update for which the associated release has a
mandatory_days_in testing set to 7 and the update is created with a
stable_days threshold of 10.
After 7 days in testing, the update gets the stable_comment, since the
maintainer can now push it to stable.
The autopush is performed after 3 more days, but this didn't happen
because of the approve_testing task ignoring all updates with
stable_comment.

Fix #4042 

### DIFF
--- a/bodhi/server/models.py
+++ b/bodhi/server/models.py
@@ -3820,8 +3820,7 @@ class Update(Base):
         """
         for comment in self.comments_since_karma_reset:
             if comment.user.name == 'bodhi' and \
-               comment.text.startswith('This update ') and \
-               'can be pushed to stable now if the maintainer wishes' in comment.text:
+               comment.text == config.get('testing_approval_msg'):
                 return True
         return False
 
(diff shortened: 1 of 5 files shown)


pypingou:
I would recommend describing more in the commit message how we ended up in this situation, but I leave this up to you and if you don't want I'll approve anyway

mattiaverga:
I tried to better describe the case, see if it fits now.
I've also added a section in the docs about the differences between the mandatory thresholds and the user defined thresholds for autopush, I saw a lot of confusion in the past about that...

pypingou (approved):
Thanks for the commit message, I understand the issue addressed here much better :)

:+1: on the doc update as well

PR #4204, opened 2021-04-10
Updates check for expiration_date in order to resolve inconsistent refusal in case of maximum duration

### LINKED ISSUE #4182: inconsistent refusal of bodhi override with maximum duration
From https://bugzilla.redhat.com/show_bug.cgi?id=1761691:

```console
$ fedpkg override create --duration=31 python-pytest-steps-1.6.2-1.fc31
Expiration date may not be longer than 31
Could not execute create_buildroot_override: Cannot create override.
```
Obviously the user is asking for a duration of 31 days which is not "longer than 31".

fedpkg-1.37-5.el8.1.noarch

Creating the update with --duration=30 works.

Looking at the code
https://github.com/fedora-infra/bodhi/blob/b81c9fb8bfc5917fa36944923c8795d2f29214ee/bodhi/server/validators.py#L1207-L1216
it's a toctou issue: `expiration_date` is calculated first, and then when `now + timedelta(days=config.get('buildroot_limit'))` is calculated the result is tiny bit off, so the check fails.

samyak-jn:
Hi :smile:,

This is my first time contributing to bodhi, please correct me if I'm wrong.
This PR is in regards to resolve issue #4182, the limit that is compared with expiration_date is updated by `timedelta(days=days+1)` so that it will correctly compare, as well the old error message that was creating confusion is updated.

### DIFF
--- a/bodhi/server/validators.py
+++ b/bodhi/server/validators.py
@@ -1200,12 +1200,12 @@ def validate_expiration_date(request, **kwargs):
         request (pyramid.request.Request): The current request.
         kwargs (dict): The kwargs of the related service definition. Unused.
     """
-    expiration_date = request.validated.get('expiration_date')
+    expiration_date = request.validated.get('expiration_date').date()
 
     if expiration_date is None:
         return
 
-    now = datetime.utcnow()
+    now = datetime.utcnow().date()
 
     if expiration_date <= now:
         request.errors.add('body', 'expiration_date',
@@ -1218,8 +1218,6 @@ def validate_expiration_date(request, **kwargs):
         request.errors.add('body', 'expiration_date',
                            'Expiration date may not be longer than %i' % days)
         return
-
-    request.validated['expiration_date'] = expiration_date
 
 
 @postschema_validator
--- a/bodhi/server/validators.py
+++ b/bodhi/server/validators.py
@@ -1200,11 +1200,12 @@ def validate_expiration_date(request, **kwargs):
         request (pyramid.request.Request): The current request.
         kwargs (dict): The kwargs of the related service definition. Unused.
     """
-    expiration_date = request.validated.get('expiration_date').date()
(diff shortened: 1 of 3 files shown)


mattiaverga:
What about rewriting the validator so that it doesn't care of hours/minutes/etc. ?
I mean:

- convert `expiration_date` to a datetime.date object
- use a datetime.date object for `now` variable

Everything else can remain the same. Just remove the final `request.validated['expiration_date'] = expiration_date` since `expiration_date` in request object is expected to be datetime.datetime.
Also, the message `Expiration date may not be longer than` is fine, since it implies "less or equal" than the limit.

mattiaverga:
Oh, I almost forgot: we lack a test for expiration_date higher than limit, please add it (see https://github.com/fedora-infra/bodhi/blob/348c3072a5c83106fad573bc35d168b27d2f7b34/bodhi/tests/server/test_validators.py#L401)

Better to add two test, one when expiration_date is equal to limit and another one for expiration_date higher than limit ;-)

samyak-jn:
I apologize for my foolishness above :rocket: 
As discussed, have converted the `expiration_date` and `limit` to `datetime.date` object (will now only compare date!).
Have added two more test functions `higherthanLimit` and `equaltoLimit` for the same!

mattiaverga:
Looks great, there's only a minor failure on flake8 checks:
```
21:59:35  ./bodhi/tests/server/test_validators.py:454:1: E302 expected 2 blank lines, found 1
21:59:35  class TestValidateOverrideNotes(BasePyTestCase):
21:59:35  ^
```

mattiaverga:
Oh, almost forgot: please add a news fragment as described [here](https://bodhi.fedoraproject.org/docs/developer/#release-notes)

samyak-jn:
Thank you @mattiaverga for all your help!  :heart:

mattiaverga (approved):
Thanks!

PR #4210, opened 2021-04-24
Fix validator issue which prevents a side-tag update owner to edit the…

### LINKED ISSUE #4209: Unable to edit update from side-tag: no commit access to one of the builds.
I added a build that I don't have commit access to into [FEDORA-2021-0cad58d4f4](https://bodhi.fedoraproject.org/updates/FEDORA-2021-0cad58d4f4) (f34) and after that I lost permissions to edit the update. I'm the owner of both the side-tag and the bodhi update.

The error message is
```
403 Forbidden
alebastr does not have commit access to cage
alebastr does not have commit access to labwc
```

My guess is that #4133 did not cover permission check for modifying existing update.

mattiaverga:
…ir update after adding a build for which they don't have commit access

When the request object only carries the 'update' without 'builds' or 'side-tag', the validator fall back to check single builds instead of validating the side-tag.
Fixes #4209 


AdamSaleh (approved):
Small change, test passes, approved!

PR #4222, opened 2021-05-21
Add "N" to the version model

nirik:
Fedora's bodhi just added a "EPEL-8N" release that has
epel8-next-updates and epel8-next-updates-testing. This is to build
against CentOS stream in those cases where normal EPEL packages are no
longer compatible due to stream changes. The bodhi module doesn't
understand the 'N' here and can't figure out the version. If we add it
into the model it's fine. :)

Signed-off-by: Kevin Fenzi <kevin@scrye.com>

### DIFF
--- a/bodhi/server/models.py
+++ b/bodhi/server/models.py
@@ -830,7 +830,7 @@ class Release(Base):
     composed_by_bodhi = Column(Boolean, default=True)
     create_automatic_updates = Column(Boolean, default=False)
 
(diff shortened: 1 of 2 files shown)


mattiaverga (changes requested):
LGTM

Just let's announce this in release notes with a news fragment in /news, shall we?
https://bodhi.fedoraproject.org/docs/developer/#release-notes

mattiaverga (approved):
I've added the news fragment, so I think we can now merge this small change.

PR #4226, opened 2021-05-30
Trying to fix pip CI tests

mattiaverga:
The PR aims to get pip CI tests pass again.

- celery currently requires Click<8.0, so we should set the same in Bodhi
- cornice_sphinx doesn't behave well with Sphinx 4, so let's pin Sphinx<4.0 in pip dockerfile. This is a temporary solution, I think we should move away from cornice_sphinx since it's not updated since years
- finally, update pip dockerfile base to F34

### DIFF
--- a/devel/ci/Dockerfile-pip
+++ b/devel/ci/Dockerfile-pip
@@ -1,4 +1,4 @@
-FROM registry.fedoraproject.org/fedora:33
+FROM registry.fedoraproject.org/fedora:34
 LABEL maintainer="Randy Barlow <bowlofeggs@fedoraproject.org>"
 
 RUN dnf install -y \
@@ -29,7 +29,7 @@ RUN pip-3 install \
     alembic \
     cornice_sphinx \
     diff-cover \
-    "docutils<0.17" \
+    docutils \
     flake8 \
     flake8-import-order \
     responses \
@@ -38,8 +38,7 @@ RUN pip-3 install \
     pytest-cov \
     "sphinx<4.0" \
     sqlalchemy_schemadisplay \
-# We can remove the following when python3-requests 2.25 will be available in Koji
-    "urllib3<1.26" \
+    urllib3 \
     webtest
 
 # Fake pungi being installed so we can avoid it and all its dependencies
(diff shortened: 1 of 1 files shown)


AdamWill:
note, I think the issue causing celery to require click < 8.0 has [been resolved](https://github.com/celery/celery/issues/6511#issuecomment-859967282), so celery upstream could also drop that constraint - I asked them to.

mattiaverga:
> note, I think the issue causing celery to require click < 8.0 has [been resolved](https://github.com/celery/celery/issues/6511#issuecomment-859967282), so celery upstream could also drop that constraint - I asked them to.

yep, it seems it now works fine with Click 8, so I've removed the related commit.
We still need to pin Sphinx<4, not sure why since I've seen Koji builds in Rawhide with Sphinx 4 runs fine...

AdamWill:
So some of the errors here are because we're using an old stable graphene (2.x) on Python 3.10. I'm not familiar enough with the CI environment to know exactly why or where/how to change that, but that's the issue. If we were using a more recent graphene - latest is 3.0.0b7, Rawhide has 3.0.0b6 which would be new enough - it wouldn't error out.

mattiaverga:
> So some of the errors here are because we're using an old stable graphene (2.x) on Python 3.10. I'm not familiar enough with the CI environment to know exactly why or where/how to change that, but that's the issue. If we were using a more recent graphene - latest is 3.0.0b7, Rawhide has 3.0.0b6 which would be new enough - it wouldn't error out.

The problem is graphene-sqlalchemy has never been updated to v3 stack and everything about graphene/graphql is nearly dead upstream.

We may want to drop all graphQL stuff in Bodhi.

lgriffin:
@StephenCoady can you take a look at this tomorrow as I believe you looked at the GraphQL side of it and @Zlopez can you mark this in as a priority for Infra and Releng as it is blocking https://github.com/fedora-infra/bodhi/pull/4221 which is now causing some community issues

Zlopez:
@lgriffin I will bring this as priority on your next infra & releng meeting

AdamSaleh:
This has been incorporated to https://github.com/fedora-infra/bodhi/pull/4244 and merged

PR #4228, opened 2021-06-09
Delete EOL sidetags

mattiaverga:
We currently rely on Koji auto-purge-when-empty for side-tags to be deleted.
This doesn't always happens: when an update is pushed to stable we untag only the builds that were pushed through the update, but there can be other builds lying in the side-tag, thus it will never be removed.

These two commits will make sure we remove the side-tag from Koji db when the associated update is pushed stable. Or, at least, we will try to do that (if a call in koji.multiCall errors out for any reason, the process doesn't crash).


cverna (approved):
LGTM

Issue #4240, opened 2021-07-26
Add/expose release EOL to bodhi API


LenkaSeg:
It would be great if bodhi API would contain end-of-life date of releases.

LenkaSeg:
PR #4241

LenkaSeg:
Closed in PR #4241

PR #4241, opened 2021-07-28
Add End of Life (eol) field in bodhi/releases

LenkaSeg:
Signed-off-by: Lenka Segura <lsegura@redhat.com>


LenkaSeg:
Fixes issue #4240

Zlopez:
How this field will be filled?

LenkaSeg:
It should be a date in format YYYY-MM-DD.
Example: 2021-05-25

Ah, you mean how. Is there some automation to which it should be added or it's filled manually for the releases?

cverna:
We need the bodhi client to be able to set that date. A few pointers

cli options for the release command --> https://github.com/fedora-infra/bodhi/blob/707998a3d45eca8b12b4afe137f3899e1f6cb193/bodhi/client/__init__.py#L171

cli command that `save` a new release --> https://github.com/fedora-infra/bodhi/blob/707998a3d45eca8b12b4afe137f3899e1f6cb193/bodhi/client/__init__.py#L1281

backend  service that save the release --> https://github.com/fedora-infra/bodhi/blob/707998a3d45eca8b12b4afe137f3899e1f6cb193/bodhi/server/services/releases.py#L348

Might be also a good idea to validate the EOL date using a validator (check format, that the date is in the future, ....) and example here --> https://github.com/fedora-infra/bodhi/blob/707998a3d45eca8b12b4afe137f3899e1f6cb193/bodhi/server/validators.py#L1195

LenkaSeg:
Added a flag `--eol` as a release option for bodhi client and a test confirming the behavior. I was able to test it only like this. I was trying to test it on local instance with `bodhi releases edit --url "http://<the_vagrant_address>:6543" --name F34 (--user <my_FAS_username>, --password <my_FAS_password> )`, but I was not able to set the permissions to be able to edit. Is there a way to do that? 
I tried uncommenting the `acl_dummy_comitter` and use my username in `development.ini` file, as suggested there, but it had not worked (even after `vagrant destroy` and `vagrant up`. Out of desperation I tried also the username and password I found in tests, but obviously it did not work either)

I got slightly lost trying to figure out the validator, but I'll still work on that.

I don't understand what needs to be added in the `def save` and `def save_release`. 

Update: tested locally with `bodhi relases edit --name F34 --eol 2016-06-14` and it works.

LenkaSeg:
Added label [WIP], cause there's a bug.

When running `bodhi release edit --eol 2016-06-14` the eol gets nicely updated in the local instance.
When running the edit again, for example with correction: `bodhi release edit --eol 2016-06-15` in get this error:

`datetime.date(2016, 6, 14) is not JSON serializable` with the first date. I'm still struggling with it.

bodhi/client/__init__.py
```diff
@@ -1224,6 +1232,8 @@ def edit_release(user: str, password: str, url: str, debug: bool, composed_by_bo
     data['csrf_token'] = csrf
     data['composed_by_bodhi'] = composed_by_bodhi
 
+#    if eol is not None:
+#        data['eol'] = str(eol)
```
AdamSaleh: Why did we need this? If we don't anymore, let's remove it.

bodhi/server/validators.py
```diff
@@ -1214,6 +1215,27 @@ def _validate_override_build(request, nvr, db):
...
+    Args:
+        request (pyramid.request.Request): The current request.
+        kwargs (dict): The kwargs of the related service definition. Unused.
+    """
+    eol_date = request.validated.get('eol')
+    if eol_date is None:
+        return
+
+    eol_date = str(eol_date)
+    if not re.match(r"20[0-9][0-9]-((0[1-9])|(1[0-2]))-(0[1-9]|[1-2][0-9]|3[0-1])", eol_date):
```
AdamSaleh: Could be better to attempt to parse it with `isoformat` we use in models.py?

bodhi/tests/server/services/test_releases.py
```diff
@@ -286,7 +293,7 @@ def test_save_release_exception_handler(self, info):
                  "pending_signing_tag": "f42-updates-testing-signing",
                  "pending_testing_tag": "f42-updates-testing-pending",
                  "override_tag": "f42-override",
-                 "csrf_token": self.get_csrf_token(),
+                 "csrf_token": self.get_csrf_token(), "eol": date(2016, 6, 16),
```
AdamSaleh, line 320: We probably should have one more test that verifies `POST /releases/` still works even without the `eol` key present in the dictionary.

AdamSaleh (changes requested):
Looks good.

AdamSaleh (approved):

bodhi/client/__init__.py
LenkaSeg: That is forgotten, removing it.

bodhi/server/validators.py
LenkaSeg: This was meant as an additional date limitation, that for example date of 3021-12-12 or 1999-12-12 would not pass, because for releases it would probably not be relevant. I remember that at some point it was discussed if to allow dates in the past. I didn't add that check for the moment, in case we would want to insert eol dates for the past releases.

bodhi/tests/server/services/test_releases.py
LenkaSeg: Ok, will add it!

bodhi/server/validators.py
LenkaSeg: I removed the regex and wanted to at least make sure that eol date will not be before release date. But I see the release date is not there :) Would it be nice to have the release date for releases?

PR #4243, opened 2021-08-05
Update to f34

humaton:
Update vagrant env and Dockerfiles to use f34.


AdamSaleh (changes requested):
In addition we should change .mergify.yml and  devel/ci/cico.pipeline so that the CI expects the the right environments to run, i.e. as we did in f33 update: https://github.com/fedora-infra/bodhi/pull/4132/files

AdamSaleh:
Closing in favour of #4244

PR #4255, opened 2021-09-13
Set a delta when quering datagrepper

mattiaverga:
Based on recent datagrepper development, we should set a `delta` parameter when we query for bodhi-related user activity, so that we will not use datagrepper default (when the new datagrepper will be deployed in prod) and we don't stress datagrepper too much (I've set a delta of 30 days).
I've also modified the script to show a useful message if no recent activity is discovered and I've set a 5 seconds timout on the query.


AdamSaleh (approved):
Looks good, but we probably need to test this in staging.

PR #4260, opened 2021-09-30
Updating 5.7 branch for 5.7.1 release

AdamSaleh:
Going through the release process in https://github.com/fedora-infra/bodhi/blob/develop/docs/developer/releases.rst


mattiaverga:
I think this should have been merged in `develop` and then backported to the 5.7 branch, otherwise the release notes in develop will not be updated correctly.
If I remember correctly, in the past all the PRs were merged in develop, then backported when appropriate with the use of the `x.x-backport` tag, which will cause mergify to cherry-pick the changes.

PR #4261, opened 2021-10-02
Correct tagging builds when editing side-tag update

mattiaverga:
On creating a side-tag update for a release composed by Bodhi (current or branched), the tag history of all builds in the update is something like:

 - upon creation: f35-build-side-xxx + f35-updates-candidate + f35-pending-signing
 - autosign signs the build: f35-build-side-xxx + f35-updates-candidate + f35-testing-pending
 - bodhi compose the testing repo: f35-build-side-xxx + f35-updates-testing

Now, if the build list is updated, Bodhi unpushed the update, removes tags from builds and tries to tag the new builds in the -pending-signing related to the side-tag, which doesn't exist.
So, the builds will end in having `f35-build-side-xxx + f35-updates-candidate`, which means new builds are not marked as signed and the update gets stuck in pending state, until `check_signed_build` task runs.

This PR should solve this issue by:

1. Use the build-side-pending-signing only for Rawhide. Current and branched release will use pending-signing for the release
2. mark all builds as pending-signed, not only the newer ones. This isn't really necessary, but will cause the older builds to be marked testing-pending to be consistent with the newer builds. Since they're already signed it should not cause excessive workload for autosign (I think).

Please look if my comprehension of the tagging process is correct, things are really garbled when speaking about side-tag workflow in Bodhi...

### DIFF
--- a/bodhi/server/models.py
+++ b/bodhi/server/models.py
@@ -2540,30 +2540,34 @@ class Update(Base):
             data['karma_critipath'] = 0
             up.date_testing = None
 
-            # Remove all koji tags and change the status back to pending
-            if up.status is not UpdateStatus.pending:
+            if up.status != UpdateStatus.pending:
+                # Remove all koji tags and change the status back to pending
                 up.unpush(db)
                 caveats.append({
                     'name': 'status',
                     'description': 'Builds changed.  Your update is being '
                     'sent back to testing.',
                 })
+                builds_to_tag = [b.nvr for b in up.builds]
+            else:
+                # No need to unpush the update, just tag new builds
+                builds_to_tag = new_builds
 
-            # Add the pending_signing_tag to all new builds
+            # Add the pending_signing_tag to builds where needed
             tag = None
-            if up.from_tag:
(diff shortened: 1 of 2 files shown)


mattiaverga:
Also, is there a reason why side-tag updates are not automatically pushed to testing right after editing builds? I can't find any:
https://github.com/fedora-infra/bodhi/blob/0d0a69b046f42b7e64668ccb1bf1fc168e25c3e5/bodhi/server/models.py#L2565-L2567

AdamSaleh:
@mattiaverga is this something we need in 5.7.1 or will it wait a month or so?

mattiaverga:
@AdamSaleh I suppose it would be nice to have it in 5.7.1, but it's not strictly necessary. In the current state, if a user updates the build list in a side-tag update, the update gets stuck in pending for a couple of days, after that the `check_signed_builds` task should unstuck the update automatically.

AdamSaleh (approved):

AdamSaleh (approved):

PR #4265, opened 2021-10-14
Remove Cgroups v1 tweak for vagrant machine

mattiaverga:
We don't need anymore the Cgroups v1 tweak in the Vagrant development machine, since docker supports Cgroups v2 in F34.


cverna (approved):
Cool :tada:

PR #4267, opened 2021-10-16
Do not override existing request when editing an update

mattiaverga:
The `data['request']` value is always filled with a Testing request by Colander validator schema.
When editing an update we should drop this value, so as to not override existing Update request (if the builds list is changed, the request is overridden later in the code).

Fixes #4263 

Signed-off-by: Mattia Verga <mattia.verga@tiscali.it>

PR #4268, opened 2021-10-25
Do not override existing request when editing an update #2

mattiaverga:
#4267 was wrong and caused some tests to fail.
This will hopefully restore the correct behavior.

### DIFF
--- a/bodhi/server/models.py
+++ b/bodhi/server/models.py
@@ -2448,8 +2448,6 @@ class Update(Base):
         buildinfo = request.buildinfo
         up = db.query(Update).filter_by(alias=data['edited']).first()
         del(data['edited'])
-        data.pop("request", None)
-        req = None
 
         caveats = []
         edited_builds = [build.nvr for build in up.builds]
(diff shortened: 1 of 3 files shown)


AdamSaleh:
Damn it, when I was running it locally I had the impression it was just the CI runner acting up.
I am reviewing it more thoroughly, an found some issues in tests, where we might be over-riding the test-status in way we didn't before (i.e. before it was Testing, now it is None)

mattiaverga:
Uh, strange, before your rebase the unit tests were passing... I will have another look at this.

mattiaverga:
I've reverted #4267 and chose to try not injecting the request in UpdateSaveSchema.
Reason is that an UpdateRequest is set by set_request (both service and method are named the same), not by the edit update method.

For the same reason I've removed a test which was failing, but it was inherently wrong: in fact, trying to editing a request for a locked update will still fail with the appropriate error message set by set_request. The only change is that now if you try to edit anything other than the buildlist of a locked update you can (changing buglist, description etc.). I suppose that should be permitted, let me know if you think it is better to lock everything.

AdamSaleh:
Nice! Thanks a lot.

AdamSaleh (approved):

Issue #4256, opened 2021-09-15
Allow case insensitive name


LenkaSeg:
# Observed
When editing release: `bodhi releases edit --name F33` the name F33 has to be capitalized. 
In case of passing `f33`there is an error:
`ERROR: Unable to create/edit release: No row was found for one()` which is not very intuitive to figure out (at least for me).

# Expected
It would be nice if command like this would pass:
 `bodhi releases edit --name f33 --long-name "FFedora 33"`
 
 # Proposed change
 Make the query case insensitive, something like this: https://stackoverflow.com/questions/16573095/case-insensitive-flask-sqlalchemy-query

mattiaverga:
I'm not sure we want such feature. It certainly have sense in Fedora perspective, but in the past I know there were other Bodhi instances out of Fedora. We cannot take for sure that `SomeRelease` is the same that `somerelease`, especially because there's no constrain at database level.

Just my opinion, I'm not in charge here...

PR #4283, opened 2021-11-16
WIP: Compare lowercase name when creating release

joshpetit:
This is for #4256. I'm not sure if this will fix it however since I've been trying to learn how to get everything set up to test. It seems like this is what's causing the error, is this a valid solution?

### DIFF
--- a/bodhi/server/services/releases.py
+++ b/bodhi/server/services/releases.py
@@ -376,7 +376,7 @@ def save_release(request):
 
         else:
             log.info("Editing release: %s" % edited)
-            r = request.db.query(Release).filter(Release.name == edited).one()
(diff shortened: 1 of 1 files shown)


AdamSaleh:
Hey, I'd leave it out for now, having more lenient comparisons would be user friendly, but I am not sure if this wouldn't introduce some sort of unanticipated breaking change somwehere in our tooling, and even @mattiaverga is wary: https://github.com/fedora-infra/bodhi/issues/4256#issuecomment-971271442

joshpetit:
@AdamSaleh Ok that makes sense! I can close the pr for now and if anything changes we can reopen it.

PR #4285, opened 2021-11-18
Correct build tagging when editing an update

### LINKED ISSUE #4284: Adding new builds to an update will let the update stuck
I think I've messed things up in #4261 and probably now when editing any update by adding new builds will let the update in a stale state because the `release-candidate` tag is not added to the new builds.

We (I) need to review my changes and fix this ASAP. And possibly find a way to test the tagging workflow when editing an update, which now isn't tested at all because there's no easy way to check that.

In the meantime, expect more and more infra tickets opened about "builds ejected from the compose"... sorry about that.

mattiaverga:
When editing a sidetag update build list, we must not forget to tag the new builds in the release candidate tag together with the release pending signing tag. The release candidate tag is required by the composer to move the update to the testing repo.
When editing a sidetag update already pushed to testing, the update is unpushed and that takes care of re-tagging all the builds in the release candidate tag. For updates still in pending-testing, we need to take care in the edit() method.

This PR also ensures that unpushing a sidetag build do not leave the build in the release candidate tag, but only in the sidetag.

I've also added some tests which provides a minimum assurance of not screw things up again... as we're using a fake dev buildsystem in tests, it's really hard to test how builds are tagged at the end, so I've used log messages to see what's going on in the background.

Fixes #4284 

### DIFF
--- a/bodhi/server/models.py
+++ b/bodhi/server/models.py
@@ -1482,12 +1482,13 @@ class Build(Base):
         """Return the creation time of the build."""
         return datetime.fromisoformat(self._get_kojiinfo()['creation_time'])
 
-    def unpush(self, koji):
+    def unpush(self, koji, from_side_tag=False):
         """
         Move this build back to the candidate tag and remove any pending tags.
 
         Args:
             koji (bodhi.server.buildsys.Buildsysem or koji.ClientSession): A koji client.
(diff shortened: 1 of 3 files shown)


AdamSaleh:
New plan, merge this now, then test on staging. Damn it, I need to add something w.r.t. multi-builds tou our release process.

PR #4309, opened 2021-12-06
5 7 2 hotfix1 - fixes for model configuration

AdamSaleh:
This is a small hotfix for issue that appeared after we put 5.7.2 into production


mattiaverga:
I've already committed the changes in the 5.7 branch through mergify, see #4306 

In your original pr #4304 I've fixed some tests and rebased to devel, so both devel and 5.7 are updated with the hotfix.

AdamSaleh:
Nice! Now I feel bad I didn't notice, but thanks for the great work :-) I will try to do a proper release this week.

PR #4315, opened 2021-12-10
WIP: BUILD in ZUUL

AdamSaleh:


abompard:
The failure of the F35 test seem to be temporary:
```Curl error (6): Couldn't resolve host name for https://mirrors.fedoraproject.org/metalink?repo=fedora-35&arch=x86_64```

(and the integration tests seem broken in `develop` too anyway)

mattiaverga:
@abompard F35 and Rawhide builds are failing due to https://bugzilla.redhat.com/show_bug.cgi?id=1988199 
That docker update has not been released in RHEL 7 so the Jenkins instance we're using for tests is not usable anymore. That's another reason why Adam started to work on porting tests to zuul.

About integration tests, well, I personally gave up trying to make them reliable...

AdamSaleh:
What do you think about the coverage of the zuul tests? 

I.e. currently the unit tests are only running against single fedora version, and I am not using pip anywhere.

Second, I forgot about the docs build, but I should have those ready.

What is missing for us to replace the required jenkins checks with zuul jobs?

abompard:
@mattiaverga  I think I fixed the integration tests in #4323 . F35 & Rawhide are still failing for the reason you mentioned though, but I guess we'll have to live with it until we can move to zuul.

mattiaverga:
Thanks Adam for working on this! Looks great.

I think we must have docs build and unit tests run against a pip pod, so we can early catch any future breakage. As I requested some time ago in a ticket, it would be nice, if possible, to have a cron job (daily or weekly) against this pip pod, so we can identify more easily what dependency causes the breakage - but that's a plus. By having a pip pod, I'm not sure if it makes sense to run the mypy, flake and pydocstyle tests against it (like we do now) or against a fixed fedora version.

Then we should have unit tests and possibly integration tests run on pods for rawhide and active Fedora versions. Does it makes sense to run a pod for EL7, since Bodhi is deployed on it?

Finally, is it possible to have single test results reported to github, instead of a generic 'fedora-check' status?

AdamSaleh:
@mattiaverga I would like to fix the "we should run periodically to avoid breaking deps" by pinning all of the dependencies and running something like dependabot. But you are certainly correct that we still need to make the pip work.

AdamSaleh:
All of Bodhi, including the VM running composes, currently sits on top of Fedora 34. W.r.t the fedora check status, I would have to ask FBO.

AdamSaleh:
retest this please

PR #4323, opened 2021-12-21
Fix integration tests

abompard:
This set of commits should fix the integrations tests and make them pass again.

### DIFF
--- a/devel/ci/integration/tests/test_bodhi_cli.py
+++ b/devel/ci/integration/tests/test_bodhi_cli.py
@@ -242,6 +242,7 @@ def test_releases_info(bodhi_container, db_container):
   Create Automatic Updates: {create_automatic_updates}
   Package Manager:          {package_manager}
   Testing Repository:       {testing_repository}
+  End of Life:              {eol}
 """.format(**release)
         assert result.output == expected
 
--- a/devel/ci/integration/tests/test_bodhi_cli.py
+++ b/devel/ci/integration/tests/test_bodhi_cli.py
@@ -478,13 +478,8 @@ def test_updates_download(bodhi_container, db_container):
     # The bodhi CLI will execute the koji CLI. Replace that executable with
     # something we can track.
     koji_mock = "#!/bin/sh\necho TESTING CALL $0 $@\n"
-    # we mock-replace the executable in two places
-    # It's because of https://pagure.io/koji/issue/912
-    # Koji is installed from RPM isn't found because it doesn't provide egg info.
-    # So Bodhi will install Koji from Pypi and the Koji-Pypi executable is used.
     with replace_file(bodhi_container, "/usr/bin/koji", koji_mock):
-        with replace_file(bodhi_container, "/usr/local/bin/koji", koji_mock):
-            result = _run_cli(bodhi_container, cmd)
+        result = _run_cli(bodhi_container, cmd)
     assert result.exit_code == 0
     for update in updates:
         assert "Downloading packages from {}".format(update['alias']) in result.output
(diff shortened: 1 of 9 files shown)


abompard:
For some reason the F35 and rawhide builds fail with this error when installing packages:
```
Errors during downloading metadata for repository 'rawhide':
- Curl error (6): Couldn't resolve host name for https://mirrors.fedoraproject.org/metalink?repo=rawhide&arch=x86_64 [getaddrinfo() thread failed to start]
Error: Failed to download metadata for repo 'rawhide': Cannot prepare internal mirrorlist: Curl error (6): Couldn't resolve host name for https://mirrors.fedoraproject.org/metalink?repo=rawhide&arch=x86_64 [getaddrinfo() thread failed to start]
```

The CI VM can't resolve mirrors.fp.o? That's weird, any idea what's going on?

abompard:
One of the commits adjusts for a possible change in Greenwave's return structure, that's the one I would most like to have a review one. I guessed the error came from a change in Greenwave's return structure but I may have been wrong there, as I haven't followed it closely.

devel/ci/integration/tests/test_bodhi_cli.py
```diff
@@ -477,13 +478,8 @@ def test_updates_download(bodhi_container, db_container):
...
     # something we can track.
     koji_mock = "#!/bin/sh\necho TESTING CALL $0 $@\n"
-    # we mock-replace the executable in two places
-    # It's because of https://pagure.io/koji/issue/912
-    # Koji is installed from RPM isn't found because it doesn't provide egg info.
-    # So Bodhi will install Koji from Pypi and the Koji-Pypi executable is used.
     with replace_file(bodhi_container, "/usr/bin/koji", koji_mock):
-        with replace_file(bodhi_container, "/usr/local/bin/koji", koji_mock):
-            result = _run_cli(bodhi_container, cmd)
+        result = _run_cli(bodhi_container, cmd)
```
mattiaverga, line 482: So, can we also remove the `pip install koji` line from dockerfiles?

devel/ci/integration/tests/test_bodhi_tasks.py
```diff
@@ -138,18 +140,17 @@ def find_bug():
...
         "--bugs",
         bug_id,
+        "--autokarma",
         update_alias,
     ]
-    celery_tasks = bodhi_container.execute(["ls", "/srv/celery-results"])
-    print(list(celery_tasks))
     try:
-        bodhi_container.execute(cmd)
+        result = bodhi_container.execute(cmd)
```
mattiaverga: I'm just pedantic here: is the result variable used anywhere?

mattiaverga (commented):
Thank you for working on this! Yes, there were some changes in greenwave stuff, but just like the EOL field for releases these changes weren't reflected in the integration tests, because no one ever looked at their unreliable results.

I've just left a couple of pedantic comments, but overall everything looks fine.

devel/ci/integration/tests/test_bodhi_cli.py
abompard: I think so, I'll remove them and give it a try.

devel/ci/integration/tests/test_bodhi_tasks.py
abompard: It's not, I'll remove the assignation, thanks.

abompard:
As a very pedantic reviewer myself, I do appreciate the comments, thanks! I fixed these issues, let's see if CI passes now.

abompard:
Alright, only F35 & Rawhide fail, as expected.

mattiaverga (approved):
LGTM thanks!

PR #4330, opened 2022-01-18
Split Bodhi into multiple package directories

### LINKED ISSUE #4301: Bodhi should be split into multiple packages
Currently Bodhi is a number of packages combined. It may be beneficial to split into its separate parts. These could be kept in a single repo for now to avoid changing how rpms are done

Currently setup.py is used in an unsupported way to manage several python packages from a single file. We need to split the file, while maintaining the internal hierarchy of the packages.

These docs show how this can be done: https://packaging.python.org/guides/packaging-namespace-packages/

abompard:
This set of commits moves the bodhi-client, bodhi-server and bodhi-messages packages into their own subdirectories, each with their own `setup.py`/`setup.cfg`.

The tests have been moved to their respective directories as well.

Fixes #4301


abompard:
Hey @mattiaverga , we're probably going to cut a Bodhi release early next week and then merge this PR. Since it's a huge change (but mostly files being moved around), I wanted to give you a heads up before the weekend, in case you have pending works that are going to be in conflict after the files are moved.

mattiaverga:
Thanks @abompard , I don't plan to push any other change until this PR is merged, as I understand it would be a pain for you to rebase it.

In fact, I'd like to review this PR in the weekend, if I find some time. I'm really interested to see how you handled it, so I can learn something! I'm also curious about how it works the `import bodhi.server` after the package was renamed to `bodhi-server`... in my previous attempt I had to change all the imports, but I see you've found the way to avoid that.

abompard:
It wouldn't necessarily be a pain to rebase, no, in fact I've done it a few times already.

I'm very happy to have your review, thanks! I suggest you look at the commits one after the other, it should be clearer. The feature I'm using to keep the imports in place is called [namespace packages](https://packaging.python.org/en/latest/guides/packaging-namespace-packages/), if you want to read up on it.
Thanks!

ryanlerch (approved):

bodhi-messages/setup.cfg
```diff
@@ -0,0 +1,55 @@
...
+    Programming Language :: Python :: 3
+    Programming Language :: Python :: 3.6
+    Programming Language :: Python :: 3.7
+    Topic :: System :: Software Distribution
+
+[options]
+zip_safe = False
+include_package_data = True
+packages = find_namespace:
+namespace_packages = bodhi
```
hroncok, line 28: @abompard Was there any particular reason you chose to use the [pkg_resources-style namespace package](https://packaging.python.org/en/latest/guides/packaging-namespace-packages/#creating-a-namespace-package)? It creates a .pth file, slowing down every Python startup when bodhi is installed. For Python 3, it should not be needed at all. Upstream says [this approach is no longer recommended](https://packaging.python.org/en/latest/guides/packaging-namespace-packages/#pkg-resources-style-namespace-packages).

bodhi-messages/setup.cfg
```diff
@@ -0,0 +1,55 @@
...
+install_requires =
```
hroncok, line 29: And when you do use the above pkg_resources-style namespace package, I believe you do need to runtime-require setuptools explicitly.

bodhi-messages/setup.cfg
hroncok: https://github.com/fedora-infra/bodhi/pull/4439

abompard: Oh I meant to do the Python 3 style but apparently I messed things up. Thanks for the PR!

PR #4340, opened 2022-01-20
Replace URLs with extlinks when possible in release notes

abompard:
This avoids a warning in recent versions of Sphinx and should make the `pip-docs` check pass.


mattiaverga (approved):
Ah, we pushed the same fix at the same time :-p

PR #4341, opened 2022-01-20
Sphinx 4 doesn't like hardlinks

mattiaverga:
Replacing all hardlinks in old release notes should make pip-docs test happy again.


mattiaverga:
See #4340

Issue #4248, opened 2021-09-02
Rawhide CI image build test fails


AdamSaleh:
The rawhide image build fails with errors during installation of packages through dnf:

```
12:01:24  [91mErrors during downloading metadata for repository 'rawhide':
12:01:24    - Curl error (6): Couldn't resolve host name for https://mirrors.fedoraproject.org/metalink?repo=rawhide&arch=x86_64 [getaddrinfo() thread failed to start]
12:01:24    - Curl error (6): Couldn't resolve host name for https://mirrors.fedoraproject.org/metalink?repo=rawhide&arch=x86_64&countme=1 [getaddrinfo() thread failed to start]
12:01:24  [0m[91mError: Failed to download metadata for repo 'rawhide': Cannot prepare internal mirrorlist: Curl error (6): Couldn't resolve host name for https://mirrors.fedoraproject.org/metalink?repo=rawhide&arch=x86_64 [getaddrinfo() thread failed to start]
```

It seems to be related to https://bugzilla.redhat.com/show_bug.cgi?id=1988199

mattiaverga:
I've pushed #4286 which fixes f34 and pip based tests.
Still not working f35 and rawhide builds, yet locally they're working and the bug you referenced has been closed.

ryanlerch:
We are now running the tests in github actions, which should resolve this issue

PR #4357, opened 2022-01-28
Make the Ci tests run on github actions

ryanlerch:
builds on https://github.com/fedora-infra/bodhi/pull/4353

This updates the CI tests to run using github actions. it also removes the old jenkins and zuul configs.

Resolves: #4357 #4248


.github/workflows/tests.yml
```diff
@@ -0,0 +1,91 @@
...
+  pull_request:
+    branches:
+      - develop
+
+name: Run tests
+
+jobs:
+
+  precommit:
+    name: Run the precommit checks
```
abompard: Maybe call it "Pre-commit checks", to be consistent with the rest?

.github/workflows/tests.yml
```diff
@@ -0,0 +1,91 @@
...
+        run: pip install click conu pytest pytest-cov munch psycopg2 fedora-messaging
+
+      # we only really need to run the precommit checks for one release,
+      # so we just do it for pip
+      - name: Run the precommit checks
+        run: devel/ci/bodhi-ci pre-commit -r pip
+  
+  
+  diff_cover:
+    name: Run the Diff Cover checks
```
abompard: Maybe call it "Diff-cover check", to be consistent with the rest?

.github/workflows/tests.yml
ryanlerch: updated. also updated the mergify commit to reflect this change

.github/workflows/tests.yml
ryanlerch: updated. also updated the mergify commit to reflect this change

abompard:
The unit test fail on bodhi-server for f34, f35 and rawhide but appear as green, and it seems simple to fix, but it reveals what is probably a bug of mine on bodhi-ci, maybe it exits with a zero code even if the tests fail. I'll check.

devel/ci/bodhi_ci/unit.py
```diff
@@ -47,13 +48,13 @@ def __init__(self, *args, **kwargs):
...
-            # Run setup.py develop in each submodule
-            f'for submodule in {modules}; do '
+            # Run setup.py develop in all 3 submodules
+            f'for submodule in {" ".join(MODULES)}; do '
             'cd $submodule; /usr/bin/python3 setup.py develop; cd ..; done; '
             # Run the tests in each submodule
             f'for submodule in {modules}; do '
             'mkdir -p /results/$submodule; '
-            f'cd $submodule; /usr/bin/python3 -m pytest {pytest_flags} || '
+            f'cd $submodule; /usr/bin/python3 -m pytest {pytest_flags}; exit $? || '
```
abompard: Nope, that can't work, it'll exit after the first module every time

abompard (changes requested):

devel/ci/bodhi_ci/unit.py
abompard: But it's the right place to fix it! :-) I have fixed it in devel, I'll push it and you can just rebase.

ryanlerch: done, removed my last commit and rebased with your fix

abompard (approved):

PR #4360, opened 2022-01-29
bash script to build all pypi packages at once

mattiaverga:
A simple bash script to build all bodhi packages in a single command and output sdist and wheel in a common top level dir, so it's easier to upload to Pypi at once.


abompard:
(I fixed the bug that made the diff-cover check fail, please rebase)

PR #4361, opened 2022-01-29
Switch to pymediawiki

mattiaverga:
[Simplemediawiki](https://github.com/iliana/python-simplemediawiki) has been abandoned years ago. Shall we consider switching to something else?
I choosed pymediawiki because it fits our needs, it's maintained and has been around for some years. I'm open to consider other candidates.


bodhi-server/bodhi/server/models.py
```diff
@@ -1558,30 +1556,26 @@ def update_test_cases(self, db):
 
         start = datetime.utcnow()
         log.debug(f'Querying the wiki for test cases of {self.nvr}')
-
-        wiki = MediaWiki(config.get('wiki_url'))
-        cat_page = f'Category:Package {self.package.external_name} test cases'
+        try:
+            wiki = MediaWiki(config.get('wiki_url'),
+                             user_agent=config.get('wiki_user_agent'))
+        except Exception:
```
abompard: Would it be useful here to also store in the `ExternalCallException` the error message that was raised?

bodhi-server/bodhi/server/models.py
```diff
@@ -1558,30 +1556,26 @@ def update_test_cases(self, db):
...
 
         def list_categorymembers(wiki, cat_page, limit=500):
-            # Build query arguments and call wiki
-            query = dict(action='query', list='categorymembers',
-                         cmtitle=cat_page, cmlimit=limit)
             try:
-                response = wiki.call(query)
-            except URLError:
+                response = wiki.categorymembers(cat_page, results=limit, subcategories=True)
+            except Exception:
```
abompard: Same comment here.

abompard:
One more comment: pymediawiki does not seem to be in Fedora yet, so we'll have to package it up.

mattiaverga:
> One more comment: pymediawiki does not seem to be in Fedora yet, so we'll have to package it up.

A new package review request is waiting in the queue: https://bugzilla.redhat.com/show_bug.cgi?id=2051107

mattiaverga:
I'll mark this as WIP because it is now mandatory that all dependencies are available in Fedora repositories. I'll rework it when pymediawiki gets approved.

mattiaverga:
One test in the integration tests is failing because it looks like it cannot connect to Fedora wiki to fetch testcases. I'm not sure if it should be able to connect through Internet or not... however, I could not find anything that previously mocked the Simplemediawiki call, so either things are not working as expected now, or it already failed to connect to the wiki previously but the test didn't fail.

At the moment I'm unable to run tests locally because running a Vagrant machine seems broken, at least in F35: https://bugzilla.redhat.com/show_bug.cgi?id=2066046

mattiaverga:
I'm going to retire this PR... I don't have much time and the recent changes to Bodhi repository (tinystage adoption and switch to poetry) it has become more and more complicated for me to develop changes.
I will try to return developing Bodhi in future times, maybe.

PR #4376, opened 2022-02-08
Make bodhi use poetry

### LINKED ISSUE #4365: Convert bodhi to poetry

LenkaSeg:
What is done so far:
- most of the setup.cfg is translated to pyproject.toml
- with all 3 subpackages the poetry install works and poetry.lock file is created
- solved the problem with the gssapi dependency by:
`sudo yum install krb5-devel`
(then there was an error with missing Python.h) =>
`sudo dnf install python3-devel`
`poetry install` => success
- figured out the entrypoints => the setup.cfg entrypoints needed to to split into 2 sections:
`[tool.poetry.plugins."paste.app_factory"]` and `[tool.poetry.scripts]`

What needs to be done:
- figure out some of the setup.cfg things like `zip_safe` and `include_package_data = True/False`
- check properly how to manage nested packages with poetry

Tested in a podman container:
`podman run -it --rm docker.io/fedora:35 /bin/bash`

bodhi-client needed these libraries:
`sudo dnf install poetry`
`sudo dnf install python`
`sudo yum install krb5-devel`
`sudo dnf install gcc`
`sudo dnf install python3-devel`
`sudo dnf install redhat-rpm-config`

bodhi-server needed:
`dnf install /usr/bin/pg_config`

After that `poetry install` worked.
Confirming that the bodhi-client works somehow:
`poetry run bodhi`
`poetry run bodhi updates query`
`poetry run bodhi releases list`
=> seems to work


bodhi-server/bodhi-server.spec
```diff
@@ -157,11 +130,6 @@ install -pm0644 docs/_build/*.1 %{buildroot}%{_mandir}/man1/
 %files -n bodhi-composer
 %license COPYING
 %doc README.rst
-%{python3_sitelib}/bodhi/server/tasks/composer.py
```
abompard, line 159: This is why the integration test fail :-)

bodhi-server/bodhi-server.spec
```diff
@@ -125,7 +101,7 @@ install apache/bodhi.wsgi %{buildroot}%{_datadir}/bodhi/bodhi.wsgi
 install -d %{buildroot}%{_mandir}/man1
 install -pm0644 docs/_build/*.1 %{buildroot}%{_mandir}/man1/
 
-%files -n %{pypi_name}
+%files -n %{pypi_name} -f %{pyproject_files}
```
hroncok: When you wish to put some files aside, don't use `%{pyproject_files}`. It is a shortcut that can be used only if all files should end up in the same RPM package. 

```suggestion
%files -n %{pypi_name}
```

bodhi-server/bodhi-server.spec
```diff
@@ -103,12 +72,19 @@ repositories.
...
+
 %build
-%py3_build
+%pyproject_wheel
 make %{?_smp_mflags} -C docs man
 
 %install
-%py3_install
+%pyproject_install
+%pyproject_save_files bodhi
```
hroncok:
```suggestion
```

bodhi-server/bodhi-server.spec
```diff
@@ -140,9 +116,6 @@ install -pm0644 docs/_build/*.1 %{buildroot}%{_mandir}/man1/
 %config(noreplace) %{_sysconfdir}/bodhi/alembic.ini
 %config(noreplace) %{_sysconfdir}/httpd/conf.d/bodhi.conf
 %dir %{_sysconfdir}/bodhi/
-%{python3_sitelib}/bodhi
-%{python3_sitelib}/bodhi_server-%{pypi_version}-py%{python3_version}-*.pth
-%{python3_sitelib}/bodhi_server-%{pypi_version}-py%{python3_version}.egg-info
 %{_mandir}/man1/bodhi-*.1*
```
hroncok, line 119:
```suggestion
%{python3_sitelib}/bodhi
%{python3_sitelib}/bodhi_server-%{pypi_version}.dist-info
%{_mandir}/man1/bodhi-*.1*
```

bodhi-server/bodhi-server.spec
```diff
@@ -157,11 +130,6 @@ install -pm0644 docs/_build/*.1 %{buildroot}%{_mandir}/man1/
 %files -n bodhi-composer
 %license COPYING
 %doc README.rst
```
hroncok, line 132: This suggestion got lost somehow, so sending it separately.

```suggestion
%doc README.rst
%pycached %{python3_sitelib}/bodhi/server/tasks/composer.py
%pycached %{python3_sitelib}/bodhi/server/tasks/metadata.py
```

bodhi-client/bodhi-client.spec
```diff
@@ -1,6 +1,7 @@
 # Created by pyp2rpm-3.3.7
 %global pypi_name bodhi-client
 %global pypi_version 5.7.5
+%global _pyproject_record %{_builddir}/%{pypi_name}-pyproject-record
```
hroncok: JFYI Trying to solve this in the pyproject macros themselves: https://src.fedoraproject.org/rpms/pyproject-rpm-macros/pull-request/260

hroncok: Fixed in:

 - [pyproject-rpm-macros-1.0.1-1.fc37](https://bodhi.fedoraproject.org/updates/FEDORA-2022-442e382b65)
 - [pyproject-rpm-macros-1.0.1-1.fc36](https://bodhi.fedoraproject.org/updates/FEDORA-2022-ba127f48b6)
 - [pyproject-rpm-macros-1.0.1-1.fc35](https://bodhi.fedoraproject.org/updates/FEDORA-2022-134c648316)
 - [pyproject-rpm-macros-1.0.1-1.fc34](https://bodhi.fedoraproject.org/updates/FEDORA-2022-9153b6e145)
 - [pyproject-rpm-macros-1.0.1-1.el9](https://gitlab.com/redhat/centos-stream/rpms/pyproject-rpm-macros/-/merge_requests/10)

bodhi-server/bodhi-server.spec
hroncok: Mea culpa, I've learned in https://bugzilla.redhat.com/show_bug.cgi?id=2049301 that you can actually keep using `%{pyproject_files}` if you add the `%exclude` lines to the package.

hroncok: tl;dr this should be possible:

```
%files -n %{pypi_name} -f %{pyproject_files}
...
%pycached %exclude %{python3_sitelib}/bodhi/server/tasks/composer.py
%pycached %exclude %{python3_sitelib}/bodhi/server/tasks/metadata.py


%files -n bodhi-composer
...
%pycached %{python3_sitelib}/bodhi/server/tasks/composer.py
%pycached %{python3_sitelib}/bodhi/server/tasks/metadata.py
```

bodhi-client/pyproject.toml
```diff
@@ -1,3 +1,49 @@
...
+    { path = "bodhi-complete.sh", format = "sdist" },
+]
+exclude = ["docs/_build"]
+
+[tool.poetry.dev-dependencies]
+Sphinx = "^3.4.3"
+pytest = ">=6.0.0"
+
+[tool.poetry.dependencies]
+python = "^3.9"
```
abompard: Why Python 3.9?

bodhi-server/pyproject.toml
```diff
@@ -1,3 +1,167 @@
...
+libcomps ="*"
+prometheus-client = "^0.13.1"
+psycopg2 = "^2.8.6"
+py3dns = "^3.2.1" # Should be required by pyLibravatar https://bugs.launchpad.net/pylibravatar/+bug/1661218
+pyLibravatar = "^1.6"
+pyasn1-modules = "^0.2.8" # Due to an unfortunate dash in its name, installs break if pyasn1 is installed first 
+pyramid = "^1.10.5"
+pyramid-fas-openid = "^0.4.0"
+pyramid-mako = "^1.0.2"
+python = "^3.9"
```
abompard: Why Python 3.9?

bodhi-server/pyproject.toml
```diff
@@ -1,3 +1,167 @@
...
+PyYAML = ">=5.4.1"
+SQLAlchemy = "^1.3.24"
+alembic = "^1.5.5"
+arrow = ">=0.17.0"
+backoff = "^1.10.0"
+bleach = ">=3.2.3"
+bodhi-messages = "^5.7.4"
+celery = [
+    {version = "<=5.2.2, >=5.0.5", python = "3.9"},
+    {version = "<=5.2.3, >=5.0.5", python = "3.10"}
```
abompard: I wonder if you could set something like:
```toml
celery = [
    {version = "<=5.2.2, >=5.0.5", python = "3.9"},
    {version = ">=5.0.5"}
]
```
What I think we want is an exception for F34 where setuptools is too old for celery 5.2.3, but we also don't want to lock celery under 5.2.3 for all other distributions.

bodhi-server/pyproject.toml
```diff
@@ -1,3 +1,167 @@
...
+py3dns = "^3.2.1" # Should be required by pyLibravatar https://bugs.launchpad.net/pylibravatar/+bug/1661218
+pyLibravatar = "^1.6"
+pyasn1-modules = "^0.2.8" # Due to an unfortunate dash in its name, installs break if pyasn1 is installed first 
+pyramid = "^1.10.5"
+pyramid-fas-openid = "^0.4.0"
+pyramid-mako = "^1.0.2"
+python = "^3.9"
+python-bugzilla = "^3.2.0"
+requests = "^2.25.1"
+responses = ">=0.10.15"
```
abompard: This is only used in the tests, it goes to `dev-dependencies`.

bodhi-server/pyproject.toml
```diff
@@ -1,3 +1,167 @@
...
+bodhi-untag-branched = "bodhi.server.scripts.untag_branched:main"
+bodhi-skopeo-lite = "bodhi.server.scripts.skopeo_lite:main"
+bodhi-sar = "bodhi.server.scripts.sar:get_user_data"
+bodhi-shell = "bodhi.server.scripts.bshell:get_bodhi_shell"
+bodhi-clean-old-composes = "bodhi.server.scripts.compat:clean_old_composes"
+bodhi-expire-overrides = "bodhi.server.scripts.compat:expire_overrides"
+bodhi-approve-testing = "bodhi.server.scripts.compat:approve_testing"
+bodhi-check-policies = "bodhi.server.scripts.compat:check_policies"
+
+#[init_catalog]
```
abompard: If babel can't read it in `pyproject.toml`, you can drop it from here.

devel/ansible/roles/bodhi/tasks/main.yml
```diff
@@ -22,19 +22,21 @@
       - pcaro-hermit-fonts
       - pcp-system-tools
       - pre-commit
+      - poetry
       - postgresql-devel
       - postgresql-server
-      - python3-alembic
```
abompard, line 27: Why remove it?

devel/ansible/roles/bodhi/tasks/main.yml
```diff
@@ -119,19 +126,34 @@
...
+- name: Create a directory for the virtualenv
+  file:
+    name: /srv/venv
+    state: directory
+    mode: 0755
+    owner: vagrant
+    group: vagrant
+
+- name: Create the virtualenv
+  shell: virtualenv --system-site-packages /srv/venv/
```
abompard: I think ansible has a `pip` module that can create virtualenvs, maybe it's worth using here?

devel/ansible/roles/bodhi/tasks/main.yml
```diff
@@ -119,19 +126,34 @@
...
+  become: yes
+  become_user: vagrant
+  creates: /srv/venv/bin/python
+
+- name: Install bodhi with poetry
+  shell: VIRTUAL_ENV=/srv/venv poetry install
```
abompard: I think ansible's `command` module can set environement variables. The `shell` module is usually a last resort option, because it's a bit less secure (not relevant in our case but we may want to copy part of this ansible playbook in the infra one day).

devel/ansible/roles/bodhi/tasks/main.yml
```diff
@@ -161,7 +183,9 @@
     state: directory
 
 - name: Apply database migrations
-  shell: PYTHONPATH=. alembic upgrade head
+  become: yes
+  become_user: vagrant
+  shell: VIRTUAL_ENV=/srv/venv poetry run alembic -c alembic.ini upgrade head
```
abompard: Same here.

devel/ci/bodhi_ci/unit.py
```diff
@@ -50,21 +50,29 @@ def __init__(self, *args, **kwargs):
         modules = " ".join(self.options["modules"])
 
         test_command = (
-            # Run setup.py develop in all 3 modules
+            # Run poetry install in all 3 modules
             f'for submodule in {" ".join(MODULES)}; do '
-            'cd $submodule; /usr/bin/python3 setup.py develop; cd ..; done; '
+            '  pushd $submodule; '
```
abompard, line 55: This bit of scripting is repeated multiple times, we should probably factor it out in an external shell script. Check out how the RPMs are built for an example.

devel/ci/build-rpms.sh
```diff
@@ -16,16 +16,20 @@ versionsuffix=\^$(date -u +%Y%m%d%H%M)git$githash
...
 
 for submodule in ${MODULES}; do
-    cd $submodule
-    /usr/bin/python3 setup.py sdist
+    pushd $submodule
+    poetry build -f sdist
     cp dist/* ~/rpmbuild/SOURCES/
     cp $submodule.spec ~/rpmbuild/SPECS/
-    moduleversion=$(python3 setup.py --version)
+    githash=$(git rev-parse --short HEAD)
```
abompard: Not sure it's useful to add this, it's already above in the script.

pyproject.toml
```diff
@@ -1,3 +1,104 @@
+[tool.poetry]
```
abompard: Let's remove that part, we don't want to build a generic `bodhi` package at all I think.

bodhi-messages/pyproject.toml
```diff
@@ -1,3 +1,76 @@
...
+
+#[options.packages.find]
+#exclude =
+#    tests
+#zip_safe = False
+
+[tool.poetry.dependencies]
+"zope.interface" = "^5.2.0"
+fedora-messaging = "^3.0.0"
+python = "^3.9"
```
abompard: Why Python 3.9?

bodhi-server/pyproject.toml
```diff
@@ -1,3 +1,167 @@
...
+    { path = "bodhi/*.woff2", format = "sdist" },
+    { path = "bodhi/*.eot", format = "sdist" },
+    { path = "bodhi/*.otf", format = "sdist" },
+    { path = "bodhi/*.svg", format = "sdist" },
+    { path = "bodhi/*.tpl", format = "sdist" },
+    { path = "bodhi/*.rst", format = "sdist" },
+    { path ="apache/*", format = "sdist" },
+    { path = "docs", format = "sdist" },
+        ]
+exclude = ["development.ini", "docs/_build"]
```
abompard: This doesn't seem to work, the `development.ini` file is in the resulting `tar.gz` :-/

bodhi-server/pyproject.toml
```diff
@@ -1,3 +1,167 @@
...
+]
+click = ">=7.1.2"
+colander = "^1.8.3"
+cornice = "^5.0.3"
+"dogpile.cache" = "^1.1.2"
+fedora-messaging = "^3.0.0"
+feedgen= "^0.9.0"
+koji = "^1.27.1"
+#libcomps ="^0.1.18"
+libcomps ="*"
```
abompard, line 97: Could you explain this in a comment please?

abompard (changes requested):
Nice work! Some additional comments:
- Looks like you forgot the `importlib_metadata` thing in `bodhi-client/bodhi/client/__init__.py`.
- You can also remove `bodhi-client/setup.cfg` and most of `bodhi-server/setup.cfg` (leave the babel stuff in unless it can be configured in `pyproject.toml`).
- You shouldn't need the `test-requirements.txt` files, poetry's `dev-dependencies` replaces them.

Thanks!

bodhi-client/pyproject.toml
LenkaSeg: Should be lower? F34 is using 3.9.

bodhi-server/pyproject.toml
LenkaSeg: I probably wanted to see which libcomps would be installed.
You mean to remove the wildcard or leave it and comment which problem we were solving with it?

bodhi-server/pyproject.toml
LenkaSeg: I don't understand this comment. You mean to remove these commented blocks?

devel/ansible/roles/bodhi/tasks/main.yml
LenkaSeg: It's also installed by poetry, so I thought it would not be necessary here. Should I return it back?

devel/ansible/roles/bodhi/tasks/main.yml
LenkaSeg: Ok, I'm gonna check it!

devel/ansible/roles/bodhi/tasks/main.yml
LenkaSeg: Ok, will change to command.

bodhi-client/pyproject.toml
abompard: Yeah but by using `importlib.metadata` we're actually only needing 3.8, since that's when it was added. Is there anything else requiring 3.9?

bodhi-server/pyproject.toml
abompard: Yeah, let's not leave commented out stuff if we don't need it.

devel/ansible/roles/bodhi/tasks/main.yml
abompard: Well we install a lot of python packages below that are also added by poetry, so let's not make an exception for this one if we don't have a good reason :-)

bodhi-client/pyproject.toml
LenkaSeg: I'm not aware. Changing it to 3.8.

bodhi-server/pyproject.toml
LenkaSeg: Tried with the `path` and seems it was finally excluded.

bodhi-server/pyproject.toml
abompard: Well, if we actually need the wildcard, then we need a comment to explain why it's needed. If we don't need it, then it should be removed :-)

devel/ansible/roles/bodhi/tasks/main.yml
LenkaSeg: Seems that the pip module does not allow to install only virtualenv without any package.

devel/ci/bodhi_ci/unit.py
LenkaSeg: This is taking me a bit more time. Would it be possible to do it as a separate PR?

abompard: Absolutely!

bodhi-server/pyproject.toml
```diff
@@ -1,3 +1,142 @@
...
+bodhi-messages = "^5.7.4"
+celery = [
+    {version = "<=5.2.2, >=5.0.5", python = "3.9"},
+    {version = ">=5.0.5", python = "^3.10" }
+]
+click = [
+    {version = "<8.0.0, >=7.1.2", python ="3.9"},
+    {version = ">=7.1.2", python ="^3.10"}
+]
+#click = ">=7.1.2"
```
abompard: Just delete that and it should be alright! :-)

abompard (changes requested):

LenkaSeg:
Rebased on abompard/drop-f34

bodhi-client/bodhi-client.spec
```diff
@@ -13,43 +13,45 @@ Source0:        %{pypi_name}-%{pypi_version}.tar.gz
...
 
 %install
-%py3_install
+%pyproject_install
+%pyproject_save_files bodhi
+
 install -d %{buildroot}%{_mandir}/man1
 install -pm0644 docs/_build/bodhi.1 %{buildroot}%{_mandir}/man1/
+install -d %{buildroot}%{_sysconfdir}/bash_completion.d
+install -pm0644 bodhi-complete.sh %{buildroot}%{_sysconfdir}/bash_completion.d/bodhi-complete.sh
```
abompard: Please call it `bodhi-client.bash` so that we know it comes from the bodhi-client package. (we may want to write completion scripts for the tools in bodhi-server one day)

devel/ansible/roles/bodhi/tasks/main.yml
```diff
@@ -22,19 +22,22 @@
       - pcaro-hermit-fonts
       - pcp-system-tools
       - pre-commit
+      - poetry
```
abompard, line 28: You'll also need to add the following packages to build `libcomps`:
- cmake
- libxml2-devel
- expat-devel

bodhi-server/pyproject.toml
```diff
@@ -1,3 +1,135 @@
...
+Jinja2 = ">=2.11.3"
+Markdown = "^3.3.6"
+PyYAML = ">=5.4.1"
+SQLAlchemy = "^1.3.24"
+alembic = "^1.5.5"
+arrow = ">=0.17.0"
+backoff = "^1.10.0"
+bleach = ">=3.2.3"
+bodhi-messages = "^5.7.4"
+celery = ">=5.0.5"
```
abompard: From 5.2.3 and up, Celery requires setuptools>=59.1.1 which we don't have in F35. We should set a require for Celery with something like `celery = "<5.2.3"`.

abompard (changes requested):
Almost there! :-)

abompard (approved):
Thanks a lot for this big piece of work!

PR #4385, opened 2022-02-14
Tweak rpm generation

### LINKED ISSUE #4378: Fix the specfiles and RPMs since the split of packages

ryanlerch:
Slightly change how we make the rpms. First is to change the way we
name them based on the suggestions in #4378 

THe second is to add a time to the end of the date in the version
for the snapshot versioning so have more flexibility.

### DIFF
--- a/bodhi-messages/bodhi-messages.spec
+++ b/bodhi-messages/bodhi-messages.spec
@@ -2,7 +2,7 @@
 %global pypi_name bodhi-messages
 %global pypi_version 5.7.4
 
-Name:           %{pypi_name}
+Name:           python-%{pypi_name}
 Version:        %{pypi_version}
 Release:        0%{?dist}
 Summary:        JSON schema for messages sent by Bodhi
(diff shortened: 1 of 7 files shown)


bodhi-messages/bodhi-messages.spec
```diff
@@ -2,7 +2,7 @@
 %global pypi_name bodhi-messages
 %global pypi_version 5.7.4
 
-Name:           %{pypi_name}
+Name:           python3-%{pypi_name}
```
nphilipp: This doesn't conform with the [Fedora Python Packaging Guidelines](https://docs.fedoraproject.org/en-US/packaging-guidelines/Python/#_library_naming): for a package containing primarily a Python library, the source package (aka the "component") must be named `python-…`, and the built package must be named `python3-…`. This means we need a new `%package python3-%{pypi_name}` section.

nphilipp (changes requested):
Other than the source package naming thing, LGTM!

bodhi-messages/bodhi-messages.spec
ryanlerch: I added this section -- i think it is correct :)

nphilipp (approved):
LGTM! 👍

PR #4388, opened 2022-02-16
Minor refactoring

abompard:
These 3 commits are independent from one another, it's just stuff that I refactored while working on OIDC.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -17,7 +17,7 @@ jobs:
       - uses: actions/checkout@v2
 
       - name: Install deps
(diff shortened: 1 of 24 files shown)


mattiaverga:
Just asking: do you think it's better to specify the testing requirements in an external file or maybe by adding them as extras_requires in setup.cfg? https://setuptools.pypa.io/en/latest/userguide/dependency_management.html#optional-dependencies

abompard:
Good question, in this case there's also a set of requirements for the integration tests and for documentation, that we want to be able to install regardless of whether we install the 3 sub-packages (the top directory is not an installable python package), so we'll have to use `requirements.txt` files for those anyway. From there, using those files for the subpackages seems coherent.

PR #4390, opened 2022-02-17
Move man page generation into the modules

### LINKED ISSUE #4378: Fix the specfiles and RPMs since the split of packages

ryanlerch:
This moves the manpage generation out of the /doc/ directory and into the 2 modules that need manpage docs. 


mattiaverga (approved):
LGTM

PR #4391, opened 2022-02-17
Prepare the client for the switch to OIDC

abompard:
This change prepares the client to be compatible with an OIDC-enabled server. It is also compatible with the current server, because bodhi-server currently ignores the new `method` parameter.

We should widely deploy the client with this change before deploying the OIDC-enabled server.


mattiaverga (approved):
Looks good

PR #4394, opened 2022-02-21
More updates to the RPM layout

### LINKED ISSUE #4378: Fix the specfiles and RPMs since the split of packages

ryanlerch:
Clean up the rpms with the following 3 commits:

#### [split out bodhi-composer package in RPMs](https://github.com/fedora-infra/bodhi/pull/4394/commits/606877367d563a42ec81910dd358360f7270acba)
this splits out the bodhi-composer into it's own seperate sub package,
as this is how it was done before the modules split.

#### [group artifacts better when uploading RPMs to github](https://github.com/fedora-infra/bodhi/pull/4394/commits/f7bb9c6eb3c2dffba88ed282fbcbf23c9a6eadd5)
previously we had 18 different zip files with one rpm in each. this
commit groups into the release, so we have 3 artifacts with all the rpms
and srpms for each release.

#### [Add missing requires for bodhi-client rpm](https://github.com/fedora-infra/bodhi/pull/4394/commits/193bc35d01ca4d26b12c2bbd82af71b42459cff1)
bodhi-client specfile was missing the requires that were in the old specfile. this adds them

### DIFF
--- a/bodhi-server/bodhi-server.spec
+++ b/bodhi-server/bodhi-server.spec
@@ -47,9 +47,54 @@ BuildRequires:  python3dist(sqlalchemy)
 BuildRequires:  python3dist(waitress)
 BuildRequires:  python3dist(whitenoise)
 
+Requires: bodhi-client == %{version}-%{release}
+Requires: python3-bodhi-messages == %{version}-%{release}
+Requires: fedora-messaging
+Requires: git
+Requires: httpd
+Requires: intltool
+Requires: python3-koji
+Requires: python3-librepo
+Requires: python3-mod_wsgi
+
+Provides:  bundled(aajohan-comfortaa-fonts)
+Provides:  bundled(abattis-cantarell-fonts)
+Provides:  bundled(bootstrap) = 3.0.1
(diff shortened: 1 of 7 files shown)


bodhi-server/bodhi-server.spec
```diff
@@ -83,6 +128,20 @@ install -pm0644 docs/_build/*.1 %{buildroot}%{_mandir}/man1/
 %{python3_sitelib}/bodhi_server-%{pypi_version}-py%{python3_version}.egg-info
 %{_mandir}/man1/bodhi-*.1*
 %{_mandir}/man1/initialize_bodhi_db.1*
+# These excluded files are in the bodhi-consumers package so don't include them here.
```
abompard: You probably mean `bodhi-composer` in the comment here, no?

abompard (commented):
Just one minor thing and then it's all good! :-)

bodhi-server/bodhi-server.spec
ryanlerch: fixed! thanks!

abompard (approved):

PR #4396, opened 2022-02-23
Bump release to 5.7.5

### LINKED ISSUE #4398: Release 5.7.5 (with just the OIDC client changes)

ryanlerch:


mattiaverga (commented):
The `Release` tags in spec files are wrong (they're set to `0` while in changelog there's `1`)

ryanlerch:
> The `Release` tags in spec files are wrong (they're set to `0` while in changelog there's `1`)

yeah, i'm not 100% sure how to work this -- the release is set to 0 because we auto bump it before doing the test rpm builds.

i suppose i can just change the clanglog back to -0

PR #4403, opened 2022-03-02
Make footer copyright year always current year

### LINKED ISSUE #4401: Copyright only to 2019 on Fedora Bodhi?

ryanlerch:
Previously, the copyright year in the master template was hardcoded to
the year. If this failed to be manually updated, it would go stale at
the time it was 2019 instead of 2022.

THis commit programatically sets the copyright year to the current year
in the template.

Resolves: #4401

### DIFF
--- a/news/4401.bug
+++ b/news/4401.bug
@@ -1,0 +1,1 @@
+Fix the copyright year in the footer going stale by programatially setting the year.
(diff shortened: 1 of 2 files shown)


news/4401.bug
```diff
@@ -0,0 +1 @@
+Fix teh copyright year in the footer going stale by programatially setting the year.
```
abompard: Typo! :-)

abompard (changes requested):

news/4401.bug
ryanlerch: thanks fixed!

abompard (approved):

PR #4431, opened 2022-03-17
Extend `save_override()` to set the expiration date of an override directly

abompard:
The fedpkg tool needs to extend the lifetime of overrides, let's make it easier for them and add the feature here.


mattiaverga (approved):
I had a thought about adding the 'expiration_date' option to the CLI, but it doesn't seem practical, so it's just fine to add it to the bindings IMO.

PR #4436, opened 2022-03-22
In Vagrant, don't put development.ini inside the repo

abompard:
This leads to errors when changing branches and rebuilding the Vagrant
machine, because the `development.ini` gets reused.

In existing Vagrant boxes, just move your `bodhi-server/development.ini`
file to `/home/vagrant/development.ini`.


ryanlerch (approved):

mattiaverga:
This breaks a few settings declared with the `%(here)s` directive... but I don't see how to fix them so that they work both for vagrant and CI tests, as the development.ini file is copied in different places.
I also have the suspect that some directories were already broken (for exampl, the `pungi.basepath`)

PR #4439, opened 2022-03-22
Use namespace packages from this decade, avoid needless .pth files

hroncok:
The `namespace_packages` setuptools options is for pkg_resources-style
namespace packages [1]. It is no longer needed nor recommended by setuptools.

Avoid `bodhi/__init__.py` and `*.pth` files altogether and use Python-native
(PEP 420) namespace packages [2] instead.

Warning: This change requires that all 3 affected packages are updated,
it is not possible to combine a bodhi package from before this change with
another bodhi package from after this change.

[1] https://packaging.python.org/en/latest/guides/packaging-namespace-packages/#pkg-resources-style-namespace-packages
[2] https://packaging.python.org/en/latest/guides/packaging-namespace-packages/#native-namespace-packages

----------

This might be complete moot considering I believe that https://github.com/fedora-infra/bodhi/pull/4376 will result in the same outcome. But OTOH maybe merging this before https://github.com/fedora-infra/bodhi/pull/4376 will make the change more explicit and allow making the poetry transition less magical (at least in this regard).

Untested, hence a draft. Will see what does the :fox_face: say, erhm... CI.

### DIFF
--- a/bodhi-client/bodhi-client.spec
+++ b/bodhi-client/bodhi-client.spec
@@ -48,7 +48,6 @@ install -pm0644 docs/_build/bodhi.1 %{buildroot}%{_mandir}/man1/
 %files -n %{pypi_name}
 %{_bindir}/bodhi
 %{python3_sitelib}/bodhi
-%{python3_sitelib}/bodhi_client-%{pypi_version}-py%{python3_version}-*.pth
 %{python3_sitelib}/bodhi_client-%{pypi_version}-py%{python3_version}.egg-info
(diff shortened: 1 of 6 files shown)


hroncok:
Not sure what's wrong on the CI.

ryanlerch:
> Not sure what's wrong on the CI.

It's not you -- the integration tests use the most recent real db dump, and there is an issue with the tests / an update that is causing the integratiuon tests to fail.

https://github.com/fedora-infra/bodhi/issues/4440

abompard (approved):

PR #4442, opened 2022-03-23
copy severity when obsoleting a security update

### LINKED ISSUE #4441: Obsoleting a security update with a non-security update does not copy severity
If a packager obsoletes a security update with a non-security update type, the incoming package is automatically set to a security update in this portion of code:

https://github.com/fedora-infra/bodhi/blob/develop/bodhi-server/bodhi/server/models.py#L2746-L2753

However, security updates appear to require a severity set:

https://github.com/fedora-infra/bodhi/blob/develop/bodhi-server/bodhi/server/validators.py#L829

ryanlerch:
Previously, If a packager obsoletes a security update with a
non-security update type, the newer update is automatically
set to a security update. However, the severity of an update is
validated to be required on security updates.

This commit copies the severity of the obsoleted update to the new
update when changing the new update to a security update.

Resolves: #4441

### DIFF
--- a/bodhi-server/bodhi/server/models.py
+++ b/bodhi-server/bodhi/server/models.py
@@ -2793,6 +2793,7 @@ class Update(Base):
                         'since it obsoletes another security update'
                     })
                     self.type = UpdateType.security
+                    self.severity = oldBuild.update.severity
 
                 if obsoletable:
                     log.info('%s is obsoletable' % oldBuild.nvr)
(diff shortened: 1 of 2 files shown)


bodhi-server/bodhi/server/models.py
```diff
@@ -213,6 +213,8 @@ def db_type(cls):
 class DeclEnumType(SchemaType, TypeDecorator):
     """A database column type for an enum."""
 
+    cache_ok = True
```
abompard: I'll address this in PR #4437, could you remove it from here please?

abompard (changes requested):

bodhi-server/bodhi/server/models.py
ryanlerch: whoops! sorry! that snuck in there....

removed now

abompard (approved):

PR #4446, opened 2022-03-27
Obsolete updates when a newer one is pushed to stable

mattiaverga:
We currently ignore obsoleting updates which are being pushed to stable. The rationale is that we can create an update with a newer build and push it to testing while allowing the older one to be pushed stable meanwhile.

Unfortunately, this doesn't work well when a release is frozen, as we may end up with several builds all stuck in testing being pushed to stable. When the freeze is lifted, not only the newer build is pushed to stable, but Bodhi will try to push the older too.

This PR is an attempt to fix that by allowing Bodhi to obsolete updates being pushed to stable, but only when the newer update is pushed to stable too. To achieve that I've added a call to `obsolete_older_updates` at the end of `set_request` and modified the query used to identify obsoletable candidates.


mattiaverga:
This is rebased on #4447

abompard (approved):

PR #4448, opened 2022-03-29
Update links to the documentation to github pages

ryanlerch:


devel/ansible/roles/bodhi/tasks/main.yml
```diff
@@ -100,12 +97,25 @@
   args:
       creates: /home/vagrant/.db-created
 
+# This isn't packaged in Fedora yet, but it's only a development tool (we should still add it)
+- name: pip install pyramid
```
mattiaverga: can you explain the reason about moving from koji provided packages to pip provided packages?

mattiaverga (commented):
Since we're moving the docs to github, shall we also remove the whitenoise configuration in the webapp, the whitenoise requirement and the `docs_path` setting from config?

Another question is: does the github docs provide a way to have multiple branches? I mean, like ReadTheDocs give the ability to have separate docs for the devel branch and for stable branches. It would be confusing to have docs updated from devel which differs from what we have in prod.

devel/ansible/roles/bodhi/tasks/main.yml
ryanlerch: the reason is i'm a silly goose.

Was playing around with trying to test out pyramid2, and left this lying around. removed it now. thanks!

ryanlerch:
> Since we're moving the docs to github, shall we also remove the whitenoise configuration in the webapp, the whitenoise requirement and the `docs_path` setting from config?
> 

will do another PR to do this

> Another question is: does the github docs provide a way to have multiple branches? I mean, like ReadTheDocs give the ability to have separate docs for the devel branch and for stable branches. It would be confusing to have docs updated from devel which differs from what we have in prod.

yeah, we are building and publishing the 5.7 branch and devel branch at the moment -- and by default the docs link redirects to 5.7 -- i.e. the current stable:

https://fedora-infra.github.io/bodhi/5.7/

but you can view the develop branch here:

https://fedora-infra.github.io/bodhi/develop/

mattiaverga (approved):
+1

PR #4450, opened 2022-04-04
Drop F34 as a supported OS

abompard:
Also move the Vagrant VM to 35, which caused a couple issues with RabbitMQ.


Zlopez:
@abompard So it's possible to run rabbitmq-server on F35?

abompard:
@Zlopez Yeah but it's a bit tricky: https://bugzilla.redhat.com/show_bug.cgi?id=2020278#c6

Zlopez:
@abompard Hm, so it just needs update for SELinux policy. I would prefer this to be done in [Fedora package](https://src.fedoraproject.org/rpms/rabbitmq-server) directly instead of adding the policy by provision script.

abompard:
Yeah me too!

abompard:
The integration test failure on rawhide seems unrelated.

PR #4550, opened 2022-06-03
Accept whitenoise 6+ and waitress 2+

abompard:


mattiaverga:
After having moved the docs to github, I think we don't need whitenoise anymore. It was used to serve the static pages of the docs and I suppose we can slim down Bodhi dependencies and remove whitenoise from the webapp.
What do you think?

abompard:
That's a very good point. Let's do that in a different PR however, I think the priority is to get CI back to green.

mattiaverga (approved):
looks good then.

PR #4564, opened 2022-06-12
Do not overwrite flags on update edits

### LINKED ISSUE #4563: `bodhi updates edit UPDATE` disables autokarma and autotime
I have been using `bodhi updates edit UPDATE` _with no options_ as workaround for unlocking submitted sidetag updates.
(Yes that is probably a hack but apart from some timing issues, it seems to work.)

However I noticed anyway that this command seems to disable autokarma and autotime?
which does not seem to desireable?

Admittedly running edit with no options is probably not that common (except for this workaround perhaps),
but I would not expect it to change the state of the update.

mattiaverga:
Fixes #4563 and also add the capability to unset flags when editing an update by CLI.


abompard (approved):

abompard:
@mergifyio rebase

abompard:
I confirm that the diff-cover check passes.

mattiaverga:
@Mergifyio refresh

mattiaverga:
@Mergifyio unqueue

PR #4584, opened 2022-06-20
Do not disable autopush or set request on rawhide updates

### LINKED ISSUE #4566: Do not allow users to set stable request on side-tag rawhide updates
Example: https://pagure.io/releng/issue/10832

Usually, side-tag rawhide updates are pushed to stable instantly after creation, but the above example pointed out a special case.

1. as critpath update, it wasn't pushed to stable due to failing gating tests
2. a negative comment disabled autopush
3. the update was edited so gating tests were passing, but still it wasn't pushed to stable because of 2. At this state, it should have been edited to re-enable autopush and wait for the approve_testing task, but no one did.
4. a user manually set request to push it to stable (I assume by CLI, as webUI shouldn't show the option - @AdamWill do you remember how you set the request? it was by CLI or webUI?)
5. the update was edited so that autopush was re-enabled. But the approve_testing task doesn't process updates with pending requests. And as rawhide update, the existing request wasn't process by composer neither. So the update was stuck.

mattiaverga:
Fixes #4566 

All Rawhide updates (or for any release which is not composed by Bodhi) relies on separate, automatic workflow and we should avoid disabling autopush or setting a request to not break the workflow.

### DIFF
--- a/bodhi-server/bodhi/server/models.py
+++ b/bodhi-server/bodhi/server/models.py
@@ -3801,8 +3801,10 @@ class Update(Base):
         if self.status not in (UpdateStatus.testing, UpdateStatus.pending):
             return
         # If an update receives negative karma disable autopush
+        # exclude rawhide updates see #4566
         if (self.autokarma or self.autotime) and self._composite_karma[1] != 0 and self.status is \
-                UpdateStatus.testing and self.request is not UpdateRequest.stable:
+                UpdateStatus.testing and self.request is not UpdateRequest.stable and \
+                self.release.composed_by_bodhi:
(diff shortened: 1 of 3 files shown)


abompard (commented):
Looks good! LGTM seems to prefer `==` to `is` with enums, but I think it's fine. I looked at zzzek's receipe where our enum implementation comes from, and he uses `is` too.

I'll check what's up with the integration tests.

abompard:
Integration tests failure should be fixed by #4585, please rebase.

abompard (approved):

PR #4601, opened 2022-06-26
Migration to Fedora Bootstrap 2.0.2

mattiaverga:
This will update fedora-bootstrap component to 2.0.2 (Bootstrap 5) and jquery to 3.6.0

I think I've updated everything for compatibility with Bootstrap 5, please test.


mattiaverga:
I'd really like someone to review this PR, since there are so many files involved that delaying the review may cause problematic rebases. This would need a visual test on every page to look for anything I could have missed. I think I've looked everything, but the automated tests tab (I cannot make the vagrant machine connect to greenwave).

As an alternative, shall we merge the PR and test the changes in staging Bodhi to check anything broken?

PR #4602, opened 2022-06-27
Provide Kerberos client authentication via OIDC

### LINKED ISSUE #4588: Add a kerberos-based login method in the client
The Packit folks have written an extension to the Bodhi Client to support kerberos authentication:

https://github.com/packit/packit/issues/1624#issuecomment-1161676764

It's worth considering it for integration in Bodhi Client itself.

TomasTomecek:
Fixes #4588

### DIFF
--- a/bodhi-client/bodhi/client/bindings.py
+++ b/bodhi-client/bodhi/client/bindings.py
@@ -41,9 +41,9 @@ except ImportError:  # pragma: no cover
     # dnf is not available on EL 7.
     dnf = None  # pragma: no cover
 from munch import munchify
-from requests.exceptions import RequestException, ConnectionError
-import requests
+from requests.exceptions import ConnectionError, RequestException
 import koji
+import requests
 
 from .constants import (
     BASE_URL,
@@ -178,7 +178,6 @@ class BodhiClient:
         id_provider: str = IDP,
         staging: bool = False,
         oidc_storage_path: typing.Optional[str] = None,
-        use_kerberos_auth: bool = True,
     ):
         """
         Initialize the Bodhi client.
@@ -189,7 +188,6 @@ class BodhiClient:
             client_id: The OpenID Connect Client ID.
             staging: If True, use the staging server. If False, use base_url.
             oidc_storage_path: Path to a file were OIDC credentials are stored
-            use_kerberos_auth: If True, use Kerberos authentication.
         """
         if staging:
             base_url = STG_BASE_URL
@@ -203,7 +201,6 @@ class BodhiClient:
         self.oidc_storage_path = (
             oidc_storage_path or os.path.join(os.environ["HOME"], ".config", "bodhi", "client.json")
         )
-        self.use_kerberos_auth = use_kerberos_auth
         self._build_oidc_client(client_id, id_provider)
 
     def _build_oidc_client(self, client_id, id_provider):
@@ -240,7 +237,7 @@ class BodhiClient:
 
     def ensure_auth(self):
         """Make sure we are authenticated."""
-        self.oidc.ensure_auth()
+        self.oidc.ensure_auth(use_kerberos=True)
         if not self.oidc.has_cookie("auth_tkt", domain=urlparse(self.base_url).hostname):
             while True:
                 resp = self.oidc.request("GET", f"{self.base_url}oidc/login-token")
@@ -248,19 +245,7 @@ class BodhiClient:
                     break
                 if resp.status_code == 401:
                     self.clear_auth()
-                    # by default, when use_kerberos_auth is True, we use Kerberos to authenticate
-                    # and if that fails, fall back to the interactive browser-based OIDC login
-                    try:
-                        # This snippet may look weird, but we want to cover 3 use cases:
-                        # 1. use_kerberos_auth is True and Kerberos succeeds: continue the loop
-                        # 2. use_kerberos_auth is True and Kerberos fails -> normal login follows
-                        # 3. use_kerberos_auth is False -> normal login follows
-                        if self.use_kerberos_auth:
-                            self.oidc.login_with_kerberos()
-                            continue
-                    except OIDCClientError as e:
-                        log.info("Failed to login with Kerberos: %s", e)
-                    self.oidc.login()
+                    self.oidc.login(use_kerberos=True)
                 else:
                     resp.raise_for_status()
 
--- a/bodhi-client/bodhi/client/bindings.py
+++ b/bodhi-client/bodhi/client/bindings.py
@@ -178,6 +178,7 @@ class BodhiClient:
         id_provider: str = IDP,
         staging: bool = False,
         oidc_storage_path: typing.Optional[str] = None,
+        use_kerberos_auth: bool = True,
     ):
         """
         Initialize the Bodhi client.
@@ -188,6 +189,7 @@ class BodhiClient:
             client_id: The OpenID Connect Client ID.
             staging: If True, use the staging server. If False, use base_url.
             oidc_storage_path: Path to a file were OIDC credentials are stored
+            use_kerberos_auth: If True, use Kerberos authentication.
         """
         if staging:
             base_url = STG_BASE_URL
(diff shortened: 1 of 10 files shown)


bodhi-client/pyproject.toml
```diff
@@ -44,6 +44,7 @@ click = ">=7.1.2"
 koji = "^1.27.1"
 authlib = ">=0.15.4"
 munch = "^2.5.0"
+requests-kerberos = ">=0.12"
```
abompard, line 48: If you change the `pyproject.toml` file, you need to regenerate the lock file with `poetry lock`.

bodhi-client/bodhi/client/oidcclient.py
```diff
@@ -149,6 +149,43 @@ def login(self):
...
+                mutual_authentication=OPTIONAL,
+            ),
+        )
+        response.raise_for_status()
+        try:
+            value = re.findall(
+                r"<title>\s*(code=[\w\-_=;&]+)\s*</title>", response.text
+            )[0]
+        except IndexError:
+            raise OIDCClientError(
```
abompard, line 205: It would be nice to have a test case for this situation, if the regexp is not found in `<title/>`.

bodhi-client/bodhi/client/oidcclient.py
```diff
@@ -149,6 +149,43 @@ def login(self):
...
+                f'Unable to locate OIDC code in the response from "{uri}".'
+            )
+        self.auth_callback(f"?{value}")
+        if not self.tokens:
+            raise OIDCClientError(
```
abompard: Same here, please add a test case for this situation. What caused this in your tests?

bodhi-client/bodhi/client/oidcclient.py
```diff
@@ -149,6 +149,43 @@ def login(self):
...
+        authorization_endpoint = self.metadata["authorization_endpoint"]
+        uri, state_ = self.client.create_authorization_url(authorization_endpoint)
+        response = requests.get(
+            uri,
+            auth=HTTPKerberosAuth(
+                # REQUIRED is not working with id.fedoraproject.org
+                mutual_authentication=OPTIONAL,
+            ),
+        )
+        response.raise_for_status()
```
abompard: I think the `RequestError` should be converted to an `OIDCClientError` for coherence with the rest of the API. No?

abompard (changes requested):
Thanks for the contribution, it looks great!

I am not entirely sure that this code should be in the `oidcclient.py` module, as I intended for this module to be a pure OIDC client that we could one day publish standalone for other apps to use, and Kerberos authentication isn't really part of OIDC.
That said, a Kerberos-capable OIDC client would be useful to other Fedora apps too, so it may indeed be the right place. I need to think about it a bit more, but wherever the code ends up living will not change the API in BodhiClient, so we can merge this and decide later.

Do you think it would be relevant to try to login with kerberos systematically, and fall back to regular login if it fails? I think it would be pretty cool for both apps and users, no?
Maybe then add a flag to the login method to disable the regular browser-based login, so that apps don't get blocked in an interactive process if they don't expect it. What do you think?

Also, please add a release notes fragment as described [here](https://fedora-infra.github.io/bodhi/6.0/developer/index.html#release-notes)

abompard:
Oh one last thing, please run `pre-commit install` so that our hooks run on commit, this should take care of one of the CI errors.

TomasTomecek:
@abompard thank you for a thorough review! I will first address #4603 and then resolve comments here, I hope I'll finish by the end of the week.

bodhi-client/bodhi/client/oidcclient.py
TomasTomecek: I'm actually going to drop this since the other login method doesn't do this check.

I don't remember honestly, it could also be a leftover from debugging :)

TomasTomecek:
```
bodhi/client/oidcclient.py     165      0   100%
```
:)

TomasTomecek:
> Thanks for the contribution, it looks great!
> 
> I am not entirely sure that this code should be in the `oidcclient.py` module, as I intended for this module to be a pure OIDC client that we could one day publish standalone for other apps to use, and Kerberos authentication isn't really part of OIDC. That said, a Kerberos-capable OIDC client would be useful to other Fedora apps too, so it may indeed be the right place. I need to think about it a bit more, but wherever the code ends up living will not change the API in BodhiClient, so we can merge this and decide later.

I'm happy to send that code to a more generic library: this is really up to you :)

> Do you think it would be relevant to try to login with kerberos systematically, and fall back to regular login if it fails? I think it would be pretty cool for both apps and users, no? Maybe then add a flag to the login method to disable the regular browser-based login, so that apps don't get blocked in an interactive process if they don't expect it. What do you think?

That's an excellent idea & I was thinking about it too. It should be implemented now.

> Also, please add a release notes fragment as described [here](https://fedora-infra.github.io/bodhi/6.0/developer/index.html#release-notes)

👍

Edit: CI tests now fail on requests-kerberos not being present: where should I specify this dependency?

abompard:
> Edit: CI tests now fail on requests-kerberos not being present: where should I specify this dependency?

In `devel/ci/Dockerfile-*`

mattiaverga:
> > Edit: CI tests now fail on requests-kerberos not being present: where should I specify this dependency?
> 
> In `devel/ci/Dockerfile-*`

Shouldn't that be added as a dependency to bodhi-client in the pyproject file?

abompard:
> Shouldn't that be added as a dependency to bodhi-client in the pyproject file?

Indeed, but it's been added already.

bodhi-client/bodhi/client/bindings.py
```diff
@@ -245,7 +248,14 @@ def ensure_auth(self):
...
                     self.clear_auth()
-                    self.oidc.login()
+                    # by default, when login_with_kerberos is True, we use Kerberos to authenticate
+                    # and if that fails, fall back to the interactive browser-based OIDC login
+                    try:
+                        if self.login_with_kerberos:
+                            self.oidc.login_with_kerberos()
+                    except OIDCClientError as e:
+                        log.info("Failed to login with Kerberos: %s", e)
+                        self.oidc.login()
```
TomasTomecek, line 258: Please let me know if this is what you had in mind. If so, I'm going to cover this by a test case and resolve the 2 test failures.

bodhi-client/bodhi/client/bindings.py
```diff
@@ -245,6 +248,18 @@ def ensure_auth(self):
...
+                        # This snippet may look weird, but we want to cover 3 use cases:
+                        # 1. login_with_kerberos is True and Kerberos succeeds: continue the loop
+                        # 2. login_with_kerberos is True and Kerberos fails -> normal login follows
+                        # 3. login_with_kerberos is False -> normal login follows
+                        if self.login_with_kerberos:
+                            self.oidc.login_with_kerberos()
+                            continue
+                    except OIDCClientError as e:
+                        log.info("Failed to login with Kerberos: %s", e)
                     self.oidc.login()
```
TomasTomecek: Please let me know if this is what you had in mind. If so, I'm going to cover this by a test case and resolve the 2 test failures.

abompard:
>
> Shouldn't that be added as a dependency to bodhi-client in the pyproject
> file?
>
It's been added there already, no?

TomasTomecek:
rebased

abompard:
Thanks! As nitpicker-in-chief there are a couple things I would have done differently as to where the methods are located, I'm going to push a commit to your PR so that we can talk on something concrete.

TomasTomecek:
> Thanks! As nitpicker-in-chief there are a couple things I would have done differently as to where the methods are located, I'm going to push a commit to your PR so that we can talk on something concrete.

it's your codebase, so please advise :)

abompard (approved):

abompard:
So, I've made a few changes, the most visible one is that you no longer have to instantiate BodhiClient with `use_kerberos_auth=True`, it will always try it (and fallback to browser-based auth if it fails).

Could you try this change with your code please?

TomasTomecek:
@abompard thank you! I will try and report my findings here

TomasTomecek:
I can confirm your changes work just fine :)
```
> /home/tt/g/packit/packit/packit/utils/bodhi.py(61)__init__()                                                                                                                                                                              
     60             import ipdb; ipdb.set_trace()
---> 61             self.oidc.login(use_kerberos=True)
     62             # in our openshift deployment, ~/.config is not writable, but $HOME is

ipdb> n
Login successful!
```

PR #4799, opened 2022-10-17
Fix linters

abompard:
Run the linters from pre-commit, upgrade flake8 and fix the errors.


mattiaverga:
The pip docker still uses F34 as base image, shall we update it to, at least, F36?
Despite all tests pass, I see several errors in the output when pip installing bodhi-*, from a quick look it seems the container misses krb5-devel package?

abompard:
Excellent idea, done. Thanks Mattia.

mattiaverga (approved):
:1st_place_medal:

PR #4814, opened 2022-10-20
Hide input when logging in via browser

mfocko:
Signed-off-by: Matej Focko <mfocko@redhat.com>


mattiaverga (changes requested):
Thanks for this, can you add a note to release notes as described [here](https://fedora-infra.github.io/bodhi/6.0/developer/index.html#release-notes)?

mattiaverga (approved):

PR #4817, opened 2022-10-21
Relax dependencies requirements

mattiaverga:
We are currently hard pinning dependencies to specific versions in `pyproject.toml` files. Dependabot only updates `poetry.lock` files, but not `pyproject.toml`, so we're actually testing no changes because the `pip` tests will always install hard pinned dependencies, while other `f*` tests will install packages from Fedora RPMs.

This is also currently breaking the vagrant deployment.


mattiaverga:
I'm going to merge this now, as I manually updated poetry.lock and dependabot will likely update it soon.

Issue #4704, opened 2022-08-19
Document DCO, Signed-off-by requirement


LenkaSeg:
Suggested by @gotmax23

Add DCO document to the repo and a note to file for contributors.

Example from rpmautospec:
https://pagure.io/fork/nphilipp/fedora-infra/rpmautospec/c/8355b525cc20a27dbe3e9af25e7c6f6ff881b55d?branch=main--random-fixes-and-additions

Related PR #4677

ronitblenz:
May I work on this issue?
Please assign me so that I can look into it.

mattiaverga:
For reference, this is already in the [developer documentation](https://github.com/fedora-infra/bodhi/blob/develop/docs/developer/index.rst#contribution-guidelines).

ronitblenz:
I have made a PR, Kindly check

PR #4820, opened 2022-10-28
DCO Documentation

ronitblenz:
Signed-off-by: Ronit Banerjee <ronitbanerjee03@gmail.com>

Document DCO, Signed-off-by requirement added to the Bodhi project.

Fixes Issue #4704


mattiaverga (changes requested):
Thanks.
In the README there is already a section about contributing to Bodhi in the Documentation paragraph. Maybe that section can be moved under Contributing and you can modify the docs with the text you put in the README (since we're going to have the DCO file in the main dir, I don't think it's useful to have the whole text reported in docs).
What do you think?

ronitblenz:
Yeah sure, Let me make those changes.

PR #4951, opened 2023-01-11
Update schemas

### LINKED ISSUE #4950: Update message schema to include __str__, agent_name, groups, and app_name
Currently the bodhi messaging schema is missing a handful of properties for proper compatibility with upcoming fmn work.

* __str__ -- two messages currently include the __str__ property (which is used to Return a human-readable representation of a message, but longer than the summary i.e. an email body). we need to implement __str__ in the majority of messages:
* agent_name -- we currently already have agent -- just need to expose this as agent_name for full compatability
* app_name -- pretty self explainatory

finally, we also have

* groups -- which i dont think we need to implement in bodhi, as it isnt really aware of FAS groups

ryanlerch:
Resolves #4950


bodhi-messages/bodhi/messages/schemas/update.py
```diff
@@ -208,6 +220,23 @@ def summary(self) -> str:
...
+
+        Returns:
+            A human readable representation of this message.
+        """
+        new_line = "\n"
+        return (
+            f"{self.update.user.name}'s Bodhi update {self.update.alias}"
+            f"completed push to {self.update.status}\n"
+            f"Builds:\n"
+            f"{new_line.join([b.nvr for b in self.update.builds])} ")
```
mattiaverga, line 238: Is there any limit to a message length? What will happen with updates composed by hundreds of builds?

bodhi-messages/bodhi/messages/schemas/update.py
```diff
@@ -243,6 +272,23 @@ def summary(self) -> str:
...
```
mattiaverga, line 290: Same as above comment.

mattiaverga (commented):
Can you drop a change note in the news directory?

bodhi-messages/bodhi/messages/schemas/update.py
ryanlerch: I don't think there is -- @abompard might be able to weigh in on this though.

the spec states that the __str__ is supposed to be like the body of an email (with the summary being the subject) -- so IMHO, its appropriate to list everything here in the __str__

ryanlerch:
> Can you drop a change note in the news directory?

Done!

mattiaverga (approved):

PR #5154, opened 2023-03-17
Copy additional config files for pungi

Zlopez:
When trying to create shared configs for pungi in bodhi deployment it failed to copy them and thus composes failed to start. This change copies over any additional pungi *.conf file that is in pungi.basepath.

See https://pagure.io/fedora-infrastructure/issue/10779 for additional info.


Zlopez:
Only the rawhide integration tests are failing, but looking at other PRs this is common.

nirik:
Looks pretty reasonable to me. Is there any documentation that needs updated?

@mattiaverga it would be nice to get this in asap and cut a minor release... I can then update it in prod before final f38 freeze and we can use these split pungi files. ;)

mattiaverga (approved):

mattiaverga:
@nirik I've pushed 7.1.1 out and cooked koji builds ;-)

PR #5172, opened 2023-03-27
Add frequencyupdateinfo.json artifact to release 

RishabhSaini:
Detailed rationale in:
https://github.com/coreos/rpm-ostree/issues/4247#issuecomment-1467819269

### DIFF
--- a/bodhi-server/bodhi/server/metadata.py
+++ b/bodhi-server/bodhi/server/metadata.py
@@ -95,6 +95,73 @@ def modifyrepo(comp_type, compose_path, filetype, extension, source, zchunk):
             repodata = os.path.join(repo_path, arch, 'os', 'repodata')
         insert_in_repo(comp_type, repodata, filetype, extension, source, zchunk)
 
+class FrequencyUpdateInfoMetadata(object):
+    """
+    This class represents the frequencyupdateinfo.json metadata
+    """
+
+    def __init__(release, update, update_status):
+        """
+        Initialize the FrequencyUpdateInfoMetadata object.
+
+        Args:
+            release (bodhi.server.models.Release): The Release that is being composed.
+            update (bodhi.server.models.Update):
+            update_status (bodhi.server.models.UpdateStatus):
+        """
+        updates = update.query.filter(
+            update.status.in_(
+                [update_status.stable])
+        ).filter(
+            update.release_id == release.id
+        ).order_by(
+            update.date_submitted
+        )
+        modified_updates = []
+        for update in updates:
+            simplified_update = {
+                "release" : {
+                    "name" : update.release.name, 
+                    "version" : update.release.version,
+                },
+                "alias" : update.alias,
+                "date_stable" : update.date_stable,
+                "builds" : [{"nvr_name" : build.nvr_name()} for build in update.builds],
+            } 
+            modified_updates.push(simplified_update)
+        self.comp_type = cr.XZ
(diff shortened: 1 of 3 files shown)


bodhi-server/bodhi/server/metadata.py
```diff
@@ -95,6 +95,73 @@ def modifyrepo(comp_type, compose_path, filetype, extension, source, zchunk):
...
+        updates = update.query.filter(
+            update.status.in_(
+                [update_status.stable])
+        ).filter(
+            update.release_id == release.id
+        ).filter(
+            release.state.in_([
+                release_state.current,
+                release_state.pending,
+            ])
```
mattiaverga: I don't think this is needed, as the previous filter is against `release.id`.

bodhi-server/bodhi/server/metadata.py
```diff
@@ -95,6 +95,73 @@ def modifyrepo(comp_type, compose_path, filetype, extension, source, zchunk):
...
+        ).order_by(
+            # Check the older updates first so there is more time for the newer to
+            # get their test results
+            update.id.asc()
```
mattiaverga: I think it is best to order by `date_submitted`

bodhi-server/bodhi/server/metadata.py
```diff
@@ -95,6 +95,73 @@ def modifyrepo(comp_type, compose_path, filetype, extension, source, zchunk):
...
+                    "version" : update.release.version
+                }
+                "alias" : update.alias
+                "date_stable" : update.date_stable
+                "builds" : [
+                ]
+            } 
+            for build in update.builds:
+                simplified_build = {"nvr_name" : build.nvr_name()}
+                simplified_update["builds"].push(simplified_build)
```
mattiaverga: `"builds" : [{"nvr_name" : build.nvr_name()} for build in update.builds]`

And missing commas between dict elements.

bodhi-server/bodhi/server/metadata.py
```diff
@@ -95,6 +95,73 @@ def modifyrepo(comp_type, compose_path, filetype, extension, source, zchunk):
...
+            modified_updates.push(simplified_update)
+        self.comp_type = cr.XZ
+        self.zchunk = True
```
mattiaverga: Could we share settings with UpdateInfoMetadata?
Otherwise we should copy them (AFAIK, zchunk is disabled for EPEL releases as it won't work)

bodhi-server/bodhi/server/models.py
```diff
@@ -757,6 +757,7 @@ class ComposeState(DeclEnum):
     pending = 'pending', 'Pending'
     initializing = 'initializing', 'Initializing'
     updateinfo = 'updateinfo', 'Generating updateinfo.xml'
+    frequencyupdateinfo = 'frequencyupdateinfo', 'Generating frequency update info'
```
mattiaverga: This will need a database migration file, I think.

mattiaverga (changes requested):
Some comments inline from a quick look.
Also, missing tests and release notes file.

RishabhSaini:
No longer need to generate artifact for getting comprehensive list of updates. RPM headers `RPMTAG_CHANGELOGTIME` was used instead
Closing due to: https://github.com/coreos/rpm-ostree/issues/4247#issuecomment-1511759881

PR #5303, opened 2023-05-09
Some Fedora Messaging schema fixes

abompard:
This PR changes two elements of the Bodhi schemas that I found annoying when subscribed to them:

---
image
→ the two messages always arrive together, so one should not generate a notification. For that, I set the most obscure one to severity DEBUG

--- 

image
→ The subject and the body aren't particularly helpful. I've added the build names to the subject and defined a message body.

### DIFF
--- a/bodhi-messages/tests/test_update.py
+++ b/bodhi-messages/tests/test_update.py
@@ -17,21 +17,23 @@
 
 
 from bodhi.messages.schemas.base import BuildV1, ReleaseV1, UpdateV1, UserV1
-from bodhi.messages.schemas.update import (UpdateCommentV1,
-                                           UpdateCompleteStableV1,
-                                           UpdateCompleteTestingV1,
-                                           UpdateEditV1,
-                                           UpdateEditV2,
-                                           UpdateEjectV1,
-                                           UpdateKarmaThresholdV1,
-                                           UpdateReadyForTestingV1,
-                                           UpdateReadyForTestingV2,
-                                           UpdateRequestObsoleteV1,
-                                           UpdateRequestRevokeV1,
-                                           UpdateRequestStableV1,
-                                           UpdateRequestTestingV1,
-                                           UpdateRequestUnpushV1,
-                                           UpdateRequirementsMetStableV1)
+from bodhi.messages.schemas.update import (
+    UpdateCommentV1,
+    UpdateCompleteStableV1,
+    UpdateCompleteTestingV1,
+    UpdateEditV1,
+    UpdateEditV2,
+    UpdateEjectV1,
+    UpdateKarmaThresholdV1,
+    UpdateReadyForTestingV1,
+    UpdateReadyForTestingV2,
+    UpdateRequestObsoleteV1,
+    UpdateRequestRevokeV1,
+    UpdateRequestStableV1,
+    UpdateRequestTestingV1,
+    UpdateRequestUnpushV1,
+    UpdateRequirementsMetStableV1,
+)
 
 from .utils import check_message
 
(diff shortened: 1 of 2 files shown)


mattiaverga:
I'm quite sure the `UpdateReadyForTesting` message is consumed by CI to start tests (it is also the message used to resubmit tests from webUI). So I don't think we can set it to DEBUG level. @AdamWill should know better.

I think the rationale to have these two different messages is that Rawhide updates never emit `UpdateCompleteTesting` messages, so they need a `UpdateReadyForTesting` to start CI tests.

AdamWill:
Yes, that's right. Both Fedora CI and openQA use the `UpdateReadyForTesting` message as a test trigger, and it is also sent (with `re-trigger: True`) when the Re-Trigger Tests button in Bodhi is clicked, which both systems also use as a trigger.

As Mattia says, these two messages do not in fact always get sent together. As [this docstring says](https://github.com/fedora-infra/bodhi/blob/7.2/bodhi-server/bodhi/server/models.py#L4261), for Rawhide, `UpdateReadyForTesting` is sent when "all packages in an update have been built by koji" (which in practice usually means it's sent immediately on update creation, as updates are usually created for builds that have already happened). I don't think any Rawhide update ever gets a `UpdateCompleteTesting` message, actually, because no Rawhide updates are ever "pushed to testing" - even the occasional Rawhide update which fails a package-specific gating policy and so winds up in the Bodhi status "testing", e.g. [this one currently](https://bodhi.fedoraproject.org/updates/FEDORA-2023-ddf133d189), don't actually go through a "testing push", so that message is never published for them.

We could definitely stand to rejig all of this, but not like this.

openQA really only needs a message for "an update has been created", because it doesn't care about repository pushes - it just asks Bodhi what builds are in the update and downloads them from Koji. openQA's update test scheduling rules boil down to:

1. Test new updates as soon as possible after they are created
2. Re-test an update if its builds are edited
3. Re-test an update if the Re-Trigger Tests button is hit

The [current implementation of this](https://pagure.io/fedora-qa/fedora_openqa/blob/schedule-cleanup-2023/f/src/fedora_openqa/consumer.py#_40) is a bit complicated (even after I cleaned it up some recently) because of how the messages we actually get differ for Rawhide and non-Rawhide updates. In practice, Rawhide updates pretty much always get a `UpdateReadyForTesting` message immediately upon creation, and non-Rawhide updates almost always get a `UpdateRequestTesting` message immediately on creation, unless the update submitter does something fairly weird which results in the update not being submitted to updates-testing.

I don't know what CI's rules are.

abompard:
Thanks for the detailed explanation! However setting the severity of a message in Fedora Messaging does not mean that the message is only sent when bodhi runs with logging set to DEBUG. The message is **always** sent. It's just that FMN looks at the severity property and by default will not notify users when the severity is DEBUG (users can opt in to that).
If OpenQA does not explicitely look at the severity property, it will not see any change.

bodhi-messages/tests/test_update.py
```diff
@@ -635,6 +644,8 @@ def test_requirements_met_stable_v1(self):
                 },
             }
         )
+        print(repr(str(msg)))
```
mattiaverga: I think these `print` aren't part of the change?

mattiaverga (commented):
Thanks for clarify the usage of `DEBUG` setting, so I think this can be merged if Adam haven't got any objections.

bodhi-messages/tests/test_update.py
abompard: Ah crap I thought I got rid of them. Thanks!

mattiaverga (approved):
Rawhide tests are currently broken due to poetry being FTI https://bugzilla.redhat.com/show_bug.cgi?id=2196879

F38 Integration tests are expected to fail (I still couldn't figure why, though).

PR #5454, opened 2023-08-13
Add ACL for QA Engineers

### LINKED ISSUE #5448: Add specific ACL for QA related activities
The "waive tests" and the "re-trigger tests" services are currently using the same ACLs used for allow editing an update.
I suppose we can add a new ACL to give members of `qa` group access to those functions, without the need to have them allowed for full editing.

mattiaverga:
Define separate ACLs to allow QA Engineers to trigger/waive tests on all updates.


AdamWill:
This looks pretty good to me. Thanks!

PR #5455, opened 2023-08-13
Support compressed comps.xml created with createrepo_c >= 1.0

mattiaverga:
createrepo_c >= 1.0 will also compress comps.xml if a compression method is set.
This PR will add support to the sanity_check_repodata method for reading compressed file, with support for all the different compression formats that can be set on createrepo_c.


bodhi-server/bodhi/server/util.py
```diff
@@ -320,7 +321,11 @@ def sanity_check_repodata(myurl, repo_type):
             # Test comps
             comps = libcomps.Comps()
             try:
-                ret = comps.fromxml_f(repo_info['group'])
+                if repo_info['group'].endswith('xz'):
+                    xml_content = lzma.open(repo_info['group']).read()
+                    ret = comps.fromxml_str(xml_content.decode())
```
dralley, line 330: You might as well just support all the options.

```
                if repo_info['group'].endswith('xz'):
                    xml_content = lzma.open(repo_info['group']).read()
                    ret = comps.fromxml_str(xml_content.decode())
                elif repo_info['group'].endswith('gz'):
                    xml_content = gzip.open(repo_info['group']).read()
                    ret = comps.fromxml_str(xml_content.decode())
                elif repo_info['group'].endswith('zst'):
                    xml_content = zstd.open(repo_info['group']).read()
                    ret = comps.fromxml_str(xml_content.decode())
                elif repo_info['group'].endswith('bz2'):
                    xml_content = bz2.open(repo_info['group']).read()
                    ret = comps.fromxml_str(xml_content.decode())
                else:
                    ret = comps.fromxml_f(repo_info['group'])
```

(I haven't tested that code)

mattiaverga:
Tests will need to wait until librepo 0.1.20 is pushed to stable F39 and F38

mattiaverga:
F38 and Rawhide tests now are fine, so the problem should be fixed. STill waiting for F39 libcomps update to hit stable, but I'm going to merge this PR

PR #5456, opened 2023-08-16
Replace dependabot with renovate

LenkaSeg:


LenkaSeg:
I wonder if the shared renovate config will be suitable for monorepo...

mattiaverga (approved):
Let's try

PR #5482, opened 2023-09-06
Show more data for Release

### LINKED ISSUE #5481: Add more information to releases
Since PDC is being retired and Bodhi is soon to become the source of truth for release info, I think it would be nice to have some more info:
- pre-beta/beta status: I think this state could be computed on the fly looking at the `fxx.status` in the config file
- release date: this would need to add a column to the Release model, the release date could be set automatically the first time the release status is changed to `current`

Do we need anything else to replace PDC?

mattiaverga:
Fixes  #5481

### DIFF
--- a/bodhi-client/bodhi/client/cli.py
+++ b/bodhi-client/bodhi/client/cli.py
@@ -17,6 +17,7 @@
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
 """The bodhi CLI client."""
 
+from datetime import datetime
 import functools
 import logging
 import os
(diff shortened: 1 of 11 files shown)


bodhi-client/bodhi/client/cli.py
```diff
@@ -17,6 +17,7 @@
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
 """The bodhi CLI client."""
 
+from datetime import datetime
```
Maxwell0010, line 20: tryh

Maxwell0010 (changes requested):

bodhi-client/bodhi/client/cli.py
mattiaverga: is that supposed to be an acronym?

PR #5484, opened 2023-09-08
Provide an endpoint to list critpath components

mattiaverga:
Let's have a way to provide information about what Bodhi consider to be in critical path.
WIP: I need to write some tests

@AdamWill If a user searches for `collection=rawhide` or `collection=f40` we need the code to redirect to `master.json`. Do we have convenient symlinks set up? Otherwise I can do that in [read_critpath_json](https://github.com/fedora-infra/bodhi/blob/f3342ba276aecaa489b7f7606412ed4c8489d479/bodhi-server/bodhi/server/util.py#L974)

### DIFF
--- a/bodhi-server/bodhi/server/__init__.py
+++ b/bodhi-server/bodhi/server/__init__.py
@@ -297,6 +297,9 @@ def main(global_config, testing=None, session=None, **settings):
     config.add_route('liveness', '/healthz/live')
     config.add_route('readyness', '/healthz/ready')
 
+    # service endpoints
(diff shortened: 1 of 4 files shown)


AdamWill:
It's not `master.json`, it's `rawhide.json`. When you trace it all out, `collection` in this context is the git branch name. No, the script does not currently generate symlinks...I kinda don't mind whether we have the script create symlinks or just make `read_critpath_json` translate...

https://pagure.io/releng/blob/main/f/scripts/critpath.py#_345

PR #5496, opened 2023-10-02
Container flatpak detection

owtaylor:
This PR fixes a problem with the new system of building Flatpaks directly with on the builders, rather than using OSBS. The new koji-flatpak plugin doesn't set 'container_koji_task_id' in the extra metadata for the build, which broke content type detection.

See https://pagure.io/releng/issue/11697

There would be more simple ways of fixing things, but I tried to clean up a bit, rather than just leaving things broken so that unknown image types (all images aren't containers or Flatpaks...) were detected as RPM packages.

The cleanup only goes so far - I've left the test case for the new error catching a 500, rather than trying to cleanly catch this in the validatators. This was because I was pretty stumped at how to fit it in - there are multiple code paths that directly or indirectly call infer_content_class() (at least validate_acls and validate_build_uniqueness).

I've also switched the code from extra['image'] to extra['typeinfo']['image'], since that's been the preferred location for the metadata for quite some time.

### DIFF
--- a/bodhi-server/bodhi/server/buildsys.py
+++ b/bodhi-server/bodhi/server/buildsys.py
@@ -248,18 +248,20 @@ class DevBuildsys:
 
                 data['extra'] = {
                     'container_koji_task_id': 19708268,
-                    'image': {
-                        'index': {
-                            'pull': ['{registry}/{repository}@sha256:{hash}'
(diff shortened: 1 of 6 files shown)


mattiaverga (approved):
I've added the news fragment, approving now

PR #5497, opened 2023-10-02
Handle <version>-<release> strings that aren't valid container tags

owtaylor:
With the new Flatpak generation inside of Koji, we are versioning the containers based on their main packages. So,

  `ghex-45~beta-2.fc39` => `ghex-flatpak-45~beta-1`

But '45~beta-1' is not a valid registry tag, so pushing fails. This PR has two parts:

 1. Don't count on the tags in the candidate registry having any particular value - pull by digest instead
 2. Make sure the tags we create in the destination registry are valid


mattiaverga (approved):
Rebased and added news fragment

PR #5521, opened 2023-10-29
Allow setting createrepo_c config per release

### LINKED ISSUE #5517: Define compose settings per Release
Currently Bodhi hardcodes the compression method and the zchunk setting in `metadata.py` with the following settings:

- compression: XZ; zchunk: enabled. Default for all releases
- compression: BZ2; zchunk: disabled. Override for all releases with a prefix starting with `FEDORA-EPEL`

Historically, this was due to the fact RHEL7 used yum and had no gain in having zchunk enabled. I'm not sure about the compression method choice.
Anyway, the idea here is to move those settings to be a per Release settings, either defined in database as a property of the Release object or in the config file (or in a separate config file, like critpath definitions).

mattiaverga:
Fixes #5517 
I've opted for defining per Release configuration or with a prefix based configuration like we do now using an external config file, rather than extending the database.

### DIFF
--- a/bodhi-server/bodhi/server/config.py
+++ b/bodhi-server/bodhi/server/config.py
@@ -351,6 +351,9 @@ class BodhiConfig(dict):
         'cors_origins_rw': {
             'value': 'https://bodhi.fedoraproject.org',
             'validator': str},
+        'createrepo_c_config': {
+            'value': '/etc/bodhi/createrepo_c.ini',
+            'validator': str},
         'critpath_pkgs': {
             'value': [],
             'validator': _generate_list_validator()},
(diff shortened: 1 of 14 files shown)


humaton:
Finally got to this, trying to get it running locally.

humaton (approved):
So, the CI fail is there just because 39 is on freeze and the latest build of libcomps-0.1.20-1.fc39 is only in testing.

I was able to run the changes locally, it looks ok. 
Not sure if my review counts but ack from me.

mattiaverga:
> So, the CI fail is there just because 39 is on freeze and the latest build of libcomps-0.1.20-1.fc39 is only in testing.
> 
> I was able to run the changes locally, it looks ok. Not sure if my review counts but ack from me.

Any review is always appreciated, thanks!

PR #5525, opened 2023-11-05
Fix regex for username mentions

### LINKED ISSUE #5453: @username doesn't work if username contains a hyphen
I attempted to reply in Bodhi to user "roman-averin" using the @ sign, but it only recognizes the part of the username before the hyphen. For example, see https://bodhi.fedoraproject.org/updates/FEDORA-2023-ee241dcf80 . I saw no way to make it work so just removed my attempt. If there is a way to do it, it needs to be documented. The link to Fedora-Flavored Markdown says nothing.

mattiaverga:
Fixes #5453


abompard:
It would be awesome if you could add a unit test… :-)

PR #5537, opened 2023-11-21
webui: add 'For help debugging' message for FCOS tests

jlebon:
We've added new `coreos.*` tests that now show up in the Bodhi UI for
some packages. These tests are not yet gating, but to be ready for when
they are, let's be nice and link to the #coreos Matrix channel where
packagers can get debugging help like for the other ones.

While we're here, add final periods to each sentence.


AdamWill:
Hum, we probably want to update the Fedora CI link to https://matrix.to/#/#fedora-ci:fedoraproject.org now too...

jlebon:
Sure, tweaked in a separate commit!

mattiaverga (approved):

PR #5572, opened 2023-12-22
Have the authentication policy honor config settings

mattiaverga:


AdamWill:
Ideally this should also include a test, right? I believe there was a test before #5288 , which 5288 removed.

mattiaverga:
Re added tests

AdamWill:
LGTM.

mattiaverga:
Cool, I'll merge this in the 8.0 and push a 8.0.1 out since this is really an urgent bug.

AdamWill:
I guess it's a good thing I got completely lost in the auth flow trying to get logout / login to work in the dev env! In the end I couldn't even get that to work so I had to hack in a workaround, but at least I found this, heh.

PR #5686, opened 2024-06-19
Fix enum initialization in python 3.13

### LINKED ISSUE #5685: Bodhi enums implementation incompatible with Python 3.13
This code, which was taken from an old recipe at http://techspot.zzzeek.org/2011/01/14/the-enum-recipe, doesn't work anymore in Python 3.13:
```
class EnumSymbol(object):
    def __init__(self, cls_, name, value, description):
        self.cls_ = cls_
        self.name = name
        self.value = value
        self.description = description

class EnumMeta(type):
    def __init__(cls, classname, bases, dict_):
        cls._reg = reg = cls._reg.copy()
        for k, v in dict_.items():
            if isinstance(v, tuple):
                sym = reg[v[0]] = EnumSymbol(cls, k, *v)
                setattr(cls, k, sym)
        return type.__init__(cls, classname, bases, dict_)

class DeclEnum(metaclass=EnumMeta):
    _reg = {}
```

bodhi-server is currently FTB in F41.

mattiaverga:
fix #5685

### DIFF
--- a/bodhi-server/bodhi/server/models.py
+++ b/bodhi-server/bodhi/server/models.py
@@ -187,7 +187,7 @@ class EnumMeta(type):
         """
         cls._reg = reg = cls._reg.copy()
         for k, v in dict_.items():
-            if isinstance(v, tuple):
+            if isinstance(v, tuple) and v:
                 sym = reg[v[0]] = EnumSymbol(cls, k, *v)
                 setattr(cls, k, sym)
         return type.__init__(cls, classname, bases, dict_)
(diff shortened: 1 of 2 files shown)


mattiaverga:
This should be considered just a quick fix to make Bodhi build again on Fedora 41.
I'm unsure if this change could have side effects as I don't quite understand all the logic behind the custom enum implementation (the code itself seems to have been taken from an online recipe).
 
Long term, I'll look to move bodhi from using a custom enum implementation to standards.

Issue #5714, opened 2024-07-22
Bodhi still has the old notification service linked under "Manage Alerts" 


markrosenbaum:
On bodhi.fp.o it looks like the management UI for the old notification service is still linked under "Manage Alerts" creating a dead link.

<img width="1280" alt="Screen Shot 2024-07-22 at 12 29 13 PM" src="https://github.com/user-attachments/assets/4dbb4b25-51a7-49d4-8a48-120a88565829">

<img width="1268" alt="Screen Shot 2024-07-22 at 12 28 52 PM" src="https://github.com/user-attachments/assets/647f142b-e51c-4f30-951e-dc79ec0129a9">

markrosenbaum:
Caused by 
https://github.com/fedora-infra/bodhi/blob/c9efd320b73af8e0c6b9f5f42cc0330c921afd9f/bodhi-server/bodhi/server/templates/master.html#L115C60-L115C132
New service does not use OpenID

PR #5717, opened 2024-07-22
Resolving https://github.com/fedora-infra/bodhi/issues/5714

markrosenbaum:


mattiaverga:
Hello, thanks for the patch and sorry for the late reply.

To have the patch approved, can you please sign the DCO? You can click on the "details" link in the checks table for an explanation and how to do so.

It would also be nice if you can drop a note in the `news` directory as explained [here](https://fedora-infra.github.io/bodhi/8.0/developer/index.html#release-notes).

PR #5764, opened 2024-10-15
Validate side-tag update as regular one if the submitter is not side-tag creator

nforro:
The point of this change is to allow a user who has commit rights to all builds in an update to edit the update, even though the creator of the update and of the associated side-tag is a different user.

Users have been able to edit/waive test failures on regular updates created by Packit provided they had commit access to all the builds. This doesn't work with side-tag updates, where Packit is the creator of the side-tag. I don't see a reason why the validation couldn't allow both, the sidetag creator and the maintainer of all packages, to make changes.

### DIFF
--- a/bodhi-server/bodhi/server/validators.py
+++ b/bodhi-server/bodhi/server/validators.py
@@ -455,9 +455,9 @@ def validate_acls(request, **kwargs):
     former update submitter doesn't have commit rights, the original submitter
     will lose their privs to edit the update.
 
-    In case of a side-tag update, we don't validate against the builds,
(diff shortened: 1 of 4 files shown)


mattiaverga (approved):

nforro:
Thanks!

lsm5:
@nforro thank you so much!!

ETA on deployment?

mattiaverga:
> ETA on deployment?

Probably a couple of weeks after the end of infrastructure freeze

mattiaverga:
This is now deployed in prod ;-)

nforro:
Excellent, thanks!

PR #5786, opened 2024-11-08
Change precommit to use python 3.13

mattiaverga:
F41 has python 3.13.
This should fix pre-commit tests


mattiaverga:
... and also fix the regex used in integration tests to match "EPEL-10.0" (the dot in release name is new)

PR #5805, opened 2024-11-20
better show the difference between autopush settings and thresholds

mattiaverga:
The details of an update page now better distinguish settings for the automatic push and the thresholds for manual push.
I've also moved the build list at the bottom of the details, because for updates with many builds it was needed to scroll the page a lot to find out useful information.

Example screenshot:
immagine


mattiaverga:
immagine

immagine

@AdamWill thoughts?

AdamWill:
yeah, I like it! I wanted to do something like this, but didn't get the roundtuits. Thanks.

PR #5882, opened 2025-04-10
generic.py: bump to 10 top testers and packagers

### LINKED ISSUE #5881: Web: list more active testers and packagers
Currently Bodhi Web lists the top 5 most active testers and packagers: I think it is a great feature.
But 5 seems a bit small, and it doesn't seem to change much over time, probably due to big gaps.

How about extending it to the top 10 at least, or even 20?
I think showing a bit more people would help encourage more participation and long-term recognition.

There could even be a separate page with a longer leaderboard, like top 50 or 100, but even just extending the current list a bit would already be a good improvement.

juhp:
closes #5881

### DIFF
--- a/bodhi-server/bodhi/server/config.py
+++ b/bodhi-server/bodhi/server/config.py
@@ -395,6 +395,9 @@ class BodhiConfig(dict):
         'fmn_url': {
             'value': 'https://apps.fedoraproject.org/notifications/',
             'validator': str},
+        'homepage_stats_num_users': {
+            'value': 10,
+            'validator': int},
         'important_groups': {
             # Defined in and tied to the Fedora Account System (limited to 16 characters)
             'value': ['proventesters', 'provenpackager', 'releng', 'security_respons', 'packager',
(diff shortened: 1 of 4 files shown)


mattiaverga (changes requested):
What about making the value configurable instead of hardcode it?
You can add a `homepage_stats_num_users` setting into `config.py` with a default value of 10 (better to add a commented out line into `production.ini` too for reference) and use that value from config into the code.

Also, don't forget to add DCO signing to commit.

juhp:
Finally got round to updating this, thanks

mattiaverga (approved):
Rebased, fixed a missing comma in config file and dropped a release note.

juhp:
Btw today is noticed the numbers displayed on https://bodhi.fedoraproject.org/ are not updating it seems?

If you click through on individuals you can see the total number of their updates does not match what is displayed: or is it a different number?  Maybe the values are cached indefinitely?

mattiaverga:
> Btw today is noticed the numbers displayed on https://bodhi.fedoraproject.org/ are not updating it seems?
> 
> If you click through on individuals you can see the total number of their updates does not match what is displayed: or is it a different number? Maybe the values are cached indefinitely?

The numbers in the homepage are referred to last x days (I don't remember exactly how many, at the moment), they're not the total.

PR #5883, opened 2025-04-11
Correct spelling of “log in”

mavit:
The verb is “log in”; “login” is the noun.


mattiaverga (changes requested):
The commit is missing DCO signing, otherwise looks good to me (as a foreign language speaker, I didn't notice the difference ;-) )

mattiaverga (approved):
Thanks!

PR #5917, opened 2025-07-14
Add user setting for disabling email notifications

mattiaverga:


mattiaverga:
I want to try to make this be set per Update rather than globally... as soon as I find some time to put in the effort.

PR #6058, opened 2026-01-09
Refactor(ipsilon): Encapsulate stack inspection logic

WillyEverGreen:
Encapsulate the stack inspection hack used to retrieve the root object into a dedicated helper method for better maintainability and documentation.

### DIFF
--- a/devel/ci/integration/ipsilon/api.py
+++ b/devel/ci/integration/ipsilon/api.py
@@ -17,11 +17,35 @@ class OpenidExtension(OpenidExtensionBase):
         super(OpenidExtension, self).__init__("insecureAPI")
 
     def enable(self):
-        # This is the most ugly hack in my history of python...
-        # But I need to find the root object, and that is not passed into
-        #  the OpenID extension system anywhere...
-        root_obj = inspect.stack()[5][0].f_locals["self"]
+        """
+        Enable the OpenID extension.
+
+        This method retrieves the root application object to register the API page.
+        """
+        root_obj = self._retrieve_root_object()
         root_obj.api = APIPage(root_obj)
+
+    def _retrieve_root_object(self):
+        """
+        Retrieve the Ipsilon root application object.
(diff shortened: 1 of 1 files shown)


devel/ci/integration/ipsilon/api.py
```diff
@@ -17,12 +17,27 @@ def __init__(self, *pargs):
...
+        Retrieve the Ipsilon root application object.
+
+        This method inspects the call stack to find the root object, as it is not
+        directly passed to the extension's enable method by the framework.
+        """
+        # Note: This approach uses stack inspection to locate the 'self' variable
+        # from the caller's frame, which corresponds to the root application object.
+        # This is necessary because the OpenID extension system does not currently
+        # expose the root object to extensions.
+        return inspect.stack()[6][0].f_locals["self"]
```
mattiaverga, line 48: Thanks for the cleaning up! Just one question as I'm not really familiar with OpenID: the old code was looking at `inspect.stack()[5]`, in the new you put `inspect.stack()[6]`... why the change?

WillyEverGreen: Good catch! Yeah, the `[5]` to `[6]` change is because wrapping this in a helper method adds an extra stack frame.

Before, when the code was inline, `inspect.stack()[5]` went straight to the root application object. Now that it's in `_get_root_object()`, that method call is an extra frame in the stack, so we need `[6]` to get to the same place.

Basically: helper method frame + the original 5 frames = 6 total.

I should have added a comment explaining this, will add one now!

mattiaverga: Thanks! Can you squash the changes in just one commit and apply the DCO signing?

WillyEverGreen:
Done! Squashed into a single commit with DCO signing. Ready for review.

mattiaverga (approved):
Thanks!

PR #6077, opened 2026-02-04
Support multiple destination registries

Zlopez:
With the move of flatpaks from registry.fedoraproject.org to quay.io (see https://forge.fedoraproject.org/infra/tickets/issues/11543 for more details) we want the bodhi to support uploading to multiple registries.

This PR changes the container.destination_registry to list and when copying from candidate registry it copies to all the destinations.

### DIFF
--- a/bodhi-server/bodhi/server/config.py
+++ b/bodhi-server/bodhi/server/config.py
@@ -337,8 +337,8 @@ class BodhiConfig(dict):
             'value': True,
             'validator': _validate_bool},
         'container.destination_registry': {
-            'value': 'registry.fedoraproject.org',
-            'validator': str},
+            'value': ['registry.fedoraproject.org', 'quay.io/fedora/flatpak'],
+            'validator': _generate_list_validator()},
         'container.source_registry': {
             'value': 'candidate-registry.fedoraproject.org',
             'validator': str},
--- a/bodhi-server/bodhi/server/config.py
+++ b/bodhi-server/bodhi/server/config.py
@@ -337,7 +337,7 @@ class BodhiConfig(dict):
             'value': True,
             'validator': _validate_bool},
         'container.destination_registry': {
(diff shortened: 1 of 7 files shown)


Zlopez:
I think this is ready for review now. I can see the integration tests are failing on the same issue in other PRs as well and the readthedocs build seems completely unrelated to the changes.

bodhi-server/bodhi/server/config.py
```diff
@@ -337,8 +337,8 @@ class BodhiConfig(dict):
             'value': True,
             'validator': _validate_bool},
         'container.destination_registry': {
-            'value': 'registry.fedoraproject.org',
-            'validator': str},
+            'value': ['registry.fedoraproject.org', 'quay.io/fedora/flatpak'],
```
mattiaverga: Production bodhi configuration in ansible just uses the defaults, so I'm unsure if we want to change these... maybe it's better to just change the value to a list with only registry.fp.o.

For staging bodhi the value is set to
```
{% if env == 'staging' %}
container.destination_registry = registry.stg.fedoraproject.org
container.source_registry = candidate-registry.stg.fedoraproject.org
{% endif %}
```
shall we use 'quay.io/fedora/flatpak' or is there any staging destination?

news/PR6077.feature
```diff
@@ -0,0 +1 @@
+Support multiple destination registries for containers
```
mattiaverga: better use the suffix .bic (backward incompatible change) for the file and specify in the notes that the `container.destination_registry` setting can now be specified as whitespace separated list of values

bodhi-server/bodhi/server/config.py
Zlopez: There is a staging target `quay.io/fedora-testing/flatpak`, I can use this.

news/PR6077.feature
Zlopez: Will do, it makes sense.

mattiaverga (approved):

PR #6078, opened 2026-02-04
Allow querying updates by side tag name

nforro:
Packit allows to automate creating multi-package updates by using side tags. However, since the lifecycle of a side tag doesn't necessarily correspond to the lifecycle of the update created from it, it is necessary for Packit, as side tag creator, to be able to query updates corresponding to a particular side tag so that it can act accordingly when an update has been obsoleted or unpushed.

In order not to break existing API, this commit adds a new field, `from_side_tag_name`, that allows to query for updates created from the specified side tag.

Alternatively, the existing `from_side_tag` field could be adjusted to accept both a boolean value and a string, but I'm not sure that's such a good idea.

### DIFF
--- a/bodhi-client/bodhi/client/bindings.py
+++ b/bodhi-client/bodhi/client/bindings.py
@@ -440,6 +440,7 @@ class BodhiClient:
             gating (str): filter by TestGatingStatus description.
             from_side_tag (bool): A boolean to filter updates created from side tag or from
                 normal workflow.
+            from_side_tag_name (str): Query for updates created from a certain side tag.
             user (str): Query for updates submitted by a specific user.
             rows_per_page (int): Limit the results to a certain number of rows per page
                 (min:1 max: 100 default: 20).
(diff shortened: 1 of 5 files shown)


mattiaverga:
> In order not to break existing API, this commit adds a new field, `from_side_tag_name`, that allows to query for updates created from the specified side tag.
> 
> Alternatively, the existing `from_side_tag` field could be adjusted to accept both a boolean value and a string, but I'm not sure that's such a good idea.

I agree, let's have the two query fields separate.

mattiaverga (changes requested):
Can you please add a [feature](https://bodhi.readthedocs.io/en/25.11.3/developer/index.html#release-notes) release note to announce the new option?

nforro:
> Can you please add a [feature](https://bodhi.readthedocs.io/en/25.11.3/developer/index.html#release-notes) release note to announce the new option?

Done.

mattiaverga (approved):
Thanks!

PR #6094, opened 2026-03-25
Update page: Rebalance main columns

ferdnyc:
The two columns of the update page were sized `col-md-9` and `col-md-3`
(a 75/25 split), which made for ugly wrapping as the right column
squashed too narrow, at lower page widths.

To compensate, make the left column `col-md-8 col-lg-9`, and the right
column `col-md-4 col-lg-3`. That way, the right column will be given
slightly more of the total space when the page width is in the 'medium'
range (but go back to 25% once the width is 'large' or greater).

All of these screenshots are taken at the full 1920px screen width, for comparison purposes.

## Before, medium-breakpoint width

Screenshot From 2026-03-25 04-32-45

## After, medium-breakpoint width

This is the only size actually affected by this change, and gives the right-hand column a bit more room to avoid the ugly wrapping (particularly the "Dates" section) seen in the previous image.

Screenshot From 2026-03-25 04-35-28

## Before, other sizes

Screenshot From 2026-03-25 04-32-25
Screenshot From 2026-03-25 04-32-34
Screenshot From 2026-03-25 04-33-04

## After, other sizes (no meaningful differences)

Screenshot From 2026-03-25 04-35-07
Screenshot From 2026-03-25 04-35-18
Screenshot From 2026-03-25 04-35-46

### DIFF
--- a/bodhi-server/bodhi/server/templates/update.html
+++ b/bodhi-server/bodhi/server/templates/update.html
@@ -164,7 +164,7 @@ elif can_edit and update.status.value == 'testing' and update.test_gating_status
   <div class="tab-content" id="updateTabContent">
     <div class="tab-pane show active" id="details-tab-pane" role="tabpanel" aria-labelledby="details-tab" tabindex="0">
       <div class="row">
-        <div class="col-md-9">
+        <div class="col-md-8 col-l-9">
 
           % if update.notes:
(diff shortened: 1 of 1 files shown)


mattiaverga (approved):
Thanks, it looks nicer indeed.

PR #6099, opened 2026-04-02
Use dogpile cache everywhere

### LINKED ISSUE #6098: current cache implementation is not thread safe
Bodhi is currently using two different caching method in its code, dogpile and lru_cache.
On production, we use running web server with multiple workers/multiple threads, but lru_cache is not shared between workers and neither is dogpile.cache.memory_pickle, which currently used.

We should port all things using cache to dogpile and use dogpile.redis (or just used redis).

mattiaverga:


mattiaverga:
I've deployed the change on the staging instance some time ago and it seems to work... so I'm going to merge the PR.

PR #6102, opened 2026-04-13
Fail `download-build` if any koji download failed

LecrisUT:


AdamWill:
I think this is too naive. Consider the case of a multi-package update containing a package that exists only on certain arches. I think with this change, when trying to download that update for another arch, we would always fail.

See [the code in openQA's downloader](https://forge.fedoraproject.org/quality/os-autoinst-distri-fedora/src/commit/a5823d99bd55d649b3f10707c96d90184ea49e04/data/setup_repos.py#L121-L124) that handles this situation and doesn't treat it as a 'failure'. I think we'd at least need that here too.

I was going to suggest making this an option as well, not just changing the behavior unilaterally.

Issue #6128, opened 2026-06-19
[Bug] Some search queries return identical/duplicate results.


beedell-roke:
Within the structure that [`issues/6127#issue-4704244527`](https://github.com/fedora-infra/bodhi/issues/6127#issue-4704244527) describes, I can observe duplicate results:

<blockquote>

~~~HTML
<li class="typeahead__item typeahead__group-packages" data-group="packages" data-index="0"><a href="updates/?packages=thunderbird"><strong>thunderbird</strong></a></li>
<li class="typeahead__item typeahead__group-packages" data-group="packages" data-index="1"><a href="updates/?packages=thunderbird"><strong>thunderbird</strong></a></li>
~~~

</blockquote>

#### My Environment

> `bodhi-server` `26.4.0`, on `bodhi-web-79764b86cc-6m5q2`

AshwinAashu:
I see that we are getting the package search response as mentioned below. Noticed that while there are two "thunderbird" , they differ by type. There is also a `thunderbird-flatpak `. Wondering how it differs from the other flatpak type thunderbird. 

 
```
 "packages": [
        {
            "name": "thunderbird",
            "type": "flatpak"
        },
        {
            "name": "thunderbird",
            "type": "rpm"
        },
        {
            "name": "thunderbird-flatpak",
            "type": "flatpak"
        }
    ],
```

The approach to fix that I am thinking of is to fetch distinct packages only at search query  and search count level. 


```python
  if search is not None:
        query = (
            query.distinct(Package.name)
                 .order_by(Package.name, case((Package.name == search, Package.name)))
        )

``` 

The another approach could be to show the type on the frontend. However, since we are navigating based on package name alone ( and therefore will land on the same page when clicked on either of the two results ) , first approach feels to have more merits to me.

mattiaverga:
I'm not sure the solution is to deduplicate query result by package name. The query result has all the information to distinguish between package types and there might be use cases where we want the complete list returned.

The solution would be to deduplicate results in javascript, however the documentation of the component used in the search bar is [no more available](http://www.runningcoder.org/jquerytypeahead/).
I think the easiest way is to modify the template used to show results so that a label with the package type is showed, e.g.:
```
display: ['name', 'type'],
template: '<span class="badge text-bg-light border">{{type}}</span>&nbsp;{{name}}',
```
 and add the type to the URL filter by `content_type`:
```
href: function (item) {
[...]

PR #6130, opened 2026-06-28
Fix duplicate package names in search results

AshwinAashu:
Problem Statement  : 

This PR aims to solve the [issue](https://github.com/fedora-infra/bodhi/issues/6128).

RCA : 

The packages have two properties : name and type. Since we are only showing package names to the user and are navigating based on the package name too , it can lead to confusion ( specially for new users ). 

Changes made : 
Grouped the packages based on name and showing based on the package id. This ensures that a single package is shown to the user if there are multiple packages with exact same name.

### DIFF
--- a/bodhi-server/bodhi/server/services/packages.py
+++ b/bodhi-server/bodhi/server/services/packages.py
@@ -74,7 +74,6 @@ def query_packages(request):
     search = data.get('search')
     if search is not None:
         query = query.filter(Package.name.ilike('%%%s%%' % search))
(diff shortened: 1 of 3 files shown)


AshwinAashu:
Hi @mattiaverga , looks like the integration test and this tests have been failing for some time now and its unrelated to these changes. I  see there has been some update and the last few recent runs have worked after the fix was merged.   Can we run the tests again ?  

Update : rebased.

AshwinAashu:
As per discussion , will make the change at the template level first and then will look at how the component can be improved.

Issue #5935, opened 2025-09-03
bodhi composer runs out of open files after some time


nirik:
Active: active (running) since Sat 2025-08-30 03:37:42 UTC; 4 days ago

```
: ERROR/ForkPoolWorker-16] Task clean_old_composes[937e35d9-e594-45d1-b90c-6e041231955d] raised unexpected: OSError(24, 'Too many open files')
Sep 03 03:03:00 bodhi-backend01.rdu3.fedoraproject.org celery-3[1169191]: Traceback (most recent call last):
Sep 03 03:03:00 bodhi-backend01.rdu3.fedoraproject.org celery-3[1169191]:   File "/usr/lib/python3.13/site-packages/celery/app/trace.py", line 477, in trace_task
Sep 03 03:03:00 bodhi-backend01.rdu3.fedoraproject.org celery-3[1169191]:     R = retval = fun(*args, **kwargs)
Sep 03 03:03:00 bodhi-backend01.rdu3.fedoraproject.org celery-3[1169191]:                  ~~~^^^^^^^^^^^^^^^^^
Sep 03 03:03:00 bodhi-backend01.rdu3.fedoraproject.org celery-3[1169191]:   File "/usr/lib/python3.13/site-packages/celery/app/trace.py", line 760, in __protected_call__
Sep 03 03:03:00 bodhi-backend01.rdu3.fedoraproject.org celery-3[1169191]:     return self.run(*args, **kwargs)
Sep 03 03:03:00 bodhi-backend01.rdu3.fedoraproject.org celery-3[1169191]:            ~~~~~~~~^^^^^^^^^^^^^^^^^
Sep 03 03:03:00 bodhi-backend01.rdu3.fedoraproject.org celery-3[1169191]:   File "/usr/lib/python3.13/site-packages/bodhi/server/tasks/__init__.py", line 123, in clean_old_composes_task
Sep 03 03:03:00 bodhi-backend01.rdu3.fedoraproject.org celery-3[1169191]:     main(num_to_keep)
Sep 03 03:03:00 bodhi-backend01.rdu3.fedoraproject.org celery-3[1169191]:     ~~~~^^^^^^^^^^^^^
Sep 03 03:03:00 bodhi-backend01.rdu3.fedoraproject.org celery-3[1169191]:   File "/usr/lib/python3.13/site-packages/bodhi/server/tasks/clean_old_composes.py", line 68, in main
[...]

PR #6155, opened 2026-08-20
Don't leak Pungi's pipes when a compose is aborted

bsmojver:
PungiComposerThread._compose_updates() started Pungi with stdout and stderr set to PIPE, then ran generate_testing_digest() and _generate_updateinfo() before reaching _wait_for_pungi(), which is what actually drains and closes those pipes.

Both of those intermediate calls talk to Koji and can raise -- in particular get_rpm_header() re-raises after three failed attempts -- so a slow or flaking Koji leaves _wait_for_pungi() unreached and the Popen object abandoned. CPython does not close the pipes in that situation: Popen.__del__ appends a still-running child to subprocess._active to keep it alive until it can be waited on, and that reference holds both file descriptors open. A garbage collection does not reclaim them.

If Pungi has written more than the 64kB pipe buffer by that point it is blocked in write() with nobody left to read, so it never exits, never gets reaped out of subprocess._active, and the descriptors are held for the remaining lifetime of the Celery worker. Composes that fail this way succeed on a retry, which is why this shows up as slow descriptor exhaustion rather than a reproducible failure.

Wrap the body of _compose_updates() in try/finally and add _abandon_pungi(), which terminates an unreaped Pungi, drains it with a timeout so a process stuck on a full pipe is unblocked and reaped, escalates to kill() if it ignores SIGTERM, and closes the streams. pungi_process is cleared after a successful _wait_for_pungi() so the cleanup is a no-op on the normal path.

Related to #5935

Assisted-by: Claude (Anthropic)

### DIFF
--- a/bodhi-server/bodhi/server/tasks/composer.py
+++ b/bodhi-server/bodhi/server/tasks/composer.py
@@ -897,6 +897,8 @@ class PungiComposerThread(ComposerThread):
     """Compose update with Pungi."""
 
     pungi_template_config_key = None
+    #: Seconds to wait for an aborted Pungi to exit after SIGTERM before killing it.
+    pungi_abort_timeout = 60
 
     def __init__(self, max_concur_sem, compose, agent, db_factory, compose_dir, resume=False):
         """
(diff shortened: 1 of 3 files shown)


decathorpe:
Hm ... but, let's assume there are no failed composes ... then there are no leaked fds, and this error condition never happens? But composes *do* start to fail every few days *without* any failed pungi runs preceding them, from what I can tell. So I'm not sure whether this actually matches the issue we're seeing in bodhi.fp.o.

bsmojver:
> Hm ... but, let's assume there are no failed composes ... then there are no leaked fds, and this error condition never happens? But composes _do_ start to fail every few days _without_ any failed pungi runs preceding them, from what I can tell. So I'm not sure whether this actually matches the issue we're seeing in bodhi.fp.o.

Analysing again...

bsmojver:
@decathorpe was this ever tried as a workaround?
```
# in /etc/bodhi/celeryconfig.py
worker_max_tasks_per_child = 200
```

mattiaverga:
I think this is useful, nonetheless. Thanks.

mattiaverga (changes requested):
There are some unit test failures (apart from the broken Rawhide).

bsmojver:
> There are some unit test failures (apart from the broken Rawhide).

Thanks, I'll follow up on it.

bsmojver:
> There are some unit test failures (apart from the broken Rawhide).

Thanks — fixed in the commit just pushed.

The three failures were all in TestPungiComposerThread__abandon_pungi. Those tests built the Pungi handle as a bare MagicMock(), so stdout.closed and stderr.closed auto-created truthy child mocks. The if stream is not None and not stream.closed guard in _abandon_pungi() therefore concluded the pipes were already closed and skipped close(), so close.assert_called_once_with() failed with "Called 0 times."

The production code was correct; only the mocks were wrong. Setting closed = False explicitly in the three tests matches what test_streams_already_closed already does for the True case.

tests/tasks/test_composer.py passes locally now.

mattiaverga (approved):
Thanks!

PR #6158, opened 2026-08-25
Opt into the hard file limit in the web server and celery tasks

jeremycline:
Resources have soft and hard limits. Processes are allowed to raise their soft limit up to the hard limit, but if they choose to not do that, the soft limit is the "real" limit.

The hard file limit in Fedora is ~500K, but for compatibility with old programs that use things like select() (which don't work with file descriptors above 1024) the soft limit is 1024.

This adjusts the Bodhi services to bump the soft limit to the hard limit. It's possible I missed some process entry-points.

See also: #5935

### DIFF
--- a/bodhi-server/bodhi/server/__init__.py
+++ b/bodhi-server/bodhi/server/__init__.py
@@ -19,6 +19,7 @@
 from collections import defaultdict
 import importlib.metadata
 import logging as python_logging
+import resource
 
 from cornice.validators import DEFAULT_FILTERS
 from dogpile.cache import make_region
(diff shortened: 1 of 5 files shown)


mattiaverga (changes requested):
Thanks, looks good.
Can you add a release note? https://bodhi.readthedocs.io/en/26.4.0/developer/index.html#release-notes

jeremycline:
> Thanks, looks good. Can you add a release note? https://bodhi.readthedocs.io/en/26.4.0/developer/index.html#release-notes

Done. I associated it with the PR since the issue has discussion of other things and I didn't actually test that this fixes the issue, just that it raises the limit.

mattiaverga (approved):
LGTM

PR #6160, opened 2026-08-30
Don't let one failed compose block every other release

bsmojver:
bodhi-push treats the presence of any Compose row as evidence that a push is already in flight:

    if not resume and session.query(Compose).count():
        ...
        resume = True
        resume_all = True

The count is unfiltered, so a single Compose left behind by a failure turns every later invocation into a resume of that compose and nothing else. The queries that find updates with a pending request live in the else branch and never run. A Compose row only survives when the compose did not finish, since remove_state() is the last statement of the success path, so the row persists for as long as the underlying problem does.

The practical effect is that one stuck lane, an F44 Flatpak compose say, stops updates for every release and every content type until someone clears the row by hand.

Nothing in the composer requires this. ComposerHandler.run() already starts one thread per Compose under a BoundedSemaphore and each thread catches its own exceptions, and the Compose primary key is (release_id, request), so lanes are independent by construction. The serialisation exists only in push.py's gate.

Collect resumable Composes and newly created ones in the same pass rather than choosing between them. A Compose being resumed keeps ownership of its (release, request) lane; every other lane is composed as usual. Updates belonging to an occupied lane are skipped before from_updates() sees them, which avoids a collision on the compound primary key and, more subtly, avoids quietly attaching a new update to a Compose that has already checkpointed past
determine_and_perform_tag_actions(). Such an update would never be tagged in Koji but would be treated as composed.

An explicit --resume is left alone. The additive behaviour applies only where resume_all was inferred from existing Composes.

Notes and caveats:

- The compose task message carries a single resume flag for the whole batch, and mixed batches are now possible. This is safe: Compose.checkpoints defaults to '{}', so load_state() on a freshly created Compose reads an empty dict and the checkpoint decorator skips nothing, and run() calls save_state(ComposeState.initializing) before work() regardless of the flag. The only visible effect is a misleading "Resuming push without any completed repos" log line for new Composes. No api_version bump is needed.

- Concurrency is unchanged. This still sends one compose task, so there is still one ComposerHandler and one semaphore, and max_concurrent_composes is still respected.

- The resume pass continues to resume every Compose regardless of --releases, while the new pass honours it. That asymmetry predates this change but becomes easier to notice: bodhi-push --releases F44 with a stuck F44F Compose will now resume F44F and create F44. It is not a regression, and fixing it seemed out of scope here.

- The first push after this is deployed will be large, since it flushes whatever accumulated behind the stuck lane. Better done deliberately than discovered from cron.

- Failed Composes are not aged out or retried differently. A failure is reported and resumed on the next run exactly as before. Resume semantics are untouched.

- Fedora's bodhi-automated-pushes.py in the infrastructure ansible repo needs a matching change to stop gating on /composes/ being empty. That change is only correct once this is deployed to the backends.

Adds test_stuck_compose_does_not_block_other_releases, covering a failed Compose on one release alongside a pending update on another and asserting that unlocked updates in the occupied lane stay out of it. Two existing expected-output fixtures gain the skip notice.

Assisted-by: Claude Opus 5 (Anthropic)

### DIFF
--- a/bodhi-server/bodhi/server/push.py
+++ b/bodhi-server/bodhi/server/push.py
@@ -108,6 +108,7 @@ def push(username, yes, **kwargs):
             resume_all = True
 
         # If we're resuming a push
+        existing_lanes = set()
         if resume:
             for compose in session.query(Compose).all():
                 if len(compose.updates) == 0:
@@ -131,7 +132,17 @@ def push(username, yes, **kwargs):
                 compose.error_message = ''
 
                 composes.append(compose)
-        else:
+                existing_lanes.add((compose.release_id, compose.request))
+
+            # Flush so that any Composes deleted above are removed from the database before we
+            # potentially create new Composes with the same compound primary key below.
+            session.flush()
+
+        # Queue up any other releases that do not already have a Compose in flight. A Compose that
+        # is being resumed owns its (release, request) lane until it succeeds, but it should not
+        # prevent unrelated lanes from being composed. An explicit --resume is left alone, since
+        # the operator asked for a resume and nothing else.
+        if not resume or resume_all:
             updates = []
             # Accept both comma and space separated request list
             requests = kwargs['request'].replace(',', ' ').split(' ')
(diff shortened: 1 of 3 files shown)


mattiaverga:
Can you add a release note to the commit? https://bodhi.readthedocs.io/en/26.4.0/developer/index.html#release-notes

bsmojver:
> Can you add a release note to the commit? https://bodhi.readthedocs.io/en/26.4.0/developer/index.html#release-notes

Just pushed that in the revised commit. Thank you for reviewing!

mattiaverga (approved):

PR #6163, opened 2026-08-31
Fix duplicate package names in search results

kadubhumika:
Package search showed identical names for different package types.
Added package type to the search result display.
Added content_type to the result URL.


mattiaverga:
Thanks, but this was already WIP by AshwinAashu in #6165

Issue #6127, opened 2026-06-19
[Bug] The search result sheet hyperlinks its headings to nothing.


beedell-roke:
When I search, [^1] I observe:

[^1]: [`discussion.fedoraproject.org/t/194042/8`](https://discussion.fedoraproject.org/t/i-want-to-upgrade-to-f45-for-newer-thunderbird-why-is-attempting-to-returning-tb-dependency-resolution-errors/194042/8)

<blockquote>

~~~HTML
<ul class="typeahead__list">
	<li class="typeahead__group" data-search-group="packages"><a href="javascript:;" tabindex="-1">
			<h3>packages</h3>
		</a></li>
~~~

</blockquote>

It does nothing, when selected:

<p></p><img width="638" height="695" alt="Image" src="https://github.com/user-attachments/assets/b87d6dc7-2d16-4334-842b-e4913be12aba" /><p></p>

#### My Environment

> `bodhi-server` `26.4.0`, on `bodhi-web-79764b86cc-6m5q2`

PR #6172, opened 2026-09-12
Fix search group hyperlink

kadubhumika:
## Fix search result group headings linking to nothing (#6127)

Issue: https://github.com/fedora-infra/bodhi/issues/6127

### Problem

The search result group headings were rendered inside an `<a href="javascript:;">` element. The heading appeared to be a link, but selecting it did not perform any action.

### Fix

* Updated `groupTemplate` in `search.js` to render the group heading as a `<div>` containing an `<h3>`, instead of wrapping it in a non-functional hyperlink.
* Updated the related CSS selectors in `site.css` to target the new `<h3>` structure directly.
* This keeps the group heading semantic and non-clickable while preserving the existing styling.

### Technical change

Before:

```html
<a href="javascript:;">
    <h3>packages</h3>
</a>
```

After:

```html
<div class="typeahead__group-header">
    <h3>packages</h3>
</div>
```

### Result

Search result group headings are now displayed as headings rather than misleading hyperlinks, so clicking/selecting the heading no longer appears to trigger a link that does nothing.

### DIFF
--- a/bodhi-server/bodhi/server/static/css/site.css
+++ b/bodhi-server/bodhi/server/static/css/site.css
@@ -75,12 +75,12 @@ pre {
     border: none;
 }
 
-#bodhi-searchbar .typeahead__group > a {
+#bodhi-searchbar .typeahead__group {
     background: none;
     padding: 0 .5rem;
 }
 
-#bodhi-searchbar .typeahead__group > a > h3 {
+#bodhi-searchbar .typeahead__group > h3 {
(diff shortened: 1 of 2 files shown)


kadubhumika:
Hi maintainers, just a gentle follow-up on this PR. The merge conflicts have been resolved and the branch has been updated. Could someone please review it when you get a chance? Thank you!

PR #6175, opened 2026-09-19
perf: reduce queries when loading update test cases

kadubhumika:
## Problem

Large updates can contain thousands of builds. `Update.full_test_cases`
accesses `Build.testcases` for every build, causing an N+1 query pattern.

For a BCD update with 3,387 builds, this resulted in 3,387 separate
queries to `build_testcase_table`.

## Changes

- Use SQLAlchemy `selectin` loading for `Build.testcases` to batch-load
  test cases instead of querying once per build.
- Add an index on `build_testcase_table.build_id`.
- Add an Alembic migration for the index.

## Benchmark

Tested with BCD update `FEDORA-2026-52976614ef` (224 test cases).

| | Before | After |
|---|---:|---:|
| SQL queries | 3,388 | 8 |
| Testcase queries | 3,387 | 7 |
| Time | ~16.71s | ~0.72s |

The main improvement comes from eliminating the N+1 query pattern.

`git diff --check` passes.
## Investigation

Traced the path through `validate_update_id()` → `Update.get()` →
`Update.full_test_cases()` and instrumented SQLAlchemy queries. This
confirmed that `Build.testcases` was being loaded separately for each
build.


kadubhumika:
Hi! I noticed the Read the Docs check is currently failing during the bodhi-server dependency installation, before the documentation build starts.
The error appears to be related to the zstandard version constraint in the package metadata. This PR only changes the test-case loading behavior and adds the related database index/migration, so I haven't changed the dependency configuration.
Just mentioning this for context while the PR is being reviewed. Thanks!

